PR #7506, opened 2023-08-10
feat(insights): add event to trigger computation

jakubdyszkiewicz:
### Checklist prior to review

Add an event to do on-demand computation of insights.

### DIFF
--- a/pkg/insights/resyncer.go
+++ b/pkg/insights/resyncer.go
@@ -227,17 +227,7 @@ func (r *resyncer) Start(stop <-chan struct{}) error {
 					log.Error(err, "could not get tenants")
 				}
 				for _, tenantId := range tenantIds {
-					meshList := &core_mesh.MeshResourceList{}
-					tenantCtx := multitenant.WithTenant(tickCtx, tenantId)
-					if err := r.rm.List(tenantCtx, meshList); err != nil {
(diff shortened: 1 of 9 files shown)


pkg/insights/resyncer.go
```diff
@@ -249,38 +239,58 @@ func (r *resyncer) Start(stop <-chan struct{}) error {
...
+	meshList := &core_mesh.MeshResourceList{}
+	tenantCtx := multitenant.WithTenant(ctx, tenantID)
+	if err := r.rm.List(tenantCtx, meshList); err != nil {
+		log.Error(err, "failed to get list of meshes", "tenantId", tenantCtx)
+		return
+	}
+	for _, mesh := range meshList.Items {
+		batch.add(time.Now(), tenantID, mesh.GetMeta().GetName(), FlagMesh|FlagService)
+	}
+	if err := batch.flush(tenantCtx, resyncEvents); err != nil {
```
lukidzi, line 289: Should it be here? I've noticed that it is duplicated in few times. In case of full resync we are going to resync after each tenantId, and after going through all the tenants

jakubdyszkiewicz: It shouldn't! Thanks for catching it

lukidzi (approved):

PR #7507, opened 2023-08-10
refactor(xds/generator): move route handling to package, use resource map in egress

michaelbeaumont:
Should not be any behavioral changes in this PR

### Checklist prior to review

### DIFF
--- a/api/common/v1alpha1/ref.go
+++ b/api/common/v1alpha1/ref.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 
 	"golang.org/x/exp/maps"
+
+	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
 )
 
 type TargetRefKind string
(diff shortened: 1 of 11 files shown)


api/common/v1alpha1/ref.go
```diff
@@ -61,6 +63,29 @@ func (in TargetRef) Hash() TargetRefHash {
 	return TargetRefHash(fmt.Sprintf("%s/%s/%s/%s", in.Kind, in.Name, strings.Join(orderedTags, "/"), in.Mesh))
 }
 
+func (in TargetRef) EnvoyTags() (map[string]string, bool) {
```
lobkovilya, line 66: To be honest, I don't see much value in making this a method. TargetRef is a core API object, it doesn't have to "know" how to represent itself in terms of Envoy tags. 

I'd just create a method somewhere under `pkg/xds/`:
```go
func TargetRefToEnvoyTags(tr TargetRef) (map[string]string, bool)
```

pkg/core/xds/types.go
```diff
@@ -177,7 +177,7 @@ type ExternalServiceDynamicPolicies map[ServiceName]PluginOriginatedPolicies
 
 type MeshResources struct {
 	Mesh                           *core_mesh.MeshResource
-	TrafficRoutes                  []*core_mesh.TrafficRouteResource
+	Resources                      map[core_model.ResourceType]core_model.ResourceList
```
lobkovilya, line 180: For new plugin-originated policy types, we already have a field `MeshResources.Dynamic`. So I'd keep this field as it is and took MeshHTTPRoute policies from the `Dynamic`.

lobkovilya: You have to implement `EgressMatchedPolicies` method on MeshHTTPRoute plugin so the policies are present in `Dynamic`

api/common/v1alpha1/ref.go
michaelbeaumont: I think maybe including "Envoy" was unnecessary. There's definitely something inherent to the semantics of the API object here, especially since `MeshServiceSubset` _includes_ tags. This is really just one direction of the isomorphism between `map[string]string`, where `kuma.io/service` is a tag, and `TargetRef`.

lobkovilya: > I think maybe including "Envoy" was unnecessary. 

How would you call the method then?

pkg/core/xds/types.go
lobkovilya: Right, `Dynamic` is for ExternalServices. In that case, I'd either rename `Resources` to `Routes` or simply create the second field:
```go
TrafficRoutes  []*core_mesh.TrafficRouteResource
MeshHTTPRoutes []*MeshHTTPRouteResource
```

michaelbeaumont: Added second field

api/common/v1alpha1/ref.go
michaelbeaumont: Moved somewhere else

PR #7519, opened 2023-08-15
feat(kuma-cp): use hash-suffix for KDS sync

lobkovilya:
Depends on #7711 

### Checklist prior to review


pkg/kds/context/context_test.go
```diff
@@ -518,7 +520,7 @@ var _ = Describe("Context", func() {
 						k8sSystemNamespace: "custom-namespace",
 					},
 					name:         "foo.custom-namespace",
-					expectedName: "foo.custom-namespace",
+					expectedName: "foo-xw46fx28d8666769.custom-namespace",
```
lahabana, line 523: Should this hash be shorter? Feels like this detail should have been ironed out in the MADR.

pkg/kds/hash/hash.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"hash/fnv"
+
+	"k8s.io/apimachinery/pkg/util/rand"
+	k8s_strings "k8s.io/utils/strings"
+
+	util_k8s "github.com/kumahq/kuma/pkg/util/k8s"
+)
+
+func ZoneName(mesh, name string) string {
```
jakubdyszkiewicz, line 14: this name is confusing. It took me a while to figure out that zone name = the name after syncing the resource to the zone.

Correct me if I'm wrong, but I think eventually we also want to use suffixes when syncing stuff from zone to global.

Maybe the `SyncedName` name would be better?

lobkovilya: When syncing from zone to global, we're going to use another function (like GlobalName, or better named) and this function is going to have a different signature:
```go
func GlobalName(zone, mesh, name string) string
```
and probably we'll need the same but with namespace:
```go
func GlobalNameNamespaced(ns, zone, mesh, name string) string
```

Alternatively, we can go with `SyncedName` as you suggested, but implement a builder pattern:
```go
SyncedName(WithZone(zone), WithMesh(mesh), ...)
```

TBH I think different functions for each situation are better and more straightforward, but maybe I should name it `NameInZone` or something like that.

jakubdyszkiewicz: `NameInZone` still does not help to indicate that it is a name in the zone _after_ the resource is synced.

We could do builder or we could just do `SyncedName(ns, zone, mesh, name string)` and pass empty strings. I'm ok with both

lobkovilya: What about:
```go
func SyncedNameInZone(mesh, name string) string
func SyncedNameInGlobal(zone, mesh, name string) string
func SyncedNameInGlobalNamespaced(ns, zone, mesh, name string) string
```
I really like the fact you don't have to check the implementation to understand what parameters have to be passed for the function to work correctly

jakubdyszkiewicz: works for me. I'd go with
```
func SyncedNameInGlobal(ns, zone, mesh, name string) string
```
soon all resources will have namespace.

jakubdyszkiewicz (approved):

Issue #5368, opened 2022-11-24
Add possibility to expose metrics endpoints with TLS


lahabana:
### Description

From @mmorel-35 

We deploy our prometheus outside of the mesh as the second option here ([Observability](https://kuma.io/docs/1.8.x/explore/observability/#configuring-prometheus)) is recommended, so we have skipMTLS: true

Prometheus then retrieves HTTP URL for the dataplanes metrics endpoints which is not allowed in our company.

The option skipMTLS disables the mtls provided by the mesh but I would like to be able to provide my own server certificate so the metrics endpoint is secured when it is accessed outside of the mesh.
In my opinion, the tls configuration shall be provided at the level of the dataplane config file and not at the level of the metrics policy.

This way, the metrics endpoint will be secured but you'll also need to ensure that the mads service of the control plane retrieves now HTTPS URL if the dataplane has TLS config while HTTP when it didn't.

lahabana:
@mmorel-35 some random thoughts:

Wouldn't it be easier to reference a cert in the Kuma secrets in the metrics definition? We could then provide this cert to all dataplanes and it would make cert rotation a little easier, I guess SAN verification would be tricky here.

Is it also imaginable to use a provided CA and expose the endpoint through TLS (but not mTLS) so the cert would be rooted on the provided CA and prometheus would be able to verify it?

mmorel-35:
It is completely imaginable that the tls config could be at the level of the metrics definition, for example, there could be something like that definition in the dataplane config file : 
```yaml
metrics:
  name: 
  type: ‘prometheus’
  conf:
  tls:
    enabled: true
    certFile:  ""
    keyFile: ""
    minVersion: "TLSv1_2"
    cipherSuites: []
```

So when mtls is disabled, the dataplane exposes it's metrics in a secured way and also the mesh is not responsible of the certificate rotation.


At the level of the mesh, I'm not sure if how it could look like ?  And the first could be done without this :

```yaml 
metrics: 
  enabledBackend: prometheus
  backends: 
    - name: prometheus
      type: prometheus
      conf: 
        path: /metrics
        port: 5670
        skipMTLS: true
        tls:
          ca: "" # Root CA ?
        tags: 
          kuma.io/service: dataplane-metrics
````

jakubdyszkiewicz:
Hey, 2 questions from me
1) How will Prometheus verify the cert used to protect the metrics endpoint? I'm talking about SAN verification. What is the expected SAN from this cert?
[...]

PR #7534, opened 2023-08-16
feat(kuma-cp): implement TLS listener for prometheus 

lukidzi:
### Checklist prior to review
* allow configuring TLS certificates for each data plane through passing path to cert in the env
* introduce new TLS model for the configuration

When certificates are not provided should we do not generate listener or do fallback to no tls one?

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: #5368 docs: https://github.com/kumahq/kuma-website/pull/1438


lukidzi:
@mmorel-35 Hi. Could you take a look if that works for you?

lahabana (commented):
I'm a little concerned with skipTLS and backward compat. To me we'd need to do double writes for 2 versions before not setting `skipTLS`

api/mesh/v1alpha1/metrics.proto
```diff
@@ -90,3 +93,25 @@ message PrometheusEnvoyConfig {
   // least once). If nil, then it is treated as false.
   google.protobuf.BoolValue usedOnly = 2;
 }
+
+// PrometheusEnvoyConfig defines Tls configuration for Prometheus listener.
+message PrometheusTlsConfig {
+  enum Mode {
+    // This is a substitute of `skipMTLS` set to false. In this situation,
+    // control-plane delivers certificates to the prometheus listener.
```
lahabana, line 101:
```suggestion
    // control-plane delivers certificates to the prometheus listener.
    // This should be used when prometheus is running inside the Mesh.
```

api/mesh/v1alpha1/metrics.proto
```diff
@@ -90,3 +93,25 @@ message PrometheusEnvoyConfig {
...
+    // In this way user is resposible for providing certificates to dataplanes.
+    // Path for the certificte and the key needs to be provided to the dataplane
+    // by environments variables:
+    // * KUMA_DATAPLANE_METRICS_CERT_PATH
+    // * KUMA_DATAPLANE_METRICS_KEY_PATH
+    delegated = 1;
+  }
+  // If true then endpoints for scraping metrics will require TLS, when false
+  // there if no tls configured. If nil, then it is treated as true.
+  google.protobuf.BoolValue enabled = 1;
```
lahabana, line 112: Instead of having this should we have a mode that is `disabled` and avoid having 2 params?

lahabana:
Can we have the docs PR too? :)

lukidzi:
> I'm a little concerned with skipTLS and backward compat. To me we'd need to do double writes for 2 versions before not setting `skipTLS`

Do you mean that someone does a rollback and there is no more skipMTLS? We could preserve that parameter and in case of rollback, we will still have it. I was translating `skipMTLS` to `tls` but it seems it is necessary to keep it.

api/mesh/v1alpha1/metrics.proto
lukidzi: Seems good, `activeMTLSBackend` could be the default.

lahabana:
Yes my understanding is that we should keep both for a bit. @jakubdyszkiewicz WDYT?

test/e2e_env/kubernetes/observability/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+	})
+	E2EAfterAll(func() {
+		Expect(kubernetes.Cluster.TriggerDeleteNamespace(namespace)).To(Succeed())
+		Expect(kubernetes.Cluster.TriggerDeleteNamespace(namespaceNoSidecar)).To(Succeed())
+		Expect(kubernetes.Cluster.DeleteMesh(mesh)).To(Succeed())
+	})
+
+	It("should scrape metrics defined in mesh and not fail when defined service doesn't exist", func() {
+		// given
+		time.Sleep(3600 * time.Hour)
```
jakubdyszkiewicz, line 94: ?

test/e2e_env/kubernetes/observability/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+		podIp, err := PodIPOfApp(kubernetes.Cluster, "test-server", namespace)
+		Expect(err).ToNot(HaveOccurred())
+
+		// when
+		stdout, _, err := client.CollectResponse(
+			kubernetes.Cluster, "test-server", "http://"+net.JoinHostPort(podIp, "1234")+"/metrics",
```
jakubdyszkiewicz, line 100: shouldn't this be HTTPS?

test/e2e_env/universal/observability/metrics.go
```diff
@@ -0,0 +1,110 @@
...
+		// given
+		host := universal.Cluster.GetApp("test-server").GetIP()
+
+		// when doing http request
+		_, _, err := client.CollectResponse(
+			universal.Cluster,
+			clientName,
+			"http://test-server.mesh:1234/metrics",
+			client.Resolve("test-server.mesh:1234", fmt.Sprintf("[%s]", host)),
+		)
```
jakubdyszkiewicz, line 73: why is it possible to still do HTTP? 🤔

test/framework/interface.go
```diff
@@ -517,6 +518,12 @@ func WithTransparentProxyV1(transparent bool) AppDeploymentOption {
 	})
 }
 
+func WithDpEnvs(envs map[string]string) AppDeploymentOption {
```
jakubdyszkiewicz, line 521: nit: maybe just `WithEnvs`?

jakubdyszkiewicz:
Yes, of course. SkipMTLS has to stay for backwards compat

test/e2e_env/kubernetes/observability/metrics.go
lukidzi: My fault, i moved the test to universal because `kuma-dp` image had no certificates. I forgot to remove this file.

test/e2e_env/universal/observability/metrics.go
lukidzi: Client can try to do the request but it fails with HTTP, I am just checking here if it fails `Expect(err).To(HaveOccurred())`

test/framework/interface.go
lukidzi: The file `interface.go` has `WithEnvs` defined already for `KumaDeploymentOption` and because both are in the same file we cannot define the one for app. Maybe we should split the file `interface` into one for app other for kuma deployment.

jakubdyszkiewicz: oh, ok :/

test/e2e_env/universal/observability/metrics.go
jakubdyszkiewicz: ah, I see

lukidzi:
> I'm a little concerned with skipTLS and backward compat. To me we'd need to do double writes for 2 versions before not setting `skipTLS`

Docs PR: https://github.com/kumahq/kuma-website/pull/1438

pkg/config/app/kuma-dp/config.go
```diff
@@ -142,6 +142,10 @@ type DataplaneRuntime struct {
 	Resources DataplaneResources `json:"resources,omitempty"`
 	// SocketDir dir to store socket used between Envoy and the dp process
 	SocketDir string `json:"socketDir,omitempty" envconfig:"kuma_dataplane_runtime_socket_dir"`
+	// MetricsCertPath path to the certificate for metrics listener
+	MetricsCertPath string `json:"metricsCertPath,omitempty" envconfig:"kuma_dataplane_metrics_cert_path"`
+	// MetricsKeyPath path to the key for metrics listener
+	MetricsKeyPath string `json:"metricsKeyPath,omitempty" envconfig:"kuma_dataplane_metrics_key_path"`
```
mmorel-35, line 148: Would you consider grouping Metrics in a dedicated Object with the prefixed properties embedded ?

pkg/core/xds/metadata.go
```diff
@@ -58,6 +60,8 @@ type DataplaneMetadata struct {
 	SocketDir           string
 	AccessLogSocketPath string
 	MetricsSocketPath   string
+	MetricsCertPath     string
+	MetricsKeyPath      string
```
mmorel-35, line 64: Same suggestion here to have all Metrics properties in the same object ?

mmorel-35 (commented):
Hi @lukidzi ,

This seems to fit what is expected.
Just a suggestion in my comments.
Thank you !

jakubdyszkiewicz (approved):

api/mesh/v1alpha1/metrics.proto
```diff
@@ -90,3 +93,25 @@ message PrometheusEnvoyConfig {
...
+    // control-plane delivers certificates to the prometheus listener.
+    // This should be used when prometheus is running inside the Mesh.
+    activeMTLSBackend = 0;
+    // In this way user is resposible for providing certificates to dataplanes.
+    // Path for the certificte and the key needs to be provided to the dataplane
+    // by environments variables:
+    // * KUMA_DATAPLANE_RUNTIME_METRICS_CERT_PATH
+    // * KUMA_DATAPLANE_RUNTIME_METRICS_KEY_PATH
+    delegated = 1;
+    // allows disabling mTLS for the prometheus listener.
```
lahabana, line 109:
```suggestion
    // allows disabling TLS for the prometheus listener.
```

api/mesh/v1alpha1/metrics.proto
```diff
@@ -90,3 +93,25 @@ message PrometheusEnvoyConfig {
...
+  enum Mode {
+    // control-plane delivers certificates to the prometheus listener.
+    // This should be used when prometheus is running inside the Mesh.
+    activeMTLSBackend = 0;
+    // In this way user is resposible for providing certificates to dataplanes.
+    // Path for the certificte and the key needs to be provided to the dataplane
+    // by environments variables:
+    // * KUMA_DATAPLANE_RUNTIME_METRICS_CERT_PATH
+    // * KUMA_DATAPLANE_RUNTIME_METRICS_KEY_PATH
+    delegated = 1;
```
lahabana, line 108: Is this maybe delegatedTLS (the active backend does mtls, this is a just server side TLS).

lukidzi: changed to `providedTLS`

Issue #7527, opened 2023-08-16
Cross zone virtual outbound with zone egress


jakubdyszkiewicz:
### What happened?

Virtual outbound cross zone should also work with zone egress

Issue #5871, opened 2023-01-30
MeshHTTPRoute cross zone support


lahabana:
### Description

TrafficRoute was working cross zone. MeshHTTPRoute should also support this.

Should we let all route work cross zone by default or not?

lahabana:
xref #5869

lobkovilya:
We need ZoneEgress support

PR #7536, opened 2023-08-17
fix(egress): routing using MeshHTTPRoute and VirtualOutbound

jakubdyszkiewicz:
Fix https://github.com/kumahq/kuma/issues/5871
Fix https://github.com/kumahq/kuma/issues/7527

### The first problem

When we do cross-zone routing using TrafficRoute, MeshHTTPRoute, VirtualOutbound etc. we configure ZoneIngress with FilterChainMatch based on SNI to route the traffic to the proper destination without TLS termination. We implemented `ingress_generator_destination.go` to generate all possible destinations from the policies.

When we add ZoneEgress to this picture, ZoneEgress passes the request to ZoneIngress without TLS termination as well. It has to be configured in the same way as ZoneIngress, but with endpoints of ZoneIngresses instead of actual services.

As we were adding new ways of routing, we were adding this to ZoneIngress (in `ingress_generator_destination.go`) without adding it to ZoneEgress.

### The solution to the first problem

I refactored the ZoneEgress code, because I noticed that a lot of it has to be almost the same between ZoneIngress. I created `zoneproxy` package where I moved code to build destinations and to build clusters, endpoints, and filter chains. This way we remove code duplication and we avoid the problem of not configuring ZoneEgress and ZoneIngress in the same way.

### The second problem

Instead of creating new clusters for every split, Zone proxies are using LB subset.
For example, there is `service:a` with `version:1` and `version:2`
ZoneIngress has a cluster `service:a` with endpoint and metadata `version:1` and `version:2` and splits based on this metadata. This works well for ZoneIngress, but if we do this for ZoneEgress we very easily end up in a situation where both endpoints for `version:1` and `version:2` point to the same ZoneIngress. Envoy then does the endpoint deduplication based on IP:port and we end up only with one endpoint instead of two.

### The solution to the second problem

When we build the LB subset, we need to check if selecting a subset of endpoints actually selects different endpoints. If not, it means that this subset is useless. For example (from Egress perspective):
Ingress1 (10.0.0.1:1000) supports `service:a,version:1` and `service:a,version:2`
Ingress2 (10.0.0.2:1000) supports `service:a,version:1` and `service:a,version:2`
and we have a destination of `service:a,version:1`. It means that we don't have to do subset because both ingresses support `service:a,version:1`.
However, ZoneIngress still needs to do routing to a `version:1` which has a different endpoint than `version:2`. All we have to do is send proper SNI regardless if we place LB subset or not. 
The good news here is that ZoneEgress passes SNI through to ZoneIngress.
This solution is implemented in `AddFilterChains#relevantTags`

### Tests

The problems existed because we did not have good enough coverage of ZoneEgress.
We _could_ add yet another entry in a test matrix to test multizone suite with and without `Mesh#zoneEgress:true` on the mesh. However, this adds CI runs which are already expensive.
The alternative is to run routing-related tests with `Mesh#zoneEgress:true/false`. Those tests should work exactly the same with or without egress. I implemented it for MeshHTTPRoute and VirtualOutbound.
I think we need a bit more coverage for E2E tests, but I want to add this as a follow-up to not expand already big PR.

Additionally, I ported Mike's MeshHTTPRoute [unit tests](https://github.com/kumahq/kuma/pull/7516/files#diff-4d4f81af0b5a2475d51c6419038dd97c09da96343763f7c08d34614f129fc3a6R1) to check if this solution works.

This PR is a superset of https://github.com/kumahq/kuma/pull/7516

### Backport

Should we backport this? Technically it's a fix, but it's pretty big change.


pkg/xds/generator/zoneproxy/destinations.go
```diff
@@ -1,45 +1,32 @@
...
-		dest[meshResources.Mesh.GetMeta().GetName()] = destForMesh
-	}
-	return dest
+func BuildMeshDestinations(
+	availableServices []*mesh_proto.ZoneIngress_AvailableService, // available services for a single mesh
+	res xds_context.Resources,
+) map[string][]envoy_tags.Tags {
+	destForMesh := map[string][]envoy_tags.Tags{}
+	meshHTTPRoutes := res.ListOrEmpty(meshhttproute_api.MeshHTTPRouteType).(*meshhttproute_api.MeshHTTPRouteResourceList).Items
+	addTrafficRouteDestinations(res.TrafficRoutes().Items, destForMesh)
```
lahabana, line 24: It's easier to read if meshHTTPRoutes is defined after this line (I struggled to see where it's actually used)

pkg/xds/generator/zoneproxy/generator.go
```diff
@@ -0,0 +1,151 @@
...
+			// relevantTags is a set of tags for which it actually makes sense to do LB split on.
+			// If the endpoint list is the same with or without the tag, we should just not do the split.
+			// However, we should preserve full SNI, because the client expects Zone Proxy to support it.
+			// This solves the problem that Envoy deduplicate endpoints of the same address and different metadata.
+			relevantTags := envoy_tags.Tags{}
+			for key, value := range destination {
+				matchedTargets := map[string]struct{}{}
+				allTargets := map[string]struct{}{}
+				for _, endpoint := range serviceEndpoints {
+					address := fmt.Sprintf("%s:%d", endpoint.Target, endpoint.Port)
```
lahabana, line 121: nit: maybe make this a method on `endpoint`

pkg/xds/generator/zoneproxy/generator.go
```diff
@@ -0,0 +1,151 @@
...
+			cluster := envoy_common.NewCluster(
+				envoy_common.WithName(clusterName),
+				envoy_common.WithTags(relevantTags),
+			)
+
+			filterChain := envoy_listeners.FilterChain(
+				envoy_listeners.NewFilterChainBuilder(apiVersion, envoy_common.AnonymousResource).Configure(
+					envoy_listeners.MatchTransportProtocol("tls"),
+					envoy_listeners.MatchServerNames(sni),
+					envoy_listeners.TcpProxyDeprecatedWithMetadata(
```
lahabana, line 141: This is deprecated? Why are we still using it? Looking at the code it doesn't look like there's much info on the deprecation. @michaelbeaumont can you explain?

test/e2e_env/multizone/meshhttproute/test.go
```diff
@@ -8,17 +8,27 @@ import (
...
 
 func Test() {
-	meshName := "meshhttproute"
+	_ = Describe("No Zone Egress", func() {
+		test("meshhttproute", samples.MeshMTLSBuilder())
+	}, Ordered)
 
+	_ = Describe("Zone Egress", func() {
+		test("meshhttproute-ze", samples.MeshMTLSBuilder().WithEgressRoutingEnabled())
+	}, Ordered)
```
lahabana, line 25: DescribeTable doesn't work because of the multiple "It" ?

lahabana (commented):
Seems to make sense to me but it's a hard code change. I want to read it again later :)

pkg/xds/generator/zoneproxy/generator.go
michaelbeaumont: It's because I [introduced Splits to unify MeshHTTP- and MeshTCPRoute](https://github.com/kumahq/kuma/pull/6943), but to rewrite everything would have been a much larger change. To be honest I forgot about including `// deprecated:` at the time.

test/e2e_env/multizone/meshhttproute/test.go
jakubdyszkiewicz: Yes, and also BeforeAll etc.

pkg/xds/generator/zoneproxy/generator.go
jakubdyszkiewicz: I've got import cycles. This requires moving some files around. I'd like to avoid it for this PR.

pkg/xds/generator/egress/internal_services_generator.go
```diff
@@ -26,29 +26,28 @@ func (g *InternalServicesGenerator) Generate(
...
-	g.addFilterChains(
-		apiVersion,
-		destinations,
-		proxy,
-		listenerBuilder,
-		meshResources,
-		services,
+	servicesMap := g.buildServices(meshResources.EndpointMap, meshResources.Mesh.ZoneEgressEnabled(), xdsCtx.ControlPlane.Zone)
+	services := maps.Keys(servicesMap)
+	sort.Strings(services)
```
lobkovilya, line 33: You can probably use https://github.com/kumahq/kuma/blob/master/pkg/util/maps/sorted_keys.go and adjust SortedKeys method to be generic

jakubdyszkiewicz: Not sure if it's worth having this util to save one line, but I added it.

michaelbeaumont (approved):

lobkovilya (approved):

PR #7542, opened 2023-08-21
ci(envoy): extract everything from the archive not just envoy binary

slonka:
### Checklist prior to review

This allows us to have multiple files in the envoy tar, so if we want to bundle it with something it will work.


jakubdyszkiewicz:
what other files do we have?

jakubdyszkiewicz (approved):

slonka:
talked on priv with Jakub

PR #7560, opened 2023-08-22
fix(api-server): return 400 when PUT/POST resource is invalid

lahabana:
We were returning 500 whenever the resource was invalid. We now return an error that is correctly handled in handle_error.

We also removed an extra json unmarshaling that was used to retrieve the type descriptor. In practice in the endpoint we already have the type descriptor.

* also add ctx info to log messages in error_handler
* output the result when using the golden matcher with a file that doesn't exist (useful when adding new tests).
* remove experimental from MeshGateway and MeshGatewayRoute

### Checklist prior to review


slonka (approved):
LGTM

api/mesh/v1alpha1/gateway.proto
```diff
@@ -29,7 +29,6 @@ message MeshGateway {
   option (kuma.mesh.resource).name = "MeshGatewayResource";
   option (kuma.mesh.resource).type = "MeshGateway";
   option (kuma.mesh.resource).package = "mesh";
-  option (kuma.mesh.resource).is_experimental = true;
```
michaelbeaumont, line 32: What's the effect of this, why is it in this PR?

slonka: I did not notice this, good catch - @lahabana ?

PR #7561, opened 2023-08-22
feat(api-server): add isTargetRefBased in /policies

### LINKED ISSUE #7485: Add `isNewStylePolicy` field to `/policies`
### Description

Add a `isNewStylePolicy` field to `/policies` indicating whether a policy type is of a new style. We specifically want this for https://github.com/kumahq/kuma-gui/issues/1290 meaning what we want to be able to do is know whether a policy of a certain policy type will have a top-level target ref object.

lahabana:
This is useful in the GUI

fix #7485

### Checklist prior to review


lahabana:
@kleinfreund FYI

pkg/api-server/policies_endpoints.go
```diff
@@ -24,6 +24,9 @@ func addPoliciesWsEndpoints(ws *restful.WebService, mode core.CpMode, readOnly b
 				SingularDisplayName: def.SingularDisplayName,
 				PluralDisplayName:   def.PluralDisplayName,
 				IsExperimental:      def.IsExperimental,
+				IsTargetRefBased:    def.IsTargetRefBased,
+				IsInbound:           def.HasFromTargetRef,
```
michaelbeaumont, line 28: Why the different name?

kleinfreund: I guess that’s related to what we want to do with this field: show whether a policy type is an inbound and/or outbound one. And I guess this can come from a few different sources one of them being whether is has a from target ref?

Automaat (approved):

jakubdyszkiewicz (approved):

PR #7565, opened 2023-08-22
fix(dataplane_watchdog): fix outdated comment

nicoche:
I think that they do use Mesh Hash now, cf. lines 178 and 226


jakubdyszkiewicz (approved):
true! thx

Issue #2483, opened 2021-08-03
Add ability to wait for sidecar container


michaelkoro:
### Summary

When deploying Kong on top of EKS into a kuma-based namespace, we noticed that the db migration job that is executed before the pods are started, is sometimes stuck with the following error - 

```
Error: [PostgreSQL error] failed to retrieve server_version_num: host or service not provided, or not known
```
After trying a few things (changing the db address, upgrading docker version) we noticed that the envoy proxy sometimes is started after the application container (in this case, the migrations container) which I guess causes network errors.

Once we deployed the chart in a namespaces that wasn't managed by Kuma, everything worked fine.
Is there a way to tell kuma to first start the envoy proxy, and only then the application itself ?

Thanks

Kuma Chart version - 0.6.0
EKS - 1.19
Kong - 2.1.4

### Additional Details & Logs
link to a related ticket in kong - https://github.com/Kong/kong/issues/4363

michaelkoro:
Did anyone encounter this kind of issue ?

jpeach:
xref https://github.com/kubernetes/kubernetes/issues/65502

skaravad:
The problem is, unless kuma DP is up and running the pod has no network, and as per K8s it appears that sidecar lifecycle is more complicated than it was thought and it is a waiting game. 

On the other hand, if you can wrap the main application command or entrypoint, you can use this logic (install netcat in ubuntu or debian, alpine has nc command by default installed)

```
## Check Network when Service Mesh is enabled
while true
do
  nc -vz www.google.com 443
  ret_code=$?
  if [ $ret_code  -ne 0 ] ; then
    echo "Network Not ready"
    sleep 3
  else
    echo "Network Ready"
  break
  fi
done

echo "starting {{.Chart.Name}} service"
MAIN COMMAND
```
Btw, if you have vault integration and you have a init container which runs , it will not init , to overcome , just add this annotation 
`vault.hashicorp.com/agent-init-first: "true"`

michaelkoro:
@skaravad I remember when working with Istio that they managed to solve the issue.
I think when deploying Istio you had to add a flag which basically tells the app container to wait for the proxy.
Are you familiar with that ? Is there a way to implement this solution in kuma as well ?

jpeach:
xref #2571

skaravad:
@michaelkoro with ISTIO it was a annotation https://github.com/istio/istio/issues/11130 
```
annotations:
[...]

PR #7583, opened 2023-08-23
feat(kuma-cp): block application container start until dp is ready

lukidzi:
### Checklist prior to review

Kubernetes creates containers sequentially and if we add `postStart` to the first container it blocks others from starting until the first is ready. This is undocumented behavior and might be changed in the future, when sidecar containers are going to be production ready.

https://github.com/kubernetes/kubernetes/pull/82282
https://github.com/kubernetes/kubernetes/issues/81450
https://github.com/kumahq/kuma/issues/2483

### DIFF
--- a/app/kuma-dp/cmd/wait.go
+++ b/app/kuma-dp/cmd/wait.go
@@ -20,13 +20,13 @@ func newWaitCmd() *cobra.Command {
 	}{
 		url:            "http://localhost:9901/ready",
 		requestTimeout: 500 * time.Millisecond,
-		timeout:        60 * time.Second,
+		timeout:        180 * time.Second,
 		checkFrequency: 1 * time.Second,
 	}
 	cmd := &cobra.Command{
 		Use:   "wait",
-		Short: "Waits for Dataplane to be ready",
-		Long:  `Waits for Dataplane (Envoy) to be ready.`,
+		Short: "Waits for data plane proxy to be ready",
+		Long:  `Waits for data plane proxy (Envoy) to be ready.`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			client := &http.Client{
 				Timeout: args.requestTimeout,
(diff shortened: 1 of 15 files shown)


pkg/core/resources/apis/mesh/zz_generated.resources.go
```diff
@@ -973,7 +973,7 @@ var MeshGatewayResourceTypeDescriptor = model.ResourceTypeDescriptor{
 	IsPolicy:            true,
 	SingularDisplayName: "Mesh Gateway",
 	PluralDisplayName:   "Mesh Gateways",
-	IsExperimental:      false,
+	IsExperimental:      true,
```
lukidzi, line 976: not sure why it changed the value

app/kuma-dp/cmd/wait.go
```diff
@@ -0,0 +1,80 @@
...
+var waitLog = dataplaneLog.WithName("wait")
+
+func newWaitCmd() *cobra.Command {
+	args := struct {
+		url            string
+		requestTimeout time.Duration
+		timeout        time.Duration
+		checkFrequency time.Duration
+	}{
+		url:            "http://localhost:9901/ready",
```
lukidzi, line 21: Maybe instead of envoy ready we should check listener that is generated by dp?

lukidzi: Or maybe we should add some additional function that checks envoy's listener and if there is specific one/or more than one

jakubdyszkiewicz: the admin listener with stats is generated at pod IP so I presume this might be a problem. What we can do however is check for listeners in stats to see if we received things already.

test/e2e_env/kubernetes/graceful/wait_for_envoy.go
```diff
@@ -0,0 +1,70 @@
...
+    - --
+    - 'while true; do curl --fail test-server_wait-for-envoy_svc_80.mesh && echo "succeeded" || { echo "failed" ; exit 1; }; done'
+    resources:
+      limits:
+        cpu: 50m
+        memory: 64Mi`, id, namespace, meshName, waitForDataplane, framework.Config.GetUniversalImage()))
+	}
+
+	It("remove Dataplane of evicted Pod", func() {
+		// TODO: flaky because config is delivered too fast, how to slow it down?
```
jakubdyszkiewicz, line 63: I'd just remove this case if we cannot make it reliable. I think it's fine to only test if we always succeed with the wait for dataplane ready

app/kuma-dp/cmd/wait.go
```diff
@@ -0,0 +1,80 @@
...
+					return fmt.Errorf("timeout occurred while waiting for Dataplane to be ready")
+				}
+			}
+		},
+	}
+
+	cmd.PersistentFlags().DurationVar(&args.checkFrequency, "check-frequency", args.checkFrequency, `frequency of checking if the Dataplane is ready`)
+	cmd.PersistentFlags().DurationVar(&args.timeout, "timeout", args.timeout, `timeout defines how long waits for the Dataplane`)
+	cmd.PersistentFlags().DurationVar(&args.requestTimeout, "request-timeout", args.requestTimeout, `requestTimeout defines timeout for the request to the Dataplane`)
+	cmd.PersistentFlags().StringVar(&args.url, "url", args.url, `url at which admin is exposed`)
```
lahabana, line 57: nit: Usually defaults are set inside these calls rather than pre-init on the object

pkg/plugins/runtime/k8s/containers/factory.go
```diff
@@ -69,6 +73,15 @@ func (i *DataplaneProxyFactory) envoyAdminPort(annotations map[string]string) (u
 	return adminPort, err
 }
 
+func (i *DataplaneProxyFactory) waitForDataplaneOrDefault(annotations map[string]string, defaultWaitForDataplane bool) (bool, error) {
+	enabled, exists, err := metadata.Annotations(annotations).GetEnabled(metadata.KumaWaitForDataplaneReady)
```
lahabana, line 77: There's GetEnabledWithDefault for this

test/e2e_env/kubernetes/graceful/wait_for_envoy.go
lahabana: Yeah I'd remove the false case.

app/kuma-dp/cmd/wait.go
jakubdyszkiewicz: what do you mean?

lahabana: I feel like is more common no?
```
	cmd.PersistentFlags().StringVar(&args.url, "url", args.url, <myDefault>, `url at which admin is exposed`)
```

app/kuma-dp/cmd/wait.go
jakubdyszkiewicz: Ok so I run multiple tests and it seems that /ready returns 200 after we've got all listeners. I first added this
```
req, err = http.NewRequest(http.MethodGet, url+"/stats?filter=listener_manager.lds.version_text", nil)
if err != nil {
	return err
}
resp, err = client.Do(req)
if err != nil {
	return err
}
defer resp.Body.Close()
bytes, err := io.ReadAll(resp.Body)
if err != nil {
	return err
}
if resp.StatusCode != 200 {
	return fmt.Errorf("HTTP status code %v", resp.StatusCode)
}
if string(bytes) == "listener_manager.lds.version_text: \"\"\n" {
	return errors.New("envoy not configured")
}
return nil
```
but it seems pointless. I'd go with just `/ready` and we can see if the test will be flaky. If so we can improve the check further

app/kuma-dp/cmd/wait.go
jakubdyszkiewicz: I think we use both, but ok - changed.

lahabana (approved):
Lgtm but I think we're missing the config in the defaults file (useful for docs)

PR #7587, opened 2023-08-24
feat(e2e): fix metrics tls test

Automaat:
### Checklist prior to review

### DIFF
--- a/pkg/api-server/auth_test.go
+++ b/pkg/api-server/auth_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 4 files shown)


pkg/api-server/auth_test.go
```diff
@@ -85,7 +86,7 @@ var _ = Describe("Auth test", func() {
 
 	It("should be able to access admin endpoints using client certs and HTTPS", func() {
 		// when
-		resp, err := httpsClient.Get(fmt.Sprintf("https://%s:%d/secrets", externalIP, httpsPort))
+		resp, err := httpsClient.Get(fmt.Sprintf("https://%s/secrets", net.JoinHostPort(externalIP, strconv.Itoa(int(httpsPort)))))
```
slonka, line 89: did it fail on ipv6?

slonka (approved):

pkg/api-server/auth_test.go
Automaat: yes

jakubdyszkiewicz (approved):

PR #7593, opened 2023-08-24
fix(kuma-cp): universal mode don't log on every lock acquire attempt

michaelbeaumont:
They chose kind of a weird interface for the logging, no option to add keys, just a list of values. So I added the arg number as a key, as opposed to just concatenating everything.

### Checklist prior to review


bartsmykla (commented):
It looks good to me, but what I don't get is, why it was `Println` and now is `Error` + `Debug`. Could you give me some more context about it?

michaelbeaumont:
[the upstream commit](https://github.com/cirello-io/pglock/commit/b2e4951c2f8fbd93bcbd5b7e669aa0ff3d1d2637)

Automaat (approved):

PR #7602, opened 2023-08-28
feat(lifecycle): make graceful shutdown duration configurable

slonka:
I suspect that during graceful shutdown 3s is not enough to gracefully shutdown grpc streams. This PR makes it configurable and adds a log line to see if shutdown actually compoleted

### Checklist prior to review


pkg/config/app/kuma-cp/config.go
```diff
@@ -362,10 +364,11 @@ func (g *GeneralConfig) Validate() error {
...
 	return &GeneralConfig{
-		DNSCacheTTL:     config_types.Duration{Duration: 10 * time.Second},
-		WorkDir:         "",
-		TlsCipherSuites: []string{},
-		TlsMinVersion:   "TLSv1_2",
+		DNSCacheTTL:              config_types.Duration{Duration: 10 * time.Second},
+		WorkDir:                  "",
+		TlsCipherSuites:          []string{},
+		TlsMinVersion:            "TLSv1_2",
+		GracefulShutdownDuration: config_types.Duration{Duration: 3 * time.Second},
```
lahabana, line 371: Should we increase the default while we're at it?

slonka:
Talked to Jakub and he suggested adding preStop hook, I'll check that out first.

slonka:
Closing for now, because adding a preStop hooked helped a bit.

PR #7624, opened 2023-08-31
feat(kds): experimental event based watchdog

jakubdyszkiewicz:
### Checklist prior to review

A test of experimental event-based watchdog for KDS.
I tested it manually with E2E test and it looks ok. We want to deploy this on our dev environment to see the perf gain. This might be a default, but for now we want to put it behind a flag.

The idea is instead of querying DB resources every X interval we react on postgres events and only generate snapshot for changed resources. Every Y interval we do a full resync if for some reason we lose the event.

### DIFF
--- a/pkg/kds/v2/reconcile/reconciler.go
+++ b/pkg/kds/v2/reconcile/reconciler.go
@@ -8,6 +8,7 @@ import (
 	envoy_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_types "github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	envoy_cache "github.com/envoyproxy/go-control-plane/pkg/cache/v3"
+	"golang.org/x/exp/maps"
 	"google.golang.org/protobuf/proto"
 
 	config_core "github.com/kumahq/kuma/pkg/config/core"
@@ -64,19 +65,37 @@ func (r *reconciler) Clear(ctx context.Context, node *envoy_core.Node) error {
 	return nil
 }
 
-func (r *reconciler) Reconcile(ctx context.Context, node *envoy_core.Node) error {
-	new, err := r.generator.GenerateSnapshot(ctx, node)
+func (r *reconciler) Reconcile(ctx context.Context, node *envoy_core.Node, changedTypes map[core_model.ResourceType]struct{}) error {
+	id, err := r.hashId(ctx, node)
+	if err != nil {
+		return err
+	}
+	old, _ := r.cache.GetSnapshot(id)
+
+	// construct builder with unchanged types from the old snapshot
+	builder := cache_v2.NewSnapshotBuilder()
+	if old != nil {
+		for _, typ := range util_kds_v2.GetSupportedTypes() {
+			resType := core_model.ResourceType(typ)
+			if _, ok := changedTypes[resType]; ok {
+				continue
+			}
+
+			oldRes := old.GetResources(typ)
+			if len(oldRes) > 0 {
+				builder = builder.With(resType, maps.Values(oldRes))
+			}
+		}
+	}
+
+	new, err := r.generator.GenerateSnapshot(ctx, node, builder, changedTypes)
 	if err != nil {
 		return err
 	}
 	if new == nil {
 		return errors.New("nil snapshot")
 	}
-	id, err := r.hashId(ctx, node)
-	if err != nil {
-		return err
-	}
-	old, _ := r.cache.GetSnapshot(id)
+
 	new = r.Version(new, old)
 	r.logChanges(new, old, node)
 	r.meterConfigReadyForDelivery(new, old)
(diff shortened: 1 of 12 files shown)


pkg/kds/v2/server/event_based_watchdog.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/kumahq/kuma/pkg/events"
+	"github.com/kumahq/kuma/pkg/kds/v2/reconcile"
+	"github.com/kumahq/kuma/pkg/multitenant"
+	util_watchdog "github.com/kumahq/kuma/pkg/util/watchdog"
+)
+
+type EventBasedWatchdog struct {
+	Ctx                  context.Context
+	Node                 *envoy_core.Node
+	StreamID             int64
```
bartsmykla, line 22: Why do we need `StreamID` here? I don't see it being used anywhere

pkg/kds/v2/reconcile/snapshot_generator.go
```diff
@@ -40,9 +40,8 @@ type snapshotGenerator struct {
 	resourceMapper  reconcile.ResourceMapper
 }
 
-func (s *snapshotGenerator) GenerateSnapshot(ctx context.Context, node *envoy_core.Node) (envoy_cache.ResourceSnapshot, error) {
-	builder := cache_kds_v2.NewSnapshotBuilder()
-	for _, typ := range s.resourceTypes {
+func (s *snapshotGenerator) GenerateSnapshot(ctx context.Context, node *envoy_core.Node, builder cache_kds_v2.SnapshotBuilder, resTypes map[model.ResourceType]struct{}) (envoy_cache.ResourceSnapshot, error) {
```
bartsmykla, line 43: Nit: Could we split the params into separate lines? It starts to be hard to read so long line, especially during the code review
image

bartsmykla (approved):
Few nits, but lgtm overall

pkg/kds/v2/server/event_based_watchdog.go
lukidzi: Removed

pkg/kds/v2/server/components.go
```diff
@@ -87,8 +97,26 @@ func newSyncTracker(log logr.Logger, reconciler reconcile_v2.Reconciler, refresh
 	if err := metrics.Register(kdsGenerationsErrors); err != nil {
 		return nil, err
 	}
+	changedTypes := map[model.ResourceType]struct{}{}
+	for _, typ := range providedTypes {
+		changedTypes[typ] = struct{}{}
+	}
 	return util_xds_v3.NewWatchdogCallbacks(func(ctx context.Context, node *envoy_core.Node, streamID int64) (util_watchdog.Watchdog, error) {
-		log := log.WithValues("streamID", streamID, "node", node)
```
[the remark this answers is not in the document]
lukidzi: I think it should stay here, reverted

pkg/kds/v2/reconcile/reconciler.go
```diff
@@ -64,19 +65,37 @@ func (r *reconciler) Clear(ctx context.Context, node *envoy_core.Node) error {
...
 		return err
 	}
-	if new == nil {
-		return errors.New("nil snapshot")
+	old, _ := r.cache.GetSnapshot(id)
+
+	// construct builder with unchanged types from the old snapshot
+	builder := cache_v2.NewSnapshotBuilder()
+	if old != nil {
+		for _, typ := range util_kds_v2.GetSupportedTypes() {
```
lobkovilya, line 78: `GetSupportTypes` returns all types we sync (both Zone->Global and Global->Zone), but we build snapshots with fewer types (i.e. `reg.ObjectTypes(model.HasKDSFlag(model.ProvidedByGlobal))`). This probably won't result in a bug because you're checking `len(oldRes)` but I'd fix it to avoid potential bugs in the future

pkg/kds/v2/server/components.go
```diff
@@ -87,8 +97,26 @@ func newSyncTracker(log logr.Logger, reconciler reconcile_v2.Reconciler, refresh
...
+	for _, typ := range providedTypes {
+		changedTypes[typ] = struct{}{}
+	}
 	return util_xds_v3.NewWatchdogCallbacks(func(ctx context.Context, node *envoy_core.Node, streamID int64) (util_watchdog.Watchdog, error) {
 		log := log.WithValues("streamID", streamID, "node", node)
+		if experimentalWatchdogCfg.Enabled {
+			return &EventBasedWatchdog{
+				Ctx:                  ctx,
+				Node:                 node,
+				Listener:             eventBus.Subscribe(),
```
lobkovilya, line 110: The listener returned from `Subscribe` should be closed somewhere.

pkg/kds/v2/reconcile/reconciler.go
jakubdyszkiewicz: Yeah, this is intentional because like you said, we just won't set those resources. We do the same with in version function

jakubdyszkiewicz: I'm merging as is. If you feel this is important, please create an issue. Thx!

Issue #7649, opened 2023-09-05
Optimize KDS V2 resource syncer


jakubdyszkiewicz:
### Description

Instead of calling
```
if err := s.resourceStore.List(ctx, downstream); err != nil {
```
we should do
```
keys := resourcekeys(ResourceAdded) + ResourceRemoved
if err := s.resourceStore.List(ctx, downstream, ListByKeys(keys); err != nil {
```

This filter does not exist. We need to implement this in paginated resource store so it's by default available for all store and additionally in postgres we need to support native query.

On the InitialRequest we don't want to set this filter.

PR #7654, opened 2023-09-05
feat(kuma-cp): pass resource keys to postgres query for delta kds

lukidzi:
### Checklist prior to review

Added option to list only elements from the database that have changed so we don't need to retrieve all the data each time. We can configure it and by default, it's turned off.

- [x] [Link to relevant issue][1] as well as docs and UI issues -- fix: #7649 

### DIFF
--- a/pkg/core/resources/store/options.go
+++ b/pkg/core/resources/store/options.go
@@ -161,7 +161,7 @@ type ListOptions struct {
 	FilterFunc   ListFilterFunc
 	NameContains string
 	Ordered      bool
-	ResourceKeys []core_model.ResourceKey
+	ResourceKeys map[core_model.ResourceKey]struct{}
 }
 
 type ListOptionsFunc func(*ListOptions)
@@ -216,7 +216,11 @@ func ListOrdered() ListOptionsFunc {
 
 func ListByResourceKeys(rk []core_model.ResourceKey) ListOptionsFunc {
 	return func(opts *ListOptions) {
-		opts.ResourceKeys = rk
+		resourcesKeys := map[core_model.ResourceKey]struct{}{}
+		for _, val := range rk {
+			resourcesKeys[val] = struct{}{}
+		}
+		opts.ResourceKeys = resourcesKeys
 	}
 }
 
--- a/pkg/core/resources/store/options.go
+++ b/pkg/core/resources/store/options.go
@@ -161,6 +161,7 @@ type ListOptions struct {
 	FilterFunc   ListFilterFunc
 	NameContains string
 	Ordered      bool
+	ResourceKeys []core_model.ResourceKey
 }
 
 type ListOptionsFunc func(*ListOptions)
@@ -210,6 +211,12 @@ func ListByFilterFunc(filterFunc ListFilterFunc) ListOptionsFunc {
 func ListOrdered() ListOptionsFunc {
 	return func(opts *ListOptions) {
 		opts.Ordered = true
+	}
+}
+
+func ListByResourceKeys(rk []core_model.ResourceKey) ListOptionsFunc {
+	return func(opts *ListOptions) {
+		opts.ResourceKeys = rk
 	}
 }
 
--- a/pkg/core/resources/store/pagination_store.go
+++ b/pkg/core/resources/store/pagination_store.go
@@ -74,11 +74,11 @@ func (p *paginationStore) List(ctx context.Context, list model.ResourceList, opt
 				}
 			}
 		}
-	}
-
-	for _, item := range fullList.GetItems() {
-		if opts.Filter(item) {
-			_ = filteredList.AddItem(item)
+	} else {
+		for _, item := range fullList.GetItems() {
+			if opts.Filter(item) {
+				_ = filteredList.AddItem(item)
+			}
 		}
 	}
 
--- a/pkg/core/resources/store/pagination_store.go
+++ b/pkg/core/resources/store/pagination_store.go
@@ -67,11 +67,9 @@ func (p *paginationStore) List(ctx context.Context, list model.ResourceList, opt
 	}
 
 	if len(opts.ResourceKeys) > 0 {
-		for _, rk := range opts.ResourceKeys {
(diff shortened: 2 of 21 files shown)


pkg/kds/v2/store/sync.go
```diff
@@ -310,3 +319,11 @@ func GlobalSyncCallback(
...
 	}
 }
+
+func ToResourceKeys(rs core_model.ResourceList) []core_model.ResourceKey {
+	rkey := []core_model.ResourceKey{}
+	for _, r := range rs.GetItems() {
+		rkey = append(rkey, core_model.MetaToResourceKey(r.GetMeta()))
+	}
+	return rkey
+}
```
lahabana, line 329: feels like this function is generally useful no? Maybe it should be moved in `core_model` ? Maybe even a method of `ResourceList` (`GetItemsAsResourceKey()`)? Or `MetaListToResourceKeys(core_model.ResourceList)` to mirror what already exists

pkg/plugins/resources/postgres/pgx_store.go
```diff
@@ -183,15 +183,28 @@ func (r *pgxResourceStore) List(ctx context.Context, resources core_model.Resour
...
+		statement += " AND ("
+		for idx, rk := range opts.ResourceKeys {
+			if idx > 0 {
+				statement += " OR "
+			}
+			statement += fmt.Sprintf("(mesh=$%d AND name=$%d)", argsIndex+1, argsIndex+2)
+			argsIndex += 2
+			statementArgs = append(statementArgs, rk.Mesh, rk.Name)
+		}
+		statement += ")"
```
lahabana, line 196: Should we be using `IN()` instead? avoids these complex string concat

lahabana: `(mesh=X AND name IN(....)) OR (mesh=Y AND name IN(...))`

pkg/kds/v2/store/sync.go
```diff
@@ -86,8 +88,15 @@ func (s *syncResourceStore) Sync(syncCtx context.Context, upstreamResponse clien
 	if err != nil {
 		return err
 	}
-	if err := s.resourceStore.List(ctx, downstream); err != nil {
-		return err
+	if len(upstream.GetItems()) >= s.maxListQueryElements || upstreamResponse.IsInitialRequest {
```
lahabana, line 91: What is the reason for doing this type of limit? I'm wondering if this should be a parameter of the store instead of KDS.

pkg/plugins/resources/postgres/pgx_store.go
```diff
@@ -183,15 +187,42 @@ func (r *pgxResourceStore) List(ctx context.Context, resources core_model.Resour
...
+				} else {
+					statement += fmt.Sprintf(",$%d", argsIndex)
+				}
+				statementArgs = append(statementArgs, name)
+			}
+			statement += "))"
+			iter++
+		}
+		statement += ")"
+	} else {
```
jakubdyszkiewicz, line 215: I think this should be outside of else, no?

pkg/core/resources/store/pagination_store.go
```diff
@@ -66,9 +66,19 @@ func (p *paginationStore) List(ctx context.Context, list model.ResourceList, opt
 		return err
 	}
 
-	for _, item := range fullList.GetItems() {
-		if opts.Filter(item) {
-			_ = filteredList.AddItem(item)
+	if len(opts.ResourceKeys) > 0 {
+		for _, rk := range opts.ResourceKeys {
+			for _, item := range fullList.GetItems() {
```
jakubdyszkiewicz, line 71: I'm slightly concerned about O(n^2) here, but I guess we can leave it and see on the flamegraphs if this is visible or not.

pkg/core/resources/store/pagination_store.go
```diff
@@ -66,9 +66,19 @@ func (p *paginationStore) List(ctx context.Context, list model.ResourceList, opt
...
+				if item.GetMeta().GetMesh() == rk.Mesh && item.GetMeta().GetName() == rk.Name {
+					_ = filteredList.AddItem(item)
+				}
+			}
+		}
+	} else {
```
jakubdyszkiewicz, line 77: same, I think this should be outside of else

pkg/core/resources/store/pagination_store.go
lahabana: Right switch the loop would likely be more friendly to CPU caches. But doesn't it make sense for resourceKeys to be set and have this in the existing for loop?

pkg/core/resources/store/options.go
```diff
@@ -161,6 +161,7 @@ type ListOptions struct {
 	FilterFunc   ListFilterFunc
 	NameContains string
 	Ordered      bool
+	ResourceKeys []core_model.ResourceKey
```
lahabana, line 164: I'd go with a `map[core_model.ResourceKey]struct{}` to use this as a set in other places.

pkg/plugins/resources/postgres/pgx_store.go
lukidzi: I am not sure if that makes sense. ResourceKey requires name and mesh to be set. It would filter this twice.

lukidzi: It might not make huge sense to create that query but it can be possible

pkg/core/resources/store/pagination_store.go
jakubdyszkiewicz: it seems the solution would be to do
```
for _, item := range fullList.GetItems() {
  _, exists := opts.ResourceKeys[model.MetaToResourceKey(item.GetMeta())]
  if len(opts.ResourceKeys) > 0 && !exists {
    continue
  }
  if !opts.Filter(item) {
    continue
  }
  _ = filteredList.AddItem(item)
}
```
but ok, we don't use both filters at once. We can fix it if we actually need it

jakubdyszkiewicz (approved):

PR #7656, opened 2023-09-06
fix(install): use bitnami/kubectl image for helm hooks

### LINKED ISSUE #6326: Review which kubectl image to use in Helm hooks
### Description

If we still need them can we use an already existing kubectl image instead of building one. If yes let's remove `tools/kubectl`

lahabana:
Fix #6326

### Checklist prior to review


lahabana:
I think one inconvenient of this is that dependabot won't catch it. I think we need to inline the 3 parts to do this.

bartsmykla:
@lahabana what do you mean by:
> I think we need to inline the 3 parts to do this.

michaelbeaumont (commented):
I think we can remove `tools/kubectl` then, right?

lahabana:
> I think we can remove tools/kubectl then, right?

Yes and just did it

lahabana:
Finally works @slonka

slonka (approved):

PR #7662, opened 2023-09-06
fix(xds): backwards compatibility on access logs paths

jakubdyszkiewicz:
### Checklist prior to review

The problem was `else if metadata.Resource != nil {`. In case of indirect lifecycle (Kubernetes), we don't have resource in the metadata therefore `AccessLogSocketPath` was always empty which results in

```
2023-09-05T23:47:22.049Z	ERROR	xds-server.dataplane-sync-watchdog	OnTick() failed	{"dataplaneKey": {"Mesh":"xxx","Name":"xxx"}, "error": "invalid resource \"kuma:metrics:hijacker\": invalid Cluster.LoadAssignment: embedded message failed validation | caused by: invalid ClusterLoadAssignment.Endpoints[0]: embedded message failed validation | caused by: invalid LocalityLbEndpoints.LbEndpoints[0]: embedded message failed validation | caused by: invalid LbEndpoint.Endpoint: embedded message failed validation | caused by: invalid Endpoint.Address: embedded message failed validation | caused by: invalid Address.SocketAddress: embedded message failed validation | caused by: invalid SocketAddress.Address: value length must be at least 1 runes", "errorVerbose": "invalid Cluster.LoadAssignment: embedded message failed validation | caused by: invalid ClusterLoadAssignment.Endpoints[0]: embedded message failed validation | caused by: invalid LocalityLbEndpoints.LbEndpoints[0]: embedded message failed validation | caused by: invalid LbEndpoint.Endpoint: embedded message failed validation | caused by: invalid Endpoint.Address: embedded message failed validation | caused by: invalid Address.SocketAddress: embedded message failed validation | caused by: invalid SocketAddress.Address: value length must be at least 1 runes
```

The problem goes away on redeploy of DPPs, but we should have a seamless upgrade.

This was not caught by E2E test, because our upgrade tests are somewhat basic. Here is the action plan how to fix it for the future https://github.com/kumahq/kuma/issues/7663

The issue only affects 2.4 version.

### DIFF
--- a/pkg/core/xds/metadata.go
+++ b/pkg/core/xds/metadata.go
@@ -142,7 +142,7 @@ func (m *DataplaneMetadata) GetVersion() *mesh_proto.Version {
 	return m.Version
 }
 
-func DataplaneMetadataFromXdsMetadata(xdsMetadata *structpb.Struct, tmpDir string) *DataplaneMetadata {
+func DataplaneMetadataFromXdsMetadata(xdsMetadata *structpb.Struct, tmpDir string, dpKey model.ResourceKey) *DataplaneMetadata {
 	// Be extra careful here about nil checks since xdsMetadata is a "user" input.
 	// Even if we know that something should not be nil since we are generating metadata,
 	// the DiscoveryRequest can still be crafted manually to crash the CP.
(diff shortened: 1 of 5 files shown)


lukidzi (approved):
Nice catch 👍

pkg/core/xds/metadata.go
```diff
@@ -142,7 +142,7 @@ func (m *DataplaneMetadata) GetVersion() *mesh_proto.Version {
 	return m.Version
 }
 
-func DataplaneMetadataFromXdsMetadata(xdsMetadata *structpb.Struct, tmpDir string) *DataplaneMetadata {
+func DataplaneMetadataFromXdsMetadata(xdsMetadata *structpb.Struct, tmpDir string, dpKey model.ResourceKey) *DataplaneMetadata {
```
lobkovilya, line 145: nit: probably we should name this `ProxyMetadataFromXdsMetadata` and not use `dp*` prefix for variable/parameter names.

pkg/xds/server/callbacks/dataplane_status_tracker.go
```diff
@@ -129,26 +129,25 @@ func (c *dataplaneStatusTracker) OnStreamRequest(streamID int64, req util_xds.Di
...
-		case mesh_proto.DataplaneProxyType:
-			dpType = core_mesh.DataplaneType
-		case mesh_proto.EgressProxyType:
-			dpType = core_mesh.ZoneEgressType
-		}
-
 		// Infer the Dataplane ID.
 		if proxyId, err := core_xds.ParseProxyIdFromString(req.NodeId()); err == nil {
 			state.dataplaneId = proxyId.ToResourceKey()
+			var dpType core_model.ResourceType
```
lahabana, line 135: I don't understand this part of the change.

lahabana (approved):
Great catch!

Feels like the api to xdsMetadata should be a bit safer for these kind of cases (Or at least have good error handling).

PR #7708, opened 2023-09-07
docs(madr): madr for extending Global Insight with info about number of policies, zones, services, dataplanes and their status

Automaat:
### Checklist prior to review


docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,90 @@
...
+Technical Story: https://github.com/kumahq/kuma/issues/7707
+
+## Context and Problem Statement
+
+We would like to extend current Global Insight to add more meaningful metrics to it. We would like to add info about number of policies, zones,
+services, dataplanes and their status.
+
+## Considered Options
+
+* extend `GlobalInsight` with data from `MeshInsight` and `ZoneInsight`.
```
bartsmykla, line 14:
```suggestion
* Extend `GlobalInsights` with data from `MeshInsight` and `ZoneInsight`.
```

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Decision Outcome
+
+We will extend Global Insight with detailed information about number of policies, zones, service, dataplanes and its status. It will be done using 
+information from `MeshInsight` and `ZoneInsight`
+
+### Solution
+
+Right now Global Insight is really simple, it only returns count of resources, example response:
```
bartsmykla, line 23: Currently, `GlobalInsights` is not a real resource. We discussed this with @lahabana, and we think it would be a good idea to make it a real resource. This would allow us to cache the results of the calculations, so that we don't have to recalculate them every time the `/global-insights` endpoint is called. I think it would be good to mention this here.

Automaat: That I didn't know, I thought we are only extending API, thanks for sharing. I've updated the MADR accordingly.

bartsmykla (approved):

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,96 @@
...
+#### How to get data
+
+Most of this data is already computed. To assemble this, we need:
+- `services` are mostly present in `MeshInsight`, missing part is `internal_by_status` data. We need to extend `MeshInsight.Services`
+   with this information. Moreover, we need to aggregate this to get the overall count.
+- `dataplanes` are present in `MeshInsight.Dataplanes` we only need to aggregate them to get the overall count.
+- `zones` info can be extracted and aggregated from `ZoneInsight`
+- `policies` are present in `MeshInsight.Policies` we just need to aggregate them to get the overall count.
+
+#### Global Insight as a resource
```
lukidzi, line 92: Is the idea to calculate them as we do for `MeshInsights`? React on events/every x min?

docs/madr/decisions/030-extended-global-insights.md
jakubdyszkiewicz: I'm not sure it's needed, but I'm not against of doing so.

docs/madr/decisions/030-extended-global-insights.md
jakubdyszkiewicz: Would we add this to `resyncer.go` ?

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,96 @@
...
+should look like this:
+
+```json
+{
+  "resources": ..., // <== this stays the same
+  "services": {
+    "total": 5,
+    "internal": 4,
+    "external": 1,
+    "internal_by_status": {
```
kleinfreund, line 59: **Suggestion**: I think the way this is typically returned is with camel-cased fields like so:

```suggestion
    "internalByStatus": {
```

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,96 @@
...
+  "zones": {
+    "cps": {
+      "online": 1,
+      "total": 1
+    }
+  },
+  "dataplanes": {
+    "online": 23,
+    "offline": 10,
+    "partially_degraded": 17,
```
kleinfreund, line 74: **Suggestion**: I think the way this is typically returned is with camel-cased fields like so:

```suggestion
    "partiallyDegraded": 17,
```

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,96 @@
...
+  "policies": {
+    "total": 100
+  }
+}
+```
+
+#### How to get data
+
+Most of this data is already computed. To assemble this, we need:
+- `services` are mostly present in `MeshInsight`, missing part is `internal_by_status` data. We need to extend `MeshInsight.Services`
```
kleinfreund, line 86: **Question**: Does this already imply that `internalByStatus` will also be available on `/mesh-insights`? (I think it should)

docs/madr/decisions/030-extended-global-insights.md
Automaat: I think it would be best to add it to resyncer, since we have all the needed data there

docs/madr/decisions/030-extended-global-insights.md
kleinfreund: Thanks, good to know.

Also, for the sake of consistency:

```suggestion
- `services` are mostly present in `MeshInsight`, missing part is `internalByStatus` data. We need to extend `MeshInsight.Services`
```

docs/madr/decisions/030-extended-global-insights.md
lahabana: Later discussion with @Automaat was to not do this in the resyncer (already complex and doesn't have everything).

His suggestion was a cache and compute on cache addition.

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,96 @@
...
+   with this information. Moreover, we need to aggregate this to get the overall count.
+- `dataplanes` are present in `MeshInsight.Dataplanes` we only need to aggregate them to get the overall count.
+- `zones` info can be extracted and aggregated from `ZoneInsight`
+- `policies` are present in `MeshInsight.Policies` we just need to aggregate them to get the overall count.
+
+#### Global Insight as a resource
+
+Right now Global Insight is not one of the Kuma resources it is computed on the fly after calling `/global-insights` endpoint.
+Global Resource should be treated as any other resource in Kuma, and stored in control plane storage. Thanks to this
+we would be able to compute them on a constant basis and cache the result.
```
lahabana, line 96: I'd actually quite like if this were a new endpoint altogether and we can deprecate the old one in the future.

This would be the occasion as well for us to design how we can generate openAPI specs for ws endpoints that are not just `Resource` (like policies as we already have this).

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,96 @@
...
+```json
+{
+  "resources": ..., // <== this stays the same
+  "services": {
+    "total": 5,
+    "internal": 4,
+    "external": 1,
+    "internalByStatus": {
+      "Online": 2,
+      "Offline": 1,
```
lahabana, line 61: Typo online/offline?

Should we also add serviceType counts?

Automaat: what do you mean by serviceType?

docs/madr/decisions/030-extended-global-insights.md
Automaat: Updated madr accordingly

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,100 @@
...
+We would like to add more detailed information about zones, services and dataplanes. New extended Global Insight 
+should look like this:
+
+```json
+{
+  "resources": ..., // <== this stays the same
+  "services": {
+    "total": 5,
+    "internal": 4,
+    "external": 1,
```
lahabana, line 58: I feel like we now have gateway_builtin and gateway_provided too no?

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,100 @@
...
+    "internalByStatus": {
+      "online": 2,
+      "offline": 1,
+      "partiallyDegraded": 1
+     }
+  },
+  "zones": {
```
lahabana, line 65: I think we probably want:

```json
"zones": {
  "total": 4,
  "controlPlanes": {
     ...
   },
   "zoneEgresses": { ... },
   "zoneIngresses": { ... }
}
```

While this might not be wired through yet. That'd be nice to have them ready as we discussed showing them.

lahabana: Also let's avoid abbreviations

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,100 @@
...
+  },
+  "dataplanes": {
+    "online": 23,
+    "offline": 10,
+    "partiallyDegraded": 17,
+    "total": 50
+  },
+  "policies": {
+    "total": 100
+  }
```
lahabana, line 79: Do we also want:

```json
"meshes": {"total": X}
```

?

lahabana (commented):
We do not mention the new path for this.

Also we should likely version it as indicated in: https://kong-aip.netlify.app/aip/3101/

lahabana:
Also we mentioned that we'd likely want to show additional data in enterprise distributions. If we were how would we go about doing this?

docs/madr/decisions/030-extended-global-insights.md
Automaat: This info is in `resources` part, but we could add it to make it consistent with the rest of the response.

Automaat:
> Also we mentioned that we'd likely want to show additional data in enterprise distributions. If we were how would we go about doing this?
@lahabana We can add a field to response called `additionalData` or `extras`, that will be of type `object` -> `map[string]interface{}` where developers can put anything in custom distributions.

docs/madr/decisions/030-extended-global-insights.md
Automaat: added

docs/madr/decisions/030-extended-global-insights.md
lahabana: Yes think that's what you've done with `builtInGateway` etc

lahabana:
That's not very good UX from a user POV right? If we were to do Go first API spec we could simply change the data type of the API in enterprise distribution Maybe that's somehting generics can help with. @jakubdyszkiewicz what's your opinion?

Automaat:
> That's not very good UX from a user POV right? If we were to do Go first API spec we could simply change the data type of the API in enterprise distribution Maybe that's somehting generics can help with. @jakubdyszkiewicz what's your opinion?

Maybe we can use [openAPI inheritance](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/) to add fields in custom distros, but we would have to replace Go types that were generated by Kuma to achieve that, so I don't have any good solution for this problem.

slonka:
I assume this is tenant aware, right?

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -51,8 +51,6 @@ should look like this:
 
 ```json
 {
-   "resources": ...,
-   // <== this stays the same
```
bartsmykla: Could we add simple sentence that it will be removed and why we decided to do it?

jakubdyszkiewicz (approved):

lukidzi (approved):

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,125 @@
...
+      "internal": 4,
+      "external": 1,
+      "gatewayBuiltin": 1,
+      "gatewayProvided": 2,
+      "internalByStatus": {
+         "online": 2,
+         "offline": 1,
+         "partiallyDegraded": 1
+      }
+   },
```
kleinfreund, line 65: **Question**: Will the `/mesh-insights` endpoint’s `services` field also have this structure or will it remain unchanged?

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -0,0 +1,125 @@
...
+         "online": 1,
+         "total": 1
+      }
+   },
+   "dataplanes": {
+      "online": 23,
+      "offline": 10,
+      "partiallyDegraded": 17,
+      "total": 50
+   },
```
johncowen, line 85: Will we ever need these `dataplane` numbers split into sidecar/gateway? I suppose maybe when knowing health and specifically unhealthy its not that important what type they are just that they are unhealthy? Thought I'd ask in case. Ah, I also just noticed below here that we aren't specifically splitting policies by type either, so I guess this is the same.

kleinfreund: I would actually like those to be only sidecar dataplanes and not gateways. It’s not meaningful to have a combined count of DPPs and Gateways in my opinion.

johncowen: > I suppose maybe when knowing health and specifically unhealthy its not that important what type they are just that they are unhealthy? 

Yeah this ☝️ I suppose, thoughts on the policies?

kleinfreund: > I suppose maybe when knowing health and specifically unhealthy its not that important what type they are just that they are unhealthy?

It’s rather meaningless if you don’t know which type is unhealthy.

johncowen: Oh wait...

> I would actually like those to be only sidecar dataplanes and not gateways. It’s not meaningful to have a combined count of DPPs and Gateways in my opinion.

Ah ok so what you are saying is these shouldn't be/aren't a mix? You would like them to be sidecars only? Is it not useful to know that a gateway is unhealthy?

kleinfreund: It’s probably useful to know both, but currently (and that’s on `/mesh-insights`, not `/global-insights`), these counts are a combination of DPPs and Gateways because they’re the same resource underneath.

In the UI, we specifically have places where the counts for DPPs are shown and this won’t be correct with them being based on both DPPs and Gateways (i.e. DPPs are more important for UI purposes currently).

johncowen: Ok so from what you just said, it sounds like in an ideal world it would be useful if these were split?

> It’s probably useful to know both

> we specifically have places where the counts for DPPs are shown and this won’t be correct with them being based on both DPPs and Gateways

i.e. useful to know both _and_ we need to know how many of only only type in the UI _currently_

I get if there is a technical reason not to do this now, but in an ideal world would we/might we want to know these total split by type at some point?

docs/madr/decisions/030-extended-global-insights.md
Automaat: I started working on implementation and it turns out it is easier to get this data from `ServiceInsight`, than update MeshInsight. I've updated MADR to take this into account.

docs/madr/decisions/030-extended-global-insights.md
johncowen: Does the same reasoning apply to policies? I know specifically of a places in the UI were we split the totals of these.

docs/madr/decisions/030-extended-global-insights.md
```diff
@@ -78,10 +83,18 @@ should look like this:
       }
    },
    "dataplanes": {
```
kleinfreund, line 85: **Question**: As far as I know, exactly this sub structure appears on `/mesh-insights` but it’s on the `dataplanesByType` field. Should we use this field name here, too?

docs/madr/decisions/030-extended-global-insights.md
kleinfreund: We might do this in a mesh-specific context (where we use `/mesh-insights`), but on a global level, I don’t think we need this split. The total number of policies should be good enough.

docs/madr/decisions/030-extended-global-insights.md
Automaat: I don't have any strong opinion on this. We probably should be consistent with `mesh-insight`

Automaat: updated

kleinfreund: Same, I’m fine with either. Just figured I bring it up.

johncowen (approved):

docs/madr/decisions/030-extended-global-insights.md
kleinfreund: This is fine for me in the current state now, so if there’s nothing more, feel free to resolve.

docs/madr/decisions/030-extended-global-insights.md
kleinfreund: So `/mesh-insights` will remain unchanged, correct?

docs/madr/decisions/030-extended-global-insights.md
Automaat: but, when I think more about this, using `dataplanesByType` would be not consistent with the rest of the `GlobalInsight` response

Automaat: so in the end I would go with `dataplanes`

docs/madr/decisions/030-extended-global-insights.md
Automaat: that is right

docs/madr/decisions/030-extended-global-insights.md
kleinfreund: Fine by me, too.

kleinfreund (approved):

PR #7711, opened 2023-09-07
fix(e2e): add test for MeshTimeout referencing MeshHTTPRoute

lobkovilya:
Hash-suffix [PR](https://github.com/kumahq/kuma/pull/7519) is going to change the way policies are named during KDS sync. This is going to affect how `targetRef{kind: MeshHTTPRoute}` is implemented. 

This PR adds an e2e test to make sure this feature is not broken.

### Checklist prior to review


lahabana:
So the goal is to show that matching works correctly even with the name change done by KDS?

michaelbeaumont (approved):

PR #7743, opened 2023-09-13
feat(kuma-cp): allow ResourceStore to be customized

bartsmykla:
This feature allows ResourceStore to be customized in BootstrapPlugins

### Checklist prior to review

  - No relevant issue
  - It won't
  - Tests were updated to reflect this change
  - There is no need
  - There is no need


pkg/core/plugins/interfaces.go
```diff
@@ -47,7 +47,7 @@ type (
 	DbVersion           = uint
 	ResourceStorePlugin interface {
 		Plugin
-		NewResourceStore(PluginContext, PluginConfig) (core_store.ResourceStore, error)
+		NewResourceStore(PluginContext, PluginConfig) (core_store.CustomizableResourceStore, error)
```
jakubdyszkiewicz, line 50: I don't think `CustomizableResourceStore` should now be a "default" interface instead of `ResourceStore`. It's now spread around the codebase.
I think that similarly to `CustomizableResourceManager`, this should be only during runtime building phase.

You introduced `WrapAll` on CustomizableResourceStore, but it's not used. I think what should happen is that in `bootstrap.go` you
* create resource store from the plugin (not customizable resource store)
* create customizable resource store from this resource store
* call `WrapAll` one by one for pagination, metrics etc.

or even simpler - just create everything as is and use `WrapAll` only if you need (outside of Kuma code)

This way the interface is still just a ResourceStore and CustomizableResourceStore is still available if you want to modify the store.

WDYT?

bartsmykla: Addressed

jakubdyszkiewicz (approved):

slonka (approved):

PR #7757, opened 2023-09-14
chore(helm): allow to disable all helm hooks

bartsmykla:
There is an option to set `noHelmHooks` which will as the name implies disable all helm hooks and add appropriate warnings in the notes after installation.

### Checklist prior to review

  - No relevant issue
  - It won't
  - There is no need
  - Yes, this change needs to be backported as it may be necessary for someone with very strict security policy to not allow custom docker images (like the ones used for helm hooks)

### DIFF
--- a/app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
+++ b/app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
@@ -18,7 +18,7 @@ installCrdsOnUpgrade:
   imagePullSecrets: []
 
 # -- Whether to disable all helm hooks
-disableHelmHooks: false
+noHelmHooks: false
 
 controlPlane:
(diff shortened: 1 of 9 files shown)


lahabana:
Looks ok to me. I'd love another pair of eyes

michaelbeaumont:
Is this necessary given `helm install --no-hooks`?

bartsmykla:
> Is this necessary given `helm install --no-hooks`?

We talked about it with @lahabana earlier and there might be situation when user cannot provide flags to helm, so we have to do it like this. The other thing is we are not able to provide any warnings in the case of `--no-hooks` (I know they'll be likely ignored, but still it's good to have them and be able to point users to them when they'll hit one of described problems).

michaelbeaumont:
OK then can we just call the option `noHooks`?

jakubdyszkiewicz (approved):

michaelbeaumont (approved):

Issue #3276, opened 2021-11-23
Add Owners to the SecretManager


lahabana:
### Description

This was done for ResourceManager it should likely be added to the SecretManager too.

jpeach:
Not clear what this issue is asking for, can we add some more detail?

lahabana:
Similar to: https://github.com/kumahq/kuma/pull/729 and https://github.com/kumahq/kuma/pull/742

PR #7770, opened 2023-09-15
feat(store): add owner reference to the secrets

slonka:
based on https://github.com/kumahq/kuma/pull/7736 

closes https://github.com/kumahq/kuma/issues/3276

you can check parent project PR to see how it works there

### Checklist prior to review

### DIFF
--- a/pkg/core/managers/apis/mesh/mesh_manager.go
+++ b/pkg/core/managers/apis/mesh/mesh_manager.go
@@ -76,11 +76,11 @@ func (m *meshManager) Create(ctx context.Context, resource core_model.Resource, 
 	if err := m.meshValidator.ValidateCreate(ctx, opts.Name, mesh); err != nil {
 		return err
 	}
-	// persist Mesh
-	if err := m.store.Create(ctx, mesh, append(fs, core_store.CreatedAt(time.Now()))...); err != nil {
+	if err := EnsureCAs(ctx, m.caManagers, mesh, opts.Name); err != nil {
 		return err
 	}
-	if err := EnsureCAs(ctx, m.caManagers, mesh, opts.Name); err != nil {
+	// persist Mesh
+	if err := m.store.Create(ctx, mesh, append(fs, core_store.CreatedAt(time.Now()))...); err != nil {
 		return err
 	}
 	if err := defaults_mesh.EnsureDefaultMeshResources(
--- a/pkg/core/managers/apis/mesh/mesh_manager.go
+++ b/pkg/core/managers/apis/mesh/mesh_manager.go
@@ -76,11 +76,11 @@ func (m *meshManager) Create(ctx context.Context, resource core_model.Resource, 
 	if err := m.meshValidator.ValidateCreate(ctx, opts.Name, mesh); err != nil {
 		return err
 	}
-	if err := EnsureCAs(ctx, m.caManagers, mesh, opts.Name); err != nil {
-		return err
-	}
 	// persist Mesh
 	if err := m.store.Create(ctx, mesh, append(fs, core_store.CreatedAt(time.Now()))...); err != nil {
+		return err
+	}
+	if err := EnsureCAs(ctx, m.caManagers, mesh, opts.Name); err != nil {
 		return err
 	}
 	if err := defaults_mesh.EnsureDefaultMeshResources(
--- a/pkg/core/managers/apis/mesh/mesh_manager.go
+++ b/pkg/core/managers/apis/mesh/mesh_manager.go
@@ -80,6 +80,7 @@ func (m *meshManager) Create(ctx context.Context, resource core_model.Resource, 
 	if err := m.store.Create(ctx, mesh, append(fs, core_store.CreatedAt(time.Now()))...); err != nil {
 		return err
 	}
+	// We need to first persist the mesh so that we can hook up secrets (cert/key) as their owner in EnsureCAs.
 	if err := EnsureCAs(ctx, m.caManagers, mesh, opts.Name); err != nil {
 		return err
 	}
--- a/pkg/core/managers/apis/mesh/mesh_manager.go
+++ b/pkg/core/managers/apis/mesh/mesh_manager.go
@@ -8,7 +8,6 @@ import (
 
 	core_ca "github.com/kumahq/kuma/pkg/core/ca"
 	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
-	"github.com/kumahq/kuma/pkg/core/resources/apis/system"
 	core_manager "github.com/kumahq/kuma/pkg/core/resources/manager"
 	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
 	core_registry "github.com/kumahq/kuma/pkg/core/resources/registry"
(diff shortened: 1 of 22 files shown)


pkg/plugins/runtime/k8s/controllers/mesh_defaults_controller_test.go
```diff
@@ -42,7 +43,9 @@ var _ = Describe("MeshDefaultsReconciler", func() {
 		store, err := k8s.NewStore(kubeClient, k8sClientScheme, k8s.NewSimpleConverter())
 		Expect(err).ToNot(HaveOccurred())
 
-		secretStore, err := secrets_k8s.NewStore(kubeClient, kubeClient, "default")
+		scheme, err := bootstrap_k8s.NewScheme()
```
slonka, line 46: we need to bring in the actual scheme we're using so that the Mesh CRD can be hooked up as owner. Otherwise we will get:

```
no kind is registered for the type v1alpha1.Mesh in scheme
```

pkg/plugins/secrets/k8s/store_test.go
```diff
@@ -77,7 +78,7 @@ var _ = Describe("KubernetesStore", func() {
 		})
 		Expect(err).ToNot(HaveOccurred())
 
-		s, err = k8s.NewStore(k8sClient, k8sClient, ns)
+		s, err = k8s.NewStore(k8sClient, k8sClient, runtime.NewScheme(), ns)
```
slonka, line 81: need the actual scheme to do the conversion as before

pkg/xds/secrets/secrets_test.go
```diff
@@ -99,14 +101,26 @@ var _ = Describe("Secrets", func() {
 		}
 	}
 
+	BeforeAll(func() {
+		core_mesh.AllowedMTLSBackends = 2
```
slonka, line 105: Now that we actually create a mesh it goes through validation and we have a validation to limit backends to one. I just bumped the limit so that the test works, maybe we could restructure it to work with 1 backend? Should we?

test/e2e_env/kubernetes/gateway/gateway.go
```diff
@@ -41,7 +41,7 @@ spec:
       tls:
         mode: TERMINATE
         certificates:
-        - secret: example-kuma-io-certificate
+        - secret: kuma-io-certificate-k8s
```
slonka, line 44: This was a weird one, please check if my reasoning is correct. Now that we're removing the secrets using owner reference and we're relying on async namespace deletion we could have a situation where the secrets are not yet deleted and we're trying to create a new mesh with the same secret name which k8s treats as changing the `mesh` of the secret https://app.circleci.com/pipelines/github/kumahq/kuma/24848/workflows/f33edb5a-74cb-45ae-b0c2-4b14bf289585/jobs/497239.

pkg/plugins/secrets/k8s/store.go
```diff
@@ -9,40 +9,47 @@ import (
...
 var _ secret_store.SecretStore = &KubernetesStore{}
 
 type KubernetesStore struct {
-	reader    kube_client.Reader
-	writer    kube_client.Writer
-	converter Converter
+	reader             kube_client.Reader
+	writer             kube_client.Writer
+	scheme             *runtime.Scheme
+	secretsConverter   Converter
```
slonka, line 33: secrets have a special converter and we need to convert the `mesh` object since it's the owner. That's why I need another converter. Not sure if this could be done with one converter.

pkg/plugins/resources/postgres/store_template_test.go
```diff
@@ -42,10 +42,8 @@ var _ = Describe("PostgresStore template", func() {
 		}
 	}
 
-	test_store.ExecuteStoreTests(createStore("pgx", 0), "pgx")
```
michaelbeaumont, line 45: Why this change?

pkg/xds/secrets/secrets_test.go
michaelbeaumont: Bumping the limit is fine IMO

test/e2e_env/kubernetes/gateway/gateway.go
michaelbeaumont: Yeah makes sense to me

pkg/core/managers/apis/mesh/mesh_manager.go
```diff
@@ -77,18 +76,17 @@ func (m *meshManager) Create(ctx context.Context, resource core_model.Resource,
...
 	}
-	if err := EnsureCAs(ctx, m.caManagers, mesh, opts.Name); err != nil {
-		return err
-	}
-
 	// persist Mesh
 	if err := m.store.Create(ctx, mesh, append(fs, core_store.CreatedAt(time.Now()))...); err != nil {
 		return err
 	}
+	if err := EnsureCAs(ctx, m.caManagers, mesh, opts.Name); err != nil {
```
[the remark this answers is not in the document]
michaelbeaumont: I guess if `EnsureCAs` fails, we don't have a misconfigured/not working `Mesh`?

pkg/core/managers/apis/mesh/mesh_manager.go
```diff
@@ -119,10 +117,7 @@ func (m *meshManager) Delete(ctx context.Context, resource core_model.Resource,
 			return err
 		}
 	}
-	// delete all secrets
-	if err := m.otherManagers.DeleteAll(ctx, &system.SecretResourceList{}, core_store.DeleteAllByMesh(opts.Name)); err != nil {
-		return errors.Wrap(err, "could not delete associated secrets")
-	}
+	// secrets are deleted via owner reference
```
lobkovilya, line 120: Did we have a bug on k8s of not cleaning up the secrets?

pkg/core/managers/apis/mesh/mesh_manager.go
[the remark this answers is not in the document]
slonka: Yea, should I clean up the mesh if `EnsureCAs` fails? WDYT?

pkg/plugins/resources/postgres/store_template_test.go
slonka: crap, was testing something and forgot to revert this, will revert - sorry!

pkg/core/managers/apis/mesh/mesh_manager.go
slonka: No, we never had a bug but we had 2 different ways of cleaning up the resources. Now we have one.

lobkovilya: IIUC secret deletion though the meshManager is happening only on Universal? How did we clean up secrets on k8s?

slonka: > IIUC secret deletion through the meshManager is happening only on Universal?

I did not check that I thought https://github.com/kumahq/kuma/blob/dd95f22e4e568ebeb899c187b107020a28703db5/pkg/core/managers/apis/mesh/mesh_manager.go#L123-L125 is called always and it just picks which store (k8s or postgres) it uses.

pkg/core/managers/apis/mesh/mesh_manager.go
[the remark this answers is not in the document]
michaelbeaumont: I'm not convinced it matters that much. I mean the mesh can always be misconfigured with a broken CA any time after it's created. So maybe not the end of the world if it's not working when it's created.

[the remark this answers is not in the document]
michaelbeaumont: I mean the mesh can always be misconfigured with a broken CA any time after it's created. So maybe not the end of the world if it's not working when it's created, though no traffic will flow

michaelbeaumont (approved):

PR #7810, opened 2023-09-21
feat(api-server): add filtering on list external-services and dataplanes

### LINKED ISSUE #5908: Provide filter by kuma.io/service on */external-services endpoints
TL;DR: We need a way to retrieve an ExternalService given a **ServiceInsight name** (which can be different than the ExternalService name) in order to show it in the service detail view.

### Description

In order to be able to look up external services by service name (which might be different than the _external_ service name), a filter on the */external-services endpoints will be needed.

This mechanism is needed for the service list and detail views in the GUI where we list services (including external services) based on the service insights API. In order to correctly display data for external services, we need to be able to reference an external service from data that exists on the service insights API.

```sh
curl 'http://localhost:5681/meshes/default/external-services?tag=kuma.io/service:httpbin'
```

This is meant to provide a proper fix for https://github.com/kumahq/kuma-gui/issues/599.

lahabana:
This also opens the path to adding more filters to the api depending on the resource descriptor

Fix #5908

### Checklist prior to review

### DIFF
--- a/pkg/api-server/filters/filtering.go
+++ b/pkg/api-server/filters/filtering.go
@@ -48,7 +48,9 @@ func Resource(resDescriptor core_model.ResourceTypeDescriptor) func(request *res
 			}, nil
 		}
 	default:
-		return nil
+		return func(request *restful.Request) (store.ListFilterFunc, error) {
+			return nil, nil
+		}
 	}
 }
 
--- a/pkg/api-server/filters/filtering.go
+++ b/pkg/api-server/filters/filtering.go
@@ -1,4 +1,4 @@
-package api_server
+package filters
 
 import (
 	"reflect"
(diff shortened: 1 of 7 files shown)


pkg/api-server/filters/filtering.go
```diff
@@ -13,6 +13,47 @@ import (
 	"github.com/kumahq/kuma/pkg/core/validators"
 )
 
+// Resource return a store filter depending on the resource. We take a descriptor so that we can do advance filtering options
+// For example we could make a filter that works on top level targetRef by looking at the descriptor info
+func Resource(resDescriptor core_model.ResourceTypeDescriptor) func(request *restful.Request) (store.ListFilterFunc, error) {
+	switch resDescriptor.Name {
+	case mesh.DataplaneType:
+		return func(request *restful.Request) (store.ListFilterFunc, error) {
```
jakubdyszkiewicz, line 21: nit: I'd extract all cases bodies to functions. `func DataplaneFilter((request *restful.Request) (store.ListFilterFunc, error) {`, `ExternalServicesFilter` etc.

jakubdyszkiewicz (approved):
cool!

pkg/api-server/filters/filtering.go
lahabana: I've looked into it and the actual refactor makes things harder to read IMO

Issue #7817, opened 2023-09-22
Bug in rest error handling for Unauthenticated


bartsmykla:
### What happened?

I believe this:
https://github.com/kumahq/kuma/blob/d76c36a102cadd50dfa1da7f38d0f4efc3451f68/pkg/core/rest/errors/error_handler.go#L153-L155
should be:
```go
	case errors.Is(err, &Unauthenticated{}):
		var unauthenticated *Unauthenticated
		errors.As(err, &unauthenticated)
```

bartsmykla:
Created an issue and not immediately fixed it as I'm addressing one more thing, so I'll do it in one PR to not waste CI resources

PR #7818, opened 2023-09-22
fix(errors): improve HandleError to handle rest_errors.Error and fix Unauthenticated error handling

bartsmykla:
* Update HandleError to process pointer to rest_errors.Error as an argument. This allows us to use errors.As directly, which simplifies the code and allows project depending on Kuma to handle not hardcoded errors.

* Fix a bug in HandleError where it was incorrectly assigning err to itself in the case of an Unauthenticated error.

### Checklist prior to review

  - Closes https://github.com/kumahq/kuma/issues/7817
  - It won't
  - Manually tested
  - There is no need
  - There is no need

### DIFF
--- a/pkg/core/rest/errors/error_handler.go
+++ b/pkg/core/rest/errors/error_handler.go
@@ -24,28 +24,28 @@ import (
 
 func HandleError(ctx context.Context, response *restful.Response, err error, title string) {
 	log := kuma_log.AddFieldsFromCtx(core.Log.WithName("error"), ctx, context.Background())
-	var kumaErr types.Error
+	var kumaErr *types.Error
 	switch {
 	case store.IsResourceNotFound(err):
-		kumaErr = types.Error{
+		kumaErr = &types.Error{
 			Status: 404,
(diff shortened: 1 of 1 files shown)


pkg/core/rest/errors/error_handler.go
```diff
@@ -119,60 +119,61 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
...
+		kumaErr = &types.Error{
 			Status: 403,
 			Title:  "Access Denied",
 			Detail: accessErr.Reason,
 		}
 	case errors.Is(err, &Unauthenticated{}):
 		var unauthenticated *Unauthenticated
-		errors.As(err, &err)
-		kumaErr = types.Error{
+		errors.As(err, &unauthenticated)
```
michaelbeaumont, line 155: Maybe just use `_` if we know this  will return true instead of ignoring

slonka (approved):

michaelbeaumont (approved):

PR #7821, opened 2023-09-22
feat(kuma-cp): introduce zone health checks

michaelbeaumont:
## Explanation

Client:
* Send health check periodically

Server:
* When we get a health check, mark it in the corresponding ZoneInsight
* Introduce new component `ZoneWatch` that listens for `ZoneOpenedStream` and marks the `(tenantID, zone)` for watching
* Periodically check all watched `(tenantID, zone)` and if the time of last health check in `ZoneInsight` is too late, send `ZoneWentOffline`
* All opened streams also listen for `ZoneWentOffline` events and the handler returns if said event is received, ending the stream

We store the info in ZoneInsight because:
All instances need to potentially kill streams but not every instance will receive a health check from connected zones

## Tests

The need for `time.Sleep` in the tests comes about because it happens asynchronously that:
1) `ZoneWatch` subscribes to `ZoneOpenedStream` events in `Start`
  in reality, this is guaranteed to happen before ZoneOpenedStream events are sent by the fact that we only send them in response to new gRPC streams being opened
2) `ZoneOpenedStream` is witnessed by `ZoneWatch`
  the test adds a `time.Sleep` because we only want to update the health check time once and in particular _after_ the zone starts being watched and then check that it's disconnected
  in reality, a zone will continually send its health check ping so it _eventually_ will be updated after the initial seen last time.  

## Todo

- [x] configurability

### Checklist prior to review


pkg/kds/mux/zone_watch.go
```diff
@@ -0,0 +1,97 @@
...
+		case <-zw.Poll.C:
+			core.Log.Info("new poll!")
+			for zone, firstSeen := range zw.zones {
+				ctx := multitenant.WithTenant(context.TODO(), zone.tenantID)
+				zoneInsight := system.NewZoneInsightResource()
+				if err := zw.RM.Get(ctx, zoneInsight, store.GetByKey(zone.zone, model.NoMesh)); err != nil {
+					//
+				}
+				lastHealthCheck := zoneInsight.Spec.GetHealthCheck().GetTime().AsTime()
+				if lastHealthCheck.After(firstSeen) && time.Since(lastHealthCheck) > 20*time.Second {
```
michaelbeaumont, line 63: race condition: zone connects but immediately disconnects before it can send a health check

pkg/kds/global/components.go
```diff
@@ -154,6 +155,15 @@ func Setup(rt runtime.Runtime) error {
 	for _, filter := range rt.KDSContext().GlobalServerFiltersV2 {
 		streamInterceptors = append(streamInterceptors, filter)
 	}
+	zwLog := kdsGlobalLog.WithName("zone-watch")
+	if err := rt.Add(component.NewResilientComponent(zwLog, mux.NewZoneWatch(
+		zwLog,
+		rt.EventBus(),
+		rt.ReadOnlyResourceManager(),
+		*time.NewTicker(5 * time.Second),
```
lobkovilya, line 163: Should we parametrize this as well?

pkg/kds/mux/zone_watch.go
```diff
@@ -0,0 +1,95 @@
...
+		select {
+		case <-zw.poll.C:
+			for zone, firstSeen := range zw.zones {
+				ctx := multitenant.WithTenant(context.TODO(), zone.tenantID)
+				zoneInsight := system.NewZoneInsightResource()
+				if err := zw.rm.Get(ctx, zoneInsight, store.GetByKey(zone.zone, model.NoMesh)); err != nil {
+					zw.log.Info("error getting ZoneInsight", "zone", zone.zone)
+				}
+				lastHealthCheck := zoneInsight.Spec.GetHealthCheck().GetTime().AsTime()
+				if lastHealthCheck.After(firstSeen) && time.Since(lastHealthCheck) > 20*time.Second {
```
lobkovilya, line 66: Probably should move the 20s to the config as well.

pkg/kds/service/server.go
```diff
@@ -136,26 +164,46 @@ func (g *GlobalKDSServiceServer) streamEnvoyAdminRPC(
...
-		if err := rpc.ResponseReceived(clientID, resp); err != nil {
-			logger.Error(err, "could not mark the response as received")
-			return status.Error(codes.InvalidArgument, "could not mark the response as received")
+	}()
+	go func() {
+		defer func() {
+			close(streamReadEnded)
+		}()
+		for {
+			resp, err := recv()
```
lobkovilya, line 189: Is it guaranteed if we return from `streamEnvoyAdminRPC` func (because of the message in `shouldDisconnectStream`) we won't block the goroutine on `recv()`?

pkg/kds/mux/client.go
```diff
@@ -282,6 +285,42 @@ func (c *client) startClusters(
...
+	log logr.Logger,
+	conn *grpc.ClientConn,
+	stop <-chan struct{},
+	errorCh chan error,
+) {
+	client := mesh_proto.NewGlobalKDSServiceClient(conn)
+	log = log.WithValues("rpc", "healthcheck")
+	log.Info("starting")
+
+	ticker := time.NewTicker(5 * time.Second)
```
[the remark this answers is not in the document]
michaelbeaumont: Yeah this is of course going to be parameterized to some degree

pkg/core/features.go
```diff
@@ -1,4 +1,4 @@
-package xds
+package core
```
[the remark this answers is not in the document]
michaelbeaumont: Well it's not just for xds anymore, so either have all features in one package or add a second set of features.

pkg/kds/service/server.go
michaelbeaumont: The assumption is that the stream ends, so we should get EOF or an error here.

pkg/core/features.go
```diff
@@ -14,3 +14,7 @@ func (f Features) HasFeature(feature string) bool {
 // FeatureTCPAccessLogViaNamedPipe indicates that the DP implements TCP accesslog
 // across a named pipe. Legacy DP versions may use structured data across GRPC.
 const FeatureTCPAccessLogViaNamedPipe string = "feature-tcp-accesslog-via-named-pipe"
+
+// FeatureZoneHealthCheck indicates that a zone control plane sends health check
+// messages as long as it's connected.
+const FeatureZoneHealthCheck string = "feature-zone-health-check"
```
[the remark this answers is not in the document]
michaelbeaumont: uh yes. I didn't see that we had that package :facepalm:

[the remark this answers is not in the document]
michaelbeaumont: got rid of all the feature stuff, I'll add in separate commit

pkg/core/features.go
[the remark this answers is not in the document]
michaelbeaumont: already have a kds set of features

pkg/kds/service/server.go
michaelbeaumont: From my testing locally we always get `codes.Canceled` and hit the second `if`.

pkg/kds/mux/client.go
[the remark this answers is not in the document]
michaelbeaumont: I think ideally the server would tell the client how often it expects a health check ping..

pkg/config/multizone/kds.go
```diff
@@ -98,3 +103,22 @@ func (k KdsClientConfig) Sanitize() {
 func (k KdsClientConfig) Validate() error {
 	return nil
 }
+
+type ZoneHealthCheckConfig struct {
+	// PollInterval is the interval between the CP checking ZoneInsight for
```
lobkovilya, line 108: nit: 
```suggestion
	// PollInterval is the interval between the Global CP checking ZoneInsight for
```

pkg/kds/mux/zone_sync.go
```diff
@@ -86,19 +99,26 @@ func (g *KDSSyncServiceServer) GlobalToZoneSync(stream mesh_proto.KDSSyncService
...
 	}
+
+	tenantID, _ := multitenant.TenantFromCtx(stream.Context())
+	shouldDisconnectStream := g.watchZoneHealthCheck(tenantID, zone)
+	defer shouldDisconnectStream.Close()
+
 	processingErrorsCh := make(chan error)
 	go g.zoneToGlobalCb.OnZoneToGlobalSyncConnect(stream, processingErrorsCh)
 	select {
+	case <-shouldDisconnectStream.Recv():
```
jakubdyszkiewicz, line 120: should we log something here?

pkg/kds/mux/zone_sync.go
```diff
@@ -46,32 +50,41 @@ func NewKDSSyncServiceServer(
...
 	}
+
+	tenantID, _ := multitenant.TenantFromCtx(stream.Context())
+	shouldDisconnectStream := g.watchZoneHealthCheck(tenantID, zone)
+	defer shouldDisconnectStream.Close()
+
 	processingErrorsCh := make(chan error)
 	go g.globalToZoneCb.OnGlobalToZoneSyncConnect(stream, processingErrorsCh)
 	select {
+	case <-shouldDisconnectStream.Recv():
```
jakubdyszkiewicz, line 86: same here about the log

pkg/kds/mux/zone_watch.go
```diff
@@ -0,0 +1,99 @@
...
+				}
+			}
+		case e := <-connectionWatch.Recv():
+			newStream := e.(service.ZoneOpenedStream)
+
+			ctx := multitenant.WithTenant(context.TODO(), newStream.TenantID)
+			zoneInsight := system.NewZoneInsightResource()
+
+			if err := zw.rm.Get(ctx, zoneInsight, store.GetByKey(newStream.Zone, model.NoMesh)); err != nil {
+				zw.log.Info("error getting ZoneInsight", "zone", newStream.Zone)
```
jakubdyszkiewicz, line 84: my feeling is that this will be almost always missing. I mean that you need to wait for status sink flush. So maybe it does not make sense even to retrieve it?

pkg/config/multizone/kds.go
lobkovilya: And at the same time, this is the interval between Zone CP pings. Should we add this to the comment?

pkg/kds/mux/zone_watch.go
```diff
@@ -0,0 +1,99 @@
...
+	defer connectionWatch.Close()
+
+	for {
+		select {
+		case <-timer.C:
+			for zone, firstSeen := range zw.zones {
+				ctx := multitenant.WithTenant(context.TODO(), zone.tenantID)
+				zoneInsight := system.NewZoneInsightResource()
+				if err := zw.rm.Get(ctx, zoneInsight, store.GetByKey(zone.zone, model.NoMesh)); err != nil {
+					zw.log.Info("error getting ZoneInsight", "zone", zone.zone)
```
jakubdyszkiewicz, line 66: add tenant ID to this log using AddFieldsFromCtx

jakubdyszkiewicz: I'd add `"cause", err`

jakubdyszkiewicz: don't we need a `continue` here?

pkg/kds/service/server.go
```diff
@@ -136,25 +181,38 @@ func (g *GlobalKDSServiceServer) streamEnvoyAdminRPC(
...
-		}
-		if err != nil {
-			logger.Error(err, "could not receive a message")
-			return status.Error(codes.Internal, "could not receive a message")
-		}
-		logger.V(1).Info("Envoy Admin RPC response received", "requestId", resp.GetRequestId())
-		if err := rpc.ResponseReceived(clientID, resp); err != nil {
-			logger.Error(err, "could not mark the response as received")
-			return status.Error(codes.InvalidArgument, "could not mark the response as received")
+	streamResult := make(chan error, 2)
```
jakubdyszkiewicz, line 184: why 2?

pkg/kds/service/server.go
```diff
@@ -136,25 +181,38 @@ func (g *GlobalKDSServiceServer) streamEnvoyAdminRPC(
...
+			logger.V(1).Info("Envoy Admin RPC response received", "requestId", resp.GetRequestId())
+			if err := rpc.ResponseReceived(clientID, resp); err != nil {
+				logger.Error(err, "could not mark the response as received")
+				streamResult <- status.Error(codes.InvalidArgument, "could not mark the response as received")
+				return
+			}
 		}
+	}()
+	select {
+	case <-shouldDisconnectStream.Recv():
```
jakubdyszkiewicz, line 212: it might be good to log here that we are terminating stream because we did not receive the HC.

pkg/kds/mux/client.go
```diff
@@ -282,6 +287,39 @@ func (c *client) startClusters(
...
+	log logr.Logger,
+	conn *grpc.ClientConn,
+	stop <-chan struct{},
+	errorCh chan error,
+) {
+	client := mesh_proto.NewGlobalKDSServiceClient(conn)
+	log = log.WithValues("rpc", "healthcheck")
+	log.Info("starting")
+
+	go func() {
```
jakubdyszkiewicz, line 301: we need more logs here

pkg/kds/service/server.go
michaelbeaumont: ah woops, left over from when I had two goroutines.

pkg/kds/mux/client.go
michaelbeaumont: where? I'm not sure we should log everytime we make a health check request. should I add a log when the timer duration changes from the last duration?

jakubdyszkiewicz: We need to know whether Zone CP sends HC and if this succeeds. We could just log, but if we think it's too much (just like it was with intercp) we can have a metric counter to at least see that it's incrementing.

Additionally, I'd add a log "global cp requested to send healthcheck every X second"  like you pointed out

pkg/kds/mux/zone_watch.go
michaelbeaumont: added explanation of why we need this

pkg/kds/mux/zone_watch.go
```diff
@@ -0,0 +1,99 @@
...
+) *ZoneWatch {
+	return &ZoneWatch{
+		log:     log,
+		poll:    cfg.ZoneHealthCheck.PollInterval.Duration,
+		timeout: cfg.ZoneHealthCheck.Timeout.Duration,
+		bus:     bus,
+		rm:      rm,
+		zones:   map[zoneTenant]time.Time{},
+	}
+}
```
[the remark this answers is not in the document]
michaelbeaumont: done

pkg/kds/mux/client.go
michaelbeaumont: added logs

lobkovilya (approved):

pkg/kds/service/server.go
```diff
@@ -34,13 +39,21 @@ type StreamInterceptor interface {
 	InterceptServerStream(stream grpc.ServerStream) error
 }
 
+type ActiveStreams struct {
+	XDSConfig chan struct{}
+	Stats     chan struct{}
+	Clusters  chan struct{}
+}
```
jakubdyszkiewicz, line 46: Is it used anywhere?

jakubdyszkiewicz (approved):
I think we could introduce tests at least for ZoneWatch. Any ideas for E2E test?

pkg/kds/service/server.go
michaelbeaumont: oops! getting rid of it

michaelbeaumont:
> I think we could introduce tests at least for ZoneWatch. Any ideas for E2E test?

I think e2e will be somewhat difficult because of the nature of isolating failure to the app level health check. The always possible, but ugly option is to introduce an option like "don't send health check pings" to the CP but IMO we should avoid this artificial, public test interface

@slonka mentioned potentially sending SIGSTOP to make the process stop sending pings but keep TCP connections open. The problem that I see is that HTTP2 PING ACKs also stop, which will trigger the gRPC keep alive to fail.

jakubdyszkiewicz:
> but IMO we should avoid this artificial, public test interface

I agree.

We could put Kong in the middle (we have test infra to deploy KIC) which apparently does not proxy PING/GOAWAY and then do SIGKILL on Zone CP. I wonder if this would work.

slonka:
> We could put Kong in the middle (we have test infra to deploy KIC) which apparently does not proxy PING/GOAWAY and then do SIGKILL on Zone CP. I wonder if this would work.

It should work 👍

michaelbeaumont:
OK, if we have some infrastructure for starting up kong gateway it becomes significantly less work

Issue #7823, opened 2023-09-21
Iptables locked and is not retried


johnharris85:
### What happened?

We're hitting a big with CNI where xtables is locked. Not sure the actual cause (but could happen I guess), but we should probably implement a retry process (like https://github.com/istio/istio/issues/31936) or something here for some resiliency.

PR #7870, opened 2023-09-26
feat(transparent-proxy): allow to wait for xtables lock and retry when installing tproxy fails

bartsmykla:
Allow to wait for xtables lock when something else is holding it concurrently with us + retry when tproxy installation fails

### Checklist prior to review

  - Closes: https://github.com/kumahq/kuma/issues/7823
  - It won't
  - manually tested and unit tests updated
  - There is no need


slonka:
@bartsmykla format does not work https://github.com/kumahq/kuma/pull/7870#issuecomment-1735290933

lahabana:
Here's the issue: https://github.com/kumahq/kuma/issues/7873

app/kumactl/cmd/install/install_transparent_proxy.go
```diff
@@ -207,6 +219,11 @@ runuser -u kuma-dp -- \
...
 	cmd.Flags().StringArrayVar(&args.VnetNetworks, "vnet", []string{}, "virtual networks in a format of interfaceNameRegex:CIDR split by ':' where interface name doesn't have to be exact name e.g. docker0:172.17.0.0/16, br+:172.18.0.0/16, iface:::1/64")
+	cmd.Flags().UintVar(&args.Wait, "wait", 0, "specify the amount of time, in seconds, that the application should wait for the xtables exclusive lock before exiting. If the lock is not available within the specified time, the application will exit with an error. Default value 0 means wait forever. To disable this behavior and exit immediately if the xtables lock is not available, use the --dont-wait flag")
+	cmd.Flags().BoolVar(&args.DontWait, "dont-wait", false, "disable the default behavior of waiting for the xtables exclusive lock. If the lock is not available when the application starts, the application will exit immediately with an error")
```
slonka, line 223: should we just have high value of `wait` and not have `dont-wait` flag?

pkg/transparentproxy/config/config.go
```diff
@@ -140,6 +151,26 @@ type Config struct {
 	// Log is the place where configuration for logging iptables rules will
 	// be placed
 	Log LogConfig
+	// RestoreLegacy indicates whether the ip{6}tables-restore version is legacy
+	// (non-nftables). If RestoreLegacy is true, the flags "--wait" and
+	// "--wait-interval" are valid. These flags are invalid with nftables.
```
slonka, line 156: Do we need this flag at all? Can't we detect if it's xtables vs nftables?

might be worth mentioning that this is not needed with nftables

```
With back end transactions being atomic, there is no need for the global xtables lock which has proven problematic in environments with large and/or rapidly changing rulesets
```

slonka (commented):
Couple of questions around switches

app/kumactl/cmd/install/install_transparent_proxy.go
lahabana: Yes. In other words why would you not want to wait?

pkg/transparentproxy/config/config.go
bartsmykla: It was not a flag, but just a Config property, as I was checking this in earlier stage of the installation, but I moved it now and got rid of the property completely

app/kumactl/cmd/install/install_transparent_proxy.go
bartsmykla: > should we just have high value of wait and not have dont-wait flag?

With value `0` being an equivalent of `wait forever`, without this flag or creating custom type to work with pointers to values it was not possible to differentiate if user provided value `0` (don't wait at all), or if it's a default one

> Yes. In other words why would you not want to wait?

Just wanted to be defensive. I'm not sure if it's possible, but if something else in the system would hold the lock forever, we would just hang.

I'm gonna remove this flag and set a default value of 5s, and change the behavior that by providing `0`, it will mean `don't wait`, not `wait forever`

app/kumactl/cmd/install/install_transparent_proxy.go
```diff
@@ -43,6 +44,11 @@ type transparentProxyArgs struct {
 	EbpfCgroupPath                 string
 	EbpfTCAttachIface              string
 	VnetNetworks                   []string
+	Wait                           uint
```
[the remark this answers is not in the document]
bartsmykla: removed

app/kumactl/cmd/install/install_transparent_proxy.go
```diff
@@ -219,9 +216,8 @@ runuser -u kuma-dp -- \
...
-	cmd.Flags().BoolVar(&args.DontWait, "dont-wait", false, "disable the default behavior of waiting for the xtables exclusive lock. If the lock is not available when the application starts, the application will exit immediately with an error")
-	cmd.Flags().UintVar(&args.WaitInterval, "wait-interval", 0, "flag can be used to specify the amount of time, in microseconds, that iptables should wait between each iteration of the lock acquisition loop. This can be useful if the xtables lock is being held by another application for a long time, and you want to reduce the amount of CPU that iptables uses while waiting for the lock")
+	cmd.Flags().UintVar(&args.Wait, "wait", args.Wait, "specify the amount of time, in seconds, that the application should wait for the xtables exclusive lock before exiting. If the lock is not available within the specified time, the application will exit with an error. Default value 0 means wait forever")
```
slonka, line 219: Here you said that 0 will mean no wait.

> I'm gonna remove this flag and set a default value of 5s, and change the behavior that by providing 0, it will mean don't wait, not wait forever

slonka (approved):
All good apart from one discrepancy

app/kumactl/cmd/install/install_transparent_proxy.go
bartsmykla: good catch, forgot about this one

PR #7901, opened 2023-09-27
feat(kuma-cp): allow to disable backend validation

lukidzi:
### Checklist prior to review


api/mesh/v1alpha1/mesh.proto
```diff
@@ -36,6 +36,9 @@ message Mesh {
 
     // List of available Certificate Authority backends
     repeated CertificateAuthorityBackend backends = 2 [ (doc.required) = true ];
+
+    // If enabled, allows to skip CA validation.
```
michaelbeaumont, line 40:
```suggestion
    // If enabled, skips CA validation.
```

pkg/core/managers/apis/mesh/mesh_manager_test.go
```diff
@@ -70,6 +70,22 @@ var _ = Describe("Mesh Manager", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		It("should create CA without validation", func() {
```
michaelbeaumont, line 73: And ultra nit pic feel free to ignore all the comments but always makes me stumble slightly mentally trying to figure out how many negations we're talking about: we should just use "skip" or some term exclusively 

Certainly a problem everywhere in the code/config, sometimes it's "without", sometimes "skip", sometimes "enabled/disabled"..

michaelbeaumont (approved):

michaelbeaumont (approved):

jakubdyszkiewicz (approved):

Issue #4985, opened 2022-09-08
`kumactl inspect dataplane <invalidName> --type clusters|config-dump|stats` return 500


lahabana:
### What happened?

I did: `~/Desktop/kuma-1.8.0/bin/kumactl inspect dataplane fake-client-7676f8f45d-s87rp   --type stats` and got `Error: Could not get dataplane resource (Internal Server Error)` 

The CP log:
````
[...]

PR #7903, opened 2023-09-28
chore(kuma-cp): remove default prefix from PreconditionError.Reason

bartsmykla:
Enterprise distributions can use this type of error to return not only format related errors.

### Checklist prior to review

  - No relevant issue
  - It won't
  - Manually tested - it's a trivial change, so I don't think there is need to introduce new unit tests
  - There is no need
  - There is no need
> Changelog: skip

### DIFF
--- a/pkg/core/resources/store/store.go
+++ b/pkg/core/resources/store/store.go
@@ -156,9 +156,13 @@ type PreconditionError struct {
 }
 
 func (a *PreconditionError) Error() string {
-	return "invalid format: " + a.Reason
+	return a.Reason
 }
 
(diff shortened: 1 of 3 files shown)


lahabana (approved):
Seems like e2e tests failed due to a changed error message: `[It] Inspect should return bad request on invalid name(#4985)` 

```
[FAILED] Expected
    <string>: error while running command: exit status 1; Error: Bad Request (invalid format: name "dummy-name" must include namespace after the dot, ex. "name.namespace")
to contain substring
    <string>: Bad Request (name "dummy-name" must include namespace after the dot, ex. "name.namespace")
```

PR #7905, opened 2023-09-28
feat(kds): execute filters on envoy admin streams

jakubdyszkiewicz:
### Checklist prior to review

Execute the same interceptors as with KDS to be able to plug in extra functionallity

### DIFF
--- a/pkg/kds/service/server.go
+++ b/pkg/kds/service/server.go
@@ -27,10 +27,15 @@ import (
 
 var log = core.Log.WithName("kds-service")
 
+type StreamInterceptor interface {
+	InterceptServerStream(stream grpc.ServerStream) error
+}
+
 type GlobalKDSServiceServer struct {
 	envoyAdminRPCs EnvoyAdminRPCs
 	resManager     manager.ResourceManager
 	instanceID     string
+	filters        []StreamInterceptor
 	mesh_proto.UnimplementedGlobalKDSServiceServer
 }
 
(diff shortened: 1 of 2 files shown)


pkg/kds/service/server.go
```diff
@@ -102,6 +109,12 @@ func (g *GlobalKDSServiceServer) streamEnvoyAdminRPC(
 	}
 	clientID := ClientID(stream.Context(), zone)
 	logger := log.WithValues("rpc", rpcName, "clientID", clientID)
+	for _, filter := range g.filters {
+		if err := filter.InterceptServerStream(stream); err != nil {
+			logger.Info("stream interceptor terminating the stream", "cause", err)
```
lobkovilya, line 114: why Info and not Error?

pkg/kds/service/server.go
```diff
@@ -102,6 +109,12 @@ func (g *GlobalKDSServiceServer) streamEnvoyAdminRPC(
...
+			return status.Error(codes.InvalidArgument, "invalid stream")
```
lobkovilya, line 115: Isn't it `codes.Internal`?

pkg/kds/service/server.go
lahabana: Cause it's a user error no?

lahabana: Actually yeah feels like the error type should from the interceptor no?

jakubdyszkiewicz: right. Changed the code to check what was the error code from interceptor. If Invalid then just do info, otherwise error

Issue #7849, opened 2023-09-25
Wait For Dataplane Ready takes too long and times out at 3 minutes


cbugneac-nex:
### What happened?

I have enabled the `waitForDataplaneReady` functionality and now starting a pod takes ~ 3 minutes due to postStart lifecycle hook timing out.
It does restart the pod eventually but waiting 3 minutes is not acceptable. Are there any other changes needed to be done apart for setting in Helm chart ?

The config is pretty generic, just couple for `externalServices`.
This is the config I have added in Helm chart values file:
```
  envVars:
    KUMA_RUNTIME_KUBERNETES_INJECTOR_SIDECAR_CONTAINER_WAIT_FOR_DATAPLANE_READY: true
```

Here are the pod events after enabling this feature:
```
Events:
  Type                    Reason               Age           From                           Message
  ----                    ------               ----          ----                           -------
  Normal                  Scheduled            3m46s         default-scheduler              Successfully assigned ***/****-5fb758f5b-vlwsx to ip-**-***-***-***.eu-west-1.compute.internal
  Warning                 FailedPostStartHook  45s           kubelet                        Exec lifecycle hook ([kuma-dp wait --url http://localhost:9901/ready]) for Container "kuma-sidecar" in Pod "****-5fb758f5b-vlwsx_nsl(5b505d6d-42cf-4138-aba7-8dba2a5bf87c)" failed - error: command 'kuma-dp wait --url http://localhost:9901/ready' exited with 1: 2023-09-25T12:02:01.897Z  INFO  kuma-dp.wait  waiting for data plane proxy to be ready  {"timeout": "3m0s"}
2023-09-25T12:02:02.898Z  INFO                 kuma-dp.wait  data plane proxy is not ready  {"err": "Get \"http://localhost:9901/ready\": dial tcp [::1]:9901: connect: connection refused"}
...
2023-09-25T12:05:01.898Z  INFO                 kuma-dp.wait  data plane proxy is not ready  {"err": "Get \"http://localhost:9901/ready\": dial tcp [::1]:9901: connect: connection refused"}
Error: timeout occurred while waiting for data plane proxy to be ready
[...]

PR #7909, opened 2023-09-28
fix(kuma-cp): specifying IPv6 Envoy Admin address breaks readiness/liveness probes

lobkovilya:
Today there is a possibility to change the Envoy Admin address to `::1` but `admin_proxy_generator.go` generates an admin cluster always with a `127.0.0.1` address.

This PR puts a specified admin address to proxy metadata and uses it when generating an admin cluster.

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/7849

### DIFF
--- a/pkg/xds/bootstrap/testdata/generator.metrics-config.kubernetes.golden.yaml
+++ b/pkg/xds/bootstrap/testdata/generator.metrics-config.kubernetes.golden.yaml
@@ -37,6 +37,7 @@ node:
   id: mesh.name.namespace
   metadata:
     accessLogSocketPath: /tmp/kuma-al-name.namespace-mesh.sock
+    dataplane.admin.address: 127.0.0.1
     dataplane.admin.port: "1234"
     dataplane.proxyType: dataplane
     features: []
--- a/pkg/xds/generator/admin_proxy_generator.go
+++ b/pkg/xds/generator/admin_proxy_generator.go
@@ -48,10 +48,14 @@ func (g AdminProxyGenerator) Generate(ctx context.Context, xdsCtx xds_context.Co
 	// since it would allow a malicious user to manipulate that value and use Prometheus endpoint
 	// as a gateway to another host.
 	envoyAdminClusterName := envoy_names.GetEnvoyAdminClusterName()
+	adminAddress := proxy.Metadata.GetAdminAddress()
+	if adminAddress == "" {
+		adminAddress = "127.0.0.1"
+	}
 	cluster, err := envoy_clusters.NewClusterBuilder(proxy.APIVersion, envoyAdminClusterName).
 		Configure(envoy_clusters.ProvidedEndpointCluster(
-			govalidator.IsIPv6(proxy.Metadata.GetAdminAddress()),
-			core_xds.Endpoint{Target: proxy.Metadata.GetAdminAddress(), Port: adminPort})).
+			govalidator.IsIPv6(adminAddress),
+			core_xds.Endpoint{Target: adminAddress, Port: adminPort})).
 		Configure(envoy_clusters.DefaultTimeout()).
 		Build()
 	if err != nil {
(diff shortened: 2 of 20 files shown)


cbugneac-nex:
Hi @lobkovilya should we also mention that `kuma-sidecar` (kuma-dp) also needs to bind to IPv6 or this is for a separate PR ?

lobkovilya:
Hi, kuma-sidecar is binding to the data plane proxy inbound address as you can see here (g.getAddress(proxy) returns it): https://github.com/kumahq/kuma/blob/21efbfb1e2d72ca9b77097859cffee81b41b2cc8/pkg/xds/generator/admin_proxy_generator.go#L87. 

And I believe we already have IPv6 support for DPP inbounds, so it should work. The only issue here was using `127.0.0.1` in the envoy admin cluster and this PR changes this to take the address from the proxy's metadata.

pkg/xds/generator/admin_proxy_generator.go
```diff
@@ -45,10 +47,15 @@ func (g AdminProxyGenerator) Generate(ctx context.Context, xdsCtx xds_context.Co
 	// In contrast to `AdminPort`, we shouldn't trust `AdminAddress` from the Envoy node metadata
 	// since it would allow a malicious user to manipulate that value and use Prometheus endpoint
```
michaelbeaumont, line 48: Is this comment no longer true/relevant?

pkg/xds/bootstrap/testdata/generator.metrics-config.kubernetes.golden.yaml
```diff
@@ -37,6 +37,7 @@ node:
   id: mesh.name.namespace
   metadata:
     accessLogSocketPath: /tmp/kuma-al-name.namespace-mesh.sock
+    dataplane.admin.address: 127.0.0.1
```
michaelbeaumont, line 40: can we add a test case with ipv6?

pkg/xds/generator/admin_proxy_generator.go
lobkovilya: I re-read the comment and figured that we shouldn't allow other IPs except `127.0.0.1, ::1, 0.0.0.0, ::`, so slightly updated the comment and added validation

pkg/xds/bootstrap/testdata/generator.metrics-config.kubernetes.golden.yaml
lobkovilya: added tests, but not for bootstrap code because there is no ipv6 logic. I added tests for the `admin_proxy_generator.go`

pkg/xds/generator/admin_proxy_generator.go
michaelbeaumont: yeah that makes complete sense actually :+1:

michaelbeaumont (approved):

lahabana:
Couple of questions:

- what happens in dual stack?
- could we have not added an e2e test to make sure it works on both ipv4 and ipv6 in real life setups?

lobkovilya:
> what happens in dual stack?

Envoy binds to `127.0.0.1` by default, so dual stack worked well before this PR.

> could we have not added an e2e test to make sure it works on both ipv4 and ipv6 in real life setups?

I'm not sure if we need a dedicated e2e test for this, but I can change one Zone CP when running kindipv6 to have:
```
KUMA_RUNTIME_KUBERNETES_INJECTOR_SIDECAR_CONTAINER_WAIT_FOR_DATAPLANE_READY: "true"
KUMA_BOOTSTRAP_SERVER_PARAMS_ADMIN_ADDRESS: "::1"
```
that way we can ensure these features are tested.

cbugneac-nex:
Apologies, will this change be included in the next release, e.g. 2.4.2 ?

michaelbeaumont:
@cbugneac-nex Yes

lahabana:
I like this option of using `::1` for our ipv6 tests yes @lobkovilya !

Issue #7922, opened 2023-09-29
Add store parameter to enforce write replica for reads


jakubdyszkiewicz:
### Description

For Upsert it's better to enforce write replica. We can have `Consistent()` store param and pick write pool

PR #7923, opened 2023-09-29
feat(store): consistent gets for read replica

jakubdyszkiewicz:
### Checklist prior to review

So whenever we use read replica for postgres and execute Upsert, we reduce chance of read replica lag.

Fix https://github.com/kumahq/kuma/issues/7922

### DIFF
--- a/pkg/core/resources/manager/manager.go
+++ b/pkg/core/resources/manager/manager.go
@@ -130,7 +130,7 @@ var ErrSkipUpsert = errors.New("don't do upsert")
 func Upsert(ctx context.Context, manager ResourceManager, key model.ResourceKey, resource model.Resource, fn func(resource model.Resource) error, fs ...UpsertFunc) error {
 	upsert := func(ctx context.Context) error {
 		create := false
-		err := manager.Get(ctx, resource, store.GetBy(key))
+		err := manager.Get(ctx, resource, store.GetBy(key), store.GetConsistent())
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/core/resources/manager/manager.go
```diff
@@ -130,7 +130,7 @@ var ErrSkipUpsert = errors.New("don't do upsert")
 func Upsert(ctx context.Context, manager ResourceManager, key model.ResourceKey, resource model.Resource, fn func(resource model.Resource) error, fs ...UpsertFunc) error {
 	upsert := func(ctx context.Context) error {
 		create := false
-		err := manager.Get(ctx, resource, store.GetBy(key))
+		err := manager.Get(ctx, resource, store.GetBy(key), store.GetConsistent())
```
Automaat, line 133: Wouldn't it be better to utilize postgres `insert ... on conflict` clause instead of reimplementing it as `upsert` in our code? With `on conflict` we are always performing operations on master

lukidzi (approved):

pkg/core/resources/manager/manager.go
Automaat: you can specify `conflict-target` so it does not only work with primary keys

Automaat: I am not sure, but probably conflict target needs to be unique, so in our case it won;t work

Automaat (approved):

Issue #7129, opened 2023-06-27
Add helm support for running zone universal CP


lahabana:
### Description

We've done it for global not for zones

PR #7924, opened 2023-09-29
feat(helm): add possibility to run universal zone cp on kubernetes

Automaat:
Fixes: #7129

### Checklist prior to review

- [ x Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) --


test/e2e/helm/kuma_helm_deploy_global_and_zone_universal_mode.go
```diff
@@ -0,0 +1,113 @@
...
+
+	E2EAfterAll(func() {
+		Expect(zoneCluster.DeleteNamespace(TestNamespace)).To(Succeed())
+		Expect(globalCluster.DeleteKuma()).To(Succeed())
+		Expect(zoneCluster.DeleteKuma()).To(Succeed())
+		Expect(globalCluster.DismissCluster()).To(Succeed())
+		Expect(zoneCluster.DismissCluster()).To(Succeed())
+	})
+
+	It("should deploy Zone and Global on 2 clusters", func() {
```
Automaat, line 105: I am not sure abut this test, I don't think it tests thoroughly this feature, but adding tests that run universal services on k8s sounds like non-trivial task, and would require adding lots of testing logic, I don't know if we would like to invest more time into this

jakubdyszkiewicz: I think it's fine as is

jakubdyszkiewicz (commented):
Looks good, can you also make sure that we don't run HELM webhook for both global and zone universal. Thsoe hook patches namespace and manages CRD which is redundant

Automaat:
> Looks good, can you also make sure that we don't run HELM webhook for both global and zone universal. Thsoe hook patches namespace and manages CRD which is redundant

@jakubdyszkiewicz isn't this achieved by running helm with `--skip-crds` option. For global, we were suggesting this in docs

jakubdyszkiewicz:
It's not. Even with `--skip-crds` HELM would still execute HELM hooks. See `pre-*` and `post-*` files

Automaat:
> It's not. Even with `--skip-crds` HELM would still execute HELM hooks. See `pre-*` and `post-*` files

ok, checked and added missing `if` in two files

jakubdyszkiewicz (approved):

Issue #2604, opened 2021-08-19
Use Go error type wrapping for  store error detection


jpeach:
### Summary

Errors from the kuma store are untyped, which means that APIs like `IsResourceNotFound` work by string comparison. Go 1.13 formalized the [error wrapping API](https://go.dev/blog/go1.13-errors), so we can now structure this cleanly.

* each error should be a distinct type
* every `ErrorResourceFoobar()` should have a matching `IsResourceFoobarr()`
* the `IsResource` APIs should be implemented in terms of [`errors.Is`](https://pkg.go.dev/errors#Is)

jakubdyszkiewicz:
> Triage: we also need a MADR with recommendations like logging.

jakubdyszkiewicz:
Seems like only `ErrorResourceNotFound` is left

PR #7937, opened 2023-10-02
fix(api-server): improve error handling and return status

lahabana:
- never return 412 (it was never appropriate
- create better error hierarchy to simplify handler
- use in some places errors.Is

### Checklist prior to review


pkg/core/resources/manager/manager.go
```diff
@@ -168,7 +168,7 @@ func Upsert(ctx context.Context, manager ResourceManager, key model.ResourceKey,
 			return err
 		}
 		err := upsert(ctx)
-		if store.IsResourceConflict(err) || store.IsResourceAlreadyExists(err) {
```
jakubdyszkiewicz, line 171: why ResourcesAlreadyExist is gone?

lahabana:
@jakubdyszkiewicz I think it makes sense to have a MADR (also because implementing things like `Is` is not always straightforward :)

pkg/core/rest/errors/error_handler.go
```diff
@@ -130,7 +98,7 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 			Title:  "Method not Allowed",
 			Detail: err.Error(),
 		}
-	case errors.Is(err, &Conflict{}):
+	case errors.Is(err, &Conflict{}) || errors.Is(err, &store.ResourceConflictError{}):
```
bartsmykla, line 101: 1. It seems this case `errors.Is(err, &Conflict{})` is invalid, as `*Conflict` doesn't have `Is(err error) bool` method, so it will always return `false`

2. 
```suggestion
	case errors.Is(err, &Conflict{}), errors.Is(err, &store.ResourceConflictError{}):
```

3.
The same suggestion as mentioned earlier, I think this PR is a great opportunity to make other Errors consistent, so I would refactor [these](https://github.com/kumahq/kuma/blob/51bb6190f28e05c4d826372975f47221f92a62c1/pkg/core/rest/errors/errors.go#L3-L25):
```go
package errors

import (
	"reflect"
)

type UnauthenticatedError struct{}

func (e *UnauthenticatedError) Error() string {
	return "Unauthenticated"
}

func (e *UnauthenticatedError) Is(err error) bool {
	return reflect.TypeOf(err) == reflect.TypeOf(e)
}

type MethodNotAllowedError struct{}

func (e *MethodNotAllowedError) Error() string {
	return "Method not allowed"
}

func (e *MethodNotAllowedError) Is(err error) bool {
	return reflect.TypeOf(err) == reflect.TypeOf(e)
}

type ConflictError struct{}

func (e *ConflictError) Error() string {
	return "Conflict"
}

func (e *ConflictError) Is(err error) bool {
	return reflect.TypeOf(err) == reflect.TypeOf(e)
}

type ServiceUnavailableError struct{}

func (e *ServiceUnavailableError) Error() string {
	return "Service unavailable"
}

func (e *ServiceUnavailableError) Is(err error) bool {
	return reflect.TypeOf(err) == reflect.TypeOf(e)
}
```

pkg/core/rest/errors/error_handler.go
```diff
@@ -94,27 +62,27 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 				Reason: violation.Message,
 			})
 		}
-	case api_server_types.IsMaxPageSizeExceeded(err):
+	case err == store.ErrorInvalidOffset:
```
bartsmykla, line 65: I know it's a simple error, but we already have similar ones: https://github.com/kumahq/kuma/blob/51bb6190f28e05c4d826372975f47221f92a62c1/pkg/core/rest/errors/errors.go#L3-L25
I think [it](https://github.com/kumahq/kuma/blob/5811eaa876a692855db5fc13da553a70f8f2203c/pkg/core/resources/store/store.go#L146) also should be renamed (to `InvalidOffsetError`) and be a separate error with `Is` method so this case statement would be similar as in other ones:
```go
case errors.Is(err, &store.InvalidOffsetError{}):
```

pkg/core/resources/store/store.go
```diff
@@ -117,24 +116,31 @@ func (s *strictResourceStore) Close() error {
...
-	return fmt.Errorf("Resource precondition failed: type=%q name=%q mesh=%q", rt, name, mesh)
+func ErrorResourceConflict(rt model.ResourceType, name, mesh string) error {
+	return &ResourceConflictError{msg: "resource conflict", rType: rt, name: name, mesh: mesh}
+}
+
+func ErrorResourceNotFound(rt model.ResourceType, name, mesh string) error {
+	return fmt.Errorf("Resource not found: type=%q name=%q mesh=%q", rt, name, mesh)
 }
 
 var ErrorInvalidOffset = errors.New("invalid offset")
```
bartsmykla, line 146:
```suggestion
type InvalidOffsetError struct {}

func (e *InvalidOffsetError) Error() string {
	return "invalid offset"
}

func (e *InvalidOffsetError) Is(err error) bool {
	return reflect.TypeOf(err) == reflect.TypeOf(e)
}
```

pkg/core/resources/store/store.go
```diff
@@ -117,24 +116,31 @@ func (s *strictResourceStore) Close() error {
...
 
-func ErrorResourcePreconditionFailed(rt model.ResourceType, name, mesh string) error {
-	return fmt.Errorf("Resource precondition failed: type=%q name=%q mesh=%q", rt, name, mesh)
+func ErrorResourceConflict(rt model.ResourceType, name, mesh string) error {
+	return &ResourceConflictError{msg: "resource conflict", rType: rt, name: name, mesh: mesh}
+}
+
+func ErrorResourceNotFound(rt model.ResourceType, name, mesh string) error {
+	return fmt.Errorf("Resource not found: type=%q name=%q mesh=%q", rt, name, mesh)
 }
```
bartsmykla, line 144:
```suggestion
type ResourceNotFoundError struct {
	rType model.ResourceType
	name  string
	mesh  string
}

func (e *ResourceNotFoundError) Error() string {
	return fmt.Sprintf("resource not found: type=%q name=%q mesh=%q", e.rType, e.name, e.mesh)
}

func (e *ResourceNotFoundError) Is(err error) bool {
	return reflect.TypeOf(err) == reflect.TypeOf(e)
}

func ErrorResourceNotFound(rt model.ResourceType, name, mesh string) error {
	return &ResourceNotFoundError{
		rType: rt,
		name:  name,
		mesh:  mesh,
	}
}
```

pkg/core/resources/store/store.go
```diff
@@ -143,12 +149,35 @@ func IsResourceNotFound(err error) bool {
 	return err != nil && strings.HasPrefix(err.Error(), "Resource not found")
 }
```
bartsmykla, line 150: I would remove this function after introducing separate  `ResourceNotFoundError`

bartsmykla (changes requested):

pkg/core/resources/store/store.go
lahabana: This is out of scope of this PR. I agree we should do this at some point and we have: https://github.com/kumahq/kuma/issues/2604 This PR is already a good size I don't want to have it even bigger

pkg/core/resources/store/store.go
lahabana: Same as previous point

pkg/core/rest/errors/error_handler.go
lahabana: Agreed it'd be good but that's out of scope of this PR

pkg/core/rest/errors/error_handler.go
lahabana: Let's be careful of scope again here.

lahabana: As for the `IsConflict` AFAIK it works because there's no variable part in Conflict so `==` will be true which is the default very of `Is`

pkg/core/resources/manager/manager.go
jakubdyszkiewicz: The reason `IsResourceAlreadyExists` was added is that you may have a conflict when you first create a resource from two places

pkg/core/rest/errors/error_handler.go
```diff
@@ -32,44 +32,12 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 			Title:  title,
 			Detail: "Not found",
 		}
-	case errors.Is(err, &rest.InvalidResourceError{}):
+	case errors.Is(err, &rest.InvalidResourceError{}) || errors.Is(err, &registry.InvalidResourceType{}) || errors.Is(err, &store.PreconditionError{}):
```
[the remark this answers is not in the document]
lahabana: I'm not sure that use `errors.As` is correct.
When you call `errors.Is` it does call `Unwrap` while `Error.Is` returns false. So you'd end up always being in a case where the would succeed.

What does `errors.As` improves here?

bartsmykla (approved):

jakubdyszkiewicz (approved):

Issue #7932, opened 2023-09-29
TrafficRoute doesn't set LB type to LeastRequest when `choiceCount` is not specified


lobkovilya:
### What happened?

The following policy doesn't change the LB type to `LEAST_REQUEST` and there are neither validation errors or errors in logs:
```yaml
type: TrafficRoute
name: route-1
mesh: mesh-1
sources:
  - match:
      kuma.io/service: 'svc-a'
destinations:
  - match:
      kuma.io/service: 'svc-b'
conf:
  loadBalancer:
    leastRequest: {}
  destination:
    kuma.io/service: 'svc-b'
```

But if `choiceCount` is specified then it works:

```yaml
type: TrafficRoute
name: route-1
mesh: mesh-1
sources:
  - match:
      kuma.io/service: 'svc-a'
destinations:
  - match:
      kuma.io/service: 'svc-b'
conf:
  loadBalancer:
    leastRequest:
      choiceCount: 2
  destination:
    kuma.io/service: 'svc-b'
```

The behavior is diverging from the docs:
> The number of random healthy hosts from which the host with the fewest active requests will be chosen. **Defaults to 2** so that we perform two-choice selection if the field is not set.

lukidzi:
I've managed to test it locally and there is an error message but after applying the policy not during applying it: 
```
[...]

PR #7938, opened 2023-10-02
fix(kuma-cp): use default value when choiceCount is 0

lukidzi:
### Checklist prior to review

Api says that when `choiceCount` is not defined we are using a default value of 2 which is correct from the envoy perspective but not from kuma API. 

`GetLeastRequest()` returns an object that has `choiceCount` equal 0, so when a user creates a policy without `choiceCount` it doesn't fail on the API but will fail on xds configuration generation because envoy requires a minimum value of 2.  
I've changed a bit how validation works now:
* validate object if the value is equal 1, that won't cause xds configuration to fail and break the sidecar
* when the value is equal to 0 we assume that it's an empty object and we set it to default 2

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/7932

### DIFF
--- a/pkg/xds/envoy/clusters/v3/lb_configurer.go
+++ b/pkg/xds/envoy/clusters/v3/lb_configurer.go
@@ -10,6 +10,8 @@ import (
 	util_proto "github.com/kumahq/kuma/pkg/util/proto"
 )
 
+const DefaultChoiceCount = 2
+
 type LbConfigurer struct {
 	Lb *mesh_proto.TrafficRoute_LoadBalancer
 }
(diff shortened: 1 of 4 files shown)


jakubdyszkiewicz (approved):

jakubdyszkiewicz:
Is it fixed in MeshLoadBalancingStrategy already?

pkg/xds/envoy/clusters/v3/lb_configurer.go
```diff
@@ -31,6 +33,11 @@ func (e *LbConfigurer) Configure(c *envoy_cluster.Cluster) error {
 		c.LbPolicy = envoy_cluster.Cluster_LEAST_REQUEST
 
 		lbConfig := e.Lb.GetLeastRequest()
+		if lbConfig == nil || lbConfig.GetChoiceCount() == 0 {
```
slonka, line 36: The way we handle things like this in new policies is that we use a pointer and detect if the value is really nil or actually set to 0. Not sure if this can be done like this in the old policies.

slonka (approved):
If we can't detect 0 vs no value like in new policies let's do this.

lukidzi:
> Is it fixed in MeshLoadBalancingStrategy already?

Yes, kubevalidator validates if the value is bigger than or equal to 2, also when an empty body it treats it as a null so it works.

pkg/xds/envoy/clusters/v3/lb_configurer.go
lukidzi: Yes, I couldn't find a way to detect if someone sets 0 or if it returned the body. In this way, we shouldn't break how the API works.

PR #7939, opened 2023-10-02
fix(kuma-cp): don't add `postStart` hook to builtin gateway even if `waitForDataplaneReady: true`

lobkovilya:
Changes:
* always pass `waitForDataplaneReady: false` when generating builtin gateway pod
* bind envoy admin to `::1` and set `waitForDataplaneReady: true` in e2e-multizone

Note: even though `postStart` hook should be completely harmless for builtin gateway for some reason it doesn't work. Builtin gateway pod spends 3m in `ContainerCreating` phase, restarts, and then works as expected. @lukidzi do you know why this can happen? 

### Checklist prior to review

### DIFF
--- a/test/framework/envs/multizone/env.go
+++ b/test/framework/envs/multizone/env.go
@@ -53,6 +53,14 @@ func SetupAndGetState() []byte {
 		},
 		framework.KumaDeploymentOptionsFromConfig(framework.Config.KumaCpConfig.Multizone.KubeZone1)...,
 	)
+	if Config.IPV6 {
+		// if the underneath clusters support IPv6, we'll configure kuma-1 with waitForDataplaneReady feature and
+		// envoy admin binding to ::1 address
+		kubeZone1Options = append(kubeZone1Options,
+			WithEnv("KUMA_RUNTIME_KUBERNETES_INJECTOR_SIDECAR_CONTAINER_WAIT_FOR_DATAPLANE_READY", "true"),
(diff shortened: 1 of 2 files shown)


test/framework/envs/multizone/env.go
```diff
@@ -53,6 +53,14 @@ func SetupAndGetState() []byte {
 		},
 		framework.KumaDeploymentOptionsFromConfig(framework.Config.KumaCpConfig.Multizone.KubeZone1)...,
 	)
+	if Config.IPV6 {
+		// if the underneath clusters support IPv6, we'll configure kuma-1 with waitForDataplaneReady feature and
+		// envoy admin binding to ::1 address
+		kubeZone1Options = append(kubeZone1Options,
+			WithEnv("KUMA_RUNTIME_KUBERNETES_INJECTOR_SIDECAR_CONTAINER_WAIT_FOR_DATAPLANE_READY", "true"),
```
lahabana, line 60: Why this?

lobkovilya: Because this feature didn't work with IPv6, also that's how I noticed we configure `waitForDataplaneReady` for builtin gateways

lahabana (approved):

lobkovilya:
I don't think we should, you can work around this by adding a `kuma.io/wait-for-dataplane-ready: false` annotation to the builtin gateway

PR #7959, opened 2023-10-04
feat(helm): add support for controlplane deployment annotations

slavogiez:
This adds ability to add annotations on the controlplane deployment.
Not sure if it's the best way to do this. Please also let me know if you'd like to get the same feature on other deployments (ingress, egress, ...)

### Checklist prior to review

### DIFF
--- a/api/mesh/v1alpha1/kds.proto
+++ b/api/mesh/v1alpha1/kds.proto
@@ -22,11 +22,9 @@ message KumaResource {
   google.protobuf.Any spec = 2;
 }
 
-message ZoneHealthCheckRequest {
-}
+message ZoneHealthCheckRequest {}
 
-message ZoneHealthCheckResponse {
-}
+message ZoneHealthCheckResponse {}
 
 service GlobalKDSService {
   // StreamXDSConfigs is logically a service exposed by Zone CP so Global CP can
--- a/app/kumactl/cmd/install/testdata/install-control-plane.cni-enabled.golden.yaml
+++ b/app/kumactl/cmd/install/testdata/install-control-plane.cni-enabled.golden.yaml
@@ -547,6 +547,8 @@ metadata:
     app: kuma-control-plane
     app.kubernetes.io/name: kuma
     app.kubernetes.io/instance: kuma
+  annotations: 
+    
 spec:
   replicas: 1
   minReadySeconds: 0
--- a/app/kumactl/cmd/install/testdata/install-control-plane.cni-legacy-enabled.golden.yaml
+++ b/app/kumactl/cmd/install/testdata/install-control-plane.cni-legacy-enabled.golden.yaml
@@ -529,6 +529,8 @@ metadata:
     app: kuma-control-plane
     app.kubernetes.io/name: kuma
     app.kubernetes.io/instance: kuma
+  annotations: 
+    
 spec:
   replicas: 1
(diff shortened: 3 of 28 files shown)


deployments/charts/kuma/templates/cp-deployment.yaml
```diff
@@ -35,6 +35,7 @@ metadata:
   name: {{ include "kuma.name" . }}-control-plane
   namespace: {{ .Release.Namespace }}
   labels: {{ include "kuma.cpLabels" . | nindent 4 }}
+  annotations: {{ include "kuma.cpDeploymentAnnotations" . | nindent 4 }}
```
michaelbeaumont, line 38: I wonder if we should do the same thing with annotations as we do with `cpLabels`, so both `Pods` and the `Deployment` get the annotations?

slavogiez: Good question. I thought about that, but in my use case (stakater reloader), I need to add annotations on the deployment only. But I agree that other use cases may need to put same annotations on resources.
My plan was to :
- use kuma.cpDeploymentAnnotations for deployment only annotations
- later use kuma.cpAnnotations for the same mechanism as kuma.cpLabels

michaelbeaumont: yeah. I think we should just have two separate annotations, one for Deployment, one for Pods.

deployments/charts/kuma/README.md
```diff
@@ -26,6 +26,7 @@ A Helm chart for the Kuma Control Plane
 | controlPlane.kdsGlobalAddress | string | `""` | Only used in `zone` mode |
 | controlPlane.replicas | int | `1` | Number of replicas of the Kuma CP. Ignored when autoscaling is enabled |
 | controlPlane.minReadySeconds | int | `0` | Minimum number of seconds for which a newly created pod should be ready for it to be considered available. |
+| controlPlane.deploymentAnnotations | object | `{}` | Control Plane Deployment Annotations |
```
michaelbeaumont, line 29: Just to make it super clear

```suggestion
| controlPlane.deploymentAnnotations | object | `{}` | Annotations applied only to the `Deployment` resource |
```

app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
```diff
@@ -51,7 +51,10 @@ controlPlane:
   # -- Minimum number of seconds for which a newly created pod should be ready for it to be considered available.
   minReadySeconds: 0
 
-  # -- Control Plane Pod Annotations
+  # -- Annotations applied only to the `Deployment` resource
+  deploymentAnnotations: {}
+
+  # -- Annotations applied only to the `Deployment` resource
```
michaelbeaumont, line 57:
```suggestion
  # -- Annotations applied only to the `Pod` resource
```

michaelbeaumont (approved):
LGTM, thanks!

michaelbeaumont:
@slavogiez is it ok if we mark this ready as review and merge? Or was there something you wanted to add/change?

slavogiez:
> @slavogiez is it ok if we mark this ready as review and merge? Or was there something you wanted to add/change?

Everything's ok. LGTM!

PR #7976, opened 2023-10-06
fix(kuma-cp): meta validation compatible with Kubernetes naming rules

lobkovilya:
### Checklist prior to review

### DIFF
--- a/pkg/api-server/resource_endpoints.go
+++ b/pkg/api-server/resource_endpoints.go
@@ -166,18 +166,21 @@ func (r *resourceEndpoints) createOrUpdateResource(request *restful.Request, res
 		return
 	}
 
-	if err := r.validateResourceRequest(request, resourceRest.GetMeta()); err != nil {
+	create := false
+	resource := r.descriptor.NewObject()
+	if err := r.resManager.Get(request.Request.Context(), resource, store.GetByKey(name, meshName)); err != nil && store.IsResourceNotFound(err) {
+		create = true
+	} else if err != nil {
+		rest_errors.HandleError(request.Request.Context(), response, err, "Could not find a resource")
+	}
+
+	if err := r.validateResourceRequest(request, resourceRest.GetMeta(), create); err != nil {
 		rest_errors.HandleError(request.Request.Context(), response, err, "Could not process a resource")
 		return
 	}
 
-	resource := r.descriptor.NewObject()
-	if err := r.resManager.Get(request.Request.Context(), resource, store.GetByKey(name, meshName)); err != nil {
-		if store.IsResourceNotFound(err) {
-			r.createResource(request.Request.Context(), name, meshName, resourceRest.GetSpec(), response)
-		} else {
-			rest_errors.HandleError(request.Request.Context(), response, err, "Could not find a resource")
-		}
+	if create {
+		r.createResource(request.Request.Context(), name, meshName, resourceRest.GetSpec(), response)
 	} else {
 		r.updateResource(request.Request.Context(), resource, resourceRest.GetSpec(), response)
 	}
@@ -272,7 +275,7 @@ func (r *resourceEndpoints) deleteResource(request *restful.Request, response *r
 	}
 }
 
-func (r *resourceEndpoints) validateResourceRequest(request *restful.Request, resourceMeta rest_v1alpha1.ResourceMeta) error {
+func (r *resourceEndpoints) validateResourceRequest(request *restful.Request, resourceMeta rest_v1alpha1.ResourceMeta, create bool) error {
 	var err validators.ValidationError
 	name := request.PathParameter("name")
 	meshName := r.meshFromRequest(request)
@@ -288,7 +291,11 @@ func (r *resourceEndpoints) validateResourceRequest(request *restful.Request, re
 	if r.descriptor.Scope == model.ScopeMesh && meshName != resourceMeta.Mesh {
 		err.AddViolation("mesh", "mesh from the URL has to be the same as in body")
 	}
-	err.AddError("", mesh.ValidateMeta(name, meshName, r.descriptor.Scope))
+	if create {
+		err.AddError("", mesh.ValidateMeta(resourceMeta, r.descriptor.Scope))
+	} else {
+		err.AddError("", mesh.ValidateMetaBackwardsCompatible(resourceMeta, r.descriptor.Scope))
+	}
 	return err.OrNil()
 }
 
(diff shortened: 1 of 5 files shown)


lahabana (commented):
Missing upgrade.md note btw

pkg/api-server/testdata/resource_name-mesh.golden.json
```diff
@@ -6,22 +6,22 @@
  "invalid_parameters": [
   {
    "field": "name",
-   "reason": "invalid characters. Valid characters are numbers, lowercase latin letters and '-', '_' symbols."
+   "reason": "invalid characters. A lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character"
   },
   {
    "field": "mesh",
-   "reason": "invalid characters. Valid characters are numbers, lowercase latin letters and '-', '_' symbols."
+   "reason": "invalid characters. Valid characters are numbers, lowercase latin letters and '-', '_', '.' symbols."
```
lahabana, line 13: can a mesh have `_`?

pkg/core/resources/apis/mesh/meta_validator.go
```diff
@@ -3,33 +3,70 @@ package mesh
...
 )
 
-var nameMeshRegexp = regexp.MustCompile("^[0-9a-z-_]*$")
+var (
+	backwardCompatRegexp = regexp.MustCompile(`^[0-9a-z-_.]*$`)
+	backwardCompatErrMsg = "invalid characters. Valid characters are numbers, lowercase latin letters and '-', '_', '.' symbols."
+)
+
+var (
+	identifierRegexp = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
```
lahabana, line 16: Isn't there a max size too?

pkg/api-server/testdata/resource_name-mesh.golden.json
lobkovilya: Yes, but you won't be able to create new meshes with `_` (because when you create new mesh you actually check `name`)

pkg/core/resources/apis/mesh/meta_validator.go
lobkovilya: If I understand https://github.com/kumahq/kuma/pull/6123/files correctly we rely on the store to validate the length

pkg/api-server/resource_endpoints.go
```diff
@@ -288,7 +291,15 @@ func (r *resourceEndpoints) validateResourceRequest(request *restful.Request, re
...
 		err.AddViolation("mesh", "mesh from the URL has to be the same as in body")
 	}
-	err.AddError("", mesh.ValidateMeta(name, meshName, r.descriptor.Scope))
+	if create {
+		err.AddError("", mesh.ValidateMeta(resourceMeta, r.descriptor.Scope))
+	} else {
+		if verr, msg := mesh.ValidateMetaBackwardsCompatible(resourceMeta, r.descriptor.Scope); verr.HasViolations() {
+			err.AddError("", verr)
+		} else if msg != "" {
+			core.Log.Info(msg)
```
michaelbeaumont, line 300: should there be more context added to this log message?

lobkovilya: Now it looks like:
> [WARNING] Kuma 2.5.x introduces more strict validation rules for resource name, it must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character. You can't CREATE resources with invalid names anymore, but it's possible to UPDATE such resources until Kuma 2.7.x is released.

What would you like to add to this message?

michaelbeaumont: isn't it just alone? Should we not indicate which resource is triggering it?

lobkovilya: Right, let me add this

michaelbeaumont (approved):

pkg/core/resources/apis/mesh/meta_validator.go
lahabana: Would it make sense to have api level validation? Or just a recipe for disaster?

lobkovilya: We don't have issues with validating name length at this moment (it's already compatible with Kubernetes), so I don't think it's necessary to include in this PR. But I can add it, should be a small change

pkg/api-server/resource_endpoints.go
```diff
@@ -288,7 +291,15 @@ func (r *resourceEndpoints) validateResourceRequest(request *restful.Request, re
...
 		err.AddViolation("mesh", "mesh from the URL has to be the same as in body")
 	}
-	err.AddError("", mesh.ValidateMeta(name, meshName, r.descriptor.Scope))
+	if create {
+		err.AddError("", mesh.ValidateMeta(resourceMeta, r.descriptor.Scope))
+	} else {
+		if verr, msg := mesh.ValidateMetaBackwardsCompatible(resourceMeta, r.descriptor.Scope); verr.HasViolations() {
+			err.AddError("", verr)
+		} else if msg != "" {
+			core.Log.Info(msg, "type", r.descriptor.Name, "mesh", resourceMeta.Mesh, "name", resourceMeta.Name)
```
jakubdyszkiewicz, line 300: we should not use unnamed logs. It does not map well in Datadog. I see there is `log` in the package. Additionally, technically we should use `AddFieldsFromCtx`, but I imagine we most likely won't hit this issue with Mesh in Konnect, won't we?

pkg/core/resources/apis/mesh/meta_validator.go
jakubdyszkiewicz: +1 to add it

pkg/api-server/resource_endpoints.go
lobkovilya: Changed this to use `log`. There isn't much we can do even if we see the message in Konnect, that's down to the user to rename their policies

pkg/core/resources/apis/mesh/meta_validator.go
lobkovilya: Added. We actually handled this pretty badly on the client side:
```
$ kumactl apply -f resource_with_long_name.yaml
Error: Could not create a resource (Internal Server Error)
```

jakubdyszkiewicz (approved):

Issue #6570, opened 2023-04-19
Use InterCP Catalog to finalize proxy subscriptions


lobkovilya:
### Description

On Universal, Kuma tracks stale DPP resources in the store by checking the `generation` counter. When DPP is connected to some CP this counter is increasing every N seconds. So if it doesn't change for 2*N seconds then the finalizer marks DPP as Offline.

Recently we introduces InterCP API, so now we have a catalog of all running CP instances. This means finalizer can figure out if the DPP resource is stale by checking `ControlPlaneInstanceId` in the catalog.

lobkovilya:
There was [attempt](https://github.com/kumahq/kuma/pull/7981) to use catalog-based approach in finalizer, but there is a blocker https://github.com/kumahq/kuma/issues/7467. We don't want to introduce catalog-based approach under the feature flag, because it brings even more complexity.

lukidzi:
> Triage: we have a gc finalizer which does a cleanup is it required?

slonka:
@lobkovilya - is this still relevant?

slonka:
> Triage: it's still relevant, not sure if this is going to be prioritised

Issue #7467, opened 2023-08-04
Provide a way to have active-active global control plane clusters


brentos:
### Description

As it currently stands, a universal Global Control Plane (GCP) can be scaled to multiple instances in a Kubernetes cluster and one will become leader by getting a database lock. The others will be followers and they all have the ability to communicate with each other via inter-cp GRPC connection. By default they communicate their IPs and everything works as expected. However, in a multi-cluster / multi-region GCP scenario, we must somehow create the inter-cp connection by providing KUMA_INTER_CP_CATALOG_INSTANCE_ADDRESS -- this could be a load balancer address, and as long as there is only 1 replica in each cluster, then this should work as expected. However, this breaks down when there are multiple replicas in multiple clusters, since we have no way of directly accessing the leader, and the LB will balance between all replicas (not what we want). There needs to be some way to accomplish this in order to support multiple clusters and multiple replicas per zone.

lahabana:
In the case of kubernetes usually people run using stateful set and add a LB with a single endpoint.

brentos:
That is a valid approach but there is no way to configure each instance properly, right? The "instanceAddress" needs to be populated with its LB address.

slonka:
Re-pinging @lahabana / @justindavies since there is a `mgmt` label.

PR #7981, opened 2023-10-09
feat(kuma-cp): ditch `generation` approach in finalizer and rely on CP catalog 

lobkovilya:
The only reason to have a `generation` counter was to detect if the CP instance is gone and the counter is not updating anymore. Now that we have Inter CP API with CP catalog we can detect this in a much simpler way.

I'm still on the edge if we should let both approaches co-exist for a few releases. But as far as I can tell there are no advantages of the `generation` approach so seems unnecessary to keep it.

### TODO

- [ ] Should we keep the `generation` approach and make a new one as `experimental`?

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/6570


pkg/gc/finalizer.go
```diff
@@ -150,32 +152,48 @@ func (f *subscriptionFinalizer) checkGeneration(ctx context.Context, typ core_mo
...
-		log.V(1).Info("mark subscription as disconnected")
-		insight.GetLastSubscription().SetDisconnectTime(now)
-
-		upsertInsight, _ := registry.Global().NewObject(typ)
-		err := manager.Upsert(ctx, f.rm, key, upsertInsight, func(r core_model.Resource) error {
-			return upsertInsight.GetSpec().(generic.Insight).UpdateSubscription(insight.GetLastSubscription())
-		})
-		if err != nil {
-			log.Error(err, "unable to finalize subscription")
+		if exist, err := f.instanceExists(ctx, insight.GetLastSubscription().GetInstanceId()); err != nil {
```
lahabana, line 164: Do we have guarantee that the `InstanceId` is always unique?

pkg/gc/finalizer.go
```diff
@@ -150,32 +152,48 @@ func (f *subscriptionFinalizer) checkGeneration(ctx context.Context, typ core_mo
...
+			}
+			delete(f.insights[tenantID(tenantId)][typ], key)
 		}
-		delete(f.insights[tenantID(tenantId)][typ], key)
 	}
 	return nil
 }
 
+func (f *subscriptionFinalizer) instanceExists(ctx context.Context, id string) (bool, error) {
+	instances, err := f.catalog.Instances(ctx)
```
lahabana, line 185: Feels like this should be a helper in the catalog (or in the catalog package at least)

lahabana (approved):

pkg/gc/finalizer.go
jakubdyszkiewicz:
```
func InstanceOfID(ctx context.Context, catalog Catalog, id string) (Instance, error) {
	instances, err := catalog.Instances(ctx)
	if err != nil {
		return Instance{}, err
	}
	for _, instance := range instances {
		if instance.Id == id {
			return instance, nil
		}
	}
	return Instance{}, ErrInstanceNotFound
}
```
there is this ^

jakubdyszkiewicz:
Looks good at first glance, but I have one problem with this approach. This issue https://github.com/kumahq/kuma/issues/7467

I'm not sure we should merge it until we solve it

lobkovilya:
@jakubdyszkiewicz I don't think #7467 will be solved in the near future, I see 2 options for how we can merge this PR:
* put catalog-based finalized under a feature flag
* make catalog-based finalizer default, but keep a way to enable generation-based finalizer if needed (i.e. when you have multiple CP replicas in multiple clusters)

lobkovilya:
Closing while it's not clear what to do with https://github.com/kumahq/kuma/issues/7467

PR #7995, opened 2023-10-10
perf(store): postgres transactions

jakubdyszkiewicz:
### Checklist prior to review

Introducing postgres transactions.
This PR does NOT introduce need for transactions in Kuma, this is just a perf optimization for stores that support transactions.

#### Enrich ResourceStore vs new interface
I introduced a new interface called Transactions, so the transaction can be used by either ResourceStore and ResourceManager. If I would add a new method in ResourceStore we would also add this to ResourceManager.

#### Transaction in Context vs store opts
I decided to propagate transaction implicitly in context instead of explicitly in store opts like `CreateInTx(tx)`. This is more flexible with less changes to the code. We could also for example open a transaction when we create new Mesh without adding `CreateInTx` to every call.

### DIFF
--- a/pkg/core/resources/store/transactions.go
+++ b/pkg/core/resources/store/transactions.go
@@ -1,0 +1,73 @@
+package store
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"go.uber.org/multierr"
+)
+
+type txCtx struct{}
+
+func CtxWithTx(ctx context.Context, tx Transaction) context.Context {
+	return context.WithValue(ctx, txCtx{}, tx)
+}
+
+func TxFromCtx(ctx context.Context) (Transaction, bool) {
+	if value, ok := ctx.Value(txCtx{}).(Transaction); ok {
+		return value, true
+	}
+	return nil, false
+}
+
+type Transaction interface {
+	Commit(ctx context.Context) error
+	Rollback(ctx context.Context) error
+}
+
+// Transactions is a producer of transactions if a resource store support transactions.
+// Transactions are never required for consistency in Kuma, because there are ResourceStores that does not support transactions.
+// However, in a couple of cases executing queries in transaction can improve the performance.
+//
+// In case of Postgres, you may set hooks when retrieve and release connections for the connection pool.
+// In this case, if you create multiple resources, you need to acquire connection and execute hooks for each create.
+// If you create transaction for it, you execute hooks once and you hold the connection.
+//
+// Transaction is propagated implicitly through Context.
+type Transactions interface {
+	Begin(ctx context.Context) (Transaction, error)
+}
+
+func InTx(ctx context.Context, transactions Transactions, fn func(ctx context.Context) error) error {
+	tx, err := transactions.Begin(ctx)
+	if err != nil {
+		return err
+	}
+	if err := fn(CtxWithTx(ctx, tx)); err != nil {
+		if rollbackErr := tx.Rollback(ctx); rollbackErr != nil {
+			return multierr.Append(errors.Wrap(rollbackErr, "could not rollback transaction"), err)
+		}
+		return err
+	}
+	return tx.Commit(ctx)
+}
+
+type NoopTransaction struct{}
+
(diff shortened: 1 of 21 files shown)


pkg/plugins/resources/postgres/pgx_store.go
```diff
@@ -465,3 +518,7 @@ func registerMetrics(metrics core_metrics.Metrics, pool *pgxpool.Pool, poolName
 	}
 	return nil
 }
+
+func (r *pgxResourceStore) Begin(ctx context.Context) (store.Transaction, error) {
+	return r.pool.Begin(ctx)
```
lobkovilya, line 523: IIUC the default IsoLevel is `read committed`, but the only thing we're trying to achieve here is batching – executing several operation on the same acquired connection. So it's probably safe to lower the IsoLevel to `read uncommitted`.

pkg/core/resources/store/transactions.go
```diff
@@ -0,0 +1,73 @@
...
+// Transactions are never required for consistency in Kuma, because there are ResourceStores that does not support transactions.
+// However, in a couple of cases executing queries in transaction can improve the performance.
+//
+// In case of Postgres, you may set hooks when retrieve and release connections for the connection pool.
+// In this case, if you create multiple resources, you need to acquire connection and execute hooks for each create.
+// If you create transaction for it, you execute hooks once and you hold the connection.
+//
+// Transaction is propagated implicitly through Context.
+type Transactions interface {
+	Begin(ctx context.Context) (Transaction, error)
```
lobkovilya, line 38: Do we even need transactions? I think we can introduce some kind of "batching" purely on Kuma side – acquire connection in `Begin` and execute all the following operation using the same connection. DB doesn't need to know about these "batches" at all

jakubdyszkiewicz: Not sure if I get your proposal here. Can you elaborate? What would it look like?

lobkovilya: In `Begin()` method, instead of:
```go
return r.pool.Begin(ctx)
```
we can do:
```go
return r.pool.Acquire(ctx) 
```

After that, method `InTx()` puts connection into ctx (instead of putting transaction into ctx).

So later `pgxResourceStore` checks if there is a connection in the context and uses this connection to call `Exec`.

That way a number of store operations are executed on the same connection, but Postgres doesn't know about it.

jakubdyszkiewicz: According to multiple sources, doing this in one transaction is more optimal on postgres CPU. 

https://www.postgresql.org/docs/current/populate.html
> 14.4.1. Disable Autocommit 
When using multiple INSERTs, turn off autocommit and just do one commit at the end. (In plain SQL, this means issuing BEGIN at the start and COMMIT at the end. Some client libraries might do this behind your back, in which case you need to make sure the library does it when you want it done.) If you allow each insertion to be committed separately, PostgreSQL is doing a lot of work for each row that is added. An additional benefit of doing all insertions in one transaction is that if the insertion of one row were to fail then the insertion of all rows inserted up to that point would be rolled back, so you won't be stuck with partially loaded data.

pkg/plugins/resources/postgres/pgx_store.go
jakubdyszkiewicz: I cannot find proof that it improves perf 🤔 Do you have sources for this?

lobkovilya: Normally that'd increase concurrency, but I see there is no difference in PostgreSQL 😞 

https://www.postgresql.org/docs/current/transaction-iso.html

> In PostgreSQL, you can request any of the four standard transaction isolation levels, but internally only three distinct isolation levels are implemented, i.e., PostgreSQL's Read Uncommitted mode behaves like Read Committed. This is because it is the only sensible way to map the standard isolation levels to PostgreSQL's multiversion concurrency control architecture.

lobkovilya (approved):

PR #7999, opened 2023-10-10
feat(api-server): add /-overview for all types that have overviews

### LINKED ISSUE #3336: Normalise apis for insights and overview
### What happened?

There is `/dataplane` and `/dataplane-insights`. The `/dataplane+insights` was supposed to be a combination of dataplane + insight therefore overview. 
Maybe change it to `/dataplane-overview`

lahabana:
There used to be some incoherency with overview endpoints they were also all reimplemented from scratch and would need us to reimplement filtering etc each time.

We now leverage our code generation to indicate which type have a `overview` we do all this with backward compatibility on the api

Fix #3336

### Checklist prior to review


pkg/core/resources/model/resource.go
```diff
@@ -122,13 +131,17 @@ type ResourceTypeDescriptor struct {
 	HasFromTargetRef bool
 	// Schema contains an unmarshalled OpenAPI schema of the resource
 	Schema *spec.Schema
+	// Insight the insights type attached to this resourceType
```
bartsmykla, line 134:
```suggestion
	// Insight contains the insight type attached to this resourceType
```

pkg/core/resources/model/resource.go
```diff
@@ -122,13 +131,17 @@ type ResourceTypeDescriptor struct {
...
+	Insight Resource
+	// Overview the overview type attached to this resourceType
```
bartsmykla, line 136:
```suggestion
	// Overview contains the overview type attached to this resourceType
```

pkg/api-server/resource_endpoints.go
```diff
@@ -43,36 +45,85 @@ type resourceEndpoints struct {
...
-	} else {
+		resource := r.descriptor.NewObject()
+		if err := r.resManager.Get(request.Request.Context(), resource, store.GetByKey(name, meshName)); err != nil {
+			rest_errors.HandleError(request.Request.Context(), response, err, "Could not retrieve a resource")
+			return
+		}
+		if withInsight {
+			insight := r.descriptor.NewInsight()
+			if err := r.resManager.Get(request.Request.Context(), insight, store.GetByKey(name, meshName)); err != nil && !store.IsResourceNotFound(err) {
+				rest_errors.HandleError(request.Request.Context(), response, err, "Could not retrieve insights")
```
bartsmykla, line 110:
```suggestion
				rest_errors.HandleError(request.Request.Context(), response, err, "Could not retrieve insights")
				return
```

bartsmykla (approved):

PR #8039, opened 2023-10-12
fix(api): 400 error on admin operations on not yet connected stream

slonka:
return 400 error on requests trying to execute admin operations on not yet connected stream

### Checklist prior to review


pkg/intercp/envoyadmin/forwarding_kds_client.go
```diff
@@ -182,3 +183,11 @@ func (f *forwardingKdsEnvoyAdminClient) clientForInstanceID(ctx context.Context,
...
 	return f.newClientFn(instance.InterCpURL())
 }
+
+type StreamNotConnectedError struct {
+	rpcName string
+}
+
+func (e StreamNotConnectedError) Error() string {
+	return fmt.Sprintf("stream to execute %s operations is not yet connected", e.rpcName)
+}
```
lahabana, line 193: You need to define is for this your fix to work as otherwise it's going to do `==` comp and not match

lahabana:
Also maybe https://github.com/kumahq/kuma/pull/7937 will influence what you do here

pkg/intercp/envoyadmin/forwarding_kds_client.go
slonka: Done

jakubdyszkiewicz (approved):

Issue #7728, opened 2023-09-11
Value `not: true` in tags ignored during reference matching 


bartsmykla:
### What happened?

when generating header matching configs for envoy's fault filters, we are ignoring tag's value: not: true (i.e. { "key": "kuma.io/service", "value": "edge-gateway", "not": true }) which basically means for some of the fault filter configs we are generating inverted regexes.

This is the reason behind problem described here: https://github.com/kumahq/kuma/issues/6630

lahabana:
> Triage: The problem is when doing service matching to figure if a policy applies to a service subset.

Issue #6630, opened 2023-04-26
Duplicated fault filter in envoy config with two MeshFaultInjection policies targeting MeshGateway


bartsmykla:
### What happened?

I deployed two `MeshFaultInjection` policies:

<details>
<summary>First: <code>targetRef: MeshService#demo-app_gateway</code> + <code>from.targetRef: Mesh</code> / percentage: <code>55</code></summary>

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshFaultInjection
metadata:
  name: mfi
  namespace: kuma-system
spec:
  targetRef:
    kind: MeshService
    name: demo-app_gateway
  from:
    - targetRef:
        kind: Mesh
      default:
        http:
          - abort:
              httpStatus: 500
              percentage: 55
```
</details>
<details>
<summary>Second: <code>targetRef: Mesh</code> + <code>from.targetRef: MeshService#demo-app_gateway</code> / percentage: <code>77</code></summary>

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshFaultInjection
metadata:
  name: mfi-2
  namespace: kuma-system
spec:
  targetRef:
    kind: Mesh
  from:
    - targetRef:
        kind: MeshService
        name: demo-app_gateway
      default:
        http:
          - abort:
              httpStatus: 500
              percentage: 77
```
</details>

and when looking at config dump we have (this is just an excerpt of the listener with less important fields removed, for clarity):
```json
{
  "name": "demo-app:HTTP:80",
  "activeState": {
    "listener": {
      "@type": "type.googleapis.com/envoy.config.listener.v3.Listener",
      "name": "demo-app:HTTP:80",
      "address": {
        "socketAddress": {
          "address": "10.42.0.9",
          "portValue": 80
        }
      },
      "filterChains": [
        {
          "filters": [
            {
              "name": "envoy.filters.network.http_connection_manager",
              "typedConfig": {
                "@type": "type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager",
                "statPrefix": "demo-app_gateway",
                "rds": {
                  "routeConfigName": "demo-app:HTTP:80"
                },
                "httpFilters": [
                  {
                    "name": "envoy.filters.http.fault",
                    "typedConfig": {
                      "@type": "type.googleapis.com/envoy.extensions.filters.http.fault.v3.HTTPFault",
                      "abort": {
                        "httpStatus": 500,
                        "percentage": {
[...]

PR #8043, opened 2023-10-13
fix(MeshFaultInjection): include tags negation in header matching

bartsmykla:
Take into account tags negation when generating header matchers for MeshFaultInjection policy.

### Checklist prior to review

  - Closes: https://github.com/kumahq/kuma/issues/7728
  - It won't
  - Unit tests updated
  - There is no need
  - It doesn't

### DIFF
--- a/pkg/plugins/policies/meshfaultinjection/plugin/v1alpha1/plugin_test.go
+++ b/pkg/plugins/policies/meshfaultinjection/plugin/v1alpha1/plugin_test.go
@@ -116,32 +116,61 @@ var _ = Describe("MeshFaultInjection", func() {
 			},
 			fromRules: core_rules.FromRules{
 				Rules: map[core_rules.InboundListener]core_rules.Rules{
-					{Address: "127.0.0.1", Port: 17777}: {{
-						Subset: core_rules.Subset{
-							{
-								Key:   "kuma.io/service",
-								Value: "demo-client",
-							},
-						},
-						Conf: api.Conf{
-							Http: &[]api.FaultInjectionConf{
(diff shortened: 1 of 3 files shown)


pkg/plugins/policies/meshfaultinjection/plugin/v1alpha1/plugin_test.go
```diff
@@ -116,32 +116,61 @@ var _ = Describe("MeshFaultInjection", func() {
...
-						Subset: core_rules.Subset{
-							{
-								Key:   "kuma.io/service",
-								Value: "demo-client",
+					{Address: "127.0.0.1", Port: 17777}: {
+						{
+							Subset: core_rules.Subset{
+								{
+									Key:   "kuma.io/service",
+									Value: "demo-client",
```
slonka, line 124: so this is not targeted because the name of this service is `backend` right?

slonka (approved):
Looks good but should we also have a test similar to this one https://github.com/kumahq/kuma/issues/6630 that has two MeshFaultInjections?

pkg/plugins/policies/meshfaultinjection/plugin/v1alpha1/plugin_test.go
bartsmykla: It's inbound policy, so MeshFaultInjection will configure faults on `backend`s listener for the traffic comming from `demo-client` I believe.

bartsmykla:
> Looks good but should we also have a test similar to this one #6630 that has two MeshFaultInjections?

Existing test assumes the policies were already merged, that's why we have two `fromRules`, so I believe we are already testing this case.

PR #8046, opened 2023-10-15
fix(shutdown): try forceful stop after 3 seconds

slonka:
### Checklist prior to review

### DIFF
--- a/pkg/kds/mux/server.go
+++ b/pkg/kds/mux/server.go
@@ -158,9 +158,15 @@ func (s *server) Start(stop <-chan struct{}) error {
 
 	select {
 	case <-stop:
+		timer := time.AfterFunc(3 * time.Second, func() {
+			muxServerLog.Info("stopping forcefully")
+			grpcServer.Stop()
+			muxServerLog.Info("stopped forcefully")
+		})
+		defer timer.Stop()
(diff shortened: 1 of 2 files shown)


pkg/kds/mux/server.go
```diff
@@ -158,9 +158,15 @@ func (s *server) Start(stop <-chan struct{}) error {
...
+		timer := time.AfterFunc(3 * time.Second, func() {
+			muxServerLog.Info("stopping forcefully")
+			grpcServer.Stop()
+			muxServerLog.Info("stopped forcefully")
+		})
+		defer timer.Stop()
 		muxServerLog.Info("stopping gracefully")
 		grpcServer.GracefulStop()
-		muxServerLog.Info("stopped")
+		muxServerLog.Info("stopped gracefully")
```
michaelbeaumont, line 169: Just FYI, this will race with https://github.com/kumahq/kuma/blob/8c3eabe51bd54c29ea4f5ecc7f22aa3912fd313b/app/kuma-cp/cmd/run.go#L152

Also `"forcefully stopped"` is very common, so we're hitting that timeout above. But I haven't been able to actually observe this `GracefulStop` blocking yet. I see `"stopped"` from this component. Seems like a different component is leading to `"forcefully stopped"`

PR #8051, opened 2023-10-16
fix(perf): put the Gatewaylisteners in the Proxy

lahabana:
We were constantly recomputing these.
This was making code hard to read and proxy had some memory impact.

We also introduce a plugin API for proxy modification which helps avoiding circular dependency between core_xds and plugins

### Checklist prior to review


pkg/plugins/runtime/gateway/plugin.go
```diff
@@ -34,6 +38,28 @@ func (p *plugin) BeforeBootstrap(context *core_plugins.MutablePluginContext, con
...
+		return err
+	}
+	proxy.RuntimeExtensions[metadata.PluginName] = l
+	return nil
+}
+
+func ExtractGatewayListener(proxy *core_xds.Proxy) []GatewayListenerInfo {
+	ext := proxy.RuntimeExtensions[metadata.PluginName]
+	if ext == nil {
+		return nil
```
Automaat, line 58: shouldn't this function return empty list instead of nil?

pkg/plugins/runtime/gateway/plugin.go
```diff
@@ -34,6 +38,28 @@ func (p *plugin) BeforeBootstrap(context *core_plugins.MutablePluginContext, con
...
+		ctx, &meshContext, proxy,
+	)
+	if err != nil {
+		return err
+	}
+	proxy.RuntimeExtensions[metadata.PluginName] = l
+	return nil
+}
+
+func ExtractGatewayListener(proxy *core_xds.Proxy) []GatewayListenerInfo {
```
Automaat, line 55: wouldn't it be better to have this function on proxy struct? Then we won't need to import gateway plugin in all policies plugins and you can simply do: `proxy.ExtractGatewayListener()` in policy.

pkg/plugins/runtime/gateway/plugin.go
```diff
@@ -34,6 +38,28 @@ func (p *plugin) BeforeBootstrap(context *core_plugins.MutablePluginContext, con
...
+	ext := proxy.RuntimeExtensions[metadata.PluginName]
```
Automaat, line 56:
```suggestion
func ExtractGatewayListeners(proxy *core_xds.Proxy) []GatewayListenerInfo {
```

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin_test.go
```diff
@@ -460,6 +460,10 @@ var _ = Describe("MeshTimeout", func() {
 					}},
 				},
 			},
+			RuntimeExtensions: map[string]interface{}{},
+		}
+		for n, p := range core_plugins.Plugins().ProxyPlugins() {
```
Automaat, line 465: should we run tese plugins in test? In my opinion we should explicitely provide `RuntimeExtension` in test, since we are not testing proxy plugin in this test

pkg/plugins/runtime/gateway/plugin.go
lahabana: I'm always unsure with arrays in go as empty array and nil is roughly the same thing...

pkg/plugins/runtime/gateway/plugin.go
lahabana: That sounds interesting. I thought a little with circular dependency which is why I think I ended up doing this this way but let me double check.

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin_test.go
lahabana: Yes that's rigth!

lahabana: Ok it was actually hard to redo it simply so I've only loaded the one gateway plugin instead of all of them

pkg/plugins/runtime/gateway/plugin.go
lahabana: Yeah it's functionnaly equivalent: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices
I think nil is easier to read

pkg/plugins/runtime/gateway/plugin.go
lahabana: Yes so I confirm what I said there's an issue with circular dep also the idea of proxy plugins is that they are type independent so it looks to me that we should keep this kind of method in the proxy plugin

lahabana:
@Automaat updated thx!

Automaat (approved):

PR #8022, opened 2023-10-11
docs(madr): add Locality Aware load balancing madr

Automaat:
### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/2689
- [x] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [x] Tests (Unit test, E2E tests, manual test on universal and k8s) --
- [x] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [x] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) --
- [x] Do you need to explicitly set a [`> Changelog:` entry here](../blob/master/CONTRIBUTING.md#submitting-a-patch) or add a `ci/` label to run fewer/more tests?

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #8082, opened 2023-10-20
feat(policy): extend meshloadbalancingstrategy with new locality aware api

Automaat:
### Checklist prior to review


pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go
```diff
@@ -38,8 +39,75 @@ type LocalityAwareness struct {
...
+	// LocalZone defines locality aware load balancing priorities between dataplane proxies inside a zone
+	LocalZone *LocalZone `json:"localZone,omitempty"`
+	// CrossZone defines locality aware load balancing priorities when dataplane proxies inside local zone
+	// are unavailable
+	CrossZone *CrossZone `json:"crossZone,omitempty"`
+	// FailoverThreshold defines the percentage of live destination dataplane proxies below which load balancing to the
+	// next priority starts.
+	// Example: If you configure failoverThreshold to 70, and you have deployed 10 destination dataplane proxies.
+	// Load balancing to next priority will start when number of live destination dataplane proxies drops below 7.
+	// Default 70
```
lukidzi, line 51: Should be 50

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go
```diff
@@ -38,8 +39,75 @@ type LocalityAwareness struct {
...
+}
+
+type AffinityTag struct {
+	// Key defines tag for which affinity is configured
+	Key string `json:"key"`
+	// Weight of the tag used for load balancing. The bigger the weight the bigger the priority.
+	// Percentage of local traffic load balanced to tag is computed by dividing weight by sum of weights from all tags.
+	// For example with two affinity tags first with weight 80 and second with weight 20,
+	// then 80% of traffic will be redirected to the first tag, and 20% of traffic will be redirected to second one.
+	// Setting weights is not mandatory. When weights are not set control plane will compute default weight based on list order.
```
lukidzi, line 67: We can add info about defaults 90%, 9% .....

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go
```diff
@@ -38,8 +39,75 @@ type LocalityAwareness struct {
...
+	Weight *int32 `json:"weight"`
```
lukidzi, line 68: should we change it to uint32?

Automaat: changed

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go
Automaat: added

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go
Automaat: changed

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go
lukidzi: Maybe we can say:

```
// Default: If you don't specify weight we will set that 90% of the traffic goes to the first entry, 9% to the next, and lower for another.
```

We could even give a math rule: `max(1, 9* 10^(len of rules - rule index))`

app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
```diff
@@ -2046,11 +2046,116 @@ spec:
...
+                                    type: object
+                                  type: array
+                              type: object
                             disabled:
                               description: Disabled allows to disable locality-aware
                                 load balancing. When disabled requests are distributed
                                 across all endpoints regardless of locality.
                               type: boolean
+                            failoverThreshold:
+                              description: 'FailoverThreshold defines the percentage
```
slonka, line 2102: nitpick: other descriptions don't start with `'`

app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
```diff
@@ -2046,11 +2046,116 @@ spec:
...
+                                  description: AffinityTags list of tags for local
+                                    zone load balancing.
+                                  items:
+                                    properties:
+                                      key:
+                                        description: Key defines tag for which affinity
+                                          is configured
+                                        type: string
+                                      weight:
+                                        description: 'Weight of the tag used for load
```
slonka, line 2133: nitpick: starts with `'` as well

app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
```diff
@@ -2046,11 +2046,116 @@ spec:
...
+                                          load balanced to tag is computed by dividing
+                                          weight by sum of weights from all tags.
+                                          For example with two affinity tags first
+                                          with weight 80 and second with weight 20,
+                                          then 80% of traffic will be redirected to
+                                          the first tag, and 20% of traffic will be
+                                          redirected to second one. Setting weights
+                                          is not mandatory. When weights are not set
+                                          control plane will compute default weight
+                                          based on list order. Default: If you don''t
```
slonka, line 2145: nitpick:

```suggestion
                                          based on list order. Default: If you don't
```

slonka:
Madr: https://github.com/kumahq/kuma/pull/8022/files

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -49,6 +50,88 @@ func validateTo(to []To) validators.ValidationError {
...
+	var verr validators.ValidationError
+	if localZone == nil {
+		return verr
+	}
+
+	var weightSpecified int
+	for idx, affinityTag := range localZone.AffinityTags {
+		path := validators.RootedAt("affinityTags").Index(idx)
+		if affinityTag.Weight != nil {
+			verr.Add(validators.ValidateIntegerGreaterThan(path.Field("weight"), *affinityTag.Weight, 1))
```
slonka, line 78: so the weight can't be 1? I thought that it has to be > 0

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -49,6 +50,88 @@ func validateTo(to []To) validators.ValidationError {
...
+			weightSpecified++
+		}
+	}
+
+	if weightSpecified > 0 && weightSpecified != len(localZone.AffinityTags) {
+		verr.AddViolation("affinityTags", "each or none affinity tags should have weight")
```
slonka, line 84: not a native speaker but I think it should be `all or none`

```suggestion
		verr.AddViolation("affinityTags", "all or none affinity tags should have weight")
```

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -49,6 +50,88 @@ func validateTo(to []To) validators.ValidationError {
...
+	}
+	return verr
```
slonka, line 86: should we be validating `.key` as well? what if the key is `nil`? what if it's an empty string?

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -49,6 +50,88 @@ func validateTo(to []To) validators.ValidationError {
...
+	var verr validators.ValidationError
+	if crossZone == nil {
+		return verr
+	}
+
+	for idx, failover := range crossZone.Failover {
+		path := validators.RootedAt("failover").Index(idx)
+		if failover.From != nil {
+			if len(failover.From.Zones) == 0 {
+				verr.AddViolationAt(path.Field("from").Field("zones"), "must not be empty")
```
slonka, line 99: should probably use https://github.com/kumahq/kuma/blob/95ab5c494973eb62d74184c75f4ee21c8f2ff569/pkg/core/validators/messages.go#L11

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -49,6 +50,88 @@ func validateTo(to []To) validators.ValidationError {
...
+			}
```
slonka, line 100: should we check that individual zones are not empty? like 

```
zones: ["", "zone-2"]
```

or maybe even weirder like:

```
zones: [nil, "zone-2"]
```

?

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -49,6 +50,88 @@ func validateTo(to []To) validators.ValidationError {
...
+		toZonesPath := path.Field("to").Field("zones")
+		switch failover.To.Type {
+		case None:
+			if len(failover.To.Zones) > 0 {
+				verr.AddViolationAt(toZonesPath, fmt.Sprintf("must be empty when type is %s", failover.To.Type))
+			}
+		case Any:
+			if len(failover.To.Zones) > 0 {
+				verr.AddViolationAt(toZonesPath, fmt.Sprintf("must be empty when type is %s", failover.To.Type))
+			}
```
slonka, line 112: I think you can collapse these into one:

```suggestion
		case None, Any:
			if len(failover.To.Zones) > 0 {
				verr.AddViolationAt(toZonesPath, fmt.Sprintf("must be empty when type is %s", failover.To.Type))
			}
```

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -49,6 +50,88 @@ func validateTo(to []To) validators.ValidationError {
...
+		case AnyExcept:
+			if len(failover.To.Zones) == 0 {
+				verr.AddViolationAt(toZonesPath, fmt.Sprintf("must not be empty when type is %s", failover.To.Type))
+			}
+		case Only:
+			if len(failover.To.Zones) == 0 {
+				verr.AddViolationAt(toZonesPath, fmt.Sprintf("must not be empty when type is %s", failover.To.Type))
+			}
```
slonka, line 120: same with these 

```suggestion
		case AnyExcept, Only:
			if len(failover.To.Zones) == 0 {
				verr.AddViolationAt(toZonesPath, fmt.Sprintf("must not be empty when type is %s", failover.To.Type))
			}
```

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -49,6 +50,88 @@ func validateTo(to []To) validators.ValidationError {
...
+
+	return verr
+}
+
+func validateFailoverThreshold(failoverThreshold *FailoverThreshold) validators.ValidationError {
+	var verr validators.ValidationError
+	if failoverThreshold == nil {
+		return verr
+	}
+	verr.Add(validators.ValidatePercentageOrNil(validators.RootedAt("percentage"), &failoverThreshold.Percentage))
```
slonka, line 134: this check for percentage `> 100` or `< 0`, does percentage `100%` or `0%` make sense in this case? I guess 100% would mean everything needs to be running, and 0 means disabled?

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -49,6 +50,88 @@ func validateTo(to []To) validators.ValidationError {
...
+		}
+	}
+
+	if weightSpecified > 0 && weightSpecified != len(localZone.AffinityTags) {
+		verr.AddViolation("affinityTags", "each or none affinity tags should have weight")
+	}
+	return verr
+}
+
+func validateCrossZone(crossZone *CrossZone) validators.ValidationError {
```
slonka, line 89: should we be validating that all `from`s are unique? like 

```
          - from:
              zones: ["zone-1"]
            to:
              type: Only
              zones: ["zone-1"]
          - from:
              zones: ["zone-1"]
            to:
              type: Only
              zones: ["zone-2"]
```

this is invalid right?

slonka (changes requested):
A couple of smaller and a bit bigger changes. Please add tests for the cases I mentioned if they make sense.

app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
slonka: These are autogenerated, if they render correctly wherever we copy them to the website then I'm fine with that. Just something to check after you merge I guess. Or maybe someone else from the team knows if they render correctly.

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
Automaat: this is correct, each entry in list will have different priority. Also you can put the same things in `from` and `to`, it will be treated as local in the end

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
Automaat: changed

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
Automaat: fixed

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
Automaat: added validation

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
Automaat: changed

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
Automaat: changed

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
Automaat: changed validation

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
Automaat: added validation

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
slonka: ok so they are only inside a certain priority. that makes sense

slonka (approved):

Issue #6551, opened 2023-04-17
Automatic reachableServices


lahabana:
### Description

[Reachable services](https://kuma.io/docs/dev/production/upgrades-tuning/fine-tuning/#reachable-services) is an ok optimization but it's very manual and annoying.

We rely on TrafficRoutes to populate all the clusters. So if you don't have a TrafficRoute there's nothing you can do (unfortunately this makes it impossible to only use MeshHTTPRoute and MeshTCPRoute).

Could we use the existing policies to only add clusters for the clusters we want to talk to.

## Open questions

1) What do we want to happen when we don't have any MeshHTTPRoute nor MeshTCPRoute?
   To me we should be able to route everywhere.
2) Should we use MeshTrafficPermissions to figure who receives the configuration for which service?
   This makes sense as you would expect to not have any information about a service you don't have access to.
3) What's the migration strategy?

## Implementation idea

Use MeshTrafficPermissions to build the reverse graph of which services can reach which services.
Only add envoy clusters for the services you have permissions to.

This has the following benefits:

1. That's what customers expect this feature to look like
2. You get added security that proxies that have not access to the service don't even it exists
3. Places that go 0 trust get the best perf out of the box
4. You can always allow everything by have a `targetRef: Mesh -- from: targetRef: Mesh -- Allow` policy (which would likely be a good default).
5. No need to figure out with 2 different policies. No need to manage to define a route that does All to All.
6. MeshTrafficPermission would do something even when mTLS is not enabled

Drawbacks:

1. Using MeshTrafficPermission kind of breaks our rule “top-level targetRef selects proxy to configure”. You change  Allow to ShadowDeny for Service A and this affects xds configurations of all other services in the mesh.

Note: If we do this you'd want `ShadowDeny` to be considered the same as `Allow` when it comes to advertising clusters to consumers of a service. Because you want to get the opportunity of a service

lahabana:
> Triage: this can also be used for auditing who can talk to whom.

lukidzi:
I think just relying on 'MeshTrafficPermissions' might not be sufficient. We could do the combination of both `MeshTrafficPermission` and `Mesh*Route`. I think sometimes you might define allow-all but define subset of routes. In this case we could cover more cases. 

Also, 2nd idea is to use stats of communication, it's less reliable but we could do something more like `suggestions`
1. Scrape metrics of clusters over some period of time
2. Build a map of services which service requested and persist in the store
[...]

Issue #6589, opened 2023-04-21
`ExternalService`s shouldn't rely on the existence of a `TrafficPermission`


lahabana:
### Description

When delete default TrafficPermission then ExternalServices don't work at all. Kuma generates outbounds for ES based on the TrafficPermission policy (no permission -> no outbound).

Can we just have a RBAC filter on the external service listener? What happens with Egress?

- [x] https://github.com/kumahq/kuma/issues/8468

lahabana:
Should we close this now? @lukidzi ?

slonka:
This is still a problem, we shouldn't close this ☝️

cbugneac-nex:
I'm on Kuma version 2.4.3 and decided to migrate to `MeshTrafficPermission` from `TrafficPermission`. Created default allow all initially to not break working things:
```yaml
---
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  namespace: kuma-system
  name: allow-all-default
  labels:
    kuma.io/mesh: default
spec:
  from:
  - default:
      action: AllowWithShadowDeny
    targetRef:
      kind: Mesh
  targetRef:
    kind: Mesh
```
And now access to `ExternalServices` doesn't work even if I add explicit `MeshTrafficPermissions` for `ExternalServices`.

lukidzi:
My idea to change the behavior:
* Remove matching service with TrafficPermisson
* Deliver all ExternalServices to Dataplanes when there is no TrafficPermission/MeshTrafficPermission defined 
** If ReachableServices are defined we filter out ExternalServices
** TrafficPermission/MeshTrafficPermission for ExternalServices has only impact when using Egress
I am not sure the best way to handle Auto ReachableServices, they are relying on the MTP so the user would need to define policy to have access or we could deliver all ES until there is a policy matching ES but it seems a bit complicated.

> Can we just have an RBAC filter on the external service listener? What happens with Egress?
I don't think that we should define RBAC in outgoing traffic, just configure RBAC when using egress. Should we allow all when no policy or deny and require the user to define it?

PR #8084, opened 2023-10-20
docs(madr): auto reachable services

slonka:
xref: https://github.com/kumahq/kuma/issues/6551

### Checklist prior to review

### DIFF
--- a/docs/madr/decisions/031-automatic-rechable-services.md
+++ b/docs/madr/decisions/031-automatic-rechable-services.md
@@ -245,6 +245,8 @@ inlineDnsTable:
 
 And that I would have a special listener in Envoy that would have an HTTP filter that would use `odcds`.
 
+You can see a demo of this in https://github.com/slonka/test-odcds.
+
 Pros/cons:
 * This is a very clean solution that would require minimal effort from the users, just marking services with `kuma.io/protocol: http`
 * It requires implementing `odcds` in `go-control-plane` and changes in Kuma
--- a/docs/madr/decisions/031-automatic-rechable-services.md
+++ b/docs/madr/decisions/031-automatic-rechable-services.md
@@ -178,16 +178,34 @@ If the system is too big to do this we need to do extra work in the iterations (
 3. We enable MeshTrafficPermission for services brought into the mesh (the performance improves because we trimmed the configs)
 4. We repeat point 2-3 until all services are in the Mesh
 
-How to ensure in the 3rd point that we defined the right services?
-Remember, once MeshTrafficPermission with deny is added we're going to start rejecting all traffic that is not allowed.
+Let's take the following service graph as an example:
 
-Even if we get the right services on the first try when do we trim the config of the outbounds?
-And when do we actually start blocking traffic?
+```mermaid
+flowchart LR
+ingress --> a --> b & c 
+c --> d & e
+b --> f
+```
 
-If we decide to do this on `MeshTrafficPermission` with `deny` then
-it only works with mTLS enabled and if there are services that aren't in the mesh they would simply be denied access.
-So for us to use `MeshTrafficPermission/deny` we already have to have the whole graph structure figured out,
-turn on Mesh for them 
+If we enable MTP on `d` that allows `c` to contact it,
+and we only use this information
+then suddenly traffic between `c -> e` does not work <sup>1</sup>.
+
+We can't enable MTP on `b` until we bring `a` into the mesh.
+
+How this only could work is that we start at the `ingress` and we bring the services level by level and
+at each level we define MTP. We have to bring services at each level at the same time to not disturb the flow of traffic (<sup>1</sup>).
+
+It seems to me that this approach requires a lot of up front knowledge (the entire service graph) and
+leaves not too much room for error.
+
+There are other small concerns that also need addressing:
+- wrong errors on the client side, instead of 403 client will get 404/"no upstream"
+  - in public facing APIs this is quote common to get a 404 to a resource that you don't have permissions to
+  - we could have a flag to distinguish between a regular 404 and this 404 caused by no upstream resulting from config trimming and return a 403
+- wrong RBAC stats on the server side
+  - we would have correct stats on the client side - if we can translate that and surface it to the user I don't think anyone would mind
+  - 
 
 
 ### {option 2}
--- a/docs/madr/decisions/031-automatic-rechable-services.md
+++ b/docs/madr/decisions/031-automatic-rechable-services.md
@@ -334,6 +334,8 @@ Pros/cons:
 * It requires implementing `odcds` in `go-control-plane` and changes in Kuma
 * No solution for TCP traffic
 * Not possible in the required timeframe
+* Delay on the first request 
+* Unable to make a request to new service when CP is down 
 
 ### Using MeshHTTPRoute / MeshTCPRoute to populate the clusters
 
--- a/docs/madr/decisions/031-automatic-rechable-services.md
+++ b/docs/madr/decisions/031-automatic-rechable-services.md
@@ -164,20 +164,11 @@ This is simply `A` can talk to anyone, this could be either adding and outgoing 
 
 3. `MeshService/MeshService` - already talked about, this is the easy case
 
-4. `MeshServiceSubset` 
+4. Subsets 
 
-Unfortunately there is one option here,
-we need to treat `MeshServiceSubset` just like `MeshService`.
-This is because we can't trim individual endpoints because the data is not available at that point, and
-we have to accept the fact that we will miss out on (small) parts of the config that could be trimmed.
+There is a strong case that makes sense using `k8s.kuma.io/namespace: xyz` to allow only traffic from a specific namespace.
 
-5. `MeshSubset`
-
-This one is a bit more tricky because the potential radious is a bit bigger.
-The approach has to be the same as in `MeshServiceSubset`, meaning: treat `MeshSubset` as just `Mesh`.
-
-There is a case that makes sense using `k8s.kuma.io/namespace: xyz` to allow only traffic from a specific namespace,
-so this needs to be documented that it won't improve the performance.
+We will support `MeshSubset` and `MeshServiceSubset` on the `from` level and support `k8s.kuma.io/*` labels on the `top` level.
 
 #### Migration
 
--- a/docs/madr/decisions/031-automatic-rechable-services.md
+++ b/docs/madr/decisions/031-automatic-rechable-services.md
@@ -46,7 +46,7 @@ Chosen option: "Using MeshTrafficPermission to generate reverse graph that would
 
 ### Negative Consequences
 
-* requires a lot of manual work
+* requires some manual work to define MTPs
 * everything in the option section
   * to have performance consequences it requires knowledge of all services that consume API of the service being migrated (you can always do `from: Mesh`)
   * breaking "top-level targetRef selects proxy to configure" rule
--- a/docs/madr/decisions/031-automatic-rechable-services.md
+++ b/docs/madr/decisions/031-automatic-rechable-services.md
@@ -51,7 +51,7 @@ Chosen option: "Using MeshTrafficPermission to generate reverse graph that would
   * to have performance consequences it requires knowledge of all services that consume API of the service being migrated (you can always do `from: Mesh`)
   * breaking "top-level targetRef selects proxy to configure" rule
   * no RBAC stats on the server side
-  * different errors on the client side, instead of 403 client will get 404/"no upstream"
+  * different errors on the client side, instead of 403 client will get connection close / connection reset (more info in mTLS section)
     * this might be desirable in some cases (e.g. github will 404 on repositories you don't have access to if you **are not** part of the organisation)
     * but it also can be undesirable (e.g. github will 403 on repositories you don't have access to if you **are** part of the organisation)
 
--- a/docs/madr/decisions/031-automatic-rechable-services.md
+++ b/docs/madr/decisions/031-automatic-rechable-services.md
@@ -1,0 +1,214 @@
+# Automatic reachable services
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/6551
+
+## Context and Problem Statement
+
+With large enough Mesh (more than 1000 services) processing the whole cluster state for each proxy is not possible due to performance reasons (what is the boundary that we're hitting, is it the CPU of the CP / DP, can we have flame graphs showing hot spots?).
+This is why a feature called [reachable services](https://github.com/kumahq/kuma-website/blob/045ff9868acdb4b322759dde3f8053f41948100f/app/_src/production/upgrades-tuning/fine-tuning.md#L8-L7) was introduced.
+Reachable services is a step in the right direction but defining the whole service graph manually by the service owners or the platform team is a big ask.
+
+How do we ensure a smooth Mesh migration (keeping in mind the performance and minimising manual labour) for large systems?
+
+## Decision Drivers <!-- optional -->
+
+* feasibility of implementation for 2.5 release
+* minimize the manual labour needed
+* smooth migration to Mesh for large systems without downtime or service communication issues
+
+## Considered Options
+
+* Using MeshTrafficPermission to generate reverse graph that would feed into reachable services
+* {option 2}
+* {option 3}
+* … <!-- numbers of options can vary -->
+
+## Decision Outcome
+
+Chosen option: "{option 1}", because {justification. e.g., only option, which meets k.o. criterion decision driver | which resolves force {force} | … | comes out best (see below)}.
+
+### Positive Consequences <!-- optional -->
+
+* {e.g., improvement of quality attribute satisfaction, follow-up decisions required, …}
+* …
+
+### Negative Consequences <!-- optional -->
+
+* {e.g., compromising quality attribute, follow-up decisions required, …}
+* …
+
+## Pros and Cons of the Options
+
+### Using MeshTrafficPermission to generate reverse graph that would feed into reachable services
+
+#### Mapping
+
+MeshTrafficPermission is defined by the service owner and states which services can communicate with that service.
+Based on that information we could build a reverse graph containing information who can talk to whom and based on that we could trim the configs created (which would be equivalent to populating reachable services field).
+
+So MTPs like this:
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshTrafficPermission
+metadata:
+  namespace: kuma-system
+  name: mtp-b
+spec:
+  targetRef:
+    kind: MeshService
+    name: d
+  from:
+    - targetRef:
+        kind: MeshService
+        name: a
+      default:
+        action: Allow
+    - targetRef:
+        kind: MeshService
+        name: b
+      default:
+        action: Allow
+    - targetRef:
+        kind: MeshService
+        name: c
+      default:
+        action: Allow
+```
+
+Would translate into:
+
+```mermaid
+graph TD
+    A-->D
+    B-->D
+    C-->D
+```
+
+Warning: this is only the case when you enable mTLS and remove default TrafficPermission and MeshTrafficPermission.
+For visibility, you could have a default "deny all" policy.
+
+Let's consider other `targetRef` with allow types in combination with `top`/`from` level:
+1. `Mesh/Mesh` - this won't help us because that's the default, anyone can talk to anyone and we can't
+
+2. `MeshService/Mesh`  
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshTrafficPermission
+metadata:
+  namespace: kuma-system
+  name: mtp-ms-to-mesh
+spec:
+  targetRef:
+    kind: MeshService
+    name: a
+  from:
+    - targetRef:
+        kind: Mesh
+      default:
+        action: Allow
+```
+
+```mermaid
+graph TD
+    B-->A
+    C-->A
+    D-->A
+```
+
+We could either define an edge between every service to that service or have a special list that contains allIncomingAllowed.
+Services from that list would always be added as an outgoing edge.
+
+- `Mesh/MeshService`
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshTrafficPermission
+metadata:
+  namespace: kuma-system
+  name: mtp-mesh-to-ms
+spec:
+  targetRef:
+    kind: Mesh
+  from:
+    - targetRef:
+        kind: MeshService
+        name: a
+      default:
+        action: Allow
+```
+
+```mermaid
+graph TD
+    A-->B
+    A-->C
+    A-->D
+```
+
+This is simply A can talk to anyone, this could be either adding and outgoing edge from this service to every other service or having it on a list called allOutgoing and
+not bothering to trim the configuration.
+
+3. `MeshService/MeshService` - already talked about, this is the easy case
+
+4. `MeshServiceSubset` 
+
+There are two options here:
+1. We could treat it as just `MeshService` and include everything, it will probably be just a small overhead in Endpoints
+2. We could trim the endpoints based on the tags
+
+For 2. we would probably need to reuse the policy matching mechanism while doing the graph computation.
+
+5. `MeshSubset`
+
+This one is a bit more tricky because the potential radious is a bit bigger. The approaches are the same as in `MeshServiceSubset`.
+Do people really use `MeshSubset` for MeshTrafficPermission?
+
+#### Migration
+
+A typical migration to Mesh involves bringing services gradually into the Mesh.
+If the system is small enough, after several iterations, it will be enabled for the whole system (every service).
+
+If the system is too big to do this we need to do extra work in the iterations (defining MeshTrafficPermissions along the way).
+
+1. 0 services are in the Mesh
+2. We bring in X services into the Mesh (where X is small enough, so the Mesh can handle it)
+3. We enable MeshTrafficPermission for services brought into the mesh (the performance improves because we trimmed the configs)
+4. We repeat point 2-3 until all services are in the Mesh
+
+How to ensure in the 3rd point that we defined the right services?
+Remember, once MeshTrafficPermission with deny is added we're going to start rejecting all traffic that is not allowed.
+
+Even if we get the right services on the first try when do we trim the config of the outbounds?
+And when do we actually start blocking traffic?
+
+If we decide to do this on `MeshTrafficPermission` with `deny` then
+it only works with mTLS enabled and if there are services that aren't in the mesh they would simply be denied access.
+So for us to use `MeshTrafficPermission/deny` we already have to have the whole graph structure figured out,
+turn on Mesh for them 
+
+
+### {option 2}
+
+{example | description | pointer to more information | …} <!-- optional -->
+
+* Good, because {argument a}
+* Good, because {argument b}
+* Bad, because {argument c}
+* … <!-- numbers of pros and cons can vary -->
+
+### {option 3}
+
+{example | description | pointer to more information | …} <!-- optional -->
+
+* Good, because {argument a}
+* Good, because {argument b}
+* Bad, because {argument c}
+* … <!-- numbers of pros and cons can vary -->
+
+## Links <!-- optional -->
+
+* {Link type} {Link to ADR} <!-- example: Refined by [ADR-0005](0005-example.md) -->
+* … <!-- numbers of links can vary -->
--- a/docs/madr/decisions/031-automatic-rechable-services.md
+++ b/docs/madr/decisions/031-automatic-rechable-services.md
@@ -36,8 +36,10 @@ Chosen option: "Using MeshTrafficPermission to generate reverse graph that would
 
 ### Positive Consequences
 
-* not invasive, composable, plug and play, can be removed if proven not efficient enough
-  * composable means you can have both trimming by MTP and add reachable services annotation, they are combined
+* not invasive, plug and play, can be removed if proven not efficient enough
+  * reachable services should take precedence because:
+    * we can change it back if we need to
+    * we have escape path if there is an edge case that would incorrectly generate the graph
 * possible to implement in the given time frame
 * it still works with reachable services and people can mix and match both options
 * is a performance benefit for users that already have MTPs defined
--- a/docs/madr/decisions/031-automatic-rechable-services.md
+++ b/docs/madr/decisions/031-automatic-rechable-services.md
@@ -209,10 +209,28 @@ spec:
         name: a
       default:
         action: Allow
-
 ```
 
-The migration assumes using Kubernetes DNS.
+And then for each of namespaces and then selectively allow services with:
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshTrafficPermission
+metadata:
+  namespace: kuma-system
+  name: ns-allow
+spec:
+  targetRef:
+    kind: MeshService
+    name: my-public
+  from:
+    - targetRef:
+        kind: Mesh
+      action: Allow
+```
+
+Scale this to 10 namespaces with 10 services per namespace with 2 services per ns accessible outside their own service.
+Then you have per instance 10 + (9 * 2) = 28 services to know about vs 100 services in our current naive case this is a massive improvement and without much sacrifice.
 
 Let's take the following service graph as an example:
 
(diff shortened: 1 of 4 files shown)


docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
+# Automatic reachable services
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/6551
+
+## Context and Problem Statement
+
+With large enough Mesh (more than 1000 services) processing the whole cluster state for each proxy is not possible due to performance reasons (what is the boundary that we're hitting, is it the CPU of the CP / DP, can we have flame graphs showing hot spots?).
```
lahabana, line 9: As mentioned before the complexity of this is O(n^2) (n is the number of services). We end up with envoy configs that are multiple MB and CP consuming many cores for small clusters. You can check this by just looking at small clusters with and without reachable services all metrics collapse when using reachable services

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+
+* not invasive, plug and play, can be removed if proven not efficient enough
+* possible to implement in the given time frame
+
+### Negative Consequences
+
+* requires a lot of manual work, in ideal situation I think we could try to do a PoC of ODCDS
+* everything in the option section
+  * requires knowledge of all services that consume API of the service being migrated
+  * breaking "top-level targetRef selects proxy to configure" rule
```
lahabana, line 45: I think that's the only negative consequence though wouldn't this also mean that MTP are in any case partly handled outside of the plugin?

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+
+A typical migration to Mesh involves bringing services gradually into the Mesh.
+If the system is small enough, after several iterations, it will be enabled for the whole system (every service).
+
+If the system is too big to do this we need to do extra work in the iterations (defining MeshTrafficPermissions along the way).
+
+1. 0 services are in the Mesh
+2. Bring X services into the Mesh (where X is small enough, so the Mesh can handle it)
+3. Enable MeshTrafficPermission for services brought into the mesh (the performance improves because we trimmed the configs)
+4. Repeat point 2-3 until all services are in the Mesh
```
lahabana, line 184: I don't think that's what happens. Having an MTP for the mesh should be a recommended way to go about a migration. Even if you use a very wide MTP (from: Mesh or from: Namespace).
If you do it organically like this you get transparent benefits.

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+
+The good thing about how this migration is set up is that we can gradually do it without any interruption to the system.
+The downside of it is that **a service owner needs to know EXACTLY which services consume their service** to avoid any disruptions.
```
lahabana, line 187: No it doesn't. It's an optimisation the problem we're trying to address is that we have an algorithm that scales in O(n^2) and must reduce the complexity.
For any service from which we define a restrictive MTP we reduce the complexity. If it's perfect then you go to the optimal solution. If it's not perfect you are still better than worst case.

Think about it just telling people to start with:

```yaml
type: MTP
name: namespace-controlled
spec:
  targetRef:
    kind: MeshSubset
    tags:
       namespace: my-ns
  from:
    - kind: MeshSubset
      tags:
        namespace: my-ns
      action: Allow
```

Just do this for each of your namespace and then selectively allow services with:

```yaml
type: MTP
name: ns-allow
spec:
   targetRef:
      kind: MeshService
      name: my-public
   from:
      - targetRef:
            kind: Mesh
         action: Allow
```

Scale this to 10 namespaces with 10 services per namespace with 2 services per ns accessible outside their own service.

Then you have per instance 10 + (9 * 2) = 28 services to know about vs 100 services in our current naive case this is a massive improvement and without much sacrifice.

Even if you take a super open mesh with 1k services who only have 3 services with very restrictive MTP the number goes down quickly 1000^2 = 1M vs  997^2 + (the few services that do actually access the 3 restricted services)

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+The migration assumes using Kubernetes DNS.
```
lahabana, line 188: I don't think that's specific to this MADR right?

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+
+```mermaid
+flowchart LR
+ingress --> a --> b & c 
+c --> d & e
+b --> f
+```
+
+We want to bring services into the mesh one by one and trim their config, let's start with service `c` (it's a good example because it has both inbound and outbound traffic):
+0. We have a default MTP `Mesh/Mesh` with `allow`
```
lahabana, line 200: I think that's the broken premise here. The MTP allow all should only be used in demos. Service owners should think about who can access their services (if you do the code parallel it's similar to public/private).

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+    name: some-service.svc
+```
+
+And that I would have a special listener in Envoy that would have an HTTP filter that would use `odcds`.
+
+Pros/cons:
+* This is a very clean solution that would require minimal effort from the users, just marking services with `kuma.io/protocol: http`
+* It requires implementing `odcds` in `go-control-plane` and changes in Kuma
+* No solution for TCP traffic
+* Not possible in the required timeframe
```
lahabana, line 252: I think a feature like this is by far our best hope. The advantage of MTP is that we would be able to fail ODCDS if we request a cluster we don't have access too so if we implement this later we'd still need the MTP feature.

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Decision Outcome
+
+Chosen option: "Using MeshTrafficPermission to generate reverse graph that would feed into reachable services"
+
+### Positive Consequences
+
+* not invasive, plug and play, can be removed if proven not efficient enough
+* possible to implement in the given time frame
+
```
lahabana, line 39: Another positive consequence is that it still works with reachable services and folks can mix and match both options.

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+This is because we can't trim individual endpoints because the data is not available at that point, and
+we have to accept the fact that we will miss out on (small) parts of the config that could be trimmed.
+
+5. `MeshSubset`
+
+This one is a bit more tricky because the potential radious is a bit bigger.
+The approach has to be the same as in `MeshServiceSubset`, meaning: treat `MeshSubset` as just `Mesh`.
+
+There is a case that makes sense using `k8s.kuma.io/namespace: xyz` to allow only traffic from a specific namespace,
+so this needs to be documented that it won't improve the performance.
```
lahabana, line 172: I feel that we take this a little upside down. What we do is when we generate the configuration of a DPP we want to figure which clusters we have access to. At this time we should be able to look at all the MTP one by one and the list of services.
So we can't manage a `MeshSubset` at the top level but we can at the `from` level right?

So in the end I think the algorithm is:

- take all MTPs
- filter out all MTP with topLevel targetRef of MeshSubset
- for each service in the mesh keep only those who have an MTP which have access to ANY endpoint in this service from the current DPP

no?

slonka:
> I'm still of the (opinion? - insertion by slonka) that this MTP option is still the pragmatically best option for the following reasons:
> 
> * it's small and easy to implement
> * it has real added security benefits

I completely agree.
 
> I think there's too much focus on the migration here

I don't fully agree. To me going through migration steps and making sure they are airtight was really important. I would highly oppose this option if I found gaps in the migration path. I did not so I'm confident that we can go forward with this.

The second thing here is that you pointed out that there is too much focus on migration but you did not say what I should focus on more.

> I guess switching this around would we resist this feature as much if we wanted added security instead of better performance? We'd have added security with the side effect of better perf.

I'm not resisting this feature. I tried to lay down all the possible options and figure out which one is the best taking into account all the real world constraints. I did not want to sound like I'm against this option. My gut feeling is it's the best option given the circumstances but in the long term we should evaluate other options. 

---

I'm going through other comments and adjusting the contents now.

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+* possible to implement in the given time frame
+
+### Negative Consequences
+
+* requires a lot of manual work, in ideal situation I think we could try to do a PoC of ODCDS
+* everything in the option section
+  * requires knowledge of all services that consume API of the service being migrated
+  * breaking "top-level targetRef selects proxy to configure" rule
+  * wrong RBAC stats on the server side
+  * wrong errors on the client side, instead of 403 client will get 404/"no upstream"
```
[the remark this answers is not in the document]
slonka: I think there are different levels so you'd get 404 if you're outside the org and you'd get 403 if you're inside the org but you don't have access to the repo. I'll change the point to reflect that.

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+### Positive Consequences
+
+* not invasive, plug and play, can be removed if proven not efficient enough
+* possible to implement in the given time frame
+
+### Negative Consequences
+
+* requires a lot of manual work, in ideal situation I think we could try to do a PoC of ODCDS
+* everything in the option section
+  * requires knowledge of all services that consume API of the service being migrated
```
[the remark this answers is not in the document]
slonka: yeah but then it does not help us, right? I'll adjust the point

docs/madr/decisions/031-automatic-rechable-services.md
slonka: yeah, to build the graph and to use it we need to hook in `mesh_context_builder.go`

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+
+There is a case that makes sense using `k8s.kuma.io/namespace: xyz` to allow only traffic from a specific namespace,
+so this needs to be documented that it won't improve the performance.
+
+#### Migration
+
+A typical migration to Mesh involves bringing services gradually into the Mesh.
+If the system is small enough, after several iterations, it will be enabled for the whole system (every service).
+
+If the system is too big to do this we need to do extra work in the iterations (defining MeshTrafficPermissions along the way).
```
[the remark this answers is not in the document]
slonka: By "from the start" you mean that people are defining MTP before even enabling the Mesh? I don't think that's the case. I think that most of the people would first want to try install it and then try out features. I will add a point of side benefit to the existing MTP users.

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+Chosen option: "Using MeshTrafficPermission to generate reverse graph that would feed into reachable services"
+
+### Positive Consequences
+
+* not invasive, plug and play, can be removed if proven not efficient enough
+* possible to implement in the given time frame
+
+### Negative Consequences
+
+* requires a lot of manual work, in ideal situation I think we could try to do a PoC of ODCDS
```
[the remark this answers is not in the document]
jakubdyszkiewicz: What manual work does it require?

docs/madr/decisions/031-automatic-rechable-services.md
jakubdyszkiewicz: > wouldn't this also mean that MTP are in any case partly handled outside of the plugin?

unfortunately - yes :(

docs/madr/decisions/031-automatic-rechable-services.md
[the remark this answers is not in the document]
jakubdyszkiewicz: I don't think you will get 404. And I don't think you would get 403 with clusters.

docs/madr/decisions/031-automatic-rechable-services.md
slonka: I agree but if I remember correctly my chat with @jakubdyszkiewicz we had a problem with `subsets` - https://github.com/kumahq/kuma/pull/8084/files#diff-28d6e9df351b905115d7d3e1c54ba2075abd0f57c370d093917530af05935e75R159 . Maybe @jakubdyszkiewicz figured out a way around this.

docs/madr/decisions/031-automatic-rechable-services.md
slonka: No, just wanted to state that out explicitly, can remove this if you want.

docs/madr/decisions/031-automatic-rechable-services.md
slonka: > The advantage of MTP is that we would be able to fail ODCDS if we request a cluster we don't have access too so if we implement this later we'd still need the MTP feature.

I don't think I understand this fully, so you'd want to `nack` the ODCDS request if someone requests a cluster they don't have access to?

docs/madr/decisions/031-automatic-rechable-services.md
slonka: > So we can't manage a MeshSubset at the top level but we can at the from level right?

I think you're right, but I guess @jakubdyszkiewicz would be the best to confirm this

docs/madr/decisions/031-automatic-rechable-services.md
michaelbeaumont: while it kind of breaks this rule, at least MTPs _still also (primarily?) affects_ the top level `targetRef` proxies. I mean no matter the solution, the consequence of _not_ breaking this rule would be to require config duplication essentially, right?
so personally I'm very OK with this.

docs/madr/decisions/031-automatic-rechable-services.md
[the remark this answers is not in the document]
slonka: defining MTPs - "a lot" could be debatable if we can define coarse grained (namespace) level MTPs

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,263 @@
...
+* Using MeshHTTPRoute / MeshTCPRoute to populate the clusters
+
+## Decision Outcome
+
+Chosen option: "Using MeshTrafficPermission to generate reverse graph that would feed into reachable services"
+
+### Positive Consequences
+
+* not invasive, plug and play, can be removed if proven not efficient enough
+* possible to implement in the given time frame
```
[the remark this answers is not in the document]
slonka: added

docs/madr/decisions/031-automatic-rechable-services.md
slonka: added

docs/madr/decisions/031-automatic-rechable-services.md
slonka: > I mean no matter the solution, the consequence of not breaking this rule would be to require config duplication essentially, right?

with on-demand there wouldn't be any policy to configure it

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,289 @@
...
+
+This is simply `A` can talk to anyone, this could be either adding and outgoing edge from this service to every other service or having it on a list called allOutgoing and not bothering to trim the configuration.
+
+3. `MeshService/MeshService` - already talked about, this is the easy case
+
+4. Subsets 
+
+There is a strong case that makes sense using `k8s.kuma.io/namespace: xyz` to allow only traffic from a specific namespace.
+
+We will support `MeshSubset` and `MeshServiceSubset` on the `from` level and support `k8s.kuma.io/*` labels on the `top` level.
```
slonka, line 171: @jakubdyszkiewicz maybe you can fill in the details on this if needed.

docs/madr/decisions/031-automatic-rechable-services.md
slonka: It's sort of similar to my previous point. I think that we also had this assumption when talking with @jakubdyszkiewicz

docs/madr/decisions/031-automatic-rechable-services.md
lahabana: Yeah I'd guess so no? Or have a cluster with a direct response? Anyway out of scope here :)

slonka: Yeah, cluster with direct response would work I think.

docs/madr/decisions/031-automatic-rechable-services.md
jakubdyszkiewicz: You can look at the implem that MTP is still handled inside of the MTP plugin.
MeshContextBuilder now has an interface with Graph and GraphBuilder and anytoany default implementation. MTP plugin provides a Graph & Builder based on MTP and this is wired together in bootstrap.

https://github.com/kumahq/kuma/pull/8125

docs/madr/decisions/031-automatic-rechable-services.md
jakubdyszkiewicz: We support any subset in from and subsets with only predefined tags in top level target ref. Those predefined tags are
`k8s.kuma.io/namespace`, `k8s.kuma.io/service` and `k8s.kuma.io/port`

docs/madr/decisions/031-automatic-rechable-services.md
jakubdyszkiewicz: see https://github.com/kumahq/kuma/pull/8084#discussion_r1371331838

docs/madr/decisions/031-automatic-rechable-services.md
michaelbeaumont: > with on-demand there wouldn't be any policy to configure it

Yeah I just mean if were to use a different policy instead of MTP there would potentially be a lot of duplication basically setting up this graph in both MTPs and the other policy

michaelbeaumont (approved):

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+The downside of it is that  to avoid any disruptions.
+
+The migration assumes using Kubernetes DNS.
+
+Let's take the following service graph as an example:
+
+```mermaid
+flowchart LR
```
lobkovilya, line 217: This graph doesn't show the situation when the service has at least 2 different clients. I think it will be useful to cover migration for such case as well. Also, worth mentioning mTLS mode, service migration should work both for Disabled and Permissive mTLS.

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,289 @@
...
+* Dynamically load clusters using [ODCDS](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/on_demand/v3/on_demand.proto)
+* Using MeshHTTPRoute / MeshTCPRoute to populate the clusters
+
+## Decision Outcome
+
+Chosen option: "Using MeshTrafficPermission to generate reverse graph that would feed into reachable services"
+
+### Positive Consequences
+
+* not invasive, plug and play, can be removed if proven not efficient enough
```
lobkovilya, line 39: How does it work if you have both old reachable services and new MTP reachable services?

docs/madr/decisions/031-automatic-rechable-services.md
lukidzi: I would add cons: delay on the first request. It seems like for some clients it might be a blocker. 
Also, another thing, what if your CP is down and you want to make a request to a new service? In this case, MTP has the knowledge about services already while ODCDS won't get it?

docs/madr/decisions/031-automatic-rechable-services.md
slonka: Will add this and a section on mTLS

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,289 @@
...
+  from:
+    - targetRef:
+        kind: MeshService
+        name: a
+      default:
+        action: Allow
+
+```
+
+The downside of it is that  to avoid any disruptions.
```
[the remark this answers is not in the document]
slonka: needed to remove that but forgot, thanks!

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,289 @@
...
+
+```mermaid
+flowchart LR
+ingress --> a --> b & c 
+c --> d & e
+b --> f
+```
+
+We want to bring services into the mesh one by one and trim their config, let's start with service `c` (it's a good example because it has both inbound and outbound traffic):
+0. We have a default MTP `Mesh/Mesh` with `allow`
```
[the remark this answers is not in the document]
slonka: done

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,289 @@
...
+Chosen option: "Using MeshTrafficPermission to generate reverse graph that would feed into reachable services"
+
+### Positive Consequences
+
+* not invasive, plug and play, can be removed if proven not efficient enough
+* possible to implement in the given time frame
+* it still works with reachable services and people can mix and match both options
+* is a performance benefit for users that already have MTPs defined
+* adds an extra layer of security as you can't look at your envoy config to figure out the IP of services you don't have access to
+* a compromised internal service can't be exploited to DOS services it does not have access to
```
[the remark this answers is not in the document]
slonka: you're right

docs/madr/decisions/031-automatic-rechable-services.md
slonka: they are combined, as mentioned in some other section you can have coarse grain MTP like from namespace to namespace and selected services in that namespace can have specific `reachable-services` will add this to the list

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,309 @@
...
+  - we would have correct stats on the client side - if we can translate that and surface it to the user I don't think anyone would mind, but
+    I think its is for future improvement
+- assuming that every service owner knows which services consume their API
+  - as a future improvement we will come up with tooling to assist the users with this problem
+    - have metrics / logs and a dashboard when `allow_with_shadow_deny` is used (this is not a silver bullet - it assumes that during "discovery" period all possible communication will happen, if you gather data for a week and there is a job that runs every quarter it probably won't be recorded)
+
+##### mTLS
+
+When mTLS is disabled, the traffic will go through the `passthrough` cluster and reach the destination.
+When either permissive or strict mTLS is enabled the requests will fail with a closed connection like this:
```
lukidzi, line 257: Could you also add a response when `passthrough` is disabled and mTLS enabled?

docs/madr/decisions/031-automatic-rechable-services.md
```diff
@@ -0,0 +1,309 @@
...
+* Dynamically load clusters using [ODCDS](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/on_demand/v3/on_demand.proto)
+* Using MeshHTTPRoute / MeshTCPRoute to populate the clusters
+
+## Decision Outcome
+
+Chosen option: "Using MeshTrafficPermission to generate reverse graph that would feed into reachable services"
+
+### Positive Consequences
+
+* not invasive, composable, plug and play, can be removed if proven not efficient enough
```
lukidzi, line 39: I have a question related to the ExternalServices and their interaction with the 'reachableServices' configuration. Currently, it appears that specifying ExternalServices in the 'reachableServices' doesn't have an effect. However, I'm curious if doing so could offer advantages in terms of security, configuration management, or performance optimization. Also, if we would like to do it how would that interact with autorechable?

docs/madr/decisions/031-automatic-rechable-services.md
jakubdyszkiewicz: > The MTP allow all should only be used in demos

_Meanwhile in the real world .. ;)_

docs/madr/decisions/031-automatic-rechable-services.md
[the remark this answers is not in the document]
jakubdyszkiewicz: It's exercised on the L4

docs/madr/decisions/031-automatic-rechable-services.md
jakubdyszkiewicz: We don't know if "other policy" would work in the similar way, so we don't know if there would be duplication. You can look at the proposed implem here https://github.com/kumahq/kuma/pull/8125

docs/madr/decisions/031-automatic-rechable-services.md
jakubdyszkiewicz: Not sure if there is much to add here. Maybe that we could support all common labels (not only k8s.kuma.io/) between all instances of the Service, but at the higher computation cost, so we want to avoid this for now.

docs/madr/decisions/031-automatic-rechable-services.md
jakubdyszkiewicz: They are not combined. Static reachable services takes precedence

slonka: So if owner of service `b` defines reachable service to `c` and `c` says only `a` and `job` can reach them then service `b` will get the cluster (know the IP) but will bounce back on RBAC. @lahabana what do you think about that?

docs/madr/decisions/031-automatic-rechable-services.md
lukidzi: Ok, maybe I haven't noticed this. When reachableServices don't specify ExternalService's name it isn't available. If that works it's fine to me. I've checked the code and thought that it might not affect them

docs/madr/decisions/031-automatic-rechable-services.md
slonka: will add

docs/madr/decisions/031-automatic-rechable-services.md
slonka: Right now auto-reachable-services are combined with rechableservices, just like mentioned in - https://github.com/kumahq/kuma/pull/8084#discussion_r1371523549 - but @jakubdyszkiewicz wanted to change that.

lukidzi: Also, does it relate to your change: https://github.com/kumahq/kuma/issues/6589

docs/madr/decisions/031-automatic-rechable-services.md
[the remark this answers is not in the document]
slonka: Adjusted the point

docs/madr/decisions/031-automatic-rechable-services.md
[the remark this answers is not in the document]
slonka: adjusted the errors and expanded on them in mTLS section

docs/madr/decisions/031-automatic-rechable-services.md
slonka: I agree, I just made the example small so it's easier to understand and illustrate. I added a paragraph above.

docs/madr/decisions/031-automatic-rechable-services.md
slonka: Added this scaling example and specific labels in `Subsets` section.

docs/madr/decisions/031-automatic-rechable-services.md
slonka: > ODCDS won't get it?

yeah you won't get a response so it won't work

added the cons mentioned

docs/madr/decisions/031-automatic-rechable-services.md
slonka: I updated the whole section

docs/madr/decisions/031-automatic-rechable-services.md
slonka: done

docs/madr/decisions/031-automatic-rechable-services.md
slonka: I think it does, if you don't define traffic permission for ES it won't get the cluster. @jakubdyszkiewicz do we want to treat ES differently?

docs/madr/decisions/031-automatic-rechable-services.md
slonka: removed it as a premise since I checked via E2E test is not needed

docs/madr/decisions/031-automatic-rechable-services.md
jakubdyszkiewicz: No, we treat it as any other service

docs/madr/decisions/031-automatic-rechable-services.md
lahabana: I'm a little concerned with this if we extend the reachable service expressiveness (For .e.g `reachableService: {prefix: foo}`). As MTP would be a way to further reduce the number of services.
Also @slonka 's point makes a little sense from a security POV.

What I first though what we would do is something like:

```go
var allServices []services
if (hasReachableServices) {
   allServices = reachableServices
} else {
   allServices = getAllServices()
}
allServices = filterByPermissionGraph(allServices)
``` 

What's the motivation for replacing instead of combining?

jakubdyszkiewicz: For me combined = services from static rs and services from graph.

We have this implem above in the current PR. We could do this, but I feel it's better to have an "escape path" if anything goes wrong.

lukidzi (approved):

docs/madr/decisions/031-automatic-rechable-services.md
slonka: haven't heard strong voices against this, I'm closing if anyone else feels strongly against this, please reopen

slonka: since Lukasz approved I'm assuming this answer was enough. Resolving, please reopen if needed.

docs/madr/decisions/031-automatic-rechable-services.md
slonka: Resolving conversation since Lukasz approved this PR (I guess this answer was enough). Please reopen otherwise.

docs/madr/decisions/031-automatic-rechable-services.md
slonka: We talked on zoom and agreed that reachable services should take precedence because:
- we can change it back if we need to
- we have escape path if there is an edge case that would incorrectly generate the graph

lobkovilya (approved):

pkg/plugins/runtime/k8s/controllers/outbound_converter.go
```diff
@@ -22,6 +22,7 @@ func (p *PodConverter) OutboundInterfacesFor(
 ) ([]*mesh_proto.Dataplane_Networking_Outbound, error) {
 	var outbounds []*mesh_proto.Dataplane_Networking_Outbound
 
+	// wire in the graph here
```
slonka, line 25:
```suggestion
```

pkg/xds/context/mesh_context_builder.go
```diff
@@ -125,6 +125,9 @@ func (m *meshContextBuilder) BuildIfChanged(ctx context.Context, meshName string
 			zoneEgresses,
 		)
 	}
+	// we should probably compute the inverted graph here
+	// using resources.MeshLocalResources[v1alpha1.MeshTrafficPermissionType]
+	// and endpoint map / dataplanes
```
slonka, line 130:
```suggestion
```

pkg/xds/sync/dataplane_proxy_builder.go
```diff
@@ -117,6 +117,7 @@ func (p *DataplaneProxyBuilder) resolveVIPOutbounds(meshContext xds_context.Mesh
 	var outbounds []*mesh_proto.Dataplane_Networking_Outbound
 	for _, outbound := range meshContext.VIPOutbounds {
 		service := outbound.GetService()
+		// here we cut down the config both for universal and for k8s
```
slonka, line 120:
```suggestion
```

pkg/xds/sync/dataplane_proxy_builder.go
```diff
@@ -159,6 +160,7 @@ func (p *DataplaneProxyBuilder) matchPolicies(meshContext xds_context.MeshContex
 		ProxyTemplate:      template.SelectProxyTemplate(dataplane, resources.ProxyTemplates().Items),
 		Dynamic:            core_xds.PluginOriginatedPolicies{},
 	}
+	// match plugin policies
```
slonka, line 163:
```suggestion
```

jakubdyszkiewicz (approved):

PR #8125, opened 2023-10-24
feat(xds): auto reachable services based on MeshTrafficPermission

jakubdyszkiewicz:
### Checklist prior to review

This PR introduces auto reachable services based on MeshTrafficPermission described here https://github.com/kumahq/kuma/pull/8084

I introduced a concept of reachable services graph which is just an interface in "core" xds_context package and default implementation that any service can reach any other service. Then in MTP plugin I created implementation based on MTP which is wired up in bootstrap. This way we avoid cyclic dependency and dependency on plugins, MTP, matching in xds_context.

At first, I created graph of service to service, but then we figured out that subset (MeshSubset / MeshServiceSubset) support is quite critical therefore the graph keeps list of rules for every service which can be evaluated by every DPP. This means that we support subsets fully in `from` section. In top level target ref it's impossible to support all subsets because client won't know which instance of the server it will reach. therefore we decided to support only predefined tags that we know are autogenerated and stable for all instances (which is `k8s.kuma.io/*`). We could have supported any common tag between instances but it would then require to call `replaceSubsets` function for every single service.

This is opt-in experimental feature.

It is not fully clear what is the impact on perf here. We will investigate this.

Existing reachable services takes precedence over the graph.

Todo as a follow-up
* A way to inspect generated graph, possibly with inspect endpoints.


test/e2e/reachableservices/auto_reachable_services_k8s.go
```diff
@@ -0,0 +1,94 @@
...
+
+		Consistently(func(g Gomega) {
+			failures, err := client.CollectFailure(
+				k8sCluster,
+				"second-test-server",
+				"first-test-server:80",
+				client.FromKubernetesPod(TestNamespace, "second-test-server"),
+			)
+			g.Expect(err).To(Not(HaveOccurred()))
+			g.Expect(failures.Exitcode).To(Equal(52))
```
slonka, line 91: I'm a bit worried that if MTLS is permissive (defined [here](https://github.com/kumahq/kuma/pull/8125/files#diff-2a35641ee0827a92c8d410d7f048c505dff6a5c9d9d934d4baa49a150a26944cR25)) this assertion will also pass. Did it work like this before?

reflected here: https://github.com/kumahq/kuma/pull/8084/files#diff-28d6e9df351b905115d7d3e1c54ba2075abd0f57c370d093917530af05935e75R257

jakubdyszkiewicz: I did an investigation. It's broken https://github.com/kumahq/kuma/issues/8152

pkg/plugins/policies/meshtrafficpermission/graph/reachable_services_graph.go
```diff
@@ -0,0 +1,161 @@
...
+	for _, dp := range dataplanes {
+		set := dp.Spec.TagSet()
+		for _, svc := range set.Values(mesh_proto.ServiceTag) {
+			if _, ok := services[svc]; ok {
+				continue
+			}
+			services[svc] = map[string]string{}
+			for tag := range SupportedTags {
+				if values := set.Values(tag); len(values) > 0 {
+					services[svc][tag] = values[0]
```
lobkovilya, line 65: Since the `Values()` method sorts the result `values[0]` can be a value from any inbound, right? This means in `services[svc]` you can assemble a not existing inbound, given:
```yaml
inbounds:
  - port: 9090
    tags:
      kuma.io/service: svc-1
      k8s.kuma.io/namespace: aaa
      k8s.kuma.io/service-port: "9090"
  - port: 8080
    tags:
      kuma.io/service: svc-2
      k8s.kuma.io/namespace: bbb
      k8s.kuma.io/service-port: "8080"
```
in `services` you can get:
```yaml
services:
  svc-1:
    k8s.kuma.io/namespace: bbb
    k8s.kuma.io/service-port: "9090"
```
and it's possible there are no policies that're matched for `svc1:bbb:9090`

pkg/plugins/policies/meshtrafficpermission/graph/reachable_services_graph.go
```diff
@@ -0,0 +1,161 @@
...
+		graph.rules[service] = rl
+	}
+
+	return graph
+}
+
+// replaceTopLevelSubsets replaces subsets present in top-level target ref.
+// Because we need to do policy matching on services instead of individual proxies, we have to handle subsets in a special way.
+// What we do is we only support subsets with predefined tags listed in SupportedTags.
+// This assumes that tags listed in SupportedTags are stable between all instances of a given service.
```
lobkovilya, line 127: What does it mean "stable between all instances of a given service"? You can have 1 service spread across many namespaces right?

pkg/plugins/policies/meshtrafficpermission/graph/reachable_services_graph.go
```diff
@@ -0,0 +1,161 @@
...
+// Otherwise, we fallback to higher level target ref (MeshSubset -> Mesh, MeshServiceSubset -> MeshService).
+//
+// Alternatively, we could have computed all common tags between instances of a given service and then allow subsets with those common tags.
+// However, this would require calling this function for every service.
+func replaceTopLevelSubsets(mtps []*mtp_api.MeshTrafficPermissionResource) []*mtp_api.MeshTrafficPermissionResource {
```
lobkovilya, line 132: Wait, but this transformation of policies can change the final configuration, i.e given:

```yaml
name: aaa
targetRef:
  kind: MeshSubset
  tags:
    k8s.kuma.io/namespace: ns1
from: 
  - targetRef:
      kind: MeshService
      name: backend
    default:
      action: Allow
---
name: bbb
targetRef:
  kind: MeshSubset
  tags:
    non-supported-tag: value
from: 
  - targetRef:
      kind: MeshService
      name: backend
    default:
      action: Deny
```
since kinds are equal and `bbb > aaa` then the final result is `Deny` for `backend`. After `replaceTopLevelSubsets` it will look like this:

```yaml
name: aaa
targetRef:
  kind: MeshSubset
  tags:
    k8s.kuma.io/namespace: ns1
from: 
  - targetRef:
      kind: MeshService
      name: backend
    default:
      action: Allow
---
name: bbb
targetRef:
  kind: Mesh
from: 
  - targetRef:
      kind: MeshService
      name: backend
    default:
      action: Deny
```
now MeshSubset is more specific than Mesh so the final result is `Allow` for `backend`.

pkg/plugins/policies/meshtrafficpermission/graph/reachable_services_graph.go
jakubdyszkiewicz: stable = the same. All instances of a service has the same value of a tag

pkg/plugins/policies/meshtrafficpermission/graph/reachable_services_graph.go
jakubdyszkiewicz: should we delete all froms with Deny whenever we replace subset?

It's what I had before when I replaced subsets in "form" I ignored the one with deny

lobkovilya: I don't think it will work, just imagine the example I gave but swap `Allow` and `Deny`:
```yaml
name: aaa
targetRef:
  kind: MeshSubset
  tags:
    k8s.kuma.io/namespace: ns1
from: 
  - targetRef:
      kind: MeshService
      name: backend
    default:
      action: Deny
---
name: bbb
targetRef:
  kind: MeshSubset
  tags:
    non-supported-tag: value
from: 
  - targetRef:
      kind: MeshService
      name: backend
    default:
      action: Allow
```
after replacing it will be:
```yaml
name: aaa
targetRef:
  kind: MeshSubset
  tags:
    k8s.kuma.io/namespace: ns1
from: 
  - targetRef:
      kind: MeshService
      name: backend
    default:
      action: Deny
---
name: bbb
targetRef:
  kind: Mesh
from: 
  - targetRef:
      kind: MeshService
      name: backend
    default:
      action: Allow
```

jakubdyszkiewicz: Why it wouldn't work in this case? Yes, the graph would create clusters for more clients than needed in a perfect scenario, but it wouldn't block the traffic, right?

lobkovilya: before you had `backend: Allow`, after replacement, you have `backend: Deny`, so `backend` won't receive the outbound

jakubdyszkiewicz: talked offline that the solution is to keep existing target ref but get rid of unsupported tags

pkg/plugins/policies/meshtrafficpermission/graph/reachable_services_graph_test.go
```diff
@@ -0,0 +1,363 @@
...
+					_, fromAll := given.expectedFromAll[to]
+					_, conn := given.expectedConnections[from][to]
+					Expect(g.CanReach(
+						map[string]string{mesh_proto.ServiceTag: from},
+						map[string]string{mesh_proto.ServiceTag: to},
+					)).To(Equal(fromAll || conn))
+				}
+			}
+		},
+		Entry("allow all", testCase{
```
lobkovilya, line 50: we can add an entry when the order of MTPs is decided by their names

pkg/plugins/policies/meshtrafficpermission/graph/reachable_services_graph.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+			if len(filteredTags) != len(mtp.Spec.TargetRef.Tags) {
+				mtp = &mtp_api.MeshTrafficPermissionResource{
+					Meta: mtp.Meta,
+					Spec: mtp.Spec.DeepCopy(),
+				}
+				mtp.Spec.TargetRef.Tags = filteredTags
+			}
+		}
+		newMtps[i] = mtp
```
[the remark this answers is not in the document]
jakubdyszkiewicz: I want to avoid copying if we don't have to. I can't modify this in place, because it's an object from the cache

PR #8129, opened 2023-10-25
fix(kuma-cp): fix Zone{In|E}gress sync when no mesh

bartsmykla:
- Add ZoneEgresses to AggregatetMeshContexts when there is no mesh.
- Sync ZoneIngresses and ZoneEgresses only when the hash changes.
- Return an error with the correct resource type (ZoneEgress) when a ZoneEgress cannot be found.

### Checklist prior to review

  - No relevant issue
  - It won't
  - Manually tested on k3d
  - There is no need
  - It does

### DIFF
--- a/pkg/xds/context/aggregate_mesh_context.go
+++ b/pkg/xds/context/aggregate_mesh_context.go
@@ -32,15 +32,28 @@ func AggregateMeshContexts(
 		meshContextsByName[mesh.Meta.GetName()] = meshCtx
 	}
 
+	hash := aggregatedHash(meshContexts)
+
 	egressByName := map[string]*core_mesh.ZoneEgressResource{}
 	if len(meshContexts) > 0 {
 		for _, egress := range meshContexts[0].Resources.ZoneEgresses().Items {
 			egressByName[egress.Meta.GetName()] = egress
 		}
+	} else {
+		var egressList core_mesh.ZoneEgressResourceList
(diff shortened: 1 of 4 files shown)


pkg/xds/context/aggregate_mesh_context.go
```diff
@@ -32,15 +32,28 @@ func AggregateMeshContexts(
...
 
+	hash := aggregatedHash(meshContexts)
+
 	egressByName := map[string]*core_mesh.ZoneEgressResource{}
 	if len(meshContexts) > 0 {
 		for _, egress := range meshContexts[0].Resources.ZoneEgresses().Items {
 			egressByName[egress.Meta.GetName()] = egress
 		}
+	} else {
+		var egressList core_mesh.ZoneEgressResourceList
```
slonka, line 43: Why are we only doing this for egress? I thought the same problems were with ingress.

slonka (approved):
Is there a test somewhere that you could extend to check for this case?

pkg/xds/context/aggregate_mesh_context.go
bartsmykla: Only ZoneEgresses are listed in AggregatedMeshContexts and ZoneIngress builder needs ZoneEgresses to properly build itc config

bartsmykla:
> Is there a test somewhere that you could extend to check for this case?

There is none I found

Issue #7535, opened 2023-08-17
ZoneInsights subscription are invalid


lahabana:
### What happened?

Multiple weird things here:
- not ordered by connectTime (the newest connection should always be last)
- a lot of non disconnected subscriptions seems to exist.

I've checked other insights and they don't seem to have this problem.

trimmed version using `pbpaste | jq '.zoneInsight.subscriptions | del(.[] | .status.stat) | del(.[] | .config) | del(.[] | .version)'`

```json
[
  {
    "id": "ef2f9da4-9b50-4b7f-9530-3efb897d9ebb",
    "globalInstanceId": "mink-control-plane-57cdd67669-bz6n9-9bd1",
    "connectTime": "2023-08-16T15:46:54.968654235Z",
    "status": {
      "lastUpdateTime": "2023-08-16T15:46:56.242911919Z",
      "total": {
        "responsesSent": "37",
        "responsesAcknowledged": "37"
      }
    },
    "generation": 383
  },
  {
    "id": "31f30740-bc80-4811-a262-4fcf467237b3",
    "globalInstanceId": "mink-control-plane-57cdd67669-bz6n9-9bd1",
    "connectTime": "2023-08-17T02:09:47.353040574Z",
    "status": {
      "lastUpdateTime": "2023-08-17T02:09:47.853503539Z",
      "total": {
        "responsesSent": "37",
        "responsesAcknowledged": "37"
      }
    },
    "generation": 133
  },
  {
    "id": "c0cd81d9-00fa-4c36-9b98-9c1d43fc6a57",
    "globalInstanceId": "mink-control-plane-57cdd67669-bz6n9-9bd1",
    "connectTime": "2023-08-16T17:09:12.650835657Z",
    "status": {
      "lastUpdateTime": "2023-08-16T17:09:13.281269475Z",
      "total": {
        "responsesSent": "37",
        "responsesAcknowledged": "37"
      }
    },
    "generation": 350
  },
  {
    "id": "6a67e101-b92c-4256-8ab9-1372534c6055",
    "globalInstanceId": "mink-control-plane-57cdd67669-bz6n9-9bd1",
    "connectTime": "2023-08-16T17:57:08.027177564Z",
    "status": {
      "lastUpdateTime": "2023-08-16T17:57:08.576920005Z",
      "total": {
        "responsesSent": "37",
        "responsesAcknowledged": "37"
      }
    },
    "generation": 330
  },
  {
[...]

PR #8144, opened 2023-10-25
fix(kuma-cp): don't unset disconnectTime in ZoneInsight subscriptions

michaelbeaumont:
I think [`ZoneInsightSink` can unset `disconnectTime`](https://github.com/kumahq/kuma/blob/152408bf7a8b02d712b25b4b32558b96d7b6f401/pkg/kds/server/status_sink.go#L142) that has been [marked disconnected in `UpdateSubscription`](https://github.com/kumahq/kuma/blob/152408bf7a8b02d712b25b4b32558b96d7b6f401/api/system/v1alpha1/zone_insight_helpers.go#L71) because it doesn't take the value from the current `ZoneInsight` but rather [takes the subscription](https://github.com/kumahq/kuma/blob/152408bf7a8b02d712b25b4b32558b96d7b6f401/pkg/kds/server/status_sink.go#L77) instead [from `streamState`](https://github.com/kumahq/kuma/blob/152408bf7a8b02d712b25b4b32558b96d7b6f401/pkg/kds/server/status_tracker.go#L139).

Changing this in `UpdateSubscription` is one way to ensure this doesn't happen.  

See https://github.com/kumahq/kuma/issues/7535

### Checklist prior to review

### DIFF
--- a/api/system/v1alpha1/zone_insight_helpers.go
+++ b/api/system/v1alpha1/zone_insight_helpers.go
@@ -66,6 +66,10 @@ func (x *ZoneInsight) UpdateSubscription(s generic.Subscription) error {
 	}
 	i, old := x.GetSubscription(kdsSubscription.Id)
 	if old != nil {
+		// We don't want to unset DisconnectTime
+		if old.DisconnectTime != nil {
+			kdsSubscription.DisconnectTime = old.DisconnectTime
+		}
 		x.Subscriptions[i] = kdsSubscription
 	} else {
 		x.finalizeSubscriptions()


jakubdyszkiewicz:
I'm not sure this is a right approach to be honest.
subscription in stream state means that it was not disconnected, right?

michaelbeaumont:
> I'm not sure this is a right approach to be honest.
> subscription in stream state means that it was not disconnected, right?

Does it? I'll admit I don't fully understand everywhere that's writing to ZoneInsight but this only triggers if the `disconnectTime` was previously set. If a subscription in stream state means it isn't disconnected, how can we tolerate having `disconnectTime` automatically set in `UpdateDescription`?

michaelbeaumont:
https://github.com/kumahq/kuma/pull/8153

PR #8148, opened 2023-10-25
feat(api-server): new inspect api

lahabana:
No kumactl for the moment will add in a follow up PR.

### Checklist prior to review


pkg/plugins/policies/meshtcproute/plugin/v1alpha1/plugin_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	core_xds "github.com/kumahq/kuma/pkg/core/xds"
 	"github.com/kumahq/kuma/pkg/metrics"
 	core_rules "github.com/kumahq/kuma/pkg/plugins/policies/core/rules"
+	_ "github.com/kumahq/kuma/pkg/plugins/policies/meshhttproute"
```
lukidzi, line 19: Is it needed just for plugin registration?

docs/generated/openapi.yaml
```diff
@@ -19,6 +19,62 @@ paths:
           $ref: '#/components/responses/InternalServerError'
       operationId: get-global-insight
       description: Returns a Global Insight object
+  /meshes/{mesh}/{policyType}/{policyName}/-resources/dataplanes:
```
lukidzi, line 22: Why `-resources` is it something specific?

pkg/api-server/resource_endpoints.go
```diff
@@ -438,3 +453,103 @@ func (r *resourceEndpoints) readOnlyMessage() string {
...
+				rest_errors.HandleError(request.Request.Context(), response, err, "failed inspect")
+				return
+			}
+			if err := r.resManager.List(request.Request.Context(), hl, store.ListByMesh(meshName)); err != nil {
+				rest_errors.HandleError(request.Request.Context(), response, err, "failed inspect")
+				return
+			}
+			dependentResources.MeshLocalResources[dependentType] = hl
+		}
+		filter := func(rs model.Resource) bool {
```
lukidzi, line 503: Could we extract this part until the response is prepared to a separate method like `GetAffectedDataplanes`?

docs/generated/openapi.yaml
lahabana: Prefix by dash avoids any possible collision with dynamically generated urls from resource paths in the future.
It makes it very clear that this is not an actual entity type

pkg/api-server/resource_endpoints.go
lahabana: I think we could but it'd be a function generator with a bunch of params (dependent resources and policy at least)

lukidzi (approved):

mk/docs.mk
```diff
@@ -56,7 +56,7 @@ docs/generated/openapi.yaml:
 	rm -rf $(OAPI_TMP_DIR)
 	mkdir -p $(dir $@)
 	mkdir -p $(OAPI_TMP_DIR)/policies
-	for i in $(API_DIRS); do mkdir -p $(OAPI_TMP_DIR)/$$(echo $${i} | cut -d: -f2); cp -r $$(echo $${i} | cut -d: -f1) $(OAPI_TMP_DIR)/$$(echo $${i} | cut -d: -f2); done
+	for i in $(API_DIRS); do mkdir -p $(OAPI_TMP_DIR)/$$(echo $${i} | cut -d: -f2); cp -R $$(echo $${i} | cut -d: -f1) $(OAPI_TMP_DIR)/$$(echo $${i} | cut -d: -f2); done
```
michaelbeaumont, line 59: Is this supposed to be here? Please at least a different commit, ideally PR

docs/generated/openapi.yaml
michaelbeaumont: Is this a common open API thing? It seems like we'd be unlikely to dynamically generate a path `../resources` and if we do, isn't that just kind of confusing in spite of the `/-` prefix?

pkg/api-server/resource_endpoints_test.go
```diff
@@ -1037,4 +1038,8 @@ var _ = Describe("Resource Endpoints", func() {
 		Expect(test_metrics.FindMetric(metrics, "api_server_http_requests_inflight")).ToNot(BeNil())
 		Expect(test_metrics.FindMetric(metrics, "api_server_http_response_size_bytes")).ToNot(BeNil())
 	})
+
+	DescribeTable("inspect", func(inputFile string) {
```
michaelbeaumont, line 1042: Does the fact that no test files have changed means the output for existing endpoints and resources is exactly the same?

mk/docs.mk
lahabana: Yeah it is I couldn't generate the change openapi without this

docs/generated/openapi.yaml
lahabana: It's not anything official in the openAPI spec for sure. But it's common in apis to have a syntax to clearly indicate that things are not a persisted but a computed resource. For example elastic search has a `_search` I think it's quite nice to have APIs that make this difference clear.

pkg/api-server/resource_endpoints_test.go
michaelbeaumont: OK, yeah. you mentioned fixing the inspect API in this PR earlier today so I wasn't sure but I guess that was referring to the new endpoint

docs/generated/openapi.yaml
lahabana: Ok moved to _resources

michaelbeaumont (approved):

PR #8156, opened 2023-10-26
fix(MeshGateway): ensure that duplicate listeners are not added when crossMesh is enabled on a listener and Routes specify hostnames

### LINKED ISSUE #8076: CrossMesh MeshGateway listener does not work when MeshGatewayRoute hostnames are set
### What happened?

I have a MeshGateway with a listener set up with no hostname set (defaults to '*') and crossMesh enabled. I'm trying to add MeshGatewayRoutes that have hostnames specified.

When the below configuration is set, the gateway proxy throws errors like

> [2023-10-19 21:15:58.269][26][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:128] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) vault:HTTP:8080: error adding listener '10.200.12.80:8080': filter chain '' has the same matching rules defined as ''

Looking at the generated Envoy config, I see that it is trying to set two identical filter chains for that listener.

Digging into the code, I think what is happening is that [RedistributeWildcardRoutes](https://github.com/kumahq/kuma/blob/6f86847aee5d50fd912aeaaa8f5470427f127676/pkg/plugins/runtime/gateway/generator.go#L429) is creating two host name configs, one for the route's host name (assuming you have one route with one hostname set) and one for the wildcard. 
The problem with having multiple hosts is that crossMesh enabled sets the protocol to HTTPS with no TLS mode set, then in [configureTls](https://github.com/kumahq/kuma/blob/6f86847aee5d50fd912aeaaa8f5470427f127676/pkg/plugins/runtime/gateway/filter_chain_generator.go#L439C33-L439C33) the switch block never sets `envoy_listeners.MatchServerNames(hostnames...),` as it does for other https modes.

The net effect is that the filter chains have identical matching blocks (since they are missing hostname matching) which causes the error.

MeshGateway:

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshGateway
mesh: mesh1
metadata:
  name: cross-mesh
  namespace: kuma
spec:
  conf:
    listeners:
      - crossMesh: true
        port: 8080
        protocol: HTTP
        tags:
          some: tag
```

Note that the hostname is not set and crossMesh is `true`

MeshGatewayRoute:
```yaml
piVersion: kuma.io/v1alpha1
kind: MeshGatewayRoute
mesh: mesh1
metadata:
  name: test
  namespace: kuma
spec:
  conf:
    http:
      hostnames:
      - myservice.external.mesh
      rules:
      - backends:
        - destination:
            kuma.io/service: test_test_svc_8000
        matches:
        - path:
            match: PREFIX
            value: /
  selectors:
  - match:
      kuma.io/service: edge-gateway
      some: tag
```

MeshGatewayInstance:
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshGatewayInstance
metadata:
  name: edge-gateway
  namespace: gateway
spec:
  replicas: 1
  serviceType: ClusterIP
  tags:
    kuma.io/service: edge-gateway
```

The namespace is tagged with `kuma.io/mesh: mesh1` and the gateway dataplace does seem to get the correct mesh

ttreptow:
In the cross-mesh case, the SNI string will be a kuma SNI string for the gateway service (e.g. edge-gateway{mesh=default,port=tcp-8080}). Thus it is not possible to distinguish hosts at the listener level and no filter chain sni matchers are added. This can lead to a duplicate listener filter chain being added if there are multiple hostnames to route.

Thus we truncate the gatewayHosts array to size 1 before creating the listener blocks.

Fixes #8076 

Supersedes #8105 
* The approach in my PR solves the problem in a way that leaves the door open for host-based cross-mesh routing (i.e. specifying mesh hostnames in MeshGatewayRoute). I have that working and will submit a follow on PR for it
* The tests in this PR are modified from the ones in #8105, the authorship of the original tests (@michaelbeaumont ) was lost in some git rebasing

### Checklist prior to review

- [X] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) -- I don't think this needs to be backported 

### DIFF
--- a/pkg/plugins/runtime/gateway/gateway_route_generator_test.go
+++ b/pkg/plugins/runtime/gateway/gateway_route_generator_test.go
@@ -1440,13 +1440,21 @@ conf:
   - port: 8080
     protocol: HTTP
     crossMesh: true
+  - port: 8081
+    protocol: HTTP
+    crossMesh: true
+    hostname: internal-cross-mesh.mesh
+    tags:
+      hostname: internal-cross-mesh.mesh
+  - port: 8082
+    protocol: HTTP
+    crossMesh: true
+    tags:
+      hostname: route-only
 `, `
 type: MeshGatewayRoute
 mesh: default
-name: echo-service
-selectors:
-- match:
-    kuma.io/service: gateway-default
+name: echo-service-default
 selectors:
 - match:
     kuma.io/service: gateway-default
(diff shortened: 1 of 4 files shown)


ttreptow:
I accidentally re-incorporated @michaelbeaumont 's changes while rebasing :see_no_evil: . I'm going to push a fixed commit in a minute

pkg/plugins/runtime/gateway/generator.go
```diff
@@ -256,6 +256,10 @@ func (g Generator) generateLDS(ctx xds_context.Context, info GatewayListenerInfo
 	for _, hostInfo := range hostInfos {
 		gatewayHosts = append(gatewayHosts, hostInfo.Host)
 	}
+	if info.Listener.CrossMesh {
+		//for cross-mesh, we can only add one listener filter chain as there will not be any (usable) SNI available for filter chain matching
```
michaelbeaumont, line 260: `gocritic` is complaining

```suggestion
		// for cross-mesh, we can only add one listener filter chain as there will not be any (usable) SNI available for filter chain matching
```

pkg/plugins/runtime/gateway/gateway_route_generator_test.go
```diff
@@ -1440,30 +1440,123 @@ conf:
   - port: 8080
     protocol: HTTP
     crossMesh: true
+    tags:
```
michaelbeaumont, line 1443: I'd like to avoid changing the existing test cases, can we leave the listener and existing MeshGatewayRoutes untouched?

pkg/plugins/runtime/gateway/generator.go
```diff
@@ -256,6 +256,10 @@ func (g Generator) generateLDS(ctx xds_context.Context, info GatewayListenerInfo
 	for _, hostInfo := range hostInfos {
 		gatewayHosts = append(gatewayHosts, hostInfo.Host)
 	}
+	if info.Listener.CrossMesh {
+		// for cross-mesh, we can only add one listener filter chain as there will not be any (usable) SNI available for filter chain matching
+		gatewayHosts = gatewayHosts[:1]
```
michaelbeaumont, line 261: Let's add this truncation to https://github.com/kumahq/kuma/blob/09097f213741535e7716703f92c5680188a7095d/pkg/plugins/runtime/gateway/filter_chain_generator.go#L94-L96

instead

michaelbeaumont (changes requested):

michaelbeaumont:
@ttreptow unfortunately you need to rebase and add the `--signoff` to satisfy the DCO

michaelbeaumont (approved):
Thanks @ttreptow !

PR #8168, opened 2023-10-27
fix(k8s): fix VIPs configmap entries with invalid keys for ExternalName services

bartsmykla:
Services with the `ExternalName` type do not allow us to get an IP address, so the generated VIPs config map contains entries with invalid keys, such as `1:`. This was hard to debug because we were getting an EOF error during unmarshalling of the VIPs config map.

This commit fixes the issue by ignoring services of that type with appropriate log when it'll happen.

### Checklist prior to review

  - No relevant issue
  - It won't
  - Added unit test
  - There is no need
  - It does


jakubdyszkiewicz (commented):
Could you help me understand how this fix relates to invalid VIPs?

`service_controller.go` is used to annotate services with an ngxing/kong annotation to use ClusterIP instead of PodIP.

pkg/plugins/runtime/k8s/controllers/service_controller.go
```diff
@@ -59,11 +59,11 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req kube_ctrl.Request
 	}
 
 	if svc.Spec.Type == kube_core.ServiceTypeExternalName {
-		log.V(1).Info(fmt.Sprintf(
-			"ignoring %s because it is of an unsupported type (%s)",
-			req.NamespacedName.String(),
-			kube_core.ServiceTypeExternalName,
-		))
+		log.V(1).Info(
+			"ignoring service because it is of an unsupported type",
```
michaelbeaumont, line 63:
```suggestion
			"ignoring Service because it is of an unsupported type",
```

test/e2e_env/kubernetes/externalname-services/externalname-services.go
```diff
@@ -0,0 +1,61 @@
...
+    app: test-server
+`, namespace)
+
+	BeforeAll(func() {
+		Expect(NewClusterSetup().
+			Install(MeshKubernetes(meshName)).
+			Install(NamespaceWithSidecarInjection(namespace)).
+			Install(testserver.Install(
+				testserver.WithName("test-server"),
+				testserver.WithNamespace(namespace),
```
jakubdyszkiewicz, line 42: please deploy this in created mesh. Just in case

bartsmykla: addressed

jakubdyszkiewicz (approved):

PR #8169, opened 2023-10-27
feat(dp-token): allow validator to define keys not scoped to a mesh

nicoche:
This permits dpServer.authn.dpPRoxy.dpToken.validator.publicKeys to omit the .mesh field. Keys without the .mesh field will be considered to verify the signature of dataplane tokens, no matter the mesh of the dataplane.

### Checklist prior to review

- [x] Link to relevant issue https://kuma-mesh.slack.com/archives/CN2GN4HE1/p1698324155036599  --


jakubdyszkiewicz:
Nice!

Could you also add E2E test for it?
We could improve test/e2e_env/universal/auth/offline_auth.go:13
* generate another sample key here test/keys/README.md
* add a new entry in cpConfig#dpServer.authn.dpProxy.validator.publicKeys in test
* add another `It(` (or change to DescribeTable if you want) with the new key

This way we can validate this functionality and also check if we can combine meshed and not meshed keys.

Let me know if you need help

nicoche:
Great! Honestly, I did not know where to look for, regarding the tests. I'll add that

nicoche:
Should be good now, the e2e tests pass locally 🙂

test/keys/README.md
```diff
@@ -1,4 +1,7 @@
 Keys for tests generated by executing:
 
-kumactl generate signing-key --format=pem > samplekey.pem
-kumactl generate public-key --signing-key-path=samplekey.pem > publickey.pem
+kumactl generate signing-key --format=pem > sample-key-static-1.pem
+kumactl generate public-key --signing-key-path=sample-key-static-1.pem > sample-publickey-static-1.pem
```
jakubdyszkiewicz, line 4: I believe this key is used by unit tests as well so renaming will blow those tests up.

btw. key itself has no motion of the mesh. It's also used in context of other signing keys. I'd just name this
```
samplekey.pem
samplekey-2.pem
```
or if you want to rename it
```
samplekey-1.pem
samplekey-2.pem
```

nicoche: Fair enough 🙂. I went with your first suggestion

jakubdyszkiewicz (approved):

Issue #7412, opened 2023-08-01
MeshTrafficPermission: prometheus listener doesn't allow traffic with allow-all policy


lukidzi:
### What happened?

When kuma uses the default `MeshTrafficPermission` that allows all the traffic, changes for RBAC are not applied to the `kuma:metrics:prometheus` listener. In this case, traffic from the scraper cannot reach the services endpoint.

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  namespace: kuma-system
  name: allow-all
spec:
  targetRef:
    kind: Mesh
  from:
    - targetRef:
        kind: Mesh
      default:
        action: Allow
```
`config_dump` of rbac `kuma:metrics:prometheus` listener when using `MeshTrafficPermission`
```yaml
        {
           "name": "envoy.filters.network.rbac",
           "typed_config": {
            "@type": "type.googleapis.com/envoy.extensions.filters.network.rbac.v3.RBAC",
            "rules": {},
            "stat_prefix": "kuma_metrics_prometheus."
           }
          },
```

When we switch to `TrafficPermisson` traffic works. 
```yaml
apiVersion: kuma.io/v1alpha1
kind: TrafficPermission
mesh: default
metadata:
  name: allow-all-traffic
spec:
  sources:
    - match:
        kuma.io/service: '*'
  destinations:
    - match:
        kuma.io/service: '*'
```

`config_dump` of rbac `kuma:metrics:prometheus` listener when using `TrafficPermission`

```yaml
        {
           "name": "envoy.filters.network.rbac",
           "typed_config": {
            "@type": "type.googleapis.com/envoy.extensions.filters.network.rbac.v3.RBAC",
            "rules": {
             "policies": {
              "allow-all-traffic": {
               "permissions": [
                {
                 "any": true
                }
               ],
               "principals": [
                {
                 "any": true
                }
               ]
              }
             }
            },
            "stat_prefix": "kuma_metrics_prometheus."
           }
          },
```

lukidzi:
We talked with @lobkovilya and concluded that it's not a quick fix to make it work. We use the DP object when building a rules view for the policies. That object doesn't have all inbounds that dataplane can have. E.g.: Prometheus, admin, probe. Because of this new policies cannot be applied to listeners that are not defined in a DP object. 

as a workaround, we can use just to allow prometheus traffic and policies for the services we can define with `MeshTrafficPermission`
```
[...]

Issue #5708, opened 2023-01-16
MeshMetric policy


lahabana:
### Description

We should have a policy for this like MeshTrace and MeshAccessLog.

jakubdyszkiewicz:
> Triage: requires MADS changes.
> DP overrides may not be needed anymore with TargetRef MeshService

lobkovilya:
Use cases to keep in mind when writing MADR:
* How to specify a list of services that can access the Prometheus listener?
* What if we want to configure timeout (or rate limit?) for the Prometheus listener?

PR #8172, opened 2023-10-27
fix(kuma-cp): don't configure RBAC rules on Prometheus listener

lobkovilya:
This issue is currently blocking us from fully switching from TrafficPermission to MeshTrafficPermissions. 

After a discussion with @lahabana we want to stop treating Prometheus listeners as an inbound for "fake" service. Instead [MeshMetric](https://github.com/kumahq/kuma/issues/5708) policy when implemented should contain a configuration of what service is allowed to access the Prometheus listener.

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/7412


slonka (approved):
sounds good, but tests are not passing

jakubdyszkiewicz (commented):
but this is intentional

https://kuma.io/docs/2.4.x/policies/traffic-metrics/
```
        tags: # tags that can be referred in Traffic Permission when metrics are secured by mTLS  
          kuma.io/service: dataplane-metrics
```

PR #8199, opened 2023-10-31
fix(kuma-dp): fix incorrect dataplane name due to mangled env vars

bartsmykla:
Kubernetes workloads may have their environment variables mangled by other components outside of our control. This can cause the `POD_NAME` or `POD_NAMESPACE` variables to be defined after the `KUMA_DATAPLANE_NAME` variable, which results in an incorrect dataplane name.

This commit fixes this issue by checking for the `POD_NAME` and `POD_NAMESPACE` variables directly during Kuma dataplane runtime initialization if `KUMA_DATAPLANE_NAME` is empty or contains the characters `$(){}`, which suggests that the dependent variables were not present before its definition. This involved adding a `PostProcess` method to the `Config` interface.

### Checklist prior to review

  - there is no relevant issue
  - it won't
  - unit tests were added
  - manually tested on local k3d cluster
  - there is no need
  - it does


pkg/config/app/kuma-dp/config.go
```diff
@@ -116,8 +131,40 @@ type Dataplane struct {
...
+		if !ok {
+			return nil
+		}
+
+		podNamespace, ok := os.LookupEnv("POD_NAMESPACE")
+		if !ok {
+			return nil
+		}
+
+		d.Name = fmt.Sprintf("%s.%s", podName, podNamespace)
```
lahabana, line 146: Should you redo validateMeshOrName here?

bartsmykla: addressed

lahabana (commented):
missing zoneProxy helm chart configuration change.

lahabana (approved):
Approved! THough I'm not sure we should backport this as it changes quite a lot.

Issue #8073, opened 2023-10-19
Implement Basic functionality of Locality Aware Load Balancing


lukidzi:
Implement MADR without Egress support

PR #8233, opened 2023-11-02
feat(kuma-cp): add locality aware implementation for egress

Automaat:
Fixes:  #8073

### Checklist prior to review


pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go
```diff
@@ -88,7 +88,7 @@ type FromZone struct {
 type ToZone struct {
 	// Type defines how target zones will be picked from available zones
 	Type  ToZoneType `json:"type"`
-	Zones []string   `json:"zones"`
+	Zones *[]string  `json:"zones,omitempty"`
```
lukidzi, line 91: This is because when Any/None we don't need to provide list of zones

test/e2e_env/multizone/localityawarelb/locality_aware_multizone_hybrid_egress.go
```diff
@@ -0,0 +1,134 @@
...
+
+	E2EAfterAll(func() {
+		Expect(multizone.UniZone1.DeleteMeshApps(mesh)).To(Succeed())
+		Expect(multizone.UniZone2.DeleteMeshApps(mesh)).To(Succeed())
+		Expect(multizone.KubeZone1.TriggerDeleteNamespace(namespace)).To(Succeed())
+		Expect(multizone.Global.DeleteMesh(mesh)).To(Succeed())
+	})
+
+	It("should route based on defined strategy with egress enabled", func() {
+		Eventually(func() (map[string]int, error) {
```
lukidzi, line 108: We can maybe make it similar to the current test without egress.
1. Check traffic without policy
2. Apply policy
3. Check when app is down
4. Check when the next is down

and we can add `E2EAfterEach` which removes the policy

pkg/xds/sync/egress_proxy_builder.go
```diff
@@ -94,24 +95,21 @@ func (p *EgressProxyBuilder) Build(
...
 			}
 			meshResources.Dynamic[es.Spec.GetService()] = policies
 		}
 
+		for serviceName := range meshResources.EndpointMap {
+			policies, err := matchExternalPolicies(map[string]string{mesh_proto.ServiceTag: serviceName}, meshCtx.Resources)
+			if err != nil {
+				return nil, err
+			}
+			meshResources.Dynamic[serviceName] = policies
```
lukidzi, line 110: Won't it affect other plugins? @lobkovilya

pkg/plugins/policies/meshloadbalancingstrategy/plugin/v1alpha1/priority.go
```diff
@@ -69,15 +69,15 @@ func getCrossZoneLbGroups(conf *api.Conf, localZone string) []CrossZoneLbGroup {
...
+				for _, zone := range *rule.To.Zones {
 					zones[zone] = true
 				}
 				lb.Zones = zones
 				lb.Priority = uint32(priority + 1)
 			case api.Only:
 				lb.Type = api.Only
 				zones := map[string]bool{}
-				for _, zone := range rule.To.Zones {
+				for _, zone := range *rule.To.Zones {
```
lukidzi, line 80: Don't we need to check if isn't nul `*rule.To.Zones`

slonka (approved):
I'm approving but I'm not the most knowledge person about egress routing

pkg/plugins/policies/meshloadbalancingstrategy/plugin/v1alpha1/priority.go
Automaat: validation on API is checking this, can there be a situation when this will be nil? After some merging?

pkg/xds/sync/egress_proxy_builder.go
Automaat: It shouldn't affect other plugins, we are only adding possible services on which we can apply policies

bartsmykla (approved):
I don't have any other suggestions/questions than already existing, so LGTM.

pkg/plugins/policies/meshloadbalancingstrategy/plugin/v1alpha1/priority.go
lukidzi: Not sure if that can happen but would be good to avoid nil pointer if it's somehow possible

test/e2e_env/multizone/localityawarelb/locality_aware_multizone_hybrid_egress.go
```diff
@@ -0,0 +1,134 @@
...
+  to:
+    - targetRef:
+        kind: MeshService
+        name: test-server_locality-aware-lb-egress_svc_80
+      default:
+        localityAwareness:
+          crossZone:
+            failover:
+              - from: 
+                  zones: ["kuma-4"]
```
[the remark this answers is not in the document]
Automaat: these are defined in multizone e2e suite setup

pkg/plugins/policies/core/matchers/egress.go
```diff
@@ -56,14 +55,14 @@ func EgressMatchedPolicies(rType core_model.ResourceType, es *core_mesh.External
 }
 
 func processFromRules(
-	es *core_mesh.ExternalServiceResource,
+	tags map[string]string,
```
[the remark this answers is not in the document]
Automaat: `TagSelector` does not fit the context, we are passing tags not the selector, and we don't have `Tag` specific type yet

test/e2e_env/multizone/localityawarelb/locality_aware_multizone_hybrid_egress.go
Automaat: changed test, found bug. And changed test again to cover more complicated use case:

```
1. Check traffic without policy
2. Apply policy
3. Check when app is down
4. Change policy priorities
5. Check traffic
6. Check when the next is down
```

pkg/plugins/policies/meshloadbalancingstrategy/plugin/v1alpha1/priority.go
Automaat: added check

test/e2e_env/multizone/localityawarelb/locality_aware_multizone_hybrid_egress.go
```diff
@@ -0,0 +1,190 @@
...
+                to:
+                  type: Any`, mesh)
+
+	BeforeAll(func() {
+		// Global
+		Expect(NewClusterSetup().
+			Install(ResourceUniversal(samples.MeshMTLSBuilder().WithName(mesh).WithEgressRoutingEnabled().Build())).
+			// Install(YamlUniversal(meshLoadBalancingStrategyDemoClient)).
```
slonka, line 85: I guess we can remove that if it's commented out, right?

slonka (approved):

pkg/xds/sync/egress_proxy_builder.go
lobkovilya: It seems like it only affects the situations when we have an internal service with the same `kuma.io/service` tag as external?

pkg/xds/sync/egress_proxy_builder.go
```diff
@@ -134,3 +132,23 @@ func (p *EgressProxyBuilder) Build(
 
 	return proxy, nil
 }
+
+func matchExternalPolicies(tags map[string]string, resources xds_context.Resources) (core_xds.PluginOriginatedPolicies, error) {
```
lobkovilya, line 136: nit: we probably should rename the method since it works for both internal and external services now

Automaat: renamed

lukidzi (approved):

Issue #8167, opened 2023-10-27
Using policy name with "." causes hash to be inserted in the wrong place on the zone


slonka:
### What happened?

When creating a policy like this:

```
type: MeshAccessLog
name: valid.name
mesh: default
spec:
  targetRef:
    kind: MeshServiceSubset
    name: frontend
    tags:
      version: canary
  to:
    - targetRef:
        kind: MeshService
        name: backend
      default:
        backends:
          - type: File
            file:
              path: /tmp/access.log
```

causes the zone to get 

```
name: valid-bv979f8v7wfw4v54.name
```

which is not right

jakubdyszkiewicz:
> Triage: we don't know yet how to solve it

PR #8240, opened 2023-11-02
fix(kuma-cp): using policy name with "." causes hash to be inserted in the wrong place on the zone

lobkovilya:
Since we allow a dot in the policy name, we can't use functions like `CoreNameToK8sName` that try to parse the name and make assumptions based on `.<namespace>` suffix.

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/8167


pkg/plugins/resources/k8s/mapper_test.go
```diff
@@ -34,19 +34,6 @@ var _ = Describe("KubernetesStore template", func() {
 			Expect(res.GetNamespace()).To(Equal("core-ns"))
 		})
 
-		It("works with namespace in name on inference", func() {
```
lobkovilya, line 37: @lahabana is it a valid test case? Maybe I'm missing something, but InferenceMapper runs only on non-k8s store types, so how can it get a name with ns-suffix as an input?

lahabana: Hmm I'm not sure. Global k8s syncing a k8s zone maybe?

lahabana:
Ok I'm completely confused with names. Can we have a doc that describes what happens to names between global and zones in all cases (global k8s, global universal, zone k8s, zone universal) ?

lobkovilya:
Sure, do you think it should be a kuma-website doc or a more developer-oriented doc in this repo?

lobkovilya:
@lahabana I pushed the doc under the `docs/guides` for now, please check if it makes sense to you

docs/guides/kds-sync-name.md
```diff
@@ -0,0 +1,186 @@
+# What is the name of the resource after KDS sync?
```
lahabana, line 1: If the code for this is contained in one place in the code (which it should probably) we should just have this as comments in the code.

docs/guides/kds-sync-name.md
```diff
@@ -0,0 +1,186 @@
...
+
+```
+nameInZone := "${truncate(nameInGlobal, 236)}-${hash(mesh, nameInGlobal)}"
+```
+
+Example: `my-policy -> my-policy-fvdw2c2wx44fv4xz`
+
+### Zone(k8s/universal) -> Global(k8s/universal)
+
+Not supported
```
lahabana, line 18: Are we sure of how it'll work once we relax this limitation?

docs/guides/kds-sync-name.md
```diff
@@ -0,0 +1,186 @@
...
+
+## New policies (i.e. MeshTrafficPermission, MeshHTTPRoute)
+
+* CRD scope: Namespace
+* Kuma scope: Mesh
+
+### Global(k8s) -> Zone(k8s)
+
+```
+nameInGlobal = trimNsSuffix(nameInGlobal)
```
lahabana, line 28: There's a ns suffix in Global k8s?

lahabana: To me on global k8s you wouldn't have nsSuffixes if they are created in k8s no?

docs/guides/kds-sync-name.md
```diff
@@ -0,0 +1,186 @@
...
+```
+nameInZone := "${truncate(nameInGlobal, 236)}-${hash(mesh, nameInGlobal)}.${systemNamespace}"
+```
+
+Example: `my-policy -> my-policy-fvdw2c2wx44fv4xz.kuma-system-zone`
+
+### Global(k8s) -> Zone(universal)
+
+```
+nameInGlobal = trimNsSuffix(nameInGlobal)
```
lahabana, line 45: Same here. How com there's an ns suffix here?

docs/guides/kds-sync-name.md
```diff
@@ -0,0 +1,186 @@
...
+nameInZone := "${truncate(nameInGlobal, 236)}-${hash(mesh, nameInGlobal)}"
+```
+
+Example: `my-policy.kuma-system-global -> my-policy-fvdw2c2wx44fv4xz`
+
+### Global(universal) -> Zone(universal)
+
+```
+nameInZone := "${truncate(nameInGlobal, 236)}-${hash(mesh, nameInGlobal)}"
```
lahabana, line 54: Why doing truncation when doing universal --> universal ? I thought there wasn't a name length limit.

docs/guides/kds-sync-name.md
```diff
@@ -0,0 +1,186 @@
...
+Example: `my-dpp.kuma-demo -> zone-1.my-dpp.kuma-demo.kuma-system` 
+
+Notes: 
+* in KDS v1 (default at this moment) instead of `.kuma-system` we have `.default`
+* DPP name collision is possible https://github.com/kumahq/kuma/issues/8249
+
+### Zone(universal) -> Global(k8s)
+
+```
+nameInGlobal := "${zone}.${nameInZone}.${systemNamespce}"
```
lahabana, line 87: Why do we have a systemNamespace here? These are already k8s resources right?

pkg/core/resources/model/resource.go
```diff
@@ -324,15 +334,21 @@ type ResourceMeta interface {
...
-	if ns := refMeta.GetNameExtensions()[v1alpha1.KubeNamespaceTag]; ns != "" {
-		return equalNames(refMeta.GetMesh(), util_k8s.K8sNamespacedNameToCoreName(refName, ns), resourceMeta.GetName())
+	nsRef := refMeta.GetNameExtensions()[K8sNamespaceComponent]
+	nsRes := refMeta.GetNameExtensions()[K8sNamespaceComponent]
+	if nsRef == "" || nsRef != nsRes {
+		return false
 	}
 
-	return false
+	return equalNames(refMeta.GetMesh(), refName, resourceMeta.GetNameExtensions()[K8sNameComponent])
```
lahabana, line 351: So here we get these NameExtensions it's never been mentioned anywhere in the previous docs.

docs/guides/kds-sync-name.md
lobkovilya: We have nsSuffixes only in k8s, Kuma core model doesn't know about "namespace" and that's why we encode it as suffix.

docs/guides/kds-sync-name.md
lobkovilya: `nameInGlobal` is how Kuma core model sees the resource name

pkg/core/resources/model/resource.go
lobkovilya: It's just a name without nsSuffix, I'm avoiding name parsing (because dot is allowed in the name) so I took name from the name extension

docs/guides/kds-sync-name.md
lobkovilya: No, we have a name length limit https://github.com/kumahq/kuma/blob/60733cb1f89c08de1f4e4fa25036ced4de2f2f5f/pkg/plugins/resources/postgres/migrations/data/1677096751_update_resources_name_limit.up.sql#L2

docs/guides/kds-sync-name.md
lahabana: Ha actually I think we allow zone side writes only for new policies right?

lobkovilya: Yes, only new ones

docs/guides/kds-sync-name.md
```diff
@@ -0,0 +1,186 @@
...
+Example: `my-dpp -> zone-1.my-dpp.kuma-system`
+
+Notes:
+* in KDS v1 (default at this moment) instead of `.kuma-system` we have `.default`
+* DPP name collision is possible https://github.com/kumahq/kuma/issues/8249
+
+### Zone(universal) -> Global(universal)
+
+```
+nameInGlobal := "${zone}.${nameInZone}"
```
lukidzi, line 99: So can we have a collision when doing 2 dps with the same name but different mesh? Or should we maybe protect from this?

lobkovilya: If global is universal, then `mesh` is part of the key. For global on k8s there the issue https://github.com/kumahq/kuma/issues/8249

docs/guides/kds-sync-name.md
```diff
@@ -0,0 +1,186 @@
...
+
+Example: `my-ingress -> zone-1.my-ingress`
+
+### Global(k8s) -> Zone(k8s)
+
+```
+nameInZone := "${nameInGlobal}.${systemNamespace}"
+```
+
+Example: `zone-1.my-ingress.kuma-system-zone-1.kuma-system-global -> zone-1.my-ingress.kuma-system-zone-1.kuma-system-global.kuma-system-zone-2`
```
lukidzi, line 153: Question: Does 2nd zone needs to know what was the namespace if ingress in zone1?

lobkovilya: Yes, what if you have ingresses in different namespaces in zone-1? We don't want names to collide in zone-2

lukidzi (approved):

PR #8241, opened 2023-11-02
feat(kuma-dp): support setting Envoy's --component-log-level

michaelbeaumont:
### Checklist prior to review

### DIFF
--- a/pkg/config/app/kuma-dp/config.go
+++ b/pkg/config/app/kuma-dp/config.go
@@ -189,6 +189,8 @@ type DataplaneRuntime struct {
 	// Available values are: [trace][debug][info][warning|warn][error][critical][off]
 	// By default it inherits Kuma DP logging level.
 	EnvoyLogLevel string `json:"envoyLogLevel,omitempty" envconfig:"kuma_dataplane_runtime_envoy_log_level"`
+	// EnvoyComponentLogLevel configures Envoy's --component-log-level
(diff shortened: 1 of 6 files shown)


pkg/config/app/kuma-dp/config.go
```diff
@@ -189,6 +189,8 @@ type DataplaneRuntime struct {
 	// Available values are: [trace][debug][info][warning|warn][error][critical][off]
 	// By default it inherits Kuma DP logging level.
 	EnvoyLogLevel string `json:"envoyLogLevel,omitempty" envconfig:"kuma_dataplane_runtime_envoy_log_level"`
+	// EnvoyComponentLogLevel configures Envoy's --component-log-level
```
lahabana, line 192: Can you add a bit more content re:format?

lahabana (approved):

pkg/config/app/kuma-dp/config.go
michaelbeaumont: Should it just be "look at the Envoy docs"? This will always correspond to the Envoy version in use, not Kuma.

bartsmykla (approved):

pkg/config/app/kuma-dp/config.go
michaelbeaumont: @lahabana I've added a sentence linking to the envoy docs

PR #8244, opened 2023-11-03
test(kumactl) adjust matcher for install tproxy test

bartsmykla:
### Checklist prior to review

  - no relevant issue
  - it won't
  - it contains only changes in tests
  - no need
  - no need

> Changelog: skip

### DIFF
--- a/app/kumactl/cmd/install/install_transparent_proxy_test.go
+++ b/app/kumactl/cmd/install/install_transparent_proxy_test.go
@@ -88,7 +88,7 @@ var _ = Describe("kumactl install transparent proxy", func() {
 				)
 			},
 			errorMatcher: HavePrefix("# [WARNING] error occurred when validating if 'conntrack' iptables module is present. Rules for DNS conntrack zone splitting won't be applied:"),
-			goldenFile:   "install-transparent-proxy.dns.golden.txt",
+			goldenFile:   "install-transparent-proxy.dns.no-conntrack.golden.txt",
 		}),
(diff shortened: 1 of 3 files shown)


app/kumactl/cmd/install/install_transparent_proxy_test.go
```diff
@@ -23,7 +23,8 @@ var _ = Describe("kumactl install transparent proxy", func() {
 	})
 
 	type testCase struct {
-		extraArgs    []string
+		skip      func(stdout, stderr *bytes.Buffer) bool
+		extraArgs []string
```
slonka: nitpick: I think this will fail linter (if not, ignore)

```suggestion
		skip         func(stdout, stderr *bytes.Buffer) bool
		extraArgs    []string
```

app/kumactl/cmd/install/install_transparent_proxy_test.go
```diff
@@ -38,6 +39,11 @@ var _ = Describe("kumactl install transparent proxy", func() {
 
 			// when
 			err := rootCmd.Execute()
+
+			if given.skip != nil && !given.skip(stdout, stderr) {
```
slonka, line 43: how do we prevent this from suddenly disabling on `CI`? I think there should be a hard force for `CI=true` so some day unnoticed we skip these tests.

slonka (commented):
It's ok to skip locally but we should really force it on CI.

slonka (approved):
talked offline with Bart, this adds a case that works with conntrack (so it works on local machine). We either run conntrack/noconntrack test. On CI we always run without conntrack.

PR #6083, opened 2023-02-22
feat(tproxy/cni): make tproxy v2 and CNI v2 default

bartsmykla:
Make transparent proxy engine v2 and CNI v2 default

This PR temporary disables re-install transparent-proxy test as it basically never worked as expected. When we are installing tproxy again, it just adds the same set of rules after the existing ones (with small exception to DNS ones, but it's irrelevant here). So the tests were passing only because rules form the second installation had been just ignored.
When mechanism to uninstall tproxy will be implemented (ref. https://github.com/kumahq/kuma/issues/6093), we will adapt this test to run the uninstaller step in between.

Closes: https://github.com/kumahq/kuma/issues/5880

- [x] [Link to relevant issue][1] as well as docs and UI issues -- https://github.com/kumahq/kuma/issues/5880
- [x] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [x] Tests (Unit test, E2E tests, manual test on universal and k8s) --
- [ ] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? -- **WIP**
- [x] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? -- it doesn't
- [x] Do you need to explicitly set a [`> Changelog:` entry here](../blob/master/CONTRIBUTING.md#submitting-a-patch) or add a `ci/` label to run fewer/more tests?

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #8262, opened 2023-11-06
feat(kds): enable delta by default

lahabana:
### Checklist prior to review

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -57,6 +57,7 @@ still supported but support for these WILL be removed in a future release.
 ### KDS Delta enabled by default
 
 KDS Delta is enabled by default. You can fallback to SOTW KDS by setting `KUMA_EXPERIMENTAL_KDS_DELTA_ENABLED=false`.
+As a side effect, on kubernetes policies synced will be persisted in the `kuma-system` namespace instead of `default`.
 
 ## Upgrade to `2.4.x`
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -54,6 +54,10 @@ That's why we decided to discontinue TrafficPermission support on the Prometheus
 We support `v1` resources and `v1.0.0` of `gateway-api`. `v1beta1` resources are
 still supported but support for these WILL be removed in a future release.
 
+### KDS Delta enabled by default
+
+KDS Delta is enabled by default. You can fallback to SOTW KDS by setting `KUMA_EXPERIMENTAL_KDS_DELTA_ENABLED=false`.
+
 ## Upgrade to `2.4.x`
 
 ### Configuration change
--- a/deployments/charts/kuma/templates/_helpers.tpl
+++ b/deployments/charts/kuma/templates/_helpers.tpl
@@ -275,7 +275,7 @@ env:
 - name: KUMA_RUNTIME_KUBERNETES_INJECTOR_EBPF_PROGRAMS_SOURCE_PATH
   value: {{ .Values.experimental.ebpf.programsSourcePath }}
 {{- end }}
-{{- if eq .Values.experimental.deltaKds "false" }}
+{{- if eq .Values.experimental.deltaKds false }}
 - name: KUMA_EXPERIMENTAL_KDS_DELTA_ENABLED
   value: "false"
 {{- end }}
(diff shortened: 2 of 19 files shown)


lobkovilya:
Should we mention in UPGRADE.md that policies will be synced to the system namespace (instead of default namespace)? It shouldn't require any actions from the user to work, but the user may search policies in the wrong place after the upgrade

UPGRADE.md
```diff
@@ -54,6 +54,11 @@ That's why we decided to discontinue TrafficPermission support on the Prometheus
 We support `v1` resources and `v1.0.0` of `gateway-api`. `v1beta1` resources are
 still supported but support for these WILL be removed in a future release.
 
+### KDS Delta enabled by default
+
+KDS Delta is enabled by default. You can fallback to SOTW KDS by setting `KUMA_EXPERIMENTAL_KDS_DELTA_ENABLED=false`.
```
slonka, line 59: This feels a bit confusing. If it's not experimental now (I guess that's why we're setting it as default) the flag should be removed/renamed to:

`KUMA_KDS_DELTA_ENABLED` and set to `true` by default.

The same way we did it for `transparent-proxy` - https://github.com/kumahq/kuma/pull/6083/files#diff-f5545ba9479927ba075dea0dc9b17e5680a7806c7860a43746cbe46d287c36ceL5205-R5244 (we removed `--experimental-transparent-proxy-engine` and added `--use-transparent-proxy-engine-v1`)

slonka (changes requested):
If we're setting something as default it shouldn't be called `experimental`.

deployments/charts/kuma/templates/_helpers.tpl
```diff
@@ -275,9 +275,9 @@ env:
 - name: KUMA_RUNTIME_KUBERNETES_INJECTOR_EBPF_PROGRAMS_SOURCE_PATH
   value: {{ .Values.experimental.ebpf.programsSourcePath }}
 {{- end }}
-{{- if .Values.experimental.deltaKds }}
+{{- if eq .Values.experimental.deltaKds false }}
```
lahabana, line 278:
```suggestion
{{- if not .Values.experimental.deltaKds }}
```

doesn't work?

test/e2e/helm/kuma_helm_upgrade.go
```diff
@@ -69,7 +69,6 @@ func UpgradingWithHelmChart() {
 			// then CRD is upgraded
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(ContainSubstring("AllowWithShadowDeny"))
-			// remove this when+then after initialChartVersion is changed to 2.1.x or later
```
lahabana, line 72: I think this whole test is not really doing anything useful now.

lukidzi: It seems like it upgrades and check MTP default value. I think we can remove it now and in case we need it just write a new one.

UPGRADE.md
slonka: closing since this is going to be resolved in https://github.com/kumahq/kuma/issues/8290

test/e2e/helm/e2e_suite_test.go
```diff
@@ -14,9 +14,9 @@ func TestE2E(t *testing.T) {
 }
 
 var (
-	_ = Describe("Test Zone and Global with Helm chart", Label("job-2"), Label("arm-not-supported"), helm.ZoneAndGlobalWithHelmChart, Ordered)
-	_ = Describe("Test Zone and Global universal mode with Helm chart", Label("job-0"), Label("arm-not-supported"), helm.ZoneAndGlobalInUniversalModeWithHelmChart, Ordered)
-	_ = Describe("Test Global and Zone universal mode with Helm chart", Label("job-0"), Label("arm-not-supported"), helm.GlobalAndZoneInUniversalModeWithHelmChart, Ordered)
-	_ = Describe("Test App deployment with Helm chart", Label("job-2"), Label("arm-not-supported"), helm.AppDeploymentWithHelmChart)
-	_ = Describe("Test upgrading with Helm chart", Label("job-2"), Label("arm-not-supported"), helm.UpgradingWithHelmChart)
```
slonka, line 21: seems not related but 👍 if we can run it on `arm` as well

slonka (approved):

PR #8268, opened 2023-11-06
fix(core): better lifecycle when context is getting cancelled

lahabana:
- For kuma-dp we properly shutdown when context is closed
- In KDS the client cancelled being closed make all the components stop

### Checklist prior to review

### DIFF
--- a/app/kuma-dp/pkg/dataplane/command/base_command.go
+++ b/app/kuma-dp/pkg/dataplane/command/base_command.go
@@ -1,5 +1,3 @@
-//go:build !linux && !windows
-
 package command
 
 import (
@@ -7,9 +5,10 @@ import (
 	"io"
 	"os/exec"
 	"syscall"
+	"time"
 )
 
-func BuildCommand(
+func baseBuildCommand(
 	ctx context.Context,
 	stdout io.Writer,
 	stderr io.Writer,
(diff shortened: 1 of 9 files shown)


app/kuma-dp/pkg/dataplane/command/base_command.go
```diff
@@ -0,0 +1,27 @@
...
+	name string,
+	args ...string,
+) *exec.Cmd {
+	command := exec.CommandContext(ctx, name, args...)
+	command.Stdout = stdout
+	command.Stderr = stderr
+	command.Cancel = func() error {
+		return command.Process.Signal(syscall.SIGTERM)
+	}
+	command.WaitDelay = time.Second * 5
```
slonka, line 24: why do we need this? is this 5 second associated with any other value that we wait on?

pkg/kds/mux/client.go
```diff
@@ -284,79 +242,71 @@ func (c *client) startClusters(
...
-			return
-		}
-		log.Error(err, "Envoy Admin rpc stream failed prematurely, will restart in background")
-		if err := stream.CloseSend(); err != nil {
-			log.Error(err, "CloseSend returned an error")
-		}
-		errorCh <- err
+	err := <-processingErrorsCh
+	if status.Code(err) == codes.Unimplemented {
+		log.Error(err, "rpc stream failed, because remote CP does not implement this rpc. Upgrade remote CP.")
```
slonka, line 297: nitpick: since https://github.com//kumahq/kuma/pull/2125 we don't use "remote CP"

```suggestion
		log.Error(err, "rpc stream failed, because zone CP does not implement this rpc. Upgrade zone CP.")
```

slonka (approved):
in general looks good, but there is something wrong with the unit tests https://app.circleci.com/pipelines/github/kumahq/kuma?branch=pull%2F8268

app/kuma-dp/pkg/dataplane/command/base_command.go
lahabana: So this is the time between a context cancelled and clean shutdown to be issued and a Kill if the process is still around.
It's a fail safe to make sure processes don't leak. I don't think it's worth making it configurable yet as it's very much just a fail safe.

pkg/kds/mux/client.go
lahabana: It's actually global cp

Issue #8254, opened 2023-11-03
Need a way to enable dns logging/debugging


timothy-spencer:
### Description

As somebody who is trying to set up a mesh in kubernetes and run terraform in it, I need a way to turn on coredns query logging or debugging on, so that I can use the logs from that to enumerate all the external services that I need to allow outbound access to.

I have tried editing the running kuma-sidecar Corefile to add dump/log/debug, and then sending a SIGHUP to coredns, but stdout seems to not be going to the same stdout as the envoy/dp stuff.  Either that, or I don't know how to turn on logging properly, or I'm not allowed to SIGHUP it or something.

In any case, either some documentation on the proper way to do this, or even better, some sort of option in the helm chart that will enable query logging would be REALLY useful to people trying to debug this sort of thing.

Thanks, and have fun!

lahabana:
Duplicate of https://github.com/kumahq/kuma/issues/5516

Issue #5516, opened 2022-12-16
Configure CoreDNS log level with Kuma DP env vars


jakubdyszkiewicz:
### Description

Similar to how we can set Envoy log level with `kuma_dataplane_runtime_envoy_log_level` we should have similar setting for CoreDNS, so it's easy to enable extra logging with `kuma.io/sidecar-env-vars: COREDNS_LOG_LEVEL=debug`

timothy-spencer:
I have a more specific use case, if that helps:

I need to enable query logs from coredns to be logged, so that I could enumerate all the external services that my pod is trying to access.  

I'm not sure if debug mode gets me that, but if it does, I would be very happy.  If it doesn't, then https://github.com/kumahq/kuma/issues/8254 would need to be opened again, or the scope of this issue would need to be expanded to enable query logs as well.

lahabana:
xref:

- https://github.com/kumahq/kuma/issues/8119

I think ^ might be a more generic way to achieve a lot of these confs we're being asked for

Issue #8119, opened 2023-10-24
Put coredns configuration in a configmap


michaelbeaumont:
### Description

This would give users
1. visibility
1. configurability

into how we configure coredns to handle DNS requests in the mesh

lahabana:
Only question is universal. Having it embedded is nice for universal

jakubdyszkiewicz:
> Triage: it's still useful to have this embedded in kuma-dp run

lahabana:
xref: https://github.com/kumahq/kuma/issues/8269

jijiechen:
Replacing https://github.com/kumahq/kuma/issues/8633

More background:

## Description
People want to customize the DNS template used by the kuma data plane to address following issues:

* Integrate with their existing DNS servers
* Resolve certain network issues introduced by the DNS proxying model of kuma dp, e.g. it's found by some Kuma clients that the DNS in kuma data plane is not compatible with certain alpine images (https://github.com/alpinelinux/docker-alpine/issues/155#issuecomment-833676652)

It's good to make our builtin Corefile template customizable.

It's best to support for both k8s and universal data planes.

## Not in scope: 

configuring command line of CoreDNS at data planes parameters from the control plane

jijiechen:
The PR has now been merged and we'll see if we need to backport to older versions.

PR #8285, opened 2023-11-08
feat(kuma-dp): add coredns logging flag

timothy-spencer:
As somebody who is trying to set up a mesh in kubernetes and run terraform in it, I need a way to turn on CoreDNS query logging or debugging on, so that I can use the logs from that to enumerate all the external services that I need to allow outbound access to.

This patch adds a flag to the helm chart that will inject an environment variable into the dataplane sidecars that will add the "log" plugin to the CoreDNS config. I added a test to make sure the config would get set in dnsserver, and I checked that `helm template` would add the `KUMA_RUNTIME_KUBERNETES_INJECTOR_BUILTIN_DNS_LOGGING` env var to the control plane, which I believe will cause the kuma-dp sidecars to get logging turned on, but I'm not easily able to test that part out, so I'm hoping to have somebody who has a good test harness try turning that on to see if dns query logs start appearing on stdout.

I also ran `make check` and `make test UPDATE_GOLDEN_FILES=true`, which updated a fair number of things as well.

Fix https://github.com/kumahq/kuma/issues/8254 and maybe https://github.com/kumahq/kuma/issues/5516

### Checklist prior to review

- [X] Relevant issues:  https://github.com/kumahq/kuma/issues/8254) and maybe https://github.com/kumahq/kuma/issues/5516 --
- [X] Tests (Unit test, E2E tests, manual test on universal and k8s) -- I ran `make test UPDATE_GOLDEN_FILES=true` and it passed except for `File limits [It] should raise the open file limit`, which is probably because my mac has all sorts of things locked down on it.
- [X] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? -- This just adds a feature that can be turned on.  No migration needed.
- [X] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) -- No


lahabana:
@timothy-spencer thanks for doing this! I'm thinking https://github.com/kumahq/kuma/issues/8119 might be a better way to achieve this as it would avoid the blow up in number of configs passed through to the kuma-dp.

Also `make k3d/restart` is an ok way to start things locally to check if things have the intended behaviour.

timothy-spencer:
Well, it's moot for now.  Thank you for the `make k3d restart` hint.  I have tried this out with the value set in the helm chart, and it's not giving any logs out, so I need to do some more debugging.

timothy-spencer:
So, I was wrong about it not working!  Turns out that `make k3d restart` installs it directly, and not with the helm chart like I thought, so once I adjusted the makefile to add `--set dataPlane.dnsLogging=true \` to the helm upgrade and did `make k3d/stop && make k3d/start && k3d/deploy/helm && make k3d/deploy/demo` it worked!

Anyways, this PR is available for you.  I'd still love it if it were incorporated!

timothy-spencer:
@lahabana , I have been looking at what it would take to do https://github.com/kumahq/kuma/issues/8119, and it looks like it will require a lot more work to implement than this PR, as you'll have to pass in config option(s) to tell it to use a configmap and what that configmap should be named, so it would be pretty much the same as what we have here.  You'd also have to set that configmap up in the chart, the tests, the k3d stuff, and so on.  If that's what you want to do, that's fine with me.  Or maybe I'm not reading this properly and there is a simple way to do it?

But in the meantime, is there any way you could consider adding this PR in?  I really would love to use kuma, but also really don't want to have to maintain a fork.  Thanks for your consideration!

lahabana:
@timothy-spencer we'll look at this very soon. We're releasing 2.5 atm and are all a little bit busy :)

lahabana:
@timothy-spencer @jijiechen from the team is going to look at providing coreDNS config as a configMap (https://github.com/kumahq/kuma/issues/8119) which should make this PR unnecessary. Is it ok if I close your PR?

timothy-spencer:
It's your project.  You are free to close whatever you want.  :-)  

That said, I really think that this feature would be a really good thing for others.  Being able to turn on dns debugging seems like a really good feature flag to support out of the box, so that people don't have to understand the inner workings of kuma and understand that they really need to edit a coredns config in a configmap to turn on dns logging so that they can see what external services are being denied.  I think that people will just want to be able to turn on a debugging flag that is in the helm chart or an env var.

If you are implementing a more general configmap solution, that's great.  But that code has yet to be written (and I suspect will take a while to figure out now to do well, from what I've seen so far), and this is working right now.  I don't understand why you wouldn't accept this PR.

Again, it's your project.  It's fine to close this if you like. I don't have the whole vision of the product in my head like you all do, and I probably am writing ugly code that doesn't fit in with your local coding standards/practices, and there are no wrong answers here.  But as a user who ran into this problem in the field, this PR is the type of solution that I would want.

timothy-spencer:
Whoops!  I didn't mean to close this.  That's your job.  :-) I managed to click on "sync repo" on my branch and wipe out my changes.  I'll try to bring this stuff back. I'm still hoping you will want to use this code.

PR #8312, opened 2023-11-09
feat(kuma-cp): map BadRequest error to 400

lukidzi:
### Checklist prior to review

### DIFF
--- a/pkg/core/rest/errors/error_handler.go
+++ b/pkg/core/rest/errors/error_handler.go
@@ -39,6 +39,12 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 			Title:  "Bad Request",
 			Detail: err.Error(),
 		}
+	case errors.Is(err, &BadRequest{}):
+		kumaErr = &types.Error{
+			Status: 400,
+			Title:  title,
+			Detail: err.Error(),
+		}
 	case manager.IsMeshNotFound(err):
 		kumaErr = &types.Error{
 			Status: 400,
--- a/pkg/core/rest/errors/error_handler.go
+++ b/pkg/core/rest/errors/error_handler.go
@@ -33,7 +33,10 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 			Title:  title,
 			Detail: "Not found",
 		}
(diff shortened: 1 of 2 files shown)


pkg/core/rest/errors/error_handler.go
```diff
@@ -39,6 +39,12 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 			Title:  "Bad Request",
 			Detail: err.Error(),
 		}
+	case errors.Is(err, &BadRequest{}):
+		kumaErr = &types.Error{
+			Status: 400,
+			Title:  title,
+			Detail: err.Error(),
+		}
```
lahabana, line 47: Should you should have this another condition in the previous case?

lukidzi: Thought about this bad then we won't have title configurable if we are ok with this

jakubdyszkiewicz (approved):

pkg/core/rest/errors/error_handler.go
```diff
@@ -33,7 +33,7 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 			Title:  title,
 			Detail: "Not found",
 		}
-	case errors.Is(err, &rest.InvalidResourceError{}) || errors.Is(err, &registry.InvalidResourceTypeError{}) || errors.Is(err, &store.PreconditionError{}):
+	case errors.Is(err, &rest.InvalidResourceError{}) || errors.Is(err, &registry.InvalidResourceTypeError{}) || errors.Is(err, &store.PreconditionError{}) || errors.Is(err, &BadRequest{}):
```
jakubdyszkiewicz, line 36: btw. I think you could do
```
case errors.Is(err, &rest.InvalidResourceError{}):
case errors.Is(err, &registry.InvalidResourceTypeError{}):
  ...common logic for all
```

lukidzi: Tried but it doesn't work

jakubdyszkiewicz: interesting, ok

PR #8319, opened 2023-11-10
fix(kuma-cni): support port exclusion for UIDs

lobkovilya:
Kuma Init container supports the following pod annotations:
* `traffic.kuma.io/exclude-outbound-tcp-ports-for-uids`
* `traffic.kuma.io/exclude-outbound-udp-ports-for-uids`

but Kuma CNI doesn't. Current PR adds support for these annotations for Kuma CNI. 

[Ideally](https://github.com/kumahq/kuma/issues/8320), we'd like to converge Kuma CNI and Transparent Proxy code bases to share parsing and config structures.

### Checklist prior to review

### DIFF
--- a/test/e2e/cni/exclude_outbound_port.go
+++ b/test/e2e/cni/exclude_outbound_port.go
@@ -68,8 +68,7 @@ func ExcludeOutboundPort() {
 			testserver.WithMesh(meshName),
 			testserver.WithNamespace(namespace),
 			testserver.WithPodAnnotations(map[string]string{
-				metadata.KumaTrafficExcludeOutboundTCPPortsForUIDs: "80:1234",
-				metadata.KumaTrafficExcludeOutboundUDPPortsForUIDs: "53:1234",
+				metadata.KumaTrafficExcludeOutboundPortsForUIDs: "tcp:80:1234;udp:53:1234",
 			}),
 			testserver.AddInitContainer(corev1.Container{
 				Name:            "init-test-server",
--- a/test/e2e/cni/exclude_outbound_port.go
+++ b/test/e2e/cni/exclude_outbound_port.go
@@ -20,7 +20,6 @@ import (
 
 func ExcludeOutboundPort() {
 	meshName := "exclude-outbound-port"
-	nodeName := fmt.Sprintf("second-%s", strings.ToLower(random.UniqueId()))
 
 	namespace := "exclude-outbound-port"
 	namespaceExternal := "exclude-outbound-port-external"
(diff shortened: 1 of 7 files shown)


bartsmykla (approved):

test/e2e/cni/exclude_outbound_port.go
```diff
@@ -0,0 +1,91 @@
...
+		Expect(cluster.DismissCluster()).To(Succeed())
+	})
+
+	It("should be able to use network from init container if we ignore ports for uid", func() {
+		Expect(cluster.Install(testserver.Install(
+			testserver.WithName("test-server"),
+			testserver.WithMesh(meshName),
+			testserver.WithNamespace(namespace),
+			testserver.WithPodAnnotations(map[string]string{
+				metadata.KumaTrafficExcludeOutboundPortsForUIDs: "tcp:80:1234;udp:53:1234",
```
slonka, line 71: we need 53 for DNS right?

slonka (approved):
looks good, two things to consider:
1. have you checked if there are other annotations that are missing parsing and only works in init-containers?
2. should we add a comment in init-container and cni that until https://github.com/kumahq/kuma/issues/8324 is resolved new annotations need to be added in both places?

lobkovilya:
1. I looked through the list of annotations, `kuma.io/sidecar-uid` is missing from CNI. But I assume this is by design since we always [set UID to default](https://github.com/kumahq/kuma/blob/bad1ab15ffeff562c7c9d46a9d51202ec5dc2ce7/app/cni/pkg/cni/annotations.go#L112).
2. Good idea, let me add a comment

test/e2e/cni/exclude_outbound_port.go
lobkovilya: yeah, I actually tried without it and it can't resolve `test-server.exclude-outbound-port-external.svc.cluster.local`

slonka:
> I looked through the list of annotations, kuma.io/sidecar-uid is missing from CNI. But I assume this is by design since we always [set UID to default](https://github.com/kumahq/kuma/blob/bad1ab15ffeff562c7c9d46a9d51202ec5dc2ce7/app/cni/pkg/cni/annotations.go#L112).

@bartsmykla this should be configurable as well, right?

bartsmykla:
I think it should be configurable as current approach of always setting the default UID is incorrect when we would change via `kuma.io/sidecar-uid` annotation on pod.

slonka:
Only 

```
1 test failed out of 144
[It] Advanced LocalityAwareness with MeshLoadBalancingStrategy and Enabled Egress should route based on defined strategy with egress enabled
```

failed so I'm merging this since improvement for it is already in the branch

PR #8325, opened 2023-11-13
chore(kuma-cp): change logger name in HandleError from error to rest

bartsmykla:
Having logger with a name "error" is confusing and incorrect in my opinion. I believe "rest" is more appropriate and it allows to filter these logs with the logger name if necessary.

### Checklist prior to review

  - no relevant issues
  - it won't
  - no tests
  - there is no need
  - there is no need

> Changelog: skip


slonka (approved):
LGTM, one Q: should we add any additional api-related fields there?

bartsmykla:
I believe we could allow to customize logger, but as a following up PR as I believe it's out of the scope of this fix

PR #8376, opened 2023-11-15
ci(actions): made build/test/check run in parallel in build-test-distribute workflow

jijiechen:
### Checklist prior to review

  - N/A
  - Confirmed.
  - No need.
  - [x] Don't forget `ci/` labels to run additional/fewer tests
  - No need.
  - labeled

This PR makes these jobs run in parallel to speed the overall run:
- build
- test
- check

A successful run can be found here: https://github.com/jijiechen/kuma/actions/runs/6878015850

This PR also includes some other related minor improvements.

### DIFF
--- a/.github/workflows/build-test-distribute.yaml
+++ b/.github/workflows/build-test-distribute.yaml
@@ -56,6 +56,7 @@ jobs:
       - run: |
           make clean
       - run: |
+          make generate
           make build
       - run: |
           make -j build/distributions
@@ -85,7 +86,7 @@ jobs:
         run: |
           make test/container-structure
   test:
-    needs: ["check", "build"]
+    needs: ["check"]
     runs-on: ubuntu-latest
     if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ci/skip-test') }}
     steps:
@@ -118,7 +119,7 @@ jobs:
           path: build/reports
           retention-days: 30
   distributions:
-    needs: test
+    needs: ["build", "test"]
     if: ${{ always() && !failure() && !cancelled() }}
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/build-test-distribute.yaml
+++ b/.github/workflows/build-test-distribute.yaml
@@ -30,15 +30,29 @@ jobs:
           make check
   build:
     runs-on: ubuntu-latest
-    needs: ["check"]
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: "Check to run on all arch/os combinations"
-        if: ${{ contains(github.event.pull_request.labels.*.name, 'ci/run-full-matrix') }}
+      - name: "Check if run on releasable ref"
+        id: "releasable_ref_checker"
+        if: |
+          ${{
+            (github.ref == 'refs/branch/master') ||
+            startsWith(github.ref, 'refs/branch/release-')  ||
+            startsWith(github.ref, 'refs/tags/') ||
+          false }}
         run: |
-          echo 'ENABLED_GOARCHES="arm64 amd64" ENABLED_GOOSES="linux darwin"' >> $GITHUB_ENV
+          echo "releasable=true" >> $GITHUB_OUTPUT
+      - name: "Check if should run on all arch/os combinations"
+        if: |
+          ${{
+            contains(github.event.pull_request.labels.*.name, 'ci/run-full-matrix') ||
+            'true' == steps.releasable_ref_checker.outputs.releasable
+          }}
+        run: |
+          echo 'ENABLED_GOARCHES=arm64 amd64' >> $GITHUB_ENV
+          echo 'ENABLED_GOOSES=linux darwin' >> $GITHUB_ENV
       - uses: actions/setup-go@v4
         with:
           go-version-file: go.mod
@@ -54,9 +68,6 @@ jobs:
       - run: |
           make dev/tools
       - run: |
-          make clean
-      - run: |
-          make generate
           make build
       - run: |
           make -j build/distributions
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-distribute.yaml
```diff
@@ -30,6 +28,42 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: "Check if run on releasable ref"
+        id: "releasable_ref_checker"
+        if: |
+          (github.ref == 'refs/branch/master') ||
+          startsWith(github.ref, 'refs/branch/release-')  ||
+          startsWith(github.ref, 'refs/tags/') ||
+          false
```
lahabana, line 43: would `github.event_name == "push"` be easier?  It also solves any problem with changing releasable branches in the future.

I think you can just have the 2 steps merged together:

```
if: github.event_name == "push" || contains(github.event.pull_request.labels.*.name, 'ci/run-full-matrix')
```

WDYT?

jijiechen: Yes, it is better. In this way, we narrow down the maintainance to one place

.github/workflows/build-test-distribute.yaml
```diff
@@ -102,23 +128,40 @@ jobs:
...
+            ;;
+            x86_64)
+            RUNNER_ARCH=amd64
+            ;;
+            *)
+            echo "Unsupported arch '$ARCH'"
+            exit 1
+            ;;
+          esac
+          echo "RUNNER_ARCH=$RUNNER_ARCH" >> $GITHUB_ENV
```
slonka, line 156: this is duplicated, can we dry it up?

.github/workflows/build-test-distribute.yaml
```diff
@@ -54,9 +96,8 @@ jobs:
       - name: Run container structure test
         if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci/skip-container-structure-test') && !contains(github.event.pull_request.labels.*.name, 'ci/skip-test') }}
         run: |
-          make test/container-structure
+          ENABLED_GOARCHES=$RUNNER_ARCH make test/container-structure
```
slonka, line 99: why is this test not under `test` job? or even a separate job just like in `circleci`

.github/workflows/build-test-distribute.yaml
```diff
@@ -75,21 +116,6 @@ jobs:
             ${{ runner.os }}-devtools
       - name: Free up disk space for the Runner
         run: |
-          # source: https://github.com/apache/flink/blob/master/tools/azure-pipelines/free_disk_space.sh
```
slonka, line 78: is it no longer needed? if so, why?

.github/workflows/build-test-distribute.yaml
```diff
@@ -102,23 +128,40 @@ jobs:
...
+            *)
+            echo "Unsupported arch '$ARCH'"
+            exit 1
+            ;;
+          esac
+          echo "RUNNER_ARCH=$RUNNER_ARCH" >> $GITHUB_ENV
       - name: "Check if force push"
+        if: |
+          contains(github.event.pull_request.labels.*.name, 'ci/force-publish') ||
+          false
```
slonka, line 160: isn't this `|| false` a noop?

.github/workflows/build-test-distribute.yaml
```diff
@@ -30,22 +28,66 @@ jobs:
...
+          ENABLED_GOARCHES=$RUNNER_ARCH make -j docker/save
+      - name: Cleanup redundant build outputs
         run: |
-          tar -czf build.tar.gz ./build
-      - name: Temporarily saving build output
+          find  ./build/artifacts-* -maxdepth 0 -type d | xargs -I % rm -rf %
+          find ./build/distributions/* -maxdepth 0 -type d | grep -v '/out' | xargs -I % rm -rf %
+          find  ./build/tools-* -maxdepth 0 -type d | xargs -I % rm -rf %
+          rm -rf ./build/oapitmp
+          rm -rf ./build/ebpf/
```
slonka, line 85: Do we have a mirror of this in circleci?

slonka (commented):
Left some questions, will do another round when answered.

.github/workflows/build-test-distribute.yaml
jijiechen: Unfortunately, they don't support setting environment variables across extracted actions.
This script will be removed because it's not needed anymore.

.github/workflows/build-test-distribute.yaml
jijiechen: Yes, it's a noop. With this noop, a maintainer may directly duplicate the whole line and don't need to worry about removing the `||` at the end.  
There are certain scenarios at other positions in the same file. (And I left a comment at here that we will bring back some commected conditions in the near future)

.github/workflows/build-test-distribute.yaml
jijiechen: CircleCI has a builtin function and it lists all items needed to persist. This is a replacement to that.

.github/workflows/build-test-distribute.yaml
jijiechen: Yes, it's no longer needed. I'll remove it. Thanks.

.github/workflows/build-test-distribute.yaml
jijiechen: I'll remove it anyway.

.github/workflows/build-test-distribute.yaml
jijiechen: Because only container images related steps will fail and other steps actual work across different arches.

jijiechen: Charly suggested a way to support cross builds for those container related steps. So this is to be removed.

lahabana (approved):

PR #8388, opened 2023-11-16
replace errors.Errorf by fmt.Errorf

mmorel-35:
### Checklist prior to review

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -39,7 +39,7 @@ run:
 linters-settings:
   forbidigo:
     analyze-types: true
-    forbidigo:
+    forbid:
       - p: ^errors.Errorf$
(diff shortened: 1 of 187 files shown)


lahabana (changes requested):
While I agree we should part from `github.com/pkg/errors` I think the presence of the stacktrace in these errors is nice and we should think more methodically about a replacement: https://github.com/pkg/errors/issues/245

I've opened: https://github.com/kumahq/kuma/issues/8390 to think a little more about this. I think it may be useful to open a MADR to decide when we should have stacktraces and when we shouldn't

jakubdyszkiewicz:
> Triage: Thanks for the contrib, but we decided to stay with pkg/errors

PR #8402, opened 2023-11-16
feat(MeshHTTPRoute): basic MeshGateway support

michaelbeaumont:
### Checklist prior to review


pkg/xds/envoy/tags/match.go
```diff
@@ -108,6 +109,28 @@ func (t Tags) String() string {
 	return strings.Join(pairs, ",")
 }
 
+func TagsFromTargetRef(targetRef common_api.TargetRef) (Tags, bool) {
```
lobkovilya, line 112: If the package is already called `tags`, probably makes sense to rename the function to `tags.FromTargetRef`

lobkovilya (commented):
MeshHTTPRoute doesn't support `kind: MeshGateway` in the top-level targetRef, right? I'm a bit lost, why some policies like MeshTimeout/MeshRetry have `kind: MeshGateway` but MeshHTTPRoute uses just `kind: MeshService` to target MeshGateway?

lobkovilya:
Is it going to work if MeshTimeout targets MeshHTTPRoute and MeshHTTPRoute targets MeshGateway?

lobkovilya (approved):
LGTM

PR #8404, opened 2023-11-17
test(policy): introduce new resources builders for tests

Automaat:
### Checklist prior to review

### DIFF
--- a/pkg/test/xds/builders/proxy_builder.go
+++ b/pkg/test/xds/builders/proxy_builder.go
@@ -64,3 +64,8 @@ func (p *ProxyBuilder) WithRouting(routing xds.Routing) *ProxyBuilder {
 	p.res.Routing = routing
 	return p
 }
+
+func (p *ProxyBuilder) WithRoutingBuilder(routing *RoutingBuilder) *ProxyBuilder {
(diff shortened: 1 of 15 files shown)


pkg/test/xds/builders/proxy_builder.go
```diff
@@ -64,3 +64,8 @@ func (p *ProxyBuilder) WithRouting(routing xds.Routing) *ProxyBuilder {
 	p.res.Routing = routing
 	return p
 }
+
+func (p *ProxyBuilder) WithRoutingBuilder(routing *RoutingBuilder) *ProxyBuilder {
+	p.res.Routing = *routing.Build()
+	return p
+}
```
bartsmykla, line 71: Nit: In other places you don't have methods for both `Foo` and `FooBuilder` and instead `Foo` takes `FooBuilder` as an argument instead of `Foo`. I would try to be consistent here.

bartsmykla (approved):
One nit but LGTM overall

pkg/test/xds/builders/proxy_builder.go
Automaat: I forgot to changer it, thank you for pointing that out

PR #8410, opened 2023-11-17
chore(kuma-cp): drop deprecated regex EngineType

michaelbeaumont:
### Checklist prior to review


test/e2e_env/universal/compatibility/dp_compatibility.go
```diff
@@ -19,9 +19,9 @@ func UniversalCompatibility() {
 			Install(Kuma(core.Zone)).
 			Install(TestServerUniversal("test-server", "default",
 				WithArgs([]string{"echo", "--instance", "universal1"}),
-				WithDPVersion("1.5.0"))).
+				WithDPVersion("2.5.0"))).
 			Install(DemoClientUniversal(AppModeDemoClient, "default",
-				WithDPVersion("1.5.0"),
+				WithDPVersion("2.5.0"),
```
lahabana, line 24: Ideally we should use `versions.yml` to pick something that's 2 version behind us so that it always stays up to date.

michaelbeaumont: Yeah that's definitely true

test/framework/universal_app.go
```diff
@@ -376,7 +376,7 @@ func (s *UniversalApp) OverrideDpVersion(version string) error {
 	// It is important to store installation package in /tmp/kuma/, not /tmp/ otherwise root was taking over /tmp/ and Kuma DP could not store /tmp files
 	err := ssh.NewApp(s.containerName, "", s.verbose, s.ports[sshPort], nil, []string{
 		"wget",
-		fmt.Sprintf("https://download.konghq.com/mesh-alpine/kuma-%s-ubuntu-amd64.tar.gz", version),
+		fmt.Sprintf("https://download.konghq.com/kuma-binaries-release/kuma-%s-linux-amd64.tar.gz", version),
```
lahabana, line 379: actually I think this should be different (cloudsmith address)

michaelbeaumont: Done

PR #8424, opened 2023-11-21
ci(actions): generate e2e tests matrix

jijiechen:
### Checklist prior to review

  - N/A 
  - Confirmed
  - Mannually tested
    - Added
  - No need
  - No need

This PR adds the ability to generate matrix for running E2E tests on GitHub Actions.

No E2E tests will be run at this time.


jijiechen:
An example can be found here: https://github.com/jijiechen/kuma/actions/runs/6938484757
The `ci/skip-*` and `ci/run-full-matrix` labels are also supported.

.github/actions/e2e/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+    default: amd64
+    required: false
+  cniNetworkPlugin:
+    description: The CNI networking plugin to use [flannel | calico]
+    default: flannel
+    required: false
+  deltaKDS:
+    description: if should run tests with new implementation of KDS
+    default: 'false'
+    required: false
```
lahabana, line 27: I think we can likely drop this option in the matrix

.github/workflows/build-test-distribute.yaml
```diff
@@ -170,3 +173,153 @@ jobs:
...
+          # non priority main e2e
+          EXCLUDE_MAIN_NON_PRIORITY='[{"k8sVersion":"kindIpv6"}, {"k8sVersion": "${{ env.K8S_MIN_VERSION }}"}]'
+
+          # Default matrix
+          SLOW=$(echo '{}' | $JQ_PATH -rc ".exclude += $EXCLUDE_ARM64" | $JQ_PATH -rc ". += $ALL_K8S_VERSION | . += $ALL_ARCH")
+          MAIN=$(echo '{}' | $JQ_PATH -rc ".exclude += $EXCLUDE_ARM64" | $JQ_PATH -rc ". += $ALL_K8S_VERSION | . += $ALL_ARCH | . += $ALL_TARGET" | $JQ_PATH -rc ".exclude += $EXCLUDE_KIND | .exclude += $EXCLUDE_UNIVERSAL" )
+          DELTA_KDS=$(echo '{}' | $JQ_PATH -rc ".exclude += $EXCLUDE_ARM64" | $JQ_PATH -rc ". += $ALL_DELTA_KDS | . += $SINGLE_ENV")
+          CNI=$(echo '{}' | $JQ_PATH -rc ".exclude += $EXCLUDE_ARM64" | $JQ_PATH -rc ". += $ALL_CNI | . += $SINGLE_ENV")
+
+          RUN_FULL_MATRIX="${{ contains(github.event.pull_request.labels.*.name, 'ci/run-full-matrix') }}"
```
lahabana, line 228: The full matrix should also run on `master` and `release-*` branches

.github/workflows/build-test-distribute.yaml
```diff
@@ -170,3 +173,153 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/e2e
+        with:
+          k8sVersion: ${{ matrix.k8sVersion }}
+          arch: ${{ matrix.arch }}
+          parallelism: 3
+          target: ""
+  e2e_main:
```
lahabana, line 262:
```suggestion
  test/e2e_env:
```
This way it's the same as the code :)

.github/workflows/build-test-distribute.yaml
```diff
@@ -170,3 +173,153 @@ jobs:
...
+            DELTA_KDS=''    # deltaKDS=="true"
+            CNI=''          # cniNetworkPlugin=="calico"
+            MAIN=$(echo $MAIN | $JQ_PATH -rc ".exclude += $EXCLUDE_ARM64 | .exclude += $EXCLUDE_MAIN_NON_PRIORITY")
+          fi
+
+          echo "slow=$SLOW" >> $GITHUB_OUTPUT
+          echo "main=$MAIN" >> $GITHUB_OUTPUT
+          echo "delta_kds=$DELTA_KDS" >> $GITHUB_OUTPUT
+          echo "cni=$CNI" >> $GITHUB_OUTPUT
+  e2e_slow:
```
lahabana, line 242:
```suggestion
  test/e2e:
```

THis way way it's the same as the code folder

lahabana: (don't know if `/` is valid in keys though)

.github/workflows/build-test-distribute.yaml
```diff
@@ -170,3 +173,153 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/e2e
+        with:
+          k8sVersion: ${{ matrix.k8sVersion }}
+          target: ${{ matrix.arch }}
+          arch: ${{ matrix.arch }}
+          deltaKDS: ${{ matrix.deltaKDS }}
+  e2e_calico:
```
lahabana, line 304: as mentioned let's remove the delta job.

.github/workflows/build-test-distribute.yaml
```diff
@@ -170,3 +173,153 @@ jobs:
...
+        run: |
+          sudo apt-get install -y jq
+      - id: generate-matrix
+        name: Generate matrix
+        run: |
+          SKIP_CI_LABEL="${{ contains(github.event.pull_request.labels.*.name, 'ci/skip-test') || contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e-test')  }}"
+          if [[ "$SKIP_CI_LABEL" == "true" ]]; then
+            echo "Skipping e2e jobs because the pull request contained a 'ci/skip-*' label."
+            exit 0
+          fi
```
lahabana, line 195: I think this would be nicer as an `if` at the github job level like we do for the rest (also speeds things up)

.github/workflows/build-test-distribute.yaml
```diff
@@ -170,3 +173,153 @@ jobs:
...
+
+          JQ_PATH=$(which jq)
+          if [ "$JQ_PATH" == "" ]; then
+            echo "Could not find devtool 'jq', please check if 'check' job has run correctly prior to this job."
+            exit 1
+          fi
```
lahabana, line 201: Would it be more readable to have the full matrix and then a bunch of exclude? Something like:

```
BASE_MATRIX=<<<EOF
{
   "target": ["test/e2e", "test/e2e-universal", "test/e2e-kubernetes", "test/e2e-multizone"],
   "arch": ["amd64", "arm64"],
   "paralellism": ["1"],
  "k8sVersion": ["kind", "kindIpv6", "${{ env.K8S_MIN_VERSION }}", "${{ env.K8S_MAX_VERSION }}"],
  // exclude all the invalid options
  "include":[
    {"target": "test/e2e", "paralellism": "4"},
    {"target": "test/e2e-kubernetes", "cniNetworkPlugin": "calico"},
  ],
  "exclude": [
    {"target": "test/e2e", "paralellism": "1"},
    {"target": "test/e2e-universal",  "${{ env.K8S_MIN_VERSION }}"},
    {"target": "test/e2e-universal", "${{ env.K8S_MAX_VERSION }}"}
  ]
}
EOF

EXCLUDE_OPTIONS=<<<EOF
[
   {"arch": "arm64},
   {"k8sVersion": "kindIpv6"},
   {"k8sVersion": "${{ env.K8S_MIN_VERSION}}"},
   {"cniNetworkPlugin": "calico"}
]
EOF

if [[ "$RUN_FULL_MATRIX" == "false" ]]; then
   BASE_MATRIX=$JQ_PATH -rc ".exclude += '${EXCLUDE_OPTIONS}'"
fi
```

There are probably mistake in the above but that's the idea.
I think this would help on readability and make this matrix more maintainable in the future WDYT?

.github/actions/e2e/action.yaml
jijiechen: What about just let it be here for now, since we are not actually running anything.
And I wrote the wrong name here, shoudl be `legacyKDS`, I'll correct it.

.github/workflows/build-test-distribute.yaml
jijiechen: Yes, just tried and the slash `/` is not allowed. Maybe it's because it will be used as a field name in some scenarios.

.github/actions/e2e/action.yaml
jijiechen: Do you want to remove the `legacyKDS` stuff? If yes, let me do it in the second PR.

.github/workflows/build-test-distribute.yaml
jijiechen: See the previous conversation.

.github/workflows/build-test-distribute.yaml
jijiechen: Sure thing, will do

.github/workflows/build-test-distribute.yaml
jijiechen: So I'll reuse the "push" event here

.github/workflows/build-test-distribute.yaml
jijiechen: Let's discuss on the call later and change accordingly in the second PR. 
Will that work?

.github/workflows/build-test-distribute.yaml
lahabana: Maybe just test_e2e and test_e2e_env

.github/workflows/build-test-distribute.yaml
```diff
@@ -170,3 +173,161 @@ jobs:
...
+            # non priority main e2e
+            EXCLUDE_ARM64='[{"arch": "arm64"}]'
+            EXCLUDE_MAIN_NON_PRIORITY='[{"k8sVersion":"kindIpv6"}, {"k8sVersion": "${{ env.K8S_MIN_VERSION }}"}]'
+            MAIN=$(echo $MAIN | jq -rc ".exclude += $EXCLUDE_ARM64 | .exclude += $EXCLUDE_MAIN_NON_PRIORITY")
+          fi
+
+          echo "slow=$SLOW" >> $GITHUB_OUTPUT
+          echo "main=$MAIN" >> $GITHUB_OUTPUT
+          echo "legacy_kds=$LEGACY_KDS" >> $GITHUB_OUTPUT
+          echo "cni=$CNI" >> $GITHUB_OUTPUT
```
lahabana, line 233: only 2 outputs to this job:

test_e2e  is `slow`
test_e2e_env is a merge of `main`, `cni` and `legacy_kds` do not hesitate to use `include` on your matrix to make this easier.

Also has the 2 matrices (test_e2e and test_e2e_env) are very different maybe just have 2 `BASE_MATRIX`.

lahabana: Other option have 4 different matrices which map make targets:

- test/e2e
- test/e2e-universal
- test/e2e-multizone
- test/e2e-kubernetes

Then each have their own matrix.

jijiechen: Updated code with the first approach.

jijiechen:
I've just updated code with the first approach of merging matrix generating.
Please have a look at your convenience. @lahabana

lahabana (approved):
LGTM This looks great and good to go I think! Do you have an example run of this?

jijiechen:
Yes, here are two:
* full matrix run: https://github.com/jijiechen/kuma/actions/runs/6958065668
* normal run: https://github.com/jijiechen/kuma/actions/runs/6958089045

In these examples, I took out dependencies from `e2e` to `build`, so the graphic could be slightly different than expected. The matrix generation was working the expected way.

PR #8426, opened 2023-11-21
ci(actions): trigger E2E tests on CircleCI manually from GitHub Actions

jijiechen:
### Checklist prior to review

  - N/A
  - Confirmed
  - Manually tested
    - Added
  - No need
  - No need

This PR adds the ability to trigger E2E tests on CircilCI manually from GitHub Actions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,27 @@ parameters:
   ubuntu_image:
     type: string
     default: "ubuntu-2204:2022.10.2"
+  manual_e2e:
+    type: boolean
+    default: false
+  e2e_param_k8s_version:
+    type: string
+    default: ""
+  e2e_param_arch:
+    type: string
+    default: ""
+  e2e_param_parallelism:
+    type: integer
+    default: 1
+  e2e_param_target:
+    type: string
+    default: ""
+  e2e_param_cni_network_plugin:
+    type: string
+    default: flannel
+  e2e_param_legacy_kds:
+    type: boolean
+    default: false
 # See https://circleci.com/docs/2.0/configuration-reference/#commands-requires-version-21.
 commands:
   install_build_tools:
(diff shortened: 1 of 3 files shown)


jijiechen:
I'll update the PR later based on the suggestions on the previous PR at https://github.com/kumahq/kuma/pull/8424
You may review the content of this PR first. @lahabana

.github/actions/e2e/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+        if [ "${{ runner.debug }}" == "1" ]; then
+          VERBOSE='-v'
+        fi
+        
+        curl $VERBOSE --fail -X POST $CIRCLE_CI_API_PATH \
+          --header "content-type: application/json" \
+          --header "x-attribution-login: ${{ github.actor }}" \
+          --header "x-attribution-actor-id: ${{ github.actor }}" \
+          --header "Circle-Token: ${{ inputs.secureCircleCIToken }}" \
+          --data "$BODY"
```
lahabana, line 76: The reason we don't use: https://github.com/CircleCI-Public/trigger-circleci-pipeline-action is that we can't pass params? I guess the problem with this is that we don't block until completion no?

.circleci/config.yml
```diff
@@ -517,3 +538,20 @@ workflows:
...
           requires: [test, container-structure]
+  manual-e2e:
+    when:
+      equal: [true, << pipeline.parameters.manual_e2e >>]
+    jobs:
+      - go_cache:
+          name: go_cache-<< pipeline.parameters.e2e_param_arch >>
+          arch: << pipeline.parameters.e2e_param_arch >>
+      - build:
+          name: build
```
lahabana, line 549: I think we should be downloading the github artifact in our e2e test as we already have everything ready right?

.github/actions/e2e/action.yaml
lahabana: Maybe instead of having parameters to the circleci job we should just read https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/tree/main#gha_meta  in the circleCI job  ?s

jijiechen: Right, we can't pass custom parameters using that action. That's why I call it manually. The `GHA_Meta` provides a way to pass custom parameters, we have to composite all params into one in this way, and we don't have a easy way to extract these params as part of the CircleCI workflow (using their `setup` job could be a way, that might requires significant changes to the whole strcture).

In terms of blocking untile completion, I believe that action does not provide this ability either. I read through the [source code](https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/blob/main/src/lib/CircleCIPipelineTrigger.ts#L82), and checkout of their [readme](https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/tree/main#outputs), they are calling the same API with the one in my code and they only provide state of the just-created pipeline as an output as well.

For blocking until complete, we may laverage their other APIs to  all workflows and periodically check latest state. Doc at https://circleci.com/docs/api/v2/index.html#operation/getWorkflowById

.circleci/config.yml
jijiechen: Yes, that'll save some time.

Issue #4318, opened 2022-05-20
Applying policies on the Zone CP


jakubdyszkiewicz:
### Description

It's more of a brainstorming issue.

Right now, all the policies are applied on Global CP and are synced to all Zone CPs. It's an easy 0-conflict model. However, it has a challenge. If the multicluster scenario in your org has a topology of cluster = team. The team needs to apply policies on the Global CP which means they need to have an access to this cluster including proper RBAC.

There are other advantages to being able to apply policies on Zone CP:
* better HA, if Global CP is down, you can manipulate policies. (technically possible today if you bypass webhook)
* tools that want to manage policies automatically like Flagger

The alternative would be to be able to apply policies on the Zone CP by specific teams. This creates multiple problems to solve.

### How to provide Mesh-wide policies?
Policies applied on Zone CP should only be executed in the given zone. How then apply policies that are applicable to all zones.
If we constraint the topology to "mesh can live in one zone", then it's not really a problem. But assuming this topology, why not deploy standalone Kuma in all clusters.

### Can policies be applied to Zone CP AND Global CP?
Even if policies are applied on Zone CP by teams, Mesh Operator should still have the last word when it comes to applying the setting.
One option is that Mesh Operator has an access to Zone CP and can modify policy, which solves the problem.
The second is that Mesh Operator can still apply policies on Global, which brings two-way sync problem.

### Two-way sync
The current way of managing resources is elegant, there is one source of truth. If we were to sync resources from Zone CP to Global CP, how do we handle it?
Policies applied on Zone CP should be synced to Global CP for visibility (not synced to other zones).
We could restrict those policies to not being managed from the Global CP (which eliminates conflicts problems etc.)

Another option would be to allow policies to be applied on both Global and Zone CP, but only managed from the CP it was applied on.
For example, teams may want to create policies that are relevant to their clusters and those are synced to Global CP for visibility. However, Mesh Operator can apply policies on Global CP that are synced to all the zones, but teams cannot manage them.

lahabana:
> Policies applied on Zone CP should only be executed in the given zone. How then apply policies that are applicable to all zones.

Can't this be left to the user by adding the zone in the selector? The policy would be synced everywhere but only apply to DPPs in the zone they want.

My idea of this was to add a "controllingZone" to the mesh. Something like:

```
type: Mesh
name: foo
conf:
   controllingZone: "zone-1"
```

[...]

PR #8427, opened 2023-11-21
docs(MADR): zone to global policy sync

lobkovilya:
### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- https://github.com/kumahq/kuma/issues/4318

### DIFF
--- a/docs/madr/decisions/032-zone-to-global-policy-sync.md
+++ b/docs/madr/decisions/032-zone-to-global-policy-sync.md
@@ -95,7 +95,7 @@ During the resource creation Zone CP will append `kuma.io/zone` tag to MeshGatew
 
 Syncing Zone originated policies to Global is needed only for visibility in GUI.
 
-During the implementation it's important to pay attention to policies not be synced from Global to Zone if they're
+During the implementation it's important to pay attention to not sync policies from Global to Zone if they're
 originated in Zone. Naming is also crucial, for policy `policy-1` from `us-east` we'll create a policy named
 `policy-1-{{hash-suffix}}` in Global with metadata `zone: us-east`. Metadata should be implemented as `labels` when
 Global on k8s and as an additional field in schema on Universal. For more information see https://github.com/kumahq/kuma/pull/7484.
--- a/docs/madr/decisions/032-zone-to-global-policy-sync.md
+++ b/docs/madr/decisions/032-zone-to-global-policy-sync.md
@@ -55,7 +55,7 @@ these policies to Global. For example, when user creates a policy in `us-east` z
 type: MeshTimeout
 mesh: mesh-1
 name: t-1
-labels: # not implemented today, requires Postgres schema 
+labels: # not implemented today, requires Postgres schema change
   kuma.io/managed-by: zone
 spec:
   targetRef:
--- a/docs/madr/decisions/032-zone-to-global-policy-sync.md
+++ b/docs/madr/decisions/032-zone-to-global-policy-sync.md
@@ -1,0 +1,185 @@
+# Zone to Global policy sync
+
+* Status: accepted
+
+Technical Story: none
+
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/032-zone-to-global-policy-sync.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+This conversion is visible inside the Zone store (so it happens **not** in Zone CP memory).
+If users provide a policy with a correct `kuma.io/zone` tag then no additional conversion is required, but we have to
+validate that zone tag is matching the zone's name. The actual security check is happening on Global but validation is
+helpful to prevent unexpected behaviour.
+
+If users apply a policy on a zone without a `kuma.io/managed-by: zone` label then Zone CP returns an error saying
+"if you want to apply a policy on the zone please add `kuma.io/managed-by: zone` label".
+This measure is going to prevent users form unintentionally applying policies at the wrong place.
```
lahabana, line 91: Is the label explicitly added by the user or it's automatically added by a webhook?

jakubdyszkiewicz (approved):

docs/madr/decisions/032-zone-to-global-policy-sync.md
```diff
@@ -0,0 +1,185 @@
...
+
+Besides policies with top level targetRef we also have MeshGateway and MeshGatewayRoute resources.
+During the resource creation Zone CP will append `kuma.io/zone` tag to MeshGateway listeners and MeshGatewayRoute selectors.
+
+Syncing Zone originated policies to Global is needed only for visibility in GUI.
+
+During the implementation it's important to pay attention to policies not be synced from Global to Zone if they're
```
lahabana, line 98:
```suggestion
During the implementation it's important to pay attention to not sync policies from Global to Zone if they're
```

docs/madr/decisions/032-zone-to-global-policy-sync.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+Kuma automatically converts this to:
+
+```yaml
+type: MeshTimeout
+mesh: mesh-1
+name: t-1
+labels: # not implemented today, requires Postgres schema 
+  kuma.io/managed-by: zone
```
lahabana, line 74: should that be the zone name?

slonka (approved):
One nitpick since my last review in the other repo

docs/madr/decisions/032-zone-to-global-policy-sync.md
lobkovilya: The idea here is to prevent users from applying things in the wrong place. They probably got used to the fact Zone CP returns an error if you unintentionally run `kubectl apply` against a zone. By requiring the `kuma.io/managed-by: zone` label we're making sure their desire is intentional. So we're **not** going to automate it with a webhook.

docs/madr/decisions/032-zone-to-global-policy-sync.md
lobkovilya: We don't want to force users to create templates with `{{ zoneName }}` if they need pretty much the same policy for different zones. That was @johnharris85 suggestion, I kind of agree with it.

lobkovilya: Also, we're going to add `kuma.io/zone` tag automatically to the top-level targetRef.

michaelbeaumont (approved):

slonka:
Can there be a conflict? If someone applies the same policy on both global and zone?

lobkovilya:
> Can there be a conflict? If someone applies the same policy on both global and zone?

No, both policies will be ordered according to the targetRef ordering rules we have and merged afterward.

PR #8435, opened 2023-11-21
fix(metrics): unify zone name in metrics for k8s and universal

slonka:
Otherwise we get different "zone" in metrics for global cp

image

vs 

image

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- not reported by issue

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -89,15 +89,14 @@ func (m *metrics) BulkRegister(cs ...prometheus.Collector) error {
 }
 
 func ZoneNameOrMode(mode config_core.CpMode, name string) string {
-	zoneName := ""
 	switch mode {
 	case config_core.Zone:
-		zoneName = name
+		return name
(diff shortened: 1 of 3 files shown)


pkg/metrics/metrics.go
```diff
@@ -85,3 +87,17 @@ func (m *metrics) BulkRegister(cs ...prometheus.Collector) error {
...
+	switch mode {
+	case config_core.Zone:
+		zoneName = name
+	case config_core.Global:
+		zoneName = "Global"
+	case config_core.Standalone:
+		zoneName = "Standalone"
+	}
+
+	return zoneName
```
bartsmykla, line 102:
```suggestion
	switch mode {
	case config_core.Zone:
		return name
	case config_core.Global:
		return "Global"
	case config_core.Standalone:
		return "Standalone"
	default:
		return ""
	}
```

bartsmykla (approved):
1 nit

PR #8439, opened 2023-11-22
chore(kuma-cp): allow to pass more apiWebServiceCustomize fns

bartsmykla:
So far when used `WithAPIWebServiceCustomize` more than once, only the last function was being used and others silently ignored. Now we can customize WebService multiple times

### Checklist prior to review

  - no relevant issue
  - it won't
  - there is no need
  - there is no need

> Changelog: skip


slonka (approved):
remember to open up an issue to add tests

lahabana:
> remember to open up an issue to add tests
Let's not get in such a habit of delaying tests for later. Let's write tests as we write fixes.

PR #8443, opened 2023-11-22
fix(ZoneIngress): subset routing when tag is present on all subsets

michaelbeaumont:
### Summary

[Recently zone proxy generation was refactored](https://github.com/kumahq/kuma/pull/7536#issue-1855312499). There was a change made to reduce the number of LB subsets we create by filtering out tags from `filterMetadata`, described in that PR's description **The second problem**. Reading that it seems this was intended only for ZoneEgress, but the refactor means both kinds of zone proxy do this filtering of tags.
[I've changed the e2e test in this PR](https://github.com/kumahq/kuma/pull/8443/commits/6873671790f6358fd05fd27ec79153f7e9bdacef#diff-fdc423194184cbb8f9cc2564e35aa449d6e3d452a69d6be9532dd415f56fa5a0L150-L158) to trigger this trimming of tags by adding a tag present on all subsets. A mismatch between the TCP proxy's `filterMetadata.envoy.lb` and the `Cluster`s `lbSubsetConfig.subsetSelectors[].keys` occurs which manifests as the subsets being ignored due to `fallbackPolicy: ANY_ENDPOINT`. [The final commit shows the change in config with this fix](https://github.com/kumahq/kuma/pull/8443/commits/dc683a9f99c9fbf43c643fb8b30d8fed5df7e251).

I'm not 100% convinced this can't occur even in the ZoneEgress case. ~In either case, another way to solve this might be the other way around, to be smarter about what we include in `subsetSelectors[].keys` but it's not clear to me how to do this in general since there can be multiple uses of the same `Cluster` with different `filterMetadata`.~ 
This PR makes it so that we're smarter about what we include in `subsetSelectors[].keys`.

### Checklist prior to review


jakubdyszkiewicz:
It took me a while, but I think it can be triggered for egress as well.

Let's say we have this
Ingress1 service:a, version:1, x:y
Ingress2 service:a, version:1, x:y
Ingress3 service:a, version:2, x:y

and we want to do the split to version:1,x:y. This way we mark `x` as irrelevant and it will go into this `fallbackPolicy: ANY_ENDPOINT` potentially routing to Ingress3.

I agree that _The second problem_ affects egress, because ingress won't have multiple endpoints with same ip:port and different metadata.

jakubdyszkiewicz:
Looking more into the code.
I think what could work is to return list of created clusters by `envoy_common.NewCluster` from `AddFilterChains` and then use it in `GenerateCDS` instead of relying on `services` and `destinationsPerService`

pkg/xds/generator/egress/testdata/04.mixed-services-custom-trafficroute.golden.yaml
```diff
@@ -85,68 +85,6 @@ resources:
         '@type': type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
         explicitHttpConfig:
           http2ProtocolOptions: {}
-- name: mesh-1:service-1-zone-2
```
michaelbeaumont, line 88: I believe these are correctly absent, since they aren't referenced by any listeners.

jakubdyszkiewicz (approved):
👍

PR #8449, opened 2023-11-23
docs(MADR): deprecate standalone mode

jakubdyszkiewicz:
### Checklist prior to review

### DIFF
--- a/docs/madr/decisions/033-deprecate-standalone.md
+++ b/docs/madr/decisions/033-deprecate-standalone.md
@@ -64,21 +64,131 @@ We create Zone object and ZoneInsights with statistic of KDS only on Global CP. 
 We want to also create Zone object for Zone CP and create ZoneInsights with KDS statistics from Zone perspective.
 Then we can show on this screen whether the Zone is connected to Global or not.
 
-#### Transitioning from Zone CP to Zone CP with KDS connection
+#### Transition
+
+Transition is described with the assumption that MADR #032 (applying policies on Zone CP) is implemented. 
+
+##### Federated Zone CP to non-federated Zone CP
+
+I tested it and it seems to just work. Nothing extraordinary happens there.
+
+##### Non-federated Zone CP to federated Zone CP 
 
 Let's assume we first start using Kuma with Zone CP without KDS connection.
 
-Then we switch the zone to connect to the Global CP. What will happen is:
-* Policies that are synced from Global CP are replaced on Zone CP
-  It's important to copy policies from Zone CP to Global CP first.
-  However, this will be easier once we support applying policies on Zone CP, because instead of replacing policies from Global, we will just sync existing policies to Global.
-* We block applying policies on Zone CP. This will change once we support apply policies on Zone CP.
-* Secrets that are synced from Global CP are replaced on Zone CP
-  It's important to copy secrets from Zone CP to Global CP first.
-  This includes signing keys and CAs.
-* You can no longer issue zone tokens on Zone CP.
-* You can no longer apply Gateway API resources. All Kuma resources created from GAPI are replaced
-  This is the case until we implement applying policies on Zone CP.
+Assuming that applying policies on Zone CP and syncing them on Global (see 032 MADR) is done, we only need to sync the following resources:
+* Secrets
+* Global Secrets
+* Meshes
 
-Ideally we should create a tool to save relevant policies from Zone CP and apply them to Global CP.
-This however won't be a part of initial implementation of this MADR.
+When we switch the zone to connect to the Global CP. The following will happen:
+* Policies are synced from Global CP. Policies applied on Zone will be synced to Global CP.
+* Secrets that are synced from Global CP are replaced on Zone CP
+* You can no longer issue zone tokens on Zone CP.
+
+**Zone on Kubernetes. Global on Kubernetes**
+
+I managed to successfully federate Zone CP with the following commands.
+```sh
+# change context to zone
+kubectl get mesh -oyaml > /tmp/meshes.yaml
+yq eval 'del(.items[].metadata.resourceVersion, .items[].metadata.ownerReferences, .items[].metadata.uid, .items[].metadata.annotations, .items[].metadata.generation)' /tmp/meshes.yaml > /tmp/meshes-clean.yaml
+kubectl get secrets -n kuma-system --field-selector='type=system.kuma.io/global-secret' -oyaml > /tmp/dump-gsecrets.yaml
+yq eval 'del(.items[].metadata.resourceVersion, .items[].metadata.ownerReferences, .items[].metadata.uid, .items[].metadata.annotations, .items[].metadata.generation)' /tmp/dump-gsecrets.yaml > /tmp/dump-gsecrets-clean.yaml
+kubectl get secrets -n kuma-system --field-selector='type=system.kuma.io/secret' -oyaml > /tmp/dump-secrets.yaml
+yq eval 'del(.items[].metadata.resourceVersion, .items[].metadata.ownerReferences, .items[].metadata.uid, .items[].metadata.annotations, .items[].metadata.generation)' /tmp/dump-secrets.yaml > /tmp/dump-secrets-clean.yaml
+
+# change context to global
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/033-deprecate-standalone.md
```diff
@@ -0,0 +1,84 @@
...
+* You can access GUI
+
+#### Visibility changes
+
+Currently, Zone tab is missing in "standalone" GUI.
+We create Zone object and ZoneInsights with statistic of KDS only on Global CP. Those objects are not synced to Zone CP.
+We want to also create Zone object for Zone CP and create ZoneInsights with KDS statistics from Zone perspective.
+Then we can show on this screen whether the Zone is connected to Global or not.
+
+#### Transitioning from Zone CP to Zone CP with KDS connection
```
lobkovilya, line 67: We should also cover the transition the other way, from Zone CP with KDS to detached Zone CP. This is important if something went wrong during the connection to Global and you want to rollback the change

docs/madr/decisions/033-deprecate-standalone.md
```diff
@@ -0,0 +1,84 @@
...
+kuma-cp run
+```
+HELM and kumactl install adds `KUMA_MODE=standalone` by default, we can also change the default to `zone`.
+
+The current way of running Zone CP will not change.
+
+The current difference of running standalone and zone is providing zone name and KDS address.
+
+Current differences between standalone and zone CP:
+* Cannot apply policies on Zone CP directly.
```
lukidzi, line 42: I would add information that `zone` requires Ingress.

docs/madr/decisions/033-deprecate-standalone.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Pros and Cons of the Options
+
+### Deprecate standalone
+
+If you want to run the "standalone" you can execute the following command
+```
+KUMA_MODE=zone kuma-cp run 
+```
+We will change the default `KUMA_MODE` value to `zone`. So in the end, you would run this the same way as before which is
```
lukidzi, line 31: Do we need to provide default name of the zone, or it isn't required when no kds connection.

docs/madr/decisions/033-deprecate-standalone.md
```diff
@@ -0,0 +1,84 @@
...
+#### Transitioning from Zone CP to Zone CP with KDS connection
+
+Let's assume we first start using Kuma with Zone CP without KDS connection.
+
+Then we switch the zone to connect to the Global CP. What will happen is:
+* Policies that are synced from Global CP are replaced on Zone CP
+  It's important to copy policies from Zone CP to Global CP first.
+  However, this will be easier once we support applying policies on Zone CP, because instead of replacing policies from Global, we will just sync existing policies to Global.
+* We block applying policies on Zone CP. This will change once we support apply policies on Zone CP.
+* Secrets that are synced from Global CP are replaced on Zone CP
```
lukidzi, line 76: How would migration look when you have 2 independent zones and you connect the first one, then you want to connect 2nd one? How would it look like secret migration in this case?

docs/madr/decisions/033-deprecate-standalone.md
```diff
@@ -0,0 +1,84 @@
...
+  It's important to copy secrets from Zone CP to Global CP first.
+  This includes signing keys and CAs.
```
lukidzi, line 78: Could we have something like `KUMA_MULTIZONE_ZONE_NAME=... KUMA_MULTIZONE_ZONE_GLOBAL_ADDRESS=....  ./kuma-cp migrate global` that would avoid the manual step of coping. Naming might be different. But that would require that Global won't have default ones and I think we have an option to not create default policies.

docs/madr/decisions/033-deprecate-standalone.md
slonka: are people running multiple standalone instances? I think when someone thinks about second standalone they would go to multizone so this situation would probably not happen. That's a good question but I think it's more related to syncing from zone to global MADR.

slonka (commented):
I think there are a lot of things that depend on zone to global sync (or could be solved/improved by that) would it make sense to prioritise that? We could take into account what would be useful for this change and only implement standalone deprecation after zone to global sync is done?

docs/madr/decisions/033-deprecate-standalone.md
jakubdyszkiewicz: Zone is only needed for cross zone traffic

docs/madr/decisions/033-deprecate-standalone.md
jakubdyszkiewicz: I think this case is unlikely. But if it's the case then they need to manually set the same secrets on the same zone, then set those secrets to global and then connect zones to global.

jakubdyszkiewicz:
> I think there are a lot of things that depend on zone to global sync (or could be solved/improved by that) would it make sense to prioritise that?

Yes, we are already working on this in parallel in one effort (I'm on this, Ilya is implementing zone -> global sync).

docs/madr/decisions/033-deprecate-standalone.md
```diff
@@ -0,0 +1,84 @@
...
+* GatewayAPI is not supported on Zone CP
+* On Zone CP, Zone Token public key is synced from Global.
+
+With deprecating standalone we still need to differentiate between zone running on its own and zone connected to the Global CP.
+To do this, instead of relying on `KUMA_MODE` `standalone` or `zone`, we will check if KDS address was provided.
+
+Existing standalone and future Zone CP without KDS connection will work the same way.
+
+Changes between existing Zone CP and upcoming Zone CP with KDS connection:
+* You can access GUI
```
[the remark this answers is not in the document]
jakubdyszkiewicz: > does this need to be changed?

It does not, same with Zone + ZoneInsights, but we want to change it for a better UX.

> also will gui work correctly on zone?

It should.

> will it know how to do a config dump / stats / etc?

yes.

docs/madr/decisions/033-deprecate-standalone.md
jakubdyszkiewicz: Ideally, we would require it, but that would require action for existing users. That's why it's required only when we have KDS connection.

docs/madr/decisions/033-deprecate-standalone.md
jakubdyszkiewicz: Yes, we could do something more clever. We could mark Global CP as "provisioning" (in Konnect this would be just checkbox when we create a new Global CP).
Then the first Zone that connects to it, sends secrets and meshes and global CP gets into provisioned (default) state.

The only tricky part is authentication.

docs/madr/decisions/033-deprecate-standalone.md
```diff
@@ -0,0 +1,84 @@
...
+Current differences between standalone and zone CP:
+* Cannot apply policies on Zone CP directly.
+* GUI is disabled on Zone CP
+* Cluster ID is synced from Global on Zone CP. Cluster ID is used for analytics
+* Zone CP connects to Global CP to pull policies
+* Default resources (Mesh, policies etc.) are disabled on Zone CP, because those are pulled from Global.
+* Standalone does not finalize (gc/finalizer) zone ingresses
+* Mesh/Service Insights are not computed on Zone CP
+* GatewayAPI is not supported on Zone CP
+* On Zone CP, Zone Token public key is synced from Global.
```
lahabana, line 50: Should we expand a little what the transition (adding a global to an existing zone) is like here?

docs/madr/decisions/033-deprecate-standalone.md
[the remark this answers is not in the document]
lahabana: There are a few issues on kuma-gui to support this

docs/madr/decisions/033-deprecate-standalone.md
lahabana: I don't think it's that unlikely but @johnharris85 can provide more input. We had cases of customers starting with standalone and later realizing that they should have used multi-zone for easier org wide management.

Do we have secrets that can't be rotated?

lobkovilya (approved):

slonka (approved):

PR #8457, opened 2023-11-27
ci(actions): triggering E2E tests on CircleCI, canceling and fail fast supported

jijiechen:
### Checklist prior to review

  - N/A
  - Confirmed
  - N/A
    - Added
  - No need
  - No need.

This PR triggers E2E test cases on CircleCI while run build/test/dist steps on GitHub actions.
It supports:
1. CI labels (`ci/skip-*` and `ci/run-full-matrix`) supported
2. trigger E2E cases in parallel on CircleCI
3. Track the results of triggered CircleCI workflows and fail GitHub actions run if needed
4. Cancel CircleCI workflows if cancelled on GitHub actions because of user operations or fail fast.


jijiechen:
Example runs:
1. A successfull run: https://github.com/jijiechen/kuma/actions/runs/7002264119
2. A full matrix run: https://github.com/jijiechen/kuma/actions/runs/7002721757 (Failed because of other reasons)
3. A run reflects the correct dependency: https://github.com/jijiechen/kuma/actions/runs/7003299010

.circleci/config.yml
```diff
@@ -355,6 +382,206 @@ jobs:
...
+        type: string
+        default: amd64
+      cniNetworkPlugin:
+        description: The CNI networking plugin to use [flannel | calico]
+        type: string
+        default: flannel
+      legacyKDS:
+        description: if should run tests with new implementation of KDS
+        type: boolean
+        default: false
```
lahabana, line 410: I don't think you should have defaults for these parameters. They should just always be set whenever this is triggered

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,226 @@
...
+        required: false
+      cniNetworkPlugin:
+        description: The CNI networking plugin to use [flannel | calico]
+        type: string
+        default: flannel
+        required: false
+      legacyKDS:
+        description: if should run tests with new implementation of KDS
+        type: string
+        default: 'false'
```
lahabana, line 32: Remove the defaults here too and make required everything. It will be easier to debug when something is wrong (it's only possibly set in 1 place)

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,226 @@
...
+              exit 1
+            fi
+          }
+
+          function check_workflow(){
+            WORKFLOW_ID=$1
+            STATUS=''
+            # status could be "success" "running" "not_run" "failed" "error" "failing" "on_hold" "canceled" "unauthorized"
+            while [[ "$STATUS" == "" ]] || [[ "$STATUS" == "not_run" ]] || [[ "$STATUS" == "running" ]] || [[ "$STATUS" == "on_hold" ]]; do
+              sleep $((RANDOM % 3 + 5))
```
lahabana, line 179: I think you can even go with 30s retry intervals.

lahabana (approved):
some nits but looks good!

jijiechen:
To trigger E2E tests  on CircleCI,  we need to add a GitHub Action secret:
* name: `CIRCLECI_TOKEN`
* value: CircleCI [Personal Access Token](https://app.circleci.com/settings/user/tokens) (not project API permission token)

.circleci/config.yml
```diff
@@ -355,6 +382,201 @@ jobs:
...
+        type: boolean
+    executor:
+      name: vm-<< parameters.arch >>
+    parallelism: << parameters.parallelism >>
+    steps:
+      - checkout
+      - run:
+          name: Download build output from GitHub action
+          command: |
+            echo "Running with: \
```
michaelbeaumont, line 418: Hmm could this be a shell script? Could be easier to test

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,222 @@
...
+              echo "{}"
+              return
+            fi
+            if [ $STATUS_CODE -lt 200 ] || [ $STATUS_CODE -gt 399 ] ; then
+              echo "Error requesting $METHOD $URL (status $STATUS_CODE)"
+              exit 1
+            fi
+          }
+
+          function check_workflow(){
```
michaelbeaumont, line 168: Also a lot of big inline shell scripts maybe better in separate files?

.circleci/config.yml
jijiechen: I did use a separated file to test when authoring locally.
This script may be refactored after the transition from CircleCI to GH Actions. In addition, this script is only useful on the CI, we don't use it anywhere else.

.github/workflows/e2e.yaml
jijiechen: The structure of the script is quite simple though.

PR #8478, opened 2023-11-28
feat(kuma-cp): deprecate standalone mode

jakubdyszkiewicz:
### Checklist prior to review

Draft because we still need to merge MADR


pkg/api-server/resource_endpoints.go
```diff
@@ -418,7 +419,7 @@ func (r *resourceEndpoints) validateResourceRequest(request *restful.Request, re
 	if name != resourceMeta.Name {
 		err.AddViolation("name", "name from the URL has to be the same as in body")
 	}
-	if r.mode == config_core.Zone && !r.doesNameLengthFitsGlobal(name) {
+	if r.mode == config_core.Zone && r.federatedZone && !r.doesNameLengthFitsGlobal(name) {
```
lahabana, line 422: Can we just remove the r.mode?

lahabana: Actually should this always be the case (even in non federated cases)?

lahabana: Do we also have a future workaround this limitation with the name stuff from Ilya?

pkg/config/app/kuma-cp/config.go
```diff
@@ -271,28 +275,11 @@ func (c *Config) Validate() error {
...
-		if err := c.Runtime.Validate(c.Environment); err != nil {
-			return errors.Wrap(err, "Runtime validation failed")
-		}
-		if err := c.Metrics.Validate(); err != nil {
-			return errors.Wrap(err, "Metrics validation failed")
-		}
 	case core.Zone:
-		if err := c.Multizone.Zone.Validate(); err != nil {
-			return errors.Wrap(err, "Multizone Zone validation failed")
+		if c.IsFederatedZoneCP() {
```
lahabana, line 279: Isn't the validation going to not happen if only one of globalAddress or Zone name is set?

pkg/api-server/resource_endpoints.go
jakubdyszkiewicz: we can, yes

lahabana (commented):
Nice!

pkg/config/app/kuma-cp/config.go
jakubdyszkiewicz: I reworked this validation.

pkg/api-server/resource_endpoints.go
lobkovilya: > Do we also have a future workaround this limitation with the name stuff from Ilya?

Yes, we won't append `.zone` suffix in the future (it will be replaced with hash-suffix + labels). This means won't don't need `doesNameLengthFitsGlobal` func

slonka (commented):
There are still references to "standalone" mode in the codebase. You can find them by running:

```
grep -rI standalone .
./app/kumactl/cmd/install/install_control_plane_test.go:                Entry("--kds-global-address is used with standalone", errTestCase{
./app/kumactl/cmd/install/install_control_plane_test.go:                        extraArgs: []string{"--kds-global-address", "192.168.0.1:1234", "--mode", "standalone"},
./app/kuma-dp/pkg/dataplane/envoy/remote_bootstrap.go:                  log.Info("could not fetch bootstrap configuration, make sure you are not trying to connect to global-cp. retrying (this could help only if you're connecting to zone-cp or standalone).", "backoff", c
fg.ControlPlane.Retry.Backoff, "err", err.Error())
...
```

A more general note: instead of renaming current "zone" to "federated zone" (and "standalone" to just "zone") wouldn't it be easier for everyone to digest if we just named "standalone" a "standalone zone"? Just asking, I'm not very firm on this.

slonka (approved):
LGTM, please remember about the docs issue / PR.

Issue #8468, opened 2023-11-27
Refactor InferProtocol 


lukidzi:
### Description

We are calling `InferProtocol` which uses `routing` from MeshContext. This routing object has data filtered by old policies and we shouldn't use them.

PR #8479, opened 2023-11-28
feat(kuma-cp): move protocol information to mesh context

lukidzi:
### Checklist prior to review

We called the method `InferProtocol` in some plugins with Endpoints already filtered by old policies. That caused some of the services not to be available. In this change, I am creating a Map with service information about protocol based on EndpointMap. This allows us to make the calculation once in a mesh context and share it between plugins. Also, we don't need `proxy.Routing` which we want to avoid using in new policies.

  - Closes: https://github.com/kumahq/kuma/issues/8468


pkg/insights/resyncer.go
```diff
@@ -473,17 +479,17 @@ func (r *resyncer) createOrUpdateServiceInsight(
 			case mesh_proto.Dataplane_Networking_Gateway_DELEGATED:
 				svcType = mesh_proto.ServiceInsight_Service_gateway_delegated
 			}
-			populateInsight(svcType, insight, gw.GetTags()[mesh_proto.ServiceTag], status, backend, "")
+			populateInsight(svcType, insight, gw.GetTags()[mesh_proto.ServiceTag], status, backend, "", core_mesh.ProtocolHTTP)
```
lahabana, line 482: I don't think that's right.

pkg/plugins/runtime/gateway/route/util.go
```diff
@@ -23,18 +21,19 @@ func InferServiceProtocol(endpoints []core_xds.Endpoint, routeProtocol core_mesh
 			return core_mesh.ProtocolHTTP
 		}
 	}
```
lahabana, line 23: why this new if?

pkg/xds/context/mesh_context_builder.go
```diff
@@ -284,19 +287,6 @@ func (m *meshContextBuilder) fetchResourceList(ctx context.Context, resType core
...
-		if mesh == nil {
-			return desc.NewList(), nil
-		}
-		// ServiceInsights in XDS generation are only used to check whether the destination is ready to receive mTLS traffic.
-		// This information is only useful when mTLS is enabled with PERMISSIVE mode.
-		// Not including this into mesh hash for other cases saves us unnecessary XDS config generations.
-		if backend := mesh.GetEnabledCertificateAuthorityBackend(); backend == nil || backend.Mode == mesh_proto.CertificateAuthorityBackend_STRICT {
-			return desc.NewList(), nil
-		}
-	}
```
lahabana, line 299: There's a performance impact here we should discuss. My biggest concern is that service-insight gets updated whenever a DPP goes up or down. That's potentially a lot of very frequent updates

pkg/xds/context/context.go
```diff
@@ -66,11 +66,16 @@ type MeshContext struct {
...
-	ServiceTLSReadiness    map[string]bool
+	ServiceInformations    ServiceInformations
 	DataSourceLoader       datasource.Loader
 	ReachableServicesGraph ReachableServicesGraph
 }
 
+type ServiceInformations struct {
+	TLSReadiness map[string]bool
+	Protocol     map[string]core_mesh.Protocol
+}
```
lahabana, line 77: Would it make more sens to have ServiceInformations `map[string]ServiceInformation` instead of the double map?

pkg/xds/generator/outbound_proxy_generator.go
```diff
@@ -309,18 +310,18 @@ func (OutboundProxyGenerator) generateEDS(
...
+func InferProtocol(servicesProtocol map[string]core_mesh.Protocol, clusters []envoy_common.Cluster) core_mesh.Protocol {
+	var protocol core_mesh.Protocol = core_mesh.ProtocolUnknown
 	for _, cluster := range clusters {
 		serviceName := cluster.Tags()[mesh_proto.ServiceTag]
-		endpoints := model.EndpointList(proxy.Routing.OutboundTargets[serviceName])
-		allEndpoints = append(allEndpoints, endpoints...)
-		endpoints = proxy.Routing.ExternalServiceOutboundTargets[serviceName]
-		allEndpoints = append(allEndpoints, endpoints...)
+		currentProtocol := servicesProtocol[serviceName]
+		if protocol == core_mesh.ProtocolUnknown && currentProtocol != "" {
```
lahabana, line 318: Isn't this handled in `GetCommonProtocol` ?

lahabana (commented):
i don't see any of the backcompat code here. What happens when we upgrade and serviceInsight is not populated? Also what happens when a new service is created and it's not in serviceInsight yet? I'd expect we need to artificially backfill for at least back compat no?

pkg/xds/context/mesh_context_builder.go
lukidzi: Yes, that's true. I think it is a potential blocker for using ServiceInsights for protocol map. Other options is to use `EndpointMap` in the mesh context and calculate by ourselves

pkg/insights/resyncer.go
lukidzi: I am not fully sure how to handle this for the gateway. Gateway can have more than one listener with different protocols. Maybe I could just take all protocols from listeners and based on this call common protocol

pkg/plugins/runtime/gateway/route/util.go
lukidzi: Had a case when Gateway was pointing to a service that wasn't in the map and that caused an empty string

pkg/xds/context/context.go
lukidzi: I think we can change it, for storing the data it might be better but using structure would require some validations

pkg/xds/generator/outbound_proxy_generator.go
lukidzi: It isn't but can add it

pkg/xds/context/mesh_context_builder.go
```diff
@@ -365,6 +356,36 @@ func modifyAllEntries(list core_model.ResourceList, fn func(resource core_model.
...
+		if _, ok := protocol[svc]; !ok {
+			protocol[svc] = util_protocol.InferServiceProtocol(endpoints)
+		}
+	}
+	return protocol
+}
+
+// backward compatibility
+// instead of doing this code in each plugin we can do this once in mesh context
+func (m *meshContextBuilder) fallbackResolveProtocol(endpointMap xds.EndpointMap) map[string]core_mesh.Protocol {
```
lukidzi, line 381: Maybe we should use this instead of ServiceInsights?

lahabana:
Jakub pointed that service insights are copied from Global to Zone. We should check and understand this.

lukidzi:
> Jakub pointed that service insights are copied from Global to Zone. We should check and understand this.

Yes, it seems like ServiceInsights calculations are running on Global/Standalone but not Zone. So if there is no connection we might have old data. The code of resyncer is responsible for serviceInsight generation. 

We have some options:
* relay on ServiceInsight - problem when Global is down
* enable ServiceInsight on zone - we would need to sync to global/create global serviceInsight? 
* use EndpointMap - instead of calling for each policy we can just build this in mesh context once per each service and provide protocol in ServiceInsight only for UI purpose?

pkg/xds/context/context.go
lahabana: Why required validation?

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -22,7 +22,7 @@ func GenerateClusters(
 
 	for _, serviceName := range services.Sorted() {
 		service := services[serviceName]
-		protocol := generator.InferProtocol(proxy, service.Clusters())
+		protocol := generator.InferProtocol(meshCtx.ServiceInformations.Protocol, service.Clusters())
```
lahabana, line 25: I'm not sure to understand why we infer a common protocol when we build a different cluster for each of the service.

pkg/util/protocol/protocol.go
```diff
@@ -51,15 +51,14 @@ func getCommonProtocol(one, another core_mesh.Protocol) core_mesh.Protocol {
 	return core_mesh.ProtocolUnknown
 }
 
-// InferServiceProtocol returns a common protocol for a given group of endpoints.
 func InferServiceProtocol(endpoints []core_xds.Endpoint) core_mesh.Protocol {
```
lahabana, line 54: I think this should just move in to be a private function in mesh_context_builder to avoid further usage of this.

pkg/xds/context/context.go
lahabana: Why you are there should you also add `isExternalService` ? Feels like it was the other things required to reduce our reliance on `core_xds.Routing` in plugins.

lahabana (commented):
I think you are right we're trying to do too much at once. Let's remove the deps of plugins on `proxy.Routing` we can later look not use all the dataplanes to figure the protocol of a service.

pkg/xds/context/context.go
lukidzi: The one issue that I noticed is circular dependency I might need to change some packages

lahabana: Circular deps between what and what?

pkg/plugins/policies/core/xds/meshroute/clusters.go
lukidzi: Fixed

pkg/plugins/policies/core/xds/meshroute/listeners.go
```diff
@@ -19,6 +20,7 @@ func MakeTCPSplit(
 	clusterCache map[common_api.TargetRefHash]string,
 	servicesAcc envoy_common.ServicesAccumulator,
 	refs []common_api.BackendRef,
+	servicesInforomations map[string]xds_context.ServiceInformations,
```
lahabana, line 23: Typo

pkg/xds/context/context.go
```diff
@@ -58,17 +58,24 @@ func (g BaseMeshContext) Hash() string {
...
+	ExternalServicesEndpointMap xds.EndpointMap
+	CrossMeshEndpoints          map[xds.MeshName]xds.EndpointMap
+	VIPDomains                  []xds.VIPDomains
+	VIPOutbounds                []*mesh_proto.Dataplane_Networking_Outbound
+	ServiceInformations         map[string]ServiceInformations
+	DataSourceLoader            datasource.Loader
+	ReachableServicesGraph      ReachableServicesGraph
+}
+
+type ServiceInformations struct {
```
lahabana, line 75: I think it should be singular (information is almost always si gular in English I think

bartsmykla (commented):
Some nits

bartsmykla (approved):

pkg/plugins/runtime/gateway/testdata/tls/tcp-route.yaml
```diff
@@ -45,6 +45,13 @@ Clusters:
                 resourceApiVersion: V3
           sni: api-service{mesh=default}
       type: EDS
+      typedExtensionProtocolOptions:
+        envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+          '@type': type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+          commonHttpProtocolOptions:
+            idleTimeout: 3600s
+          explicitHttpConfig:
+            httpProtocolOptions: {}
```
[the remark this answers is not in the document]
lukidzi: It's because we are building map of protocol in mesh context based on EndpointMap. Test EndpointMap has a definition of this service with a `kuma.io/protocol: http`

pkg/xds/context/context.go
```diff
@@ -101,6 +108,25 @@ func (mc *MeshContext) GetLoggingBackend(tl *core_mesh.TrafficLogResource) *mesh
...
+	tlsReady := map[string]bool{}
+	for serviceName, info := range mc.ServicesInformation {
+		if info != nil {
+			tlsReady[serviceName] = info.TLSReadiness
+		} else {
+			tlsReady[serviceName] = false
+		}
+	}
+	return tlsReady
+}
```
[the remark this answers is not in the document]
lukidzi: They are a bit different. `GetServiceProtocol` return protocol for specific service and if not found - UNKNOWN
Other returns tls readiness for all services.

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go
```diff
@@ -48,10 +49,10 @@ func generateListeners(
 				NormalizePath:            true,
 			}))
 
-		protocol := plugins_xds.InferProtocol(proxy.Routing, serviceName)
+		protocol := meshCtx.GetServiceProtocol(serviceName)
 		var routes []xds.OutboundRoute
 		for _, route := range prepareRoutes(rules, serviceName, protocol) {
-			split := meshroute_xds.MakeHTTPSplit(proxy, clusterCache, servicesAcc, route.BackendRefs)
+			split := meshroute_xds.MakeHTTPSplit(proxy, clusterCache, servicesAcc, route.BackendRefs, meshCtx)
```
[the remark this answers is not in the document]
lahabana: ok then

pkg/plugins/runtime/gateway/testdata/tls/tcp-route.yaml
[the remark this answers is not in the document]
lahabana: Should we fix the test case?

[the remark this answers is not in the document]
lukidzi: The difference is that previously we were taking `tags` from the destination field in the policy and putting them into the endpoints structure. Based on these endpoints control-plane was getting the protocol. Because in the destination there is no tag with the policy only the service name we were getting the TCP. Now I've changed that instead of creating endpoints based on the tags in the destinations we are retrieving the data from the protocol map. It has HTTP protocol tag so that's why we are getting it now.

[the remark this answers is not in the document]
lukidzi: Maybe to keep it constant I should take the value of the protocol tag from the `destination`?

pkg/xds/context/context.go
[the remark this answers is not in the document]
lahabana: I tlsReadiness exists for all services then we should be able to set UNKNOWN when we need to no?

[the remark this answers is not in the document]
lukidzi: TLS readiness is set only when using permissive TLS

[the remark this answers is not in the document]
lukidzi: I've changed a bit that we init a new object with UNKNOWN but still, if there is no service we are returning UNKNOWN.

lahabana (approved):
LGTM! Sorry I wanted to sit down to properly look at this change

PR #8485, opened 2023-11-28
feat(kuma-dp): add coredns logging flag

timothy-spencer:
As somebody who is trying to set up a mesh in kubernetes and run terraform in it, I need a way to turn on CoreDNS query logging or debugging on, so that I can use the logs from that to enumerate all the external services that I need to allow outbound access to.

This patch adds a flag to the helm chart that will inject an environment variable into the dataplane sidecars that will add the "log" plugin to the CoreDNS config. I added a test to make sure the config would get set in dnsserver, and I checked that `helm template` would add the `KUMA_RUNTIME_KUBERNETES_INJECTOR_BUILTIN_DNS_LOGGING` env var to the control plane, which will cause the kuma-dp sidecars to get logging turned on and going to stdout.

I also ran `make check` and `make test` cleanly.

Fix https://github.com/kumahq/kuma/issues/8254 and maybe https://github.com/kumahq/kuma/issues/5516

This PR was originally discussed under https://github.com/kumahq/kuma/pull/8285, where I messed up by syncing my branch to main by accident, as this is not my normal workflow.  However, I hope I have redeemed myself by rebasing properly and getting everything up to date in this one at least.  :-)

### Checklist prior to review

- [X] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? -- No.
- [X] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) -- No.


timothy-spencer:
Hey, not sure why it failed the test.  It is in code that is nowhere near anything I did, from what I can see.  I suspect that it was some sort of ephemeral failure, as the test that failed timed out rather than actually failing to pass the test.

In any case, please let me know what you think you'd like to do with this PR.  I still think it will be a really useful feature for your users.  Thanks!

timothy-spencer:
Another timeout on a different test.  I just am not lucky today.  :-)

lahabana:
Thanks for the contrib! (The failing check is for an incubating part of our CI)

lahabana (approved):

PR #8487, opened 2023-11-29
test(policy): add extra testing of the matching algorithm

### LINKED ISSUE #8484: Rule merging alghoritm is not sorting lexographically on MeshTrafficPermission
### What happened?

My intention is to create a `MeshTrafficPermission` set of rules that will allow connectivity from some services and ShadowDeny (or Deny) everything else. 

To achieve that i did a 2 `MeshTrafficPermission` objects, both on  `MeshSubset` targetRef level.

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  labels:
    kuma.io/mesh: default
  name: default-demo-app
  namespace: kuma-system
spec:
  targetRef:
    kind: MeshSubset
    tags:
  from:
    - default:
        action: Allow
      targetRef:
        kind: MeshSubset
        tags:
          app.kubernetes.io/name: ui
    - default:
        action: Allow
      targetRef:
        kind: MeshSubset
        tags:
          app.kubernetes.io/name: service-order
    - default:
        action: Allow
      targetRef:
        kind: MeshSubset
        tags:
          app.kubernetes.io/name: service-payment
---

apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  labels:
    kuma.io/mesh: default
  name: default-demo-app-b
  namespace: kuma-system
spec:
  targetRef:
    kind: MeshSubset
    tags:
  from:
    - default:
        action: Deny
      targetRef:
        kind: Mesh
```

This setup works fine if naming is like `myrule` for allow set and `myrule-a` for denyset. I see in UI that rules are applied in correct order:
Screenshot 2023-11-28 at 15 54 03

My tests also shows that traffic works as expected

However, when I try to use names like `rule-a` and `rule-b` order is opposite to expected and tests also failing.

Screenshot 2023-11-28 at 15 56 25

I am not sure how sorting is working, but looks like that not according to doc which state that `2. At the same level we use lexicographic order on name`.

lahabana:
- Add some tests to the policy matching algo
- Add possibility to add a test as pending in `test.EntriesForFolder`

Ref #8484

### Checklist prior to review


lobkovilya:
Tests in this PR are not expected to pass, because sorting normally is happening here https://github.com/kumahq/kuma/blob/b249dab0157cf18dee9968ef24a51a4901bc9c6b/pkg/plugins/policies/core/matchers/dataplane.go#L63-L67

But the test calls `core_rules.BuildFromRules` function that already expects sorted policies as input.

lobkovilya:
If you place the same test here `pkg/plugins/policies/core/matchers/dataplane_test.go`, it works as expected.

lahabana:
Yeah that's what I've noticed later

lahabana:
Yeah I noticed that after so I added tests there as well and it seems to be working correctly. So need to understand the issue some more

Issue #8142, opened 2023-10-25
Support MeshHTTPRoute/MeshTCPRoutes with MeshGateways


lahabana:
### Description

MeshGatewayRoute should just be using `MeshHTTPRoute` and `MeshTCPRoute`.

For this to happen we need the top level targetRef to support `kind: MeshGateway` and be able to identify a specific listener.

```yaml
targetRef:
  kind: MeshGateway
  tags:
     kuma.io/service: gw-1
     listener: http
```

In the case of MeshGateway we do the usual thing of merging together gateway and listener tags.

Things we need to take care of:

- [x] [Allow selecting listeners](https://github.com/kumahq/kuma/issues/8549)
- [x] [Need Hostname support in MeshHTTPRoute which would need to be supported for regular routes too](https://github.com/kumahq/kuma/pull/8663).
- [x] [Add rewrite hostname to backend like in MeshGatewayRoute](https://github.com/kumahq/kuma/pull/8772)
- Migration for MeshGatewayRoute to MeshHTTPRoute: https://github.com/kumahq/kuma/issues/8923
- One point is we do RDS for Gateway but not for Mesh*Route. We should probably to RDS for everything: https://github.com/kumahq/kuma/issues/8921
- Update our GatewayAPI controller stop using MeshGatewayRoute: https://github.com/kumahq/kuma/issues/8922

michaelbeaumont:
Implemented

Issue #6070, opened 2023-02-20
Policy merging discrepancy between Kubernetes and Universal with defaults


michaelbeaumont:
### What happened?

We use the `kubebuilder:default` annotation to generate an OpenAPI spec describing our resources. When we create our CRDs with these OpenAPI specs, Kubernetes then _sets_ the value of omitted fields to the default. On Universal, we don't do the same, this part of the spec serves only as documentation

This means:
* Merging won't work as intended. Given a `kubebuilder:default=30s` for `interval`:
```
---
targetRef:
kind: Mesh
to:
  - default:
      interval: 100s
      timeout: 200s
---
targetRef:
  kind: MeshService
  name: backend
to:
  - default:
      timeout: 300s
```
on Universal we get `{interval: 100s, timeout: 300s}` but on Kubernetes we get `{interval: 30s, timeout: 300s}` because the API server sets `interval: 30s` on the second resource.

* when writing policy implementations, we have to duplicate the default everywhere we use the fields for the universal case, even though on Kubernetes they'll always be set

lahabana:
Looking at the code we have a `Defaulter` interface that does this fairly well: https://github.com/kumahq/kuma/blob/f5f63bd91b7a5384b11d87c772d66704247e2dbd/pkg/core/resources/manager/manager.go#L59-L66

We should likely make it work in a similar way as `Validate` in the generator ([if there's a `validate` method do something](https://github.com/kumahq/kuma/blob/17fe0560f9b09e4049dbef4a42cd2b6bfb8a124b/tools/policy-gen/generator/cmd/core_resource.go#L157-L163)) we could then auto generate the defaulter that use kubebuilder annotations.

If we are to do this and always rely on kubebuilder annotations for defaults we should remove this:
https://github.com/kumahq/kuma/blob/e590c3d8f47eee36e63f18b047151dd8d18975fb/pkg/plugins/runtime/k8s/webhooks/defaulter.go#L53-L58 and fix MeshService and Mesh which are the 2 existing implems of `Defaulter`

lahabana:
Btw MeshService implements the Defaulter when likely it should use the comment

slonka:
> This also seems to be in the openAPI spec, can't we just use openAPI magic to have the defaults ensured? A bit like we do in:

We can, and this is what I do here: https://github.com/kumahq/kuma/pull/12829/files#r1952173703

> If we are to do this and always rely on kubebuilder annotations for defaults we should remove this:

[...]

PR #8489, opened 2023-11-29
docs(MADR): support `MeshHTTPRoute` with `MeshGateway`

michaelbeaumont:
### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- #8142 


docs/madr/decisions/033-meshhttproute-meshgateway.md
```diff
@@ -0,0 +1,104 @@
...
+- have users on each rule do hostname matching using the header match on `host`
+
+This MADR proposes adding `spec.to[].hostnames`. Hostname matching is in some sense the
+"highest level" matching possible with HTTP and it's likely that all rules of a
+given resource are intended to apply to a given hostname set, whether that's a
+specific hostname, a wildcard pattern or to all hostnames.
+
+The second option might be feasible but it only makes sense if the cost of having
+host matching separate is very high. It also differs from Gateway API, which is
+a disadvantage.
```
lahabana, line 51: +1 for option 1. I think even if hostnames had an effect on regular routes that wouldn't be a problem. So to be clear this would be part of the `Match` part of the spec? Maybe an example would be good here.

docs/madr/decisions/033-meshhttproute-meshgateway.md
```diff
@@ -0,0 +1,104 @@
...
+### `to.targetRef`
+
+When using `MeshHTTPRoute` for service to service communication, it's very
+useful to limit the effects based on request source and destination. We do
+this by setting `spec.targetRef`, the source, and `spec.to[].targetRef`, the destination.
+
+With `MeshGateway`, once a request hits the gateway, the routing resources
+_define_ the destination. We have to put the rules somewhere, so this MADR
+proposes `spec.to` but there's no semantic value to a `spec.to[].targetRef`
+or for that matter, a `spec.from[].targetRef`.
```
lahabana, line 62: So to be clear here is that `spec.to[].target` would be unset if the top level targetRef is MeshGateway?
Again an example would be good.
Would it be not allowed to bind a route using MeshSubset or MeshService? .i.e if the DPP is a gateway then we only look at MeshGateway top level targetRefs? Does that make policy matching quite different for routes?

docs/madr/decisions/033-meshhttproute-meshgateway.md
```diff
@@ -0,0 +1,104 @@
...
+  - rules:
+    - matches:
+      - path:
+          type: PathPrefix
+          value: /
+      default:
+        backendRefs:
+        - kind: MeshService
+          name: backend_demo_svc_8080
+```
```
lahabana, line 95: What about MeshTCPRoute?

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: Yes it would be not allowed. I guess it's different but the intention AFAIU at the moment anyway is to avoid MeshService and friends for matching on gateways.

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: Yeah good point, I think that could belong in this MADR. I'll add it

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: Good point I'll add some more examples.

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: Yeah this is already present in the example

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: added `hostnames` to the example. it's not in match. it's next to `rules`.

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: added

docs/madr/decisions/033-meshhttproute-meshgateway.md
```diff
@@ -0,0 +1,133 @@
...
+### Route matching
+
+This MADR doesn't address supporting policy matching of routes
+that attach to `MeshGateways` because we can't policy match on `MeshGatewayRoute` at
+the moment anyway. The feasability/details should be addressed in a separate MADR.
+
+### `MeshHTTPRoute` example
+
+```
+apiVersion: kuma.io/v1alpha1
```
lobkovilya, line 77: What backend will be picked if I send a request to `test.example.com`?
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshHTTPRoute
metadata:
  name: edge-routes-a
spec:
  targetRef:
    kind: MeshGateway
    name: edge-gateway
  to:
  - hostnames:
    - test.example.com
    rules:
    - matches:
      - path:
          type: PathPrefix
          value: /
      default:
        backendRefs:
        - kind: MeshService
          name: backend_a
---
apiVersion: kuma.io/v1alpha1
kind: MeshHTTPRoute
metadata:
  name: edge-routes-b
spec:
  targetRef:
    kind: MeshGateway
    name: edge-gateway
  to:
  - hostnames:
    - *.example.com
    rules:
    - matches:
      - path:
          type: PathPrefix
          value: /
      default:
        backendRefs:
        - kind: MeshService
          name: backend_b
```

michaelbeaumont: The first.

michaelbeaumont: Semantically, they'd create a  set of rules that can only be merged with other rule sets that have the exact same `hostname`. I tend to think it's not feasible to do this across wildcards, only between exact matches.

My assumption here was "nothing new, it works just like MeshGatewayRoute." Which I think is still basically true but I'll make it explicit and write up some examples as well.

As far as the implementation, we'll see

docs/madr/decisions/033-meshhttproute-meshgateway.md
```diff
@@ -0,0 +1,133 @@
...
+When using `MeshHTTPRoute`/`MeshTCPRoute` for service to service communication, it's very
+useful to limit the effects based on request source and destination. We do
+this by setting `spec.targetRef`, the source, and `spec.to[].targetRef`, the destination.
+
+With `MeshGateway`, once a request hits the gateway, the routing resources
+_define_ the destination. We have to put the rules somewhere, so this MADR
+proposes `spec.to` but there's no semantic value to a `spec.to[].targetRef`
+or for that matter, a `spec.from[].targetRef`.
+
+This MADR proposes allowing `spec.to[].targetRef` to be omitted. Another
```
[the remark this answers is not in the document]
michaelbeaumont: OK, yeah I dont even have that strong of an opinion. I'll switch to `kind: Mesh`.

With open API schemas in the CRDs this would actually just be a `kubebuilder:default` annotation. But we really need https://github.com/kumahq/kuma/issues/6070

docs/madr/decisions/033-meshhttproute-meshgateway.md
lahabana: Yeah ideally I feel like this should just be similar to being part of `matches` section.

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: @lahabana WDYT? The problem with "allowing" `MeshService` and friends for gateways is that it prevents validation. In particular, atm we _disallow_ `kind: Mesh` for non-MeshGateway MeshHTTPRoutes.

lahabana: I'm not sure I understand the precise question.
But I'm fine with having only MeshGateway work for gateways.

michaelbeaumont: > But I'm fine with having only MeshGateway work for gateways 

That was the question. Yeah we should probably ignore such routes.

docs/madr/decisions/033-meshhttproute-meshgateway.md
```diff
@@ -0,0 +1,318 @@
...
+When using `MeshHTTPRoute`/`MeshTCPRoute` for service to service communication, it's very
+useful to limit the effects based on request source and destination. We do
+this by setting `spec.targetRef`, the source, and `spec.to[].targetRef`, the destination.
+
+With `MeshGateway`, once a request hits the gateway, the routing resources
+_define_ the destination. We have to put the rules somewhere, so this MADR
+proposes `spec.to`. There's no real semantic value to a distinct
+`spec.to[].targetRef`, or for that matter, a `spec.from[].targetRef`.
+
+This MADR proposes requiring `spec.to[].targetRef.kind: Mesh`.
```
jakubdyszkiewicz, line 64: Maybe I'm missing something, but it says that it requires `spec.to[].targetRef.kind: Mesh`, but I see no such thing in the examples.

michaelbeaumont: Woops, forgot to update the examples

docs/madr/decisions/033-meshhttproute-meshgateway.md
```diff
@@ -0,0 +1,338 @@
...
+    - matches:
+      - path:
+          type: PathPrefix
+          value: /
+      default:
+        backendRefs:
+        - kind: MeshServiceSubset
+          name: backend_demo_svc_8080
+          tags:
+            version: v1
```
lobkovilya, line 184: both examples have `v1`, I assume one of them should be `v2`

docs/madr/decisions/033-meshhttproute-meshgateway.md
```diff
@@ -0,0 +1,338 @@
...
+---
+apiVersion: kuma.io/v1alpha1
+kind: MeshHTTPRoute
+metadata:
+  name: wildcard-hostname
```
lobkovilya, line 164: do we care about the lexicographic order of the names in this case?

docs/madr/decisions/033-meshhttproute-meshgateway.md
lobkovilya: Are `hostnames` basically a syntactic sugar for `matches`? For example
```yaml
to:
- hostnames:
  - test1.example.com
  - *.domain.com
  rules:
    - matches:
      - path:
          type: PathPrefix
          value: /
```
results in 
```yaml
to:
- rules:
    - matches:
      - path:
          type: PathPrefix
          value: /
      - headers:
        - type: Exact
          name: Host
          value: test1.example.com
        - type: RegularExpression
          name: Host
          value: *.domain.com
```

If we treat `hostnames` as syntactic sugar, we don't have to introduce additional complexity into the matching algorithm. Also, `hostnames` naturally becomes part of the `mergeKey`

docs/madr/decisions/033-meshhttproute-meshgateway.md
```diff
@@ -0,0 +1,338 @@
...
+```
+---
+apiVersion: kuma.io/v1alpha1
+kind: MeshHTTPRoute
+metadata:
+  name: specific-hostname
+spec:
+  targetRef:
+    kind: MeshGateway
+    name: edge-gateway
```
lobkovilya, line 236: If I want to target a specific listener on the MeshGateway, should I use `targetRef.tags`?

lahabana (approved):

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: Yeah that's what we decided on in [the `kind: MeshGateway` MADR](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/027-meshgateway-targetref.md#decision-outcome).

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: Of the resource names? Don't we use them now to sort the rules?

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: I can't come up with an example where it has different semantics.

jakubdyszkiewicz (approved):

docs/madr/decisions/033-meshhttproute-meshgateway.md
lobkovilya: I just realized we always have pairs of kinds (one with tags and one without), i.e. Mesh (no tags) and MeshSubset (with tags), same for MeshService and MeshServiceSubset. Should the version of `kind: MeshGateway` with tags have a different name, like `MeshGatewayListener`?

docs/madr/decisions/033-meshhttproute-meshgateway.md
lobkovilya: Yes, but it seems like for MeshHTTPRoute we build the order that makes the most sense, no? i.e. generic matches like `PathPrefix: /` on top and more specific `Exect: /bar` at the bottom?

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: Honestly I was going to propose deprecating `MeshServiceSubset` :grin: 
IMO a different kind isn't necessary. Not sure what others think

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: We will have to ignore such routes in some cases, like `from` with `MeshFaultInjection`

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: @lobkovilya If you feel strongly about this then should we discuss it separately? It's not particular to `MeshHTTPRoute`

docs/madr/decisions/033-meshhttproute-meshgateway.md
michaelbeaumont: Yes this is already part of the MeshHTTPRoute spec independent of this PR.

lahabana:
@michaelbeaumont what are we missing to get this merged?

michaelbeaumont:
@lahabana I've been discussing things with @lobkovilya and trying to figure out if the `hostnames` merging makes sense, I updated the implem PR but need to make a change in the MADR

lobkovilya (approved):

Issue #8483, opened 2023-11-28
InboundProxyGenerator: UpstreamBindConfig is always setup when using transparent proxying


nicoche:
### What happened?

Hi!

It looks like this condition https://github.com/kumahq/kuma/blob/ee9cfca281a27322f425005679256d7b067eb6e1/pkg/xds/generator/inbound_proxy_generator.go#L45 is always evaluated to true whenever TransparentProxy is enabled. Is it expected? Should it be the following instead?

```
if proxy.Dataplane.IsUsingTransparentProxy() && (endpoint.WorkloadIP == core_mesh.IPv4Loopback.String() || endpoint.WorkloadIP == core_mesh.IPv6Loopback.String()) {
```

(see https://kuma-mesh.slack.com/archives/CN2GN4HE1/p1701184894860479 for more details)

lukidzi:
Hi. I think the condition should be 
```
if proxy.Dataplane.IsUsingTransparentProxy() && (endpoint.WorkloadIP != core_mesh.IPv4Loopback.String() && endpoint.WorkloadIP != core_mesh.IPv6Loopback.String())
```

We should set upstream bind when:
1. Transparent proxy is enabled
2. Workload IP is different than Loopback IPv4 and IPv6

PR #8490, opened 2023-11-29
fix(kuma-cp): fix address check to not be loopback ipv4 and ipv6

lukidzi:
### Checklist prior to review
There was a bug that we didn't correctly check for WorkloadIP. Now we are setting `upstreamBind` only when the transparent proxy and the `WorkloadIP` is different than loopback v4 and v6.

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/8483


nicoche:
Hey! Thanks for looking into this. Heads up, the same condition exists in `pkg/hds/tracker/healthcheck_generator.go`

pkg/xds/generator/inbound_proxy_generator.go
```diff
@@ -42,7 +42,8 @@ func (g InboundProxyGenerator) Generate(ctx context.Context, _ *core_xds.Resourc
 			Configure(envoy_clusters.ProvidedEndpointCluster(false, core_xds.Endpoint{Target: endpoint.WorkloadIP, Port: endpoint.WorkloadPort})).
 			Configure(envoy_clusters.Timeout(defaults_mesh.DefaultInboundTimeout(), protocol))
 		// localhost traffic is routed dirrectly to the application, in case of other interface we are going to set source address to
-		if proxy.Dataplane.IsUsingTransparentProxy() && (endpoint.WorkloadIP != core_mesh.IPv4Loopback.String() || endpoint.WorkloadIP != core_mesh.IPv6Loopback.String()) {
+		// 127.0.0.6 to avoid redirections and thanks to first iptables rule just return fast
+		if proxy.Dataplane.IsUsingTransparentProxy() && (endpoint.WorkloadIP != core_mesh.IPv4Loopback.String() && endpoint.WorkloadIP != core_mesh.IPv6Loopback.String()) {
```
lahabana, line 46: @nicoche I think that's something you pointed out so would love a review :)

nicoche: tbh I don't fully grasp how this all is supposed to work; I'm not familiar with k8s setup nor with the default iptables, so take my review with a grain of salt.

In our case, `WorkloadIP` is a `10.0. ...` address. We would validate this new condition and Envoy's configuration would have:

```
"upstream_bind_config": {
  "source_address": {
    "address": "127.0.0.6",
    "port": 0,
  },
}
```

This does not work for us as trying to initiate a connection with a loopback source address and a non-loopback destination address seems illegal. Envoy requests fail and log `invalid argument`.

This is why I was thinking that the expected condition should be `if proxy.Dataplane.IsUsingTransparentProxy() && (endpoint.WorkloadIP == core_mesh.IPv4Loopback.String() || endpoint.WorkloadIP == core_mesh.IPv6Loopback.String()) {`

But maybe we're just doing something esoteric? 😕

jakubdyszkiewicz: @bartsmykla to double check please

bartsmykla: It's correct

Issue #8466, opened 2023-11-27
MeshMetric MADR


Automaat:
### Description

Write MADR for new policy MeshMetric

PR #8526, opened 2023-12-04
docs(MADR): `MeshMetric` policy

Automaat:
### Checklist prior to review

Fixes: #8466 


docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,126 @@
...
+1. I think the aggregation that [uses the aggregate name and pod annotations](https://kuma.io/docs/2.5.x/policies/traffic-metrics/#expose-metrics-from-applications) no longer makes sense, as we can use targetRef 
+policies to point to specific DPs. I imagine there would be a default wide policy (possibly `Mesh` scoped) and per app / framework / language overrides.
+2. What I also don’t like about this is that “filters” are under envoy - I’d probably rename it to filter and introduce 
+the concept of “profiles” like we discussed. Maybe something like this:
+
+```yaml
+filter:
+  usedOnly: true # true or false
+  profile: minimal # one of minimal, default, full
+  regex: http2_act.* # only profile or regex can be defined
```
lahabana, line 59: maybe worth adding a list of metrics to ignore? Sometimes you just have 1 or 2 noisy metrics you want to ignore.

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,126 @@
...
+apiVersion: kuma.io/v1alpha1
+kind: MeshMetrics
+metadata:
+  name: default
+  namespace: kuma-system
+spec:
+  targetRef:
+    kind: Mesh | MeshSubset | MeshService | MeshServiceSubset
+  default:
+    filter:
```
lahabana, line 80: should this just be sidecar?

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,126 @@
...
+      profile: minimal
+      regex: http2_act.* # only profile or regex can be defined
+      usedOnly: true
+    aggregate:
```
lahabana, line 84: should this be application and also have a `regex` option?

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Configure MADS based on policy data
+MADS server is responsible for aggregating information about dataplanes needed for Prometheus monitoring. 
+It now extracts information from the `Mesh` resource which proxy should be monitored and creates a `MonitoringAssignment` response.
+
+We need to change the source of our information about monitored proxies from `Mesh` to `MeshMetric` policy. We cannot treat 
+new policy as usual in this context, we should not merge policies, but create different `MonitoringAssignment` for each 
+defined policy, this is needed for scenarios like migrating to a new backend. For example, when we define two `MeshMetric` 
+policies that target the same data plane proxy, we should create two separate `MonitoringAssignment` responses. 
+We can achieve this by extending Prometheus integration, so that it sends backend name with requests to MADS API.
```
lahabana, line 108: Ideally we shouldn't modify MADS. If we need to maybe we should consider using https://prometheus.io/docs/prometheus/latest/configuration/configuration/#http_sd_config instead

Feels like we should have an identifier of some sort for the request for a backend so that it matches the correct backend.

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,126 @@
...
+It needs to make HTTP requests to Envoy to gather stats, apply our custom transformations and convert them to OTLP format. 
+Thanks to usage of [PeriodicReader](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#PeriodicReader) 
+from OTEL SDK, we will be able to make these calls periodically. This reader can use both Prometheus and OTLP exporters 
+to publish metrics to desired backends.
+
+### Using xDS to configure DPP
+In order to pick which backend we will be using for metrics collection, we need to somehow configure our data plane dynamically. 
+The easiest way to do this is to use xDS and create an Envoy listener that will serve configuration for DPP, 
+then we will need to periodically make requests from DPP to it to check if configuration changed. This will enable us
+to not only configure metrics backend, but will make DPP easily configurable from control-plane with policies or any other mechanism.
```
lahabana, line 126: We need more detail about this in this MADR. What's the API like, what's to see on it...

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,126 @@
...
+In order to push metrics to the OTEL collector, and preserve current Prometheus metrics, we need to start using [OTEL metrics sdk](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric). 
+It allows us to export metrics in multiple formats, especially the ones that we are interested in, Prometheus and OTLP. 
+Thanks to this, we will be able to preserve our old functionality and add the possibility to push metrics to the OTEL collector.
+
+### Metrics hijacker
+To preserve metrics hijacker functionality, we need to implement our own OTEL [Producer](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#Producer). 
+It needs to make HTTP requests to Envoy to gather stats, apply our custom transformations and convert them to OTLP format. 
+Thanks to usage of [PeriodicReader](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#PeriodicReader) 
+from OTEL SDK, we will be able to make these calls periodically. This reader can use both Prometheus and OTLP exporters 
+to publish metrics to desired backends.
```
lahabana, line 120: I think we need a schema to show everything here. What exposes what and how :)

docs/madr/decisions/035-mesh-metric-policy.md
jakubdyszkiewicz: agreed. But it's a clever idea 👍

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,126 @@
...
+      - path: "/metrics/prometheus" # is name still needed since targetRef can just point to a name?
+        port: 8888
+      - port: 8000 # default path is /metrics
+    backends:
+      - type: Prometheus
+        prometheus:
+          port: 5670
+          path: /metrics
+          tls:
+            mode: disabled | providedTLS | activeMTLSBackend
```
jakubdyszkiewicz, line 94: nit: we are trying to keep enums as PascalCase, so it's `Disabled`, `ProvidedTLS` and `ActiveMTLSBackend`

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,126 @@
...
+metadata:
+  name: default
+  namespace: kuma-system
+spec:
+  targetRef:
+    kind: Mesh | MeshSubset | MeshService | MeshServiceSubset
+  default:
+    filter:
+      profile: minimal
+      regex: http2_act.* # only profile or regex can be defined
```
jakubdyszkiewicz, line 82: can this be a list?

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,126 @@
...
+#### Changes to the API
+1. I think the aggregation that [uses the aggregate name and pod annotations](https://kuma.io/docs/2.5.x/policies/traffic-metrics/#expose-metrics-from-applications) no longer makes sense, as we can use targetRef 
+policies to point to specific DPs. I imagine there would be a default wide policy (possibly `Mesh` scoped) and per app / framework / language overrides.
+2. What I also don’t like about this is that “filters” are under envoy - I’d probably rename it to filter and introduce 
+the concept of “profiles” like we discussed. Maybe something like this:
+
+```yaml
+filter:
+  usedOnly: true # true or false
+  profile: minimal # one of minimal, default, full
```
jakubdyszkiewicz, line 58: Is it only predefined profiles? Is a list of metrics for each profile a separate MADR?

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: this cannot be a list, it is rewritten to: https://www.envoyproxy.io/docs/envoy/latest/operations/admin#get--stats?filter=regex

jakubdyszkiewicz: ok

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: changed

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: changed

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: changed

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,127 @@
...
+spec:
+  targetRef:
+    kind: Mesh | MeshSubset | MeshService | MeshServiceSubset
+  default:
+    sidecar:
+      profile: minimal
+      regex: http2_act.* # only profile or regex can be defined
+      usedOnly: true
+    application:
+      - path: "/metrics/prometheus" # is name still needed since targetRef can just point to a name?
```
lahabana, line 85: Agreed we don't need a name anymore. Except if we want to provide an extra label to each metric with its source

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: I've added diagram to ilustrate it

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: added more details

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: If we want, we can go further and provide field `extraLabels` that will contain key value pairs of labels to add to all metrics from envoy and application

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: I think that, at least at the beginning, we should allow only predefined profiles. I will work with the envoy stats and prepare a list of metrics for the given profiles to include in this MADR

lahabana: Just realising that this is cool because you can add more stats just for the dataplane you are investigating

docs/madr/decisions/035-mesh-metric-policy.md
lahabana: We could though maybe it's just adding too many features here. Let's keep this MADR manageable and forget about this :)

docs/madr/decisions/035-mesh-metric-policy.md
lahabana: Let's discuss with @jakubdyszkiewicz but I think mads v1 was pretty painful to contribute.

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,171 @@
...
+We need to change the source of our information about monitored proxies from `Mesh` to `MeshMetric` policy. We cannot treat 
+new policy as usual in this context, we should not merge policies, but create different `MonitoringAssignment` for each 
+defined policy, this is needed for scenarios like migrating to a new backend. For example, when we define two `MeshMetric` 
+policies that target the same data plane proxy, we should create two separate `MonitoringAssignment` responses. 
+We can achieve this by extending Prometheus integration, so that it sends backend name with requests to MADS API.
+
+### Pushing metrics from DP to OTEL collector
+In order to push metrics to the OTEL collector, and preserve current Prometheus metrics, we need to start using [OTEL metrics sdk](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric). 
+It allows us to export metrics in multiple formats, especially the ones that we are interested in, Prometheus and OTLP. 
+Thanks to this, we will be able to preserve our old functionality and add the possibility to push metrics to the OTEL collector.
```
lahabana, line 114: I think it's fine if the otel collector doesn't land in 2.6. It'd be great but worth having this lower in the priority list

lahabana (commented):
This is really starting to look good!

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: I will add section `future considerations` with this

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: Right now you can just extend regex with `or` clause. But we can add this in future

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: I would like to push this to separate MADR, as there is too much metrics to easily create profiles out of them, and since we don't have such a possibility currently we won't need this in first version of this policy

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Using xDS to configure DPP
+In order to pick which backend we will be using for metrics collection, we need to somehow configure our data plane dynamically. 
+The easiest way to do this is to use xDS and create an Envoy listener that will serve configuration for DPP, 
+then we will need to periodically make requests from DPP to it to check if configuration changed. This will enable us
+to not only configure metrics backend, but will make DPP easily configurable from control-plane with policies or any other mechanism.
+
+In case of metrics, this enables us to do basically anything with metrics itself: filter, combine, mutate etc.
+
+Our listener could expose `/dpp-config` endpoint, that can serve configuration. Right now we can start with providing only
```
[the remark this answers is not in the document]
Automaat: added more details

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+The question is if we want to allow configuring multiple backends for a single DPP, I think that this is technically possible,
+and can be useful when migrating from `Prometheus` to `OTEL`.
+
+How to implement this: 
+
+We can add another generator like [prometheus_endpoint_generator](https://github.com/kumahq/kuma/blob/f8f5b40a649c4ea5b5dac4ea56ff4fc289da07bc/pkg/xds/generator/prometheus_endpoint_generator.go). 
+This generator will create some basic dataplane configuration. Then we can create new plugin type, eg. `DataplaneConfigurationPlugin` 
+that will be called in the generator to update the config.
```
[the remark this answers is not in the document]
Automaat: Updated madr accordingly

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,197 @@
...
+  namespace: kuma-system
+spec:
+  targetRef:
+    kind: Mesh | MeshSubset | MeshService | MeshServiceSubset
+  default:
+    sidecar:
+      profile: minimal
+      regex: http2_act.* # only profile or regex can be defined
+      usedOnly: true
+    application:
```
jakubdyszkiewicz, line 84: nit: if this is a list, shouldn't this be `applications`?

docs/madr/decisions/035-mesh-metric-policy.md
jakubdyszkiewicz: I'm trying to wrap my head around this. Is the strategy then to _name_ the backend and put it in Prometheus configuration? Schema above does not have a name.

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,197 @@
...
+  }
+}
+```
+
+We should allow configuring multiple metrics backends. 
+
+How to implement this: 
+
+We can add another generator like [prometheus_endpoint_generator](https://github.com/kumahq/kuma/blob/f8f5b40a649c4ea5b5dac4ea56ff4fc289da07bc/pkg/xds/generator/prometheus_endpoint_generator.go). This generator will create separate listener for configuration.
+This listener should be only bound to localhost. We don't need to expose it to the world. In the future we can extend `DataplaneInsight`
```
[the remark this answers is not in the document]
Automaat: added

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: good catch, changed

docs/madr/decisions/035-mesh-metric-policy.md
Automaat: About your last comment, it was in my head, but I forgot about it, added now

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Implementation roadmap
+
+1. New policy API + validation
+2. Move existing functionality to new policy
+   1. Move prometheus_endpoint_generator logic to policy plugin
+   2. Extend MADS to use information from `MeshMetric` plugin data instead of `Mesh` resource
+3. Extend MADS to allow using multiple Prometheus backends
+4. Implement sidecar metrics profiles
+5. Add possibility to dynamically configure DPP
```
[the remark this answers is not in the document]
Automaat: Until this change, everything will work like previously, so restart is required when configuring metrics aggregation from application

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,201 @@
...
+### Configure MADS based on policy data
+MADS server is responsible for aggregating information about dataplanes needed for Prometheus monitoring. 
+It now extracts information from the `Mesh` resource which proxy should be monitored and creates a `MonitoringAssignment` response.
+
+We need to change the source of our information about monitored proxies from `Mesh` to `MeshMetric` policy. We cannot treat 
+new policy as usual in this context, we should not merge policies, but create different `MonitoringAssignment` for each 
+defined policy, this is needed for scenarios like migrating to a new backend. For example, when we define two `MeshMetric` 
+policies that target the same data plane proxy, we should create two separate `MonitoringAssignment` responses. 
+We can achieve this by extending Prometheus integration, so that it sends backend name with requests to MADS API, and we will
+create `MonitoringAssignment` based on `backends[].name` field.
```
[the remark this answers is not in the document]
Automaat: we need to extend prometheus integration to enable user to add id/name to it. This name should be the same as the one provided in `MeshMetric` policy

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,201 @@
...
+The API is going to be similar to what we have under `mesh.spec.metrics`. The current API looks like this:
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: Mesh
+metadata:
+  name: default
+spec:
+  metrics:
+  enabledBackend: prometheus-1
```
lukidzi, line 27: nit: just incorrect indentation

docs/madr/decisions/035-mesh-metric-policy.md
```diff
@@ -0,0 +1,201 @@
...
+1. I think the aggregation that [uses the aggregate name and pod annotations](https://kuma.io/docs/2.5.x/policies/traffic-metrics/#expose-metrics-from-applications) no longer makes sense, as we can use targetRef 
+policies to point to specific DPs. I imagine there would be a default wide policy (possibly `Mesh` scoped) and per app / framework / language overrides.
+2. What I also don’t like about this is that “filters” are under envoy - We could rename it to `sidecar` and introduce 
+the concept of “profiles” like we discussed. Maybe something like this:
+
+```yaml
+sidecar:
+  usedOnly: true # true or false
+  profile: minimal # one of minimal, default, full
+  regex: http2_act.* # only profile or regex can be defined
```
lukidzi, line 59: Just an idea, not sure if useful. Sometimes you want to get `http2_act` and maybe `cx_active` for this you need to create some more complicated regex (if works) but you could maybe do 2 separate call. I feel like doing x separate calls is not the best option but we could verify if more complicated regexes works.

Automaat: I just checked and more complicated regexes with `or` works, for example: `/stats?filter=circuit_breakers|update_success` 

In this mechanism, we are utilizing envoy filtering functionality, and I don't think making additional calls it good way to do it. In the future with dynamic dp config and our producers we can add more advanced filtering that will be easier for user to configure

lahabana (approved):

lukidzi (approved):

tepley:
might not be the right place to comment, but some metrics are much much busier than others:

envoy_cluster_upstream_rq_completed
envoy_cluster_upstream_cx_connect_fail
envoy_cluster_upstream_cx_connect_ms

perhaps having the ability to further filter down specific ones from the larger profile pack could be nice.

Issue #8467, opened 2023-11-27
Create a MADR how to enable new policies


lukidzi:
### Description

We should write how we want to enable new policies since 2.6

PR #8528, opened 2023-12-04
docs(MADR): targetRef policy as a default engine

lukidzi:
### Checklist prior to review

We want to make `targetRef` policies as the new default policies, while still accepting previous mesh as a legacy.

- [X] [Link to relevant issue][1] as well as docs and UI issues
  - closes: https://github.com/kumahq/kuma/issues/8467

> Changelog: skip


docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,104 @@
...
+    // When creating, the default is set to TargetRef.
+    // During retrieval of old settings, distinguishing it as Legacy.
+    Undefined = 0;
+    // A Legacy mode creates old default policies.
+    Legacy = 0;
+    // A TargetRef mode creates new default targetRef policies.
+    TargetRef = 1;
+  }
+
+  // policyEngine implies which policy engine should be used as a default.
```
jakubdyszkiewicz, line 58: Is it only about defaults or is it about which policies are used?

Let's say I've set policyEngine to targetRef. Are old policies still applied?
If so, is "how to migrate to new policies" a part of the current "policy 2.0" effort as well?

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+New configuration would look like:
+
+```yaml
+policyEngine: TargetRef
+```
+
+This approach enables the creation of new default target ref policies without enforcing users to only use new polcies. The control-plane will support both old and new policies, with the default `Mesh` being created using the new policies. New policies will not require `MeshTrafficPermissions`, `MeshHTTPRoute`, or `MeshTCPRoute` to enable traffic across the cluster.
```
lobkovilya, line 69: > New policies will not require `MeshTrafficPermissions`, `MeshHTTPRoute`, or `MeshTCPRoute` to enable traffic across the cluster.

I can understand MeshHTTPRoute/MeshTCPRoute, but why do we want to do this with MeshTrafficPermission? If there are no "permissions" in the cluster, why then the traffic should work? Also, if we have implicit `action: Allow` users are going to create denylists as opposed to allowlist. Allowlists are better for security and for the auto reachable services. 

I'm not sure if this should be discussed in this MADR, but I haven't found a relevant issue.

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,104 @@
...
+
+Potential future enhancements may include introducing another mode, such as `TargetRefOnly`, which disables old policies. This could be implemented in `MeshContext` by excluding existing legacy policies from the context. On the API level, checks could be performed to verify if the `Mesh` permits the creation of legacy resources.
```
lobkovilya, line 71: I don't think we need `TargetRefOnly`, eventually the lack of `policyEngine` field should provide such behavior. Ideally, we want to remove `policyEngine` field when the transition is over.

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,104 @@
...
+
+#### Existing users behaviour
+
+When a user already has a `Mesh` and doesn't define the `policyEngine` field, we treat it as an `Undefined` engine. Only when a user creates a new mesh does the mesh defaulter set an engine to `TargetRef`.
+
+Why do we need `Undefined`?
+
+Protobuf, by default, takes the first value if there is no value provided. Without `Undefined`, we would default to `Legacy`. The issue arises when a user intends to create a new `Mesh` without specifying the `policyEngine` field, resulting in the selection of the first option due to the lack of an optional `policyEngine` field on the protobuf level.
+
+This situation creates problem in discovering between cases where the user did not provide the definition or explicitly chose `Legacy`.
```
lobkovilya, line 81: For meshes that already exist, are we going to react to the `policyEngine` changes? If I change `Legacy` to `TargetRef` do we remove old default policies and create new ones?

docs/madr/decisions/034-default-target-ref-policy.md
lobkovilya: Ah I see, I was thinking we want traffic by default even with strict mTLS. We should make it clear in docs that MTPs must be created before enabling the mTLS, otherwise, no requests are allowed.

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,104 @@
...
+policyEngine: TargetRef
+```
+
+This approach enables the creation of new default target ref policies without enforcing users to only use new polcies. The control-plane will support both old and new policies, with the default `Mesh` being created using the new policies. New policies will not require `MeshTrafficPermissions`, `MeshHTTPRoute`, or `MeshTCPRoute` to enable traffic across the cluster.
+
+Potential future enhancements may include introducing another mode, such as `TargetRefOnly`, which disables old policies. This could be implemented in `MeshContext` by excluding existing legacy policies from the context. On the API level, checks could be performed to verify if the `Mesh` permits the creation of legacy resources.
+
+#### Existing users behaviour
+
+When a user already has a `Mesh` and doesn't define the `policyEngine` field, we treat it as an `Undefined` engine. Only when a user creates a new mesh does the mesh defaulter set an engine to `TargetRef`.
```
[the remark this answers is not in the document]
lukidzi: We could make a default legacy but then we cannot distinguish if a user wants to use a legacy engine. Maybe we would need to call the store to check if mesh exists.

docs/madr/decisions/034-default-target-ref-policy.md
lukidzi: Good question, I think we shouldn't. When you create mesh now with `skipCreatingInitialPolicies` and later remove it you won't get them

docs/madr/decisions/034-default-target-ref-policy.md
bartsmykla: I don't think that's the case. `MeshHTTPRoute` and `MeshTCPRoute` have expectation that traffic flows by default without distinction if mTLS is enabled or not. This is the same behavior as Kubernetes. If you want to change it, you could create an appropriate `MeshTrafficPermission` policy.

docs/madr/decisions/034-default-target-ref-policy.md
jakubdyszkiewicz: If this is only abou initial policies then I think we need more explicit name. We already have `skipCreatingInitialPolicies`. Should we name this `initialPoliciesEngine`?

docs/madr/decisions/034-default-target-ref-policy.md
jakubdyszkiewicz: +1 to we should not do this

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,109 @@
...
+
+```yaml
+policyEngine: TargetRef
+```
+
+This approach allows the creation of new default `TargetRef` policies without mandating users to solely adopt these new policies. To implement this, we need to introduce an environment variable called `KUMA_DEFAULTS_USE_LEGACY_POLICY_ENGINE`. This variable will define the default policy engine when no specific one is provided during the creation or loading process.
+
+#### Existing users behaviour
+
+When a user already possesses a `Mesh` but doesn't define the `policyEngine` field, we consider it as being set to an `Undefined` engine state. The mesh defaulter only assigns an engine based on `KUMA_DEFAULTS_USE_LEGACY_POLICY_ENGINE` when a user creates a new mesh. By default, this variable is set to `false`, meaning each new mesh will use `policyEngine: TargetRef`.
```
jakubdyszkiewicz, line 73: can we change this to
```
KUMA_DEFAULTS_INITIAL_POLICY_ENGINE=TargetRef
```
?

instead of true/false

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,109 @@
...
+Problem:
+What if users update the mesh using CD/Terraform? 
+
+During an upgrade, existing users have the option to set `KUMA_DEFAULTS_USE_LEGACY_POLICY_ENGINE=true`. This setting will result in all existing meshes or newly created ones without the `policyEngine` field being treated as `Legacy`.
+
+#### New kuma users behaviour
+
+When a new user creates a `Mesh` and does not define the `policyEngine` field, it defaults to a `TargetRef` engine.
+
+The control-plane does not generate default `MeshTrafficPermissions`, `MeshHTTPRoute`, or `MeshTCPRoute` because they are not necessary for the traffic to function.
```
jakubdyszkiewicz, line 93: so what policies do we actually generate with new policy engine?

docs/madr/decisions/034-default-target-ref-policy.md
lukidzi: But it's not only about initial policies but also about the flow of the code. In some cases, we need to verify which version you want to use and change the flow

lukidzi: MeshHTTPRoute/MeshTCPRoute we don't need them so we need to remove check if policy exists, but for old engine we don't want to configure it again

jakubdyszkiewicz: I feel like `policyEngine` is a too generic field.

Doc "implies which policy engine should be used as a default" also does not explain much if I'm unfamiliar with the MADR and the discussion.

If I were to guess what will happen, setting this to targetRef would just ignore old policies. But what it does is:
* When mesh is created, create MeshTimeout, MeshRetry, and MeshCircuitBreaker policies instead of old policies
* If TrafficRoute is missing, let the traffic flow
* If TrafficPermission to external is missing (+ no egress with mtls), let the traffic flow

Maybe we need something more explicit
```
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: default
spec:
  initialPoliciesEngine: sourcesDestinations # | targetRef. "sourcesDestinations" by default if missing
  routing:
    allowOnMissingPolicies: false # | true. "false" by default
```

We can document that if you want to only use new policies, create such mesh
```
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: default
spec:
  initialPoliciesEngine: targetRef
  routing:
    allowOnMissingPolicies: true
```

Eventually, in X versions we will remove those fields.

I think introducing `KUMA_DEFAULTS_USE_LEGACY_POLICY_ENGINE` does not add much because you need to set this on every zone and global cp (also multitenant globalcp).
I think setting it in the mesh is more explicit and easier to understand.

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,121 @@
+# TargetRef policies as a default
+
+* Status: accepted
+
+Technical Story: none
```
slonka, line 5:
```suggestion
Technical Story: https://github.com/kumahq/kuma/issues/8467
```

I think we link the issue here but not sure

docs/madr/decisions/034-default-target-ref-policy.md
slonka: side q: wasn't there an idea by @michaelbeaumont that there shouldn't be default policies that could break traffic flow? what was our decision there?

slonka (commented):
In general I feel like we're trying to switch to new policies but we're not really firm about the switch. If we allow options to force old policies we will be stuck with them for longer. Also adding the switch adds more complexity, we need to handle cases like: "what if the user switches back to old policies - etc.".

Could we do it like this:
- for new meshes / new deployments only allow new policies
- for current meshes warn on old policies creation, when we detect user has switched to new policies (no custom s&d policies and default policies are the same as original defaults) disallow creating old policies

docs/madr/decisions/034-default-target-ref-policy.md
lukidzi: I've added a change based on the conversation we had. We will not create any default policy and add upgrade note information for users so they can create them by themselves.

slonka (commented):
Do we prevent users from creating old policies somehow? Also I think we should document all options that we considered, not remove ones that we passed on.

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,135 @@
...
+The legacy policies currently serve as the defaults. However, starting from release 2.6, the intention is to switch the default policy to TargetRef.
+
+## Considered Options
+
+1. Prefer avoiding the creation of default policies and utilize plugin code in cases where there are no existing default legacy policy.
+2. Introduce a `policyEngine` field within the `Mesh` object, allowing the selection of the default policy version.
+
+## Decision Outcome
+
+The decision is to stop creating default policies and utilize plugin code in cases where there are no existing default legacy policy.
```
slonka, line 23: can you go into the specifics of `utilize plugin code` ? do you mean creating envoy resources needed if we detect no default policy? is that easy to do?

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -0,0 +1,135 @@
...
+
+* Clients who are recreating their deployment using continuous delivery (CD) need to manually add default `Timeout`, `CircuitBreaker`, and `Retry` policies
+
+## Pros and Cons of the Options
+
+### Prefer avoiding the creation of default policies and utilize plugin code in cases where there are no existing default legacy policy.
+
+As of the 2.6 release, we have discontinued the creation of default policies when initializing a new mesh. This change in the plugin's code allows us to verify the existence of an older matching policy. In cases where there is no existing policy, we generate the configuration using the plugin code.
+
+To support older versions of zones we are going to introduce `KUMA_DEFAULTS_CREATE_MESH_RESOURCES` which by default is disabled. The user can enable `KUMA_DEFAULTS_CREATE_MESH_RESOURCES` and that allows new meshes to be created with default policies.
```
slonka, line 41: so what's the case for this? someone upgrades global to 2.6 and the zones are on 2.5, the global no longer creates default policies (but I guess it does not remove the existing ones, right?) so nothing changes? then someone updates the zones to 2.6 and still nothing changes 🤔 or are you thinking about someone creating a new mesh while global is on 2.6 and zone is on 2.5 and the new mesh does not work because there are no default policies? 🤔 should we care about that case?

lukidzi: > are you thinking about someone creating a new mesh while global is on 2.6 and zone is on 2.5 and the new mesh does not work because there are no default policies
It's about this one and I think we should care about it.

docs/madr/decisions/034-default-target-ref-policy.md
lukidzi: By `utilize plugin code` i mean pluggable policies (targerRef)

docs/madr/decisions/034-default-target-ref-policy.md
slonka: > a new mesh while global is on 2.6 and zone is on 2.5 and the new mesh does not work because there are no default policies

yeah, you're right because we have hosted global and can't control zone upgrades

lukidzi: Added note

docs/madr/decisions/034-default-target-ref-policy.md
jakubdyszkiewicz: Basically just hardcode XDS config regardless if a resource is present or not, right?

lukidzi: Tried to clarify it a bit, if we don't have an old policy, code related to the targetRef policies generates configuration even if there is no default new policy, which makes everything work. This is the difference to old policies because when there is no default policy means there is no traffic.

docs/madr/decisions/034-default-target-ref-policy.md
jakubdyszkiewicz: I'm a bit sceptical whether we should introduce this env. This can confuse users further what to do.

What are the chances that in the middle of Kuma upgrade someone creates a Mesh, spins up new services and DPPs in the mesh and cannot afford this behavior?

jakubdyszkiewicz (approved):
looks ok besides this env that I'm not sure we really need

docs/madr/decisions/034-default-target-ref-policy.md
bartsmykla: I'm still not entirely clear on this part. I think it would be helpful if you could specify that you're referring specifically to the code that manages traffic flow.

docs/madr/decisions/034-default-target-ref-policy.md
bartsmykla: I agree with @jakubdyszkiewicz that introducing a separate environment variable might introduce unnecessary complexity and confusion, potentially outweighing its marginal benefits. Instead, we should focus on providing clear and concise documentation that guides users through this specific behavior.

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -36,10 +36,14 @@ The decision is to stop creating default policies and utilize plugin code in cas
...
+As of the 2.6 release, we have discontinued the creation of default policies when initializing a new mesh. This change in the plugin's code allows us to verify the existence of an older matching policy. In cases when there is no existing legacy policy, we generate the configuration using the code for targetRef policies.
 
 To support older versions of zones we are going to introduce `KUMA_DEFAULTS_CREATE_MESH_RESOURCES` which by default is disabled. The user can enable `KUMA_DEFAULTS_CREATE_MESH_RESOURCES` and that allows new meshes to be created with default policies.
 
+Why do we need need `KUMA_DEFAULTS_CREATE_MESH_RESOURCES`?
+
+When global is newer than a zone, and a new mesh is created it won't have default policies. Without default policies the older zone won't be able to generate configuration required to route between services. In this case the user can deploy global or federated zone with `KUMA_DEFAULTS_CREATE_MESH_RESOURCES` and create the new mesh that will work with the old zone.
```
slonka, line 45: I thought it was a setting only on the `global` 🤔. Global creates the policies and syncs them to federated zones right?

michaelbeaumont (approved):

docs/madr/decisions/034-default-target-ref-policy.md
lukidzi: What about multitenant deployment when clients might not be at the same versions? In this case, it's easier to allow the creation of old ones policies to make everything works

docs/madr/decisions/034-default-target-ref-policy.md
lukidzi: Yes, global or federated zone(legacy standalone)

docs/madr/decisions/034-default-target-ref-policy.md
```diff
@@ -36,13 +36,13 @@ The decision is to stop creating default policies and utilize plugin code in cas
...
 
-Why do we need need `KUMA_DEFAULTS_CREATE_MESH_RESOURCES`?
+Why do we need need `KUMA_DEFAULTS_CREATE_MESH_ROUTING_RESOURCES`?
 
-When global is newer than a zone, and a new mesh is created it won't have default policies. Without default `TrafficPermission` and `TrafficRoute` policies the older zone won't be able to generate configuration required to route between services. In this case the user can deploy global or federated zone with `KUMA_DEFAULTS_CREATE_MESH_RESOURCES` and create the new mesh that will work with the old zone.
+When global is newer than a zone, and a new mesh is created it won't have default policies. Without default `TrafficPermission` and `TrafficRoute` policies the older zone won't be able to generate configuration required to route between services. In this case the user can deploy global or federated zone with `KUMA_DEFAULTS_CREATE_MESH_ROUTING_RESOURCES` and create the new mesh that will work with the old zone.
```
michaelbeaumont, line 45:
```suggestion
When global is newer than a zone, and a new mesh is created it won't have default `TrafficPermission` and `TrafficRoute` policies, without which the older zone won't be able to generate configuration required to route between services. In this case the user can deploy global or federated zone with `KUMA_DEFAULTS_CREATE_MESH_ROUTING_RESOURCES` and create the new mesh that will work with the old zone.
```

michaelbeaumont (approved):

docs/madr/decisions/034-default-target-ref-policy.md
michaelbeaumont: since it will now still have _some_ default policies

slonka (approved):

PR #8529, opened 2023-12-04
feat(api-server): add /_resources endpoint

lahabana:
Similar to the previous /policies but does it for all resources. This should enable us to more easily write automation around resources that are not limited to policies.

### Checklist prior to review


pkg/core/resources/model/resource.go
```diff
@@ -158,6 +158,8 @@ type ResourceTypeDescriptor struct {
 	Insight Resource
 	// Overview contains the overview type attached to this resourceType
 	Overview Resource
+	// DumpForGlobal whether resources of this type should be dumped when exporting a zone to migrate to global
+	DumpForGlobal bool
```
jakubdyszkiewicz, line 162: I'm not yet sure it will be simple on/off on a resource. I can imagine that for secrets/configs we might need to pick resources, but let's see.

jakubdyszkiewicz (approved):

pkg/core/resources/model/resource.go
lahabana: Ok we may simply drop this field then

Issue #5827, opened 2023-01-25
MeshHTTPRoute to support `Mesh` in `to` targetRef


lobkovilya:
### Description

Currently, MeshHTTPRoute supports only `MeshService` in `to` targetRef, but it shouldn't be a problem to support `Mesh`. That's useful if a user wants to add a default route for all outbound listeners.

lahabana:
We should also think about what happens when there are no routes defined.
With TrafficRoute this meant no traffic was possible.
With MeshHTTPRoute all traffic is possible.

We need further discussion to figure what the right behaviour is.

slonka:
pinging again @lahabana @jakubdyszkiewicz

lahabana:
Curious to see how this would work with multiple policies and merging. Probably worth looking at inspect output in the end.

michaelbeaumont:
Not for sidecars

lahabana:
Example Use cases:

- A user wants to add a header to all requests to any service


to triage: is this just a validation and test issue or there's actually more to be figured here?

lukidzi:
> Triage: it might require only validation change (to allow Mesh) and validate if everything works correctly. Also, we should add E2E test covering it

lahabana:
@lobkovilya is this still actually an issue? I think a lot of things have changed in this area

lobkovilya:
> Triage: the original motivation for the issue was having a default MeshHTTPRoute policy. Eventually, we changed the concept so that traffic flows with no MeshHTTPRoutes. There were no asks for this functionality since, so we're closing the issue.

lahabana:
You can also achieve this with using `MeshService` and using a label that almost everything has.

PR #8540, opened 2023-12-05
feat(api-server): add _rules api to meshgateways

lahabana:
This is useful to get all the policies that apply to a mesh gateway.

### Checklist prior to review

### DIFF
--- a/pkg/api-server/testdata/inspect/meshgateways/_rules/meshhttproute.golden.json
+++ b/pkg/api-server/testdata/inspect/meshgateways/_rules/meshhttproute.golden.json
@@ -1,6 +1,17 @@
 {
  "httpMatches": [
   {
+   "hash": "P6k5V5Y/QDNzTnF1vr7gY33PAAoDt+H+DtspboMfNMs=",
+   "match": [
+    {
+     "path": {
+      "value": "/api/v2",
+      "type": "PathPrefix"
+     }
+    }
+   ]
+  },
+  {
    "hash": "waFsoIISmZDTfFWYqxvY265/GASHYEWvHQTwMh/bpuU=",
    "match": [
     {
@@ -14,8 +25,8 @@
  ],
  "resource": {
   "mesh": "default",
-  "name": "dp-1",
-  "type": "Dataplane"
+  "name": "the-gateway",
+  "type": "MeshGateway"
  },
  "rules": [
   {
@@ -28,7 +39,7 @@
         "matches": [
          {
           "path": {
-           "value": "/api",
+           "value": "/api/v2",
            "type": "PathPrefix"
           }
          }
@@ -44,27 +55,7 @@
           }
          ]
         }
-       }
-      ]
-     },
-     "matchers": [
-      {
-       "key": "kuma.io/service",
-       "not": false,
-       "value": "backend_kuma-demo_svc_3001"
-      }
-     ],
-     "origin": [
-      {
-       "mesh": "default",
-       "name": "the-http-route",
-       "type": "MeshHTTPRoute"
-      }
-     ]
-    },
-    {
-     "conf": {
-      "rules": [
+       },
        {
         "matches": [
          {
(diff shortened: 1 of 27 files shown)


pkg/api-server/testdata/inspect/meshgateways/_rules/meshhttproute.golden.json
```diff
@@ -0,0 +1,162 @@
...
+     "matchers": [
+      {
+       "key": "__rule-matches-hash__",
+       "not": false,
+       "value": "P6k5V5Y/QDNzTnF1vr7gY33PAAoDt+H+DtspboMfNMs="
+      },
+      {
+       "key": "kuma.io/service",
+       "not": false,
+       "value": ""
```
lahabana, line 125: @michaelbeaumont this doesn't really added to me. Feels like this empty service shouldn't exist right? Is this a problem in my resource code or a more general policy matching issue?

pkg/api-server/testdata/inspect/meshgateways/_rules/meshhttproute.golden.json
```diff
@@ -0,0 +1,162 @@
...
+  "name": "the-gateway",
+  "type": "MeshGateway"
+ },
+ "rules": [
+  {
+   "fromRules": [],
+   "toRules": [
+    {
+     "conf": {
+      "Rules": [
```
lahabana, line 37: This `Rules` instead of `rules` is very odd.

pkg/api-server/testdata/inspect/meshgateways/_rules/meshhttproute.golden.json
lahabana: the `on-route` policy:

```yaml
type: MeshTimeout
name: on-route
mesh: default
spec:
  targetRef:
    kind: MeshHTTPRoute
    name: the-http-route
  to:
    - targetRef:
        kind: Mesh
      default:
        http:
          requestTimeout: 15s
```

Interesting that it wants to add a service match in there...

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -301,8 +301,7 @@ func buildToList(p core_model.Resource, httpRoutes []core_model.Resource) ([]cor
 			for _, to := range policyWithTo.GetToList() {
 				rv = append(rv, &artificialPolicyItem{
 					targetRef: common_api.TargetRef{
-						Kind: common_api.MeshServiceSubset,
-						Name: mhrRules.TargetRef.Name,
```
michaelbeaumont, line 309: @lobkovilya This seems like it should depend on the `to` `targetRef`?

lahabana: You mean it should be the service of the `to targetRef` ? Maybe but then what happens when the `to targetRef` is `kind: Mesh` ?

lobkovilya: > Maybe but then what happens when the to targetRef is kind: Mesh ?

We don't support `kind: Mesh` today and we have an open issue https://github.com/kumahq/kuma/issues/5827. 

But even when we add `kind: Mesh` support in the future, we shouldn't lose `kuma.io/service` info, right? So at least it has to be included in `matchesHash`

michaelbeaumont: @lahabana in the interest of getting this change in for MeshGatewayRoute, should we make a note to implement this for MeshHTTPRoute separately?

lahabana: This doesn't support MeshGatewayRoute AFAIK. This adds matching for MeshGateways.

lahabana: Ok reverted this change as it should be handled as part of the MeshGateway support for MeshHTTPRoute

michaelbeaumont (approved):

PR #8547, opened 2023-12-06
ci(actions): stabilize the E2E tests  for running on GH Actions by preventing binding to ipv6 ports when creating test containers

jijiechen:
A few issues were found in E2E tests when they are run on GitHub Actions. These issues were caused by port conflicts between ports between the ipv6 addr and ipv4 addr.  This PR is to fix this problem, and the runs based on this fix have been observed to be stable.

When a test container is being created, the current test will make the container expose addresses on both ipv4 and ipv6 addresses. When the random ports are the same on these addresses, this can mess things up in a few test cases.

Here is an example of the flaky run:
* https://github.com/kumahq/kuma/actions/runs/7037707580/job/19153296059#step:9:14439

The latest version is stable to run on GitHub Actions, here are two runs:
* https://github.com/jijiechen/kuma/actions/runs/7109600243
* https://github.com/jijiechen/kuma/actions/runs/7099208668

### Checklist prior to review

  - N/A
  - Confirmed 
  - N/A
    - Added
  - No need
  - No need


test/framework/docker.go
```diff
@@ -29,7 +29,11 @@ func GetPublishedDockerPorts(
 		if len(addresses) < 1 {
 			return nil, errors.Errorf("there are no addresses for port %d", port)
 		}
-		_, pubPortStr, err := net.SplitHostPort(addresses[0])
+		addr := addresses[0]
+		if strings.HasPrefix(addr, ":::") {
+			addr = "[::]:" + addr[3:]
+		}
+		_, pubPortStr, err := net.SplitHostPort(addr)
```
lahabana, line 36: A little comment to explain this would be nice

test/framework/k8s_cluster.go
```diff
@@ -515,28 +519,42 @@ func (c *K8sCluster) DeployKuma(mode core.CpMode, opt ...KumaDeploymentOption) e
...
-		}
+		appsToInstall = append(appsToInstall, appInstallation{Config.ZoneIngressApp, Config.KumaNamespace, 1})
 	}
-
 	if c.opts.zoneEgress {
-		if err := c.WaitApp(Config.ZoneEgressApp, Config.KumaNamespace, 1); err != nil {
-			return err
-		}
+		appsToInstall = append(appsToInstall, appInstallation{Config.ZoneEgressApp, Config.KumaNamespace, 1})
+	}
```
lahabana, line 532: Good call that should speed things up no?

lahabana (approved):
LGTM just a few nits

test/framework/k8s_cluster.go
jijiechen: Right, should be faster by installing the apps in parallel. A drawback is it makes the log output disordered instead of sequencial.

test/framework/docker.go
jijiechen: Just added and updated the code.
This format is seen on CircleCI runs ([an example run](https://app.circleci.com/pipelines/github/jijiechen/kuma/794/workflows/597fa41c-4fc9-4dc2-be08-cc150b2280cb/jobs/1223))

PR #8552, opened 2023-12-06
ci(actions): pin git SHA to the same value as GH Actions when running e2e tests on CircleCI

jijiechen:
CircleCI failed to load downloaded artifacts if there are new commits pushed in the branch during an ongoing run.

This PR is to pin the SHA that CircleCI workflows run on.

It should fix failures like this: https://app.circleci.com/pipelines/gh/kumahq/kuma/26972/workflows/be3668af-9369-4cf2-81d4-39e61694606a/jobs/566681

### Checklist prior to review

  - N/A
  - Confirmed 
  - N/A
    - Added
  - No need
  - No need

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -747,6 +747,9 @@ workflows:
         equal: ["", << pipeline.parameters.gh_action_build_artifact_name >>]
     jobs:
       - e2e_testing:
+          filters:
+            tags:
+              only: /.*/
           k8sVersion: << pipeline.parameters.e2e_param_k8s_version >>
           target: << pipeline.parameters.e2e_param_target >>
           arch: << pipeline.parameters.e2e_param_arch >>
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -747,6 +747,9 @@ workflows:
         equal: ["", << pipeline.parameters.gh_action_build_artifact_name >>]
     jobs:
       - e2e_testing:
+          filters:
+            tags:
+              only: /.*/
```
lahabana, line 752: Why is this needed?

lahabana (approved):
LGTM!

.circleci/config.yml
jijiechen: Because CircleCI by default will run on branches and not any tags. They treat SHAs the same way as tags. 
Refer to this thread https://discuss.circleci.com/t/v2-api-pass-git-sha-to-trigger-new-pipeline/39811

PR #8582, opened 2023-12-11
chore(kuma-cp): deprecate driverName=postgres (pq)

lobkovilya:
Doesn't make sense to maintain both Postgres drivers, we added `pgx` to use `BeforeAcquire` a while ago, no we can deprecate `pq`.

### Checklist prior to review


pkg/config/plugins/resources/postgres/config.go
```diff
@@ -65,6 +65,7 @@ type PostgresStoreConfig struct {
 	MaxOpenConnections int `json:"maxOpenConnections" envconfig:"kuma_store_postgres_max_open_connections"`
 	// MaxIdleConnections (applied only when driverName=postgres) is the maximum number of connections in the idle connection pool
 	// <0 value means no idle connections and 0 means default max idle connections
+	// Deprecated: it's only used when driverName=postgres (lib/pq) which is deprecated, use driverName=pgx instead.
```
jakubdyszkiewicz, line 68: I think we should add this to kuma-cp default config YAML

jakubdyszkiewicz (approved):

PR #8584, opened 2023-12-11
feat(zone): create Zone resources on zone cp automatically and generate ZoneInsights

jakubdyszkiewicz:
### Checklist prior to review

Create a default Zone resource on Zone CP and generate Zone Insight
This is needed for a GUI so we have a visibility whether the zone is connected to global.

Described in MADR https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/033-deprecate-standalone.md


pkg/config/app/kuma-cp/config.go
```diff
@@ -177,6 +177,13 @@ func (c Config) IsNonFederatedZoneCP() bool {
 	return c.Mode == core.Zone && !c.IsFederatedZoneCP()
 }
 
+func (c Config) ZoneName() string {
+	if c.Multizone.Zone.Name == "" {
+		return core.DefaultZoneName
+	}
+	return c.Multizone.Zone.Name
+}
```
lahabana, line 185: Should you use `Default()` instead?

lobkovilya (approved):

PR #8606, opened 2023-12-12
ci(publish): don't use -i in docker

lahabana:
This was causing issues in CI like GH actions which don't have tty

addresses the failure: https://github.com/kumahq/kuma/actions/runs/7185343654/job/19571263923#step:11:17

### Checklist prior to review

### DIFF
--- a/mk/distribution.mk
+++ b/mk/distribution.mk
@@ -75,7 +75,7 @@ publish/pulp/$(DISTRIBUTION_TARGET_NAME)-$(1)-$(2):
 		-e USE_CLOUDSMITH=x \
 		-e USE_PULP=x \
 		-v $(TOP)/build/distributions/out:/files:ro \
-		-it \
+		-t \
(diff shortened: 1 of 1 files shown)


mk/distribution.mk
```diff
@@ -75,7 +75,7 @@ publish/pulp/$(DISTRIBUTION_TARGET_NAME)-$(1)-$(2):
 		-e USE_CLOUDSMITH=x \
 		-e USE_PULP=x \
 		-v $(TOP)/build/distributions/out:/files:ro \
-		-it \
+		-t \
```
michaelbeaumont, line 78: Is it enough to remove `-i`? That just controls stdin IIUC

lahabana: The error is: "the input device is not a TTY" so isn't that what this problem is?

michaelbeaumont (approved):

Issue #8580, opened 2023-12-11
Extend MADS service to use data from MeshMetric policy


Automaat:
### Description

As described in MeshMetric MADR.

PR #8608, opened 2023-12-13
feat(mads): extend MADS service to use data from MeshMetric policy

slonka:
E2E tests will be added after merging https://github.com/kumahq/kuma/pull/8617

closes https://github.com/kumahq/kuma/issues/8580

To me MADS server seems to have some weird parts I'd like to understand before making refactoring / suggestions:
- we start both http and grpc services, but only http is used to configure prometheus https://github.com/kumahq/kuma/blob/a6d05c1f35bdd06e02947a74da586da7a7c01472/app/kumactl/cmd/install/testdata/install-observability.defaults.golden.yaml#L448 . **Why do we need grpc?**
- `assigment_test.go` and `snapshot_generator_test.go` seem to be covering pretty much the same thing (take meshes and dataplanes and in one case produce `core_xds.Resource` and in another case a snapshot that wraps `map[string]envoy_types.Resource`. Even the test cases are common:
  - "no Meshes, no Dataplanes" - [1](https://github.com/kumahq/kuma/blob/9ed757d3e5955ea57d7940badf4468057ed46663/pkg/mads/v1/reconcile/snapshot_generator_test.go#L68), [2](https://github.com/kumahq/kuma/blob/fc86a9d3ae1fa14fe0a223dc74103a4948f8b83e/pkg/mads/v1/generator/assignments_test.go#L40)
  - "Dataplane without Mesh / Mesh with Prometheus enabled but no Dataplanes" - [1](https://github.com/kumahq/kuma/blob/fc86a9d3ae1fa14fe0a223dc74103a4948f8b83e/pkg/mads/v1/generator/assignments_test.go#L43), [2](https://github.com/kumahq/kuma/blob/9ed757d3e5955ea57d7940badf4468057ed46663/pkg/mads/v1/reconcile/snapshot_generator_test.go#L87)
- using go-control-plane constructs like snapshot, snapshot cache, envoy node (disguised as prometheus client) to manage the state.  
- interfaces with single implementation
  - I don't think we foresee multiple implementations of mads
  - The module is pretty self contained, it's not pulled from multiple places

To me the whole module is just "get all the dataplanes (or the ones that match policies), transform the data to fit prometheus format, send a response" and all the other stuff just makes it a bit harder to understand what's going on.

We could refactor this up now but we probably will be doing some clean up when we deprecate/remove `mesh.metrics` way of configuring prometheus so should it wait for that moment?

So for now I only created a new generator that generates the responses for MeshMetrics and added test coverage for it. After this PR review we can decide if we do something with the current code.

### Checklist prior to review

### DIFF
--- a/pkg/mads/v1/reconcile/snapshot_generator.go
+++ b/pkg/mads/v1/reconcile/snapshot_generator.go
@@ -2,8 +2,8 @@ package reconcile
 
 import (
 	"context"
+
 	envoy_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
-	"github.com/kumahq/kuma/pkg/xds/cache/mesh"
 	"github.com/pkg/errors"
 
 	"github.com/kumahq/kuma/pkg/core"
@@ -17,6 +17,7 @@ import (
 	"github.com/kumahq/kuma/pkg/plugins/policies/core/matchers"
 	"github.com/kumahq/kuma/pkg/plugins/policies/meshmetric/api/v1alpha1"
 	util_xds_v3 "github.com/kumahq/kuma/pkg/util/xds/v3"
+	"github.com/kumahq/kuma/pkg/xds/cache/mesh"
 )
 
 var log = core.Log.WithName("mads").WithName("v1").WithName("reconcile")
@@ -25,7 +26,7 @@ func NewSnapshotGenerator(resourceManager core_manager.ReadOnlyResourceManager, 
 	return &snapshotGenerator{
 		resourceManager:   resourceManager,
 		resourceGenerator: resourceGenerator,
-		meshCache: meshCache,
+		meshCache:         meshCache,
 	}
 }
 
--- a/pkg/mads/v1/reconcile/snapshot_generator.go
+++ b/pkg/mads/v1/reconcile/snapshot_generator.go
@@ -13,7 +13,7 @@ import (
 	core_xds "github.com/kumahq/kuma/pkg/core/xds"
 	"github.com/kumahq/kuma/pkg/mads/generator"
 	mads_v1_cache "github.com/kumahq/kuma/pkg/mads/v1/cache"
-	"github.com/kumahq/kuma/pkg/mads/v1/meshmetrics"
+	meshmetrics_generator "github.com/kumahq/kuma/pkg/mads/v1/generator"
 	"github.com/kumahq/kuma/pkg/plugins/policies/core/matchers"
 	"github.com/kumahq/kuma/pkg/plugins/policies/meshmetric/api/v1alpha1"
 	util_xds_v3 "github.com/kumahq/kuma/pkg/util/xds/v3"
@@ -77,7 +77,7 @@ func (s *snapshotGenerator) GenerateSnapshot(ctx context.Context, node *envoy_co
 			return nil, err
 		}
 
-		resources, err = meshmetrics.Generate(dataplanes)
+		resources, err = meshmetrics_generator.Generate(dataplanes)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/mads/v1/reconcile/snapshot_generator.go
+++ b/pkg/mads/v1/reconcile/snapshot_generator.go
@@ -48,7 +48,7 @@ func (s *snapshotGenerator) GenerateSnapshot(ctx context.Context, node *envoy_co
 	}
 
 	if len(meshes) > 0 && len(meshesWithMeshMetrics) > 0 {
-		log.Info("it is not supported to use both MeshMetrics policy and 'metrics' under Mesh resource. If migrating please remove the 'metrics' section and apply an equivalent MeshMetrics resource")
+		log.Info("it is not supported to use both MeshMetrics policy and 'metrics' under Mesh resource. For now MeshMetrics will take precedence. If migrating please remove the 'metrics' section and apply an equivalent MeshMetrics resource")
 	}
 
 	var resources []*core_xds.Resource
--- a/pkg/mads/v1/reconcile/snapshot_generator.go
+++ b/pkg/mads/v1/reconcile/snapshot_generator.go
@@ -2,6 +2,7 @@ package reconcile
 
 import (
 	"context"
+	"github.com/kumahq/kuma/pkg/xds/cache/mesh"
 
 	envoy_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	"github.com/pkg/errors"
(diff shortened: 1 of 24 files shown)


pkg/mads/v1/reconcile/snapshot_generator.go
```diff
@@ -26,34 +35,97 @@ type snapshotGenerator struct {
...
+func (s *snapshotGenerator) getMatchingDataplanes(ctx context.Context, meshMetrics []*v1alpha1.MeshMetricResource) (map[*v1alpha1.MeshMetricResource]*core_mesh.DataplaneResourceList, error) {
+	meshMetricToDataplanes := map[*v1alpha1.MeshMetricResource]*core_mesh.DataplaneResourceList{}
+
+	for _, meshMetric := range meshMetrics {
+		dataplaneList := &core_mesh.DataplaneResourceList{}
+		filter := func(rs model.Resource) bool {
+			dpp := rs.(*core_mesh.DataplaneResource)
+			res, _ := matchers.PolicyMatches(meshMetric, dpp, xds_context.NewResources()) // what are the dependant resources? do I need them?
+			return res
+		}
```
slonka, line 118: This is similar to what we filter in `insights` we get all dpps and filter by the ones that match the policy. I guess in insights it might be ok because it's on demand. Here we will be doing this on interval (I think it's 15s for prometheus).

pkg/mads/v1/reconcile/snapshot_generator.go
```diff
@@ -26,34 +35,97 @@ type snapshotGenerator struct {
...
+	if err := s.resourceManager.List(ctx, &meshMetricsList); err != nil {
 		return nil, err
 	}
 
-	return mads_v1_cache.NewSnapshot("", core_xds.ResourceList(resources).ToIndex()), nil
+	meshMetrics := make([]*v1alpha1.MeshMetricResource, 0)
+	for _, meshMetric := range meshMetricsList.Items {
+		for _, backend := range *meshMetric.Spec.Default.Backends { // can backends be nil?
+			// match against client ID or fallback to "" when specified by user
+			if backend.Type == v1alpha1.PrometheusBackendType && (*backend.Prometheus.ClientId == clientId || *backend.Prometheus.ClientId == "") {
```
slonka, line 100: this is the matching part that we talked about, if the clientId matches or if it's empty we "match" the policy.

pkg/plugins/policies/core/interfaces.go
```diff
@@ -1 +0,0 @@
-package core
```
slonka, line 1: not related but was just an empty file

pkg/mads/v1/service/components.go
```diff
@@ -87,6 +88,7 @@ func NewXdsContext(log logr.Logger) (envoy_cache.NodeHash, util_xds_v3.SnapshotC
 type hasher struct{}
 
 func (_ hasher) ID(node *envoy_core.Node) string {
-	// in the very first implementation, we don't differentiate clients
-	return ""
+	// now that we start differentiating between clients are we ok with this config growing for old mechanism (under `mesh.metrics`)
```
slonka, line 91: this needs to be addressed, I think it's fine, snapshots have cleanup mechanism and I don't think there's going to be 1000s of prometheus instances

pkg/mads/v1/reconcile/snapshot_generator.go
```diff
@@ -4,56 +4,131 @@ import (
...
-	dataplanes, err := s.getDataplanes(ctx, meshes)
+	meshes, err := s.getMeshesWithPrometheusEnabled(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	meshGateways, err := s.getMeshGateways(ctx, meshes)
-	if err != nil {
-		return nil, err
+	if len(meshes) > 0 && len(meshesWithMeshMetrics) > 0 {
```
Automaat, line 50: Do I understand correctly that if both are configured `MeshMetric` will be used to generate `MonitoringAssignment`? Maybe it is worth to add this info to this log?

slonka: > Maybe it is worth to add this info to this log? Or don't generate MonitoringAssignment at all?

I think it makes sense to add it to logs. I think we should allow a flow from old to new policies with the least interruptions as possible.

pkg/mads/v1/reconcile/snapshot_generator.go
```diff
@@ -4,56 +4,131 @@ import (
...
+func (s *snapshotGenerator) getMeshesWithMeshMetrics(ctx context.Context, clientId string) ([]string, error) {
+	meshMetricsList := v1alpha1.MeshMetricResourceList{}
+	if err := s.resourceManager.List(ctx, &meshMetricsList); err != nil {
 		return nil, err
 	}
 
-	return mads_v1_cache.NewSnapshot("", core_xds.ResourceList(resources).ToIndex()), nil
+	var meshToMeshMetrics []string
+	for _, meshMetric := range meshMetricsList.Items {
+		for _, backend := range *meshMetric.Spec.Default.Backends { // can backends be nil?
```
Automaat, line 99: backends can be `nil` if you have one policy for `Mesh` with backend and seperate policies for `MeshServices` that only change aggregation or sidecar config

pkg/mads/v1/reconcile/snapshot_generator.go
```diff
@@ -4,56 +4,131 @@ import (
...
+	meshes, err := s.getMeshesWithPrometheusEnabled(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	meshGateways, err := s.getMeshGateways(ctx, meshes)
-	if err != nil {
-		return nil, err
+	if len(meshes) > 0 && len(meshesWithMeshMetrics) > 0 {
+		log.Info("it is not supported to use both MeshMetrics policy and 'metrics' under Mesh resource. If migrating please remove the 'metrics' section and apply an equivalent MeshMetrics resource")
```
slonka, line 51:
```suggestion
		log.Info("it is not supported to use both MeshMetrics policy and 'metrics' under Mesh resource. For now MeshMetrics will take precedence. If migrating please remove the 'metrics' section and apply an equivalent MeshMetrics resource")
```

@Automaat WDYT?

Automaat: LGTM

Automaat (approved):

pkg/mads/v1/reconcile/snapshot_generator.go
slonka: added check for it

slonka:
Merging because https://github.com/kumahq/kuma/actions/runs/7261250426/job/19782420377?pr=8608#step:13:9595 is a flake described in https://github.com/kumahq/kuma/issues/8675

PR #8614, opened 2023-12-13
feat(routes): handle routing if there are no TrafficRoutes

michaelbeaumont:
Depends on https://github.com/kumahq/kuma/pull/8602

- [x] don't warn in outbound proxy generator

### Checklist prior to review


pkg/plugins/policies/meshtcproute/plugin/v1alpha1/listeners_backendrefs.go
```diff
@@ -45,12 +17,34 @@ func getBackendRefs(
...
-	return getTCPBackendRefs(toRulesTCP, service)
+	if tcpConf != nil {
+		return tcpConf.Default.BackendRefs
+	}
+	return []common_api.BackendRef{{
+		TargetRef: common_api.TargetRef{
+			Kind: common_api.MeshService,
+			Name: serviceName,
+		},
+		Weight: pointer.To(uint(100)),
```
lukidzi, line 48: Shouldn't be `1`?

lukidzi (approved):

pkg/plugins/policies/meshtcproute/plugin/v1alpha1/listeners_backendrefs.go
michaelbeaumont: I think it doesn't matter since it's the only backend

lukidzi: Seems like there is a nil pointer https://github.com/kumahq/kuma/blob/master/pkg/plugins/policies/meshhttproute/plugin/v1alpha1/gateway.go#L37

PR #8615, opened 2023-12-13
fix(kuma-cp): treat envoy admin errors as 4xx

lobkovilya:
### Checklist prior to review


pkg/envoy/admin/kds_client.go
```diff
@@ -163,3 +165,20 @@ func resNameInZone(nameInGlobal string, k8sStore bool) (string, string, error) {
...
+func (e *KDSTransportError) Error() string {
+	if e.reason == "" {
+		return fmt.Sprintf("could not send %s", e.requestType)
+	} else {
+		return fmt.Sprintf("could not send %s: %s", e.requestType, e.reason)
+	}
+}
+
+func (e *KDSTransportError) Is(err error) bool {
+	return reflect.TypeOf(e) == reflect.TypeOf(err)
```
michaelbeaumont, line 183: Shouldn't this just be like:

```suggestion
	_, ok := target.(*KDSTransportError)
	return ok
```

?

michaelbeaumont (approved):

pkg/envoy/admin/kds_client.go
lobkovilya: All other `Is` implementations are using `reflect`:
```
$ grep -r ") Is(err" .
./pkg/api-server/types/errors.go:func (a *InvalidPageSizeError) Is(err error) bool {
./pkg/core/access/errors.go:func (a *AccessDeniedError) Is(err error) bool {
./pkg/core/resources/store/store.go:func (e *ResourceConflictError) Is(err error) bool {
./pkg/core/resources/store/store.go:func (e *AssertionError) Is(err error) bool {
./pkg/core/resources/store/store.go:func (a *PreconditionError) Is(err error) bool {
./pkg/core/rest/errors/errors.go:func (e *BadRequest) Is(err error) bool {
./pkg/intercp/envoyadmin/forwarding_kds_client.go:func (e *StreamNotConnectedError) Is(err error) bool {

$ grep -r "reflect.TypeOf(.*) == reflect.TypeOf(err)" .
./pkg/api-server/types/errors.go:       return reflect.TypeOf(a) == reflect.TypeOf(err)
./pkg/core/access/errors.go:    return reflect.TypeOf(a) == reflect.TypeOf(err)
./pkg/core/resources/store/store.go:    return reflect.TypeOf(e) == reflect.TypeOf(err)
./pkg/core/resources/store/store.go:    return reflect.TypeOf(e) == reflect.TypeOf(err)
./pkg/core/resources/store/store.go:    return reflect.TypeOf(a) == reflect.TypeOf(err)
./pkg/core/rest/errors/errors.go:       return reflect.TypeOf(e) == reflect.TypeOf(err)
./pkg/intercp/envoyadmin/forwarding_kds_client.go:      return reflect.TypeOf(e) == reflect.TypeOf(err)

```

If we want to change this I can open another PR that updates all the places

lahabana: Is there much a difference?

pkg/core/rest/errors/error_handler.go
```diff
@@ -145,6 +146,12 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 			Title:  title,
 			Detail: err.Error(),
 		}
+	case errors.Is(err, &admin.KDSTransportError{}):
+		kumaErr = &types.Error{
+			Status: 400,
```
lahabana, line 151: Is 400 really the right status here? There's no user error afaiu. I think if we want to use 4xx we should use another status code (potentially a non standard one)

lahabana:
@lobkovilya what's missing to get this in?

pkg/core/rest/errors/error_handler.go
jakubdyszkiewicz: I'm fine either with 400 or with non-standard 4xx error. It both indicates user-facing error which is technically not correct

lobkovilya: It's tricky to pick a good status code here, but I was thinking "who can do something to resolve the problem" and in this case, it's a user. So it's a bit of a stretch, but I think it's acceptable to treat it as a user error

pkg/envoy/admin/kds_client.go
lobkovilya: Avoiding reflections (if possible) is a good approach in general. But I don't think we should change it in this PR.

pkg/core/rest/errors/error_handler.go
michaelbeaumont: Do we also need to consider how the GUI responds to this error?

lobkovilya: Not really, it's just going to display the error text, we already have a similar 400 error for StreamNotConnected https://github.com/kumahq/kuma/blob/9146696f3a05a8c0bc70c0b711a1b1ac8e75914e/pkg/core/rest/errors/error_handler.go#L131

jakubdyszkiewicz (approved):

Issue #8581, opened 2023-12-11
Implement MeshMetric xds


Automaat:
### Description

Implement MeshMetric plugin xds

PR #8617, opened 2023-12-13
feat(policy): implement MeshMetric xds

Automaat:
Fixes: #8581

### Checklist prior to review


pkg/core/managers/apis/dataplane/additional_networking.go
```diff
@@ -19,6 +19,8 @@ func AdditionalInbounds(dataplane *core_mesh.DataplaneResource, mesh *core_mesh.
 	return inbounds, nil
 }
 
+// TODO: is this even used?
```
slonka, line 22: AFAIK this is used to generate an inbound that configures the listener for `PrometheusBackend`. From the code lookup we also use this to match against MeshTrafficPermission MeshFaultInjection etc

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -27,5 +31,76 @@ func (p plugin) MatchedPolicies(dataplane *core_mesh.DataplaneResource, resource
 
 func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *core_xds.Proxy) error {
 	log.V(1).Info("apply is not implemented")
```
slonka, line 33:
```suggestion
```

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -27,5 +31,76 @@ func (p plugin) MatchedPolicies(dataplane *core_mesh.DataplaneResource, resource
...
+	policies, ok := proxy.Policies.Dynamic[api.MeshMetricType]
+	if !ok || len(policies.SingleItemRules.Rules) == 0 {
+		return nil
+	}
+
+	conf := policies.SingleItemRules.Rules[0].Conf.(api.Conf)
```
slonka, line 39: don't we need to call "Compute" or something like this on it?

pkg/core/managers/apis/dataplane/additional_networking.go
Automaat: I assume that at this moment this policies won't take it into account and we need to implement this separately, or extend gathering [listeners](https://github.com/kumahq/kuma/blob/master/pkg/plugins/policies/core/xds/listeners.go) and [clusters](https://github.com/kumahq/kuma/blob/master/pkg/plugins/policies/core/xds/clusters.go) so it extracts Prometheus resources

Automaat: I think we don't need that at the moment, we can add this later on when needed

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
Automaat: since it is `SingleItemRule` we don't need to run `Compute` because there will be only one rule

slonka: xrel: https://github.com/kumahq/kuma/issues/8681

pkg/plugins/policies/meshmetric/plugin/xds/prometheus_configurer.go
```diff
@@ -0,0 +1,130 @@
...
+			return nil, err
+		}
+	} else {
+		listener, err = pc.unsecuredListener(proxy)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return listener, nil
```
slonka, line 56: will that work?

```suggestion
	switch {
	case pc.useMeshMtls(mesh):
		listener, err = pc.mtlsListener(proxy, mesh)
	case pc.useProvidedTls(proxy.Metadata):
		listener, err = pc.providedTlsListener(proxy)
	default:
		listener, err = pc.unsecuredListener(proxy)
	}

	return listener, err
```

pkg/plugins/policies/meshmetric/plugin/xds/prometheus_configurer.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (pc *PrometheusConfigurer) useMeshMtls(mesh *core_mesh.MeshResource) bool {
+	return pc.Backend.Tls != nil && pc.Backend.Tls.Mode == api.ActiveMTLSBackend && mesh.MTLSEnabled()
```
slonka, line 60: I think we talked about not having `ActiveMTLSBackend`, right?

slonka (commented):
I think you did not remove everything that we agreed on (activeMTLSBackend)

pkg/plugins/policies/meshmetric/plugin/xds/prometheus_configurer.go
Automaat: yes, I didn't have time to remove it, now it is gone

slonka (approved):
I think it's alright now.

Issue #6592, opened 2023-04-21
Make new `targetRef` default


lahabana:
### Description

New mesh should be created with default policies using targetRef style.

There should still be a flag for having the previous behaviour on (so that folks with existing automation don't have to change anything).

xref:

- #6588

johnharris85:
Should this be configurable at the mesh level or at the CP level?

lahabana:
I think it's at the CP level. I don't think it's worth having this in the Mesh.

I think we'd just tell users to set this if they don't want to move on new policies yet and this flag is only read when a mesh is created

PR #8629, opened 2023-12-14
feat(kuma-cp): don't create default policies by default and use targetRef

lukidzi:
### Checklist prior to review

Why is KUMA_DEFAULTS_CREATE_MESH_ROUTING_RESOURCES necessary? 
- In situations where `Global` is newer than the zone, the zone might not be able to route traffic due to the absence of default policies. To facilitate an easier migration process, users can modify the default behavior exclusively within the global or zone (federated) settings.

Most of the changes are test files. There is an issue with gateway timeouts. Previously, the gateway code was setting specific timeouts for the gateway but since we created the default MeshTimeout policy it overrides the defaults. So far we cannot create a default policy just for gateways so we might need to document how the policy should look like and what good timeouts are for the gateway.

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/6592


pkg/plugins/policies/meshhttproute/plugin/v1alpha1/plugin.go
```diff
@@ -119,6 +123,9 @@ func ApplyToGateway(
 	rules []ToRouteRule,
 ) error {
 	var limits []plugin_gateway.RuntimeResoureLimitListener
+	if len(rules) == 0 {
+		return nil
+	}
 
 	for _, info := range plugin_gateway.ExtractGatewayListeners(proxy) {
```
lukidzi, line 130: This function `ExtractGatewayListeners` reuses some old part of gateway code which has services filtered by the traffic permissions

test/e2e_env/universal/grpc/grpc.go
```diff
@@ -62,7 +62,7 @@ func GRPC() {
 		}, "30s", "1s").Should(Succeed())
 	})
 
-	It("MeshHTTPRoute DOES NOT split the traffic between two gRPC services", func() {
+	It("MeshHTTPRoute does split the traffic between two gRPC services", func() {
```
slonka, line 65: why is this now the case? why is default policy creation tied to MeshHTTPRoute splitting gRPC traffic?

test/e2e_env/universal/grpc/grpc.go
```diff
@@ -96,9 +96,8 @@ spec:
 				universal.Cluster, "second-test-server", "http://localhost:9901/stats?format=prometheus",
 			)
 			g.Expect(err).ToNot(HaveOccurred())
-			// todo switch .ToNot to .To when https://github.com/kumahq/kuma/issues/3325 is implemented
```
slonka, line 99: same question as above

slonka (commented):
Added some comments and have a question about this:

> Why is KUMA_DEFAULTS_CREATE_MESH_RESOURCES necessary?
>
> In situations where Global is never set as a zone, the zone might not be able to route traffic due to the absence of default policies. To facilitate an easier migration process, users can modify the default behavior exclusively within the global or zone (federated) settings.

`In situations where Global is never set as a zone` -> what do you mean by this? you mean a non federated zone?

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go
```diff
@@ -90,6 +89,12 @@ func generateListeners(
 		filterChainBuilder.
 			Configure(envoy_listeners.AddFilterChainConfigurer(outboundRouteConfigurer))
 
+		// TODO: https://github.com/kumahq/kuma/issues/3325
```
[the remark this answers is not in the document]
michaelbeaumont: Agree, this should really be a different PR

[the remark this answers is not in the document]
lukidzi: I agree but currently, we don't have the policy `MeshGRPCRoute` and without this change we won't be able to fully switch to new policies and still partially relies on TrafficRoute for grpc traffic.

pkg/defaults/mesh/mesh.go
```diff
@@ -52,44 +53,45 @@ func EnsureDefaultMeshResources(
...
-	}
-
-	defaultResourceBuilders := map[string]func() model.Resource{
-		"allow-all":           defaultTrafficPermissionResource,
-		"route-all":           defaultTrafficRouteResource,
-		"timeout-all":         DefaultTimeoutResource,
-		"circuit-breaker-all": defaultCircuitBreakerResource,
-		"retry-all":           defaultRetryResource,
-	}
+	if createMeshDefaultResources {
```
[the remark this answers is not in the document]
lukidzi: I think that is correct because that is a new behavior that doesn't require default policies.

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go
```diff
@@ -167,19 +171,14 @@ func prepareRoutes(
...
-				}
-			}
 			route.BackendRefs = []common_api.BackendRef{
 				{
-					TargetRef: targetRef,
-					Weight:    pointer.To(uint(100)),
+					TargetRef: common_api.TargetRef{
+						Kind: common_api.MeshService,
+						Name: serviceName,
+						Tags: tags,
```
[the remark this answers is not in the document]
lukidzi: Previously, we only required the `mesh` tag. However, to support virtual outbound, we now require additional tags.

test/e2e_env/universal/grpc/grpc.go
lukidzi: To fully switch to targetRef policies we need grpc support and because of the lack of MeshGRPCRoute we can currently support them with MeshHTTPRoute

pkg/defaults/mesh/mesh_test.go
```diff
@@ -30,9 +30,43 @@ var _ = Describe("EnsureDefaultMeshResources", func() {
...
+		// when
+		err := mesh.EnsureDefaultMeshResources(context.Background(), resManager, defaultMesh, []string{}, context.Background(), false)
+		Expect(err).ToNot(HaveOccurred())
+
+		// and Dataplane Token Signing Key for the mesh exists
+		err = resManager.Get(context.Background(), system.NewSecretResource(), core_store.GetBy(tokens.SigningKeyResourceKey(issuer.DataplaneTokenSigningKeyPrefix(model.DefaultMesh), tokens.DefaultKeyID, model.DefaultMesh)))
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("should not create default policies", func() {
```
michaelbeaumont, line 43: kind of a nit but IMO the test case could be named better, like reading this `It` and the next `It` are just confusing together since they're opposite. I guess this one should be "should not create... if resources are already present"?

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/gateway.go
```diff
@@ -37,7 +37,12 @@ func generateGatewayListeners(
 	if info.Listener.CrossMesh {
 		protocol = mesh_proto.MeshGateway_Listener_HTTPS
 	}
-	res, filterChainBuilders, err := FilterChainGenerators[protocol].Generate(ctx, info, gatewayHosts)
+	filterGen, found := FilterChainGenerators[protocol]
+	if !found {
```
michaelbeaumont, line 41: Is this the gRPC thing or why is this necessary?

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/gateway_routes.go
```diff
@@ -17,7 +17,9 @@ import (
 
 func GenerateEnvoyRouteEntries(host plugin_gateway.GatewayHost, toRules []ToRouteRule) []route.Entry {
 	var entries []route.Entry
-
+	if len(toRules) == 0 {
```
michaelbeaumont, line 20: I tihnk this can be removed? The check already happens in `plugin.go` now

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go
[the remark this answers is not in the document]
michaelbeaumont: Can we please make different PRs for this in the future, it will now be significantly harder to figure out why this change was made after the fact.

pkg/plugins/runtime/gateway/gateway_route_generator_test.go
```diff
@@ -1606,13 +1597,7 @@ conf:
       - destination:
           kuma.io/service: echo-service
 `,
-			WithoutResource{
-				Resource: core_mesh.TimeoutType,
-				Mesh:     "default",
-				Name:     "timeout-all-default",
```
michaelbeaumont, line 1612: How exactly should these tests be changing, don't we kind of need both cases? So for example here we now need to _add_ `Timeout` resources?

pkg/plugins/runtime/gateway/testdata/http/no-timeout.yaml
```diff
@@ -25,7 +25,7 @@ Clusters:
         envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
           '@type': type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
           commonHttpProtocolOptions:
-            idleTimeout: 0s
+            idleTimeout: 3600s
```
michaelbeaumont, line 28: Why is this changing?

lukidzi:
> Added some comments and have a question about this:
> 
> > Why is KUMA_DEFAULTS_CREATE_MESH_RESOURCES necessary?
> > In situations where Global is never set as a zone, the zone might not be able to route traffic due to the absence of default policies. To facilitate an easier migration process, users can modify the default behavior exclusively within the global or zone (federated) settings.
> 
> `In situations where Global is never set as a zone` -> what do you mean by this? you mean a non federated zone?

I fixed the comment i meant newer than zone

pkg/plugins/policies/meshtcproute/plugin/v1alpha1/listeners.go
```diff
@@ -27,6 +27,9 @@ func generateListeners(
 	// For one outbound we pick one traffic route, so LB and Timeout are
 	// the same.
 	clusterCache := map[common_api.TargetRefHash]string{}
+	if proxy == nil || proxy.Dataplane == nil || proxy.Dataplane.Spec.GetNetworking() == nil {
```
michaelbeaumont, line 30: when can `networking` be `nil`?

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/gateway.go
lukidzi: Yes, I had a nil pointer in case of unknown protocol

pkg/plugins/policies/meshtcproute/plugin/v1alpha1/listeners.go
lukidzi: Only in case of Egress/Ingress

pkg/plugins/runtime/gateway/gateway_route_generator_test.go
lukidzi: Reverted change, I am going to switch in the next PR and add default policy

pkg/plugins/runtime/gateway/testdata/http/no-timeout.yaml
lukidzi: This value changes because the gateway code is setting a different timeout. Because MeshTimeout is a new default it overrides the timeouts set by gateway code. There is no way (yet) to create only a policy that targets all gateways

pkg/config/app/kuma-cp/config.go
```diff
@@ -380,6 +380,13 @@ func DefaultGeneralConfig() *GeneralConfig {
 	}
 }
 
+func DefaultDefaultsConfig() *Defaults {
+	return &Defaults{
+		SkipMeshCreation:    false,
+		CreateMeshResources: true,
```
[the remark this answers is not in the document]
lukidzi: In the 2nd PR because might change some more tests

lukidzi:
[closing in favor of ](https://github.com/kumahq/kuma/pull/8920)

PR #8631, opened 2023-12-15
ci(.github): improve ci caching by including dynamic go-deps version in deps lock

jijiechen:
This PR changes the calculation of dependency lock to include the dynamic versions get from the `go-deps.sh` file.
In addition, the PR also differentiates caches between modes of run-full-matrix or not.

### Checklist prior to review

  - N/A
  - Confirmed, the GH Actions yaml is compatible with current and vnext of KM sync script.
- [X] Tests (Unit test, E2E tests, manual test on universal and k8s)
  - Manual tested.
    - Added
  - No need
  - No need.


.github/workflows/build-test-distribute.yaml
```diff
@@ -24,20 +28,21 @@ jobs:
...
-          key: ${{ runner.os }}-devtools-${{ hashFiles('mk/dependencies/deps.lock') }}
+          key: devtools-${{ env.RUN_FULL_MATRIX && 'full-matrix' || 'normal' }}-${{ runner.os }}-${{ hashFiles('mk/dependencies/deps.lock') }}
           restore-keys: |
-            ${{ runner.os }}-devtools
+            devtools-${{ runner.os }}-${{ env.RUN_FULL_MATRIX && 'full-matrix' || 'normal' }}
+            devtools-${{ runner.os }}
       - run: |
           make dev/tools
-      - run: |
-          make clean
```
lahabana, line 33: Still need this to do a proper gen check

.github/workflows/build-test-distribute.yaml
```diff
@@ -24,20 +28,21 @@ jobs:
         with:
           path: |
             ${{ env.CI_TOOLS_DIR }}
-          key: ${{ runner.os }}-devtools-${{ hashFiles('mk/dependencies/deps.lock') }}
+          key: devtools-${{ env.RUN_FULL_MATRIX && 'full-matrix' || 'normal' }}-${{ runner.os }}-${{ hashFiles('mk/dependencies/deps.lock') }}
```
lahabana, line 31: Actually dev-tools is the same whether we do full matrix or not. It's only the code cache (from setup-go)

.github/workflows/build-test-distribute.yaml
jijiechen: Got it, I'll revert this change. 
I saw a failure caused by missing a zz_generated file in KM repo. I fixed it by removing `clean`. Now I think the correct fix should be fixing the installation of `controller-gen` tool so that it installs the tool correctly and generate the files as expected.

.github/workflows/build-test-distribute.yaml
jijiechen: Got it, just updated code to implement it.

lahabana (approved):

PR #8634, opened 2023-12-15
feat(bootstrap): support customizing corefile template from kuma-cp

jijiechen:
### Checklist prior to review

  - https://github.com/kumahq/kuma/issues/8119
  - Should not block as it's adding a new feature.
  - Added
    - Should run normally
  - No, it's compatible with old versions
  - No backport needed


lahabana:
No need for the label :)

app/kuma-dp/cmd/run_test.go
```diff
@@ -116,35 +121,22 @@ var _ = Describe("run", func() {
...
-			Expect(actualArgs[1]).To(Equal("--config-path"))
-			actualConfigFile := actualArgs[2]
-			Expect(actualConfigFile).To(BeARegularFile())
-
-			// then
-			if given.expectedFile != "" {
-				Expect(actualArgs[2]).To(Equal(given.expectedFile))
-			}
+			corednsPid := verifyComponentProcess("coredns", corednsPidFile, corednsCmdlineFile, func(actualArgs []string) {
+				Expect(len(actualArgs)).To(HaveLen(3))
```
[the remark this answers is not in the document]
slonka: you need to pass array to `HaveLen`

app/kuma-dp/cmd/run.go
```diff
@@ -226,6 +203,37 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
 			}
 			opts.AdminPort = bootstrap.GetAdmin().GetAddress().GetSocketAddress().GetPortValue()
 
+			if cfg.DNS.Enabled && !cfg.Dataplane.IsZoneProxy() {
+				if cfg.DNS.CoreDNSConfigTemplatePath == "" && len(kumaSidecarConfiguration.Networking.CorefileTemplate) > 0 {
+					templateFile, err := writeTempFile("kuma-dp-corefile-template-", "Corefile", kumaSidecarConfiguration.Networking.CorefileTemplate, 0o644)
+					if err != nil {
+						return errors.Errorf("Failed to write corefile template to disk. %v", err)
```
jakubdyszkiewicz, line 210: this is "nit", but errors shoudl be lowercase `failed to write...`. Check all errors in this PR for it.

jakubdyszkiewicz (approved):
Nice contrib!

Can you also add this new setting to `pkg/config/app/kuma-cp/kuma-cp.defaults.yaml` ?

app/kuma-dp/cmd/run.go
jijiechen: Thanks, will update code in an upcoming commit.

jijiechen:
I'm still working on passing the E2E tests

jijiechen:
The E2E tests are passing.

test/e2e_env/kubernetes/bootstrap/corefile_template.go
```diff
@@ -0,0 +1,84 @@
+package bootstrap
```
jakubdyszkiewicz, line 1: The test is nice, but I don't think this should go into `e2e_env/kubernetes`.

The idea of `test/e2e_env/kubernetes` is to be able to parallelize tests. This test breaks this promise. Yes, we use Kuma2 here, but imagine we also have another test that needs to deploy Kuma with a custom config using Kuma2. It cannot be then used in parallel with this test.

Another disadvantage is that we need to run a second Kube cluster just for this test. When developing a kube e2e test, usually you focus the suite you are writing and run `make test/e2e-kubernetes`. Forcing running a second kube cluster every single time prolongs the development loop.

I'd move this to `test/e2e`.

jijiechen: Yeah, that'll make more sense, you may go ahead moving it and revert the changes in `e2e.mk`. Thanks a lot.
I wasn't sure if it's good to place it under `e2e_env/kubernetes`, but I felt like it's also not very suitable to but in `test/e2e ` when compared to existing tests.

jijiechen: A new PR has been created to conduct this change, please review:
https://github.com/kumahq/kuma/pull/8736

PR #8635, opened 2023-12-15
ci(github): fix make dev/tools

lahabana:
Calling make dev/tools multiple times would fail

### Checklist prior to review


mk/dependencies/go-deps.sh
```diff
@@ -12,6 +12,7 @@ PGKUMADOC=github.com/kumahq/protoc-gen-kumadoc@$(go list -f '{{.Version}}' -m gi
 GINKGO=github.com/onsi/ginkgo/v2/ginkgo@$(go list -f '{{.Version}}' -m github.com/onsi/ginkgo/v2)
 CONTROLLER_GEN=sigs.k8s.io/controller-tools/cmd/controller-gen@$(go list -f '{{.Version}}' -m sigs.k8s.io/controller-tools)
 
+> mk/dependencies/go-deps.versions
```
michaelbeaumont, line 15: Uhhh what does this mean? :grinning:

jijiechen: This statement resets the file content to empty, identical to the following:
```
echo '' > mk/dependencies/go-deps.versions
```

jijiechen:
Let me add some background to this PR:
This PR is a fix to a [previous PR](https://github.com/kumahq/kuma/pull/8631) that tries to improve the cache in the CI. 
The PR collects all versions of the go tools involved and save it to the file `mk/dependencies/go-deps.versions` and using its content as part of the cache key. 

However, there is a bug in it: when executed multiple times, the file will contain duplicated entries, and will potentially lead to  unexpected cache misses.
So the fix here is to reset the file content every time before the installation to mitgrate the duplication.

mk/dependencies/go-deps.sh
jijiechen, line 15:
```suggestion
echo '' > mk/dependencies/go-deps.versions
```

jijiechen: This is to pass the shell check, so fixes
https://github.com/kumahq/kuma/actions/runs/7221117859/job/19675420378?pr=8635

Issue #8639, opened 2023-12-15
Remove completely EmptyDNSPort


jakubdyszkiewicz:
### Description

Remove EmptyDNSPort since it's not used, but it's useful for rollbacks

Icarus9913:
I take it

PR #8640, opened 2023-12-15
chore(dns): drop usage of empty dns port

jakubdyszkiewicz:
### Checklist prior to review

If I remember correctly (pre-MADR days) there was no way to configure Envoy's DNSFilter to not forward DNS requests if they were not on the predefined list (our VIPs). That's why we had to configure Envoy to fallback on extra port in CoreDNS that always respond with NXDOMAIN.

This behavior seems has changed.
https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/udp_filters/dns_filter
"Users can disable external DNS resolution by omitting the client configuration object."

I confirmed manually that it's the case.

I hold off on completely removing all occurrences of EmptyDNSPort across the codebase so we can support rollback of CP.
`dns_generator.go` in stable 2.5.1 has if on `if dnsPort == 0 || emptyDnsPort == 0 {`.
Let say we remove emptyDnsPort. If user deploys 2.6.0, then all DPPs and then they want to rollback starting from CP. The moment CP is on 2.5.0 and DPPs on 2.6.0 we would not generate DNS configuration which could cause interruption in the traffic.

Issue to remove this completely https://github.com/kumahq/kuma/issues/8639

This simplifies our DNS flow with Kuma.


pkg/xds/envoy/listeners/v3/dns_configurer.go
```diff
@@ -52,32 +47,8 @@ func (c *DNSConfigurer) dnsFilter() *envoy_dns.DnsFilterConfig {
...
 	return &envoy_dns.DnsFilterConfig{
 		StatPrefix: "kuma_dns",
-		ClientConfig: &envoy_dns.DnsFilterConfig_ClientContextConfig{
-			// We configure upstream resolver to resolver that always returns that it could not find the domain (NXDOMAIN)
-			// As for this moment there is no setting to disable upstream resolving.
-			TypedDnsResolverConfig: &envoy_core.TypedExtensionConfig{
-				Name:        "envoy.network.dns_resolver.cares",
-				TypedConfig: util_proto.MustMarshalAny(&resolverConfig),
-			},
-			MaxPendingLookups: 256,
```
michaelbeaumont, line 79: I guess we're intentionally taking the default value?

michaelbeaumont (approved):

pkg/xds/envoy/listeners/v3/dns_configurer.go
jakubdyszkiewicz: yes. Default value is "do not pass this to external DNS"

PR #8654, opened 2023-12-18
docs(MADR): internal listener naming

Automaat:
### Checklist prior to review

### DIFF
--- a/docs/madr/decisions/036-internal-listeners.md
+++ b/docs/madr/decisions/036-internal-listeners.md
@@ -1,0 +1,15 @@
+# Internal listeners naming convention
+* status: accepted
+
+## Context and Problem statement
+At the moment we don't clearly distinguish between listeners used for service to service communication
+and internal listeners, like Prometheus listeners.
+
+## Decision Outcome
+
+We will add prefix `_` to internal listeners.
+
+## Solution
+At the moments internal listeners names start with `kuma` prefix, problem is that service can also have this prefix.
+To be able to easily distinguish between service listeners and internal listeners, we can
+add `_` prefix to listener name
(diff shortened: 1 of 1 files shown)


lahabana (commented):
Should this be more generally applicable to all XDS resources?

docs/madr/decisions/036-internal-listeners.md
```diff
@@ -0,0 +1,15 @@
+# Internal listeners naming convention
+* status: accepted
+
+## Context and Problem statement
+At the moment we don't clearly distinguish between listeners used for service to service communication
+and internal listeners, like Prometheus listeners.
+
+## Decision Outcome
+
+We will add prefix `_` to internal listeners.
```
slonka, line 10: what are the implications of this? are there any downsides? for example we have metrics based on `envoy_cluster_name` so changing cluster names would influence metrics/dashboards/queries - for the listeners I only found https://github.com/kumahq/kuma/blob/7061107e586ef6b7dcf1eba453a76974addafef2/app/kumactl/data/install/k8s/metrics/grafana/kuma-dataplane.json#L1064 . Are there other things that we should look for?

slonka (commented):
If there are no negative consequences then I'll 👍 this, I'm just missing a bit of investigation on how / if this can negatively impact other parts of the system.

docs/madr/decisions/036-internal-listeners.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Solution
+At the moments internal listeners names start with `kuma` prefix, problem is that service can also have this prefix.
```
lukidzi, line 13: Could we add a note about a migration? Users might have dashboards that rely on cluster names and that would break it after an upgrade. Should we add a configuration option to turn off this behavior?

docs/madr/decisions/036-internal-listeners.md
Automaat: I've checked and didn't found any downsides of this, also we use this in MeshMetric right now, and no one found any issues with it

docs/madr/decisions/036-internal-listeners.md
Automaat: added section about updating UPGRADE.md after changing naming of already existing resources

slonka (approved):

lukidzi (approved):

PR #8663, opened 2023-12-18
feat(MeshHTTPRoute): support hostnames with MeshGateway

michaelbeaumont:
### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- #8142 


pkg/plugins/policies/core/rules/merge.go
```diff
@@ -10,57 +10,71 @@ import (
...
 
 	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
+	"github.com/kumahq/kuma/pkg/util/pointer"
 )
 
 // indicates that all slices that start with this prefix will be appended, not replaced
 const appendSlicesPrefix = "Append"
 
-func MergeConfs(confs []interface{}) (interface{}, error) {
+func MergeConfs(confs []interface{}) ([]interface{}, error) {
```
lobkovilya, line 19: Please add a doc comment for this function, we have to explain `[]interface{}` return value

pkg/plugins/policies/core/rules/merge.go
```diff
@@ -10,57 +10,71 @@ import (
...
-	}
 
-	confType := reflect.TypeOf(confs[0])
-	result, err := newConf(confType)
-	if err != nil {
-		return nil, err
-	}
+		if err := json.Unmarshal(resultBytes, result); err != nil {
+			return nil, err
+		}
```
lobkovilya, line 61: I think this code can be moved to a separate private function:
```go
// mergeJSONPatches merges a list of confs to a single conf using the algorithm described in https://www.rfc-editor.org/rfc/rfc7396
func mergeJSONPatches(confs []interface{}) (interface{}, error) {
```
that's only for readability reasons

pkg/plugins/policies/meshhttproute/api/v1alpha1/meshhttproute.go
```diff
@@ -29,6 +29,10 @@ type MeshHTTPRoute struct {
 }
 
 type To struct {
+	// Hostnames limits the effects of the rules to requests to this hostname.
+	// Given hostnames must intersect with the hostname of the listeners the
```
lobkovilya, line 33: I'd expand the comment mentioning that by "listener" we mean both the outbound listener (with implicit "*" hostname) and meshgateway listener

lobkovilya (commented):
Do we already have an e2e test for MeshHTTPRoute attached to MeshGateway?

pkg/plugins/policies/meshhttproute/api/v1alpha1/meshhttproute.go
michaelbeaumont: clarified about MeshGateway only in the doc string

lobkovilya (approved):
LGTM

PR #8703, opened 2023-12-20
fix(kds): avoid rare cases where onStreamClosed is called with no state

lahabana:
There were rare cases where state is not in the streams map. We simply avoid a nil pointer by returning early in this case. There shouldn't be any problems with this as the entry is not in the map anyway

However, maybe there's a race where streamClosed is called before streamOpen

### Checklist prior to review

### DIFF
--- a/pkg/kds/v2/server/status_tracker.go
+++ b/pkg/kds/v2/server/status_tracker.go
@@ -99,6 +99,7 @@ func (c *statusTracker) OnDeltaStreamClosed(streamID int64, _ *envoy_core.Node) 
 
 	state := c.streams[streamID]
 	if state == nil {
+		c.log.Info("[WARNING] OnDeltaStreamClosed but no state in the status_tracker", "streamid", streamID)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


Automaat (approved):

pkg/kds/v2/server/status_tracker.go
```diff
@@ -98,6 +98,9 @@ func (c *statusTracker) OnDeltaStreamClosed(streamID int64, _ *envoy_core.Node)
 	defer c.mu.Unlock()
 
 	state := c.streams[streamID]
+	if state == nil {
```
slonka, line 101: can the same thing happen in dataplane status tracker https://github.com/kumahq/kuma/blob/f80621987afd7e01fb1ff3b65757014a4fbebde5/pkg/xds/server/callbacks/dataplane_status_tracker.go#L94 ? if so should we also add a guard there?

slonka (approved):
Makes sense, only one question

pkg/kds/v2/server/status_tracker.go
lahabana: Let me add a guard and logs in all places because it does seem odd

lahabana: Done!

PR #8705, opened 2023-12-20
feat(resources): add kuma.io/display-name label

### LINKED ISSUE #8484: Rule merging alghoritm is not sorting lexographically on MeshTrafficPermission

jakubdyszkiewicz:
### Checklist prior to review

Fix #8484

This PR adds a new standard label `kuma.io/display-name`.

`kuma.io/display-name` is needed for the following cases:
* To preserve original Kubernetes name without namespace
* To preserve original name after KDS sync, because we add hashes now.
* To use it in matching to fix issue linked above.

We could add always it (postgres, memory store), but it's a bit awkward to repeat `name` field when it's equal.

If needed, I can write MADR for this change, but I'm not sure it's worth it.


api/mesh/v1alpha1/dataplane_helpers.go
```diff
@@ -42,6 +42,11 @@ const (
 
 	// Used for Service-less dataplanes
 	TCPPortReserved = 49151 // IANA Reserved
+
+	// DisplayName is a standard label that can be used to easier recognize policy name.
+	// On Kubernetes, Kuma resource name contains namespace. Display name is original name without namespace.
+	// The name contains hash when the resource is synced from global to sync. In this case, display name is original name from originated CP.
```
lobkovilya, line 48: typo
```suggestion
	// The name contains hash when the resource is synced from global to zone. In this case, display name is original name from originated CP.
```

lobkovilya (approved):

lahabana (approved):
Hmm wouldn't it just make sense to use it always? After all how long would it be until we do things differently in memory and zones?

jakubdyszkiewicz:
by "use it always" do you mean add it always to every single store?

lahabana:
> by "use it always" do you mean add it always to every single store?

Yeah but I could be wrong...

jakubdyszkiewicz:
We could always add it if we find this to be problematic. It's hard to go the other way around

PR #8710, opened 2023-12-21
ci(github): use GitHub public artifact API when downloading artifacts from CircleCI

jijiechen:
To fix CI errors here:
https://github.com/kumahq/kuma/actions/runs/7273842586/job/19819186863#step:16:73

### Background
The artifact downloading failures started to occur after this PR: https://github.com/kumahq/kuma/pull/8701

There are some behavior changes between these two versions of the artifact plugin, according to GH documents:
https://github.com/actions/toolkit/tree/main/packages/artifact#improvements

These changes broke our download artifact implementation for CircleCI: artifacts uploaded by version 4 are not available to the "internal API" we are using (which worked for version 3).

Fortunately, according to GH docs, artifacts uploaded by version 4 are now available to their public rest APIs. So we are switching to use their public API. This will also be a more reliable way to achieve our goals.

### Checklist prior to review

  - N/A
  - Confirmed, it will work with existing sync scripts
  - Manually tested
    - Added
  - No need
  - No need

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,10 +16,7 @@ parameters:
   gh_action_build_artifact_name:
     type: string
     default: ""
-  gh_action_artifact_list_url:
-    type: string
-    default: ""
-  gh_action_runtime_token:
+  gh_action_run_id:
     type: string
     default: ""
   e2e_param_k8sVersion:
@@ -415,6 +412,7 @@ jobs:
       - run:
           name: Download build output from GitHub action
           command: |
+            set -e
             echo "Running with: \
               k8s:<< parameters.k8sVersion >> \
               target:<< parameters.target >> \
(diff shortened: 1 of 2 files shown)


jijiechen:
Here is an example of successful runs:
https://github.com/jijiechen/kuma/actions/runs/7285468977/job/19853703909

.circleci/config.yml
```diff
@@ -468,49 +464,56 @@ jobs:
...
+            if [[ "$TARBALL" == "" ]]; then
               echo "Could not find built artifact for linux($ARCH)."
               exit 1
             fi
-            cat $ARTIFACT.sha256 | sha256sum --check
+            cat ${TARBALL}.sha256 | sha256sum --check
 
             mkdir $BASE_DIR/distributions/artifacts
-            tar -xzf $ARTIFACT -C $BASE_DIR/distributions/artifacts
+            tar -xzf $TARBALL -C $BASE_DIR/distributions/artifacts
```
lahabana, line 516: If this uses public APIs now can we just use `gh` and not have to deal with all the bash code?

jijiechen:
We'll need to set an env var with `GITHUB_TOKEN` on CircleCI to run the E2E tests on CircleCI.
The token should have `read` permission to the repo that the CI is operating on.

.circleci/config.yml
jijiechen: Sure, good idea! 
Just had a look at the `gh` CLI, it worked like a charm. So I've just updated the code.

jijiechen: We'll still need this block throug. What it does is restoring the folder structure after the downloading, so it's required regardless of how we download.

lahabana (approved):
LGTM!

PR #8730, opened 2024-01-02
feat(xds): skip generating clusters for external services without any endpoints

jijiechen:
In certain cases,  endpoints of external services can be invalidated by `TrafficPermission` policies. In such case,  these external services are actually treated as "non-external" in our current implementation of xDS generating. 

While it didn't cause any actual problems, it's still need to be fixed as they may mess up some time. For now, we can see, mTLS certificates can be unexpectedly issued for data planes that have outbound only to these external services. 

### Checklist prior to review

  - N/A
  - Should not break.
  - Added
  - No need
  - No need
  - No need

### DIFF
--- a/pkg/xds/generator/outbound_proxy_generator.go
+++ b/pkg/xds/generator/outbound_proxy_generator.go
@@ -49,7 +49,8 @@ func (g OutboundProxyGenerator) Generate(ctx context.Context, _ *model.ResourceS
 	for _, outbound := range outbounds {
 		// Determine the list of destination subsets
 		// For one outbound listener it may contain many subsets (ex. TrafficRoute to many destinations)
-		routes := g.determineRoutes(proxy, outbound, clusterCache, xdsCtx.Mesh.Resource.ZoneEgressEnabled())
+		routes := g.determineRoutes(proxy, outbound, clusterCache,
+			xdsCtx.Mesh.ExternalServicesEndpointMap, xdsCtx.Mesh.Resource.ZoneEgressEnabled())
 		clusters := routes.Clusters()
 
 		protocol := inferProtocol(xdsCtx.Mesh, clusters)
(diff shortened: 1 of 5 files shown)


pkg/xds/generator/outbound_proxy_generator.go
```diff
@@ -367,12 +367,23 @@ func (OutboundProxyGenerator) determineRoutes(
...
+			var hasEndpoints bool
 			if endpoints := proxy.Routing.OutboundTargets[service]; len(endpoints) > 0 {
+				hasEndpoints = true
 				isExternalService = endpoints[0].IsExternalService()
 			}
 			if endpoints := proxy.Routing.ExternalServiceOutboundTargets[service]; len(endpoints) > 0 {
+				hasEndpoints = true
 				isExternalService = true
 			}
+			if _, ok := proxy.Routing.ExternalServiceUnavailable[service]; ok {
```
slonka, line 379: do we need this new field? can we just pass `xdsCtx.Mesh.ExternalServicesEndpointMap` to `determineRoutes` and do a lookup like here - https://github.com/kumahq/kuma/pull/8730/files#diff-f85d1997072108108f860d61dab0a92e0002f6edbdfab7be9faf0ef66ea89e42R196 ?

slonka (commented):
added a question - maybe we don't need a new struct

pkg/xds/generator/outbound_proxy_generator.go
jijiechen: Good point, yes, I think by using this existing data, we don't need to introduce a new field there.
I've just learnt more about the existing design. Thank you very much.

jijiechen: I've just updated code. Thanks again.

jijiechen:
The E2E failures in this PR are reflecting code changes, I'll take a look.

jijiechen:
I can't reproduce the E2E test failures on local env after a few trails, so I'll close this PR for now. If we decide this to be a high priority change, I'll file a new PR later.

PR #8760, opened 2024-01-05
feat(toggles): add plugin policy toggles

slonka:
This is a draft of making it possible to dynamically disable/enable policies. It ties it back with helm's `.Values.plugins.policies` that is already used to do that in k8s (only via rbac/webhooks). This extends the feature to also work in Universal and Universal on k8s modes by simply not registering the plugin. What I'm not sure about here is the fact that we initialize / register policies in some places:
- https://github.com/kumahq/kuma/blob/cc843e69667f068bbd63a9dbebf748e292b7cd8f/app/kumactl/cmd/root.go#L27
- https://github.com/kumahq/kuma/blob/12ab22c7673ac76d6e8b84ddd46bd647888ad56b/pkg/xds/sync/proxy_builder_test.go#L24

are we ok with manually initialising policies there?

The second thing is we need to keep this default env var up to date (~~before we had to remember to manually import the plugin for side effects in `github.com/kumahq/kuma/pkg/plugins/policies` - now we have to do that~~ ~~edit: this turned out to not be true, it's autogenerated, so now we additionally have to update the env var~~ edit2: made it autogenerated as well). 

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- linked from outside
- [ ] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS -- this needs coordination with child repos
- [ ] Tests (Unit test, E2E tests, manual test on universal and k8s) -- TBA
- [ ] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? -- I changed the definition of `.Values.plugins.policies` so maybe but I might just revert that because it could be not needed. Also this is more of an internal change, users probably don't need to touch `.Values.plugins.policies`
- [X] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) -- no

> Changelog: skip


pkg/plugins/policies/imports.go
```diff
@@ -1,18 +1,44 @@
...
+	"github.com/kumahq/kuma/pkg/plugins/policies/meshratelimit"
+	"github.com/kumahq/kuma/pkg/plugins/policies/meshretry"
+	"github.com/kumahq/kuma/pkg/plugins/policies/meshtcproute"
+	"github.com/kumahq/kuma/pkg/plugins/policies/meshtimeout"
+	"github.com/kumahq/kuma/pkg/plugins/policies/meshtrace"
+	"github.com/kumahq/kuma/pkg/plugins/policies/meshtrafficpermission"
 )
+
+var nameToModule = map[string]func(bool){
+	"meshaccesslog":             meshaccesslog.InitPlugin,
```
slonka, line 21: this should be autogenerated

pkg/config/app/kuma-cp/config.go
```diff
@@ -167,6 +167,8 @@ type Config struct {
 	Tracing tracing.Config `json:"tracing"`
 	// EventBus is a configuration of the event bus which is local to one instance of CP.
 	EventBus eventbus.Config `json:"eventBus"`
+	// Configuration of enabled policies
+	PluginPoliciesEnabled []string `json:"-" envconfig:"KUMA_PLUGIN_POLICIES_ENABLED" default:"meshaccesslog,meshcircuitbreaker,meshfaultinjection,meshhealthcheck,meshhttproute,meshloadbalancingstrategy,meshmetric,meshproxypatch,meshratelimit,meshretry,meshtcproute,meshtimeout,meshtrace,meshtrafficpermission"`
```
slonka, line 171: this should be autogenerated

slonka: Now it is.

pkg/plugins/policies/imports.go
```diff
@@ -1,18 +1,44 @@
...
+	"meshproxypatch":            meshproxypatch.InitPlugin,
+	"meshratelimit":             meshratelimit.InitPlugin,
+	"meshretry":                 meshretry.InitPlugin,
+	"meshtcproute":              meshtcproute.InitPlugin,
+	"meshtimeout":               meshtimeout.InitPlugin,
+	"meshtrace":                 meshtrace.InitPlugin,
+	"meshtrafficpermission":     meshtrafficpermission.InitPlugin,
+}
+
+func InitPolicies(enabledPluginPolicies []string) {
```
lobkovilya, line 37: If we change this to `func init()` does it mean the existing imports like ` _ "github.com/kumahq/kuma/pkg/plugins/policies"` will work?

slonka: we can't use init unless we manually read the env variable responsible for handling switching on / off policies, do you think it's a better approach?

slonka: I'm changing it now, we will see if it works better

pkg/plugins/policies/imports.go
```diff
@@ -1,18 +1,60 @@
...
+
+func initAllPolicies() {
+	for _, initializer := range nameToModule {
+		initializer()
+	}
+}
+
+func init() {
+	// we read this manually otherwise we would have to wire in enabling plugins in every test
+	rawEnabledPluginPolicies := os.Getenv("KUMA_PLUGIN_POLICIES_ENABLED")
```
jakubdyszkiewicz, line 48: What is unfortunate about this approach is that this env var is not documented and it's going with "different flow" than the regular config. Can't we include this in our config and just load policies in bootstrap phase?

You said that this might be a problem for tests. If so, we could maybe init policies as part of `RunSpecs` which is called for every test

slonka: I'll try that approach

pkg/plugins/policies/imports.go
slonka: https://github.com/kumahq/kuma/pull/8828 created alternative PR with that approach

pkg/plugins/policies/imports.go
slonka: @jakubdyszkiewicz - please look at https://github.com/kumahq/kuma/pull/8828

slonka:
closing in favour of https://github.com/kumahq/kuma/pull/8828

Issue #5045, opened 2022-09-22
HTTP stats for `MeshTrafficPermission`


lobkovilya:
### Description

Both TrafficPermission and MeshTrafficPermission are using a network RBAC filter, which produces an inaccurate allow/deny stat (check happens on the connection level). For HTTP services an http RBAC filter should be used.

PR #8777, opened 2024-01-09
chore(MeshTrafficPermission): use http rbac envoy filter when applicable

bartsmykla:
Instead of always using `envoy.filters.network.rbac` we are gonna use `envoy.filters.http.rbac` when applicable.

### Checklist prior to review

  - Closes https://github.com/kumahq/kuma/issues/5045
  - It won't
  - Unit tests were updated
  - There is no need
  - There is no need


pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/testdata/apply-egress.golden.yaml
```diff
@@ -13,9 +13,9 @@ resources:
         - external-service-1{mesh=mesh-1}
         transportProtocol: tls
       filters:
-      - name: envoy.filters.network.rbac
+      - name: envoy.filters.http.rbac
```
jakubdyszkiewicz, line 16: have you checked that it actually works? I thought that HTTP filters have to be placed in httpFilters inside of HTTP connection manager

bartsmykla: Good catch.

bartsmykla: addressed

jakubdyszkiewicz (commented):
We use TCP RBAC stats in Grafana dashboards. Can you check what happens there if we use HTTP RBAC filter?

test/e2e_env/universal/meshtrafficpermission/meshtrafficpermission.go
```diff
@@ -57,7 +57,7 @@ func MeshTrafficPermissionUniversal() {
 				universal.Cluster, "demo-client", "test-server.mesh",
 			)
 			g.Expect(err).ToNot(HaveOccurred())
-			g.Expect(response.ResponseCode).To(Equal(503))
+			g.Expect(response.ResponseCode).To(Equal(403))
```
lukidzi, line 60: Could we have an another case for TCP traffic?

bartsmykla:
> We use TCP RBAC stats in Grafana dashboards. Can you check what happens there if we use HTTP RBAC filter?

I've checked on local k3d cluster and it seems that stats `envoy_rbac_*` stats we are using in dashboards are still correctly updated

test/e2e_env/universal/meshtrafficpermission/meshtrafficpermission.go
bartsmykla: added

jakubdyszkiewicz (approved):

PR #8782, opened 2024-01-09
chore(kuma-cp): pass context.Context to plugin.Apply

michaelbeaumont:
Get rid of the current [use of `context.TODO()` in MeshHTTPRoute plugin](https://github.com/kumahq/kuma/pull/8782/files#diff-2d4e9cddeb9533b7e939184577a3c8dd84271a49d7b58919a8fc2612c595ab63L79-R79).

### Checklist prior to review

### DIFF
--- a/pkg/core/plugins/interfaces.go
+++ b/pkg/core/plugins/interfaces.go
@@ -93,9 +93,9 @@ type PolicyPlugin interface {
 	Plugin
 	// MatchedPolicies return all the policies of the plugins' type matching this dataplane. This is used in the inspect api and accessible in Apply through `proxy.Policies.Dynamic`
 	MatchedPolicies(dataplane *core_mesh.DataplaneResource, resources xds_context.Resources) (core_xds.TypedMatchingPolicies, error)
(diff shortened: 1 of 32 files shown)


bartsmykla:
I don't see it being used now. Could you update the PR description with information why it's necessary?

michaelbeaumont:
> I don't see it being used now. Could you update the PR description with information why it's necessary?

https://github.com/kumahq/kuma/pull/8782/files#diff-2d4e9cddeb9533b7e939184577a3c8dd84271a49d7b58919a8fc2612c595ab63L79-R79

jakubdyszkiewicz:
so in the end this is used in the `datasource.Loader` only? We know that the implem used in xds config does not use I/O therefore context is irrelevant.

I'd consider passing `context.TODO()` or nil or whatever in `loadBytes` func instead of bringing the context through all layers. Alternatively, we could make `staticLoader` public, add a `LoadStatically` (or similar) func that does not require context at all.

I'd like to discourage users of using I/O in xds generation. My long-term plan was to ban I/O in XDS generators so instead of goroutines with tickers we could do a queue-based approach, because the process would be purely CPU-bound.

michaelbeaumont:
Yeah it's a good point

PR #8804, opened 2024-01-10
ci(github): don't rely on artifacts, instead rebuild each time

lahabana:
- Storing build artifacts was slow and unstable, we now rebuild each time.
- We also reduce the number of arm test cases (there's no real reason to rerun tests on arm so many times).

Example full run: https://github.com/kumahq/kuma/actions/runs/7478606778/job/20353924056

### Checklist prior to review

### DIFF
--- a/.github/workflows/build-test-distribute.yaml
+++ b/.github/workflows/build-test-distribute.yaml
@@ -240,7 +240,7 @@ jobs:
               "test_e2e": {
                 "target": [""],
                 "k8sVersion": ["kindIpv6", "${{ env.K8S_MIN_VERSION }}", "${{ env.K8S_MAX_VERSION }}"],
-                "arch": ["amd64", "arm64"],
+                "arch": ["amd64"],
                 "parallelism": [3],
                 "cniNetworkPlugin": ["flannel"],
                 "legacyKDS": [false]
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test-distribute.yaml
```diff
@@ -1,7 +1,7 @@
 name: "build-test-distribute"
 on:
   push:
-    branches: ["master", "release-*", "!*-merge-master"]
+    branches: ["master", "release-*", "!*-merge-master", "reworkCI"]
```
lahabana, line 4:
```suggestion
    branches: ["master", "release-*", "!*-merge-master"]
```

jijiechen:
Given the actual binary build steps are quite productive, I think this change will be benefitial!

Had a look on the build steps on existing runs, here are timing:
- make build **~110s**
- make images   **~60s**
- upload artifact **~70s**
- download artifact **~30s**, sometimes **10m**

By inline the steps into tests and distributions, we save the time on artifact uploading/downloading.

.github/workflows/e2e.yaml
```diff
@@ -77,37 +77,19 @@ jobs:
...
-          for BIN in $(ls $SRC_DIR); do
-            mkdir -p $DEST_DIR/$BIN
-            cp $SRC_DIR/$BIN $DEST_DIR/$BIN/
-          done
+          make -j build/distributions
+      - name: "Github Actions: images"
+        if: steps.eval-params.outputs.run-type == 'github'
+        run: |
+          make -j images
+          make -j docker/save
```
jijiechen, line 92: Do we need `docker/save` here?

jijiechen (approved):
LGTM

.github/workflows/e2e.yaml
lahabana: Yes because of: https://github.com/kumahq/kuma/blob/5fba9cfe5e813dc940509b4dc3b75ed4d7f9e349/mk/e2e.new.mk#L15-L24 

and I wanted to keep changes to the minimum

PR #8806, opened 2024-01-11
Revert "ci(github): remove docker DNS setting on ipv6 e2e tests"

lahabana:
Reverts kumahq/kuma#8676


jijiechen:
it looks like some E2E cases are now failing because of DNS issues, after the PR #8676

Here is a example failure: https://github.com/kumahq/kuma/actions/runs/7484780838/job/20372243275#step:13:11196

So let's revert that PR for now.

PR #8822, opened 2024-01-11
test(e2e): run all test/e2e tests

lahabana:
When moving from circleci we missed adding parallelisation to test/e2e so a portion of the jobs were not running.

### Checklist prior to review


test/e2e/helm/kuma_helm_deploy_global_zone.go
```diff
@@ -169,23 +169,22 @@ interCp:
 		})
 	})
 
-	It("should execute admin operations on Global CP", func() {
+	// Seems like admin operations are broken:
+	// error while running command: exit status 1; Error: Could not execute admin operation (could not send XDSConfigRequest: Resource not found: type="Dataplane" name="demo-client-84b6dc7cb8-wq44w.kuma-system" mesh="default")
```
lahabana, line 173: @lobkovilya I think this is a result of the rename. I've changed the tests but it now seems to be reaching the wrong DP (this test wasn't running anymore which is why you didn't catch it :( ).

lukidzi (approved):

lahabana:
Covered the new ignored tests in: https://github.com/kumahq/kuma/issues/8838

PR #8827, opened 2024-01-12
feat(policy): support targeting many gateways

lukidzi:
### Checklist prior to review

There was no way to target all gateways with the common configuration. Added new type `MeshGatewaysSubset` which allows pointing many gateways at once.

### DIFF
--- a/pkg/core/resources/apis/mesh/validators.go
+++ b/pkg/core/resources/apis/mesh/validators.go
@@ -348,10 +348,13 @@ func ValidateTargetRef(
 		}
 		err.Add(disallowedField("mesh", ref.Mesh, ref.Kind))
 		err.Add(disallowedField("tags", ref.Tags, ref.Kind))
-	case common_api.MeshSubset:
+	case common_api.MeshSubset, common_api.MeshGatewaysSubset:
 		err.Add(disallowedField("name", ref.Name, ref.Kind))
 		err.Add(disallowedField("mesh", ref.Mesh, ref.Kind))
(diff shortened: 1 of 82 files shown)


pkg/core/resources/apis/mesh/validators.go
```diff
@@ -348,10 +348,13 @@ func ValidateTargetRef(
 		}
 		err.Add(disallowedField("mesh", ref.Mesh, ref.Kind))
 		err.Add(disallowedField("tags", ref.Tags, ref.Kind))
-	case common_api.MeshSubset:
+	case common_api.MeshSubset, common_api.MeshGatewaysSubset:
 		err.Add(disallowedField("name", ref.Name, ref.Kind))
 		err.Add(disallowedField("mesh", ref.Mesh, ref.Kind))
 		err.Add(ValidateTags(validators.RootedAt("tags"), ref.Tags, ValidateTagsOpts{}))
+		if ref.Kind == common_api.MeshGatewaysSubset && len(ref.Tags) > 0 && !opts.GatewayListenerTagsAllowed {
+			err.Add(disallowedField("tags", ref.Tags, ref.Kind))
```
michaelbeaumont, line 356: Hmm this may be a bit tricky. What if you want to select a strict subset of mesh gateways, but all listeners of each of them?

michaelbeaumont: probably we should ensure that MeshGatewaysSubset can only select _all_ listeners of a subset of gateways. So we would ignore listener tags when matching

michaelbeaumont: but we can fix that in a later PR

PR #8828, opened 2024-01-12
feat(toggles): add plugin policy toggles

slonka:
alternative to https://github.com/kumahq/kuma/pull/8760

### Checklist prior to review


pkg/plugins/policies/imports.go
```diff
@@ -1,18 +1,61 @@
...
+		}
+	}
+}
+
+func InitPolicies(enabledPluginPolicies []string) {
+	for _, policy := range enabledPluginPolicies {
+		initializer, ok := nameToModule[policy]
+		if ok && !initializer.initialized {
+			initializer.initFn()
+			initializer.initialized = true
```
slonka, line 56: This is not nice but needed, we initialise the policies in both `RunSpec` and `kumactl` code and if both are reached by test code then we get panics on double registration. I don't see a nice way around this, suggestions welcomed.

docs/generated/raw/kuma-cp.yaml
```diff
@@ -753,19 +730,32 @@ experimental:
...
   openTelemetry:
     endpoint: "" # e.g. otel-collector:4317
-
 # Configuration of the event bus which is local to one instance of CP
 eventBus:
   # BufferSize controls the buffer for every single event listener.
   # If we go over buffer, additional delay may happen to various operation like insight recomputation or KDS.
   bufferSize: 100 # ENV: KUMA_EVENT_BUS_BUFFER_SIZE
+policies:
+  pluginPoliciesEnabled:
```
jakubdyszkiewicz, line 747: Comment + ENV: ?

jakubdyszkiewicz (approved):

pkg/config/plugins/policies/policies.go
```diff
@@ -0,0 +1,22 @@
+package policies
+
+func DefaultPoliciesConfig() *Config {
+	return &Config{
+		PluginPoliciesEnabled: []string{
+			"meshaccesslogs",
```
jijiechen, line 6: Is it better to make these strings constants?

jijiechen (approved):

pkg/config/plugins/policies/policies.go
slonka: not sure it matters that much, I already have this in this format and it's autogenerated so would require more changes. If you really want to change it please open up an issue.

jakubdyszkiewicz (approved):

PR #8829, opened 2024-01-12
docs(MADR): targeting all gateways in policy

lukidzi:
### Checklist prior to review

### DIFF
--- a/docs/madr/decisions/037-configure-all-gateways.md
+++ b/docs/madr/decisions/037-configure-all-gateways.md
@@ -92,3 +92,4 @@ We could remove requirement of the name when using `kind: MeshGateway` and when 
 
 - Users can impact the configuration of all gateways by mistake
 - Change of the current api
+- Problem with ordering when there is more specific policy with the same kind but with name defined
--- a/docs/madr/decisions/037-configure-all-gateways.md
+++ b/docs/madr/decisions/037-configure-all-gateways.md
@@ -131,7 +131,7 @@ targetRef:
   kind: Mesh|MeshSubset
   proxyTypes: ["Sidecar", "Gateway"]
 ```
-When field is not defined or empty we assume that the user wants to apply configuration to all sidecar, not gateways. 
+When field is not defined or empty we assume that the user wants to apply configuration to all sidecar and gateways. 
 In this situtation we won't break current behaviour. Also, worth mentioning that `proxyTypes` can only be set when 
 kind `Mesh` or `MeshSubset`. Another thing is that `proxyTypes` won't affect ordering and we base order on name.
 
--- a/docs/madr/decisions/037-configure-all-gateways.md
+++ b/docs/madr/decisions/037-configure-all-gateways.md
@@ -119,19 +119,21 @@ targetRef:
 - What about merging `MeshSubset`'s
 - Can break current behavior if the user has already defined the tag
 
-## Add a new field to targetRef `gateways: include/exclude/only`
+## Add a new field to targetRef `proxyTypes: [sidecar, gateway]`
 
-We can add a new field `gateways` to targetRef which allows to apply policy to: 
+We can add a new field `proxyTypes` to targetRef which allows to apply policy to: 
 * only gateways
 * all dataplanes including gateways
 * all dataplanes excluding gateways
 
 ```yaml
 targetRef:
-  kind: MeshSubset
-  gateways: "only/include/exclude"
+  kind: Mesh|MeshSubset
+  proxyTypes: ["sidecar", "gateway"]
 ```
-By default we are going to include gateways to don't change the  current behavior.
+When field is not defined or empty we assume that the user wants to apply configuration to all dataplanes. 
+In this situtation we won't break current behaviour. Also, worth mentioning that `proxyTypes` can only be set when 
+kind `Mesh` or `MeshSubset`. Another thing is that `proxyTypes` won't affect ordering and we base order on name.
 
 ### Positive Consequences
 
@@ -143,4 +145,3 @@ By default we are going to include gateways to don't change the  current behavio
 ### Negative Consequences
 
 - Extending targetRef object
-- Merging might be a bit different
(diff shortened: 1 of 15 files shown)


docs/madr/decisions/037-configure-all-gateways.md
```diff
@@ -0,0 +1,64 @@
...
+
+We can introduce a new kind `MeshGatewaysSubset` It only picks gateways based on tags, if you provide tags. But, it can't pick a specific gateway by name.
+
+```yaml
+targetRef:
+  kind: MeshGatewaysSubset
+  tags: {} # not required
+```
+
+Thanks to this we can allow users to target all gateways and apply some good default configuration for them.
```
michaelbeaumont, line 32: I think we can add a note that this will have to match either only an entire gateway, i.e. ignoring listener tags OR can match specific listeners, depending on the policy.

docs/madr/decisions/037-configure-all-gateways.md
```diff
@@ -0,0 +1,69 @@
...
+
+This document aims to clarify how to configure all gateways or a subset of them, especially when adjustments, such as in `MeshTimeout`, are needed compared to regular Mesh traffic.
+
+## Decision Drivers
+
+- When we set up default rules, we want to use different settings for gateways than we do for regular Mesh applications.
+
+## Considered Options
+
+- Introduce new kind `MeshGatewaysSubset`
```
lahabana, line 15: Should be MeshGatewaySubset

lahabana (approved):

docs/madr/decisions/037-configure-all-gateways.md
```diff
@@ -0,0 +1,76 @@
...
+- ?
+
+## Introduce new kind `MeshGatewaysSubset`
+
+We can introduce a new kind `MeshGatewaysSubset` It only picks gateways based on tags, if you provide tags. But, it can't pick a specific gateway by name.
+
+```yaml
+targetRef:
+  kind: MeshGatewaysSubset
+  tags: {} # not required
```
lahabana, line 29: Please just add a note that no tags means any gateway :)

docs/madr/decisions/037-configure-all-gateways.md
```diff
@@ -0,0 +1,78 @@
...
+## Considered Options
+
+- Introduce new kind `MeshGatewaysSubset`
+- Add label to all gateways and target them by kind `MeshSubset`
+
+## Decision Outcome
+
+- ?
+
+## Introduce new kind `MeshGatewaysSubset`
```
lukidzi, line 22: The naming of this kind is a bit confusing but hard to find other name
- MeshGatewaysSubset - Subset of all gateways
- MeshGateway - specific gateway by name
- MeshGateways - all gateways, but it's too similar to MeshGateway
- others?

Another option could be to introduce an additional field in a `targetRef` object but it just moves functionality of kind outside so I don't think this is a good option.

lobkovilya: MeshGatewaySuperset :) i.e. if we want to select all MeshServices we can add `MeshServiceSuperset`

michaelbeaumont: Hmm true, or actually just `MeshGatewaySet` right?

lobkovilya: TBH I'd prefer `...Superset`, there are definitions in Wiki on "superset/subset" and it seems more clear we're trying to oppose "superset" to "subset". While if you're reading `...Set` it doesn't tell you much on its own

docs/madr/decisions/037-configure-all-gateways.md
```diff
@@ -0,0 +1,147 @@
...
+
+We can add a new field `proxyTypes` to targetRef which allows to apply policy to: 
+* only gateways
+* all dataplanes including gateways
+* all dataplanes excluding gateways
+
+```yaml
+targetRef:
+  kind: Mesh|MeshSubset
+  proxyTypes: ["sidecar", "gateway"]
```
lobkovilya, line 132: PascalCase :)
```suggestion
  proxyTypes: ["Sidecar", "Gateway"]
```

docs/madr/decisions/037-configure-all-gateways.md
```diff
@@ -0,0 +1,147 @@
...
+
+- Introduce new kind `MeshGatewaysSuperset`
+- Add label to all gateways and target them by kind `MeshSubset`
+- Remove requirement of `name` when using kind `MeshGateway`
+- Use internal tag `gateways: include/exclude/only` when using with `MeshSubset`
+- Add a new field to targetRef `gateways: include/exclude/only`
+
+## Decision Outcome
+
+- Add a new field to targetRef `gateways: include/exclude/only`
```
bartsmykla, line 23:
```suggestion
- Add a new field to targetRef `proxyTypes: ["Sidecar", "Gateway"`
```

docs/madr/decisions/037-configure-all-gateways.md
```diff
@@ -0,0 +1,147 @@
...
+
+- When we set up default rules, we want to use different settings for gateways than we do for regular Mesh applications.
+
+## Considered Options
+
+- Introduce new kind `MeshGatewaysSuperset`
+- Add label to all gateways and target them by kind `MeshSubset`
+- Remove requirement of `name` when using kind `MeshGateway`
+- Use internal tag `gateways: include/exclude/only` when using with `MeshSubset`
+- Add a new field to targetRef `gateways: include/exclude/only`
```
bartsmykla:
```suggestion
- Add a new field to targetRef `proxyTypes: ["Sidecar", "Gateway"]`
```

bartsmykla (approved):
lgtm

docs/madr/decisions/037-configure-all-gateways.md
```diff
@@ -0,0 +1,147 @@
...
+* all dataplanes including gateways
+* all dataplanes excluding gateways
+
+```yaml
+targetRef:
+  kind: Mesh|MeshSubset
+  proxyTypes: ["Sidecar", "Gateway"]
+```
+When field is not defined or empty we assume that the user wants to apply configuration to all sidecar, not gateways. 
+In this situtation we won't break current behaviour. Also, worth mentioning that `proxyTypes` can only be set when
```
lobkovilya, line 135: Hmm, but today `kind: Mesh` means both gateways and sidecars, no?

lukidzi: Yes, I think I forgot to update it. Let me fix

lobkovilya (approved):

PR #8873, opened 2024-01-16
feat(kuma-cp): require `kuma.io/origin: zone` label when creating zone-origination policies

lobkovilya:
### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix https://github.com/kumahq/kuma/issues/8934


jakubdyszkiewicz (commented):
This is optional UX feature. I remember we talked about disabling it via env var if someone don't want to have this validation. Should we introduce such env var?

pkg/config/app/kuma-cp/kuma-cp.defaults.yaml
```diff
@@ -528,6 +528,7 @@ multizone:
       # Response backoff is a time Zone CP waits before sending ACK/NACK.
       # This is a way to slow down Global CP from sending resources too often.
       responseBackoff: 0s # ENV: KUMA_MULTIZONE_ZONE_KDS_RESPONSE_BACKOFF
+    disableOriginLabelValidation: false # ENV: KUMA_MULTIZONE_ZONE_DISABLE_ORIGIN_LABEL_VALIDATION
```
jakubdyszkiewicz, line 531: we need description for this

pkg/plugins/runtime/k8s/webhooks/validation.go
```diff
@@ -103,40 +109,109 @@ func (h *validatingHandler) decode(req admission.Request) (core_model.Resource,
...
+	}
+	return true
+}
+
+func (h *validatingHandler) isResourceAllowed(r core_model.Resource) bool {
+	if !h.federatedZone || !r.Descriptor().IsPluginOriginated {
+		return true
+	}
+	if !h.disableOriginLabelValidation {
+		if r.GetMeta().GetLabels() == nil || r.GetMeta().GetLabels()[mesh_proto.ResourceOriginLabel] != string(mesh_proto.ZoneResourceOrigin) {
```
jakubdyszkiewicz, line 174: nit: shouldn't we use this `ResourceOrigin()`?

jakubdyszkiewicz (approved):

pkg/plugins/runtime/k8s/webhooks/validation.go
lobkovilya: ah missed that place! thanks

Issue #4946, opened 2022-08-31
Sidecars are not allocated ephemeral-storage by default


michaelgoodrx:
### What happened?

We have some workloads that use ephemeral-storage. When they run on the same node as a pod running the kuma sidecar, the pod running the kuma sidecar can be evicted, because it uses some ephemeral storage even though it doesn't actually request any. This caused pod evictions which have caused other bugs.

Kuma should specify a small ephemeral-storage request by default, along with cpu and memory. I can add it myself using a patch, but most people are going to want at least a little bit of storage for cluster stability.

jakubdyszkiewicz:
xref https://github.com/kumahq/kuma-net/issues/66

lahabana:
This is done already right?

michaelbeaumont:
I think actually it'd be https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#setting-requests-and-limits-for-local-ephemeral-storage

lahabana:
@stu_at_kong

PR #8882, opened 2024-01-17
feat(injector): add ephemeral-storage resource request/limit for sidecars

jijiechen:
fixes https://github.com/kumahq/kuma/issues/4946

### Checklist prior to review

  - https://github.com/kumahq/kuma/issues/4946
  - Confirmed
  - test case updated
    - No need
  - No
  - No


pkg/plugins/runtime/k8s/webhooks/injector/injector.go
```diff
@@ -73,27 +73,10 @@ type KumaInjector struct {
 }
 
 func (i *KumaInjector) InjectKuma(ctx context.Context, pod *kube_core.Pod) error {
-	ns, err := i.namespaceFor(ctx, pod)
-	if err != nil {
-		return errors.Wrap(err, "could not retrieve namespace for pod")
-	}
 	logger := log.WithValues("pod", pod.GenerateName, "namespace", pod.Namespace)
-	// Log deprecated annotations
-	for _, d := range metadata.PodAnnotationDeprecations {
```
jijiechen, line 82: Refactoring: moved to the separate file `precheck.go`

pkg/plugins/runtime/k8s/webhooks/injector/injector.go
```diff
@@ -404,6 +301,8 @@ func (i *KumaInjector) NewSidecarContainer(
 	}
 
 	container.Name = k8s_util.KumaSidecarContainerName
+	container.Resources.Requests[kube_core.ResourceEphemeralStorage] = pointer.Deref(kube_api.NewScaledQuantity(10, kube_api.Mega))
+	container.Resources.Limits[kube_core.ResourceEphemeralStorage] = pointer.Deref(kube_api.NewScaledQuantity(10, kube_api.Mega))
```
jakubdyszkiewicz, line 305: should this be added only for injected sidecar or also for builtin gateway? CC @michaelbeaumont If for both, it's better to place it in `NewContainer()` func

jijiechen: I've moved this into `factory.go` so it can be shared for both sidecars and gateways

jakubdyszkiewicz (approved):

pkg/plugins/runtime/k8s/containers/factory.go
```diff
@@ -158,12 +159,14 @@ func (i *DataplaneProxyFactory) NewContainer(
...
+				kube_core.ResourceCPU:              kube_api.MustParse(i.ContainerConfig.Resources.Requests.CPU),
+				kube_core.ResourceMemory:           kube_api.MustParse(i.ContainerConfig.Resources.Requests.Memory),
+				kube_core.ResourceEphemeralStorage: pointer.Deref(kube_api.NewScaledQuantity(50, kube_api.Mega)),
 			},
 			Limits: kube_core.ResourceList{
-				kube_core.ResourceCPU:    kube_api.MustParse(i.ContainerConfig.Resources.Limits.CPU),
-				kube_core.ResourceMemory: kube_api.MustParse(i.ContainerConfig.Resources.Limits.Memory),
+				kube_core.ResourceCPU:              kube_api.MustParse(i.ContainerConfig.Resources.Limits.CPU),
+				kube_core.ResourceMemory:           kube_api.MustParse(i.ContainerConfig.Resources.Limits.Memory),
+				kube_core.ResourceEphemeralStorage: pointer.Deref(kube_api.NewScaledQuantity(1, kube_api.Giga)),
```
michaelbeaumont, line 169: What's the advantage and implications of setting a limit?

jijiechen: Technically, setting a limit to every resource allocation is a "good manner" to make sure we don't over consume. The implication is when we exceed that limit, the pod get evicted. 

Right now, the [default allowed logging](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration) on Kubernetes clusters is, 5 x 10Mi

michaelbeaumont: Actually [we shouldn't even be setting CPU limits](https://home.robusta.dev/blog/stop-using-cpu-limits). What if users configure access logging to file in `/tmp` for example, now the Pod will be evicted if this file hits 10GiB?

PR #8887, opened 2024-01-17
feat(meshmetrics): rename used only to include unused 

slonka:
rename used only to include unused and revert the logic

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- reported via slack


pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -143,7 +143,7 @@ func envoyMetricsFilter(conf api.Conf) string {
 	if query != "" {
 		query += "&"
 	}
-	if pointer.Deref(conf.Sidecar.UsedOnly) {
```
Automaat, line 146: i think that this won't work by default because of check: in line `136`

slonka: modified, can you check again?

Automaat (approved):

PR #8896, opened 2024-01-17
feat(MeshTimeout): add RequestHeadersTimeout option and configure gateway

lukidzi:
### Checklist prior to review

Gateway has a recommendation to configure RequestHeadersTimeout. Also, gateways timeouts have different recommendation for defaults.


lukidzi:
I need this change https://github.com/kumahq/kuma/pull/8892 because here mostly we want to configure inbound listnener

pkg/plugins/policies/core/defaults/consts.go
```diff
@@ -10,6 +10,11 @@ const (
 	DefaultIdleTimeout           = time.Hour
 	DefaultStreamIdleTimeout     = 30 * time.Minute
 	DefaultRequestTimeout        = 15 * time.Second
+	DefaultRequestHeadersTimeout = 0
 	DefaultMaxStreamDuration     = 0
 	DefaultMaxConnectionDuration = 0
+	// Gateway
+	DefaultGatewayIdleTimeout           = 5 * time.Minute
+	DefaultGatewayStreamIdleTimeout     = 5 * time.Second
+	DefaultGatewayRequestHeadersTimeout = 500 * time.Millisecond
```
bartsmykla, line 19: I think if we are putting these defaults here, we should get rid of these and just reuse ones you just added https://github.com/kumahq/kuma/blob/8cc520ea4c14e20b386702fcf41056c66d44f33b/pkg/plugins/runtime/gateway/filter_chain_generator.go#L56-L58

pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go
```diff
@@ -148,6 +159,58 @@ func ConfigureRouteAction(
...
+			policies_defaults.DefaultGatewayRequestHeadersTimeout,
+		)
+		hcm.StreamIdleTimeout = toProtoDurationOrDefault(
+			pointer.Deref(conf.Http).StreamIdleTimeout,
+			policies_defaults.DefaultGatewayStreamIdleTimeout,
+		)
+		if httpConf := pointer.Deref(conf.Http); httpConf.RequestTimeout != nil {
+			hcm.RequestTimeout = util_proto.Duration(httpConf.RequestTimeout.Duration)
+		}
+		return nil
```
bartsmykla, line 190: I'm kinda thinking now should we have logic for applying default timeouts in two places, here and: https://github.com/kumahq/kuma/blob/8cc520ea4c14e20b386702fcf41056c66d44f33b/pkg/plugins/runtime/gateway/filter_chain_generator.go#L205-L209

bartsmykla (commented):
I have few questions and suggestions

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
```diff
@@ -221,6 +223,23 @@ func applyToGateway(
...
+		var protocol core_mesh.Protocol
+		switch listenerInfo.Listener.Protocol {
+		case mesh_proto.MeshGateway_Listener_HTTP, mesh_proto.MeshGateway_Listener_HTTPS:
+			protocol = core_mesh.ProtocolHTTP
+		case mesh_proto.MeshGateway_Listener_TCP, mesh_proto.MeshGateway_Listener_TLS:
+			protocol = core_mesh.ProtocolTCP
+		}
+		if err := plugin_xds.ConfigureGatewayListener(conf, protocol, gatewayListeners[key]); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
michaelbeaumont: Yea true, if we have `ConfigureGatewayListener` it can accept `mesh_proto.MeshGateway_Listener_Protocol` instead of `core_mesh.Protocol`

bartsmykla (approved):
lgtm

pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go
lukidzi: One thought occurred to me: Previously, users were unable to configure these timeouts, and if the default configuration did not suit them, they had to use a ProxyTemplate to modify it. By establishing defaults in the plugin code, we risk overriding the user's preferences. I am thinking about tweaking the logic to only update values if the current ones are the default settings.

lukidzi: I've tested it locally it seems ProxyTemplate is applied later for Gateway. It's kinda interesting because it's applied even after MeshProxyPatch

PR #8911, opened 2024-01-18
chore(kds): stop using deprecated otelgrpc.UnaryServerInterceptor

michaelbeaumont:
We were intentionally using the deprecated version but [the issue](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4576) seems to have been fixed by https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4577 which was released in `v0.46.1` which we're using

### Checklist prior to review

### DIFF
--- a/pkg/kds/mux/server.go
+++ b/pkg/kds/mux/server.go
@@ -131,8 +131,7 @@ func (s *server) Start(stop <-chan struct{}) error {
 	grpcOptions = append(
 		grpcOptions,
 		grpc.ChainUnaryInterceptor(s.unaryInterceptors...),
-		//nolint:staticcheck
-		grpc.ChainUnaryInterceptor(otelgrpc.UnaryServerInterceptor()),
+		grpc.StatsHandler(otelgrpc.NewServerHandler()),
 	)
 	grpcServer := grpc.NewServer(grpcOptions...)
 


pkg/kds/mux/server.go
```diff
@@ -131,8 +131,7 @@ func (s *server) Start(stop <-chan struct{}) error {
 	grpcOptions = append(
 		grpcOptions,
 		grpc.ChainUnaryInterceptor(s.unaryInterceptors...),
-		//nolint:staticcheck
-		grpc.ChainUnaryInterceptor(otelgrpc.UnaryServerInterceptor()),
+		grpc.StatsHandler(otelgrpc.NewServerHandler()),
```
lahabana, line 134: I remember us rolling back to this because of an issue in grpc. This is fixed now?

michaelbeaumont: Ah didn't realize that. In fact yes [the issue](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4576) seems to have been fixed by https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4577

michaelbeaumont: which was released in the version we're using

bartsmykla (approved):

PR #8912, opened 2024-01-18
ci(github): cleanup build-test-distribute action a little

lahabana:
There's a few unecessary, unclear or redundant things that we should remove

### Checklist prior to review

### DIFF
--- a/.github/workflows/build-test-distribute.yaml
+++ b/.github/workflows/build-test-distribute.yaml
@@ -87,8 +87,9 @@ jobs:
         run: |
           echo 'ENABLED_GOARCHES=arm64 amd64' >> $GITHUB_ENV
           echo 'ENABLED_GOOSES=linux darwin' >> $GITHUB_ENV
-      - run: |
-          echo '${ENABLED_GOARCHES}|${ENABKED_GOOSES}' > .run-full-matrix
+      - name: "Add matrix to .run-full-matrix for cache"
+        run: |
+          echo '${ENABLED_GOARCHES}|${ENABLED_GOOSES}' > .run-full-matrix
       - name: "Maybe set flag to push build artifacts"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-distribute.yaml
```diff
@@ -100,18 +81,15 @@ jobs:
...
-          echo 'RUN_FULL_MATRIX=true' > .run-full-matrix
-          echo 'RUN_FULL_MATRIX=true' >> $GITHUB_ENV
-      - name: "Set switches based on if run full matrix"
-        if: ${{ env.RUN_FULL_MATRIX == 'true' }}
         id: set-full-matrix-switches
         run: |
           echo 'ENABLED_GOARCHES=arm64 amd64' >> $GITHUB_ENV
           echo 'ENABLED_GOOSES=linux darwin' >> $GITHUB_ENV
-      - name: "Check if force push"
+      - run: |
```
michaelbeaumont, line 90: maybe worth always setting a `name` if what the `run` is doing isn't super obvious

.github/workflows/e2e.yaml
```diff
@@ -72,7 +64,7 @@ jobs:
       - if: steps.eval-params.outputs.run-type == 'github'
         run: |
           make dev/tools
-      - name: Free up disk space for the Runner
+      - name: "Github Actions: Free up disk space for the Runner"
```
michaelbeaumont, line 67: Why this?

michaelbeaumont (approved):

.github/workflows/e2e.yaml
lahabana: This job runs 2 options on github directly or through circleci. Other steps in this job are prefixed to make it easier to read

michaelbeaumont: OK I thought these names only show up in GH workflow runs, but :shrug:

Issue #4562, opened 2022-07-07
Deprecate support for `redirect-inbound-port-v6` 


lukidzi:
### What happened?

It seems that `redirect-inbound-port-v6` is not required and the listener for IPv4 and IPv6 might work on the same port. Instead of `redirect-inbound-port-v6` we can introduce a flag that allows disabling support for IPv6.

jijiechen:
Is `--ipv6-disabled` a good argument name to do this?

PR #8939, opened 2024-01-22
feat(transparent-proxy): deprecate argument 'redirect-inbound-port-v6' and introduce 'ip-family-mode'

jijiechen:
fixes https://github.com/kumahq/kuma/issues/4562

### Checklist prior to review

   https://github.com/kumahq/kuma/issues/4562
  - Confirmed
  - Unit tests
  - Updated
  - No need


pkg/plugins/runtime/k8s/webhooks/injector/testdata/inject.01.golden.yaml
```diff
@@ -10,7 +10,6 @@ metadata:
     kuma.io/transparent-proxying: enabled
     kuma.io/transparent-proxying-ebpf: disabled
     kuma.io/transparent-proxying-inbound-port: "15006"
-    kuma.io/transparent-proxying-inbound-v6-port: "15010"
```
jijiechen, line 13: this annotation will not be injected by default

pkg/plugins/runtime/k8s/webhooks/injector/testdata/inject.33.golden.yaml
```diff
@@ -0,0 +1,169 @@
...
+    kuma.io/mesh: default
+    kuma.io/sidecar-injected: "true"
+    kuma.io/sidecar-uid: "5678"
+    kuma.io/transparent-proxying: enabled
+    kuma.io/transparent-proxying-ebpf: disabled
+    kuma.io/transparent-proxying-inbound-port: "15006"
+    kuma.io/transparent-proxying-outbound-port: "15001"
+    kuma.io/virtual-probes: enabled
+    kuma.io/virtual-probes-port: "9000"
+    kuma.io/disable-ipv6: "true"
```
jijiechen, line 16: test we should honor this new annotation

pkg/plugins/runtime/k8s/webhooks/injector/testdata/inject.33.input.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: busybox
+  labels:
+    run: busybox
+  annotations:
+    kuma.io/disable-ipv6: "true"
```
jijiechen, line 8: test we should honor this new annotation

pkg/plugins/runtime/k8s/webhooks/injector/testdata/inject.34.input.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: busybox
+  labels:
+    run: busybox
+  annotations:
+    kuma.io/transparent-proxying-inbound-v6-port: "15010"
```
jijiechen, line 8: test we should still honor this deprecated annotation

pkg/transparentproxy/kubernetes/kubernetes_test.go
```diff
@@ -92,11 +92,11 @@ var _ = Describe("kubernetes", func() {
 				"--redirect-outbound-port", "25100",
 				"--redirect-inbound=" + "true",
 				"--redirect-inbound-port", "25204",
-				"--redirect-inbound-port-v6", "25206",
 				"--kuma-dp-uid", "12345",
 				"--exclude-inbound-ports", "12000",
 				"--exclude-outbound-ports", "11000",
 				"--verbose",
+				"--redirect-inbound-port-v6", "25206",
```
jijiechen, line 99: this argument is moved to tail when added manually by user via an annotation.

UPGRADE.md
```diff
@@ -33,16 +33,21 @@ With the release of Kuma 2.6.0, we've made some changes to the implementation of
...
+These are not valid units according to the [International System of Units](https://en.wikipedia.org/wiki/International_System_of_Units) they are respectively corrected to `Gbps` and `Mbps` if using
+these invalid units convert them into `kbps` prior to upgrade to avoid invalid format.
+
 ### Deprecation of postgres driverName=postgres (lib/pq)
 
 The postgres driver `postgres` (lib/pq) is deprecated and will be removed in the future.
 Please migrate to the new postgres driver `pgx` by setting `DriverName=pgx` configuration option or `KUMA_STORE_POSTGRES_DRIVER_NAME=pgx` env variable.
 
-### Make format SI valid for bandwidth in MeshFaultInjection policy
+### Deprecation of argument to transparent-proxy
```
bartsmykla, line 47: Wdyt?

```suggestion
### Deprecation of `--redirect-inbound-port-v6` flag and `kuma.io/transparent-proxying-inbound-v6-port` annotation
```

UPGRADE.md
```diff
@@ -33,16 +33,21 @@ With the release of Kuma 2.6.0, we've made some changes to the implementation of
...
 
-Prior to this upgrade `mbps` and `gbps` were used for units for parameter `conf.responseBandwidth.percentage`.
-These are not valid units according to the [International System of Units](https://en.wikipedia.org/wiki/International_System_of_Units) they are respectively corrected to `Gbps` and `Mbps` if using
-these invalid units convert them into `kbps` prior to upgrade to avoid invalid format.
+Parameter `--redirect-inbound-port-v6` is deprecated and will be removed in the future. The proxy will now redirect IPv6 traffic using the same port (15006) as ipv4 traffic, if you want to disable traffic redirection for IPv6 traffic, please use `--disable-ipv6`.
+Hence, the matching annotation `kuma.io/transparent-proxying-inbound-v6-port` is now deprecated and we've also added a new Kubernetes annotation `kuma.io/disable-ipv6` for disabling IPv6 traffic redirection on Kubernetes pods.
```
bartsmykla, line 50: Wdyt?

```suggestion
The `--redirect-inbound-port-v6` flag and the corresponding annotation `kuma.io/transparent-proxying-inbound-v6-port` are deprecated and will be removed in a future release of Kuma. These flags and annotations were used to configure the port used for redirecting IPv6 traffic to Kuma.

In the upcoming release, Kuma will redirect IPv6 traffic to the same port as IPv4 traffic (15006). This means that you no longer need to configure a separate port for IPv6 traffic.

If you want to disable traffic redirection for IPv6 traffic, you can use the `--disable-ipv6` flag.

We have also added a new Kubernetes annotation `kuma.io/disable-ipv6` for disabling IPv6 traffic redirection on Kubernetes pods. This annotation can be applied to any pod that you want to exclude from IPv6 traffic redirection.

We recommend that you update your configurations to use the new defaults for IPv6 traffic redirection. If you need to retain separate ports for IPv4 and IPv6 traffic, you can continue to use the deprecated flags and annotations until they are removed.
```

bartsmykla (commented):
Few suggestions around UPGRADE.md.

UPGRADE.md
```diff
@@ -33,16 +33,21 @@ With the release of Kuma 2.6.0, we've made some changes to the implementation of
 
 To ensure a smooth transition to Kuma 2.6.0, carefully review your existing configuration files and make necessary adjustments related to denied request responses and RBAC-related Envoy stats.
 
+### Make format SI valid for bandwidth in MeshFaultInjection policy
```
slonka, line 36: Nitpick: 

```suggestion
### Make SI format valid for bandwidth in MeshFaultInjection policy
```

slonka (commented):
Just one nitpick.

UPGRADE.md
jijiechen: This part is not directly related to task at this PR. What I've done was shifted this section above. 
I'll try to apply the suggestions, but want the original author to be informed. @lahabana

UPGRADE.md
jijiechen: I was aware the annotation was not intended to be set by the user, but by our `injector.go`, the user does not have a way to exclude certain Kubernetes pod.

jijiechen: I'll update this section and reflect the latest code.

UPGRADE.md
```diff
@@ -80,16 +80,24 @@ With the release of Kuma 2.6.0, we've made some changes to the implementation of
 
 To ensure a smooth transition to Kuma 2.6.0, carefully review your existing configuration files and make necessary adjustments related to denied request responses and RBAC-related Envoy stats.
 
+### Make SI format valid for bandwidth in MeshFaultInjection policy
```
jijiechen, line 83: This section was shifted up from below.

UPGRADE.md
```diff
@@ -80,7 +80,7 @@ With the release of Kuma 2.6.0, we've made some changes to the implementation of
 
 To ensure a smooth transition to Kuma 2.6.0, carefully review your existing configuration files and make necessary adjustments related to denied request responses and RBAC-related Envoy stats.
 
-### Make format SI valid for bandwidth in MeshFaultInjection policy
+### Make SI format valid for bandwidth in MeshFaultInjection policy
```
slonka, line 83: this is not targetting 2.6 and it's in `UPGRADE.md` https://github.com/kumahq/kuma/blob/master/UPGRADE.md#make-format-si-valid-for-bandwidth-in-meshfaultinjection-policy - or am I confused about something?

UPGRADE.md
slonka: is SI the international term for it? We're talking about this - https://en.wikipedia.org/wiki/International_System_of_Units right?

app/cni/pkg/cni/annotations.go
```diff
@@ -94,6 +97,20 @@ func validatePortList(ports string) error {
 	return nil
 }
 
+func validateBool(val string) error {
+	if val == "" {
+		return errors.New("value is empty")
+	}
+
+	validValues := []string{"yes", "no", "true", "false"}
```
slonka, line 105: is this  the first time we're allowing "yes" for boolean? this should be consistent across all toggles and we shouldn't add it just here

app/cni/pkg/cni/annotations.go
```diff
@@ -144,3 +164,13 @@ func mapAnnotation(annotations map[string]string, field *string, fieldName strin
 	*field = val
 	return nil
 }
+
+func assignIPv6InboundRedirectPort(allFields map[string]*string) {
+	fieldPointer := allFields["inboundPortV6"]
+	disabledAnno := *allFields["ipv6Disabled"]
```
slonka, line 170: is it `annotation`? also it's a pointer so a possible NPE here, we should use comma-ok check

app/kumactl/cmd/install/install_transparent_proxy.go
```diff
@@ -168,10 +168,11 @@ runuser -u kuma-dp -- \
 				}
 			}
 
+			defaultCfg := config.DefaultConfig()
 			// Backward compatibility
 			if args.RedirectPortInBoundV6 != "" &&
-				args.RedirectPortInBoundV6 != "15006" /* new default value, identical to ipv4 port */ &&
-				args.RedirectPortInBoundV6 != "15010" /* old default value, dedicated for ipv6 */ {
+				args.RedirectPortInBoundV6 != fmt.Sprintf("%d", defaultCfg.Redirect.Inbound.Port) /* new default value, identical to ipv4 port */ &&
```
slonka, line 174: how is sprintf better here?

pkg/plugins/runtime/k8s/metadata/annotations_test.go
```diff
@@ -55,6 +55,53 @@ var _ = Describe("Kubernetes Annotations", func() {
...
+	Describe("GetBoolean()", func() {
+		type testCase struct {
+			input    string
+			expected bool
+		}
+		annotations := map[string]string{
+			"key1": "yes",
+			"key2": "no",
+			"key3": "true",
+			"key4": "false",
```
slonka, line 67: probably need an error cases like:

```suggestion
			"key4": "false",
			"key5": "foo",
```

slonka (commented):
I think there is a change to get an NPE

app/cni/pkg/cni/annotations.go
jijiechen: Yes, this can be the first time introducing "yes/no". Right, agreed, we'd better make it consistent across the product. I search over the whole product and found only "enabled/disabled" for things like enabling the sidecar injector. 
I personally think adding "yes/no" is more intuitive when setting boolean values. Here it is setting a boolean value rather than an "enabled state", in fact, the name of the switch here contains a "disable", so it will be misleading to reuse "enabled/disabled" in this case.

app/cni/pkg/cni/annotations.go
jijiechen: The value was originated from an annotation from the parent level of callsites, and they are parsed from pod annotations.

app/kumactl/cmd/install/install_transparent_proxy.go
jijiechen: On, it's not Sprintf, but we read the value from `defaultCfg`, instead of a literal.

pkg/plugins/runtime/k8s/metadata/annotations_test.go
jijiechen: I'll add, thanks!

UPGRADE.md
jijiechen: I'm not sure, actually. Because the orginal code block was not added by me and I don't know what was it.

app/cni/pkg/cni/annotations.go
jijiechen: Had a chat with Charly, I'll change the argument to `ipv6`, so we can reuse existing values like `enabled/disabled`

jijiechen:
I'll make it a draft at the moment since there is still some work to do.

jijiechen:
> I'll make it a draft at the moment since there is still some work to do.
This is done now.

jakubdyszkiewicz (approved):
All of this works of course, but if we can expose ipv4 and v6 on the same port wouldn't it be simpler to just change the default here?

```
        # Redirect port for inbound traffic.
        redirectPortInbound: 15006 # ENV: KUMA_RUNTIME_KUBERNETES_INJECTOR_SIDECAR_CONTAINER_REDIRECT_PORT_INBOUND
        # Redirect port for inbound traffic.
        redirectPortInboundV6: 15010 # ENV: KUMA_RUNTIME_KUBERNETES_INJECTOR_SIDECAR_CONTAINER_REDIRECT_PORT_INBOUND_V6
```
so v6 and v4 is on the same port?

jijiechen:
> wouldn't it be simpler to just change the default here?

Technically yes, but it needs people to configure it and they don't know what will be the side effects of doing so. So I guess poeple would normally not do this.
And we'll help them get rid of this unnecessary configuration item.

PR #8945, opened 2024-01-22
fix(kuma-cp): set creationTime on KDS sync

lobkovilya:
Resource meta doesn't have creationTime after KDS sync so we can't call `r.GetMeta().GetCreationTime()`.

### Checklist prior to review

### DIFF
--- a/pkg/kds/store/sync.go
+++ b/pkg/kds/store/sync.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/go-logr/logr"
 
+	"github.com/kumahq/kuma/pkg/core"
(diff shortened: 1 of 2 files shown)


pkg/kds/store/sync.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/go-logr/logr"
 
+	"github.com/kumahq/kuma/pkg/core"
```
lahabana, line 10: I'd love to avoid depending on this global time. It makes unit tests harder. We should have the clock passed as a param

michaelbeaumont: You can also usually write the tests in such a way as to not depend on the time itself

PR #8977, opened 2024-01-22
chore(mk): fix make kind/start

kleinfreund:
### Checklist prior to review

Fix `kind create cluster --config` argument value to refer to existing cluster.yaml file. Without this, `make kind/start` fails with `No kind clusters found. make: *** [mk/kind.mk:32: kind/start] Error 1` but actually, the issues is `ERROR: failed to create cluster: error reading file: open ./test/kind/cluster-kuma.yaml: no such file or directory` which isn’t shown due to a `--quiet` argument on the `kind create cluster` command.

See also https://github.com/kumahq/kuma/pull/8676.

### DIFF
--- a/mk/kind.mk
+++ b/mk/kind.mk
@@ -11,9 +11,9 @@ KIND_KUBECONFIG := $(KIND_KUBECONFIG_DIR)/kind-$(KIND_CLUSTER_NAME)-config
 unexport KUBECONFIG
 
 ifdef IPV6
-KIND_CONFIG ?= $(KUMA_DIR)/test/kind/cluster-ipv6.yaml
+KIND_CONFIG ?= $(KUMA_DIR)/test/kind/cluster-ipv6-$(KIND_CLUSTER_NAME)-1.yaml
 else
-KIND_CONFIG ?= $(KUMA_DIR)/test/kind/cluster.yaml
+KIND_CONFIG ?= $(KUMA_DIR)/test/kind/cluster-$(KIND_CLUSTER_NAME)-1.yaml
 endif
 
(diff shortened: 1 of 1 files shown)


mk/kind.mk
```diff
@@ -32,7 +26,7 @@ kind/start: ${KUBECONFIG_DIR}
 	$(KIND) get clusters | grep $(KIND_CLUSTER_NAME) >/dev/null 2>&1 && echo "Kind cluster already running." && exit 0 || \
 		($(KIND) create cluster \
 			--name "$(KIND_CLUSTER_NAME)" \
-			--config "$(KUMA_DIR)/test/kind/cluster-$(if $(IPV6),ipv6-,)$(KIND_CLUSTER_NAME).yaml" \
+			--config "$(KUMA_DIR)/test/kind/cluster$(if $(IPV6),-ipv6,).yaml" \
```
kleinfreund, line 29: **Question**: What about the other two sets of files (i.e. the ones with `-kuma-1` and `-kuma-2` suffix)?

lahabana: Should this just be `$(KIND_CONFIG)` and the lines above kept and test/kind-cluster.yaml become test/kind-cluster-kuma.yaml ?

kleinfreund: I made this change, but I needed to pick between `cluster-ipv6-kuma-1.yaml` and `cluster-ipv6-kuma-2.yaml`. Arbitrarily picked `cluster-ipv6-kuma-1.yaml`. What should we use here?

bartsmykla (approved):

Issue #9005, opened 2024-01-23
Improve `kumactl export`


slonka:
### Description

xrel: https://github.com/kumahq/kuma/issues/8212

Things left to change:
- add a line number where the problem is - otherwise it's hard to figure out

Everything from here was done in https://github.com/kumahq/kuma/pull/9009 :
- fix NPE when trying to export in k8s format
- spec field missing on k8s format for new policies
- rename IncludeInFederationWithPolicies to IncludeInFederation
- remove ServiceInsight from IncludeInFederation
- change federation profile to export everything based on IncludeInFederation minus policies from targetRef, minus MeshGateway

slonka:
Closing and opening a new issue for the remaining one point.

PR #9009, opened 2024-01-23
feat(kumactl): improve kumactl export

slonka:
xrel https://github.com/kumahq/kuma/issues/9005

Solves:
- [x] fix NPE when trying to export in k8s format
- [x] spec field missing on k8s format for new policies
- [ ] ~~add a line number where the problem is - otherwise it's hard to figure out~~
- [x] rename IncludeInFederationWithPolicies to IncludeInFederation
- [x] remove ServiceInsight from IncludeInFederation
- [x] change federation profile to export everything based on IncludeInFederation minus policies from targetRef, minus MeshGateway

### Checklist prior to review

> Changelog: feat(kumactl): basic export command


app/kumactl/cmd/export/export_test.go
```diff
@@ -36,7 +36,8 @@ var _ = Describe("kumactl export", func() {
 
 	BeforeEach(func() {
 		store = core_store.NewPaginationStore(memory_resources.NewStore())
-		rootCtx, err := test_kumactl.MakeRootContext(rootTime, store, mesh.MeshResourceTypeDescriptor, system.SecretResourceTypeDescriptor, system.GlobalSecretResourceTypeDescriptor)
+		defs := registry.Global().ObjectDescriptors()
```
slonka, line 39: otherwise we can't base our tests `includeInFederation`

app/kumactl/cmd/export/export.go
```diff
@@ -180,3 +180,9 @@ func resourcesTypesToDump(ctx context.Context, ectx *exportContext) ([]model.Res
 	}
 	return resTypes, nil
 }
+
+func includeInFederationProfile(res api_common.ResourceTypeDescription) bool {
+	return res.IncludeInFederation && // base decision on `IncludeInFederation` field
+		(res.Policy == nil || (res.Policy != nil && !res.Policy.IsTargetRef)) && // do not include new policies (q: should this just be reflected in IncludeInFederation?)
```
slonka, line 186: mostly question to @jakubdyszkiewicz

app/kumactl/cmd/export/testdata/export-kube.golden.yaml
```diff
@@ -1,12 +1,3 @@
-apiVersion: kuma.io/v1alpha1
```
slonka, line 1: order change because I'm using filter function

app/kumactl/cmd/export/testdata/export-kube.golden.yaml
```diff
@@ -7,12 +17,56 @@ metadata:
...
+  annotations:
+    kubectl.kubernetes.io/last-applied-configuration: |
+      {"apiVersion":"kuma.io/v1alpha1","kind":"MeshAccessLog","metadata":{"annotations":{},"labels":{"kuma.io/mesh":"default"},"name":"default","namespace":"kuma-system"},"spec":{"from":[{"default":{"backends":[{"file":{"path":"/tmp/access.log"},"type":"File"}]},"targetRef":{"kind":"Mesh"}}],"targetRef":{"kind":"Mesh"},"to":[{"default":{"backends":[{"file":{"path":"/tmp/access.log"},"type":"File"}]},"targetRef":{"kind":"Mesh"}}]}}
+  creationTimestamp: "2024-01-24T09:52:46Z"
+  labels:
+    kuma.io/mesh: default
+  name: default
   namespace: kuma-system
-type: system.kuma.io/global-secret
+spec:
```
slonka, line 31: spec is present

app/kumactl/cmd/export/export.go
jakubdyszkiewicz: I don't think it should. If you set `includeInFederation:true` to target ref policies then you would need to filter this out in profileFederation

pkg/test/resources/samples/serviceinsights.go
```diff
@@ -0,0 +1,21 @@
+package samples
+
+import (
+	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
+	"github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
+	"github.com/kumahq/kuma/pkg/test/resources/builders"
+)
+
+func ServiceInsight(mesh string) *mesh.ServiceInsightResource {
```
jakubdyszkiewicz, line 9: let's not add arguments like this to samples. You add one argument, then in a couple of months there will be 5 arguments there. Then you have tests that look like this `ServiceInsight("default", 2, true, false, "something)` and adding any new argument means that you need to go over all usages and fix it.
If you want to parameterize anything then you should use builder functions.

do
```
func SampleServiceInsightBuilder() *ServiceInsightBuilder {
	return ServiceInsightBuilderBuilder().
		AddService("test-service", &mesh_proto.ServiceInsight_Service{
			ServiceType: mesh_proto.ServiceInsight_Service_internal,
			Status:      mesh_proto.ServiceInsight_Service_online,
		})
}
```
and call
```
SampleServiceInsightBuilder().WithMesh(mesh).Build()
```

jakubdyszkiewicz: but kudos that you actually added builders and samples to this PR. Thanks for that

app/kumactl/cmd/export/export.go
```diff
@@ -91,7 +85,11 @@ $ kumactl export --profile federation --format universal > policies.yaml
 				}
 				if resDesc.Scope == model.ScopeGlobal {
 					list := resDesc.NewList()
-					if err := rs.List(cmd.Context(), list); err != nil {
+					// filter out envoy-admin-ca and inter-cp-ca otherwise it will cause TLS handshake errors
+					if err := rs.List(cmd.Context(), list, store.ListByFilterFunc(func(rs model.Resource) bool {
```
jakubdyszkiewicz, line 89: please double check that it works with remote store on a real system. Becuase I think it does not. I'd filter this after we retrieve the list

slonka: this is how I checked:

```
build/artifacts-darwin-arm64/kumactl/kumactl export --format kubernetes | grep -i serviceinsight

```

```
build/artifacts-darwin-arm64/kumactl/kumactl export --format kubernetes --profile all | grep -i serviceinsight
type: ServiceInsight
type: ServiceInsight
type: ServiceInsight
```

pkg/test/resources/samples/serviceinsights.go
slonka: I was meant to clean up this example, thanks for pointing it out.

app/kumactl/cmd/export/export.go
slonka: not it is filtered out 

```
build/artifacts-darwin-arm64/kumactl/kumactl export --format kubernetes | grep -i envoy-admin-ca

```

jakubdyszkiewicz (approved):

PR #9012, opened 2024-01-23
fix(MeshHTTPRoute): isolate listeners

michaelbeaumont:
The idea behind this PR is to make sure the effects of MeshHTTPRoutes attached to MeshGateway listeners are limited to those listeners, thus the need to map `(address, port, hostname)` to routes instead of `(address, port)`.

### Checklist prior to review

### DIFF
--- a/pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/01.golden.yaml
+++ b/pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/01.golden.yaml
@@ -39,67 +39,133 @@ FromRules:
       type: MeshAccessLog
     Subset: []
 ToRules:
-  127.0.0.1:8080:
-  - Conf:
-      backends:
-      - file:
-          path: /gateway-listener
-        type: File
-    Origin:
-    - creationTime: "0001-01-01T00:00:00Z"
-      mesh: mesh-1
-      modificationTime: "0001-01-01T00:00:00Z"
-      name: gateway
-      type: MeshAccessLog
-    - creationTime: "0001-01-01T00:00:00Z"
-      mesh: mesh-1
-      modificationTime: "0001-01-01T00:00:00Z"
-      name: gatewaylistener
-      type: MeshAccessLog
-    - creationTime: "0001-01-01T00:00:00Z"
-      mesh: mesh-1
-      modificationTime: "0001-01-01T00:00:00Z"
-      name: mesh
-      type: MeshAccessLog
-    Subset: []
-  127.0.0.1:8081:
-  - Conf:
-      backends:
-      - file:
-          path: /servicesubset
-        type: File
-    Origin:
-    - creationTime: "0001-01-01T00:00:00Z"
-      mesh: mesh-1
-      modificationTime: "0001-01-01T00:00:00Z"
-      name: gateway
-      type: MeshAccessLog
-    - creationTime: "0001-01-01T00:00:00Z"
-      mesh: mesh-1
-      modificationTime: "0001-01-01T00:00:00Z"
-      name: mesh
-      type: MeshAccessLog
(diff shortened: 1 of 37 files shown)


pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/02.policies.yaml
```diff
@@ -3,68 +3,96 @@ type: MeshAccessLog
...
+    - port: 8082
+      protocol: HTTP
+      tags:
+        listener: three
+    - port: 8083
+      protocol: HTTP
+      hostname: four-hostname-1
+      tags:
+        listener: four-hostname-1
+    - port: 8083
```
bartsmykla, line 94: Is it intentional to have the same port as a listener above?

pkg/plugins/runtime/gateway/match/hostnames.go
```diff
@@ -35,6 +35,9 @@ func (h *hostname) contains(n hostname) bool {
 }
 
 func makeHostname(name string) hostname {
+	if name == "" {
+		name = "*"
+	}
```
slonka, line 40: we have a bunch of these scattered around the codebase, should we make them point to this function?

pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/01.golden.yaml
```diff
@@ -39,67 +39,133 @@ FromRules:
...
+        modificationTime: "0001-01-01T00:00:00Z"
+        name: gateway
+        type: MeshAccessLog
+      - creationTime: "0001-01-01T00:00:00Z"
+        mesh: mesh-1
+        modificationTime: "0001-01-01T00:00:00Z"
+        name: mesh
+        type: MeshAccessLog
+      Subset: []
+  ByListenerAndHostname:
```
slonka, line 107: Isn't just `ByListener` a `ByListenerAndHostname` with `*` as a hostname? Does having two entries give us anything else?

pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/01.golden.yaml
```diff
@@ -39,67 +39,133 @@ FromRules:
...
+    127.0.0.1:8080:*:
```
slonka, line 108: Isn't this notation problematic for ipv6? should we use a different delimiter?

slonka (commented):
I understand the idea and agree with it, left a couple of Qs.

Automaat (approved):

pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/01.golden.yaml
michaelbeaumont: In some sense yes, but the key thing is that it depends on the policy whether targeting hostnames are supported and currently only MeshHTTPRoute supports it. So for MeshHTTPRoute only `ByListenerAndHostname` is set and for other policies only `ByListener`. It makes the distinction a little clearer that for other policies the hostname can _only_ be `*`.

pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/01.golden.yaml
michaelbeaumont: We only serialize this for tests

slonka (approved):

pkg/plugins/runtime/gateway/match/hostnames.go
michaelbeaumont: tried to reduce the `"" => "*"` logic in the codebase

Issue #9011, opened 2024-01-23
Cannot call admin api when Global is on 2.6 and running on Kubernetes


lukidzi:
### What happened?

I've tested a few scenarios 

Global on K8s and Zone on Universal  = works
Global on K8s and Zone on K8s  - don't work with the old version(2.5.1) and with the new version of the control-plane

Global on Universal - both works 

It seems like this part of the code https://github.com/kumahq/kuma/blame/master/pkg/plugins/resources/k8s/store.go#L276 breaks the behavior

PR #9014, opened 2024-01-23
fix(kds): fix resource sync

lukidzi:
### Checklist prior to review

There were few issues when running Global on K8s
- `k8s.kuma.io/namespace` was always set to `kuma-system` on global and couldn't sync data
- ZoneIngress namespace was changed all the time and when caused that zone resource was synced all the time in the loop
- I've tested the code with 
  - Global on K8s and K8s Zone 2.5.1, K8s zone Snapshot, Universal Snapshot and Universal 2.5.1 and I had no issues

I've introduced a new value in the mapper. Since 2.6 we cannot distinguish if the resource is from universal so we were always adding `kuma-system` to requests on the zone. That caused that we cannot get the response from the zone stats

- [x] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/9011


bartsmykla (approved):
LGTM

jakubdyszkiewicz:
please wait with merging. I want to look at it

jakubdyszkiewicz (changes requested):
I'm not a fan of this. I'd try to avoid adding a new label to every single resource. Could we fetch ZoneInsight in `resNameInZone` and check store type from CP config?

pkg/plugins/resources/k8s/store.go
```diff
@@ -273,9 +273,10 @@ func (m *KubernetesMetaAdapter) GetLabels() map[string]string {
 	if _, ok := labels[v1alpha1.DisplayName]; !ok {
 		labels[v1alpha1.DisplayName] = m.GetObjectMeta().GetName()
 	}
-	if m.Namespace != "" {
+	if _, ok := labels[v1alpha1.KubeNamespaceTag]; !ok {
```
jakubdyszkiewicz, line 276: I assume you want to preserve a namespace from the Zone CP. Is that correct?

I think we still need `if m.Namespace != "" {`. Otherwise, cluster-scoped objects (old policies etc.) will receive a label with an empty value.

lukidzi: Yes, that was my idea to preserve the namespace. In case of calling this code on global it was always overriding the namespace and causing that on k8s calling the stats endpoint was incorrect. There was always a request with a namespace of the global sent to the zone.

lukidzi:
> I'm not a fan of this. I'd try to avoid adding a new label to every single resource. Could we fetch ZoneInsight in `resNameInZone` and check store type from CP config?

I think we can, the only thing is that we might need to take the first subscription (or iterate), and map string config to object might be more complicated but I think we can do this

pkg/kds/context/context.go
```diff
@@ -69,7 +70,10 @@ func DefaultContext(
 			),
 			MapZoneTokenSigningKeyGlobalToPublicKey),
 		reconcile.If(
-			reconcile.IsKubernetes(cfg.Store.Type),
+			reconcile.And(
+				reconcile.IsKubernetes(cfg.Store.Type),
+				reconcile.Not(reconcile.TypeIs(mesh.ZoneIngressType)),
```
jakubdyszkiewicz, line 75: Could you help me understand why we need this change?

pkg/config/util.go
```diff
@@ -23,3 +23,11 @@ func ToJson(cfg Config) ([]byte, error) {
...
 	return yaml.YAMLToJSON(yamlBytes)
 }
+
+func FromJson(cfg string, config Config) error {
+	yamlConfig, err := yaml.JSONToYAML([]byte(cfg))
+	if err != nil {
+		return err
+	}
+	return FromYAML(yamlConfig, config)
+}
```
jakubdyszkiewicz, line 33: nit: shouldn't FromYAML just work in the end? YAML is a superset of JSON in the end

pkg/kds/context/context.go
lukidzi: If I noticed that correctly, when we have 2 zones: one old and one new, we can observe that every reconciliation tick zone ingress changes the name:

```
 2024-01-24T16:08:23.305Z    INFO    kds-delta-zone    creating a new resource from upstream    {"type": "ZoneIngress", "name": "zone-251.kuma-ingress-85df4875bd-rnkml.kuma-system", "mesh": ""}                   │
│ 2024-01-24T16:08:24.292Z    INFO    kds-delta-zone    deleting a resource since it's no longer available in the upstream    {"type": "ZoneIngress", "name": "zone-251.kuma-ingress-85df4875bd-rnkml.kuma-system",  │
│ 2024-01-24T16:08:24.299Z    INFO    kds-delta-zone    creating a new resource from upstream    {"type": "ZoneIngress", "name": "zone-251.kuma-ingress-85df4875bd-rnkml.kuma-system.kuma-system", "mesh": ""}       │
│ 2024-01-24T16:08:26.294Z    INFO    kds-delta-zone    deleting a resource since it's no longer available in the upstream    {"type": "ZoneIngress", "name": "zone-251.kuma-ingress-85df4875bd-rnkml.kuma-system.ku │
│ 2024-01-24T16:08:26.310Z    INFO    kds-delta-zone    creating a new resource from upstream    {"type": "ZoneIngress", "name": "zone-251.kuma-ingress-85df4875bd-rnkml.kuma-system", "mesh": ""}```

That causes issue with constant changes being propagated

Comment (jakubdyszkiewicz) [reply pkg/kds/context/context.go:75]: I understand what you see, but I don't understand the cause. I'm not sure that's the right approach. Let me test a thing or two...

Comment (lukidzi) [reply pkg/kds/context/context.go]: We did a test and the cause was that we were not cloning an object after trimming a suffix and instead we were changing an object in the resource cache. Each iteration had a different name and that caused the issue

File: pkg/kds/context/context_test.go [hunk 4]
```diff
@@ -476,10 +477,14 @@ var _ = Describe("Context", func() {
 		}
 
 		resource := func(given testCase) model.Resource {
-			descriptor := model.ResourceTypeDescriptor{
-				IsPluginOriginated: given.isResourcePluginOriginated,
-				Resource:           core_mesh.NewCircuitBreakerResource(),
-				Scope:              given.scope,
+			// var descriptor model.ResourceTypeDescriptor
```
jakubdyszkiewicz, line 480: leftover?

jakubdyszkiewicz (approved):

PR #9033, opened 2024-01-25
fix(kuma-cp): use pagination store for secret store

lukidzi:
### Checklist prior to review

SecretStore wasn't using a pagination store so there was a risk that resources might not be sorted. Changed SecretStore to use pagination store.

### DIFF
--- a/pkg/plugins/secrets/k8s/store.go
+++ b/pkg/plugins/secrets/k8s/store.go
@@ -3,7 +3,6 @@ package k8s
 import (
 	"context"
 	"fmt"
-	"sort"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/plugins/secrets/k8s/store.go
```diff
@@ -302,9 +301,6 @@ func (c *SimpleConverter) ToCoreList(in *kube_core.SecretList, out core_model.Re
 			}
 			secOut.Items[i] = r
 		}
-		sort.SliceStable(secOut.Items, func(i, j int) bool {
```
slonka, line 305: to get this right, the previous store could have partial response here and this is how it was not sorted?

slonka (commented):
does a pagination store enforce sorting?

lukidzi:
> does a pagination store enforce sorting?

Yes, https://github.com/kumahq/kuma/blob/master/pkg/core/resources/store/pagination_store.go#L82

pkg/plugins/secrets/k8s/store.go
lukidzi: Not sure if that could be the case. I think we may have overlooked sorting in this situation due to the delegation model, where tasks are delegated from one entity to another. The secret store wasn't positioned behind the pagination store, which ensures that results are sorted.

slonka (approved):

PR #9062, opened 2024-01-29
WIP: perf(vips): reduce DB calls for CreateOrUpdateVIPConfigs

nicoche:
`CreateOrUpdateVIPConfigs` goes over every mesh and each time queries resources from the store. With more than a few meshes, it:
  * can be pretty heavy on the database
  * can be pretty slow

This changes the behavior of `CreateOrUpdateVIPConfigs` to pull all necessary resources beforehand. They are then grouped by mesh in memory and used as if they were pulled from the store.

With less than 200 meshes, this speeds up VIP generation from 5-20s to <= 500ms.

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- no issue
- [x] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? -- no
- [x] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) -- no


pkg/dns/vips_allocator.go
```diff
@@ -199,63 +265,113 @@ func addFromMeshGateway(outboundSet *vips.VirtualOutboundMeshView, dnsSuffix, me
...
+	out := map[string][]*core_mesh.MeshGatewayResource{}
+	for _, meshGateway := range meshGateways.Items {
+		mesh := meshGateway.Meta.GetMesh()
+		out[mesh] = append(out[mesh], meshGateway)
+	}
+
+	return out, nil
+}
+
+func (d *VIPsAllocator) buildVirtualOutboundMeshView(mesh string, virtualOutbounds []*core_mesh.VirtualOutboundResource, dataplanes []*core_mesh.DataplaneResource, zoneIngresses []*core_mesh.ZoneIngressResource, externalServices []*core_mesh.ExternalServiceResource, meshGatewaysByMesh map[string][]*core_mesh.MeshGatewayResource) (*vips.VirtualOutboundMeshView, error) {
```
nicoche, line 336: That's a lot of parameters. We could define a struct holding them if you think that's necessary for lisibility

pkg/dns/vips/persistence.go
```diff
@@ -47,42 +47,45 @@ func NewPersistence(resourceManager manager.ReadOnlyResourceManager, configManag
 	}
 }
 
-func (m *Persistence) GetByMesh(ctx context.Context, mesh string) (*VirtualOutboundMeshView, error) {
-	name := fmt.Sprintf(template, mesh)
-	resource := config_model.NewConfigResource()
-	err := m.configManager.Get(ctx, resource, store.GetByKey(name, ""))
+func (m *Persistence) Get(ctx context.Context, meshes core_mesh.MeshResourceList) (map[string]*VirtualOutboundMeshView, error) {
```
nicoche, line 50: I'm adding a test for this one

pkg/dns/vips_allocator.go
jakubdyszkiewicz: it's a private function. It's not ideal, but I think it's fine. The only thing is that I'd format this in this way
```
func (d *VIPsAllocator) buildVirtualOutboundMeshView(
  mesh string,
  virtualOutbounds []*core_mesh.VirtualOutboundResource,
  dataplanes []*core_mesh.DataplaneResource,
  zoneIngresses []*core_mesh.ZoneIngressResource,
  externalServices []*core_mesh.ExternalServiceResource,
  meshGatewaysByMesh map[string][]*core_mesh.MeshGatewayResource,
) (*vips.VirtualOutboundMeshView, error) {
```

jakubdyszkiewicz:
Interesting. The assumption was that because we use ReadOnlyResourceManager, the queries will be effectively cached between this and XDS reconciliation process. This component is run by leader, so only in the case where there is no DPP of a specific mesh connected to the leader, we would run additional queries. This has to be your case, because of the high number of meshes.

This change make sense for me, but I believe we should target master now. I merged release into master, so there should be no conflicts

pkg/dns/vips_allocator.go
nicoche: 🤝 done

jakubdyszkiewicz (approved):

PR #9082, opened 2024-01-31
fix(kuma-cp): don't run outbound proxy generator when there is no TrafficRoute

michaelbeaumont:
### Checklist prior to review

### DIFF
--- a/pkg/xds/generator/outbound_proxy_generator.go
+++ b/pkg/xds/generator/outbound_proxy_generator.go
@@ -30,7 +30,7 @@ const OriginOutbound = "outbound"
 type OutboundProxyGenerator struct{}
 
 func (g OutboundProxyGenerator) Generate(ctx context.Context, _ *model.ResourceSet, xdsCtx xds_context.Context, proxy *model.Proxy) (*model.ResourceSet, error) {
-	if len(proxy.Policies.TrafficRoutes) == 0 {
+	if len(proxy.Routing.TrafficRoutes) == 0 {
 		return nil, nil
 	}
 
--- a/pkg/xds/generator/outbound_proxy_generator.go
+++ b/pkg/xds/generator/outbound_proxy_generator.go
@@ -12,8 +12,6 @@ import (
 	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
 	"github.com/kumahq/kuma/pkg/core/user"
 	model "github.com/kumahq/kuma/pkg/core/xds"
-	meshhttproute_api "github.com/kumahq/kuma/pkg/plugins/policies/meshhttproute/api/v1alpha1"
-	meshtcproute_api "github.com/kumahq/kuma/pkg/plugins/policies/meshtcproute/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/xds/generator/outbound_proxy_generator.go
```diff
@@ -62,8 +64,12 @@ func (g OutboundProxyGenerator) Generate(ctx context.Context, _ *model.ResourceS
 
 		protocol := inferProtocol(xdsCtx.Mesh, clusters)
 		switch protocol {
-		case core_mesh.ProtocolHTTP, core_mesh.ProtocolHTTP2:
-			if hasMeshRoutes {
+		case core_mesh.ProtocolHTTP, core_mesh.ProtocolHTTP2, core_mesh.ProtocolGRPC:
+			if hasMeshHTTPRoutes {
+				continue
+			}
+		case core_mesh.ProtocolUnknown, core_mesh.ProtocolTCP:
```
lukidzi, line 71: I think kafka missing
```suggestion
		case core_mesh.ProtocolUnknown, core_mesh.ProtocolKafka, core_mesh.ProtocolTCP:
```

lukidzi: I think we have a code change in MeshTCPRoute(and test) which allows kafka protocol https://github.com/kumahq/kuma/blob/2df58666a6a80e39b78e7ac8a0f2b9c59f8e244b/pkg/plugins/policies/meshtcproute/plugin/v1alpha1/testdata/kafka.listeners.golden.yaml

pkg/xds/generator/outbound_proxy_generator.go
```diff
@@ -31,7 +30,9 @@ const OriginOutbound = "outbound"
 type OutboundProxyGenerator struct{}
 
 func (g OutboundProxyGenerator) Generate(ctx context.Context, _ *model.ResourceSet, xdsCtx xds_context.Context, proxy *model.Proxy) (*model.ResourceSet, error) {
-	hasMeshRoutes := len(proxy.Policies.Dynamic[v1alpha1.MeshHTTPRouteType].ToRules.Rules) > 0
+	if len(proxy.Routing.TrafficRoutes) == 0 {
```
lukidzi, line 33: We can use routing because we set `proxy.Policies.TrafficRoutes` based on the routing https://github.com/kumahq/kuma/blob/master/pkg/xds/sync/dataplane_proxy_builder.go#L45

lukidzi (approved):

pkg/xds/generator/outbound_proxy_generator.go
michaelbeaumont: Is this `> 0` exactly when there is _any_ `TrafficRoutes` in the mesh, not just ones that affect this `Dataplane`?

lukidzi: I think it's about this DP. Should we validate if there is any TrafficRoute in the mesh?

lahabana (approved):

PR #9094, opened 2024-01-31
perf(policies): if no trafficRoute don't read old policies

lahabana:
- Old policies are not usable if you don't have trafficRoute
- Therefore we don't put them in the MeshContext
- Refactored things to make clear what is used by old policies and that we shouldn't use anymore

This is a first step to get away from having proxy generation be in O(number of dataplanes).

### Checklist prior to review


pkg/core/xds/inspect/attachments_test.go
```diff
@@ -301,12 +302,12 @@ var _ = Describe("GroupByAttachment", func() {
...
 					"backend": &core_mesh.RetryResource{Meta: meta1},
 				},
-				Dynamic: map[core_model.ResourceType]core_xds.TypedMatchingPolicies{
-					core_mesh.TrafficLogType: {
-						ServicePolicies: map[core_xds.ServiceName][]core_model.Resource{
-							"redis": {
-								&core_mesh.TrafficLogResource{Meta: meta6},
-							},
+			},
+			policyPlugins: core_xds.PluginOriginatedPolicies{
```
lukidzi, line 306: Just a bit confusing that we are using `PluginOriginatedPolicies` with old policies

lukidzi (approved):

pkg/core/xds/inspect/attachments_test.go
lahabana: Are we?

pkg/api-server/inspect_endpoints_test.go
```diff
@@ -60,6 +60,22 @@ func serviceSelector(name, protocol string) *mesh_proto.Selector {
 	}
 }
 
+func defaultTrafficRoute(mesh string) *core_mesh.TrafficRouteResource {
```
slonka, line 63: nitpick: could've been a reusable `route-all-default` test resource - but we're not going to develop old policies more so I think this can stay.

pkg/core/xds/inspect/attachments_test.go
slonka: We're using both right? `GroupByAttachment` takes both `PluginOriginatedPolicies` (new policies) and `MatchedPolicies` (old policies) and then in (e.g.) `getInboundMatchedPolicies` we combine them.

pkg/plugins/policies/meshfaultinjection/plugin/v1alpha1/plugin.go
```diff
@@ -71,9 +71,6 @@ func applyToInbounds(
 	for _, inbound := range proxy.Dataplane.Spec.GetNetworking().GetInbound() {
 		iface := proxy.Dataplane.Spec.Networking.ToInboundInterface(inbound)
 		protocol := core_mesh.ParseProtocol(inbound.GetProtocol())
-		if _, exists := proxy.Policies.FaultInjections[iface]; exists {
```
slonka, line 74: why was this here in the first place @lukidzi ? why do we not need it now @lahabana ?

slonka (approved):
Just a couple of Qs. This is a nice addition but have we seen this being a perf bottleneck anywhere?

lahabana:
I'm going to close this. I think the right option is to potentially not install old policies in this case.

PR #9110, opened 2024-02-01
feat(MeshTrafficPermission): apply default deny

jakubdyszkiewicz:
### Checklist prior to review

This PR changes the default behavior on no TrafficPermission and no MeshTrafficPermission for HTTP apps. It fixes a flake that we have on CI.

mtls.go "should enforce traffic permissions" (the one that already has FlakeAttempts) test is flaking in a weird way. When we apply mTLS the traffic is blocked. When we add TrafficPermission/MeshTrafficPermission it starts working, then it fails for a couple of requests and then it's running fine after this. I'd rather debug this in separate PR.

### DIFF
--- a/pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/plugin.go
+++ b/pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/plugin.go
@@ -55,11 +55,7 @@ func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *
 		return nil
 	}
 
-	mtp, ok := proxy.Policies.Dynamic[api.MeshTrafficPermissionType]
-	if !ok {
-		return nil
-	}
-
+	mtp := proxy.Policies.Dynamic[api.MeshTrafficPermissionType]
 	for _, res := range rs.Resources(envoy_resource.ListenerType) {
 		if res.Origin != generator.OriginInbound {
 			continue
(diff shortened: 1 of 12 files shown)


pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/plugin.go
```diff
@@ -74,7 +70,11 @@ func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *
 		}
 		rules, ok := mtp.FromRules.Rules[key]
 		if !ok {
-			continue
+			if len(proxy.Policies.TrafficPermissions) == 0 {
```
lahabana, line 73: It is not supported to mix TP and MTP. Let's not add custom code trying to do their own thing in each policy

test/e2e_env/universal/trafficpermission/traffic_permission.go
```diff
@@ -204,7 +204,7 @@ destinations:
 		Expect(err).ToNot(HaveOccurred())
 
 		// then
-		trafficBlocked()
+		trafficBlocked(503)
```
lahabana, line 207: Would be useful to explain why this is a 503...

pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/plugin.go
jakubdyszkiewicz: ok, but how do you want to add a default behavior on the lack of MTP? If we miss this check, then we will override any trafficpermission

lahabana: Ha you are right indeed.

lahabana (approved):

PR #9118, opened 2024-02-01
test(e2e): print cp logs on failure

jakubdyszkiewicz:
### Checklist prior to review

We had a test flake failure. I see that now we print logs of a container that failed, so we printed that Envoy did not receive XDS config (initial fetch timeout). However, there is no way to inspect CP logs during this time (at least at kubernetes). This PR prints CP logs after test fails.

### DIFF
--- a/test/framework/envs/universal/env.go
+++ b/test/framework/envs/universal/env.go
@@ -3,6 +3,7 @@ package universal
 import (
 	"encoding/json"
 
+	"github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	"github.com/kumahq/kuma/pkg/config/core"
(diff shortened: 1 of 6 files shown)


lukidzi (approved):

test/framework/envs/universal/env.go
```diff
@@ -55,3 +56,14 @@ func RestoreState(bytes []byte) {
 	Expect(Cluster.AddNetworking(state.ZoneEgress, framework.Config.ZoneEgressApp)).To(Succeed())
 	Cluster.SetCp(cp)
 }
+
+func PrintCPLogsOnFailure(report ginkgo.Report) {
```
slonka, line 60: we probably need a separate function because the cluster implem is different but theoretically could be done with one fn.

slonka (approved):

Issue #8819, opened 2024-01-11
MeshReconciler and MeshDefaultsReconciler can be merged into a single controller


lobkovilya:
### Description

They both have similar watch configurations:
```go
return kube_ctrl.NewControllerManagedBy(mgr).
	For(&mesh_k8s.Mesh{}).
	Complete(r)
```

From the code point of view, there are still 2 different things so we don't have to merge code-wise.

PR #9124, opened 2024-02-02
fix(kubernetes): create builtin CA once

jakubdyszkiewicz:
### Checklist prior to review

We had two controllers that reconciled the Mesh object. One for default mesh resources and another to call EnsureCA.
If you apply such Mesh
```
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: nondefault
spec:
  mtls:
    enabledBackend: ca-1
    backends:
      - name: ca-1
        type: builtin
        dpCert:
          rotation:
            expiration: 1d
        conf:
          caCert:
            RSAbits: 2048
            expiration: 10y
```
Both are triggered at the same time. Default mesh resources one was calling `ResourceManager#Update` which would trigger `EnsureCA`. It means that we were calling `EnsureCA` twice in parallel. The best case scenario is that we were just overriding one CA with another (easy to reproduce)
```
2024-02-02T12:20:41.790Z	INFO	ca.builtin	CA created	{"mesh": "nondefault"}
2024-02-02T12:20:41.875Z	INFO	ca.builtin	CA created	{"mesh": "nondefault"}
```
We store builtin CA in two secrets (public and private key).
The worst case scenario is that one call writes public call, another writes private key and we then have a mismatch :) This is what I noticed in our E2E test run https://app.circleci.com/pipelines/gh/kumahq/kuma/30885/workflows/2a36132b-972d-49b6-b721-9bea4a57e155/jobs/578481

I squashed two controllers into one so there are no parallel EnsureCA calls.

Also: fix https://github.com/kumahq/kuma/issues/8819


slonka (approved):
LGTM

PR #9153, opened 2024-02-06
fix(kuma-cp): map instance not found to 404

lukidzi:
### Checklist prior to review

### DIFF
--- a/pkg/core/rest/errors/error_handler.go
+++ b/pkg/core/rest/errors/error_handler.go
@@ -19,6 +19,7 @@ import (
 	"github.com/kumahq/kuma/pkg/core/tokens"
 	"github.com/kumahq/kuma/pkg/core/validators"
 	"github.com/kumahq/kuma/pkg/envoy/admin"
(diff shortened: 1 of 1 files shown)


slonka (approved):

pkg/core/rest/errors/error_handler.go
```diff
@@ -88,6 +89,12 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 				},
 			},
 		}
+	case errors.Is(err, catalog.ErrInstanceNotFound):
+		kumaErr = &types.Error{
+			Status: 404,
+			Title:  "Dataplane not found",
```
jakubdyszkiewicz, line 95: why dataplane not found? 🤔

lukidzi: While testing, I observed that the globalCp instance contains information about the specific global CP instance connected to a zone, which, in turn, has a connection to a DP. I believe the message `instance not found` may be a bit misleading when making a request from the Global CP. In my case, zone was offline while DP was online 🤔

PR #9157, opened 2024-02-07
test(e2e): upgrade KIC

jakubdyszkiewicz:
### Checklist prior to review

We recently saw some flakiness for Kong tests, so I decided to upgrade Kong to see if this may help. I switched to Gateway API which is now a recommended way to handle kong.

I also noticed that we deploy two Kong in parallel with the same ingress (gateway) class, so I think this can potentially cause flakiness, one route would override another, so now each deployment gets its own class name.


lukidzi:
I think the newer version of KIC supports `arm64`, maybe we can remove the flag https://github.com/kumahq/kuma/blob/master/test/e2e_env/kubernetes/kubernetes_suite_test.go#L61

lukidzi (approved):

PR #9201, opened 2024-02-12
test(e2e): multizone upgrade

lobkovilya:
Add an e2e test to check multizone upgrades.

### Checklist prior to review


test/e2e/helm/kuma_helm_upgrade_multizone.go
```diff
@@ -0,0 +1,154 @@
...
+			n, err := numberOfPolicies(global)
+			g.Expect(err).ToNot(HaveOccurred())
+			return n
+		}, "10s", "100ms").Should(Equal(1))
+		// and
+		Consistently(func(g Gomega) int {
+			n, err := numberOfPolicies(zone)
+			g.Expect(err).ToNot(HaveOccurred())
+			return n
+		}, "10s", "100ms").Should(Equal(1))
```
jakubdyszkiewicz, line 152: nit: can we squeeze this to one consistently? 20s overhead is a lot :( Should this be 5s?

pkg/util/versions/versions.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"os"
+
+	"github.com/Masterminds/semver/v3"
+	"sigs.k8s.io/yaml"
+)
+
+const previewVersion = "preview"
+
+func Supported(path string) ([]*semver.Version, error) {
```
jakubdyszkiewicz, line 12: Naming is a bit weird here. It's hard to guess what this function does by the name.
It has nothing to do with supported versions, no?

It's more of `ParseFromFile`?

jakubdyszkiewicz (approved):

test/e2e/helm/kuma_helm_upgrade_multizone.go
lobkovilya: yeah, can be squeezed into 1 Consistently. Unfortunately, 5s is not enough to track the issue, but maybe I can wait for a new subscription on ZoneInsight with Eventually and then do a 5s Consistently that checks the number of policies

pkg/util/versions/versions.go
lobkovilya: ah right, some of them are out of support, I'll rename the func

pkg/util/versions/versions.go
```diff
@@ -0,0 +1,43 @@
...
+			return nil, err
+		}
+		versions = append(versions, ver)
+	}
+
+	return versions, nil
+}
+
+func OldestUpgradableToLatest(version []*semver.Version) string {
+	return version[len(version)-1-2].String()
```
michaelbeaumont, line 42: This isn't quite right I don't think. Shouldn't it be something like `{current.major, current.minor-2, 0}`? This just returns the 3rd to last version in the list which at `2.6.3` could be `2.6.1`

michaelbeaumont (changes requested):

pkg/util/versions/versions.go
michaelbeaumont: `versions.yaml` only has the latest for a minor version but https://kuma.io/docs/2.6.x/production/upgrades-tuning/upgrades/#upgrade-kuma explicitly says the patch version doesn't matter, so I suggest setting the patch of the returned version to `0`.

test/e2e/helm/kuma_helm_upgrade_multizone.go
lobkovilya: Added Eventually that waits for the new ZoneInsight subscription. So now I was able to create 1 Consistently with 5s.

pkg/util/versions/versions.go
michaelbeaumont: @lobkovilya let's leave it like you have it and open a website PR to make the support policy explicit.

michaelbeaumont (approved):

PR #9222, opened 2024-02-12
chore(deps): upgrade envoy to v1.29.1

lukidzi:
### Checklist prior to review

### DIFF
--- a/tools/releases/version.sh
+++ b/tools/releases/version.sh
@@ -8,7 +8,7 @@ function envoy_version() {
   # - if ENVOY_TAG is a real git tag like 'v1.20.0' then the version is equal to '1.20.0' (without the first letter 'v').
   # - if ENVOY_TAG is a commit hash then the version will look like '1.20.1-dev-b16d390f'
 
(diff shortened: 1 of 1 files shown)


slonka (approved):
Why do we need to bump? I guess CVEs right? https://github.com/envoyproxy/envoy/releases/tag/v1.29.1

tools/releases/version.sh
```diff
@@ -8,7 +8,7 @@ function envoy_version() {
   # - if ENVOY_TAG is a real git tag like 'v1.20.0' then the version is equal to '1.20.0' (without the first letter 'v').
   # - if ENVOY_TAG is a commit hash then the version will look like '1.20.1-dev-b16d390f'
 
-  ENVOY_TAG=${ENVOY_TAG:-"v1.29.0"}
+  ENVOY_TAG=${ENVOY_TAG:-"v1.29.1"}
```
slonka, line 11: Should this be backported? I think it should.

slonka: oh, it is - weird it's not through a label 🤔

PR #9257, opened 2024-02-14
fix(postgres): add ping when creating a pool

lahabana:
This may avoid cases where the pool is not properly warmed on startup

### Checklist prior to review

### DIFF
--- a/pkg/plugins/common/postgres/connection.go
+++ b/pkg/plugins/common/postgres/connection.go
@@ -73,8 +73,7 @@ func ConnectToDbPgx(ctx context.Context, postgresStoreConfig config.PostgresStor
 	if err != nil {
 		return nil, err
 	}
-	err = p.Ping(ctx)
-	if err != nil {
+	if err := p.Ping(ctx); err != nil {
 		return nil, err
 	}
 	return p, nil
--- a/pkg/plugins/common/postgres/connection.go
+++ b/pkg/plugins/common/postgres/connection.go
@@ -41,7 +41,7 @@ func ConnectToDb(cfg config.PostgresStoreConfig) (*sql.DB, error) {
 // ref. https://github.com/DataDog/dd-trace-go/blob/3d97fcec9f8b21fdd821af526d27d4335b26da66/contrib/database/sql/conn.go#L290
 var spanTypeSQLAttribute = attribute.String("span.type", "sql")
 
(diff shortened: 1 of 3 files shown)


lahabana:
@jakubdyszkiewicz curious if maybe this would fix: https://github.com/Kong/mink-charts/issues/957

pkg/plugins/common/postgres/connection.go
```diff
@@ -69,5 +69,13 @@ func ConnectToDbPgx(postgresStoreConfig config.PostgresStoreConfig, customizers
...
 
-	return pgxpool.NewWithConfig(context.Background(), pgxConfig)
+	p, err := pgxpool.NewWithConfig(ctx, pgxConfig)
+	if err != nil {
+		return nil, err
+	}
+	err = p.Ping(ctx)
+	if err != nil {
+		return nil, err
+	}
```
michaelbeaumont, line 79:
```suggestion
	if err := p.Ping(ctx); err != nil {
		return nil, err
	}
```

nit obviously

pkg/plugins/common/postgres/connection.go
```diff
@@ -69,5 +69,13 @@ func ConnectToDbPgx(postgresStoreConfig config.PostgresStoreConfig, customizers
 		return nil, err
 	}
 
-	return pgxpool.NewWithConfig(context.Background(), pgxConfig)
+	p, err := pgxpool.NewWithConfig(ctx, pgxConfig)
+	if err != nil {
+		return nil, err
+	}
+	err = p.Ping(ctx)
```
michaelbeaumont, line 76: Does this need a retry?

michaelbeaumont: Or does the component lifecycle handle restarting

michaelbeaumont (approved):

pkg/plugins/common/postgres/connection.go
jakubdyszkiewicz: It wouldn't. The CP would restart. Yeah, we might need a retry there

lahabana: So you think I should retry around this?

lahabana:
I'm not sure this helps. There is another theory around this closing this until we pursue the other theory

Issue #9184, opened 2024-02-09
Default policies are doubled after the upgrade


lobkovilya:
### What happened?

Kuma 2.4.3 zone with new default policies gets doubled default policies after the upgrade to 2.6.0:
```
MESH      NAME                                                 AGE
default   gw-timeouts                                          33m
default   gw-timeouts-xf29454cfxf4b984                         6m
default   mesh-gateways-timeout-all-default                    51m
default   mesh-gateways-timeout-all-default-dwb9v4724c244745   6m
default   mesh-timeout-all-default                             51m
default   mesh-timeout-all-default-2x9x48578wc4zcc7            6m
```

PR #9259, opened 2024-02-14
fix(kuma-cp): kds sync on upgrade doubles the number of policies

lobkovilya:
The PR makes KDS sync rely on `kuma.io/origin: zone` label. But we still want `KUMA_MULTIZONE_ZONE_DISABLE_ORIGIN_LABEL_VALIDATION: true` to disable origin label validation. That's why the PR introduces a webhook that automatically adds `kuma.io/origin: zone` after the validation.

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix https://github.com/kumahq/kuma/issues/9184

### DIFF
--- a/pkg/api-server/resource_endpoints.go
+++ b/pkg/api-server/resource_endpoints.go
@@ -360,6 +360,13 @@ func (r *resourceEndpoints) createResource(
 		return
 	}
 
+	if r.federatedZone && r.descriptor.IsPluginOriginated {
+		if labels == nil {
+			labels = map[string]string{}
+		}
+		labels[mesh_proto.ResourceOriginLabel] = string(mesh_proto.ZoneResourceOrigin)
+	}
+
 	res := r.descriptor.NewObject()
 	_ = res.SetSpec(spec)
 	if err := r.resManager.Create(ctx, res, store.CreateByKey(name, meshName), store.CreateWithLabels(labels)); err != nil {
--- a/pkg/api-server/resource_endpoints.go
+++ b/pkg/api-server/resource_endpoints.go
@@ -360,7 +360,7 @@ func (r *resourceEndpoints) createResource(
 		return
 	}
 
-	if r.federatedZone && r.descriptor.IsPluginOriginated {
+	if r.mode == config_core.Zone {
 		if labels == nil {
 			labels = map[string]string{}
 		}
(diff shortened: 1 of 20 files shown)


pkg/api-server/resource_endpoints.go
```diff
@@ -360,6 +360,13 @@ func (r *resourceEndpoints) createResource(
 		return
 	}
 
+	if r.federatedZone && r.descriptor.IsPluginOriginated {
```
jakubdyszkiewicz, line 363: should this be only in federated zone? Shouldn't we need this in non-federated zone and potentially global too?

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -0,0 +1,127 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
+	"github.com/kumahq/kuma/pkg/config/core"
+	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
+	"github.com/kumahq/kuma/pkg/version"
+)
+
+type ResourceAdmissionChecker struct {
```
jakubdyszkiewicz, line 19: nit: I'd strick to Validator suffix, but it's up to you

lobkovilya: That was my initial choice, but then I discovered there is already a thing called [AdmissionValidator](https://github.com/kumahq/kuma/blob/2962747ee1e4350a23e1ec0581d4f78132ac9dcb/pkg/plugins/common/k8s/composite_validator.go#L11) and since I'm not implementing the interface probably it's better not to use `AdmissionValidator` so it's less confusing

pkg/api-server/resource_endpoints.go
lobkovilya: Ah right, a non-federated zone should have this webhook for sure. Also, we should probably add `kuma.io/origin` to all resources, not just plugin-originated policies.

jakubdyszkiewicz (approved):

Issue #8845, opened 2024-01-15
Implement metrics profiles in MeshMetric policy 


Automaat:
### Description

Implement profiles in MeshMetric policy. Brief description in [MeshMetric MADR](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/035-mesh-metric-policy.md#metrics-profiles), but we need to write MADR with profiles itself

Automaat:
Look into: https://github.com/kumahq/kuma/issues/8108

Issue #8108, opened 2023-10-23
Metrics are likely incorrect when cross mesh clusters are present


michaelbeaumont:
### What happened?

The metrics hijacker doesn't understand the [`<cluster>_<mesh>` format](https://github.com/kumahq/kuma/blob/39f99ff5f2b0760ea08511831623e2ab5ad0e6bb/pkg/plugins/policies/core/xds/meshroute/listeners.go#L47)

lukidzi:
Is it still the case when using MeshMetrics? @slonka @Automaat

bartsmykla:
> Triage: Correct way to do cross mesh will be to use MeshIdentity/MeshTrust

PR #9269, opened 2024-02-15
docs(meshmetric): meshmetric profiles MADR

slonka:
Don't have any other ideas, maybe someone reads this before weekend and can chip in some ideas.

closes: https://github.com/kumahq/kuma/issues/8845

Also I can remove the files i added in `assets/038/` - I used them to get the data for `combined.txt` - these are metrics combined from all dashboards available - and compare that to what's default in Envoy and in DD metadata list.

### Checklist prior to review


docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,228 @@
...
+
+Just like we [mutate responses for metrics hijacker](https://github.com/kumahq/kuma/blob/d6c9ce64ac5e7ba1f5dbb9fb410e7d9410b67815/app/kuma-dp/pkg/dataplane/metrics/server.go#L348)
+we can add a filter mutator to reduce the number of metrics (same thing for [OTEL](https://github.com/kumahq/kuma/blob/d6c9ce64ac5e7ba1f5dbb9fb410e7d9410b67815/app/kuma-dp/pkg/dataplane/metrics/metrics_producer.go#L106)).
+
+#### Validation
+
+After all profiles are compiled from regexes make sure that they include the ones on the lower levels (all includes default, default includes minimal etc.)
+Make sure that with `default` profile (or the profile chosen for dashboards) all dashboards are populated.
+
+Can we somehow track if users are happy with the defined profiles?
```
Automaat, line 204: I think it would be hard to track, but I guess that if someone will have problem with profiles they will come

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,228 @@
...
+# 378
+```
+
+And datadog lists 990 metrics in total https://github.com/DataDog/integrations-core/blob/master/envoy/metadata.csv and 329 non-legacy ones
+
+```bash
+curl -s https://raw.githubusercontent.com/DataDog/integrations-core/master/envoy/metadata.csv | grep -v Legacy | wc -l                                                                                                                                         -- INSERT --
+```
+
+We can build automation on top of this, so when we update dashboards or Envoy changes metrics it emits we know about this and can adjust accordingly
```
Automaat, line 71: does the metrics change that often to build automation around this?

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,228 @@
...
+docker run --rm -it -p 9901:9901 -p 10000:10000 envoyproxy/envoy:v1.29.1
+curl -s localhost:10000 > /dev/null
+curl -s localhost:9901/stats/prometheus | grep -E -o '\benvoy_[a-zA-Z0-9_]+\b' | sort | uniq | wc -l
+# 378
+```
+
+And datadog lists 990 metrics in total https://github.com/DataDog/integrations-core/blob/master/envoy/metadata.csv and 329 non-legacy ones
+
+```bash
+curl -s https://raw.githubusercontent.com/DataDog/integrations-core/master/envoy/metadata.csv | grep -v Legacy | wc -l                                                                                                                                         -- INSERT --
```
slonka, line 68:
```suggestion
curl -s https://raw.githubusercontent.com/DataDog/integrations-core/master/envoy/metadata.csv | grep -v Legacy | wc -l
```

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+to
+
+```yaml
+sidecar:
+  usedOnly: true # true or false
+  profile:
+    name: default # one of `nothing`, `minimal`, `default`, `comprehensive`, `all`
+    exclude: regex2.* # first exclude
```
Automaat, line 190: could be a list so that it is easier to configure multiple filters

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,228 @@
+# MeshMetric profiles (predefined subsets of all metrics)
```
Automaat, line 1: we can approach profiles differently, one profile is a set of metrics, so we can combine profiles how we need

slonka: https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/017-policy-array-merging.md#pros-and-cons-of-the-options how to combine things ☝️

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,228 @@
...
+- Just an empty profile and people manually can add things to this.
+
+#### Schema
+
+I suggest changing the current schema of
+
+```yaml
+sidecar:
+  usedOnly: true # true or false
+  profile: minimal # one of minimal, default, full
```
johnharris85, line 179: What are these profiles? Where / how are they defined?

docs/madr/decisions/038-meshmetric-profiles.md
johnharris85: Not sure I'm a fan of the augmenting mechanic here (with excludes / includes). Why wouldn't we just allow people to define their own profiles and reference them, rather than bloating the MeshMetric schema?

docs/madr/decisions/038-meshmetric-profiles.md
slonka: I don't think it was implemented, this was just a suggestion in the previous MADR.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: What if someone wants to only have dashboard metrics (our "default" profile) + for some reason they care about 1 specific metric (e.g. `downstream_cx_drain_close`). So they build their own profile based on our profile (without augmenting mechanic, so just copy-paste) then we update our dashboards and add 1 new metric/graph. After kuma update they have one empty graph and have to manually update their profile. This augmenting behaviour allows us to keep clients up to date with what we think is needed + their special use cases.

slonka:
Moving this back to draft until I make the adjustments.

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,352 @@
...
+  name: my-profile
+  namespace: kuma-system
+spec:
+  default:
+    appendMetricExact:
+      - envoy_cluster_upstream_rq_xx
+      - envoy_http_downstream_rq_xx
+    appendMetricRegex:
+      - .*error.*
+```
```
lahabana, line 101: Why a different resource and not directly in MeshMetric?

slonka: This was a question from John Harris: https://github.com/kumahq/kuma/pull/9269#discussion_r1496529297

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,352 @@
...
+  profile:
+    name: default # one of `nothing`, `minimal`, `default`, `comprehensive`, `all`
+    exclude: regex2.* # first exclude
+    include: regex1.* # then include (include takes over)
+```
+
+#### Implementation
+
+Just like we [mutate responses for metrics hijacker](https://github.com/kumahq/kuma/blob/d6c9ce64ac5e7ba1f5dbb9fb410e7d9410b67815/app/kuma-dp/pkg/dataplane/metrics/server.go#L348)
+we can add a filter mutator to reduce the number of metrics (same thing for [OTEL](https://github.com/kumahq/kuma/blob/d6c9ce64ac5e7ba1f5dbb9fb410e7d9410b67815/app/kuma-dp/pkg/dataplane/metrics/metrics_producer.go#L106)).
```
jakubdyszkiewicz, line 321: Can we double check if we won't get a subsential memory usage reduction when we do this directly in Envoy?

I know that it's not dynamic then, because it's configured in bootstrap, but it might be worth checking if this is neglegable difference or not.

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,352 @@
...
+  - `.*reject.*`
+  - `envoy_listener_no_filter_chain_match`
+  - `.*denied.*` (envoy_rbac_denied, envoy_rbac_shadow_denied)
+  - `envoy_server_days_until_first_cert_expiring`
+- Saturation
+  - `.*memory.*` (allocated, heap, physical)
+
+##### Nothing
+
+- Just an empty profile and people manually can add things to this.
```
jakubdyszkiewicz, line 226: I wonder if the profiles below are really useful. Especially *App profiles.

I expect there are two cases:
* You go with full profile and you want to exclude noisy metrics
* You go with extensive and you add metrics you are missing

IMHO it's very unlikely that you start from 0 and then you are adding metrics on top of this. I'm sceptical whether people would actually use those small profiles below

docs/madr/decisions/assets/038/all-dd.txt
```diff
@@ -0,0 +1,328 @@
+envoy_access_logs_grpc_access_log_logs_dropped_count
```
lahabana, line 1: Can we put these in gist to avoid adding 100s of lines to this repo?

docs/madr/decisions/038-meshmetric-profiles.md
slonka: @johnharris85 - do you have anything else to add to https://github.com/kumahq/kuma/pull/9269#discussion_r1496529297 ?

docs/madr/decisions/038-meshmetric-profiles.md
slonka: > Can we double check if we won't get a subsential memory usage reduction when we do this directly in Envoy?

Yeah if we do this in Envoy then we'll reduce the memory:
- Envoy could be more memory efficient at filtering stuff
- 1 smaller response to our hijacker
- No need to allocate anything to process the response

I'll add the two approaches with pros and cons. I think we can start with doing this in Envoy (we pretty much get it for free with `filter`) and in the future if we need more processing we can rewrite that (it's not exposed to the user so it won't be a breaking change).

> I know that it's not dynamic then

It is dynamic, now we can reconfigure MeshMetric dynamically.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: We need to settle down on something, @lahabana - for the troubleshooting case would you be ok with having "full" profile for that period? I mean if you troubleshoot something you probably not 100% sure what's wrong so turning on full profile just for a couple of hours / day shouldn't hurt that much, right? Then if people are not happy with this approach we can add more profiles in next iterations. WDYT?

lahabana: Sounds good. So we'd go with 2 profiles:

- golden (more or less what's required for dashboards and what looks useful)
- all (no filter whatsoever)

These are still additive even though it's not that useful ftm.

Right?

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,352 @@
...
+or should they be hidden from the user (and only referencable)?
+
+#### Profiles suggested:
+
+Suggestions to the merge / split and naming are welcomed.
+The first ones are granularity based the last ones are feature focused (see [additive profiles](#behaviour)).
+
+##### All
+
+Nothing is removed, everything included in Envoy, people can manually remove stuff.
```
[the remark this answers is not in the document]
slonka: Yeah, I just wanted to have that so it explicit and consistent (we have "none" profile). User does not have to figure out what missing profile name means.

Its:

```
sidecar:
  profile:
    exclude: regex2.*
```

vs

```
sidecar:
  profile:
    name: all
    exclude: regex2.*
```

I'm not very firm on that so up to you.

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,352 @@
...
+- Just an empty profile and people manually can add things to this.
+
+##### ADS / XDS / Management server
+
+All stats from: https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/mgmt_server#subscription-statistics
+divided into CDS/LDS/RDS/SDS/VHDS by `listener_manager.[cds, lds, rds, sds, vhds]`.
+
+This would enable ADS (aggregated xDS) stats, might be helpful when debugging DP-CP communication on various resources (HDS, LDS, SDS).
+On the other hand we would have to parse labels as well because these metrics are under normal metrics like `envoy.cluster.upstream_rq_timeout.count`,
+but are published with `envoy_cluster:ads_cluster` label.
```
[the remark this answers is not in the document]
slonka: I like the `internal_cluster_stats` idea.

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,352 @@
...
+
+##### Comprehensive
+
+- All available dashboards
+
+##### Default
+
+- Our dashboards
+
+##### Minimal
```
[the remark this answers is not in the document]
slonka:
```suggestion
##### Golden
```

docs/madr/decisions/038-meshmetric-profiles.md
lahabana: I think @jakubdyszkiewicz was mentioning that: https://envoyproxy.io/docs/envoy/latest/api-v3/config/metrics/v3/stats.proto#envoy-v3-api-msg-config-metrics-v3-statssink  is dynamically configurable in envoy. 

Currently we have:

```json
        "statsConfig": {
          "statsTags": [
            {
              "tagName": "name",
              "regex": "^grpc\\.((.+)\\.)"
            },
            {
              "tagName": "status",
              "regex": "^grpc.*streams_closed(_([0-9]+))"
            },
            {
              "tagName": "kafka_name",
              "regex": "^kafka(\\.(\\S*[0-9]))\\."
            },
            {
              "tagName": "kafka_type",
              "regex": "^kafka\\..*\\.(.*?(?=_duration|$))"
            },
            {
              "tagName": "worker",
              "regex": "(worker_([0-9]+)\\.)"
            },
            {
              "tagName": "listener",
              "regex": "((.+?)\\.)rbac\\."
            }
          ]
        },
```

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,352 @@
...
+
+##### CircuitBreaker
+
+For circuit breaker we use both circuit breaker and outlier detection so we could combine these:
+- https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats.html#circuit-breakers-statistics
+- https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats.html#outlier-detection-statistics
+
+##### Tracing
+
+- https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/stats.html#tracing-statistics
```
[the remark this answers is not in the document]
slonka: These are in EnvoyServer

docs/madr/decisions/assets/038/all-dd.txt
slonka: Yes, I'll add all of them to a gist

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,228 @@
...
+
+#### Profiles suggested:
+
+Suggestions to the merge / split and naming are welcomed.
+
+##### All
+
+Nothing is removed, everything included in Envoy, people can manually remove stuff.
+
+##### Extensive
```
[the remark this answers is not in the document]
slonka: Will squeeze them together.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: Will add a comment explaining the situation.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: Yeah so we could use `stats_matcher` but we need to be careful about https://github.com/envoyproxy/envoy/issues/8771

docs/madr/decisions/038-meshmetric-profiles.md
slonka: > These are still additive even though it's not that useful ftm.

Yes

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,352 @@
...
+    - ref: default # our predefined profile
+```
+
+The question here is: do we want our builtin profiles to be defined like that (this possibly means a user can edit them a bit like default policies)
+or should they be hidden from the user (and only referencable)?
+
+#### Profiles suggested:
+
+Suggestions to the merge / split and naming are welcomed.
+The first ones are granularity based the last ones are feature focused (see [additive profiles](#behaviour)).
```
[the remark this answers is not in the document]
slonka: We're starting with a small set of profiles for now. Could be discussed in the future.

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,367 @@
...
+### Negative Consequences
+
+* more processing power needed to filter metrics in kuma-dp
+
+## Links
+
+* https://github.com/lahabana/demo-scene/blob/a48ec6e0079601d340f79613549e1b2a4ea715a1/mesh-localityaware/k8s/otel-collectors.yaml#L174
+* https://docs.datadoghq.com/integrations/envoy
+* https://docs.datadoghq.com/integrations/istio
+* https://gist.github.com/slonka/f8fd490b8918cf027bbeba1df8ae2419
```
slonka, line 367:
```suggestion
* https://gist.github.com/slonka/f8fd490b8918cf027bbeba1df8ae2419
* https://docs.google.com/document/d/1CnyeIi5hMBqXZRKm-0_yW3yy69HM80BHg-EGwkJb-qw/edit
```

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,368 @@
...
+
+We could adjust our dashboards to reflect which graphs are going to be populated in which profile.
+
+## Decision Outcome
+
+Chosen option: "Base profiles on expert knowledge, external dashboards and our grafana dashboards" with:
+- implementation in kuma-dp because it's dynamically configurable
+- profiles being additive
+- manual escape hatches using include/exclude
+- start with 2 profiles all, golden (+dashboards) and extend if needed
```
Automaat, line 346: so it will be three profiles? `all`, `default` and `golden`?

slonka: Charly suggested to merge that, but I did not push that change, will change that.

Automaat (approved):

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+to
+
+```yaml
+sidecar:
+  usedOnly: true # true or false
+  profiles:
+    appendProfiles:
+      - name: default
+      - name: internal_cluster_stats # not available now, in the future
+    exclude:
```
jakubdyszkiewicz, line 312: I'd consider naming this `excludeRegex`. `sidecar.profiles.exclude` with a list of strings would kind of suggest that I can exclude profiles. Same with include.

slonka:
```suggestion
    excludeRegex:
```

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+      - regex2.* # first exclude
+    include:
```
slonka, line 314:
```suggestion
    includeRegex:
```

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+sidecar:
+  usedOnly: true # true or false
+  appendProfiles:
+    - ref: my-profile
+    - ref: default # our predefined profile
+```
+
+The question here is: do we want our builtin profiles to be defined like that (this possibly means a user can edit them a bit like default policies)
+or should they be hidden from the user (and only referencable)?
+Since we're starting without this it can be skipped for now but in the future this needs reconsideration.
```
lahabana, line 132: I think it is not actually written out in the MADR that we push this feature to the future

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+We probably don't want to have a another "default policies" situation because that causes problems.
```
lahabana, line 133: What are the problems?

docs/madr/decisions/038-meshmetric-profiles.md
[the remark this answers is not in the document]
lahabana: Yeah we can it doesn't cost us anything

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+  - `envoy_cluster_upstream_rq_pending_failure_eject`
+  - `.*overflow.*`
+  - `.*no_cluster.*`
+  - `.*no_route.*`
+  - `.*reject.*`
+  - `envoy_listener_no_filter_chain_match`
+  - `.*denied.*` (envoy_rbac_denied, envoy_rbac_shadow_denied)
+  - `envoy_server_days_until_first_cert_expiring`
+- Saturation
+  - `.*memory.*` (allocated, heap, physical)
```
lahabana, line 218: Do we use cpu?

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+sidecar:
+  usedOnly: true # true or false
+  profiles:
+    appendProfiles:
+      - name: default
+      - name: internal_cluster_stats # not available now, in the future
+    excludeRegex:
+      - regex2.* # first exclude
+    includeRegex:
+      - regex1.* # then include (include takes over)
```
lahabana, line 315: I'd go with:

```yaml
include:
  - type: prefix|regex|exact # default is prefix (I guess)
     match:  "my_match"
```

Or at least:

```yaml
include:
  - regex: foo
```

This would make things flexible for future changes.

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+  - `envoy_cluster_upstream_cx_.*`
+  - `envoy_cluster_upstream_rq_.*`
+  - `envoy_cluster_circuit_breakers_.*`
+  - `envoy_http_downstream_.*`
+  - `envoy_listener_downstream_.*`
+  - `envoy_listener_http_.*`
+
+##### Default
+
+- Our dashboards stats plus [golden 4 signals](https://sre.google/sre-book/monitoring-distributed-systems/) (by regex / or exact):
```
lahabana, line 156: Do we filter out internal clusters? ( I think we should)

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+we can add a filter mutator to reduce the number of metrics (same thing for [OTEL](https://github.com/kumahq/kuma/blob/d6c9ce64ac5e7ba1f5dbb9fb410e7d9410b67815/app/kuma-dp/pkg/dataplane/metrics/metrics_producer.go#L106)).
+
+##### In Envoy
+
+We could implement this directly in Envoy using [stats_matcher](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/metrics/v3/stats.proto#envoy-v3-api-msg-config-metrics-v3-statssink).
+This would probably be faster, more memory efficient, but it's in boostrap config, so we couldn't change it without restarting Envoy.
+
+#### Validation
+
+After all profiles are compiled from regexes make sure that they include the ones on the lower levels (all includes default, default includes golden etc.)
```
lahabana, line 332: As these profiles are not provided by users it doesn't feel like they need to be regexs (which are known to be slow).

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+##### CircuitBreaker
+
+For circuit breaker we use both circuit breaker and outlier detection so we could combine these:
+- https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats.html#circuit-breakers-statistics
+- https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats.html#outlier-detection-statistics
+
+##### Tracing
+
+- https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/stats.html#tracing-statistics
+
```
lahabana, line 291: It's not clear which profiles we implement in this release.
My guess is none, all, default? Also I guess the default is all?  So is the name `default` no right. Should it be `basic` ? 

Also is this our opportunity to fix metrics hijacker and use otel libs instead of our manual and problematic merge fuction: https://github.com/kumahq/kuma/blob/master/app/kuma-dp/pkg/dataplane/metrics/merge.go ? 

If there are no profiles we use the old way but as soon as a profile is selected we use the new way and fix problems like: https://github.com/kumahq/kuma/issues/8108 .

lahabana: It will also be a lot easier to drop internal clusters in metrics.

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+    includeRegex:
+      - regex1.* # then include (include takes over)
+```
+
+#### Implementation
+
+##### In kuma-dp
+
+Just like we [mutate responses for metrics hijacker](https://github.com/kumahq/kuma/blob/d6c9ce64ac5e7ba1f5dbb9fb410e7d9410b67815/app/kuma-dp/pkg/dataplane/metrics/server.go#L348)
+we can add a filter mutator to reduce the number of metrics (same thing for [OTEL](https://github.com/kumahq/kuma/blob/d6c9ce64ac5e7ba1f5dbb9fb410e7d9410b67815/app/kuma-dp/pkg/dataplane/metrics/metrics_producer.go#L106)).
```
lahabana, line 323: IIUC now we use otel everywhere so I think filtering should really be done using otel libs so it's both solid and efficient (and less to maintain on our side ;) ).

docs/madr/decisions/038-meshmetric-profiles.md
```diff
@@ -0,0 +1,371 @@
...
+  usedOnly: true # true or false
+  profiles:
+    appendProfiles:
+      - name: default
+      - name: internal_cluster_stats # not available now, in the future
+    excludeRegex:
+      - regex2.* # first exclude
+    includeRegex:
+      - regex1.* # then include (include takes over)
+```
```
lahabana, line 316: Please explicitly state that include takes precedence over exclude

docs/madr/decisions/038-meshmetric-profiles.md
lahabana: e.g: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/filterprocessor

lahabana (approved):
Some minor comments around syntax and implem but looking good!

docs/madr/decisions/038-meshmetric-profiles.md
slonka: e.g. someone manually editing a default profile, and then we need to modify it / synchronize somewhere etc.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: I don't think Envoy can measure CPU because it's relative to the infra, right? So it needs to come from k8s / OS etc.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: Added note

docs/madr/decisions/038-meshmetric-profiles.md
slonka: This is what I meant. When we compile the static list of all metrics needed make sure we keep the hierarchy. I'll update this sentence.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: > It's not clear which profiles we implement in this release. My guess is none, all, default?

Yes, and it's written down in decision outcome section.

> Also I guess the default is all? 

I thought that we wanted `basic` to be the default profile. I've also renamed `default` to `basic`.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: agreed 👍

docs/madr/decisions/038-meshmetric-profiles.md
slonka: Closing this since there is no movement on this comment. If anyone feels strongly about this, please open up a thread on slack or chime in during the implem PR.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: Closing this since there is no movement on this comment. If anyone feels strongly about this, please open up a thread on slack or chime in during the implem PR.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: I know it's SO but I think I'm right here https://stackoverflow.com/a/21535618 - please ping me if you think otherwise.

docs/madr/decisions/038-meshmetric-profiles.md
slonka: If we want to switch the name or which one is default it's not going to be big change. Closing this for now, will resolve this in implem PR if needed.

Issue #9185, opened 2024-02-09
Support `from[].targetRef.kind: MeshGateway` in `MeshTrafficPermission`


lahabana:
### Description

Tried:

```shell
echo "
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  namespace: kuma-system
  name: demo-app
spec:
  targetRef:
    kind: MeshService
    name: demo-app_kuma-demo_svc_5000
  from:
    - targetRef:
        kind: MeshGateway
        name: my-gateway
      default:
        action: Allow
" | kubectl apply -f -
The MeshTrafficPermission "demo-app" is invalid: spec.from[0].targetRef.kind: value is not supported
```

So I had to use:

```shell
echo "
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  namespace: kuma-system
  name: demo-app
spec:
  targetRef:
    kind: MeshService
    name: demo-app_kuma-demo_svc_5000
  from:
    - targetRef:
        kind: MeshService
        name: edge-gateway
      default:
        action: Allow
" | kubectl apply -f -
meshtrafficpermission.kuma.io/demo-app created
```

MeshGateway is quite useful in this case and should likely be preferred.

michaelbeaumont:
Wasn't this reverted? https://github.com/kumahq/kuma/pull/9359

lahabana:
Why did we decide to not go with MeshGateway? Was it simply because we can't really tell which listener traffic is coming from?

lukidzi:
`MeshGateway` allows you to select many different dataplanes. It's a bit strange that we can have different sets of dataplanes with different names matching the same listener.

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshGateway
mesh: default
metadata:
  name: edge-gateway
spec:
  selectors:
    - match:
        kuma.io/service: edge-gateway
    - match:
        kuma.io/service: other-gateway
        kuma.io/version: v2
  conf:
    listeners:
      - port: 8080
        protocol: HTTP
        hostname: foo.example.com
        tags:
          vhost: foo.example.com
      - port: 8080
        protocol: HTTP
        hostname: bar.example.com
        tags:
          vhost: bar.example.com
```
When we have a policy:

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
...
  from:
    - targetRef:
        kind: MeshGateway
        name: edge-gateway
```

[...]

PR #9298, opened 2024-02-19
feat(meshtrafficpermission): support from MeshGateway

lukidzi:
### Checklist prior to review

When using `MeshGateway` as the `from.kind`, we didn't map it for subsets. Additionally, `MeshGateway` kind uses the name of a gateway, so to make it work with `MeshTrafficPermission` we need to map matched dataplanes to services. I didn't add support for `MeshGateway` for `to` because I think there is no reason to target `MeshGateway`s from the mesh. Also, the single-match policy (`MeshTrace`) has no support for it either.

The matching of the `MeshGateway` to services:
1. Find `MeshGateway` resource with the name provided in `from.name`
2. Get all selectors and for each create a new `targetRef` with tags from selector
3. Each `targetRef` map to subset

1st commit shows the failing unit test with `from.kind: MeshGateway`
2nd commit shows the change required to make it work and the changed test file
3rd commit adds E2E test 

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/9185

### DIFF
--- a/pkg/plugins/policies/core/rules/rules.go
+++ b/pkg/plugins/policies/core/rules/rules.go
@@ -14,6 +14,7 @@ import (
 
 	common_api "github.com/kumahq/kuma/api/common/v1alpha1"
 	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
+	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
 	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
 	"github.com/kumahq/kuma/pkg/plugins/policies/meshhttproute/api/v1alpha1"
 	"github.com/kumahq/kuma/pkg/util/pointer"
@@ -262,6 +263,7 @@ func ComputeConf[T any](rs Rules, sub Subset) *T {
 
 func BuildFromRules(
 	matchedPoliciesByInbound map[InboundListener][]core_model.Resource,
+	gateways []*core_mesh.MeshGatewayResource,
 ) (FromRules, error) {
 	rulesByInbound := map[InboundListener]Rules{}
 	for inbound, policies := range matchedPoliciesByInbound {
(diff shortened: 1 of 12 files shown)


michaelbeaumont (commented):
Can we add a unit test with Envoy config?

lukidzi:
> Can we add a unit test with Envoy config?

I think it might not represent the actual change. We are providing an already processed policy https://github.com/kumahq/kuma/blob/master/pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/plugin_test.go#L102 instead of providing a policy and later building a subset based on it.

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -558,6 +569,27 @@ func toStringList(nodes []graph.Node) []string {
 	return rv
 }
 
+func mapToTargetRefs(tr common_api.TargetRef, gateways []*core_mesh.MeshGatewayResource) []common_api.TargetRef {
+	targetRefs := []common_api.TargetRef{}
+	switch tr.Kind {
+	case common_api.MeshGateway:
+		for _, gw := range gateways {
+			if gw.GetMeta().GetName() == tr.Name {
+				selectors := gw.Spec.GetSelectors()
```
michaelbeaumont, line 578: Should this include `gw.GetTags()`?

lukidzi: I don't think we need them https://github.com/kumahq/kuma/blob/master/api/mesh/v1alpha1/gateway.pb.go#L156 based on the comment it means tags on the listener, so I think they are not used to build SNI.

michaelbeaumont: I think the `spec.tags` don't matter because if the selector matches, it can only be a `Dataplane` configured by the named MeshGateway, we don't need to check the MeshGateway tags.

michaelbeaumont (approved):

PR #9321, opened 2024-02-20
feat(k8s): sidecar containers

### LINKED ISSUE #7541: add support for k8s sidecar-containers
### Description

It made it to alpha in 1.28 https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#api-for-sidecar-containers

This solves some of our problems so it'd be good to implement experimental support for this.

michaelbeaumont:
This PR adds a `.experimental` config flag. The feature isn't really experimental but it shouldn't be the default yet since [the feaure is still beta](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/).
It adds a new job to the matrix that runs the Kubernetes tests with the features enabled.
**Question**: Do we think it's ok that it runs only on master or should it be run on PRs?

Note: all the added lines are the golden files for injector tests with sidecars enabled.

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- Closes #7541 


.circleci/config.yml
```diff
@@ -285,7 +285,7 @@ jobs:
               - {equal: [calico, << parameters.cniNetworkPlugin >>]}
               - {equal: [kindIpv6, << parameters.k8sVersion >>]}
               - {equal: [arm64, << parameters.arch >>]}
-              - {equal: [true, << parameters.legacyKDS >>]}
+              # - {equal: [true, << parameters.sidecarContainers >>]}
```
michaelbeaumont, line 288: Should this run on PRs or not?

.github/workflows/build-test-distribute.yaml
```diff
@@ -221,20 +221,23 @@ jobs:
...
                 "cniNetworkPlugin": ["flannel"],
+                "sidecarContainers": [""],
                 "exclude":[
                   {"target": "kubernetes", "k8sVersion":"kind"},
                   {"target": "multizone", "k8sVersion":"kind"},
                   {"target":"universal", "k8sVersion":"${{ env.K8S_MIN_VERSION }}"},
                   {"target":"universal", "k8sVersion":"${{ env.K8S_MAX_VERSION }}"}
                 ],
                 "include":[
+                  {"sidecarContainers": "sidecarContainers", "k8sVersion": "${{ env.K8S_MAX_VERSION }}", "target": "kubernetes", "arch": "amd64"},
```
lahabana, line 240: If you only run this with amd64 I think you can skip the change on circleci

michaelbeaumont: removed!

slonka:
CI seems to not be running. I'll do "update branch" to see if that triggers a that.

pkg/plugins/runtime/k8s/plugin.go
```diff
@@ -302,10 +306,26 @@ func addValidators(mgr kube_ctrl.Manager, rt core_runtime.Runtime, converter k8s
...
+		k8sVersion, err := discClient.ServerVersion()
+		if err != nil {
+			return err
+		}
+		var sidecarContainersEnabled bool
+		if v, err := semver.NewVersion(
+			fmt.Sprintf("%s.%s.0", k8sVersion.Major, k8sVersion.Minor),
+		); err == nil && !v.LessThan(sidecarContainerVersion) {
+			sidecarContainersEnabled = rt.Config().Experimental.SidecarContainers
+		}
```
jakubdyszkiewicz, line 323: if someone enables this in experimental but kube cluster is too old can we log a message that we won't enable it?

pkg/plugins/runtime/k8s/util/util.go
```diff
@@ -122,8 +123,8 @@ func FindPort(pod *kube_core.Pod, svcPort *kube_core.ServicePort) (int, *kube_co
 	return 0, nil, fmt.Errorf("no suitable port for manifest: %s", pod.UID)
 }
 
-func FindContainerStatus(pod *kube_core.Pod, containerName string) *kube_core.ContainerStatus {
-	for _, cs := range pod.Status.ContainerStatuses {
+func FindContainerStatus(containerName string, status []kube_core.ContainerStatus, otherStatuses ...[]kube_core.ContainerStatus) *kube_core.ContainerStatus {
```
jakubdyszkiewicz, line 126: nit: this func definition is kind of weird. Can we just pass one slice here and do append if needed when we call it?

jakubdyszkiewicz (approved):

pkg/plugins/runtime/k8s/util/util.go
michaelbeaumont: Suspected this might be controversial...
I agree the definition is odd at first glance but I kind of like how it looks when calling it
```
util_k8s.FindContainerStatus(
  util_k8s.KumaSidecarContainerName,
  pod.Status.ContainerStatuses, 
  pod.Status.InitContainerStatuses,
)

util_k8s.FindContainerStatus(
  util_k8s.KumaSidecarContainerName,
  pod.Status.ContainerStatuses, 
)
```

I think I'll just expose two functions `FindContainerStatus` and `FindContainerOrInitContainerStatus` instead

pkg/plugins/runtime/k8s/plugin.go
michaelbeaumont: done

pkg/plugins/runtime/k8s/util/util.go
jakubdyszkiewicz: that works as well. ok.

Issue #9356, opened 2024-02-22
Implement targeting `MeshHTTPRoute` for `MeshRetry`


lukidzi:
### Description

Implement targeting `MeshHTTPRoute` for `MeshRetry`

PR #9365, opened 2024-02-22
feat(meshretry): allow configuration for MeshHTTPRoute

lukidzi:
### Checklist prior to review

The best to review by commit
- 8cfcfc25313dbae678626bfbda984b49f231335d adds validator
- 32ded9067eefb8a8ed029ea9fd06231397f175ad extract MeshHTTPRoute to sample 
- fb03e3b2e375207be13d266b1fdcb14caa0b6e4d add tests which have an incorrect config
- 72448e42b252d5cf0d269285d52dadb74cec6a6e implementation 
- 01d2656920976b1050dd19d3a976ea32a114e9e4 generate test file with correct configuration per route
- 38db787f10467355f16c23a221251bdf2b6291ae adds E2E test with small refactor to make it a bit faster

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/9356


pkg/plugins/policies/meshretry/api/v1alpha1/validator_test.go
```diff
@@ -169,6 +169,19 @@ to:
         retryOn: 
           - 500
           - 409
+`),
+			Entry("target MeshHTTPRoute", `
```
lobkovilya, line 173: What about
```yaml
targetRef:
  kind: MeshHTTPRoute
  name: route-1
to:
  - targetRef:
      kind: MeshService
      name: backend
    default:
      http:
        retryOn: 
          - 500
          - 409
```
I don't think we should allow anything other than `Mesh` in `to` when top level kind is `MeshHTTPRoute`

pkg/test/resources/samples/meshhttproute_samples.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	plugins_xds "github.com/kumahq/kuma/pkg/plugins/policies/core/xds"
+	meshhttproute_api "github.com/kumahq/kuma/pkg/plugins/policies/meshhttproute/api/v1alpha1"
+	meshhttproute_xds "github.com/kumahq/kuma/pkg/plugins/policies/meshhttproute/xds"
+	"github.com/kumahq/kuma/pkg/util/pointer"
+	envoy_common "github.com/kumahq/kuma/pkg/xds/envoy"
+)
+
+func MeshHttpOutboundRouteWithSeveralRoutes(serviceName string) *meshhttproute_xds.HttpOutboundRouteConfigurer {
```
lobkovilya, line 11: nit: do we need the word `Route` twice in this name? Should it just be `MeshHttpOutboundWithSeveralRoutes`?

test/e2e_env/universal/meshretry/http.go
```diff
@@ -71,26 +87,114 @@ spec:
 				universal.Cluster, "demo-client", "test-server.mesh",
 			)
 			g.Expect(err).ToNot(HaveOccurred())
-		}).Should(Succeed())
-		Consistently(func(g Gomega) {
-			// --max-time 8 to wait for 8 seconds to beat the default 5s connect timeout
+		}, "30s", "1s", MustPassRepeatedly(5)).Should(Succeed())
```
lobkovilya, line 90: Is it possible to set the interval to 100ms instead of 1s?

test/e2e_env/universal/meshretry/http.go
```diff
@@ -71,26 +87,114 @@ spec:
...
+		for i := 0; i < 50; i++ {
+			time.Sleep(time.Millisecond * 100)
+			_, err := client.CollectEchoResponse(
+				universal.Cluster, "demo-client", "test-server.mesh",
+			)
+			if err != nil {
+				errs = append(errs, err)
+			}
+		}
+		Expect(errs).ToNot(BeEmpty())
```
lobkovilya, line 106: can't we express the same thing but using `Eventually`?

lobkovilya: also, you should probably `break` the loop when there is at least 1 error in `errs`, but I'd go with `Eventually`

pkg/test/resources/samples/meshhttproute_samples.go
```diff
@@ -0,0 +1,75 @@
...
+		},
+		DpTags: map[string]map[string]bool{
+			"kuma.io/service": {
+				"web": true,
+			},
+		},
+	}
+}
+
+func MeshHttpOutboundRouteWithSingleRoute(serviceName string) *meshhttproute_xds.HttpOutboundRouteConfigurer {
```
lobkovilya, line 51: nit: same here, should be `MeshHttpOutboundWithSingleRoute`

test/e2e_env/universal/meshretry/http.go
```diff
@@ -71,32 +86,109 @@ spec:
...
+
+		By("Adding a MeshFaultInjection for test-server")
+		Expect(universal.Cluster.Install(YamlUniversal(meshFaultInjection))).To(Succeed())
+
+		By("Check some errors happen")
+		Eventually(func(g Gomega) {
+			_, err := client.CollectEchoResponse(
+				universal.Cluster, "demo-client", "test-server.mesh",
+			)
+			g.Expect(err).To(HaveOccurred())
```
lobkovilya, line 99: What error are we getting here? Can we check more, maybe we can see the status code is 500. It would be nice to rule out non-relevant errors here (like connectivity or DNS)

pkg/plugins/policies/meshretry/plugin/xds/configurer.go
```diff
@@ -59,27 +59,28 @@ func (c *Configurer) ConfigureListener(ln *envoy_listener.Listener) error {
 }
 
 func (c *Configurer) ConfigureRoute(route *envoy_route.RouteConfiguration) error {
-	if c.Retry == nil || route == nil {
+	conf := c.getConf(c.Subset)
+	if route == nil {
```
bartsmykla, line 63: nit part 1
```suggestion
	if route == nil {
```

pkg/plugins/policies/meshretry/plugin/xds/configurer.go
```diff
@@ -59,27 +59,28 @@ func (c *Configurer) ConfigureListener(ln *envoy_listener.Listener) error {
...
-	case "http":
-		policy, err = genHttpRetryPolicy(c.Retry.HTTP)
-		if err != nil {
-			return err
-		}
-	case "grpc":
-		policy = genGrpcRetryPolicy(c.Retry.GRPC)
-	default:
-		return nil
+	defaultPolicy, err := c.getRouteRetryConfig(conf)
```
bartsmykla, line 67: nit part 2
```suggestion
	defaultPolicy, err := c.getRouteRetryConfig(c.getConf(c.Subset))
```

pkg/plugins/policies/meshretry/plugin/xds/configurer.go
```diff
@@ -59,27 +59,28 @@ func (c *Configurer) ConfigureListener(ln *envoy_listener.Listener) error {
...
+			policy, err := c.getRouteRetryConfig(routeConfig)
+			if err != nil {
+				return err
+			}
+			if policy == nil && defaultPolicy == nil {
+				continue
+			}
+			if policy == nil {
+				policy = defaultPolicy
+			}
```
bartsmykla, line 83: nit
```suggestion
			if policy == nil {
				if defaultPolicy == nil {
					continue
				}
				policy = defaultPolicy
			}
```

test/e2e_env/universal/meshretry/http.go
```diff
@@ -71,32 +86,117 @@ spec:
...
+
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(response.ResponseCode).To(Equal(500))
+		}, "10s", "100ms").Should(Succeed())
+
+		By("Apply a MeshRetry policy")
+		Expect(universal.Cluster.Install(YamlUniversal(meshRetryPolicy))).To(Succeed())
+
+		By("Eventually all requests succeed consistently")
+		Eventually(func(g Gomega) {
```
bartsmykla, line 110: Maybe instead of rely on other policy (`MeshFaultInjection`) you could use new functionality of our test-server (https://github.com/kumahq/kuma/pull/9423, example usage: https://github.com/kumahq/kuma/pull/9437)

lukidzi: I don't think it will work better in this case. MeshFaultInjection creates a better setup because, in my experience, failing the 1st request and passing the 2nd one while the header is causing only 'x' consecutive requests to fail. When testing, if we retry and all requests pass, in my opinion, it creates a better setup

bartsmykla (approved):

PR #9391, opened 2024-02-23
feat(meshhttproute): set name of route action equal to hash of matches

lukidzi:
### Checklist prior to review

We are setting the name of a route equal to the hash of `match` for Dataplane routes but not for the gateway. This is useful when the `targetRef` is `kind: MeshHTTPRoute`. The change introduces setting RDS route names equal to the hash of the `match`.


slonka (approved):
In general ok, but are we sure we're not going to need to manually match using the name from somewhere outside? In other words, is it ok to make it a hash and not a more easily parseable name?

lukidzi:
> In general ok, but are we sure we're not going to need to manually match using the name from somewhere outside? In other words, is it ok to make it a hash and not a more easily parseable name?

We don't have a manual match of the route, but this is a standard way we are doing this for static routes. I wanted to make it the same for dynamic routes by setting the same hash. Hash is equal to the hash of a match section in the policy

PR #9394, opened 2024-02-23
fix(policy): don't set empty kuma.io service when using MeshHTTPRoute

lukidzi:
#### Problem
When policy targets `MeshHTTPRoute` with `to.kind: Mesh` we are creating an artificial object that has a tag with a hashed value of a matcher from `MeshHTTPRoute`. By default, we are creating it as a `MeshServiceSubset` which by default has `kuma.io/service` with an empty value. That works well for policies that have `to.kind: MeshService/MeshServiceSubset` but not for `Mesh/MeshSubset`. Later we cannot match the subset because `kuma.io/service: ""` is set. This change checks the kind and sets the correct value.

The best to review by commits 
1st shows the issue
2nd implementation
3rd shows the file

### DIFF
--- a/pkg/api-server/testdata/resources/inspect/meshgateways/_rules/meshhttproute.golden.json
+++ b/pkg/api-server/testdata/resources/inspect/meshgateways/_rules/meshhttproute.golden.json
@@ -122,13 +122,18 @@
       {
        "key": "kuma.io/service",
        "not": false,
-       "value": ""
+       "value": "backend_kuma-demo_svc_3001"
       }
      ],
      "origin": [
       {
        "mesh": "default",
        "name": "on-route",
+       "type": "MeshTimeout"
+      },
+      {
+       "mesh": "default",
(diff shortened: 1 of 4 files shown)


lukidzi:
When `kuma.io/service` is empty string we cannot match policies. When doing `isSubset` we are trying to find `kuma.io/service: ""` but in case if targetRef `Mesh` we don't have `kuma.io/service` label

michaelbeaumont (approved):
OK I guess it isn't clear to me whether this is a user-visible problem?

lukidzi:
I think it could be visible when targeting `MeshHTTPRoute` with `to.Kind: Mesh` because we wouldn't be able to match

lukidzi:
Ok so it isn't an issue with the MeshTimeout policy because MeshHTTPRoute for sidecar (not gateway) doesn't accept MeshHTTPRoute `to.kind: Mesh` which is a root of the issue

Issue #9380, opened 2024-02-23
Implement targeting `MeshHTTPRoute` in `MeshRateLimit` for `MeshGateway`


lukidzi:
### Description

Implement targeting MeshHTTPRoute from MeshRateLimit for Gateway

PR #9396, opened 2024-02-23
feat(meshratelimit): support targetRef: MeshHTTPRoute for Gateway

lukidzi:
### Checklist prior to review

The best to review by commits.

Added option to support targeting MeshHTTPRoute for MeshRateLimit policy when configuring MeshGateway. 

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/9380


pkg/plugins/policies/meshratelimit/plugin/v1alpha1/plugin_test.go
```diff
@@ -627,73 +631,218 @@ var _ = Describe("MeshRateLimit", func() {
...
+		meshhttproutes core_rules.GatewayRules
+		meshtcproutes  core_rules.GatewayRules
+		rules          core_rules.GatewayRules
+	}
+	DescribeTable("should generate proper Envoy config for MeshGateways",
+		func(given gatewayTestCase) {
+			Expect(given.name).ToNot(BeEmpty())
+			resources := xds_context.NewResources()
+
+			gateway := &core_mesh.MeshGatewayResource{
```
slonka, line 646: can we rewrite all of this to use YAML file as an input? I know it's already been written like this but it makes it really hard to review

lukidzi: I think we could do it as a separate task. I am not fully sure if it's worth adding file policy input just for one test.

slonka (approved):
Might be good if @michaelbeaumont or someone with better gateway knowledge would look at this as well

PR #9405, opened 2024-02-26
feat(tools): generate core resource

jakubdyszkiewicz:
### Checklist prior to review

Modified policy-gen tool to generate non-policy "core" resources like MeshService or ExternalService 2.0.

The pkg for core resources is `pkg/core/resources/apis`. This pkg has `mesh` and `system` subpackages, but I find this separation to be nonsense.

I had to modify K8S registration, because it was in policy plugins package, and it's not great to have a dependency on it in `pkg/core`.

I added DoNothingResource just like DoNothingPolicy, although I'd remove it after we have the first real resource.

Ideally, the tool should be named `resource-gen`, not `policy-gen`, but I did not want to do a big renaming in this PR.


mk/generate.mk
```diff
@@ -46,12 +46,18 @@ resources/type: $(RESOURCE_GEN)
 clean/resources:
 	find pkg -name 'zz_generated.*.go' -delete
 
-POLICIES_DIR := pkg/plugins/policies
+POLICIES_DIR ?= pkg/plugins/policies
 COMMON_DIR := api/common
 
-policies = $(foreach dir,$(shell find pkg/plugins/policies -maxdepth 1 -mindepth 1 -type d | grep -v -e core | sort),$(notdir $(dir)))
+policies = $(foreach dir,$(shell find $(POLICIES_DIR) -maxdepth 1 -mindepth 1 -type d | grep -v -e '/core$$' | grep -v -e '/system$$' | grep -v -e '/mesh$$' | sort),$(notdir $(dir)))
```
lahabana, line 52: Does this work for KM btw?

mk/generate.mk
```diff
@@ -46,12 +46,18 @@ resources/type: $(RESOURCE_GEN)
...
 kuma_policies = $(foreach dir,$(shell find $(KUMA_DIR)/pkg/plugins/policies -maxdepth 1 -mindepth 1 -type d | grep -v -e core | sort),$(notdir $(dir)))
 
+generate/resources: POLICIES_DIR=pkg/core/resources/apis
+generate/resources:
+	POLICIES_DIR=$(POLICIES_DIR) $(MAKE) $(addprefix generate/policy/,$(policies))
+	POLICIES_DIR=$(POLICIES_DIR) $(MAKE) generate/policy-import
+	POLICIES_DIR=$(POLICIES_DIR) HELM_VALUES_FILE_POLICY_PATH=".plugins.resources" $(MAKE) generate/policy-helm
```
lahabana, line 59: All this very alike. Shouldn't we have `RESOURCES_DIR` and have both pkg/core/resources/apis and pkg/plugins/policies?

pkg/core/resources/apis/imports.go
```diff
@@ -0,0 +1,32 @@
...
+	for _, policy := range enabledPluginPolicies {
+		initializer, ok := NameToModule[policy]
+		if ok && !initializer.Initialized {
+			initializer.InitFn()
+			initializer.Initialized = true
+		} else {
+			panic("policy " + policy + " not found")
+		}
+	}
+}
```
lahabana, line 32: Really seems to me that this code doesn't need to be generated (only NameToModule is dynamic here)

pkg/plugins/policies/meshretry/zz_generated.plugin.go
```diff
@@ -1,16 +1,15 @@
...
 	k8s_v1alpha1 "github.com/kumahq/kuma/pkg/plugins/policies/meshretry/k8s/v1alpha1"
 	plugin_v1alpha1 "github.com/kumahq/kuma/pkg/plugins/policies/meshretry/plugin/v1alpha1"
 )
 
 func InitPlugin() {
-	core.Register(
-		api_v1alpha1.MeshRetryResourceTypeDescriptor,
-		k8s_v1alpha1.AddToScheme,
-		plugin_v1alpha1.NewPlugin(),
-	)
```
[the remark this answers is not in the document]
jakubdyszkiewicz: Maybe, but then you need to move this helper from `github.com/kumahq/kuma/pkg/plugins/policies/core` to... ?

It's kind of awkward to pass nil

mk/generate.mk
jakubdyszkiewicz: make sense, let me try this

lobkovilya (approved):
> Ideally, the tool should be named resource-gen, not policy-gen, but I did not want to do a big renaming in this PR.

We already have `tools/resource-gen` for old stuff. Until we drop it I'd stick with `policy-gen` for now to avoid confusion. 

Overall LGTM.

mk/generate.mk
michaelbeaumont: Let's do this in a separate PR @lahabana ?

pkg/config/plugins/policies/policies.go
```diff
@@ -1,3 +1,6 @@
+// Generated by tools/policy-gen
```
lahabana, line 1: Nit but could the file be named `zz_generated_policies.go` ?

pkg/core/resources/apis/modules.go
```diff
@@ -0,0 +1,9 @@
+// Generated by tools/policy-gen
```
lahabana, line 1: Nit but could the file be named `zz_generated_modules.go` ?

PR #9411, opened 2024-02-26
feat(kds): add span for admin requests to zone CPs

michaelbeaumont:
### Checklist prior to review

### DIFF
--- a/pkg/envoy/admin/kds_client.go
+++ b/pkg/envoy/admin/kds_client.go
@@ -7,6 +7,11 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/codes"
+	semconv "go.opentelemetry.io/otel/semconv/v1.24.0"
+	"go.opentelemetry.io/otel/trace"
 
 	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
 	system_proto "github.com/kumahq/kuma/api/system/v1alpha1"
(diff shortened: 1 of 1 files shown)


pkg/envoy/admin/kds_client.go
```diff
@@ -42,124 +53,137 @@ func (k *kdsEnvoyAdminClient) PostQuit(context.Context, *core_mesh.DataplaneReso
...
+	grpc.ReverseUnaryMessage
+	GetError() string
+}
+
+func startTrace(ctx context.Context, tracer trace.Tracer, name string) context.Context {
+	ctx, _ = tracer.Start(
+		ctx,
+		name,
+		trace.WithSpanKind(trace.SpanKindClient),
+		// We fake attributes for the reverse unary gRPC service
```
slonka, line 66: why do we need fake attributes?

slonka (approved):
just Qs and nitpick

pkg/envoy/admin/kds_client.go
michaelbeaumont: "Fake" is maybe the wrong word but the problem is that normally these are set by `otelgrpc`, and we have a kind of gRPC service tunneled through around service here. Perhaps I should just remove `RPCSystemGRPC`

pkg/envoy/admin/kds_client.go
```diff
@@ -42,124 +53,137 @@ func (k *kdsEnvoyAdminClient) PostQuit(context.Context, *core_mesh.DataplaneReso
...
+		trace.WithAttributes(
+			semconv.RPCSystemGRPC,
+			semconv.RPCService(reverseUnaryRPCService),
+			semconv.RPCMethod(name),
+		),
+	)
+	return ctx
+}
+
+func doRequest[T message]( // nolint:nonamedreturns
```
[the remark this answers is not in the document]
michaelbeaumont: the `nolint`? So that we can use `defer`.

pkg/envoy/admin/kds_client.go
michaelbeaumont: Removed

pkg/envoy/admin/kds_client.go
```diff
@@ -42,124 +53,137 @@ func (k *kdsEnvoyAdminClient) PostQuit(context.Context, *core_mesh.DataplaneReso
...
 }
 
-func (k *kdsEnvoyAdminClient) ConfigDump(ctx context.Context, proxy core_model.ResourceWithAddress) ([]byte, error) {
+type message interface {
+	grpc.ReverseUnaryMessage
+	GetError() string
+}
+
+func startTrace(ctx context.Context, tracer trace.Tracer, name string) context.Context {
+	ctx, _ = tracer.Start(
```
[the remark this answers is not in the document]
michaelbeaumont: changed

slonka (approved):

Issue #8928, opened 2024-01-19
Decide if we want to migrate from old metrics stack in DPP to OTEL Prometheus exporter


Automaat:
### Description

Right now for Prometheus we use old server with lots of custom logic, we could replace it with our new `AggregatedProducer` and `Prometheus` exporter for `OpenTelemetry`. 
Benefits:
- simplify DPP code around metrics
- improve performance when using multiple backend, for example `Prometheus` and `OpenTelemetry` at the same time
Drawbacks:
- previously, we were accepting anything that app responded to us, and added this to the response. With this change we would have to accept only Prometheus, or other specific format that we could map to `OpenTelemetry` metrics

PR #9424, opened 2024-02-27
feat(kuma-dp): migrate to prometheus otel sdk when using meshmetric

Automaat:
Fixes: #8928

### Checklist prior to review


app/kuma-dp/pkg/dataplane/metrics/metrics_producer.go
```diff
@@ -85,7 +85,9 @@ func (ap *AggregatedProducer) Produce(ctx context.Context) ([]metricdata.ScopeMe
 func combineMetrics(metrics <-chan *metricdata.ScopeMetrics) []metricdata.ScopeMetrics {
 	var combinedMetrics []metricdata.ScopeMetrics
 	for metric := range metrics {
-		combinedMetrics = append(combinedMetrics, *metric)
+		if metric != nil {
```
slonka, line 88: when can this be nil?

slonka (approved):
one q

app/kuma-dp/pkg/dataplane/metrics/metrics_producer.go
Automaat: when [fetchStats](https://github.com/kumahq/kuma/pull/9424/files#diff-1e652cf18e10c3d480b68f787cb5dba80b4847cc44ea6b53857f24e6c7aeef0cL93) failed, and we could not get metrics or map them. I stumbled upon this while testing this with an application that returned metrics in wrong format

Issue #8942, opened 2024-01-22
Add support for multiple `OpenTelemetry` backends in MeshMetric policy


Automaat:
### Description

Right now, we can configure only one `OpenTelemetry` backend. We need to decide if we want to support multiple `OpenTelemetry` backends and implement it if needed

PR #9445, opened 2024-02-28
feat(policy): add possibility to configure multiple opentelemetry backends

Automaat:
Fixes: #8942

### Checklist prior to review

### DIFF
--- a/pkg/core/xds/sockets.go
+++ b/pkg/core/xds/sockets.go
@@ -22,8 +22,8 @@ func MeshMetricsDynamicConfigurationSocketName(workdir string) string {
 	return socketName(filepath.Join(workdir, "kuma-mesh-metric-config"))
 }
 
-func OpenTelemetrySocketName(workdir string) string {
-	return socketName(filepath.Join(workdir, "kuma-open-telemetry"))
+func OpenTelemetrySocketName(workdir string, backendName string) string {
+	return socketName(filepath.Join(workdir, "kuma-otel-"+backendName))
 }
 
 func socketName(s string) string {
--- a/pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
+++ b/pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
@@ -66,18 +66,17 @@ func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *
 	removeResourcesConfiguredByMesh(rs, listeners.Prometheus, clusters.Prometheus)
 
 	prometheusBackends := filterPrometheusBackends(conf.Backends)
-	// TODO multiple backends of the same type support. Issue: https://github.com/kumahq/kuma/issues/8942
-	openTelemetryBackend := firstOpenTelemetryBackend(conf.Backends)
+	openTelemetryBackends := filterOpenTelemetryBackends(conf.Backends)
 
 	err := configurePrometheus(rs, proxy, prometheusBackends)
 	if err != nil {
 		return err
 	}
-	err = configureOpenTelemetry(rs, proxy, openTelemetryBackend)
+	err = configureOpenTelemetry(rs, proxy, openTelemetryBackends)
 	if err != nil {
 		return err
 	}
-	err = configureDynamicDPPConfig(rs, proxy, conf, prometheusBackends, openTelemetryBackend)
+	err = configureDynamicDPPConfig(rs, proxy, conf, prometheusBackends, openTelemetryBackends)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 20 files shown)


pkg/core/xds/sockets.go
```diff
@@ -22,8 +22,8 @@ func MeshMetricsDynamicConfigurationSocketName(workdir string) string {
 	return socketName(filepath.Join(workdir, "kuma-mesh-metric-config"))
 }
 
-func OpenTelemetrySocketName(workdir string) string {
-	return socketName(filepath.Join(workdir, "kuma-open-telemetry"))
+func OpenTelemetrySocketName(workdir string, backendName string) string {
+	return socketName(filepath.Join(workdir, "kuma-otel-"+backendName))
```
Automaat, line 26: Probably this needs better naming, looking for suggestions

pkg/xds/envoy/names/resource_names.go
```diff
@@ -65,6 +65,14 @@ func GetMetricsHijackerClusterName() string {
 	return Join("kuma", "metrics", "hijacker")
 }
 
+func GetOpenTelemetryListenerName(backendName string) string {
```
Automaat, line 68: Again I am not sure about naming of resources, adding `listener` and `cluster` to names feel weird, but they need to be unique

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -268,3 +283,7 @@ func filterPrometheusBackends(backends *[]api.Backend) []*api.PrometheusBackend
 	}
 	return prometheusBackends
 }
+
+func backendNameFrom(endpoint string) string {
```
Automaat, line 287: I am not sure about this backend name on Kubernetes it looks like: `otel-collector.observability.svc4317` it is unique for backend, and I didn't find any issues with it yet. But this ends up as part of listener and cluster name. Can this end up too long, or be problematic because of characters like: `.` or `-`?

pkg/core/xds/sockets.go
slonka: maybe but I don't think it's really that important 🤷

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
slonka: From docs:

> A string must always contain UTF-8 encoded or 7-bit ASCII text, and cannot be longer than 2^32.

So I don't think it's ever going to be longer than `4 294 967 296`

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin_test.go
```diff
@@ -268,5 +268,46 @@ var _ = Describe("MeshMetric", func() {
 				).
 				Build(),
 		}),
+		Entry("multiple_otel", testCase{
```
slonka, line 271: Can this be done via golden file? it's really hard to read.

pkg/plugins/policies/meshmetric/plugin/v1alpha1/testdata/basic.listeners.golden.yaml
```diff
@@ -22,7 +22,7 @@ resources:
               routes:
               - directResponse:
                   body:
-                    inlineString: '{"observability":{"metrics":{"applications":[{"path":"/metrics","port":8080,"address":""}],"backends":[{"type":"Prometheus"}],"sidecar":{"regex":"http.*","includeUnused":false}}}}'
+                    inlineString: '{"observability":{"metrics":{"applications":[{"path":"/metrics","port":8080,"address":""}],"backends":[{"type":"Prometheus","name":null}],"sidecar":{"regex":"http.*","includeUnused":false}}}}'
```
slonka, line 25: why is the name `null`?

pkg/xds/envoy/names/resource_names.go
slonka: we can't have a cluster and a listener with the same name?

slonka (approved):
In general looks ok, just a couple of Qs

pkg/plugins/policies/meshmetric/plugin/v1alpha1/testdata/basic.listeners.golden.yaml
Automaat: in prometheus we don't need that name, and we are always sending one backend to dataplane, we can change this in the future if we need this

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin_test.go
Automaat: probably, but we never do it via golden file in policies test. I think adding this won't be trivial and I think that there are more important things to take care of

Issue #8924, opened 2024-01-19
Support targeting gateway MeshHTTPRoutes in MeshTimeout


michaelbeaumont:
### Description

If a MeshHTTPRoute targets a MeshGateway, it should be possible to target that MeshHTTPRoute with MeshTimeout

PR #9446, opened 2024-02-28
feat(meshtimeout): added possibility to target MeshHTTPRoute for MeshGateway

lukidzi:
### Checklist prior to review

Added support for MeshGateway. The best to review by commit

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/8924


test/e2e_env/kubernetes/gateway/gateway.go
```diff
@@ -578,6 +579,121 @@ spec:
...
+
+			// no timeout on another route
+			Eventually(func(g Gomega) {
+				start := time.Now()
+				_, err := client.CollectEchoResponse(
+					kubernetes.Cluster, "demo-client", "http://simple-gateway.simple-gateway:8080/no-timeout",
+					client.FromKubernetesPod(clientNamespace, "demo-client"),
+					client.WithHeader("host", "example.kuma.io"),
+					client.WithHeader("x-set-response-delay-ms", "3000"),
+					client.WithMaxTime(10), // we don't want 'curl' to return early
```
slonka, line 689: why would curl return early? 🤔

test/e2e_env/kubernetes/gateway/gateway.go
```diff
@@ -578,6 +579,121 @@ spec:
...
+				)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(time.Since(start)).To(BeNumerically(">", 3*time.Second))
```
slonka, line 692: nitpick: not sure we need this, but it can stay

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
```diff
@@ -193,13 +193,23 @@ func applyToGateway(
...
+						routeConf := getConf(toRules, core_rules.MeshSubset().WithTag(core_rules.RuleMatchesHashTag, r.Name, false))
+						if conf == nil && routeConf == nil {
+							continue
+						}
+						if routeConf == nil && conf != nil {
+							routeConf = conf
+						}
+						if routeConf == nil {
+							continue
+						}
```
slonka, line 208: this seems a bit magic to me, what are we picking here?

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin_test.go
```diff
@@ -541,6 +549,73 @@ var _ = Describe("MeshTimeout", func() {
 				},
 			},
 		},
+	}), Entry("route-level-timeouts", gatewayTestCase{
+		meshhttproutes: core_rules.GatewayRules{
+			ToRules: core_rules.GatewayToRules{
+				ByListenerAndHostname: map[core_rules.InboundListenerHostname]core_rules.Rules{
```
slonka, line 555: if we could make this a golden file, that would be great

slonka (commented):
not sure about the nil / route / routeConf - might need to add some explanation there

test/e2e_env/kubernetes/gateway/gateway.go
lukidzi: Setting explicitly ensures that it won't finish in less than 10 seconds and it should finish in > 3.

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
lukidzi: So the logic is that if there is no config for the specific route we are using config for all routes. 
You might have a configuration for Mesh and MeshHTTPRoute, in case there is no policy for the route, use Mesh one

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin_test.go
lukidzi: IMO we don't have that many tests to introduce that logic. We could maybe have 2-3 and most of tests don't require this

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
slonka: can you add a comment about this? that we fallback

test/e2e_env/kubernetes/gateway/gateway.go
slonka: But why is it important that we won't finish in less than 10 seconds? can you explain this and add a comment?

test/e2e_env/kubernetes/gateway/gateway.go
slonka: same question as ☝️

test/e2e_env/kubernetes/gateway/gateway.go
lukidzi: The default timeout is 5 seconds, and since we timeout in 2 seconds, the default is sufficient. There was no need for a timeout of 10 seconds for curl; it must be higher than the Mesh timeout to prevent curl from failing but still timing out due to the Mesh timeout.

Issue #8746, opened 2024-01-04
Add e2e tests for new policies with delegated gateway


michaelbeaumont:
### Description

Write and run e2e tests for policies applied to a delegated gateway

- [x] https://github.com/kumahq/kuma/pull/8826
- [x] https://github.com/kumahq/kuma/pull/9423
- [x] https://github.com/kumahq/kuma/pull/8832
- [x] https://github.com/kumahq/kuma/pull/9048
- [x] https://github.com/kumahq/kuma/pull/9419
- [x] https://github.com/kumahq/kuma/pull/9437
- [x] https://github.com/kumahq/kuma/pull/9449
- [x] https://github.com/kumahq/kuma/pull/9438
- [x] https://github.com/kumahq/kuma/pull/9447
- [x] https://github.com/kumahq/kuma/pull/9450
- [x] https://github.com/kumahq/kuma/pull/9463
- [x] https://github.com/kumahq/kuma/pull/9520

PR #9449, opened 2024-02-29
test(e2e): add e2e test for MeshHTTPRoute with delegated gateway

bartsmykla:
### Checklist prior to review

  - https://github.com/kumahq/kuma/issues/8746
  - It won't
  - There is no need
  - There is no need

> Changelog: skip

### DIFF
--- a/test/e2e_env/kubernetes/gateway/delegated/meshhttproute.go
+++ b/test/e2e_env/kubernetes/gateway/delegated/meshhttproute.go
@@ -1,0 +1,107 @@
+package delegated
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
+	"github.com/kumahq/kuma/pkg/plugins/policies/meshhttproute/api/v1alpha1"
+	"github.com/kumahq/kuma/test/framework"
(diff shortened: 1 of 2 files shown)


test/e2e_env/kubernetes/gateway/delegated/meshhttproute.go
```diff
@@ -0,0 +1,107 @@
...
+			Eventually(func(g Gomega) {
+				response, err := client.CollectEchoResponse(
+					kubernetes.Cluster,
+					"demo-client",
+					fmt.Sprintf("http://%s/test-server", config.KicIP),
+					client.FromKubernetesPod(config.NamespaceOutsideMesh, "demo-client"),
+				)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(response.Instance).To(HavePrefix("external-service"))
+			}, "30s", "1s").Should(Succeed())
```
slonka, line 104: Are we probing just for one request? This is going to be flaky, we can't guarantee in which order Envoy will route, right? I would change this to issue 40 requests and check that the ratio is around 50% each.

bartsmykla: It's the direct copy of our normal existing test (without delegated gateway). I'm also not sure how would it be flaky as it would be only possible if for 30s all requests would go to just one service, and if it would happen than that would indicate there might be an issue.

slonka (approved):

PR #9472, opened 2024-03-01
fix(MeshHTTPRoute): order rules by match priority

### LINKED ISSUE #9428: Sort `MeshHTTPRoute` for `kind: MeshService` according to Gateway API rules
### What happened?

https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteRule

michaelbeaumont:
### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- Closes #9428 


pkg/plugins/policies/meshhttproute/plugin/v1alpha1/testdata/match-priority.listeners.golden.yaml
```diff
@@ -0,0 +1,71 @@
...
+                  cluster: backend
+                  timeout: 0s
+              - match:
+                  path: /v1
+                name: bzv64o+joRk33jfDADgGv5ar4QyONO0NEVKnoBfJW/E=
+                route:
+                  cluster: backend
+                  timeout: 0s
+              - match:
+                  prefix: /v1/
```
lukidzi, line 44: Shouldn't it be higher because it's longer than the one above? I'm uncertain if this applies to the path that we generated.

lukidzi (approved):

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare.go
```diff
@@ -0,0 +1,150 @@
...
+	Filters     []Filter
+	BackendRefs []common_api.BackendRef
+	Hash        string
+}
+
+// SortRules orders the rules according to Gateway API precedence:
+// https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteRule
+func SortRules(rules []Rule) []Route {
+	type keyed struct {
+		sortKey Match
```
lahabana, line 125: Shoudn't Match simply implemet the `Ordered` interface?

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare.go
```diff
@@ -0,0 +1,150 @@
...
+	case a != nil && b == nil:
+		return -1
+	case a == nil && b != nil:
+		return 1
+	case a == nil && b == nil:
+		return 0
+	}
+
+	switch {
+	case a.Type == Exact && b.Type == Exact:
```
lahabana, line 21: Shouldn't it return 0 in this case?

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare.go
```diff
@@ -0,0 +1,150 @@
...
+
+type Route struct {
+	Match       Match
+	Filters     []Filter
+	BackendRefs []common_api.BackendRef
+	Hash        string
+}
+
+// SortRules orders the rules according to Gateway API precedence:
+// https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteRule
```
lahabana, line 122: The spec mentions:
> Note: The precedence of RegularExpression path matches are implementation-specific.

We should make this explicit what our behaviour is in this comment

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare.go
lahabana: Would this be a bit simpler to rewrite this this way:

```go
switch{
case a.Type == b.Type:
   // imperfect syntax
   return cm.Compare(a.Value, b.Value)
case a.Type == Exact:
   return -1
case a.Type == PathPrefix:
  if b.Type == Exact { return 1 } else { return -1}
case a.Type == RegularExpression:
  return 1
}
```

In plain text:
- If types is the same order by length (for exact it's not required but at least we have an absolute order).
- If a.Type lower priority than b.Type then a.Type is priority (and vice versa)

We order the cases by priority so we don't have to do all types always.

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	switch {
+	case a.Type == PathPrefix && b.Type != PathPrefix:
+		return -1
+	case a.Type != PathPrefix && b.Type == PathPrefix:
+		return 1
+	case a.Type == PathPrefix && b.Type == PathPrefix:
+		switch {
+		case len(a.Value) > len(b.Value):
```
lahabana, line 35: Any reason not to use `cmp.Compare` ?

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare_test.go
```diff
@@ -0,0 +1,156 @@
...
+		}))).To(Equal([]api.Route{{
+			Match: exactMatch,
+			Hash:  api.HashMatches(singleOrExact),
+		}, {
+			Match: prefixMatch,
+			Hash:  api.HashMatches(prefixMatchOnly),
+		}, {
+			Match: singleHeaderMatch,
+			Hash:  api.HashMatches(singleOrExact),
+		}}))
```
lahabana, line 154: Might make sense to have a test with no rules or tests with exact match but different methods

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare.go
michaelbeaumont: Why should it return 0? What if one has a method match and the other doesn't? I find that switch statement pretty hard to read, to be honest. We don't compare length for exact matches, just prefix matches.

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare.go
michaelbeaumont: Wouldn't that just do a lexicographic compare of the strings?

michaelbeaumont: Yeah this could be something like

```
if lenCmp := cmp.Compare(len(a.Value), len(b.Value)); lenCmp != 0 {
  return lenCmp
}
```

I guess

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare.go
michaelbeaumont: I don't think non primitive types can implement that interface.

pkg/plugins/policies/meshhttproute/api/v1alpha1/compare.go
michaelbeaumont: > If types is the same order by length (for exact it's not required but at least we have an absolute order).

We actually need them not to be reordered because we need stability when sorting.

michaelbeaumont: I fixed this but made it more explicit because I prefer not to have an implicit "RegularExpression is the 'last' kind" assumption built in.

lahabana (approved):

PR #9477, opened 2024-03-04
test(e2e/framework): don't fail when deleting non-existing k8s namespace

bartsmykla:
When removing kic deployment we are removing a configured ingress namespace. There might be situation when this namespace might be already deleted during different test suite stage, so if doesn't exist, we don't have to fail during the deletion.

Updated kic deployment delete function to call `TraiggerDeleteNamespace` (which potentially might be more performant) instead of `DeleteNamespace`.

### Checklist prior to review

  - No relevant issues
  - It won't
  - There is no need
  - There is no need

> Changelog: skip

### DIFF
--- a/test/framework/deployments/kic/kubernetes.go
+++ b/test/framework/deployments/kic/kubernetes.go
@@ -82,7 +82,7 @@ func (t *k8sDeployment) Deploy(cluster framework.Cluster) error {
 }
 
 func (t *k8sDeployment) Delete(cluster framework.Cluster) error {
-	return cluster.DeleteNamespace(t.ingressNamespace)
+	return cluster.(*framework.K8sCluster).TriggerDeleteNamespace(t.ingressNamespace)
 }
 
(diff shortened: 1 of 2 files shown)


slonka (approved):

test/framework/deployments/kic/kubernetes.go
```diff
@@ -82,7 +82,7 @@ func (t *k8sDeployment) Deploy(cluster framework.Cluster) error {
 }
 
 func (t *k8sDeployment) Delete(cluster framework.Cluster) error {
-	return cluster.DeleteNamespace(t.ingressNamespace)
+	return cluster.(*framework.K8sCluster).TriggerDeleteNamespace(t.ingressNamespace)
```
lahabana, line 85: We use helm to install. Why don't we use helm to uninstall? Also I don't see why we delete a namespace that wasn't created in install...

PR #9479, opened 2024-03-04
ci(github): collapse auto-backport and check

lahabana:
In a single PR check job. This may reduce job queueing as we'll have 1 instead of 2 job ask on PR changes.
It also removes an unecessary go-setup which was taking 10s

### Checklist prior to review


.github/workflows/check.yaml
```diff
@@ -8,30 +8,37 @@ on:
...
+        # https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#merge-message-for-a-squash-merge
+        # for more info. We have "Default to PR title for squash merge commits" enabled.
+        if: github.event.action != 'synchronize'
         # Inject as env variable to escape properly
         env:
           TITLE: ${{ github.event.pull_request.title }}
         run: |
+          npm install -g @commitlint/cli@19.0.1 @commitlint/config-conventional@19.0.0
           commitlint --config .github/commitlint.config.js --edit=<(echo "${TITLE}")
+      - name: Maybe add backport label
```
slonka, line 36: I don't think this works properly :( didn't have time to check why that's happening. Should we use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore instead? (like Mike suggested some time ago). 

E.g. this one https://github.com/kumahq/kuma/pull/9430#event-11946827074 - Bart had to add a label himself.

slonka (approved):

.github/workflows/check.yaml
lahabana: Yeah I've noticed it not working properly. 

Another option is to use:

```
$ gh pr view 9479 --json files  --jq '.files[].path'
.github/workflows/auto-backport.yaml
.github/workflows/check.yaml
```

In any case that's a different issue than merging this inline: https://github.com/kumahq/kuma/issues/9482

PR #9500, opened 2024-03-05
test(container-structure): skip iptable test on kuma-init

lahabana:
When using nftables based iptables it fails on arm64 because of https://github.com/multiarch/qemu-user-static/issues/191

### Checklist prior to review

### DIFF
--- a/test/container-structure/kuma-init.yaml
+++ b/test/container-structure/kuma-init.yaml
@@ -10,10 +10,13 @@ commandTests:
 - name: "Contains kumactl"
   command: kumactl
   args: ["version"]
-- name: "Contains iptables"
-  command: iptables
-  args: ["--version"]
-  expectedOutput: ["iptables v.*"]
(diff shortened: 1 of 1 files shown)


bartsmykla (approved):

michaelbeaumont (approved):

test/container-structure/kuma-init.yaml
```diff
@@ -10,10 +10,13 @@ commandTests:
 - name: "Contains kumactl"
   command: kumactl
   args: ["version"]
-- name: "Contains iptables"
-  command: iptables
-  args: ["--version"]
-  expectedOutput: ["iptables v.*"]
+# Skipping iptables test as it fails on arm64 with qemu-user-static
```
slonka, line 13: Normally I'd suggest skipping only on arm64 but if iptables is not present at all then other tests will fail so 👍

slonka (approved):

PR #9504, opened 2024-03-05
feat(meshgatewayinstance): deprecate kuma.io/service and generate serviceName

lukidzi:
### Checklist prior to review

To increase the security of `MeshGateway` deployment on k8s we decided to deprecate setting custom `kuma.io/service` tag value and instead of it generating one: `${MeshGatewayInstance name}_${MeshGatewayInstance namespace}_svc`. Because it's a breaking change we decided first to deprecate it and support the new format. 

* deprecate `kuma.io/service` in `MeshGatewayInstance.Spec.Tags`
* generate `kuma.io/service` when not set
* moved one e2e suite to use new service name
* added upgrade.md


pkg/plugins/runtime/k8s/controllers/gateway_instance_controller.go
```diff
@@ -112,6 +127,7 @@ func (r *GatewayInstanceReconciler) createOrUpdateService(
 		return nil, nil, err
 	}
 	gateway := xds_topology.SelectGateway(gatewayList.Items, func(selector mesh_proto.TagSelector) bool {
+		r.resolveGatewayInstancServiceTag(gatewayInstance)
```
michaelbeaumont, line 130:
```suggestion
		r.resolveGatewayInstanceServiceTag(gatewayInstance)
```

michaelbeaumont (approved):

pkg/plugins/runtime/k8s/controllers/gateway_instance_controller.go
```diff
@@ -100,6 +101,20 @@ func k8sSelector(name string) map[string]string {
 	return map[string]string{"app": name}
 }
 
+func (r *GatewayInstanceReconciler) resolveGatewayInstancServiceTag(gatewayInstance *mesh_k8s.MeshGatewayInstance) {
```
lahabana, line 104:
```suggestion
func (r *GatewayInstanceReconciler) resolveGatewayInstanceServiceTag(gatewayInstance *mesh_k8s.MeshGatewayInstance) {
```

pkg/plugins/runtime/k8s/controllers/gateway_instance_controller.go
```diff
@@ -421,20 +437,37 @@ func GatewayToInstanceMapper(l logr.Logger, client kube_client.Client) kube_hand
...
+			if err != nil {
+				l.Error(err, "failed to fetch GatewayInstances")
+				continue
+			}
+			instance := &mesh_k8s.MeshGatewayInstance{}
+			err = client.Get(ctx, name, instance)
+			if err != nil {
+				l.Error(err, "failed to fetch GatewayInstances")
+				continue
 			}
```
lahabana, line 466: I think the right behaviour is to enqueue the request and then the reconcile ignores the request if the entity is not found.

Issue #8510, opened 2023-12-02
Automatically convert JSON patch paths in `MeshProxyPatch`es from `snake_case` to `camelCase`


bartsmykla:
### Description

Currently, Envoy's config dump presents paths in `snake_case` format, while JSON patches in `MeshProxyPatch` require `camelCase` format. Instead of documenting this discrepancy, we can implement logic to automatically convert paths from `snake_case` to `camelCase` when encountered. This approach would eliminate the need for users to manually convert paths, improving the overall usability of the system.

---

xref. https://github.com/kumahq/kuma-website/pull/1539

lukidzi:
I did some debugging, it is caused by not setting OrigName to true https://github.com/kumahq/kuma/blob/master/pkg/util/proto/proto.go#L44
if we change it to this we are going to have proto fields name (like in a config_dump)

`MeshProxyPatch` allows the creation of JSON patch and accepts JSON formatted fields `camelCase`; if we change the dump, a user might have to rewrite them. The same is true with the value field.

Envoy's `config_dump` shows only `snake_case` when directly requesting the envoy, but not through the Control-plane.

I think this issue might not be the case if we agree that what shows UI is a source of truth, but if we decide that we should stick to the same format as Envoy(which would be the best for cross docs reading) then we need to rewrite a lot of things and we need to keep backward compatibility for old patches in JSON format

jakubdyszkiewicz:
> Triage: let's change API so the config dump is `snake_cased` (just like 9901) not `camelCased`

lahabana:
This because of the sanitization:

https://github.com/kumahq/kuma/blob/62335211632105c2efacbc3045cd0990e2079cb8/pkg/envoy/admin/client.go#L145-L161

How about we retrieve the tokens that needs to be redacted in the first step and then redacted from the raw payload? This would avoid the reserialization.

lukidzi:
We are sanitizing `DataplaneToken` from the output. We don't store it anywhere so we won't be able to retrieve it.

lahabana:
What I mean is something like:

```go
	configDump, err := a.executeRequest(ctx, proxy, "config_dump")
	if err != nil {
		return nil, err
	}

        toReplace := []string{}
	cd := &envoy_admin_v3.ConfigDump{}
	if err := util_proto.FromJSON(configDump, cd); err != nil {
		return nil, err
	}
	for _, config := range configDump.Configs {
[...]

PR #9519, opened 2024-03-06
feat(admin): return `config_dump` response in the same format as envoy admin 

lukidzi:
### Checklist prior to review

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/8510


pkg/plugins/policies/core/jsonpatch/merge_test.go
```diff
@@ -67,9 +67,17 @@ var _ = Describe("Json Patch merge", func() {
 		}
 
 		// when
-		_, err := jsonpatch.MergeJsonPatchAny(util_proto.MustMarshalAny(hcm), patches)
+		hcmAny := util_proto.MustMarshalAny(hcm)
+		mergedHcmAny, err := jsonpatch.MergeJsonPatchAny(hcmAny, patches)
```
slonka, line 71: how does it create camelCase?

slonka (approved):
if it does not screw up anything else, and  you're sure in your test you're creating `camelCase`

pkg/plugins/policies/core/jsonpatch/merge_test.go
lukidzi: I've used a library but if we think it is error prone we could try do regex and match->replace https://github.com/gobeam/stringy/blob/master/stringy.go#L112

lahabana:
How come the dump and the values do not match?

lukidzi:
> How come the dump and the values do not match?

I did some debugging, it is caused by not setting `OrigName` to true https://github.com/kumahq/kuma/blob/master/pkg/util/proto/proto.go#L44
```golang
marshaler := &jsonpb.Marshaler{
		OrigName: true,
	}
```

if we change it to this we are going to have proto fields name (like in a config_dump)

`MeshProxyPatch` allows the creation of `jsonPatch` and accepts JSON formatted fields `camelCase`; if we change the dump, a user might have to rewrite them. The same is true with the `value` field.

lukidzi:
I think this issue might not be the case if we agree that what presents UI is true, but if we decide that we should stick to the same format as Envoy(which would be the best for cross docs reading) then we need to rewrite a lot of things and we need to keep backward compatibility for old patches in JSON format

jakubdyszkiewicz (approved):

pkg/envoy/admin/testdata/golden.full_config.json
```diff
@@ -7,41 +7,41 @@
           "id": "default.backend-1",
```
lukidzi, line 7: When compared with the source file it has only one change:
image

pkg/envoy/admin/sanitize.go
```diff
@@ -1,35 +1,42 @@
...
+	"strings"
 
-func Sanitize(configDump *envoy_admin_v3.ConfigDump) error {
-	for _, config := range configDump.Configs {
+	envoy_admin_v3 "github.com/envoyproxy/go-control-plane/envoy/admin/v3"
+
+	util_proto "github.com/kumahq/kuma/pkg/util/proto"
+)
+
+func Sanitize(configDump []byte) ([]byte, error) {
```
lukidzi, line 11: @lahabana I've changed to the way you mentioned in the issue

api/common/v1alpha1/jsonpatch.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"strconv"
 
-	"github.com/evanphx/json-patch/v5"
+	jsonpatch "github.com/evanphx/json-patch/v5"
```
lahabana, line 8: Feels like it's left over
```suggestion
                    "github.com/evanphx/json-patch/v5"
```

PR #9520, opened 2024-03-07
test(e2e): add e2e test for MeshTCPRoute with delegated gateway

bartsmykla:
### Checklist prior to review

  - https://github.com/kumahq/kuma/issues/8746
  - It won't
  - There is no need
  - There is no need

> Changelog: skip

### DIFF
--- a/test/e2e_env/kubernetes/gateway/delegated/meshtcproute.go
+++ b/test/e2e_env/kubernetes/gateway/delegated/meshtcproute.go
@@ -1,0 +1,111 @@
+package delegated
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
+	"github.com/kumahq/kuma/pkg/plugins/policies/meshtcproute/api/v1alpha1"
+	"github.com/kumahq/kuma/test/framework"
+	"github.com/kumahq/kuma/test/framework/client"
+	"github.com/kumahq/kuma/test/framework/envs/kubernetes"
+	"github.com/kumahq/kuma/test/server/types"
+)
+
+func MeshTCPRoute(config *Config) func() {
+	GinkgoHelper()
+
+	return func() {
+		framework.E2EAfterEach(func() {
+			Expect(framework.DeleteMeshResources(
+				kubernetes.Cluster,
+				config.Mesh,
+				v1alpha1.MeshTCPRouteResourceTypeDescriptor,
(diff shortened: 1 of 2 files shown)


test/e2e_env/kubernetes/gateway/delegated/meshtcproute.go
```diff
@@ -0,0 +1,111 @@
...
+
+			// then
+			Eventually(func() ([]types.EchoResponse, error) {
+				return client.CollectResponses(
+					kubernetes.Cluster,
+					"demo-client",
+					fmt.Sprintf("http://%s/test-server", config.KicIP),
+					client.FromKubernetesPod(config.NamespaceOutsideMesh, "demo-client"),
+					client.WithNumberOfRequests(10),
+					client.WithHeader("x-set-response-delay-ms", "2000"),
```
slonka, line 101: why do we need response delay here?

test/e2e_env/kubernetes/gateway/delegated/meshtcproute.go
```diff
@@ -0,0 +1,111 @@
...
+					client.WithoutRetries(),
+				)
+			}, "30s", "1s").Should(ContainElements(
```
slonka, line 104: should we add `MustPassRepeatedly` so it does not pass if the behaviour is flaky?

test/e2e_env/kubernetes/gateway/delegated/meshtcproute.go
bartsmykla: We do, because if request is quick there is high chanse of tcp connection being reused

test/e2e_env/kubernetes/gateway/delegated/meshtcproute.go
```diff
@@ -0,0 +1,111 @@
...
+`, config.CpNamespace, config.Mesh))(kubernetes.Cluster)).To(Succeed())
+
+			// then
+			Eventually(func() ([]types.EchoResponse, error) {
+				return client.CollectResponses(
+					kubernetes.Cluster,
+					"demo-client",
+					fmt.Sprintf("http://%s/test-server", config.KicIP),
+					client.FromKubernetesPod(config.NamespaceOutsideMesh, "demo-client"),
+					client.WithNumberOfRequests(10),
```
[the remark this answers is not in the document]
bartsmykla: It's inside `Eventually` block, so it shouldn't be flaky

test/e2e_env/kubernetes/gateway/delegated/meshtcproute.go
bartsmykla: I'm gonna follow up with this addition if it ever flakes

Issue #9349, opened 2024-02-22
Fix flakines in universal reachable services e2e test


bartsmykla:
### What happened?

[Failed Job: 21729081749](https://github.com/kumahq/kuma/actions/runs/7960326410/job/21729081749)

```
Summarizing 1 Failure:
  [FAIL] Reachable Services [It] should not be able to non reachable services
  github.com/kumahq/kuma/test/e2e_env/universal/reachableservices/reachableservices.go:50
```

---

Logs archive: [logs_59272.zip](https://github.com/kumahq/kuma/files/14372389/logs_59272.zip)

```
• [FAILED] [9.439 seconds]
Reachable Services [It] should not be able to non reachable services
github.com/kumahq/kuma/test/e2e_env/universal/reachableservices/reachableservices.go:41

  Captured StdOut/StdErr Output >>
  Reachable Services should not be able to non reachable services 2024-02-19T13:45:48Z ssh.go:24: Running /usr/bin/ssh -q -tt -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@localhost -p 32786 curl --max-time 5 --fail --request GET --silent --write-out %{json} --output /dev/null 'second-test-server.mesh'
  Reachable Services should not be able to non reachable services 2024-02-19T13:45:48Z ssh.go:24: Running /usr/bin/ssh -q -tt -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@localhost -p 32786 curl --max-time 5 --fail --request GET --silent --write-out %{json} --output /dev/null 'second-test-server.mesh'
  Reachable Services should not be able to non reachable services 2024-02-19T13:45:49Z ssh.go:24: Running /usr/bin/ssh -q -tt -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@localhost -p 32786 curl --max-time 5 --fail --request GET --silent --write-out %{json} --output /dev/null 'second-test-server.mesh'
  Reachable Services should not be able to non reachable services 2024-02-19T13:45:49Z ssh.go:24: Running /usr/bin/ssh -q -tt -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@localhost -p 32786 curl --max-time 5 --fail --request GET --silent --write-out %{json} --output /dev/null 'second-test-server.mesh'
[...]

PR #9535, opened 2024-03-07
test(e2e): enable reachable services test and increase interval

lukidzi:
### Checklist prior to review

It seems a time of default 5s https://github.com/kumahq/kuma/blob/master/pkg/test/ginkgo.go#L39 in consistently wasn't enough in this case

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/9349


jakubdyszkiewicz:
> The default time is 100ms with a 10ms interval

Is it though? https://github.com/kumahq/kuma/blob/master/pkg/test/ginkgo.go#L39

lukidzi:
> 

> > The default time is 100ms with a 10ms interval
> 
> Is it though? https://github.com/kumahq/kuma/blob/master/pkg/test/ginkgo.go#L39

Didn't notice this configuration 😕

slonka:
@jakubdyszkiewicz and @lukidzi what's the way forward here? should we run it a couple of times to see it's not flaky any more and merge it?

michaelbeaumont:
@lukidzi should we close this until it can make progress?

jakubdyszkiewicz:
We now have much more stable CI. Let's merge it if it passes. If it fails we will revert once again

jakubdyszkiewicz (approved):

PR #9538, opened 2024-03-08
feat(xds): don't merge listeners of headless service instances

jijiechen:
Should be part of the cause of issue 9006, but may not be complete.

xref: https://github.com/kumahq/kuma/issues/9006

### Checklist prior to review

  - Described
  - Confirmed
  - Tested using unit tests and manually
  - No
  - No, this is new in 2.6.1


jakubdyszkiewicz (changes requested):
I don't think it's a correct fix. You can have multiple tags in outbound in the end. I think we need to compare set of tags in `Outbound#Tags`

jakubdyszkiewicz:
Closing for more general fix https://github.com/kumahq/kuma/pull/9542

PR #9539, opened 2024-03-08
chore(workflow): deduct the least common version for security updates

lobkovilya:
Today the automation picks the lowest version. It doesn't work if there are 2 CVEs, i.e. for helm.sh/helm/v3:
* CVE-1 fixed in 3.14.1
* CVE-2 fixed in 3.14.2

It won't be correct to update to 3.14.1 because the second CVE won't be fixed.

At the same time, we don't want to take the newest version, i.e.:
* CVE-1 fixed in 1.0.1 and 1.1.0
* CVE-2 fixed in 1.0.2

It won't be correct to update to 1.1.0, because a bigger version jump most likely requires code changes.

### Checklist prior to review


michaelbeaumont (approved):

mk/check.mk
```diff
@@ -89,4 +89,4 @@ check: format/common lint ## Dev: Run code checks (go fmt, go vet, ...)
 
 .PHONY: update-vulnerable-dependencies
 update-vulnerable-dependencies:
-	@$(KUMA_DIR)/tools/ci/update-vulnerable-dependencies.sh
+	@$(KUMA_DIR)/tools/ci/update-vulnerable-dependencies/update-vulnerable-dependencies.sh
```
lahabana, line 92: Can we avoid a shell script that calls a go tool? I think we can as all of the go toolset is available in the SDK.
Also we could also wrap the osv-scanner in there.

Maybe it's even worth moving this to: https://github.com/kumahq/ci-tools

lahabana: cf: https://pkg.go.dev/golang.org/x/mod/modfile

lahabana: I think the whole tool would a lot simpler too.

PR #9540, opened 2024-03-08
chore(e2e): slow down requests and cleanup

lukidzi:
### Checklist prior to review

* added delay between requests because it seems there might be some overload of envoy - curl return 28 which means connection timeout
* added cleanup of resources between VO test
* added more retries on cluster setup for multizone
* 


test/framework/client/collect.go
```diff
@@ -102,6 +113,12 @@ func WithoutRetries() CollectResponsesOptsFn {
 	}
 }
 
+func WithoutConcurrentRequestDelay() CollectResponsesOptsFn {
+	return func(opts *CollectResponsesOpts) {
+		opts.withoutConcurrentRequestDelay = true
```
lahabana, line 118: Why not just use 0?

test/framework/client/collect.go
```diff
@@ -84,12 +88,19 @@ func WithCACert(cacert string) CollectResponsesOptsFn {
 	}
 }
 
-func WithMaxConcurrentReqeusts(maxConcurrentRequests int) CollectResponsesOptsFn {
+func WithMaxConcurrentRequests(maxConcurrentRequests int) CollectResponsesOptsFn {
```
lahabana, line 91: Unit please :)

test/framework/client/collect.go
```diff
@@ -25,33 +26,36 @@ import (
...
 	return CollectResponsesOpts{
-		numberOfRequests:      10,
-		maxConcurrentRequests: 10,
-		maxTime:               5,
-		verbose:               false,
-		Method:                "GET",
-		Headers:               map[string]string{},
-		ShellEscaped:          utils.ShellEscape,
+		numberOfRequests:          10,
+		maxConcurrentRequests:     10,
```
lahabana, line 52: Should we reduce this? like 5 or something like this?

lukidzi: We can do it because it produces a bit of load on Envoy.

.github/workflows/_e2e.yaml
```diff
@@ -130,6 +130,9 @@ jobs:
             export GINKGO_E2E_LABEL_FILTERS="job-${{ matrix.parallelRunnerId }}"
           fi
           env
+          if [[ "${{ env.E2E_PARAM_TARGET }}" == "multizone" ]]; then
```
lukidzi, line 133:
```
Summarizing 3 Failures:
  [FAIL] [SynchronizedBeforeSuite] 
  github.com/kumahq/kuma/test/framework/ginkgo.go:46
  [FAIL] [SynchronizedBeforeSuite] 
  github.com/kumahq/kuma/test/framework/envs/multizone/env.go:106
  [PANICKED!] [DeferCleanup (Suite)] 
```

there was a message:
```
          Err: <retry.MaxRetriesExceeded>{
              Description: "get default mesh",
              MaxRetries: 30,
          },```

want to see if that helps

lukidzi: seems to help

Automaat (approved):

PR #9542, opened 2024-03-08
fix(xds): duplicated listeners

### LINKED ISSUE #9006: Duplicated listeners after federating a zone
### What happened?

got:

```
 2024-01-23T09:59:44.403Z    INFO    xds.nack-backoff    config was previously rejected by Envoy. Applying backoff before resending it    {"backoff": "5s", "nodeID": "default.demo-app-c7cd6588b-wxn6t.kuma-demo", "reason": "Error adding/updating listener(s) outbound:2 │
│ 40.0.0.1:80: error adding listener: 'outbound:240.0.0.1:80' has duplicate address '240.0.0.1:80' as existing listener\noutbound:240.0.0.0:80: error adding listener: 'outbound:240.0.0.0:80' has duplicate address '240.0.0.0:80' as existing listener\n"}                 │
```

after federating a zone with demo installed, @bartsmykla mentioned that this could be related to https://github.com/kumahq/kuma/issues/6717 - that's why I assigned @jijiechen

Steps to reproduce:
1. deploy a zone
2. install demo
3. deploy a global
4. export resources using `kumactl export`
5. federate zone (kumactl install with kds global address)
6. check logs

jakubdyszkiewicz:
### Checklist prior to review

Fix #9006 (at least for headless services).

I've rewritten the `buildAdditionalAddresses` function. It now just groups outbound by tags and puts addresses in proper order. It's O(n) as well. I also don't get why `08` test case was not squashed to additional addresses before.

### DIFF
--- a/pkg/xds/generator/outbound_proxy_generator_test.go
+++ b/pkg/xds/generator/outbound_proxy_generator_test.go
@@ -889,6 +889,36 @@ var _ = Describe("OutboundProxyGenerator", func() {
 `,
 			expected: "10.envoy.golden.yaml",
 		}),
+		Entry("11. service vips with outbound of multiple tags (headless service)", testCase{
+			ctx: serviceVipCtx,
+			dataplane: `
+            networking:
+              address: 10.0.0.1
+              inbound:
+              - port: 8080
+                tags:
+                  kuma.io/service: web
+              outbound:
+              - port: 80
+                address: 240.0.0.3
+                tags:
+                  kuma.io/service: backend
+              - port: 80
+                address: 10.0.0.1
+                tags:
+                  kuma.io/service: backend
+                  kuma.io/instance: instance-1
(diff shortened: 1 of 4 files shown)


pkg/xds/generator/outbound_proxy_generator_test.go
```diff
@@ -889,6 +889,36 @@ var _ = Describe("OutboundProxyGenerator", func() {
 `,
 			expected: "10.envoy.golden.yaml",
 		}),
+		Entry("11. service vips with outbound of multiple tags (headless service)", testCase{
+			ctx: serviceVipCtx,
+			dataplane: `
+            networking:
+              address: 10.0.0.1
```
lukidzi, line 896: Should this case have a different address?

lukidzi (approved):

Issue #5039, opened 2022-09-21
How to manage transition from old to new policies


lahabana:
### Description

We should start thinking what it means to have both old and new policies.
In some cases both may be able to coexist (Observability for example), in others a decision has to be made (TrafficPermissions).

We should write a MADR to discuss each of these the idea is that a user should have a smooth way to transition from one method to the other.

Issue #9151, opened 2024-02-06
odd behaviour with multiple policies with same top level targetRef


lahabana:
### What happened?

I've created a branch with a test Case: https://github.com/lahabana/kuma/commit/62c17a5c0bc8a5c5ac9061c9f32738322edba538

The idea is to have multiple policies with the same top level targetRef but different `to` targetRefs.

One is at Mesh and the other is MeshService.
What happens is that we sort them by policy name and completely lose the `meshService` specific entry.

I believe this is not what a user would expect. At the same level we should sort lexicography but also by kind.

So instead of having an intermediary representation:

```yaml
    - targetRef: # from mesh-timeout-all-default
        kind: Mesh
      default:
        connectionTimeout: 5s
        idleTimeout: 1h0m0s
        http:
          requestTimeout: 15s
          streamIdleTimeout: 30m0s
    - targetRef: # from mesh-timeout-1
        kind: Mesh
      default:
        http:
          requestTimeout: 1s
    - targetRef: # from mesh-timeout-4
        kind: MeshService
        name: api-play-002_microservice-mesh_svc_8080
      default:
        http:
          requestTimeout: 3s
```

It should be:

```yaml
    - targetRef: # from mesh-timeout-4
        kind: MeshService
        name: api-play-002_microservice-mesh_svc_8080
      default:
        http:
          requestTimeout: 3s
    - targetRef: # from mesh-timeout-all-default
        kind: Mesh
      default:
        connectionTimeout: 5s
        idleTimeout: 1h0m0s
        http:
          requestTimeout: 15s
          streamIdleTimeout: 30m0s
    - targetRef: # from mesh-timeout-1
        kind: Mesh
      default:
        http:
          requestTimeout: 1s
```

I'm not sure if it's a bug or if we're diverging from the design we already thought about

lahabana:
Clarification: the problem is that we'd expect the requestTimeout to be 1s not 3s when talking to api-play-002

slonka:
> Clarification: the problem is that we'd expect the requestTimeout to be 1s not 3s when talking to api-play-002

why would we expect it to be `1s` when `3s` has a more specific [targetRef: MeshService](https://github.com/lahabana/kuma/commit/62c17a5c0bc8a5c5ac9061c9f32738322edba538#diff-67bd3c0da4e254a827bfdc151a192a32ac08e66bbc57f640e7e6d2aee741f8bdR45-R46)?

lobkovilya:
[...]

PR #9571, opened 2024-03-12
docs(MADR): transition to new policies

lobkovilya:
### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/5039


docs/madr/decisions/038-transition-to-new-policies.md
```diff
@@ -0,0 +1,473 @@
...
+  to: [...]
+```
+
+it doesn't apply on the real proxies. Instead, by using a special endpoint we can get a list of jsonPatches that would be applied to the Envoy configuration
+
+```shell
+$ curl :5681/meshes/defaults/dataplane/dpp-1/xds?shadow=true # show potential changes for a dpp-1 in the "default" mesh
+```
+
+Implementation wise, shadow mode requires generating 2 snapshots: one with and without the new policy.
```
lukidzi, line 53: How would it behave if we have one MeshTimeout policy (global) with shadow and another MeshTimeout without shadow. After merge is it going to be shadow or not?

docs/madr/decisions/038-transition-to-new-policies.md
```diff
@@ -0,0 +1,473 @@
...
+            threshold: 36
+```
+
+Using shadow mode we see there is an empty jsonPatches list!
+
+### FaultInjection -> MeshFaultInjection
+
+It should be always safe to remove FaultInjection and create new MeshFaultInjection when necessary. 
+
+### TrafficPermission -> MeshTrafficPermission
```
lukidzi, line 364: Do we need to provide more explanation for MeshTrafficPermission since in 2.6 it creates HTTP rbac configuration?

docs/madr/decisions/038-transition-to-new-policies.md
```diff
@@ -0,0 +1,473 @@
...
+## ProxyTemplate -> MeshProxyTemplate
+
+This is where shadow mode shines the most even not in the context of the transition but for a regular usage.
+
+```yaml
+type: MeshProxyPatch
+mesh: default
+name: custom-template-1
+labels:
+  kuma.io/dryrun: yesplease
```
lukidzi, line 431: Should it be also shadow?

docs/madr/decisions/038-transition-to-new-policies.md
```diff
@@ -0,0 +1,473 @@
...
+- Status: accepted
+
+## Context and Problem Statement
+
+The current state of the policies as of Kuma 2.6.x:
+* new policies are in GA, they're a recommended way to configure new deployments in Kuma
+* old policies are not deprecated, they're still in use by majority of Kuma users
+
+The goal is to provide a way to transition from old policies to new policies. Having a smooth transition is going to enable 
+us to deprecate old policies in the future.
```
lahabana, line 12: We should refresh people's mind in this MADR by mentioning that old policies do not apply as soon as new policies are applied. I think it's important to understand this MADR

lahabana (commented):
Sounds interesting. I'm curious to hear what @johnharris85 thinks

lahabana:
@lobkovilya also missing this and `_rules` API

docs/madr/decisions/038-transition-to-new-policies.md
lobkovilya: yeah, my PoC had this label and I forgot to remove it 😂

lobkovilya:
```suggestion
  kuma.io/effect: shadow
```

docs/madr/decisions/038-transition-to-new-policies.md
lobkovilya: It's pretty bad even with TCP rbac, but with HTTP the diff looks even worse:
```yaml
- value:
    - name: envoy.filters.network.http_connection_manager
      typedConfig:
        '@type': type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
        commonHttpProtocolOptions:
          idleTimeout: 0s
        forwardClientCertDetails: SANITIZE_SET
        httpFilters:
          - name: envoy.filters.http.rbac
            typedConfig:
              '@type': type.googleapis.com/envoy.extensions.filters.http.rbac.v3.RBAC
              rules:
                policies:
                  MeshTrafficPermission:
                    permissions:
                      - any: true
                    principals:
                      - any: true
          - name: envoy.filters.http.router
            typedConfig:
              '@type': type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
        requestHeadersTimeout: 0s
        routeConfig:
          name: inbound:frontend_kuma-demo_svc_8080
          requestHeadersToRemove:
            - x-kuma-tags
          validateClusters: false
          virtualHosts:
            - domains:
                - '*'
              name: frontend_kuma-demo_svc_8080
              routes:
                - match:
                    prefix: /
                  route:
                    cluster: localhost:8080
                    idleTimeout: 3600s
                    timeout: 0s
        setCurrentClientCertDetails:
          uri: true
        statPrefix: localhost_8080
        streamIdleTimeout: 3600s
  op: replace
  path: /type.googleapis.com~1envoy.config.listener.v3.Listener/inbound:10.42.0.16:8080/filterChains/0/filters
```
I'll add this to the doc

docs/madr/decisions/038-transition-to-new-policies.md
lahabana: I want proper testing here. Because I'm not 100% sure the composition works this way:

However is the output a+b(shadow)+c(shadow) useful to figure what a+b+c(shadow) does?
May it's not an actual problem.

What I mean is that I'd like to see more advanced experiments than just using 2 policies.

docs/madr/decisions/038-transition-to-new-policies.md
```diff
@@ -0,0 +1,473 @@
...
+
+```shell
+$ curl :5681/meshes/defaults/dataplane/dpp-1/xds?shadow=true # show potential changes for a dpp-1 in the "default" mesh
+```
+
+Implementation wise, shadow mode requires generating 2 snapshots: one with and without the new policy. 
+By comparing these snapshots we can generate a list of jsonPatches that would be applied to the Envoy configuration. 
+
+Depending on how much time do we want to spend on the implementation, we can choose to implement shadow mode for Zone CP or Global CP.
+Implementing shadow mode for Global CP requires either generating snapshots on Global CPs or forwarding the request to Zone CPs.
```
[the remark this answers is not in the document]
lahabana: The user will need to do this for each of their DPP so the zone part is the least of their problem no? :)

I think for sure the UX from the global can be improved. However, I think any operation that is O(number of dpp across all zones) is a no go.

docs/madr/decisions/038-transition-to-new-policies.md
```diff
@@ -0,0 +1,473 @@
...
+            exact: spiffe://default/frontend_kuma-demo_svc_8080
+  op: add
+  path: envoy.config.listener.v3.Listener/inbound:10.42.0.28:6379/filterChains/0/filters/0/typedConfig/rules/policies/MeshTrafficPermission
+- op: remove
+  path: envoy.config.listener.v3.Listener/inbound:10.42.0.28:6379/filterChains/0/filters/0/typedConfig/rules/policies/backend-to-httpbin-4b4zbcc4422zz784
+```
+
+## TrafficRoute -> MeshHTTPRoute(MeshTCPRoute)
+
+Covered by [036-route-transition.md](036-route-transition.md)
```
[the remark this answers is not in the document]
lobkovilya: The link works if clicking not from a "rich diff" but from a "rich file" https://github.com/kumahq/kuma/blob/fd1e38de7197dc53a875c5a62cfb199d872f95f6/docs/madr/decisions/038-transition-to-new-policies.md

docs/madr/decisions/038-transition-to-new-policies.md
```diff
@@ -0,0 +1,473 @@
...
+spec:
+  targetRef:
+    kind: Mesh
+  to: [...]
+```
+
+it doesn't apply on the real proxies. Instead, by using a special endpoint we can get a list of jsonPatches that would be applied to the Envoy configuration
+
+```shell
+$ curl :5681/meshes/defaults/dataplane/dpp-1/xds?shadow=true # show potential changes for a dpp-1 in the "default" mesh
```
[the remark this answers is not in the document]
lobkovilya: added an openapi schema for `_shadow` endpoint

docs/madr/decisions/038-transition-to-new-policies.md
```diff
@@ -0,0 +1,473 @@
...
+      kind: Mesh
+    default:
+      connectionTimeout: 10s
+      idleTimeout: 2h
+      http:
+        requestTimeout: 0s
+        streamIdleTimeout: 1h
+---
+type: MeshTimeout
+name: aaa-timeout-to-backend
```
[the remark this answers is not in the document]
lobkovilya: Same reason as https://github.com/kumahq/kuma/issues/9151, both `timeout-global` and `aaa-timeout-to-backend` have top-level targetRef `Mesh`, but we want the second one to be more specific

lobkovilya:
> also missing this and `_rules` API

I updated the MADR, for `_rules` we're going to introduce only `effect=shadow` query param that returns full rules response as if shadow policies are applied for real (it **doesn't** return jsonPatches diff).

Since `/_rules` is rarely used without GUI or kubectl I think it's okay to not introduce `_rules-diff` endpoint. At least now, while we don't know how we're going to present it in GUI.

docs/madr/decisions/038-transition-to-new-policies.md
lobkovilya: added a note under the `current state` section

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,642 @@
...
+  targetRef:
+    kind: Mesh
+  to: [...]
+```
+
+it doesn't apply on the real proxies. Instead, by using Inspect API we can inspect what config dump would look like if the shadow policies was applied.
+
+### Changes in the Inspect API
+
+Introduce a new non-physical endpoint `/_snapshot` that returns a config dump of a DPP without the actual `/config_dump` call to Envoy.
```
jakubdyszkiewicz, line 53: what is non-physical?

lobkovilya: I was using the terminology from Charly's [comment](https://github.com/kumahq/kuma/pull/9571#discussion_r1521620338):
> Non physical endpoints are usually prefixed by _ like _rules or _resources.

IIUC it's non-physical when there is no such resource type

jakubdyszkiewicz (approved):

docs/madr/decisions/038-transition-to-new-policies.md
slonka: > However is the output a+b(shadow)+c(shadow) useful to figure what a+b+c(shadow) does?
> May it's not an actual problem.

There is no composition for s&d policies right? I might be getting this wrong but what I would image happening is:
1. You have s&d policies (in normal mode because there is no other mode)
2. You add a couple of targetRef policies in shadow mode (might be even doing doing some composition)
3. We generate normal snapshot for s&d policies and another snapshot that has both s&d policies and targetRef policies and we call this snapshot "shadow snapshot"
4. We show the diff to the user, if they like the diff they can remove "shadow" from all policies that have that 
5. We repeat 2-4 until everything is migrated
6. Most important thing and a bit of leap of faith, users remove the old s&d policies. This **heavily** relies on our assumption that targetRef policies take precedence over s&d policies. If this is broken anywhere then this step might mess some things up.

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -44,12 +50,121 @@ spec:
...
+
+We don't want to introduce `/_rules-diff` at this point, as `/_rules` API is rarely used without the client and heavily depends on the visualization in GUI.
+
+OpenAPI spec for new endpoints:
+
+```yaml
+openapi: 3.0.3
+info:
+  title: Shadow Policies Inspector
+  description: API for getting a diff in JSON Patch format as if shadow policies were applied
```
slonka, line 79: the diff was supposed to be under `_snapshot-diff` right? not under `_snapshot`?

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -451,23 +644,3 @@ spec:
...
-
-TBD
-
-## TrafficLog -> MeshTrafficLog
-
-TBD
-
-## TrafficTrace -> MeshTrafficTrace
-
-TBD
```
slonka: Are these not interesting? Why are we removing them?

slonka (commented):
Done another round, sorry it took some time. I think that the spec is wrong or I got something wrong there and I'm curious to know why we don't want examples for the couple of policies mentioned in the last comment.

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,646 @@
...
+it doesn't apply on the real proxies. Instead, by using Inspect API we can inspect what config dump would look like if the shadow policies was applied.
+
+### Changes in the Inspect API
+
+Introduce a new non-physical endpoint `/_snapshot` that returns a config dump of a DPP without the actual `/config_dump` call to Envoy.
+If we want to be able to compute diff on the client side we can't use `xds` endpoint as its response lightly diverges from what CP has in SnapshotCache.
+
+Add a query parameter `effect=shadow`:
+
+* `GET /meshes/{mesh}/dataplanes/{name}/_snapshot` -> `GET /meshes/{mesh}/dataplanes/{name}/_snapshot?effect=shadow`
```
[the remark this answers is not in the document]
lobkovilya: Yes, it's a config dump but from the CP's point of view (not Envoy's).

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,646 @@
...
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successfully retrieved snapshot.
+          content:
+            application/json:
+              schema:
+                type: object
+                description: The value to be used within the operations.
```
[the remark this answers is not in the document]
lobkovilya: Ah I accidentally copied it from the `_snapshot-diff` endpoint

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,646 @@
...
+        streamIdleTimeout: 3600s
+  op: replace
+  path: /type.googleapis.com~1envoy.config.listener.v3.Listener/inbound:10.42.0.16:8080/filterChains/0/filters
+```
+
+## TrafficRoute -> MeshHTTPRoute(MeshTCPRoute)
+
+Covered by [036-route-transition.md](036-route-transition.md)
+
+## ProxyTemplate -> MeshProxyTemplate
```
lahabana, line 615:
```suggestion
## ProxyTemplate -> MeshProxyPatch
```

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -44,12 +50,121 @@ spec:
...
+  title: Shadow Policies Inspector
+  description: API for getting a diff in JSON Patch format as if shadow policies were applied
+  version: 1.0.0
+paths:
+  /meshes/{mesh}/dataplanes/{name}/_snapshot:
+    get:
+      summary: Get a DPP snapshot on a CP
+      description: Returns a DPP snapshot computed by the CP without the actual Envoy call.
+      parameters:
+        - in: path
```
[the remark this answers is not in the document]
lobkovilya: Added `shadow=true` to the spec

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: right, I updated the description

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -44,12 +50,121 @@ spec:
...
+                      type: string
+                      description: A JSON Pointer path indicating the part of the document to operate on.
+                    value:
+                      type: object
+                      description: The value to be used within the operations.
+        '400':
+          description: Bad request. Parameters were invalid.
+        '404':
+          description: The specified mesh/name combination was not found.
+```
```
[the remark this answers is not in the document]
lobkovilya: Right, I added the line:
> Endpoint `/_rules` is getting the same `shadow` and `include` query parameters.

So I guess we don't have to re-define the whole OpenAPI spec for `_rules`

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: TBH it takes time to produce meaningful examples so I thought the existing examples were sufficient enough. But maybe I can do a few more, WDYT?

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,646 @@
...
+Introduce a new non-physical endpoint `/_snapshot` that returns a config dump of a DPP without the actual `/config_dump` call to Envoy.
+If we want to be able to compute diff on the client side we can't use `xds` endpoint as its response lightly diverges from what CP has in SnapshotCache.
+
+Add a query parameter `effect=shadow`:
+
+* `GET /meshes/{mesh}/dataplanes/{name}/_snapshot` -> `GET /meshes/{mesh}/dataplanes/{name}/_snapshot?effect=shadow`
+* `GET /meshes/{mesh}/dataplanes/{name}/_rules` -> `GET /meshes/{mesh}/dataplanes/{name}/_rules?effect=shadow`
+
+when `effect=shadow` is set, the response will take policies with `kuma.io/effect: shadow` label into account.
+Please note, the response DOES NOT yet contain the json patch diff. These endpoints are only useful when calculating the diff on the client side.
```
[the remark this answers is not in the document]
lobkovilya: Removed the whole sentence

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,646 @@
...
+          required: true
+          description: The mesh of the DPP to get the diff for.
+          schema:
+            type: string
+        - in: path
+          name: name
+          required: true
+          description: The name of the DPP within the mesh to get the diff for.
+          schema:
+            type: string
```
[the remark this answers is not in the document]
lobkovilya: updated the spec with all relevant params

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,646 @@
...
+
+Add a query parameter `effect=shadow`:
+
+* `GET /meshes/{mesh}/dataplanes/{name}/_snapshot` -> `GET /meshes/{mesh}/dataplanes/{name}/_snapshot?effect=shadow`
+* `GET /meshes/{mesh}/dataplanes/{name}/_rules` -> `GET /meshes/{mesh}/dataplanes/{name}/_rules?effect=shadow`
+
+when `effect=shadow` is set, the response will take policies with `kuma.io/effect: shadow` label into account.
+Please note, the response DOES NOT yet contain the json patch diff. These endpoints are only useful when calculating the diff on the client side.
+
+Introduce a new non-physical endpoint `/_snapshot-diff`. This endpoint will return a list of changes in a JSON Patch format
```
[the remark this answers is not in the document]
lobkovilya: Nice, I really like the idea, updated the spec

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,646 @@
...
+
+### Changes in the Inspect API
+
+Introduce a new non-physical endpoint `/_snapshot` that returns a config dump of a DPP without the actual `/config_dump` call to Envoy.
+If we want to be able to compute diff on the client side we can't use `xds` endpoint as its response lightly diverges from what CP has in SnapshotCache.
+
+Add a query parameter `effect=shadow`:
+
+* `GET /meshes/{mesh}/dataplanes/{name}/_snapshot` -> `GET /meshes/{mesh}/dataplanes/{name}/_snapshot?effect=shadow`
+* `GET /meshes/{mesh}/dataplanes/{name}/_rules` -> `GET /meshes/{mesh}/dataplanes/{name}/_rules?effect=shadow`
```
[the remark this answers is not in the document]
lobkovilya: it's `shadow=true` now

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,646 @@
...
+## Shadow mode for policies
+
+Shadow mode is a way to test new policies without affecting the real traffic. When the policy is created with `kuma.io/effect: shadow` label
+
+```yaml
+type: MeshTimeout
+name: timeout-global
+mesh: default
+labels:
+  kuma.io/effect: shadow
```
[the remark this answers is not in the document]
lobkovilya: discussed offline, `kuma.io/ignore` and `kuma.io/mode` don't work well for this case, so keeping the `kuma.io/effect: shadow`?

docs/madr/decisions/038-transition-to-new-policies.md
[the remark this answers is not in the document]
lahabana: Ok clearly this indicates that there's work to be done on this API

docs/madr/decisions/039-transition-to-new-policies.md
lahabana: yeah "non physical" would be this. It's a very crap naming though :)

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,644 @@
...
+    kind: Mesh
+  to: [...]
+```
+
+it doesn't apply on the real proxies. Instead, by using Inspect API we can inspect what config dump would look like if the shadow policies was applied.
+
+### Changes in the Inspect API
+
+Introduce a new non-physical endpoint `/_snapshot` that returns a config dump of a DPP without the actual `/config_dump` call to Envoy.
+If we want to be able to compute diff on the client side we can't use `xds` endpoint as its response slightly diverges from what CP has in the SnapshotCache.
```
lahabana, line 58: How does it differ?

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,644 @@
...
+
+* `GET /meshes/{mesh}/dataplanes/{name}/_snapshot` -> `GET /meshes/{mesh}/dataplanes/{name}/_snapshot?shadow=true`
+* `GET /meshes/{mesh}/dataplanes/{name}/_rules` -> `GET /meshes/{mesh}/dataplanes/{name}/_rules?shadow=true`
+
+when `shadow=true` is set, the response will take policies with `kuma.io/effect: shadow` label into account.
+
+For the debugging purposes it makes sense to represent the shadow snapshot as the original snapshot + jsonPatch that can be applied to it.
+Kuma GUI and kumactl also can benefit from such representation. To make server produce such response we can introduce a query parameter `include=diff`.
+Query param `include` contains a list of fields that should be included in the response. By default, only `xds` field is set by the API server. 
+From the validation point of view the request `_snapshot?shadow=true&include=diff` makes no sense and should be rejected.
```
lahabana, line 70: It's the other way round no?

1. `_snapshot` or `_snapshot?shadow=false` : Returns the snaphost as is (no shadow policies)
2. `_snapshot?shadow=true` : Returns the snapshot as if shadow policies were enabled
3. `_snapshot?shadow=true&include=diff`: same as 2 but also returns the diff between this and what 1 would return
4. (not here but in the future) `_snapshot?proxy=true`: Returns the xds config from the proxy (equivalent to the current `xds` endpoint)
5. (not here but in the future) `_snapshot?proxy=true&include=diff`: 4 but also returns the diff between the proxy xds config and the snapshot of the CP (what 1 would return)
6. (not here but in the future) `_snapshot?proxy=true&shadow=true` ??? Is this even possible?

In light of 4,5,6 should we just have `_snapshot?mode=proxy|shadow` ? 
WDYT?

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,644 @@
...
+openapi: 3.0.3
+info:
+  title: Shadow Policies Inspector
+  description: API for inspecting policies' effect on the Data Plane Proxy configurations
+  version: 1.0.0
+paths:
+  /meshes/{mesh}/dataplanes/{name}/_snapshot:
+    get:
+      summary: Get a DPP snapshot on a CP
+      description: Returns a DPP snapshot computed by the CP without the actual Envoy call.
```
lahabana, line 84: That's not very clear.

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,644 @@
...
+OpenAPI spec for new endpoints:
+
+```yaml
+openapi: 3.0.3
+info:
+  title: Shadow Policies Inspector
+  description: API for inspecting policies' effect on the Data Plane Proxy configurations
+  version: 1.0.0
+paths:
+  /meshes/{mesh}/dataplanes/{name}/_snapshot:
```
lahabana, line 81: Is `_snapshot` just to generic? Should this just be `_config` ? Istio uses `proxy-config` https://istio.io/latest/docs/ops/diagnostic-tools/proxy-cmd/

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,644 @@
...
+          name: shadow
+          description: |
+            When computing XDS config the CP takes into account policies with 'kuma.io/effect: shadow' label
+          schema:
+            type: boolean
+            default: false
+        - in: query
+          name: include
+          description: |
+            An array of field names that also should be included in the response. By default, only 'xds' field is set by the API server.
```
lahabana, line 108: It's not exactly this right? What this param does is potentially do extra stuff. For example here it computes the diff and adds it in the response.

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,644 @@
...
+                type: object
+                required: [xds]
+                properties:
+                  xds:
+                    description: The raw XDS config as an inline JSON object
+                    type: object
+                  diff:
+                    description: |
+                      When '?include=diff' is specified, the field is set to an array of json patches between the current XDS config
+                      and XDS config that takes into account policies with 'kuma.io/effect: shadow'.
```
lahabana, line 129: This doesn't actually say what the output is. This is a list of json patches so we should likely link to the spec of json patch for example.

Put yourself in the user's shoes, what do you think they'd want to know?

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,644 @@
...
+                        value:
+                          type: object
+                          description: The value to be used within the operations.
+        '400':
+          description: Bad request. Parameters were invalid.
+        '404':
+          description: The specified mesh/name combination was not found.
+```
+
+Endpoint `/_rules` is getting the same `shadow` and `include` query parameters.
```
lahabana, line 153: So there's a diff extra field? Where is it because that's the actual response.

https://github.com/kumahq/kuma/blob/9d93621900c57c3569a53924f0ec836c88eba559/api/openapi/specs/api.yaml#L148-L163

docs/madr/decisions/039-transition-to-new-policies.md
```diff
@@ -0,0 +1,644 @@
...
+kumactl inspect dataplane dpp-1 --config-dump --shadow > with-shadow.json && jd <(kumactl inspect dataplane dpp-1 --config-dump) with-shadow.json
+```
+
+### Implementation
+
+Implementation wise, shadow mode requires generating 2 snapshots: one with and without the new policy. 
+By comparing these snapshots we can generate a list of jsonPatches that would be applied to the Envoy configuration. 
+
+Depending on how much time do we want to spend on the implementation, we can choose to implement shadow mode for Zone CP or Global CP.
+Implementing shadow mode for Global CP requires either generating snapshots on Global CPs or forwarding the request to Zone CPs.
```
lahabana, line 170: So we said this wasn't an option right? Should this be persisted in the MADR?

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: I called it "snapshot" because CP operates with Snapshots and stores them in the SnapshotCache. This is contrasted to the config dump (or proxy-config) that is supposed to be taken from the proxy (not from CP). 

But if we want to converge `_snapshot` and `xds` in the future and introduce something like `mode=proxy|shadow` then `_config` is a better name

docs/madr/decisions/038-transition-to-new-policies.md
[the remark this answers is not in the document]
lobkovilya: Yeah, and I think we're planning to introduce `weight` to the targetRef no?

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: The description is not wrong, it's just missing the fact that "before including in the response the field has to be computed". If you think it's important for users to know I surely can add it

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: But we have a json patch spec right here https://github.com/kumahq/kuma/blob/7d5702d38fae50ac72021d0225dbd77593a45274/docs/madr/decisions/039-transition-to-new-policies.md?plain=1#L130-L146

I can add a link to the RFC https://datatracker.ietf.org/doc/html/rfc6902/

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: Discussed [here](https://github.com/kumahq/kuma/pull/9571#discussion_r1542645970) that probably we don't need to re-define `_rules` spec in this MADR. But if it's not clear I can add it

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: What about this
```suggestion
      description: |
        Returns a proxy xds config with or without taking into account policies with 'kuma.io/effect: shadow' label.
```

lahabana: How about:

> Returns the configuration of the proxy ([xds](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol) configuration).

The fact that there are options is documented on the option not in the generic description of the endpoint.

docs/madr/decisions/039-transition-to-new-policies.md
lahabana: Just where do you put the `diff` field?

docs/madr/decisions/039-transition-to-new-policies.md
lahabana: Only thing that I was saying is that your doc doesn't say it's a JSONPatch sure the spec is here but folks have to parse/understand it whereas if they just see "it's a JSONPatch" they get it quicker.

docs/madr/decisions/039-transition-to-new-policies.md
lahabana: Hmm isn't this what I say:

> 3. _snapshot?shadow=true&include=diff: same as 2 but also returns the diff between this and what 1 would return

Of course extra query_params doesn't change the rest of the response.

lobkovilya: I see, I got a wrong understanding of 3 initially, so `_snapshot?shadow=true&include=diff` returns
```yaml
xds: {...} # 'shadow' config
diff: [...] # jsonPatch that is applicable to 'xds' so you could get a 'current' config
```

But I guess if we call just `_snapshot?include=diff` then it returns:
```yaml
xds: {...} # 'current' config
diff: [...] # jsonPatch that is applicable to 'xds' so you could get a 'shadow' config
```

So in the first case `diff` is `shadow -> current` and in another `diff` is `current -> shadow`.

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: Thanks, updated

docs/madr/decisions/039-transition-to-new-policies.md
slonka: we can always add it afterwards, it's fine now

docs/madr/decisions/039-transition-to-new-policies.md
lahabana: Hmm to me `_snapshot?shadow=false&include=diff` would just return an empty diff as it compares current config with current config. I think it's just less suprising that the base of the diff remains unchanged (the CP snapshot as it's sent to the proxy)

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: Right, fixed it

docs/madr/decisions/039-transition-to-new-policies.md
lobkovilya: Right, I mention adding it at the root level of `InspectRules` object

docs/madr/decisions/039-transition-to-new-policies.md
lahabana: I'd go for _config because I find _snapshot a lot more confusing (makes me think about DB snapshot as well).

lahabana (approved):

slonka (approved):
LGTM

PR #9510, opened 2024-03-06
feat(MeshService): add first iteration of resource

michaelbeaumont:
- [ ] helm chart changes

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues --
- [x] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [x] Tests (Unit test, E2E tests, manual test on universal and k8s) --
  - Don't forget `ci/` labels to run additional/fewer tests
- [x] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [x] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) --

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #9583, opened 2024-03-12
feat(MeshService): xds generation

michaelbeaumont:
- [x] Depends on #9510 

### Checklist prior to review


pkg/plugins/policies/core/xds/meshroute/listeners.go
```diff
@@ -55,9 +56,59 @@ func MakeHTTPSplit(
...
+		}
+		for _, port := range svc.Spec.Ports {
+			dests = append(dests, DestinationService{
+				Outbound: mesh_proto.OutboundInterface{
+					DataplaneIP:   svc.Spec.Status.VIPs[0].IP,
+					DataplanePort: port.Port,
+				},
+				Port:        pointer.To(port.Port),
+				Protocol:    port.Protocol,
+				ServiceName: fmt.Sprintf("%s_svc_%d", svc.GetMeta().GetName(), port.Port),
```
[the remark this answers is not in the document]
michaelbeaumont: Perhaps... it wasn't immediately clear to me how to get this from the core `MeshService` object, which doesn't have a namespace.

pkg/plugins/policies/core/xds/meshroute/listeners.go
```diff
@@ -58,7 +58,7 @@ func MakeHTTPSplit(
 
 type DestinationService struct {
 	Outbound    mesh_proto.OutboundInterface
-	Port        *uint32
+	Port        *uint32 // todo(jakubdyszkiewicz): move to TargetRef when we introduce port in TargetRef
```
michaelbeaumont, line 61: This might just be a clumsy structuring of this intermediate type on my part. I think we don't necessarily need TargetRef with a port.
I think the `TargetRef` in this struct should be a `BackendRef` instead, whose weight we ignore, similar to how we handle mirroring in MeshHTTPRoute. And we put the port there in the "real `MeshService`" case.

jakubdyszkiewicz:
This is just the first step. I manually verified that it works. I disabled VIP controller on Kubernetes and manually applied MeshService. We'll start writing E2E tests when we'll have autogeneration from Kube Services.

jakubdyszkiewicz (approved):

pkg/xds/context/mesh_context_builder.go
```diff
@@ -156,6 +157,11 @@ func (m *meshContextBuilder) BuildIfChanged(ctx context.Context, meshName string
 	for _, dp := range dataplanes {
 		dataplanesByName[dp.Meta.GetName()] = dp
 	}
+	meshServices := resources.MeshServices().Items
+	meshServicesByName := make(map[string]*v1alpha1.MeshServiceResource, len(dataplanes))
```
lahabana, line 161: This len seems weird no?

Issue #8963, opened 2024-01-22
pass proper context to OpenTelemetry exporter


Automaat:
### Description

https://github.com/kumahq/kuma/pull/8893#discussion_r1459097662 related to review discussion

PR #9606, opened 2024-03-14
fix(kuma-dp): pass proper context in meshmetric component

Automaat:
Fixes: #8963

### Checklist prior to review

### DIFF
--- a/app/kuma-dp/pkg/dataplane/meshmetrics/component.go
+++ b/app/kuma-dp/pkg/dataplane/meshmetrics/component.go
@@ -68,6 +68,8 @@ func (cf *ConfigFetcher) Start(stop <-chan struct{}) error {
 		"socketPath", fmt.Sprintf("unix://%s", cf.socketPath),
 	)
 
+	ctx := context.Background()
+
 	for {
 		select {
 		case <-cf.ticker.C:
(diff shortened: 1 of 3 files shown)


app/kuma-dp/pkg/dataplane/meshmetrics/component.go
```diff
@@ -68,6 +68,8 @@ func (cf *ConfigFetcher) Start(stop <-chan struct{}) error {
 		"socketPath", fmt.Sprintf("unix://%s", cf.socketPath),
 	)
 
+	ctx := context.Background()
```
michaelbeaumont, line 71: Should this be canceled when `stop` is closed?

Automaat: you are right, should be good now

michaelbeaumont: @Automaat I think it needs to be closed in a separate go routine, if we're executing `cf.configureOpenTelemetryExporter`, we can't react to the channel closing in `case <-stop:` I don't think

app/kuma-dp/pkg/dataplane/meshmetrics/component.go
```diff
@@ -285,6 +296,17 @@ func (cf *ConfigFetcher) reconfigureBackendIfNeeded(backendName string, backend
 	return nil
 }
 
+func (cf *ConfigFetcher) shutDownMetricsExporters() {
+	ctx, cancel := context.WithTimeout(context.Background(), cf.drainTime)
+	defer cancel()
+	for _, exporter := range cf.runningBackends {
+		err := exporter.exporter.Shutdown(ctx)
```
michaelbeaumont, line 303: I _suppose_ this could be done in parallel and with waitgroup to increase the chances that all are shutdown? But I don't think it's likely to matter in practice.

app/kuma-dp/pkg/dataplane/meshmetrics/component.go
```diff
@@ -68,6 +70,14 @@ func (cf *ConfigFetcher) Start(stop <-chan struct{}) error {
 		"socketPath", fmt.Sprintf("unix://%s", cf.socketPath),
 	)
 
+	ctx, ctxCancel := context.WithCancel(context.Background())
+	go func() {
+		select {
+		case <-stop:
+			ctxCancel()
+		}
```
michaelbeaumont, line 78: Nit, feel free to ignore but

```suggestion
		_ <-stop
		ctxCancel()
```

michaelbeaumont (approved):

app/kuma-dp/pkg/dataplane/meshmetrics/component.go
Automaat: I think that users will very rarely use more than one exporter

app/kuma-dp/pkg/dataplane/meshmetrics/component.go
Automaat: changed

PR #9579, opened 2024-03-12
feat(meshmetric): profiles API

slonka:
schema from MADR: https://github.com/kumahq/kuma/pull/9269/files#diff-eb2f6c70e315748d491c7b85736fd012c0bb46738a1f39631c0341b1c0ac28a6R306-R319

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- xrel https://github.com/kumahq/kuma/issues/8845
- [X] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [X] Tests (Unit test, E2E tests, manual test on universal and k8s) --
  - Don't forget `ci/` labels to run additional/fewer tests
- [X] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [X] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) --

> Changelog: feat(meshmetric) profiles

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #9624, opened 2024-03-15
feat(meshmetrics): profiles implementation

slonka:
closes https://github.com/kumahq/kuma/issues/8845

### Checklist prior to review


app/kuma-dp/pkg/dataplane/metrics/metrics_mutator.go
```diff
@@ -0,0 +1,61 @@
...
+	"io"
+
+	io_prometheus_client "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+)
+
+type (
+	MetricsMutator    func(in io.Reader, out io.Writer) error
+	PrometheusMutator func(in map[string]*io_prometheus_client.MetricFamily) error
+	OtelMutator       func(in io.Reader) (map[string]*io_prometheus_client.MetricFamily, error)
```
Automaat, line 13: we could rename this type from `OtelMutator` to `MeshMetricMutator`

app/kuma-dp/pkg/dataplane/metrics/profiles.go
```diff
@@ -0,0 +1,236 @@
...
+	selectorToFilterFunction(v1alpha1.Selector{
+		Type:  v1alpha1.ExactSelectorType,
+		Match: "envoy_cluster_membership_total",
+	}),
+	// end of dashboards
+}
+
+func ProfileMutatorGenerator(sidecar *v1alpha1.Sidecar) PrometheusMutator {
+	effectiveSelectors := []selectorFunction{alwaysSelect} // default is All
+	if sidecar != nil && sidecar.Profiles != nil && sidecar.Profiles.AppendProfiles != nil && len(*sidecar.Profiles.AppendProfiles) == 1 {
```
Automaat, line 157: if I understand it correctly we will only use 1st profile? Don't we want to support setting multiple profiless?

slonka: If I remember correctly since we only have 3 profiles now we said that we will only allow 1 profile and allow multiple profiles in the future, that's why this validation is here https://github.com/kumahq/kuma/pull/9579/files#diff-e28548cca6e7c15cb7915892c16c7b8a4916a29ab0004f64c7a4f3c1da18b4c5R44

Automaat: right, with current set of profiles we don't need this

Automaat (approved):

PR #9632, opened 2024-03-15
fix(policy): fix otel backend naming and add proper error handling wh…

Automaat:
…en closing backend

### Checklist prior to review

### DIFF
--- a/pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
+++ b/pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
@@ -290,5 +290,6 @@ func filterPrometheusBackends(backends *[]api.Backend) []*api.PrometheusBackend 
 }
 
 func backendNameFrom(endpoint string) string {
-	return strings.ReplaceAll(endpoint, ":", "")
+	// we need to remove "/" as this name will be used as directory name
+	return strings.ReplaceAll(endpoint, "/", "")
 }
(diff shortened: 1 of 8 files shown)


pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -290,5 +290,6 @@ func filterPrometheusBackends(backends *[]api.Backend) []*api.PrometheusBackend
 }
 
 func backendNameFrom(endpoint string) string {
-	return strings.ReplaceAll(endpoint, ":", "")
+	// we need to remove "/" as this name will be used as directory name
```
slonka, line 293: Should we just allow [POSIX](https://en.wikipedia.org/wiki/POSIX) "Fully portable filenames" `A–Z a–z 0–9 . _ -` from - https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations

slonka (approved):
one Q

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
Automaat: this is partially fixed by [updated validation](https://github.com/kumahq/kuma/pull/9634). I am now replacing "/" and ":", so it should now work with valid url without schema

PR #9653, opened 2024-03-19
feat(MeshService): backend ref outbound to mesh service on Dataplane

jakubdyszkiewicz:
### Checklist prior to review

Adds a way to address MeshService using outbound (non-transparent proxy). I also moved generating ServiceDestination based on MeshService from httproute to build implicit outbound (just like we do for VIPs). This makes the code consistent.


api/mesh/v1alpha1/dataplane.proto
```diff
@@ -183,8 +183,22 @@ message Dataplane {
       // specific data plane proxy of a service (for example: `version=v2`) the
       // better practice is to use TrafficRoute.
       map<string, string> tags = 5;
+
+      // BackendRef is a way to target MeshService.
+      // Experimental. Do not use on production yet.
```
lahabana, line 188: Should mention that if both tags and backendRef is set. backendRef takes over.

api/mesh/v1alpha1/dataplane_helpers.go
```diff
@@ -127,6 +127,26 @@ func (i OutboundInterface) String() string {
 		strconv.FormatUint(uint64(i.DataplanePort), 10))
 }
 
+func NonBackendRefFilter(outbound *Dataplane_Networking_Outbound) bool {
```
michaelbeaumont, line 130: Any plans for this API that we don't just have `GetNonBackendRefOutbounds`, `GetAllOutbounds` instead? Fine either way I suppose

michaelbeaumont (approved):

api/mesh/v1alpha1/dataplane.proto
jakubdyszkiewicz: that should not be allowed. Added validation

api/mesh/v1alpha1/dataplane_helpers.go
jakubdyszkiewicz: I feel like we may need to filter out external services outbound in the future. Maybe I'm wrong. I'll leave it as is for this PR.

bartsmykla (approved):

PR #9670, opened 2024-03-20
fix(GAMMA/MeshHTTPRoute): allow "kuma.io/unresolved-backend" service name

bartsmykla:
Normally, service names cannot contain forward slashes ("/"). However, there are exceptions during resource conversion:
* Gateway API to Kuma MeshHTTPRoute Conversion When converting an HTTPRoute from Gateway API to a MeshHTTPRoute (Kuma's resource definition), there might be situations where the targeted backend reference cannot be found. In such cases, Kuma sets the service name to "kuma.io/unresolved-backend". This name includes a forward slash, but it's allowed as an exception to handle unresolved references.

### Checklist prior to review

  - No relevant issues
  - It won't
  - Manually tested
  - There is no need
  - There is no need


lahabana:
I think this value seems weird. Weakening validation seems pretty risky to me... what's @michaelbeaumont 's opinion?

bartsmykla:
@lahabana that's how it currently work for `MeshGatewayRoute`'s (re. https://github.com/kumahq/kuma/commit/97552c1d07148434ac1b80bd85d048237098742f) and according to code, that's how it should work for GAMMA, but it doesn't because of validation is not allowing to convert `HTTPRoute`s to `MeshHTTPRoute`s

michaelbeaumont:
Yeah, the option as Bart pointed out would work. I wouldn't look at it as weakening validation per se, it's just to allow us to use a reserved string, same as we do for keys. Is there another way to prevent a collision with a real service?

michaelbeaumont (approved):

PR #9675, opened 2024-03-20
feat(pgx): configure idle timeout

lukidzi:
### Checklist prior to review

- add environment variable `KUMA_STORE_POSTGRES_MAX_CONNECTION_IDLE_TIME` as duration to configure pgx connection idle time
- default connection idle time equal 30m, the same as in pgx

### DIFF
--- a/pkg/config/plugins/resources/postgres/config.go
+++ b/pkg/config/plugins/resources/postgres/config.go
@@ -15,6 +15,7 @@ import (
 const (
 	DriverNamePgx             = "pgx"
 	DriverNamePq              = "postgres"
+	DefaultConnectionTimeout  = 5
 	DefaultMaxIdleConnections = 50
 	DefaultMinOpenConnections = 0
 )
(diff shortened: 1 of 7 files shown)


pkg/config/plugins/resources/postgres/config.go
```diff
@@ -50,7 +52,10 @@ type PostgresStoreConfig struct {
 	// Driver to use, one of: pgx, postgres
 	DriverName string `json:"driverName" envconfig:"kuma_store_postgres_driver_name"`
 	// Connection Timeout to the DB in seconds
+	// Deprecated: it's used for MaxIdleConnectionTime (applied only when driverName=pgx), use `MaxConnectionIdleTime` instead.
```
Automaat, line 55: isn't this connection timeout used in: https://github.com/kumahq/kuma/blob/master/pkg/config/plugins/resources/postgres/config.go#L105 which is used for default postgres driver?

Automaat (approved):

PR #9676, opened 2024-03-20
feat(resources): add status

jakubdyszkiewicz:
### Checklist prior to review

This PR adds a possibility to have a `status` field next to `spec`. It only works with new resources (`Mesh*` prefix).

* Rename `GetStatus()` helper to `Status()` for overview resources to avoid clash.
* It's not clear if we want users to be able to modify status or should we ignore this. Currently I left it so that it's possible to modify it.
* Add `status` field to `MeshService` as an example of policy with status field. I did not remove `status` in `spec` yet to avoid unnecessary changes in this PR.


pkg/plugins/resources/k8s/caching_converter.go
```diff
@@ -61,6 +50,19 @@ func (c *cachingConverter) ToCoreResource(obj k8s_model.KubernetesObject, out co
...
+		status, err := obj.GetStatus()
+		if err != nil {
+			return err
+		}
+		if err := out.SetStatus(status); err != nil {
+			return err
+		}
+	}
+	if obj.GetResourceVersion() != "" {
+		// an absent of the ResourceVersion means we decode 'obj' from webhook request,
```
lukidzi, line 63:
```suggestion
		// an absence of the ResourceVersion means we decode 'obj' from webhook request,
```

lukidzi (approved):

lahabana (commented):
That's good to see this appear. We should be especially careful that IIRC k8s status are in practice a separate resource in etcd where it looks here in universal we just make it a different row. I'm not sure what the implications are but we should consider this

pkg/api-server/resource_endpoints.go
```diff
@@ -379,25 +382,27 @@ func (r *resourceEndpoints) createResource(
...
 		r.descriptor,
 		user.FromCtx(ctx),
 	); err != nil {
 		rest_errors.HandleError(ctx, response, err, "Access Denied")
 		return
 	}
 
-	_ = currentRes.SetSpec(newSpec)
+	_ = currentRes.SetSpec(newResRest.GetSpec())
+	if r.descriptor.HasStatus { // todo(jakubdyszkiewicz) should we always override this?
```
lobkovilya, line 401: All resources on Universal already have `status`. Is it going to simplify the code if we generate `Status` field for all resources and drop the `HasStatus` logic altogether?

As long as we're using a pointer and `omitempty`, the `Status` field won't be visible during marshaling. Also, it'll be always safe to call `SetStatus`.

jakubdyszkiewicz: > All resources on Universal already have status

What do you mean?

I was debating whether to implement noop SetStatus instead of if on HasStatus, but my conclusion was that it's less error prone.

lobkovilya: > > All resources on Universal already have status
> 
> What do you mean?

Just the fact that PostgreSQL transaction creates a `status` column. 

> I was debating whether to implement noop SetStatus instead of if on HasStatus, but my conclusion was that it's less error prone.

Ok if you've already explored the idea then I'm fine with it

lobkovilya (approved):

pkg/plugins/resources/k8s/caching_converter.go
```diff
@@ -40,17 +40,6 @@ func (c *cachingConverter) ToCoreResource(obj k8s_model.KubernetesObject, out co
 		obj.GetResourceVersion(),
 		obj.GetObjectKind().GroupVersionKind().String(),
 	}, ":")
-	if obj.GetResourceVersion() == "" {
```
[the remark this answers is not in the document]
jakubdyszkiewicz: to avoid repetition of this block
```
if out.Descriptor().HasStatus {
		status, err := obj.GetStatus()
		if err != nil {
			return err
		}
		if err := out.SetStatus(status); err != nil {
			return err
		}
	}
```

I also think there was a mistake in this logic - the lack of return in the `if obj.GetResourceVersion() == "" {`.

PR #9687, opened 2024-03-21
feat(api): add a helper function to get all kuma targetRef kinds to be used in child repos

jijiechen:
### Checklist prior to review

  - N/A
  - Confirmed
  - No need
  - No need
  - No need

### DIFF
--- a/api/common/v1alpha1/ref.go
+++ b/api/common/v1alpha1/ref.go
@@ -42,6 +42,14 @@ func (k TargetRefKind) Less(o TargetRefKind) bool {
 	return order[k] < order[o]
 }
 
+func AllTargetRefKinds() []TargetRefKind {
+	keys := make([]TargetRefKind, 0, len(order))
+	for k := range order {
+		keys = append(keys, k)
+	}
+	return keys
+}
+
 // TargetRef defines structure that allows attaching policy to various objects
 type TargetRef struct {
 	// Kind of the referenced resource


api/common/v1alpha1/ref.go
```diff
@@ -42,6 +42,14 @@ func (k TargetRefKind) Less(o TargetRefKind) bool {
 	return order[k] < order[o]
 }
 
+func AllTargetRefKinds() []TargetRefKind {
+	keys := make([]TargetRefKind, 0, len(order))
+	for k := range order {
+		keys = append(keys, k)
+	}
+	return keys
```
jakubdyszkiewicz, line 50: nit: I'd use `return maps.Keys(order)` or even `maps.SortedKeys(order)`

jakubdyszkiewicz (approved):

api/common/v1alpha1/ref.go
jijiechen: I'll update.

api/common/v1alpha1/ref.go
```diff
@@ -42,6 +42,10 @@ func (k TargetRefKind) Less(o TargetRefKind) bool {
 	return order[k] < order[o]
 }
 
+func AllTargetRefKinds() []TargetRefKind {
+	return maps.Keys(order)
```
lahabana, line 46: SortedKeys may be better here as it's stable?

jijiechen: Yes, a sorted slice is definite better. But type of the keys is "TargetRefKind" which is not easy to be sorted here. 
We leave this responsibility to futures caller who care about the order?

PR #9689, opened 2024-03-21
docs(MADR): mesh service api

jakubdyszkiewicz:
### Checklist prior to review

MADR for MeshService API.

### DIFF
--- a/docs/madr/decisions/039-meshservice-api.md
+++ b/docs/madr/decisions/039-meshservice-api.md
@@ -1,0 +1,112 @@
+# MeshService API
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/6331
+
+## Context and Problem Statement
+
+Currently, our definition of "service" is derived from `kuma.io/service` tag placed in a Dataplane object. It has a several problems.
+
+**Scalability**
+* There are a number of objects whose size grows `O(#services)`: ServiceInsights, VIP ConfigMap, ZoneIngress availableServices. At scale storing objects of this size becomes a problem. There are limits for a size of an object in Kube API server.
+* Listing available services requires iterating through all Dataplane objects.
+* Separating Envoy endpoint generation from cluster generation. Right now a single change of Dataplane object would trigger cluster generation even though the cluster may not have changed.
+
+**Indirection**
+* There is no way to add metadata to the service. You can add metadata to a workload (Dataplane).
+  While we can compute shared metadata of workloads of same `kuma.io/service` and consider this a service metadata, there is a confusing indirection.
+
+## Considered Options
+
+* Shard objects
+* MeshService
+
+## Decision Outcome
+
+Chosen option: "MeshService", because it helps us to provide better foundation to build things on top of it. 
+
+### Positive Consequences
+
+* Solved scalability problems
+
+### Negative Consequences
+
+* A pretty big effort (implementation, a strategy of migration).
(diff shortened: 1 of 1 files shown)


lahabana (approved):

michaelbeaumont (approved):

docs/madr/decisions/039-meshservice-api.md
```diff
@@ -0,0 +1,119 @@
...
+kind: MeshService
+metadata:
+  name: redis
+  namespace: redis-system
+  labels: # you could then select this in when you use `MeshService` in `to` section
+    team: db-operators
+    kuma.io/mesh: default
+spec:
+  selector:
+    dataplaneTags: # tags in Dataplane object, see below
```
lobkovilya, line 59: Is it necessary to nest `dataplaneTags` into `selector`? Can we use a k8s-like API:
```yaml
spec:
  selector:
    app: redis
    k8s.kuma.io/namespace: redis-system
    kuma.io/zone: east-1
```

docs/madr/decisions/039-meshservice-api.md
```diff
@@ -0,0 +1,119 @@
...
+  ports:
+  - port: 6739
+    targetPort:
+      value: 6739
+    appProtocol: tcp
+  - name: some-port
+    port: 16739
+    targetPort:
+      name: target-port-from-container # name of the inbound
+    appProtocol: tcp
```
lobkovilya, line 72: k8s has both `appProtocol` and `protocol`, are we fine having only `appProtocol`?

docs/madr/decisions/039-meshservice-api.md
```diff
@@ -0,0 +1,119 @@
...
+status: # managed by CP. Not shared cross zone, but synced to global
+  availability: Online # | Offline | NotAvailable | PartialyDegraded
+  tls:
+    state: Ready
+    issuedBackends:
+      ca-1: 5
+  proxies:
+    offline: 3
+    online: 5
+    total: 8
```
lobkovilya, line 82: nit: do we need all 3? I guess when CP sets the status it's going to ensure anyway that `offline+online=total` right?

docs/madr/decisions/039-meshservice-api.md
jakubdyszkiewicz: yes, we care only about appProtocol

docs/madr/decisions/039-meshservice-api.md
jakubdyszkiewicz: I'd rather be explicit about this. I expect that in the future we will refactor Dataplane object to MeshProxy object that will have a one set of tags for all inbounds.

docs/madr/decisions/039-meshservice-api.md
jakubdyszkiewicz: I guess we don't

docs/madr/decisions/039-meshservice-api.md
lobkovilya: But this `selector` is a result of the conversion of the k8s service to MeshService? IIUC users can't change MeshService resources directly. So you'll need a way to create different selectors on k8s service, right?

Also if you replace some Dataplane objects with MeshProxy objects, don't you want `selector` to still work without changing `dataplaneTags` to `meshProxyTags`?

lobkovilya (approved):

PR #9692, opened 2024-03-22
docs(MADR): mesh service vips and hostnames

jakubdyszkiewicz:
MADR for #10034 

### Checklist prior to review

MADR of managing vips and hostnames using mesh service


docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,128 @@
...
+* List MeshServices
+* If a MeshService has no VIP, assign a Kuma VIP. Kubernetes Services would have a VIP already (ClusterIP), so we don’t need to assign an extra VIP.
+* To assign a VIP, we build an IPAM state reserving every address we already assigned and allocate the next address.
+  To solve the problem of “assigning immediately” a VIP of an old service to a new Service we can also store in memory recently removed VIPs for X seconds and set TTL of DNS responses for X seconds.
+  The current value of 30s is too long, especially that the DNS server is run locally in kuma-dp. Leader election should take more than X seconds, so if we were to lose this list, it’s not a problem.
+
+Following this process, we don’t have to store VIPs in any central place like we do currently (VIP ConfigMap for each Mesh).
+We also assign VIPs for ExternalServices, but this can work as long as we consider both ExternalServices and MeshServices for this process.
+
+To avoid collision with an existing VIPs mechanism, we can take a completely new CIDR (241.0.0.0/4).
```
lahabana, line 51: This remains configurable for users

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,128 @@
...
+MeshService has to be addressable with a hostname.
+In case of services in a Kubernetes cluster, we don't have to do anything because we can just rely on Kube DNS that resolves to a Kube VIP we placed for MeshService.
+
+In case of services in a Universal cluster, currently we provide two ways:
+* `<kuma.io/service>.mesh` address that resolves to Kuma VIP
+* VirtualOutbound resource to provide custom addresses
+
+Hostnames should be managed only by MeshOperator. Otherwise, we get into security issues such as one service owner can override a hostname of other service.
+Hostnames are created by convention in the majority of cases - Kubernetes hostnames are also such example.
+It is still possible for service owner to create unique hostname for their app by asking reaching out to mesh operator.
```
lahabana, line 66:
```suggestion
It is still possible for service owner to create unique hostname for their app by reaching out to mesh operator.
```

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,128 @@
...
+  kuma.io/zone: east
+spec: ...
+```
+We would generate such hostname `redis.demo-app.svc.cluster.east`
+
+Possible template functions/keys are:
+* `{{ name }}` - name of the MeshService
+* `{{ label "x" }}` - value of label `x`.
+
+If the template cannot be resolved (label is missing), the hostname won't be generated.
```
lahabana, line 100:
```suggestion
If the template cannot be resolved (label is missing), the hostname won't be generated with a `NotAvailable` status and an appropriate `reason`.
```

Though what do we put for hostname in this case?

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,128 @@
...
+
+Hostnames should be managed only by MeshOperator. Otherwise, we get into security issues such as one service owner can override a hostname of other service.
+Hostnames are created by convention in the majority of cases - Kubernetes hostnames are also such example.
+It is still possible for service owner to create unique hostname for their app by asking reaching out to mesh operator.
+
+To manage hostname, we introduce new object called `HostnameGenerator`
+
+```yaml
+type: HostnameGenerator
+name: k8s-zone-hostnames
```
lahabana, line 72: There can be a mesh here?

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,128 @@
...
+
+### VIPs management
+
+We have to assign a VIP for each MeshService to be able to address it using transparent proxy.
+
+MeshService object contains VIPs in status.
+```yaml
+status: # managed by CP. Not shared cross zone, but synced to global
+  vips:
+  - ip: <kube_cluster_ip> # or kuma VIP
```
lahabana, line 32: What about headless service? These are pod Ips?

michaelbeaumont: Yes. I'll cover that in https://github.com/kumahq/kuma/pull/9722

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,128 @@
...
+* To assign a VIP, we build an IPAM state reserving every address we already assigned and allocate the next address.
+  To solve the problem of “assigning immediately” a VIP of an old service to a new Service we can also store in memory recently removed VIPs for X seconds and set TTL of DNS responses for X seconds.
+  The current value of 30s is too long, especially that the DNS server is run locally in kuma-dp. Leader election should take more than X seconds, so if we were to lose this list, it’s not a problem.
+
+Following this process, we don’t have to store VIPs in any central place like we do currently (VIP ConfigMap for each Mesh).
+We also assign VIPs for ExternalServices, but this can work as long as we consider both ExternalServices and MeshServices for this process.
+
+To avoid collision with an existing VIPs mechanism, we can take a completely new CIDR (241.0.0.0/4).
+
+It's not expected for service to have multiple VIPs, but we want expose this possibility for the future (therefore `vips` field is a list).
```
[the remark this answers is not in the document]
michaelbeaumont: Or put differently, will any usage of `vips` implicitly assume there's one entry? If so, we don't need a list.

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,128 @@
...
+type: HostnameGenerator
+name: k8s-zone-hostnames
+spec:
+  meshServiceSelector:
+    matchLabels:
+      kuma.io/zone: east 
+  template: {{ label "k8s.kuma.io/service-name" }}.{{ label "k8s.kuma.io/namespace" }}.svc.cluster.{{ label "kuma.io/zone" }}
+```
+
+`HostnameGenerator` is a namespaced-scoped object on Kubernetes, but can only be applied in system namespace.
```
michaelbeaumont, line 80: Why is it namespaced?

docs/madr/decisions/040-meshservice-hostname-vips.md
michaelbeaumont: I think maybe it's worth covering here how we update status in the case of a headless MeshService

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: It's just safer. We can limit the namespace to a system namespace, but I'd rather have it just in case.

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: I'm not sure. We don't have a mechanism yet to have a resource that can optionally belong to a mesh. For example, in case of Secrets we created Secret (mesh-bound) and GlobalSecret (not mesh-bound).

I think we need to decide on one way or another. I think I'd go with "global" and then extend this to be optionally "mesh-bound" if needed.

WDYT?

docs/madr/decisions/040-meshservice-hostname-vips.md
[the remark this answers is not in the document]
jakubdyszkiewicz: You have a point. We could go with just `vip` and create `vips` if needed. All usages then would need to read both, but it seems we will control all of them so it's fine.

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,128 @@
...
+To manage hostname, we introduce new object called `HostnameGenerator`
+
+```yaml
+type: HostnameGenerator
+name: k8s-zone-hostnames
+spec:
+  meshServiceSelector:
+    matchLabels:
+      kuma.io/zone: east 
+  template: {{ label "k8s.kuma.io/service-name" }}.{{ label "k8s.kuma.io/namespace" }}.svc.cluster.{{ label "kuma.io/zone" }}
```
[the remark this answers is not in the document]
jakubdyszkiewicz: `clustermesh` is strange. `svc.mesh.{{ label "kuma.io/zone" }}` on the other hand sounds good to me.

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: added comment about it

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: added that it can be pod ip

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+
+If MeshService was created based on Kubernetes Service, we can just take ClusterIP, but otherwise we need to assign each MeshService a Kuma VIP.
+Kuma VIP is a non-routable IP used only by Envoy to indicate destination (like a host header, but on L4). We currently do this with 240.0.0.0/4 CIDR.
+
+Kuma VIPs are assigned asynchronously with MeshService creation. Assigning Kuma VIPs as sync operation on MeshService creation would create a race condition on VIP allocation.
+
+VIP assigning process (single-threaded):
+* List MeshServices
+* If a MeshService has no VIP, assign a Kuma VIP. Kubernetes Services would have a VIP already (ClusterIP), so we don’t need to assign an extra VIP.
+* To assign a VIP, we build an IPAM state reserving every address we already assigned and allocate the next address.
```
lukidzi, line 44: How is it going to work with multiple instances of control-plane?

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+```
+We would generate such hostname `redis.demo-app.svc.cluster.east`
+
+Possible template functions/keys are:
+* `{{ name }}` - name of the MeshService
+* `{{ label "x" }}` - value of label `x`.
+
+If the template cannot be resolved (label is missing), the hostname won't be generated with a `NotAvailable` status and an appropriate `reason`.
+
+Generated hostnames will be placed in the status of MeshService
```
lobkovilya, line 101: It would be nice to have similar step-by-step algorithm explained (as you did for VIPs)

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+    ip: <kube_cluster_ip> # or kuma VIP or Pod IP for Headless Service
+    type: Kubernetes # | Kuma
+```
+
+If MeshService was created based on Kubernetes Service, we can just take ClusterIP, but otherwise we need to assign each MeshService a Kuma VIP.
+Kuma VIP is a non-routable IP used only by Envoy to indicate destination (like a host header, but on L4). We currently do this with 240.0.0.0/4 CIDR.
+
+Kuma VIPs are assigned asynchronously with MeshService creation. Assigning Kuma VIPs as sync operation on MeshService creation would create a race condition on VIP allocation.
+
+VIP assigning process (single-threaded):
```
lobkovilya, line 41: is it a controller on k8s?

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+## Pros and Cons of the Options
+
+### VIPs management
+
+We have to assign a VIP for each MeshService to be able to address it using transparent proxy.
+
+MeshService object contains VIPs in status.
+```yaml
+status: # managed by CP. Not shared cross zone, but synced to global
+  vip:
```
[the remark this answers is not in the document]
michaelbeaumont: See earlier discussion https://github.com/kumahq/kuma/pull/9692#discussion_r1539935011

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+    ip: <kube_cluster_ip> # or kuma VIP or Pod IP for Headless Service
```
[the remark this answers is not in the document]
michaelbeaumont: I don't think so, or what port wouldn't be found in the `ports` list?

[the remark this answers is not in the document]
lobkovilya: Today we always use port 80 for VIPs and it's configured in `KUMA_DNS_SERVER_SERVICE_VIP_PORT`. If we plan to keep this functionality, I guess we need a way to specify a port inside the vip, no?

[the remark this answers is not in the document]
michaelbeaumont: I don't think we're keeping this functionality but we should call that fact out @jakubdyszkiewicz

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
+# Hostnames and VIP management in MeshService
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/6331
+
+## Context and Problem Statement
+
+With introduction of MeshService, we need to have a mechanism to assign VIPs and hostnames to MeshServices.
+Currently, VIPs are managed in a ConfigMap per Mesh. Hostnames are managed using VirtualOutbound.
```
lahabana, line 10: Might be worth explaining the shortcoming of the current implem

docs/madr/decisions/040-meshservice-hostname-vips.md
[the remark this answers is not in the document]
lahabana: This functionality was not very good so I'd argue for dropping!

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+If MeshService was created based on Kubernetes Service, we can just take ClusterIP, but otherwise we need to assign each MeshService a Kuma VIP.
+Kuma VIP is a non-routable IP used only by Envoy to indicate destination (like a host header, but on L4). We currently do this with 240.0.0.0/4 CIDR.
+
+Kuma VIPs are assigned asynchronously with MeshService creation. Assigning Kuma VIPs as sync operation on MeshService creation would create a race condition on VIP allocation.
+
+VIP assigning process (single-threaded):
+* List MeshServices
+* If a MeshService has no VIP, assign a Kuma VIP. Kubernetes Services would have a VIP already (ClusterIP), so we don’t need to assign an extra VIP.
```
lahabana, line 43: IIRC ClusterIP are set asynchronously. How do we avoid the assigning of a Kuma VIP in this case? (What's the code logic there)

docs/madr/decisions/040-meshservice-hostname-vips.md
lahabana: I guess this only happens on the master?

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+* To assign a VIP, we build an IPAM state reserving every address we already assigned and allocate the next address.
+  To solve the problem of “assigning immediately” a VIP of an old service to a new Service we can also store in memory recently removed VIPs for X seconds and set TTL of DNS responses for X seconds.
+  The current value of 30s is too long, especially that the DNS server is run locally in kuma-dp. Leader election should take more than X seconds, so if we were to lose this list, it’s not a problem.
+
+Following this process, we don’t have to store VIPs in any central place like we do currently (VIP ConfigMap for each Mesh).
+We also assign VIPs for ExternalServices, but this can work as long as we consider both ExternalServices and MeshServices for this process.
```
lahabana, line 49: Should ExternalServices just use a different CIDR?

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+Kuma VIP is a non-routable IP used only by Envoy to indicate destination (like a host header, but on L4). We currently do this with 240.0.0.0/4 CIDR.
+
+Kuma VIPs are assigned asynchronously with MeshService creation. Assigning Kuma VIPs as sync operation on MeshService creation would create a race condition on VIP allocation.
+
+VIP assigning process (single-threaded):
+* List MeshServices
+* If a MeshService has no VIP, assign a Kuma VIP. Kubernetes Services would have a VIP already (ClusterIP), so we don’t need to assign an extra VIP.
+* To assign a VIP, we build an IPAM state reserving every address we already assigned and allocate the next address.
+  To solve the problem of “assigning immediately” a VIP of an old service to a new Service we can also store in memory recently removed VIPs for X seconds and set TTL of DNS responses for X seconds.
+  The current value of 30s is too long, especially that the DNS server is run locally in kuma-dp. Leader election should take more than X seconds, so if we were to lose this list, it’s not a problem.
```
lahabana, line 46: Are we sure we're ok about the "in-memory" part.

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+Following this process, we don’t have to store VIPs in any central place like we do currently (VIP ConfigMap for each Mesh).
+We also assign VIPs for ExternalServices, but this can work as long as we consider both ExternalServices and MeshServices for this process.
+
+To avoid collision with an existing VIPs mechanism, we can take a completely new CIDR by default (241.0.0.0/4).
+This CIDR can be configured by the user.
+
+### Hostname Generator
+
+MeshService has to be addressable with a hostname.
+In case of services in a Kubernetes cluster, we don't have to do anything because we can just rely on Kube DNS that resolves to a Kube VIP we placed for MeshService.
```
lahabana, line 57: Multizone services are a thing though right?

docs/madr/decisions/040-meshservice-hostname-vips.md
lahabana: Sounds good but then let's name it `GlobalHostnameGenerator` should like k8s convention of Cluster scoped resources should be named `Cluster...`

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+Hostnames are created by convention in the majority of cases - Kubernetes hostnames are also such example.
+It is still possible for service owner to create unique hostname for their app by reaching out to mesh operator.
+
+To manage hostname, we introduce new object called `HostnameGenerator`
+
+```yaml
+type: HostnameGenerator
+name: k8s-zone-hostnames
+spec:
+  meshServiceSelector:
```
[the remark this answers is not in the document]
lahabana: I agree it's nicer

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,127 @@
...
+  kuma.io/zone: east
+spec: ...
+```
+We would generate such hostname `redis.demo-app.svc.cluster.east`
+
+Possible template functions/keys are:
+* `{{ name }}` - name of the MeshService
+* `{{ label "x" }}` - value of label `x`.
+
+If the template cannot be resolved (label is missing), the hostname won't be generated with a `NotAvailable` status and an appropriate `reason`.
```
lahabana, line 99: I guess that's on the MeshService? Shouldn't the status be more a `HostnameNotAvailable` or something like this? Feels like NotAvailable is a bit too generic.
I guess we also send an event right?

lahabana: Ok got confused because status is explained after.

lahabana: Should there be a Condition on MeshService?

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: It's only processed on one instance - a leader.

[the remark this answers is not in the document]
jakubdyszkiewicz: I'd go with unique across all meshes.

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: I'd say it's Kuma component with a ticker. We need to aggregate all MeshServices to compute this.

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: I think it's fine. The alternative is to keep this in configmap which means more I/O and more dependencies for VIP allocation process

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: yes, of course. It's for local services in the cluster

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: Hard to say. I did some reading on conditions https://maelvls.dev/kubernetes-conditions/
To try to convert this to conditions it would be something like
```
status:
  conditions:
  - type: HostnameAssigned
    status: "False"
    message: "There is a conflict..."
```
this `HostnameAssigned` would be for all hostnames. Now the questions are
1) What to do when there is one hostname and second cannot be assigned because of conflict. Is it True or False?
2) What to do when we have a Kube service with kube hostname. We don't put Kube DNS hostname in list of hostnames, so is it True or False?

My feeling is that the status of each hostname is more useful. Unless we push conditions into each item of `addresses`, but sounds like a simple status & reason would be simpler API?

docs/madr/decisions/040-meshservice-hostname-vips.md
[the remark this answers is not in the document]
jakubdyszkiewicz: feels like the other way around. Like described here, we will take list of all MeshService, then assign vips. Sharding this process for multiple meshes sounds like unnecessary work

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: good idea!

docs/madr/decisions/040-meshservice-hostname-vips.md
lahabana: I don't know but It's nice in k8s to have a systematic way to understand if something is cluster scoped or not.

In anycase I'd stay away from GlobalHostnameGenerator for the multizone service as it's confusing with GlobalSecret

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: added to madr

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: added

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: added

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: added clarification on it

docs/madr/decisions/040-meshservice-hostname-vips.md
lahabana: It could be `AllHostnameAssigned` no?

jakubdyszkiewicz: and if no hostnames then it's `True`?

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,140 @@
...
+      origin:
+        kind: HostnameGenerator
+        name: "k8s-zone-hostnames" # name of the HostnameGenerator
+      reason: "not available because of the clash with ..." # reason if there was a problem
+```
+
+Hostname generation process (only computed by leader):
+* List MeshService and HostnameGenerators
+* Sort HostnameGenerators by precedence
+* Take each HostnameGenerator and add hostnames for a Service (on the side, do not modify MeshService object). Track generated hostnames to detect collisions
```
lukidzi, line 123: What about a case where we have 2 services in 2 different mesh but in the same namespace? In this case, we might not have a clash of IP addresses but hostnames.

jakubdyszkiewicz: > What about a case where we have 2 services in 2 different mesh but in the same namespace?

How to define this using Kubernetes objects? Can you give an example?

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: talked offline: let's try this solution for now. We can evolve if needed

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: talked offline: let's start with status and then we can do conditions if needed

docs/madr/decisions/040-meshservice-hostname-vips.md
lukidzi: True, it's not possible but is it possible in universal?

jakubdyszkiewicz: Ok, true. This process should be sharded by mesh. I mean this step "Track generated hostnames to detect collisions" should be per mesh. Let me update the MADR

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,140 @@
...
+
+To manage hostname, we introduce new object called `HostnameGenerator`
+
+```yaml
+type: HostnameGenerator
+name: k8s-zone-hostnames
+spec:
+  selector:
+    meshService:
+      Labels:
```
lahabana, line 79: I guess the capital L is a typo

docs/madr/decisions/040-meshservice-hostname-vips.md
```diff
@@ -0,0 +1,140 @@
...
+  kuma.io/zone: east
+spec: ...
+```
+We would generate such hostname `redis.demo-app.svc.mesh.east`
+
+Possible template functions/keys are:
+* `{{ name }}` - name of the MeshService
+* `{{ label "x" }}` - value of label `x`.
+
+If the template cannot be resolved (label is missing), the hostname won't be generated with a `NotAvailable` status and an appropriate `reason`.
```
lahabana, line 106: hostname in addresses would be empty then?

The status would look like:

```yaml
status:
  addresses:
    - hostname: ""
      status: NotAvailable # if there was a collision
      origin:
        kind: HostnameGenerator
        name: "k8s-zone-hostnames" # name of the HostnameGenerator
      reason: "not available because of the clash with ..." # reason if there was a problem
```

lahabana (approved):

lukidzi (approved):

docs/madr/decisions/040-meshservice-hostname-vips.md
jakubdyszkiewicz: I think we would just skip the template. Imagine you have this template
```
{{ label "kuma.io/display-name" }}.{{ label "k8s.kuma.io/namespace" }}.svc.mesh.{{ label "k8s.kuma.io/zone" }}
```
You have a universal zone that syncs the service to you that does not have such labels.
You can of course select all kubernetes zones one by one for this generator, but that's annoying.

Issue #9431, opened 2024-02-28
Pod does not start when turnning on waitForDataplaneReady with kuma-cni enabled


jijiechen:
### What happened?

It's reported by some kuma users that the data plane pods can not be started on Kubernetes when they enable these features together:

1. kuma-cni
2. waitForDataplaneReady

When check the pod state, we can find the pod hangs in a "ContainerCreating" status with no IP address assigned.

The pod will recover after a few minutes when the postStartHook timeout to complete and the `kuma-sidecar` container restarts. When retrieve the logs from the restarted container using `kubectl logs <pod> -c kuma-sidecar -p`, we'll get these log entries:

```log
2024-02-28T02:31:18.011Z	INFO	kuma-dp.run	generating bootstrap configuration
2024-02-28T02:31:18.011Z	INFO	dataplane	trying to fetch bootstrap configuration from the Control Plane
2024-02-28T02:31:18.020Z	INFO	dataplane	Dataplane entity is not yet found in the Control Plane. If you are running on Kubernetes, the control plane is most likely still in the process of converting Pod to Dataplane. If it takes too long, check pod events and control plane logs to see possible cause. Retrying.	{"backoff": "3s"}
```

This is reproducable on multi-versions of Kuma, and it only happens when kuma-cni enabled.

jijiechen:
When checking the control plane logs, we can see the pods are not ready to be comsumed by kuma cp at the first event and the reconcilation was successful 3 minutes later:

```log
2024-01-18T02:20:21.522Z	LEVEL(-5)	kube-manager	Reconciling	{"controller": "pod", "controllerGroup": "", "controllerKind": "Pod", "Pod": {"name":"workload-74c94f6878-r7p8n","namespace":"workload-ns-1"}, "namespace": "workload-ns-1", "name": "workload-74c94f6878-r7p8n", "reconcileID": "e0caa01d-59cd-4912-8ec2-612e4391c170"}
2024-01-18T02:20:21.522Z	DEBUG	controllers.Pod	reconcile	{"pod": {"name":"workload-74c94f6878-r7p8n","namespace":"workload-ns-1"}}
[...]

PR #9699, opened 2024-03-25
feat(kuma-cni): add a init container to validate iptables rules are applied

jijiechen:
fixes https://github.com/kumahq/kuma/issues/9431

Like what Istio does, I'm trying to introduce a container named "kuma-cni-validation" to validate the iptables rules.

Such a container can also mitegate the issue of "IP address assignment hang" during the postStart check.

### Checklist prior to review

  - https://github.com/kumahq/kuma/issues/9431
  - Yes, I'll update
  - No need

> Changelog: Introduce a new init container kuma-cni-verification


api/common/v1alpha1/ref.go
```diff
@@ -43,9 +43,17 @@ func (k TargetRefKind) Less(o TargetRefKind) bool {
 }
 
 func AllTargetRefKinds() []TargetRefKind {
-	return maps.Keys(order)
+	keys := maps.Keys(order)
+	sort.Sort(TargetRefKindSlice(keys))
+	return keys
 }
 
+type TargetRefKindSlice []TargetRefKind
```
jijiechen, line 51: This is an enhancement to a previous PR at https://github.com/kumahq/kuma/pull/9687

pkg/transparentproxy/validate/validator.go
```diff
@@ -0,0 +1,191 @@
...
+			ServerListenPort:    port,
+			ClientConnectIP:     clientConnectIP,
+			ClientRetryInterval: validationInterval,
+		},
+		Logger: logger,
+	}
+}
+
+func (validator *Validator) Run() error {
+	validator.Logger.Info("Starting iptables validation...")
```
slonka, line 59:
```suggestion
	validator.Logger.Info("starting iptables validation")
```

Don't use elipsis and start lowercase - check out our logging madr https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/003-logging.md?plain=1#L122-L130 https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/003-logging.md?plain=1#L34

pkg/transparentproxy/validate/validator.go
```diff
@@ -0,0 +1,191 @@
...
+	sExit := make(chan struct{})
+
+	sError := validator.runServer(sExit)
+	select {
+	case serverErr := <-sError:
+		if serverErr == nil {
+			serverErr = fmt.Errorf("server exited unexpectedly")
+		}
+		validator.Logger.Error(serverErr, "Validation failed")
+		return serverErr
```
slonka, line 69: either log or return the error 

https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/003-logging.md?plain=1#L56-L79

pkg/transparentproxy/validate/validator.go
```diff
@@ -0,0 +1,191 @@
...
+			return
+		default:
+		}
+	}
+}
+
+func (validator *Validator) runClient() error {
+	c := LocalClient{ServerIP: validator.Config.ClientConnectIP, ServerPort: validator.Config.ClientConnectPort}
+	backoff := retry.WithMaxRetries(validationRetries, retry.NewConstant(validator.Config.ClientRetryInterval))
+	return retry.Do(context.TODO(), backoff, func(ctx context.Context) error {
```
slonka, line 148: missing context

pkg/transparentproxy/validate/validator_test.go
```diff
@@ -0,0 +1,74 @@
...
+		err := validator.Run()
+
+		// then
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("should return fail when no iptables rules setup", func() {
+		// given
+		validator := createValidator(false)
+		validator.Config.ClientRetryInterval = 30 * time.Millisecond
```
slonka, line 58: Not sure I understand this test completely. There is no iptables injection going on in the test we just make te retry interval so small that it fails, right? Shouldn't this be even lower to be less sensitive to flakes?

slonka (commented):
A bunch of questions. Might need another round here.

app/kumactl/cmd/install/install_transparent_proxy_validator.go
```diff
@@ -0,0 +1,56 @@
...
+func newInstallTransparentProxyValidator() *cobra.Command {
+	args := transparentProxyValidatorArgs{
+		IpFamilyMode:         "dualstack",
+		ValidationServerPort: validate.ValidationServerPort,
+	}
+	cmd := &cobra.Command{
+		Use:   "transparent-proxy-validator",
+		Short: "Validates if transparent proxy has been set up successfully",
+		Long: `Validates the transparent proxy setup by testing if the applied 
+hosts iptables rules are working correctly.
```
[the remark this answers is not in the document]
jijiechen: It's a little confusing here. Host here simply means the iptables applied onto the NET namespace of the pod. The CNI is running on the host process space, so I wrote "host". But as you know the rules are actually generated and applied in the CNI process, so there is not anything changed in terms of the iptables rules generation and applying

app/kumactl/cmd/install/install_transparent_proxy_validator.go
```diff
@@ -0,0 +1,56 @@
...
+Follow the following steps to validate:
+ 1) install the transparent proxy using 'kumactl install transparent-proxy'
+ 2) run this command to validate if the installation was successful
+
+The result will be shown as text in stdout as well as the exit code
+`,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			log := core.NewLoggerTo(os.Stdout, kuma_log.InfoLevel).WithName(defaultLogName)
+
+			ipv6Supported, _ := transparentproxy.HasLocalIPv6()
```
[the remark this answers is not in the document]
jijiechen: Because if there is any errors, the first return value will be false, and it'll be correctly handled in that case

pkg/plugins/runtime/k8s/webhooks/injector/injector.go
```diff
@@ -180,6 +184,22 @@ func (i *KumaInjector) InjectKuma(ctx context.Context, pod *kube_core.Pod) error
 		} else {
 			pod.Spec.InitContainers = append(pod.Spec.InitContainers, patchedIc)
 		}
+	} else if podRedirect.RedirectInbound {
+		ic := i.NewValidationContainer(podRedirect.IpFamilyMode, fmt.Sprintf("%d", podRedirect.RedirectPortInbound), sidecarTmp.Name)
+		patchedIc, err := i.applyCustomPatches(logger, ic, initPatches)
+		if err != nil {
+			return err
+		}
+		enabled, _, err := metadata.Annotations(pod.Annotations).GetEnabled(metadata.KumaInitFirst)
```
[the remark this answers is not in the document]
jijiechen: I didn't test that for now

pkg/transparentproxy/validate/validator.go
```diff
@@ -0,0 +1,191 @@
...
+func NewValidator(useIpv6 bool, port uint16, logger logr.Logger) *Validator {
+	// Traffic to lo (but not 127.0.0.1) by sidecar will be redirected to  KUMA_MESH_INBOUND_REDIRECT, so:
+	// connect to 127.0.0.6 should be redirected to 127.0.0.1
+	// connect to ::6       should be redirected to ::1
+	serverListenIP, _ := netip.AddrFromSlice([]byte{127, 0, 0, 1})
+	clientConnectIP, _ := netip.AddrFromSlice([]byte{127, 0, 0, 6})
+
+	if useIpv6 {
+		serverListenIP, _ = netip.AddrFromSlice([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1})
+		clientConnectIP, _ = netip.AddrFromSlice([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 6})
```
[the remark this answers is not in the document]
jijiechen: These are obviously correct IP addresses, because they are hard coded, so there is no need to parse, and they should not generate errors.

pkg/transparentproxy/validate/validator_test.go
jijiechen: Right, we are not injecting any iptables rules in the test cases, the cases here are only verifying the logic of the validator itself should behave correctly 
I think 30ms is a reasonable value for testing just to make it faster and there should be no flakiness here because the connection will never establish successfully without the redirection set up by the iptables rules.

pkg/transparentproxy/validate/validator.go
jijiechen: The whole validator should be running relatively quickly (at most 10s), is it still needed to make it cancelable? Or do you think a context can be beneficial for other reasons?

app/kumactl/cmd/install/install_transparent_proxy_validator.go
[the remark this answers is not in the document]
jijiechen: I'll reword this statement.

pkg/transparentproxy/validate/validator.go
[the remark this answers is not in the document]
jijiechen: I'll use `netip.MustParseAddr` to replace them.

jijiechen:
Hi @slonka  I've updated the code according to the discussions and your suggestions, some threads are left open and waiting for your further response.
Could you review them at your convenience?

pkg/transparentproxy/validate/validator.go
slonka: so I guess we can use `context.Background()` right? I know they are eventually the same but `TODO` indicates that PR is not finished yet

pkg/transparentproxy/validate/validator_test.go
slonka: we could add a comment in the code that "the connection will never establish successfully" and that this value does not matter that much as long as it's low so the test is fast, right?

slonka:
@jijiechen - answered, just 2 small things and we can 👍

pkg/transparentproxy/validate/validator_test.go
slonka:
```suggestion
		validator.Config.ClientRetryInterval = 30 * time.Millisecond // just to make test faster and there should be no flakiness here because the connection will never establish successfully without the redirection 
```

slonka (approved):

PR #9701, opened 2024-03-25
feat(images/kuma-init): use iptables-wrapper to use correct iptables version

bartsmykla:
As of iptables 1.8, the iptables command line clients come in two different versions/modes: "legacy", which uses the kernel iptables API just like iptables 1.6 and earlier did, and "nft", which translates the iptables command-line API into the kernel nftables API.

Because they connect to two different subsystems in the kernel, you cannot mix and match between them; in particular, if you are adding a new rule that needs to run either before or after some existing rules (such as the system firewall rules), then you need to create your rule with the same iptables mode as the other rules were created with, since otherwise the ordering may not be what you expect. (eg, if you prepend a rule using the nft-based client, it will still run after all rules that were added with the legacy iptables client.)

### Checklist prior to review

  - No relevant issues
  - It shouldn't
  - Tested manually
  - There is no need


mk/build.mk
```diff
@@ -83,6 +83,10 @@ $(foreach target,$(BUILD_RELEASE_BINARIES) $(BUILD_TEST_BINARIES),$(eval $(call
 # Build_Go_Application is a build command for the Kuma Go applications.
 Build_Go_Application = GOOS=$(1) GOARCH=$(2) $$(GOENV) go build -v $$(GOFLAGS) $$(LD_FLAGS) -o $$@/$$(notdir $$@)
 
+build/iptables-wrapper:
+	git clone https://github.com/kubernetes-sigs/iptables-wrappers.git --single-branch $(@)
+	git -C $(@) checkout 5792812d9e5a5bb7f22d79d557bbfeece253343d # latest commit on master (March 25, 2024)
```
lahabana, line 88: We add a lot of issues with random git clones in our builds. Is there anyway we can work around this?

bartsmykla: We can add it as a git submodule

michaelbeaumont: Can we fork it and add the binaries to a release?

lahabana: @bartsmykla can we do the fork thing?

bartsmykla: Could you elaborate on what do you mean by putting in the release? It's binary nad shell script which needs to be placed in the container

bartsmykla: Done

michaelbeaumont (approved):

Issue #9704, opened 2024-03-25
MADR for Kubernetes UX of MeshService 


jakubdyszkiewicz:
### Description

Create MADR for Kubernetes UX of MeshService.

Include headless service if MADR is not too long.

Issue #9707, opened 2024-03-25
MADR for Policy Matching and Inspect API of MeshService


jakubdyszkiewicz:
### Description

Create a MADR for Policy Matching and Inspect API of MeshService

Remember:
- StatefulSet (considering that MeshService is per Pod)

lahabana:
Duplicate of https://github.com/kumahq/kuma/issues/9707

PR #9722, opened 2024-03-26
docs(MADR): add Kubernetes UX for `MeshService`

michaelbeaumont:
### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- #9704 

### DIFF
--- a/docs/madr/decisions/041-meshservice-kubernetes-ux.md
+++ b/docs/madr/decisions/041-meshservice-kubernetes-ux.md
@@ -47,6 +47,35 @@ including supporting named `targetPorts`.
 Note that we only support `Service.ports[].protocol: TCP`, which is also the
 default.
 
+### Headless Service with selectors
+
+In Kubernetes, a headless Service with selectors is used to create a DNS record
+for every Pod selected by the Service that points directly to the Pod's IP.
+
+To support this with Kuma, we will create a `MeshService` per Pod, each
+represented by the hostname allocated by the headless Service and the Pod
+IP as the "VIP" and single endpoint.
+
+In order to do this we need to have a list of all the Pods selected by the
+Service, which we can get by looking at `EndpointSlices`. These resources hold a
+list of endpoints, each of which has a `targetRef`. If the `targetRef` is `kind:
+Pod`, we can rely on the naming of `Dataplane` objects and directly select a
+given `Dataplane` by setting `spec.selector.dataplaneName` to the name of the
+`Pod`.
+
+```
+kind: MeshService
+spec:
+  selector:
+    dataplaneName: pod-1
+    # dataplaneTags: ...
+```
+
+#### Policy matching
+
+Note that this prevents using `kind: MeshService` to select all Pods of a
+StatefulSet. In another MADR, we will cover this use case.
+
 ### Positive Consequences
 
 * Users don't have to think about creating `MeshService`
--- a/docs/madr/decisions/041-meshservice-kubernetes-ux.md
+++ b/docs/madr/decisions/041-meshservice-kubernetes-ux.md
@@ -1,0 +1,64 @@
+# `MeshService` Kubernetes UX
+
+* Status: accepted
+
+Technical Story: #9704
+
+## Context and Problem Statement
+
+The new `MeshService` resource in Kubernetes zones maps 1-1 to Kubernetes
+`Service` resources. Additionally, users should not have to manage
+`MeshServices` directly but instead a `MeshService` should be derived directly
+from its `Service` with the possibility of users annotating the `Service`
+to control specifics.
+
+## Considered Options
+
+* CP manages `MeshService` from `Service`
+* Users have to create `MeshService`
+
+## Decision Outcome
+
+We will convert `Services` to `MeshServices` in a Kubernetes controller,
+assuming:
+
+1. The namespace is part of the mesh with `kuma.io/sidecar-injection: "true"`
+1. The `Mesh` of the `Service` or its namespace exists.
+1. The `Service` hasn't opted out via `kuma.io/ignore: "true"`
+
+### Metadata
+
+The `ownerReference` of the `MeshService` points to the `Service`.
+Labels are copied from the `Service` object to the `MeshService` object.
+
+### VIP
+
+For `MeshServices` in Kubernetes zones, we no longer create a VIP but instead
+reuse the `ClusterIP` field allocated by the Kubernetes control plane and add
+this to `status.vips` on `MeshService`.
+
+Note this happens asynchronously.
+
+### `ports`
+
+The `MeshService` ports are derived from the `ports` field on `Service`. Note
+that `meshservice_api.Port.Protocol` is derived from `ServicePort.AppProtocol`.
+
+Note that we only support `Service.ports[].protocol: TCP`, which is also the
+default.
+
+### Positive Consequences
+
+* Users don't have to think about creating `MeshService`
+* We don't need to allocate additional VIPs
+
+### Negative Consequences
+
+* None?
+
+## Pros and Cons of the Options
+
+### Manual management
+
+* Bad, because almost always the `MeshService` should be entirely derived from
+  the `Service` object
--- a/docs/madr/decisions/041-meshservice-kubernetes-ux.md
+++ b/docs/madr/decisions/041-meshservice-kubernetes-ux.md
@@ -74,6 +74,11 @@ spec:
     # dataplaneTags: ...
 ```
 
+Note that Kubernetes does not allocate a `ClusterIP` for headless services, it
+only creates a round-robin DNS record to point to PodIPs. Kuma does not
+allocate a VIP either. In a zone, users can rely on kube-dns. Cross-zone
+behavior will be covered in a subsequent MADR.
+
 #### Policy matching
 
 Note that this prevents using `kind: MeshService` to select all Pods of a
--- a/docs/madr/decisions/041-meshservice-kubernetes-ux.md
+++ b/docs/madr/decisions/041-meshservice-kubernetes-ux.md
@@ -46,7 +46,7 @@ The `MeshService` ports are derived from the `ports` field on `Service`,
 including supporting named `targetPorts`.
 
 Note that we only support `Service.ports[].protocol: TCP`, which is also the
-default.
+default. We copy `appProtocol` to the `MeshService` port entry as well.
 
 ### Headless Service with selectors
 
@@ -77,6 +77,71 @@ spec:
 Note that this prevents using `kind: MeshService` to select all Pods of a
 StatefulSet. In another MADR, we will cover this use case.
 
+### Examples
+
+The Kubernetes Service
+
+```
+kind: Service
+metadata:
+  name: redis
+  namespace: redis-system
+  labels:
+    team: db-operators
+    kuma.io/mesh: default
+spec:
+  selector:
+    app.kubernetes.io/name: redis
+  ports:
+  - port: 6739
+    targetPort: 6739
+    protocol: TCP
+  - name: admin
+    port: 8080
+    protocol: TCP
+    appProtocol: http
+```
+
+would be converted to:
+
+```yaml
+kind: MeshService
+metadata:
+  name: redis
+  namespace: redis-system
+  labels:
+    team: db-operators
+    kuma.io/mesh: default
+spec:
+  selector:
+    dataplaneTags: # tags in Dataplane object, see below
+      app.kubernetes.io/name: redis
+      k8s.kuma.io/namespace: redis-system # added automatically
+      kuma.io/zone: east-1 # added automatically
+  ports:
+  - port: 6739
+    targetPort:
+      value: 6739
+    appProtocol: tcp
+  - name: admin
+    port: 8080
+    appProtocol: http
+  clusterIP: 192.168.10.1
+  type: ClusterIP
+status:
+  availability: Online
+  tls:
+    state: Ready
+    issuedBackends:
+      ca-1: 5
+  proxies:
+    offline: 3
+    online: 5
+  vips:
+  - ip: 192.168.10.1
+    type: Kubernetes
+```
+
 ### Positive Consequences
 
 * Users don't have to think about creating `MeshService`
@@ -84,7 +149,7 @@ StatefulSet. In another MADR, we will cover this use case.
 
 ### Negative Consequences
 
-* None?
+* One API object per Pod, as opposed to one per Service
 
 ## Pros and Cons of the Options
 
--- a/docs/madr/decisions/041-meshservice-kubernetes-ux.md
+++ b/docs/madr/decisions/041-meshservice-kubernetes-ux.md
@@ -56,6 +56,8 @@ for every Pod selected by the Service that points directly to the Pod's IP.
 To support this with Kuma, we will create a `MeshService` per Pod, each
 represented by the hostname allocated by the headless Service and the Pod
 IP as the "VIP" and single endpoint.
+The name of said `MeshServices` has the format `<service-name>-<pod-hash>`
+where the `Service` name is appropriately truncated.
 
 In order to do this we need to have a list of all the Pods selected by the
 Service, which we can get by looking at `EndpointSlices`. These resources hold a
--- a/docs/madr/decisions/041-meshservice-kubernetes-ux.md
+++ b/docs/madr/decisions/041-meshservice-kubernetes-ux.md
@@ -32,6 +32,9 @@ assuming:
 The `ownerReference` of the `MeshService` points to the `Service`.
 Labels are copied from the `Service` object to the `MeshService` object.
 
+If a `MeshService` object exists in the namespace with the same name, the
+`MeshService` won't be created and an error is bubbled up to the user.
+
 ### VIP
 
 For `MeshServices` in Kubernetes zones, we no longer create a VIP but instead
@@ -76,8 +79,8 @@ spec:
 
 Note that Kubernetes does not allocate a `ClusterIP` for headless services, it
 only creates a round-robin DNS record to point to PodIPs. Kuma does not
-allocate a VIP either. In a zone, users can rely on kube-dns. Cross-zone
-behavior will be covered in a subsequent MADR.
+allocate a VIP either. In a zone, users can rely on kube-dns. This record or
+behavior won't be exposed in anyway cross-zone.
 
 #### Policy matching
 
--- a/docs/madr/decisions/041-meshservice-kubernetes-ux.md
+++ b/docs/madr/decisions/041-meshservice-kubernetes-ux.md
@@ -66,15 +66,16 @@ In order to do this we need to have a list of all the Pods selected by the
 Service, which we can get by looking at `EndpointSlices`. These resources hold a
 list of endpoints, each of which has a `targetRef`. If the `targetRef` is `kind:
 Pod`, we can rely on the naming of `Dataplane` objects and directly select a
-given `Dataplane` by setting `spec.selector.dataplaneName` to the name of the
+given `Dataplane` by setting `spec.selector.dataplane.name` to the name of the
 `Pod`.
 
 ```
 kind: MeshService
 spec:
   selector:
-    dataplaneName: pod-1
-    # dataplaneTags: ...
+    dataplane:
+      name: pod-1
+      # tags: ...
 ```
 
 Note that Kubernetes does not allocate a `ClusterIP` for headless services, it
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,64 @@
...
+For `MeshServices` in Kubernetes zones, we no longer create a VIP but instead
+reuse the `ClusterIP` field allocated by the Kubernetes control plane and add
+this to `status.vips` on `MeshService`.
+
+Note this happens asynchronously.
+
+### `ports`
+
+The `MeshService` ports are derived from the `ports` field on `Service`. Note
+that `meshservice_api.Port.Protocol` is derived from `ServicePort.AppProtocol`.
```
lahabana, line 45: We said we'd use AppProtocol too to avoid confusion

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,64 @@
...
+assuming:
+
+1. The namespace is part of the mesh with `kuma.io/sidecar-injection: "true"`
+1. The `Mesh` of the `Service` or its namespace exists.
+1. The `Service` hasn't opted out via `kuma.io/ignore: "true"`
+
+### Metadata
+
+The `ownerReference` of the `MeshService` points to the `Service`.
+Labels are copied from the `Service` object to the `MeshService` object.
```
lahabana, line 32: What happens with conflicts?

michaelbeaumont: Conflicts where?

michaelbeaumont: Should we also copy annotations?

jakubdyszkiewicz: I'd avoid this for now if we don't have to do this

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,93 @@
...
+
+* CP manages `MeshService` from `Service`
+* Users have to create `MeshService`
+
+## Decision Outcome
+
+We will convert `Services` to `MeshServices` in a Kubernetes controller,
+assuming:
+
+1. The namespace is part of the mesh with `kuma.io/sidecar-injection: "true"`
```
jakubdyszkiewicz, line 25: or Service is labeled with `kuma.io/mesh`. I'd like to still give a possibility for users to opt in only one service/pod in a namespace.

WDYT?

michaelbeaumont: Yeah makes sense.

docs/madr/decisions/041-meshservice-kubernetes-ux.md
lahabana: No I meant if by some random reason there's already a MeshService with this name

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,94 @@
...
+given `Dataplane` by setting `spec.selector.dataplaneName` to the name of the
+`Pod`.
+
+```
+kind: MeshService
+spec:
+  selector:
+    dataplaneName: pod-1
+    # dataplaneTags: ...
+```
```
lahabana, line 73: Is there still a service that covers all pods? Similar to clusterIP (I can't remember what happens in this case...)

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,94 @@
...
+StatefulSet. In another MADR, we will cover this use case.
+
+### Positive Consequences
+
+* Users don't have to think about creating `MeshService`
+* We don't need to allocate additional VIPs
+
+### Negative Consequences
+
+* None?
```
lahabana, line 87: Many MeshService when using headless services a lot

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,94 @@
...
+
+Note this happens asynchronously.
+
+### `ports`
+
+The `MeshService` ports are derived from the `ports` field on `Service`,
+including supporting named `targetPorts`.
+
+Note that we only support `Service.ports[].protocol: TCP`, which is also the
+default.
```
lahabana, line 49: We're not talking about appProtocol should we mention it's copied as is

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,94 @@
...
+* CP manages `MeshService` from `Service`
+* Users have to create `MeshService`
+
+## Decision Outcome
+
+We will convert `Services` to `MeshServices` in a Kubernetes controller,
+assuming:
+
+1. The namespace is part of the mesh with `kuma.io/sidecar-injection: "true"`
+  or the `Service` is labelled with `kuma.io/mesh`.
```
lobkovilya, line 26: We should consider having the same label for all k8s objects to enable the mesh. Today namespaces and pods need `kuma.io/sidecar-injection` to enable sidecars. It can be surprising for users if the service needs only `kuma.io/mesh` to get sidecars

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,94 @@
...
+
+Note that we only support `Service.ports[].protocol: TCP`, which is also the
+default.
+
+### Headless Service with selectors
+
+In Kubernetes, a headless Service with selectors is used to create a DNS record
+for every Pod selected by the Service that points directly to the Pod's IP.
+
+To support this with Kuma, we will create a `MeshService` per Pod, each
```
lobkovilya, line 56: What's the MeshService `name` in this case?

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,94 @@
...
+list of endpoints, each of which has a `targetRef`. If the `targetRef` is `kind:
+Pod`, we can rely on the naming of `Dataplane` objects and directly select a
+given `Dataplane` by setting `spec.selector.dataplaneName` to the name of the
+`Pod`.
+
+```
+kind: MeshService
+spec:
+  selector:
+    dataplaneName: pod-1
```
lobkovilya, line 71: why can't we use [kuma.io/instance](https://github.com/kumahq/kuma/blob/17025b30b70ad353a772172eeeadfc8496d9c41f/api/mesh/v1alpha1/dataplane_helpers.go#L35) tag that's already set to the pod's name on k8s?

michaelbeaumont: We might be able to do this. IMO it seems better to be explicit about the fact that we're selecting _exactly one_ `Dataplane`. I also don't like that this is _only_ set if the Service is headless at the moment, though we can change that.

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: Yeah I agree but `kuma.io/sidecar-injection` fits so badly. @jakubdyszkiewicz @lahabana opinions?

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: Yeah maybe I'll add it back. I removed it because it's `appProtocol` -> `appProtocol` but I suppose once we mention `protocol` we should mention `appProtocol`.

docs/madr/decisions/041-meshservice-kubernetes-ux.md
jakubdyszkiewicz: > It can be surprising for users if the service needs only kuma.io/mesh to get sidecars

They don't enable sidecars this way. They opt-in a Service to a mesh.

Let's also keep in mind that it's probably 10-20% of use cases. The majority of time people are placing sidecar-injection on the namespace and then disable mesh on specific pods if needed.

docs/madr/decisions/041-meshservice-kubernetes-ux.md
Automaat, line 1: can we add one bigger example of `Service` and how created `MeshService` would look like?

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: IMO it's the same answer to the same question but for `Dataplanes`

docs/madr/decisions/041-meshservice-kubernetes-ux.md
slonka: Would like to see that too.

slonka (commented):
Not really to much to add, would be good to see an example as mentioned by Marcin.

docs/madr/decisions/041-meshservice-kubernetes-ux.md
lobkovilya: > They don't enable sidecars this way. They opt-in a Service to a mesh.

I see, so the `kuma.io/mesh` label just means we're going to convert the Service to a MeshService. But what's the usecase for that, why'd you want to introduce a Service to a mesh while all pods implementing the Service don't have a sidecar? Is it something related to ExternalName?

jakubdyszkiewicz: The use case is that there needs to be a mechanism to convert Service to MeshService for a corresponding Dataplane object that has sidecar-injection.

We could retrieve Pods in Service Controller to check this, but the controller would have to have a dependency on pods now which we want to avoid.

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: I think we'll need to include the pod name via a hash suffix.

michaelbeaumont: added to MADR

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: @lahabana Added a note about this, WDYT? I think most interesting is probably the behavior of headless services cross-zone but a more comprehensive MADR on cross-zone behavior is coming.

slonka (approved):
It's really hard to do another review when it's force pushed. I see you added the example (the thing that I mentioned in the previous review).

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,169 @@
...
+  tls:
+    state: Ready
+    issuedBackends:
+      ca-1: 5
+  proxies:
+    offline: 3
+    online: 5
+  vips:
+  - ip: 192.168.10.1
+    type: Kubernetes
```
lahabana, line 152: Do we need conditions? I think it would be nice from a user PoV

docs/madr/decisions/041-meshservice-kubernetes-ux.md
jakubdyszkiewicz: I like the explicit thing, because it guarantees that it's only one DPP although `kuma.io/instance` works for me as well

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: Yeah let's add conditions

docs/madr/decisions/041-meshservice-kubernetes-ux.md
lahabana: I'm fine with the error. Is this an event on the service or the meshService?

lahabana: Also do we merge the labels even if the ownerReference don't match?

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,169 @@
...
+
+Note this happens asynchronously.
+
+### `ports`
+
+The `MeshService` ports are derived from the `ports` field on `Service`,
+including supporting named `targetPorts`.
+
+Note that we only support `Service.ports[].protocol: TCP`, which is also the
+default. We copy `appProtocol` to the `MeshService` port entry as well.
```
lahabana, line 52: What happens if a service has another protocol? Can we send an event?

docs/madr/decisions/041-meshservice-kubernetes-ux.md
lobkovilya: Ok, I don't mind doing it explicitly here. 

Do we know how this selector is going to evolve in the future (i.e. when we modernize the Dataplane resource)? Maybe it makes sense to create a `dataplane` section like:
```yaml
selector:
  dataplane:
    name: dppName
    tags: {...}
```
so that later we could do `selector.meshDpp` or something like this

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,169 @@
...
+### Headless Service with selectors
+
+In Kubernetes, a headless Service with selectors is used to create a DNS record
+for every Pod selected by the Service that points directly to the Pod's IP.
+
+To support this with Kuma, we will create a `MeshService` per Pod, each
+represented by the hostname allocated by the headless Service and the Pod
+IP as the "VIP" and single endpoint.
+The name of said `MeshServices` has the format `<service-name>-<pod-hash>`
+where the `Service` name is appropriately truncated.
```
lahabana, line 63: I guess we could just use the index `<serviceName>-<index>` this would be an example where a collision would happen:

- I create a service with clusterIP: `kafka-broker-0`
- I create a headless service with name `kafka-broker`

Then I have 2 meshService `kafka-broker-0` first one wins, the other gets an error. Is it an event on the MeshService on the Service?

docs/madr/decisions/041-meshservice-kubernetes-ux.md
lahabana: Should we have a label `k8s.kuma.io/serviceName` so we can easily filter `MeshService`s behind a `Service` ?

docs/madr/decisions/041-meshservice-kubernetes-ux.md
lahabana: Ok makes sense we should really talk about it cross-zone for sure!

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,169 @@
...
+
+Note that Kubernetes does not allocate a `ClusterIP` for headless services, it
+only creates a round-robin DNS record to point to PodIPs. Kuma does not
+allocate a VIP either. In a zone, users can rely on kube-dns. This record or
+behavior won't be exposed in anyway cross-zone.
+
+#### Policy matching
+
+Note that this prevents using `kind: MeshService` to select all Pods of a
+StatefulSet. In another MADR, we will cover this use case.
```
lahabana, line 88: This MADR is `MeshServiceGroup` or something like this right? Do we have an issue to cover this use case?

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,169 @@
...
+    issuedBackends:
+      ca-1: 5
+  proxies:
+    offline: 3
+    online: 5
+  vips:
+  - ip: 192.168.10.1
+    type: Kubernetes
+```
+
```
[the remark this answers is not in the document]
michaelbeaumont: I think events is kind of an open set but I'll document in this MADR everywhere where we do create an event.

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: I think nothing happens, as it is now the port is essentially ignored/continues to work via passthrough. You mean an event noting that the port is being skipped? Yeah we can do that

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: On the Service I would say, the user likely doesn't know that the MeshService exists and would first look at the Service.

michaelbeaumont: > Also do we merge the labels even if the ownerReference don't match?

Is this a different case than the MeshService existing? How exactly would you get to this state otherwise? If the MeshService exists, we don't overwrite it and we don't touch the labels.

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: @lobkovilya @lahabana please take another look

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: The problem with this is that not every Pod matched by a headless service comes from a StatefulSet/has an index.

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: @lahabana I'm going to add these in a separate PR to the original MADR since they have nothing to do with Kubernetes

michaelbeaumont: https://github.com/kumahq/kuma/issues/10057

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: https://github.com/kumahq/kuma/issues/9707

docs/madr/decisions/041-meshservice-kubernetes-ux.md
lahabana: No I think are right

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,177 @@
...
+### `ports`
+
+The `MeshService` ports are derived from the `ports` field on `Service`,
+including supporting named `targetPorts`.
+
+Note that we only support `Service.ports[].protocol: TCP`, which is also the
+default. We copy `appProtocol` to the `MeshService` port entry as well.
+
+We add an event to the `Service` to indicate that there are ports not supported
```
lahabana, line 57: Maybe we don't need such event. I don't see that being very usable as an event

docs/madr/decisions/041-meshservice-kubernetes-ux.md
lahabana: How do headless service work in that case?

docs/madr/decisions/041-meshservice-kubernetes-ux.md
```diff
@@ -0,0 +1,177 @@
...
+kind: MeshService
+metadata:
+  name: redis
+  namespace: redis-system
+  labels:
+    team: db-operators
+    kuma.io/mesh: default
+spec:
+  selector:
+    dataplane:
```
lahabana, line 134: nit: I think the spec is `dataplaneTags`

lahabana (approved):
I'm still not very convinced by the headless service implem...

The rest looks great though!

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: This alternative was suggested above https://github.com/kumahq/kuma/pull/9722#discussion_r1574691093

michaelbeaumont: @lahabana gonna merge, let me know if this is a problem

docs/madr/decisions/041-meshservice-kubernetes-ux.md
michaelbeaumont: @lahabana gonna merge, let me know if you have more on this

Issue #8922, opened 2024-01-19
Use `MeshHTTPRoute` everywhere for Gateway API


michaelbeaumont:
### Description

Currently only used for GAMMA, we can use it for gateways with https://github.com/kumahq/kuma/issues/8142

> Kuma includes controllers that reconcile Gateway API CRDs and convert them into the corresponding Kuma CRDs. This is why in the GUI, Kuma MeshGateways/MeshGatewayRoutes/MeshHTTPRoutes/MeshTCPRoutes are visible and not Kubernetes Gateway API resources.
> ref. https://kuma.io/docs/2.6.x/using-mesh/managing-ingress-traffic/gateway-api/#how-it-works

- [ ] Remove webhook check when creating Gateway API resources that they're "only supported in a single-zone"

PR #9732, opened 2024-03-27
feat(gatewayapi): use MeshHTTPRoutes instead of MeshGatewayRoutes internally

bartsmykla:
### Checklist prior to review

  - Closes https://github.com/kumahq/kuma/issues/8922
  - It won't
  - Unit tests updated
  - Existing e2e (including GatewayAPI conformance) tests passes
- [ ] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)?
  - There is no need


pkg/plugins/runtime/k8s/controllers/gatewayapi/http_route_controller.go
```diff
@@ -105,148 +112,165 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req kube_ctrl.Reque
 
 type ParentConditions map[gatewayapi.ParentReference][]kube_meta.Condition
 
+func generateMeshHTTPRouteName(route *gatewayapi.HTTPRoute, parentRef gatewayapi.ParentReference) string {
+	hash := fnv.New32()
+
+	parentRefElems := slices.DeleteFunc(
```
michaelbeaumont, line 118: I think these can just go in the hash without filtering them right?

bartsmykla: You are right, this is leftover when `parentRefElems` were used to generate name. Removed filtering logic

michaelbeaumont (approved):

PR #9742, opened 2024-03-28
fix(kds): fix memory leak on kds error

Automaat:
Goroutines were hanging on writing to channel that no one was listening anymore

### Checklist prior to review


pkg/util/channels/nonblocking.go
```diff
@@ -0,0 +1,11 @@
+package channels
+
+import "github.com/labstack/gommon/log"
```
jakubdyszkiewicz, line 3: we don't use this logger. I think you need to pass a logger to this function because it can contain extra useful key value pairs

pkg/util/channels/nonblocking.go
```diff
@@ -0,0 +1,11 @@
...
+
+func NonBlockingErrorWrite(ch chan error, err error) {
+	select {
+	case ch <- err:
+	default:
+		log.Error(err, "Failed to write error to closed channel")
```
jakubdyszkiewicz, line 9: nit: lowercase https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/003-logging.md#convention

pkg/kds/global/components.go
```diff
@@ -115,16 +116,17 @@ func Setup(rt runtime.Runtime) error {
...
-			errChan <- err
+			channels.NonBlockingErrorWrite(errChan, err)
+			return
 		}
 		log := kdsDeltaGlobalLog.WithValues("peer-id", zoneID)
 		log = kuma_log.AddFieldsFromCtx(log, stream.Context(), rt.Extensions())
 		log.Info("Global To Zone new session created")
 		if err := createZoneIfAbsent(stream.Context(), log, zoneID, rt.ResourceManager()); err != nil {
-			errChan <- errors.Wrap(err, "Global CP could not create a zone")
+			channels.NonBlockingErrorWrite(errChan, errors.Wrap(err, "Global CP could not create a zone"))
```
jakubdyszkiewicz, line 126: shouldn't we return after this?

pkg/util/channels/nonblocking.go
```diff
@@ -0,0 +1,11 @@
+package channels
+
+import "github.com/labstack/gommon/log"
+
+func NonBlockingErrorWrite(ch chan error, err error) {
```
michaelbeaumont, line 5: Might this just as well be generic?

pkg/kds/global/components.go
```diff
@@ -115,16 +116,17 @@ func Setup(rt runtime.Runtime) error {
 	onGlobalToZoneSyncConnect := mux.OnGlobalToZoneSyncConnectFunc(func(stream mesh_proto.KDSSyncService_GlobalToZoneSyncServer, errChan chan error) {
 		zoneID, err := util.ClientIDFromIncomingCtx(stream.Context())
 		if err != nil {
-			errChan <- err
+			channels.NonBlockingErrorWrite(errChan, err)
```
lahabana, line 119: This is confusing I find.

Folks used to go understand the pattern:

```go
select {
case <-channel:
  // Do something
default:
  // Do something
```

I think we should just keep it as is.
A more generic question is how come we have this issue in the first place?

Automaat: we are writing to channel [here](https://github.com/kumahq/kuma/blob/ee20a34318dcd497070c2dbdf55cac932260fe0e/pkg/kds/global/components.go#L127) and none is listening on this channel because we've caught context cancelled and exited [here](https://github.com/kumahq/kuma/blob/ee20a34318dcd497070c2dbdf55cac932260fe0e/pkg/kds/mux/zone_sync.go#L88)

pkg/util/channels/nonblocking.go
Automaat: changed

pkg/kds/global/components.go
Automaat: added return

pkg/util/channels/nonblocking.go
Automaat: no longer apply

pkg/kds/global/components.go
Automaat: moved out of util function

pkg/util/channels/nonblocking.go
Automaat: I was too quick to commit and didn't look at it, shuld be fine now

pkg/kds/global/components.go
michaelbeaumont: maybe instead of `default:` we should check if the context is canceled?

Automaat: > maybe instead of default: we should check if the context is canceled?

but there still can be a race, we can check and see that context was not cancelled, and it can be cancelled in the meantime and still write to channel that noone is listening

jakubdyszkiewicz (approved):

pkg/kds/global/components.go
michaelbeaumont: indeed. the only way to be sure is for the reader to communicate back to the writer that it's no longer listening.

pkg/kds/global/components.go
```diff
@@ -133,7 +147,11 @@ func Setup(rt runtime.Runtime) error {
 	onZoneToGlobalSyncConnect := mux.OnZoneToGlobalSyncConnectFunc(func(stream mesh_proto.KDSSyncService_ZoneToGlobalSyncServer, errChan chan error) {
 		zoneID, err := util.ClientIDFromIncomingCtx(stream.Context())
 		if err != nil {
-			errChan <- err
+			select {
+			case errChan <- err:
+			default:
```
michaelbeaumont, line 152: why are we sometimes dropping the error and sometimes logging it?

pkg/kds/global/components.go
```diff
@@ -115,16 +115,30 @@ func Setup(rt runtime.Runtime) error {
...
 		}
 		log := kdsDeltaGlobalLog.WithValues("peer-id", zoneID)
 		log = kuma_log.AddFieldsFromCtx(log, stream.Context(), rt.Extensions())
 		log.Info("Global To Zone new session created")
 		if err := createZoneIfAbsent(stream.Context(), log, zoneID, rt.ResourceManager()); err != nil {
-			errChan <- errors.Wrap(err, "Global CP could not create a zone")
+			select {
+			case errChan <- err:
+			default:
+				log.Error(errors.Wrap(err, "Global CP could not create a zone"), "failed to write error to closed channel")
```
michaelbeaumont, line 131: shouldn't the wrapped error be sent to the channel?

Automaat: correct, forgot to change it

pkg/kds/global/components.go
Automaat: I thought I was missing logger, I've added logging

pkg/kds/global/components.go
Automaat: isn't it overengineering in this case? Isn't just non-blocking send of error (fire and forget) enough?

michaelbeaumont: yeah sorry, I wasn't saying we should do it. just thinking it through

Issue #9458, opened 2024-02-29
ZoneIngress doesn't have listener if there are no services in the mesh


lahabana:
### Description

While this might not seem to be a problem it makes Load balancer healthchecks for zone ingresses fail which is poor UX.

It should also be quite simple to fix.

Issue #8469, opened 2023-11-27
Review naming of Envoy entities


lahabana:
### Description

We currently don't have a unified and standard way to name things in our XDS config this has led to a few issues:

- https://github.com/kumahq/kuma/issues/2363
- https://github.com/kumahq/kuma/issues/3249

We should write how we want to do this in the future and how things are exposed as APIs to avoid folks making up new ones.

Also should we have tags in cluster subsets somewhere in the metadata? 

How do we roll this out from a breaking change POV?

lahabana:
Routes also don't have good names. Also in matching API the hash is not related to the one in the resource name. Finally is there anyway to add metadata to make these easier to map in the xds conf?

lahabana:
Listeners are standardised: https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/036-internal-listeners.md (we still need to migrate our existing APIs over).

We should have similar MADRs for:

- clusters (`_` prefix for internal, what about name + hash? )
- routes (MeshHTTPRoute, MeshTCPRoute) would be good to have names on these.

anything else?

lobkovilya:
Covered by https://github.com/kumahq/kuma/issues/13256

Issue #1907, opened 2021-04-29
Health checks and dashboard for connectivity between zones


jakubdyszkiewicz:
## Summary

Right now we have pretty good observability of a connection from Remote to Global. You can see easily see in Global that Remote is online or not.

The problem is that once everything zone is connected to global, we don't provide any tools to check connectivity from one zone to another.

### Observability

I'd like to have a dashboard in the GUI to see connectivity between zones. Something like this

![image](https://user-images.githubusercontent.com/5459824/116520706-5acecc00-a8d3-11eb-8776-a24e72eb40d5.png)

where there is a separate Zone on each end. If the connection is up, the line is green, if not the line is red.

(It also can be just a simple table with every zone in column and every zone in row)

### Traffic reliability

If Global <-> Remote communication works, but Zone <-> Zone does not we still include Ingress endpoint into EDS. Of course, Envoy can exclude this endpoint with Health Checks / Circuit Breaker + Retries, but we should do this beforehand.

If Remote A knows that it cannot connect to Ingress of Remote B, it should not include this in the EDS.

## Overview of potential implementation

* Build a component that runs on Remote CP (leader) that healthchecks (just open TCP connection) Ingress from other zones and save this information in Config called `zone-healthcheck`
* When building XDS config, check `zone-healthcheck` Config and exclude ingress if we see that HC failed
* Propagate `zone-healthcheck` from every Remote CP to Global CP
* Build an API based on all `zone-healthcheck` from every zone
* Build a dashboard based on this API

## Considerations

* Zone can have 2 ingresses, one can fail, the other don't. Probably it would be good to have "partially degrated" status just to follow the same pattern as with DP health checks.

lahabana:
This ticket will need to be split into a milestone and this only tracks the design of it.
The proposed design/idea from @jakubdyszkiewicz probably needs to be adapted with the creation of ZoneEgress.

mohamedawnallah:
Hey @jakubdyszkiewicz @lahabana, I'm excited to take on this issue! Could you kindly provide me with some guidance and details on how to get started? Thanks!

slonka:
hi @mohamedawnallah I'm moving this back to triage to discuss this on the next meeting and we'll get back to you with more details.

lobkovilya:
> Triage: please reopen if needed

PR #9745, opened 2024-03-28
feat(xds): generate an empty direct response listener for empty zone ingress gateway

jijiechen:
fixes https://github.com/kumahq/kuma/issues/9458

### Checklist prior to review

  - https://github.com/kumahq/kuma/issues/9458
  - Confirmed
  - Done
  - No need
  - No need


pkg/xds/generator/ingress_generator.go
```diff
@@ -70,13 +71,26 @@ func (i IngressGenerator) Generate(
 	if err != nil {
 		return nil, err
 	}
-	if len(listener.(*envoy_listener_v3.Listener).FilterChains) > 0 {
+
+	hasFilterChain := len(listener.(*envoy_listener_v3.Listener).FilterChains) > 0
+	if !hasFilterChain {
+		var blackHoleCluster envoy_common.NamedResource
+		listener, blackHoleCluster, err = zoneproxy.GenerateBlackHoleResources(proxy.APIVersion, address, port)
```
lahabana, line 78: Thinking about this in the context of: https://github.com/kumahq/kuma/issues/1907

Would it make sense to use: https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/direct_response_filter this way we don't add a cluster we don't need and can also implement healthchecks quite easily.

WDYT?

pkg/xds/generator/zoneproxy/generator.go
```diff
@@ -172,3 +172,45 @@ func AddFilterChains(
 
 	return servicesAcc.Services()
 }
+
+const blackHoleClusterName = "kuma:blackhole"
```
lahabana, line 176: This should like go to `pkg/xds/envoy/names/resource_names.go`.

Also in light of: https://github.com/kumahq/kuma/issues/8469 and the related issues. We should not introduce more names with `:`. @Automaat created a MADR for listeners: https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/036-internal-listeners.md . I think we should have a similar one for cluster names.
I think we should have `_blackhole` here which is good enough.

pkg/xds/generator/ingress_generator.go
jijiechen: What should we send into the response then?

pkg/xds/generator/zoneproxy/generator.go
jijiechen: Agreed, we definitely need a unified naming convension. I'll take a look at #8469
I introduced the colon because we had done so for previous listeners like passthrough ones and a colon simply prevent duplication to user defined services.

pkg/xds/generator/ingress_generator.go
lahabana: Hmm could be a json with zoneIngress name and zone. @jakubdyszkiewicz any idea?

pkg/xds/envoy/listeners/v3/direct_response_configurer_test.go
```diff
@@ -50,4 +50,24 @@ filters:
...
+
+		// then
+		actual, err := util_proto.ToYAML(filterChain)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(actual).To(MatchYAML(`filters:
+          - name: envoy.filters.network.direct_response
+            typedConfig:
+              '@type': type.googleapis.com/envoy.extensions.filters.network.direct_response.v3.Config
+              response:
+                  inlineBytes: eyJqc29uIjoidmFsdWUifQ==`))
```
slonka, line 71: decodes ok

slonka (approved):
LGTM

PR #9748, opened 2024-03-28
fix(kds): skip error on context cancelled

Automaat:
We are changing this because with context cancelled we will log it as error, that is irrelevant since we are logging it on info level later on

### Checklist prior to review

> Changelog: skip

### DIFF
--- a/pkg/kds/global/components.go
+++ b/pkg/kds/global/components.go
@@ -114,7 +114,7 @@ func Setup(rt runtime.Runtime) error {
 
 	onGlobalToZoneSyncConnect := mux.OnGlobalToZoneSyncConnectFunc(func(stream mesh_proto.KDSSyncService_GlobalToZoneSyncServer, errChan chan error) {
 		zoneID, err := util.ClientIDFromIncomingCtx(stream.Context())
-		if err != nil && !errors.Is(err, context.Canceled) {
+		if err != nil {
 			select {
 			case errChan <- err:
 			default:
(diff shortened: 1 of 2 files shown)


pkg/kds/global/components.go
```diff
@@ -114,7 +114,7 @@ func Setup(rt runtime.Runtime) error {
 
 	onGlobalToZoneSyncConnect := mux.OnGlobalToZoneSyncConnectFunc(func(stream mesh_proto.KDSSyncService_GlobalToZoneSyncServer, errChan chan error) {
 		zoneID, err := util.ClientIDFromIncomingCtx(stream.Context())
-		if err != nil {
+		if err != nil && !errors.Is(err, context.Canceled) {
```
jakubdyszkiewicz, line 117: I think it's not fully correct. Here for example we won't return on context cancelled error from `util.ClientIDFromIncomingCtx`

Automaat: you are right, removed. As for the rest changes in this review, there could be context cancelled error returned

pkg/kds/global/components.go
```diff
@@ -134,7 +137,9 @@ func Setup(rt runtime.Runtime) error {
 			}
 			return
 		}
-		if err := kdsServerV2.GlobalToZoneSync(stream); err != nil {
+
+		err = kdsServerV2.GlobalToZoneSync(stream)
+		if err != nil && !errors.Is(err, context.Canceled) {
```
slonka, line 142: some of them could be inlined - that would make me a bit less worried that we're overriding some other err

slonka (commented):
In general why do we need this? Was this causing some side effect or only to have cleaner logs?

jakubdyszkiewicz (approved):

pkg/kds/global/components.go
Automaat: changed

slonka (approved):

PR #9749, opened 2024-03-28
feat(kds): stats of kds client versions

jakubdyszkiewicz:
### Checklist prior to review

This PR adds stats of number of clients per version. For now it only works with Delta KDS, but I added this to generic stats callbacks so in the future we can add this to other DSes if we want to.

I considered just adding a dimension to `_streams_active` metric, but streams active counts all DS streams, whereas this metric only counts streams with at least one DiscoveryRequest, because version is carried in metadata of request. This should be equal, but technically _can_ be different.

### DIFF
--- a/pkg/util/xds/stats_callbacks.go
+++ b/pkg/util/xds/stats_callbacks.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/kumahq/kuma/pkg/core"
 	core_metrics "github.com/kumahq/kuma/pkg/metrics"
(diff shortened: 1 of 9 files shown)


pkg/util/xds/stats_callbacks.go
```diff
@@ -31,10 +38,13 @@ type statsCallbacks struct {
 	NoopCallbacks
 	responsesSentMetric    *prometheus.CounterVec
 	requestsReceivedMetric *prometheus.CounterVec
+	versions               *prometheus.GaugeVec
```
lobkovilya, line 41: nit: probably makes sense to name it `versionsMetric` to be aligned with other fields of this struct

pkg/util/xds/stats_callbacks.go
```diff
@@ -114,17 +134,30 @@ func (s *statsCallbacks) OnStreamOpen(context.Context, int64, string) error {
 	return nil
 }
 
-func (s *statsCallbacks) OnStreamClosed(int64) {
+func (s *statsCallbacks) OnStreamClosed(streamID int64) {
```
lobkovilya, line 137: Do you know if these OnClosed methods are called during the graceful CP shutdown? It'd be not ideal to have stale metrics after each CP rollout

jakubdyszkiewicz: What do you mean? Even if OnStreamClosed is not closed, the CP goes down, so it won't report this metric.

lobkovilya (approved):

Issue #8413, opened 2023-11-20
[Transparent Proxy] Implement handling of different types of `iptables`


bartsmykla:
### Description

In [our docs](https://github.com/kumahq/kuma-website/blob/master/app/_src/production/dp-config/transparent-proxying.md?plain=1#L39-L54) we are suggesting users using the latest Ubuntu to run:

```sh
update-alternatives --set iptables /usr/sbin/iptables-legacy
update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
```

1. I think this suggestion is incorrect and this is unnecessary
  - After some research I don't see any feature we are using which would force us to use `iptables-legacy` and not `iptables-nft`
  - I tested on a VM with the latest Ubuntu and `kumactl install transparent-proxy` works with the default `iptables-nft`
  - I think this incorrect instruction was added as a result of confusion created by https://github.com/kumahq/kuma/issues/8412
2. I think we should add a CLI flag (i.e. `--iptables-mode {nft,legacy}` or `--iptables-binary {iptables-nft, iftables-legacy, /sbin/iptables-nft [...]}` where we could enforce usage of the legacy or nft modes.
  - The default behavior should be to automatically figure out the proper mode

michaelbeaumont:
Can we use the k8s upstream [iptables-wrappers](https://github.com/kubernetes-sigs/iptables-wrappers)?

PR #9750, opened 2024-03-28
feat(transparent-proxy): add automatic iptables type detection

bartsmykla:
### Checklist prior to review

  - Closes https://github.com/kumahq/kuma/issues/8413
  - It won't
  - There is no need


pkg/transparentproxy/iptables/builder/builder.go
```diff
@@ -240,13 +216,13 @@ func RestoreIPTables(cfg config.Config) (string, error) {
 		}
 	}
 
-	output, err := restoreIPTables(cfg, dnsIpv4, false)
+	output, err := restoreIPTables(context.TODO(), cfg, dnsIpv4, false)
```
lahabana, line 219: Can we wire the ctx through shouldn't be that complicated...

michaelbeaumont (approved):

pkg/transparentproxy/iptables/builder/builder.go
bartsmykla: done

PR #9767, opened 2024-03-29
feat(components): exponential backoff for resilient components

jakubdyszkiewicz:
### Checklist prior to review

Introduce exponential backoff when resilient component fails.
Components that are using resilient component:
* KDS
* OTEL
* Postgres events
* Zone watcher
* Insights resync

In all cases, it makes sense to have exponential backoff because those components relies on external systems - API/DB etc.

I tested it manually.

### DIFF
--- a/pkg/core/runtime/component/resilient.go
+++ b/pkg/core/runtime/component/resilient.go
@@ -5,10 +5,12 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	"github.com/sethvargo/go-retry"
 )
 
 const (
-	backoffTime = 5 * time.Second
+	backoffBaseTime = 5 * time.Second
+	backoffMaxTime  = 1 * time.Minute
 )
 
 type resilientComponent struct {
@@ -25,6 +27,7 @@ func NewResilientComponent(log logr.Logger, component Component) Component {
 
 func (r *resilientComponent) Start(stop <-chan struct{}) error {
 	r.log.Info("starting resilient component ...")
+	backoff := retry.WithCappedDuration(backoffMaxTime, retry.NewExponential(backoffBaseTime))
 	for generationID := uint64(1); ; generationID++ {
 		errCh := make(chan error, 1)
 		go func(errCh chan<- error) {
@@ -51,7 +54,8 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 				r.log.WithValues("generationID", generationID).Error(err, "component terminated with an error")
 			}
 		}
-		<-time.After(backoffTime)
+		dur, _ := backoff.Next()
+		<-time.After(dur)
 	}
 }
 
--- a/pkg/core/runtime/component/resilient.go
+++ b/pkg/core/runtime/component/resilient.go
@@ -29,6 +29,7 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 	r.log.Info("starting resilient component ...")
 	backoff := newBackoff()
 	for generationID := uint64(1); ; generationID++ {
+		lastStart := time.Now()
 		errCh := make(chan error, 1)
 		go func(errCh chan<- error) {
 			defer close(errCh)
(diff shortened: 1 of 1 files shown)


slonka (commented):
A bit scary to do a change like this that touches so many parts of the system without any tests changing 🤔 Are we sure there aren't any resilient components in parent project that could be impacted by this?

jakubdyszkiewicz:
There are, but by principle those are also I/O, so they can fail the same way - an external system is not available.

pkg/core/runtime/component/resilient.go
```diff
@@ -5,10 +5,12 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	"github.com/sethvargo/go-retry"
 )
 
 const (
-	backoffTime = 5 * time.Second
+	backoffBaseTime = 5 * time.Second
+	backoffMaxTime  = 1 * time.Minute
```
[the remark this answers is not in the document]
jakubdyszkiewicz: We could, but why is KDS special?
If we go with 5mins for KDS, why 5mins is not ok, but 1min is ok for insight resync for example?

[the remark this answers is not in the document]
slonka: I think that we're the most exposed to KDS errors via parent project hosted platform, so it's just for convenience. [Redis's max backoff](https://cloud.google.com/memorystore/docs/redis/exponential-backoff) is 64 seconds so 1m is close enough.

pkg/core/runtime/component/resilient.go
```diff
@@ -51,7 +54,8 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 				r.log.WithValues("generationID", generationID).Error(err, "component terminated with an error")
 			}
 		}
-		<-time.After(backoffTime)
+		dur, _ := backoff.Next()
```
lahabana, line 57: Do we need to reset this backoff if we've worked fine for a while? I'd say if the component stays up for more than the max backoff reseting the backoff sounds good.

pkg/core/runtime/component/resilient.go
```diff
@@ -25,6 +27,7 @@ func NewResilientComponent(log logr.Logger, component Component) Component {
 
 func (r *resilientComponent) Start(stop <-chan struct{}) error {
 	r.log.Info("starting resilient component ...")
+	backoff := retry.WithCappedDuration(backoffMaxTime, retry.NewExponential(backoffBaseTime))
```
lahabana, line 30:
```suggestion
	backoff := retry.WithJitter(retry.WithCappedDuration(backoffMaxTime, retry.NewExponential(backoffBaseTime)), backoffBaseTime)
```

Can't hurt no?

pkg/core/runtime/component/resilient.go
```diff
@@ -42,23 +42,38 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
...
+			case err := <-errCh:
+				if err != nil {
+					r.log.WithValues("generationID", generationID).Error(err, "component terminated with an error")
+				}
+				receivedErr = true
+			case <-time.After(backoffMaxTime):
+				// reset backoff so in a case of events
+				// 1) Component is unhealthy until max backoff is reached
+				// 2) Component is healthy for at least backoff max time
+				// 3) Component is unhealthy
```
slonka, line 63: I don't understand 1 and 3 - IMO we should only reset backoff when the component is healthy again. Also isn't 3 just superset of 1 so we don't need 1?

jakubdyszkiewicz: it's steps. First step then second step then third

pkg/core/runtime/component/resilient.go
```diff
@@ -25,6 +27,7 @@ func NewResilientComponent(log logr.Logger, component Component) Component {
 
 func (r *resilientComponent) Start(stop <-chan struct{}) error {
 	r.log.Info("starting resilient component ...")
+	backoff := newBackoff()
 	for generationID := uint64(1); ; generationID++ {
 		errCh := make(chan error, 1)
```
lahabana, line 32: lastStart := time.Now()

pkg/core/runtime/component/resilient.go
```diff
@@ -39,22 +42,38 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
...
-
 			errCh <- r.component.Start(stop)
 		}(errCh)
-		select {
-		case <-stop:
-			r.log.Info("done")
-			return nil
-		case err := <-errCh:
-			if err != nil {
-				r.log.WithValues("generationID", generationID).Error(err, "component terminated with an error")
```
lahabana, line 51: if lastStart.Sub(time.Now)) > backoffMaxTime {
   backoff = newBackoff()
}

PR #9779, opened 2024-04-02
fix(kumactl): correctly print new style resources

lahabana:
Prior to this, spec was stripped out which was making the yaml
output invalid

Also rewrite tests to remove a lot of boilerplate and repetitive code.

### Checklist prior to review


app/kumactl/cmd/get/testdata/get/get-meshtimeout.golden.yaml
```diff
@@ -0,0 +1,23 @@
...
+    proxyTypes:
+    - Gateway
+  to:
+  - default:
+      http:
+        requestTimeout: 4s
+    targetRef:
+      kind: MeshService
+      name: api-play-000_microservice-mesh_svc_8080
+type: MeshTimeout
```
slonka, line 23: I think the original author cared about `type` being at the top / having some sort of ordering. But if we can't do that I think you can always pipe to `yq` if you care ( https://github.com/mikefarah/yq/issues/1199#issuecomment-1471145764 )

slonka (approved):
I did not check every fixture. I did check the test / printer and one fixture - `spec` is there so should be 👍

app/kumactl/cmd/get/testdata/get/get-meshtimeout.golden.yaml
lahabana: Yeah it's imperfect. I think that's why k8s picked `kind` instead of `type`. I think this is the best we can do with low effort. We can always improve in the future

PR #9782, opened 2024-04-02
chore(meshservice): remove status in spec

jakubdyszkiewicz:
### Checklist prior to review

We introduced status next so spec. We should get rid of status in mesh service spec.


pkg/plugins/runtime/k8s/controllers/meshservice_controller.go
```diff
@@ -131,7 +131,10 @@ func (r *MeshServiceReconciler) Reconcile(ctx context.Context, req kube_ctrl.Req
 			})
 		}
 
-		ms.Spec.Status.VIPs = []meshservice_api.VIP{
+		if ms.Status == nil {
```
michaelbeaumont, line 134: Theoretically we could do https://github.com/kumahq/kuma/blob/fdebf4b7dff3035f89fc740debddf6e1b4b02bba/pkg/plugins/runtime/k8s/controllers/gatewayapi/gateway_status.go#L34 to just update the status subresource

jakubdyszkiewicz: Would this help here though? I see how this is useful for "insights" update in MeshService, but what would be the gain here if we already fetch MeshService?

michaelbeaumont: yeah probably not very much. it's only useful if we were going to update Status when we wouldn't otherwise update Spec

Issue #9774, opened 2024-04-02
policy(meshmetric): remove sidecar.regex since it's been moved under profiles


slonka:
### Description

remove 

https://github.com/kumahq/kuma/blob/master/pkg/plugins/policies/meshmetric/api/v1alpha1/meshmetric.go#L30-L31

in favour of

https://github.com/kumahq/kuma/blob/master/pkg/plugins/policies/meshmetric/api/v1alpha1/meshmetric.go#L48

PR #9793, opened 2024-04-03
feat(meshmetric): manually remove regex

slonka:
closes https://github.com/kumahq/kuma/issues/9774

### Checklist prior to review


lahabana:
Don't we need to follow some deprecation time here?

slonka:
> Don't we need to follow some deprecation time here?

MeshMetric was in alpha and noone is using it on our hosted platform. And even if someone was using it on-prem the only bad thing that could happen is that their metric is not filtered.

slonka:
> Don't we need to follow some deprecation time here?

image

Hmm... actually only we marked OTEL as `alpha` not the whole policy :|

lahabana:
Ha this is part of MeshMetric. I think it's fine then I thought it was a more global config

app/kuma-dp/pkg/dataplane/metrics/server.go
```diff
@@ -74,7 +74,6 @@ func AddSidecarParameters(sidecar *v1alpha12.Sidecar) func(queryParameters url.V
 	values := v1alpha1.EnvoyMetricsFilter(sidecar)
 
 	return func(queryParameters url.Values) url.Values {
-		queryParameters.Set("filter", values.Get("filter"))
 		queryParameters.Set("usedonly", values.Get("usedonly"))
```
Automaat, line 77: is the usedonly still relevant here? don't we set this on otel producer?

Automaat (approved):

PR #9795, opened 2024-04-03
ci(update-vulnerable-dependencies): pass KUMA_DIR to the script

bartsmykla:
When this target is being called from the projects which depends on Kuma, path to `tools/ci/update-vulnerable-dependencies/main.go` is different, so script needs to take into account this root path.

### Checklist prior to review

  - No relevant issues
  - It actually fixes child repos
  - Manually tested locally
  - There is no need

> Changelog: skip

### DIFF
--- a/tools/ci/update-vulnerable-dependencies/update-vulnerable-dependencies.sh
+++ b/tools/ci/update-vulnerable-dependencies/update-vulnerable-dependencies.sh
@@ -5,6 +5,7 @@ set -e
 command -v osv-scanner >/dev/null 2>&1 || { echo >&2 "osv-scanner not installed!"; exit 1; }
 command -v jq >/dev/null 2>&1 || { echo >&2 "jq not installed!"; exit 1; }
 
+KUMA_DIR=${1:-"."}
 
 for dep in $(osv-scanner --lockfile=go.mod --json | jq -c '.results[].packages[] | .package.name as $vulnerablePackage | {
   name: $vulnerablePackage,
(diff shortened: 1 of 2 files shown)


tools/ci/update-vulnerable-dependencies/update-vulnerable-dependencies.sh
```diff
@@ -5,14 +5,15 @@ set -e
...
+KUMA_DIR=${1:-"."}
 
 for dep in $(osv-scanner --lockfile=go.mod --json | jq -c '.results[].packages[] | .package.name as $vulnerablePackage | {
   name: $vulnerablePackage,
   current: .package.version,
   fixedVersions: [.vulnerabilities[].affected[] | select(.package.name == $vulnerablePackage) | .ranges[].events |
   map(select(.fixed != null) | .fixed)] | map(select(length > 0)) } | select(.name != "github.com/kumahq/kuma")'); do
 
-  fixVersion=$(go run tools/ci/update-vulnerable-dependencies/main.go <<< "$dep")
+  fixVersion=$(go run "$KUMA_DIR"/tools/ci/update-vulnerable-dependencies/main.go <<< "$dep")
```
lahabana, line 16: Isn't just the same dir as the one in the script? Really feel like this should be moved to `kumahq/ci-tools`

lahabana:
```suggestion
  dir=$( dirname -- "${BASH_SOURCE[0]}" )
  fixVersion=$(go run "$dir"/main.go <<< "$dep")
```

bartsmykla: You are right with the same dir. I'm gonna update the PR. I would prefer to first fix the current solution, and in a following up PR to potentially move it to `kumahq/ci-tools`

lahabana (approved):
LGTM and I guess you already validated this with entreprise anyway

bartsmykla:
> LGTM and I guess you already validated this with entreprise anyway

I did

PR #9826, opened 2024-04-04
feat(auth): add possibility to restrict /config access

### LINKED ISSUE #9825: add possibility to restrict /config to authenticated user
### Description

This endpoint may contain sensitive information. It may therefore be useful to require auth on it

lahabana:
fix #9825

### Checklist prior to review


pkg/envoy/admin/access/static.go
```diff
@@ -35,39 +39,26 @@ func NewStaticEnvoyAdminAccess(
 
 func mapAccess(users []string, groups []string) accessMaps {
 	m := accessMaps{
-		usernames: map[string]bool{},
-		groups:    map[string]bool{},
+		usernames: make(map[string]struct{}, len(users)),
```
slonka, line 42: probably for the next PR: this looks pretty similar to the other file, should we DRY it up?

slonka (approved):
A bit of repetition here and there but it was introduced before so I'm ok to address it in a separate PR

pkg/envoy/admin/access/static.go
lahabana: Yeah wanted to be careful at changing too much (already doing a bit of this...

PR #9877, opened 2024-04-09
fix{meshloadbalancingstrategy): configure bultiin gateway

lukidzi:
### Checklist prior to review

While checking the MeshLoadBalancingStrategy I noticed the builtin gateway is not configured when using targetRef to, `MeshService`. It was caused by a condition that we were trying to get configuration using MeshSubset from `MeshService`.

```
meshConf := core_rules.ComputeConf[api.Conf](rules, core_rules.MeshSubset())
```
which returns nil because a subset of a specific service cannot have all services.

* use per-service configuration and configure listener once to avoid duplicate configuration
* allow more than one filter chain (gateway might have more)
* added E2E test

### DIFF
--- a/pkg/plugins/policies/meshloadbalancingstrategy/plugin/v1alpha1/plugin.go
+++ b/pkg/plugins/policies/meshloadbalancingstrategy/plugin/v1alpha1/plugin.go
@@ -187,12 +187,6 @@ func (p plugin) configureGateway(
 			continue
 		}
 
-		meshConf := core_rules.ComputeConf[api.Conf](rules, core_rules.MeshSubset())
-
-		if err := p.configureListener(listener, gatewayRoutes, meshConf); err != nil {
-			return err
-		}
-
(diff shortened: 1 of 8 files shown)


pkg/plugins/policies/meshloadbalancingstrategy/plugin/v1alpha1/plugin.go
```diff
@@ -291,31 +293,37 @@ func (p plugin) configureListener(
 		return nil
 	}
 
-	if l.FilterChains == nil || len(l.FilterChains) != 1 {
-		return errors.Errorf("expected exactly one filter chain, got %d", len(l.FilterChains))
+	if l.FilterChains == nil {
+		return errors.New("expected more or one filter chain, got 0")
```
Automaat, line 297: this error message sounds weird, maybe: 
```
expected at least one filter chain
```
or
```
filter chains should not be empty
```
?

Automaat (approved):

bartsmykla (approved):

Issue #9258, opened 2024-02-14
Deprecate `from.kind: MeshService`


lahabana:
### What happened?

This doesn't really make a lot of sense and everything can be achieved with `kind: MeshSubset tags: kuma.io/service`.

This will make things less confusing when we separate service and workload identity.

This issue covers updating the docs and starting to warn about the deprecation. We'll need a follow up issue to remove support.

bartsmykla:
This is breaking change, and is modifying some of the previous MADRs, so there should be an MADR created before implementing this.

lahabana:
This should start with a deprecation so that people are just warned that this is being removed

PR #9881, opened 2024-04-09
feat(kuma-cp): deprecated `from[].targetRef.kind: MeshService`

lobkovilya:
### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/9258


app/kumactl/cmd/apply/apply.go
```diff
@@ -153,21 +160,29 @@ $ kumactl apply -f https://example.com/resource.yaml
...
+			return nil, err
 		}
 	}
 	if err := newRes.SetSpec(res.GetSpec()); err != nil {
-		return err
+		return nil, err
 	}
-	return rs.Update(ctx, newRes, store.UpdateWithLabels(meta.GetLabels()))
+	uerr := rs.Update(warnContext, newRes, store.UpdateWithLabels(meta.GetLabels()))
+	return warnings, uerr
```
jakubdyszkiewicz, line 187: we really need to introduce proper client. It's quite a big workaround. Ofc not a subject of this PR

pkg/core/resources/model/resource.go
```diff
@@ -128,6 +128,13 @@ func Validate(resource Resource) error {
 	return nil
 }
 
+func CheckDeprecated(resource Resource) []string {
```
jakubdyszkiewicz, line 131: nit: For me `check` should result in true/false or error/not error. I'd name it maybe `Deprecations()`

jakubdyszkiewicz (approved):

app/kumactl/cmd/apply/apply.go
```diff
@@ -138,9 +139,15 @@ $ kumactl apply -f https://example.com/resource.yaml
 						return err
 					}
 				} else {
-					if err := upsert(cmd.Context(), pctx.Runtime.Registry, rs, resource); err != nil {
+					warnings, err := upsert(cmd.Context(), pctx.Runtime.Registry, rs, resource)
+					if err != nil {
 						return err
 					}
+					for _, w := range warnings {
+						if _, err := fmt.Fprintf(cmd.ErrOrStderr(), "Warning: %v\n", w); err != nil {
```
lahabana, line 147: Worth adding a `#` ?

app/kumactl/cmd/apply/apply.go
lahabana: Yeah we need to finish the openAPI spec and just autogen most of the client!

lahabana:
FaultInjection is the only one that allows MeshService in from?

lobkovilya:
> FaultInjection is the only one that allows MeshService in from?

There are 2 – MeshFaultInjection and MeshTrafficPermission

app/kumactl/cmd/apply/apply.go
lobkovilya: sorry not sure I understand, you mean adding `#` somewhere in the output line?

lahabana: That way if we return a yaml the warning is a comment. Though this is Stderr so it's fine

pkg/core/resources/model/resource.go
```diff
@@ -128,6 +128,13 @@ func Validate(resource Resource) error {
 	return nil
 }
 
+func Deprecations(resource Resource) []string {
+	if v, ok := interface{}(resource).(interface{ Deprecated() []string }); ok {
```
lahabana, line 132: `GetDeprecations()` or just `Deprecations()` might be a better name here no?

lobkovilya: Right, `Deprecations` sounds better

lukidzi (approved):

PR #9882, opened 2024-04-09
feat(kuma-cp): do not set mesh owner reference on synced resources

jakubdyszkiewicz:
### Checklist prior to review

I investigated KDS NACKs in E2E tests and all of them are caused by mesh owner reference.
What is interesting is that we don't set Mesh owner reference on Universal (because we go directly thorough store, not resource manager), that's why the only NACKs we saw were from Kube zones.

Setting mesh owner reference caused the following problems:
* Policy is synced before mesh - fails because mesh is not present so we cannot set an owner
* Resource not found on delete - fails because resource was first deleted by owner mechanism

Additionally, I disabled secret delete validation on synced resources. This also caused a race when you delete Secret before Mesh. It was also only applied on Kube because of same reasons (store vs manager).

After this change, I see 0 NACKs on KDS in the multizone suite.

Also, first I tried ordering resources over KDS, but not waiting for ACK resulted races as well. I think it's just more correct to trust the source of truth (other CP that manages resource).

### DIFF
--- a/pkg/plugins/runtime/k8s/webhooks/owner_reference_mutator.go
+++ b/pkg/plugins/runtime/k8s/webhooks/owner_reference_mutator.go
@@ -10,6 +10,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
+	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
+	config_core "github.com/kumahq/kuma/pkg/config/core"
 	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
 	meshservice_api "github.com/kumahq/kuma/pkg/core/resources/apis/meshservice/api/v1alpha1"
 	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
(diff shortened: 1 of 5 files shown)


Automaat (approved):

pkg/plugins/runtime/k8s/webhooks/owner_reference_mutator.go
```diff
@@ -77,3 +84,11 @@ func (m *OwnerReferenceMutator) Handle(ctx context.Context, req admission.Reques
 	}
 	return admission.PatchResponseFromRaw(req.Object.Raw, mutatedRaw)
 }
+
+func syncedResource(cpMode config_core.CpMode, labels map[string]string) bool {
+	syncedOrigin := mesh_proto.GlobalResourceOrigin
+	if cpMode == config_core.Global {
+		syncedOrigin = mesh_proto.ZoneResourceOrigin
+	}
+	return len(labels) > 0 && labels[mesh_proto.ResourceOriginLabel] == string(syncedOrigin)
```
lobkovilya, line 93: you can use [ResourceOrigin](https://github.com/kumahq/kuma/blob/b0789c8fda5bd8d8724cd06b5c97de049e22c6c4/pkg/core/resources/model/resource.go#L443) func

lobkovilya (approved):
Nice! I like the idea 👍

pkg/plugins/runtime/k8s/webhooks/owner_reference_mutator.go
jakubdyszkiewicz: it takes meta. I don't have a core resource here, only kube resource

PR #9850, opened 2024-04-08
feat(kuma-cp): shadow mode for policies

lobkovilya:
Implementation of the [madr-40](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/040-transition-to-new-policies.md)

Does **not** cover:
* Global CP support (`_config` endpoint works only on Zone CP)
* no support for `_rules` + `shadow`
* **upd:** no support for `kumactl` in this PR, I'm going to open another one

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues --
- [ ] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [X] Tests (Unit test, E2E tests, manual test on universal and k8s) --
  - Don't forget `ci/` labels to run additional/fewer tests
- [X] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [X] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) --

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #9887, opened 2024-04-09
feat(kumactl): support for new Inspect API endpoint `_config`

lobkovilya:
Follow up for https://github.com/kumahq/kuma/pull/9850 

Now it's possible to see the diff with shadow in a nice and more compact `jd` format. For example,

<details>
  <summary>we apply MeshFaultInjection with "kuma.io/effect: shadow"</summary>

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshFaultInjection
metadata:
  name: default
  namespace: kuma-system
  labels:
    kuma.io/origin: zone
    kuma.io/mesh: default
    kuma.io/effect: shadow
spec:
  targetRef:
    kind: Mesh
    proxyTypes:
      - Sidecar
  from:
    - targetRef:
        kind: Mesh
      default:
        http:
          - abort:
              httpStatus: 500
              percentage: 51
```      

</details>

```bash
$ kumactl inspect dataplane demo-app-d789d49c-sd9d7.kuma-demo --type=config --shadow=true --include=diff | jq '.diff' | jd -t patch2jd

@ ["type.googleapis.com/envoy.config.listener.v3.Listener","inbound:10.42.0.7:5000","filterChains","0","filters","0","typedConfig","httpFilters","1"]
- {"name":"envoy.filters.http.router","typedConfig":{"@type":"type.googleapis.com/envoy.extensions.filters.http.router.v3.Router"}}
@ ["type.googleapis.com/envoy.config.listener.v3.Listener","inbound:10.42.0.7:5000","filterChains","0","filters","0","typedConfig","httpFilters","0","name"]
- "envoy.filters.http.fault"
@ ["type.googleapis.com/envoy.config.listener.v3.Listener","inbound:10.42.0.7:5000","filterChains","0","filters","0","typedConfig","httpFilters","0","name"]
+ "envoy.filters.http.router"
@ ["type.googleapis.com/envoy.config.listener.v3.Listener","inbound:10.42.0.7:5000","filterChains","0","filters","0","typedConfig","httpFilters","0","typedConfig","@type"]
- "type.googleapis.com/envoy.extensions.filters.http.fault.v3.HTTPFault"
@ ["type.googleapis.com/envoy.config.listener.v3.Listener","inbound:10.42.0.7:5000","filterChains","0","filters","0","typedConfig","httpFilters","0","typedConfig","@type"]
+ "type.googleapis.com/envoy.extensions.filters.http.router.v3.Router"
@ ["type.googleapis.com/envoy.config.listener.v3.Listener","inbound:10.42.0.7:5000","filterChains","0","filters","0","typedConfig","httpFilters","0","typedConfig","abort"]
- {"httpStatus":500,"percentage":{"numerator":51}}

```

### Checklist prior to review


app/kumactl/pkg/resources/inspect_envoy_proxy_client.go
```diff
@@ -45,6 +47,10 @@ func (h *httpInspectEnvoyProxyClient) Clusters(ctx context.Context, rk core_mode
 	return h.executeInspectRequest(ctx, rk, "clusters")
 }
 
+func (h *httpInspectEnvoyProxyClient) Config(ctx context.Context, rk core_model.ResourceKey, shadow bool, include []string) ([]byte, error) {
+	return h.executeInspectRequest(ctx, rk, fmt.Sprintf("_config?shadow=%t&include=%s", shadow, strings.Join(include, ",")))
```
lahabana, line 51: Please build urls correctly not with `fmt.Sprintf`

Automaat (approved):

app/kumactl/pkg/resources/inspect_envoy_proxy_client.go
```diff
@@ -34,22 +36,30 @@ type httpInspectEnvoyProxyClient struct {
...
+		"include": include,
+	})
+}
+
+func (h *httpInspectEnvoyProxyClient) executeInspectRequest(ctx context.Context, rk core_model.ResourceKey, inspectionPath string, queryParams url.Values) ([]byte, error) {
 	resUrl, err := h.buildURL(rk, inspectionPath)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not construct the url")
 	}
+	resUrl.RawQuery = queryParams.Encode()
```
slonka, line 62: am I missing something or are we going to call `.Encode()` on `nil` from lines [47](https://github.com/kumahq/kuma/pull/9887/files#diff-f767b855e17c699f1100e5f7bd6f7bed1d7b7d683f7746c8cd1bb26fc2c970f3R47), 43, 39?

slonka (commented):
One thing that worries me.

app/kumactl/pkg/resources/inspect_envoy_proxy_client.go
lobkovilya: There is a check inside so I was thinking it's okay https://github.com/golang/go/blob/master/src/net/url/url.go#L973. But we can check on the Kuma side as well

lobkovilya: or maybe it's better to pass empty `url.Values` instead

slonka: just something I noticed if it won't NPE then I guess either one is good. Side q: do other builtin structures also have this? what's the idiomatic go way of doing this?

lobkovilya: not sure about the idiomatic way, but I think passing an empty `url.Values{}` is better than `nil` because that way code-readers don't have to check `.Encode()` implementation every time to ensure it's NPE safe.

PR #9890, opened 2024-04-10
docs(UPGRADE.md): add section about using MeshHTTPRoute in GAPI

bartsmykla:
### Checklist prior to review

  - No relevant issues
  - It won't
  - No tests necessary
  - It doesn't

> Changelog: skip


UPGRADE.md
```diff
@@ -126,6 +126,27 @@ In the upcoming release, Kuma will redirect IPv6 traffic to the same port as IPv
...
 We recommend that you update your configurations to use the new defaults for IPv6 traffic redirection. If you need to retain separate ports for IPv4 and IPv6 traffic, you can continue to use the deprecated flags and configuration options until they are removed.
 
+### Change in internal resources with Kubernetes Gateway API
+
+This section describes changes to internal resources used by Kuma when configuring the built-in gateway using the Kubernetes Gateway API.
+
+#### Prior Behavior (Before Kuma 2.7.0):
+
+  * Applying a `Gateway` resource resulted in the creation of corresponding `MeshGateway` and `MeshGatewayInstance` resources.
+  * Applying an `HTTPRoute` resource was converted to a `MeshGatewayRoute` resource.
```
slonka, line 136: not a native speaker but, shouldn't this be like?

```suggestion
  * An applied `HTTPRoute` resource was converted to a `MeshGatewayRoute` resource.
```

UPGRADE.md
```diff
@@ -126,6 +126,27 @@ In the upcoming release, Kuma will redirect IPv6 traffic to the same port as IPv
...
+  * `HTTPRoute` resources are now converted to `MeshHTTPRoute` resources instead of `MeshGatewayRoute` resources.
+
+#### Upgrade Impact:
+
+  * Existing `MeshGatewayRoute` resources automatically created from `HTTPRoute` definitions will be deleted during the upgrade.
+  * New `MeshHTTPRoute` resources will be created to replace the deleted ones.
+
+#### Important Note:
+
+This change is transparent with regard to the generated Envoy configuration. There should be no impact on existing traffic routing.
```
slonka, line 149: shouldn't we start with this? that there should be no impact to the live traffic?

bartsmykla: I don't think so. This information may seem extensive, but it's always worth to read it fully. While I've tried to anticipate all potential impacts of this change, there's always a chance something unexpected could arise. Reading this entire section will ensure you're fully informed of the changes and any potential consequences, even those I may not have foreseen.

michaelbeaumont: Yeah I agree, ultimately `UPGRADE.md` just isn't that long, we can try to push users to read it all.

Automaat (approved):

PR #9896, opened 2024-04-10
perf(k8s): ignore serviceless pods from outbounds

jakubdyszkiewicz:
### Checklist prior to review

Ignore serviceless pods from being added as an outbound to other dataplanes.

Serviceless pod is a pod that has no service selected. In this case we build `kuma.io/service` out of Deployment/Job name. This pod is not meant to be addressed. We were adding a mesh VIP to this `kuma.io/service` and were adding outbound to all DPPs (if reachable service is not used). This can cause quite a lot of unnecessary xds changes.

This PR marks such inbound with state `Ignored`. It means that such Dataplane preserves `kuma.io/service`. However, it's not included in XDS nor in ServiceInsights.

I left `health` field as is. We need to keep it for one more minor release.


pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -118,15 +116,13 @@ func inboundForServiceless(zone string, pod *kube_core.Pod, name string, nodeLab
...
 	); cs != nil && !cs.Ready {
-		state = mesh_proto.Dataplane_Networking_Inbound_NotReady
 		health.Ready = false
 	}
 
 	return &mesh_proto.Dataplane_Networking_Inbound{
-		Port:   mesh_proto.TCPPortReserved,
-		Tags:   tags,
-		State:  state,
-		Health: &health, // write health for backwards compatibility with Kuma 2.5 and older
```
Automaat, line 129: you wrote in pr description that you left health as is, but you removed if from here, is it correct?

jakubdyszkiewicz: it's not 🤦

Automaat (approved):

slonka (approved):

jakubdyszkiewicz:
This has unfortunately more impact than I anticipated. Policy matching does not apply to ignored listener, we need to go with ignore reserved port on VIPs level. I'll open another PR.

PR #9904, opened 2024-04-10
feat(k8s): add drain when using native sidecars

michaelbeaumont:
The idea here is to take advantage of the separation of SIGTERM from `preStop` hook for native sidecar containers and use Envoy to drain connections for applications. The effect is that Envoy will drain incoming connections as long as the application runs and when the application exits and `kuma-dp` gets SIGTERM, `kuma-dp` exits immediately.

This PR uses SIGUSR2 to signal that kuma-dp should drain.

- ~[ ] port~

### Checklist prior to review


app/kuma-dp/pkg/dataplane/envoy/dpserver.go
```diff
@@ -0,0 +1,75 @@
...
+var dpServerLog = core.Log.WithName("kuma-dp").WithName("run").WithName("dpServer")
+
+type DPServerComponent struct {
+	mux         *http.ServeMux
+	port        int
+	notifyDrain chan<- struct{}
+	done        chan struct{}
+}
+
+func NewDPServer(port int, notifyDrain chan<- struct{}) component.GracefulComponent {
```
jijiechen, line 25: I understand this `DPServerComponent` is designed to be a generic server container that can hold arbitrary logic that is fired by this channel. If this is the idea, should we name the `notifyDrain` to something else, like "triggerer". In addition, the the http path and method should also be generic to make this a generic container. 
Otherwise, if we don't make it generic, should we put the calling to the `envoyComponent.DrainForever` into this class?

app/kuma-dp/cmd/run.go
```diff
@@ -239,22 +239,46 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
...
-						case <-ctx.Done():
+				var draining bool
+				for {
+					select {
+					case _, ok := <-usr2Recv:
+						if !ok {
+							// If our channel is closed, never take this branch
+							// again
+							usr2Recv = nil
+							continue
```
jakubdyszkiewicz, line 250: looks like infinite loop to me?

run this
```
func main() {
	ch := make(chan int)
	close(ch)

	for {
		println("log")
		select {
		case _, ok := <-ch:
			if !ok {
				continue
			}
		}
	}
}
```

pkg/core/alias.go
```diff
@@ -22,24 +22,47 @@ var (
...
 	Now                   = time.Now
 
-	SetupSignalHandler = func() (context.Context, context.Context) {
+	SetupSignalHandler = func() (context.Context, context.Context, <-chan struct{}) {
 		gracefulCtx, gracefulCancel := context.WithCancel(context.Background())
 		ctx, cancel := context.WithCancel(context.Background())
 		c := make(chan os.Signal, 3)
-		signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+		signal.Notify(c, syscall.SIGINT, syscall.SIGTERM, syscall.SIGUSR2)
+		usr2Notify := make(chan struct{}, 1)
```
jakubdyszkiewicz, line 30: I'd drop `, 1`. Why buffer if we only write in a non-blocking way

app/kuma-dp/cmd/run.go
jakubdyszkiewicz: Ok, it's fine because this `user2Recv = nil`...

jakubdyszkiewicz (approved):
TIL: nil channels behavior is much different than closed channels...

pkg/core/alias.go
lahabana: Agreed with this understanding

pkg/core/alias.go
```diff
@@ -22,24 +22,47 @@ var (
...
+				switch s {
+				case syscall.SIGINT, syscall.SIGTERM:
+					stopSignalCancel(s)
+				case syscall.SIGUSR2:
+					select {
+					case usr2Notify <- struct{}{}:
+					default:
+					}
+				}
+			}
```
lahabana, line 63: How about something that doesn't iterate over function references?

```go
	signalReceived := 0
	for {
		s:=<-c
		switch s {
		case syscall.SIGINT, syscall.SIGTERM:
			signalReceived += 1
			logger.Info("received signal, stopping instance gracefully", "signal", s.String(), "signalCount", signalReceived)
		case syscall.SIGUSR2:
			select {
			case usr2Notify <- struct{}{}:
			default:
			}
			break
		}
		switch signalReceived {
		case 1:
			gracefulCancel()
			close(usr2Notify)
			usr2Notify = nil
		case 2:
			cancel()
		case 3:
			s.Exit(1)
		default:
			// ??
		}
	}
```

michaelbeaumont: I think having a direct pointer from signal -> transition makes it easier to create the state machine in my head but if you both find this way easier I can change it

lahabana: Yeah I find your new way much easier to read

PR #9912, opened 2024-04-11
fix(transparent-proxy): port conflict in validator in unit tests

jijiechen:
### Checklist prior to review

  - N/A
  - Confirmed
  - No need
  - No need

> Changelog: skip


pkg/transparentproxy/validate/validator_test.go
```diff
@@ -64,11 +67,16 @@ var _ = Describe("Should Validate iptables rules", func() {
...
 	})
 })
 
-func createValidator(ipv6Enabled bool) *Validator {
-	return NewValidator(ipv6Enabled, ValidationServerPort, core.NewLoggerTo(os.Stdout, kuma_log.InfoLevel).WithName("validator"))
+func createValidator(ipv6Enabled bool, validationServerPort uint16) *Validator {
+	return NewValidator(ipv6Enabled, validationServerPort, core.NewLoggerTo(os.Stdout, kuma_log.InfoLevel).WithName("validator"))
+}
+
+func randomPort() uint16 {
```
jakubdyszkiewicz, line 79: We need to check if something is listening on a port you pick. Use `pkg/util/net#PickTCPPort`

jakubdyszkiewicz: sorry, it's `pkg/test#GetFreePort`. We should remove the `pkg/util/net#PickTCPPort` util.

jakubdyszkiewicz:
should this be a PR to branch release-2.7?

pkg/transparentproxy/validate/validator.go
```diff
@@ -105,32 +105,41 @@ type LocalServer struct {
...
-		s.logger.Error(err, fmt.Sprintf("error listening on %v", s.config.ServerListenIP))
+		close(readiness)
 		return err
 	}
 
+	// if the port is not set, we will use the port that is assigned by the OS
+	// this is useful for testing purposes when cases are ran in parallel, to make sure ports are not conflicting
+	if s.config.ServerListenPort == 0 {
+		s.config.ServerListenPort = uint16(l.Addr().(*net.TCPAddr).Port)
+		if s.config.ClientConnectPort == 0 {
```
michaelbeaumont, line 121: does the randomization behavior still happen?

pkg/transparentproxy/validate/validator_test.go
```diff
@@ -54,7 +53,8 @@ var _ = Describe("Should Validate iptables rules", func() {
 
 	It("should return fail when no iptables rules setup", func() {
 		// given
-		validator := createValidator(false)
+		validator := createValidator(false, uint16(0))
+		validator.Config.ClientConnectPort = ValidationServerPort
```
michaelbeaumont, line 57: why is this still necessary if we set the port in `validator.go`?

michaelbeaumont: Should ClientConnectPort even be configurable?

jijiechen: It actually is only usable in this particular case when we want the client to connect to an invalid port. Without this, we can not control the validator to distingwish from these two scenarios:

1. auto generate a port and connect to that port (useful to test the validation logic without any iptables rules applied, this is used in the case above)
2. connect to a different port from the server listening port  (this should succed when iptables rules are applied, but it only holds in a real product run)

michaelbeaumont: Why even have the logic for 1. in `validator.go` though? It's only ever executed by the test. It's not really testing anything.

michaelbeaumont: Where is `ClientConnectPort` ever set otherwise? I don't see it in the code.

michaelbeaumont: I don't know the flow seems odd to me. `LocalServer` is mutating `Config.ClientConnectPort` so that we can later pass `ClientConnectPort` to `LocalClient`, it's difficult to follow what's happening here.

michaelbeaumont: I would expect something like

1. run Server(porttolistenon), either 0 or some port, which returns the port the server is listening on
2. run Client(porttoconnectto)

Between 1 and 2 we have to decide whether we want to use `porttolistenon` or a random port but we should only have code to do the first in the tests and only execute "find random port to connect to" in the real world.
The interface shouldn't know that it's in test mode or not test mode IMO.

pkg/transparentproxy/validate/validator.go
```diff
@@ -142,18 +127,28 @@ func (s *LocalServer) handleTcpConnections(l net.Listener, cExit chan struct{})
...
 			return retry.RetryableError(e)
 		}
-		validator.Logger.Info("Client: connection established")
+		validator.Logger.Info("client: connection established")
 		return nil
 	})
+
+	if clientErr != nil {
+		clientErr = fmt.Errorf("validation failed, client failed to connect to the verification server: %w", clientErr)
+		close(sExit)
```
michaelbeaumont, line 145: nit: maybe a `defer close(sExit)` to be safer

michaelbeaumont (approved):
LGTM!!

PR #9957, opened 2024-04-16
feat(api-server): add errors to trace spans

michaelbeaumont:
### Checklist prior to review

### DIFF
--- a/pkg/core/rest/errors/error_handler.go
+++ b/pkg/core/rest/errors/error_handler.go
@@ -162,7 +162,9 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 		}
 	}
 	if ctx != nil {
-		if span := trace.SpanFromContext(ctx); span.IsRecording() {
+		span := trace.SpanFromContext(ctx)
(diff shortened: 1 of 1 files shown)


pkg/core/rest/errors/error_handler.go
```diff
@@ -162,7 +162,9 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
 		}
 	}
 	if ctx != nil {
-		if span := trace.SpanFromContext(ctx); span.IsRecording() {
+		span := trace.SpanFromContext(ctx)
+		span.RecordError(err, trace.WithStackTrace(true))
+		if span.IsRecording() {
```
slonka, line 167: are you sure RecordError does not have to be inside `IsRecording` if?

slonka (approved):

pkg/core/rest/errors/error_handler.go
michaelbeaumont:
```
	// be set to Error, as this method does not change the Span status. If this
	// span is not being recorded or err is nil then this method does nothing.

```

Automaat (approved):

PR #9962, opened 2024-04-16
fix(kuma-cp): use display-name label to check if resource is referenced

lobkovilya:
The existing `IsReferenced` method didn't work properly on Global CP when policies where synced from the Zone.

### Checklist prior to review

### DIFF
--- a/pkg/core/resources/model/resource_test.go
+++ b/pkg/core/resources/model/resource_test.go
@@ -32,7 +32,6 @@ var _ = Describe("Resource", func() {
 })
 
 var _ = Describe("IsReferenced", func() {
-
 	metaFuncs := map[string]map[string]func(mesh, name string) model.ResourceMeta{
 		"zone": {
 			"k8s": func(mesh, name string) model.ResourceMeta {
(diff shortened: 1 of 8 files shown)


pkg/core/resources/model/resource_test.go
```diff
@@ -29,60 +32,136 @@ var _ = Describe("Resource", func() {
...
-		})
-		It("should return false when t1 is referencing route-2", func() {
-			Expect(model.IsReferenced(meta("m1", "t1"), "route-2", meta("m1", "route-1"))).To(BeFalse())
-		})
-		It("should return false when meshes are different", func() {
-			Expect(model.IsReferenced(meta("m1", "t1"), "route-1", meta("m2", "route-1"))).To(BeFalse())
-		})
-	})
+	}
+	fmt.Println("DEBUG", len(entries))
```
michaelbeaumont, line 132:
```suggestion
```

lukidzi (approved):

Automaat (approved):

Issue #3875, opened 2022-02-18
Support TCP and grpc liveness probes


lahabana:
### Description

We only support httpGet for probes. It would be great to also support TCP and the coming grpc probs too.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe

lahabana:
will need this: https://github.com/kumahq/kuma/issues/3876

lahabana:
Do grpc first and then TCP as grpc will be very similar to existing http.

lahabana:
Also curious to see if this works with `named-port`: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#use-a-named-port

jijiechen:
So we are going to support all the probe types, right? (readiness, liveness and startup)

My idea is to extend our current virtual probe handler and use this listener filter to identify if the inbound request is an HTTP, and switch filter chains between a newly introduced tcp proxy and the existing HCM
https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/listener_filters/http_inspector

WDYT?

lahabana:
As discussed we should have a MADR to support this

jijiechen:
A MADR doc has been created at https://github.com/kumahq/kuma/pull/9964

Yeah, and I found the `http_inspector` probably doesn't help us as the TCP socket probes will always succeed when whenever the connection is established. So the "universal virtual probes port" will make all TCP probes success regardless how the traffic is actually handled with the listener.

PR #9964, opened 2024-04-17
docs(MADR): grpc and tcp virtual probes

jijiechen:
### Checklist prior to review

MADR to support TCP and gRPC probes on Kubernetes.

  - https://github.com/kumahq/kuma/issues/3875
  - Confirmed
  - N/A
  - No need
  - No need


docs/madr/decisions/042-tcp-grpc-probe.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Decision Drivers
+
+- Provide the best user experience by designing a simple and intuitive way of declaring virtual probes
+
+- Keep the compatibility with the existing HTTP based Virtual Probe and data plane data model
+
+## Decision Outcome
+
+Chosen option: (to be chosen)
```
jakubdyszkiewicz, line 246: Two questions.

1) TCP probes
Do we even need to do anything special for TCP probes? Correct me if I'm wrong, but I think TCP probe just opens a connection and that's it. The connection can be mtlsed, because we don't send anything on it.

If we don't need to do anything about TCP then...
2) gRPC probe
gRPC is just HTTP under the hood. Do we even need to have any translator? Isn't it just a matter of providing proper HTTP path? Maybe we don't even need to change Dataplane model if we can do
```
  probes:
    endpoints:
    - inboundPath: /metrics
      inboundPort: 8081
      path: /8081/metrics
    - inboundPath: /pathToGrpcService # I don't remember syntax
      method: POST # In gRPC it's POST, right now AFAIR we only support GET
      inboundPort: 8082
      path: /8081/metrics
    port: 19000
```

docs/madr/decisions/042-tcp-grpc-probe.md
```diff
@@ -0,0 +1,246 @@
...
+      httpGet:
+        path: /3001/metrics
+        port: 19000
+      initialDelaySeconds: 3
+      periodSeconds: 3
+  initContainers:
+  - name: kuma-init
+    image: kuma/kuma-init:latest
+  ...
+```
```
lahabana, line 118: Worth mentioning the dataplane object `probes` section in these cases.

docs/madr/decisions/042-tcp-grpc-probe.md
```diff
@@ -0,0 +1,246 @@
...
+        service: liveness
+    readinessProbe:
+      tcpSocket:
+        host: some-other-host
+        port: 1433
+```
+
+The `host` field in the `tcpSocket` probe is only used to resolve the network address when one wants to probe a different host than the pod's IP address and it's optional. The `service` field on gRPC probes is also optional. There is little space for us to carry contextual information on them.
+
+A pod can have multiple containers with each defines multiple and different probes, the worst case is a pod has all 3 different types of probes with many probes for each type.
```
lahabana, line 42: Isn't one of the main reason for these probes to control readiness of the sidecar? @michaelbeaumont what's the story of this with sidecar containers?

docs/madr/decisions/042-tcp-grpc-probe.md
```diff
@@ -0,0 +1,246 @@
...
+      periodSeconds: 3
+  initContainers:
+  - name: kuma-init
+    image: kuma/kuma-init:latest
+  ...
+```
+
+## Considered Options
+
+- Allocate a new and separated port for each of the TCP probes and gRPC probes
```
lahabana, line 122: Why do gRPC probes need separate ports? AFAIU gRPC is just HTTP so it could just be a different route on the http probe no?

docs/madr/decisions/042-tcp-grpc-probe.md
```diff
@@ -0,0 +1,246 @@
...
+HTTP Probes --> HTTP Probes --->  Virtual Probe Listener ->  HTTP Probes  --->  Application
+              ↗                                           ↘
+   TCP Probes                                               HTTP to TCP translator  --->  Application
+```
+
+#### Positive Consequences
+
+- Don't need to introduce new ports for TCP and gRPC probes
+
+- Better user experience: don't need to introduce new annotations
```
lahabana, line 226: It's worth mentioning that with the existing annotation folks only need to play with it if they use port `19000` for their app

lahabana (commented):
I think there should be something simpler

docs/madr/decisions/042-tcp-grpc-probe.md
lahabana: +1 to this I think some of my comments follow this point.

For TCPProbe @jakubdyszkiewicz seems right too:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe

Though wasn't one of the point of these listeners to also bind sidecar probes with application ?

docs/madr/decisions/042-tcp-grpc-probe.md
michaelbeaumont: Without a startupProbe, immediately after a sidecar container is started, the next initContainer is started. With a startupProbe, kubelet waits until it succeeds. Init containers otherwise can't have probes.

docs/madr/decisions/042-tcp-grpc-probe.md
jijiechen: Multiple gRPC probes to different ports will use the same path and method, so we can't differentiate them within one single listener.

docs/madr/decisions/042-tcp-grpc-probe.md
```diff
@@ -0,0 +1,246 @@
...
+        service: liveness
+    readinessProbe:
+      tcpSocket:
+        port: 1433
+```
+
+In which:
+- `http` is the port to be used for all the HTTP probes, replacing the existing annotation `kuma.io/virtual-probes-port`
+- `grpc-*` is the ports to be used for each of gRPC probes
+- `tcp-*` is the ports to be used for each of TCP probes
```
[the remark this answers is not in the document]
jijiechen: It is not required. It's us that needs this as an implementation detail.

docs/madr/decisions/042-tcp-grpc-probe.md
```diff
@@ -0,0 +1,246 @@
...
+
+A user can now specify the Virtual Probe ports like this:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  annotations:
+    kuma.io/virtual-probes: "enabled"
+    kuma.io/virtual-probes-port-mapping: '{"http": 19000, "grpc-2379": 19001, "tcp-1433": 19002}'
```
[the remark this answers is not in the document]
jijiechen: Yeah, it also works with the range specified.

docs/madr/decisions/042-tcp-grpc-probe.md
jijiechen: Sure, I'll do.

docs/madr/decisions/042-tcp-grpc-probe.md
jijiechen: Added onto line 48

docs/madr/decisions/042-tcp-grpc-probe.md
jijiechen: I think the probes are to support the probes of the applications, rather than to probe the Kuma sidecar.
The kuma containers are run before the application container and probes of kuma containers are working as expected without any actions. So our virtual probes are to fix the probing features broken by the service mesh.

docs/madr/decisions/042-tcp-grpc-probe.md
jakubdyszkiewicz: TCP probes - true. I think TCP probes are very uncommon. Also by the time Envoy actually starts and receives config over XDS for inbound passthrough listener, TCP app next it is probably up already.

GRPC proves - true. But maybe we can route by host header? We would need to verify this, but when kubelet does gRPC HC to app on port 1234 and on port 1235, what does it send in host (:authority) header?

docs/madr/decisions/042-tcp-grpc-probe.md
lahabana: Is this a realistic use case? Can we find an example where we'd end up in this case?

docs/madr/decisions/042-tcp-grpc-probe.md
lahabana: I'd double check this assumption about the Host header. Usually http clients do the right thing of setting the host header to whatever is in the request (like Jakub suggests).
Let's be fact based please and not make assumptions.

jakubdyszkiewicz: > An application running with other protocols would need TCP probes, for example, a gRPC application would need it before the cluster supports gRPC probes.

maybe... however gRPC probes went beta in 1.24 and GA with Kube 1.27 which is already EOL in two months (2024-06-28). What I've seen is that gRPC apps rather expose extra port with HTTP probes. We saw this pattern in our apps in Konnect as well.

jijiechen: gRPC apps on legacy clusters is only an example. I think TCP probes are to support probing application that has other custom protocols.

docs/madr/decisions/042-tcp-grpc-probe.md
jijiechen: We can also make it support only one gRPC probe per pod, if that makes sense

docs/madr/decisions/042-tcp-grpc-probe.md
```diff
@@ -0,0 +1,246 @@
...
+      inboundPort: 8081
+      path: /8081/metrics
+    port: 19000
+  tcpProbes:
+  - port: 19001
+    applicationPort: 1433
+  grpcProbes:
+  - port: 19002
+    applicationPort: 2379
+    service: abcd
```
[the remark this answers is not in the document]
jijiechen: Sure, we can make it that way

docs/madr/decisions/042-tcp-grpc-probe.md
jijiechen: When gRPC requests are being sent to different ports, the HTTP hosts look like this: `10.42.0.12:5086`, which means the target port is part of the host header.

docs/madr/decisions/042-tcp-grpc-probe.md
jijiechen: It turns out that we can support mulple gRPC probes on a single port.

docs/madr/decisions/042-tcp-grpc-probe.md
jijiechen: For TCP probes, `tcp_proxy` listener with a  "blackhole" cluster will always pass the probe. So it looks like the only choice left is to transform TCP probes to HTTP probes.

docs/madr/decisions/046-grpc-probes.md
```diff
@@ -0,0 +1,239 @@
...
+
+We need to allocate a new port for each of them. To allocate listener ports, we'll need a new Pod annotation to support user specifying the range of ports to be used. This annotation, with name `kuma.io/virtual-probes-port-range`, will replace the existing `kuma.io/virtual-probes-port` annotation. The annotation should be optional. When user does not specify the annotation, the default range of ports will be used.
+
+gRPC probe requests are actually HTTP2 requests, which are based on TCP. So with separated virtual probe listener created for each probe, we can use TCP proxies with mTLS disabled in these listeners to forward requests back to application ports. Thus, on the `Dataplane` there is no need to introduce extra fields capturing properties of user defined gRPC probes, except the port number.
```
jakubdyszkiewicz, line 159: I have a concern here. 

Why do people even have gRPC probes instead of HTTP probes? My guess would be that they want to HC the same port as their grpc application port. If this is on separate port then why even bother with gRPC.

Assuming that probe and app is on the same port if we just expose the port without explicit restriction to this specific gRPC service, we basically open the whole app without mTLS.

I think rewriting probes and exposing them without mTLS only works if we are very careful of not exposing the rest of the app

jijiechen: Good point. To prevent security issues, we should use HCM to handle probe requests and only forward the health check endpoint.

jijiechen: Updated.

docs/madr/decisions/046-grpc-probes.md
```diff
@@ -0,0 +1,239 @@
...
+      grpc:
+        port: 2379
+        service: liveness
+```
+
+When a gRPC probe request is being sent from kubelet, it will:
+- include the port as part of the underlying HTTP `Host` header 
+- include the `service` as part of gRPC message body
+
+A pod can have multiple containers with each defines multiple and different probes, the worst case is a pod has all different types of probes with many probes for each type.
```
lahabana, line 40:
```suggestion
A pod can have multiple containers with each defining multiple and different probes, the worst case is a pod has all different types of probes with many probes for each type. 
```

lahabana: There's one detail though here as a port won't be used by different containers as there's a single netns for the whole pod.

docs/madr/decisions/046-grpc-probes.md
```diff
@@ -0,0 +1,239 @@
...
+A pod can have multiple containers with each defines multiple and different probes, the worst case is a pod has all different types of probes with many probes for each type. 
+
+The key design goal is to solve these two problems:
+
+1. Support the probing capability
+2. Differ multiple probes with different types and ports
+
+## Existing design: HTTP Virtual Probes
+
+As per our current design, a pod with multiple probes will be merged and will be handled by a single virtual listener. This listener will forward/redirect probe requests to the corresponding application port without any intermediate transforming except path rewriting, etc.
```
lahabana, line 49:
```suggestion
As per our current design, a pod with multiple probes will be merged and will be handled by a single virtual listener. This listener will forward/redirect probe requests to the corresponding application port without any intermediate modification except path rewriting, etc.
```

docs/madr/decisions/046-grpc-probes.md
```diff
@@ -0,0 +1,239 @@
...
+  ...
+```
+
+## Considered Options
+
+**Option 1:** Transform gRPC probes to HTTP probes, merging all the HTTP probes into the existing HTTP based Virtual Probes listener
+
+**Option 2:** Allocate a new and separated port for each of the gRPC probes, and route them back to the application probe handling port accordingly. 
+
+### Proposed option: option 2 - allocate a new and separated port for each of the gRPC probes, and route them back to the application probe handling port accordingly
```
lahabana, line 153: Odd to start with option 2...

docs/madr/decisions/046-grpc-probes.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Considered Options
+
+**Option 1:** Transform gRPC probes to HTTP probes, merging all the HTTP probes into the existing HTTP based Virtual Probes listener
+
+**Option 2:** Allocate a new and separated port for each of the gRPC probes, and route them back to the application probe handling port accordingly. 
+
+### Proposed option: option 2 - allocate a new and separated port for each of the gRPC probes, and route them back to the application probe handling port accordingly 
+
+The only information we can use to differ from multiple gRPC probes is the port number, while the port number is included in the underlying HTTP2 request as `Host` header, it will be the same if we want to use a single virtual probe listener to support forwarding traffic for all gRPC probes. So we need to allocate a new virtual probe port for each of them. To allocate these ports, we'll need a new Pod annotation to support user specifying the range to be used. This annotation, with name `kuma.io/virtual-probes-port-range`, will replace the existing annotation `kuma.io/virtual-probes-port`. The annotation should be optional: when user does not specify, a default range will be used.
```
lahabana, line 155: multiple gRPC probes to the same service no? Because otherwise the service enables us to make a difference between probes no?

lahabana: What is this range?

lahabana (commented):
Can't we just play with the service field of the probe like we do for the path on the http probe?

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe

https://cloud.google.com/endpoints/docs/grpc/transcoding

My understanding is that the GRPC probe could just be going through the same routes we do ATM. We'd just have extra routes that match: `method: POST, some differentiator` and we'd rewrite the path from one to the other.

For example a probe:

```
grpc:
  port: 4500
  service: foo
```

would become:

```
grpc:
  port: 19000
  service: port4500-foo
```

Is this not possible?

docs/madr/decisions/046-grpc-probes.md
jijiechen: I meant defining different types (httpGet/gRPC/tcpSocket) of probes by:
- different probes (one on startupProe, the other on readinessProbe, etc.)
- on different containers

docs/madr/decisions/046-grpc-probes.md
jijiechen: Yeah, let me shift the ordering.

docs/madr/decisions/046-grpc-probes.md
jijiechen: > multiple gRPC probes to the same service no? Because otherwise the service enables us to make a difference between probes no?

the `service` value will actually be carried over the gRPC payload, insteand of any HTTP header, so it's not available for routing.

jijiechen: I'll add the proposed range.

jijiechen:
> Can't we just play with the service entry in the request?

The `service` is actually part of the gRPC message which means it'll be part of the HTTP body. So it's not unavailabe to routing.
https://github.com/grpc/grpc/blob/master/doc/health-checking.md#service-definition

jijiechen:
> what do other implementations do with probes?

Istio is converting both TCP and gRPC probes to HTTP ones.
https://github.com/istio/istio/blob/master/pkg/kube/inject/app_probe.go#L99-L122

Linkerd looks doesn't support it yet.
https://github.com/linkerd/linkerd2/issues/7777

lahabana:
Feels like this needs updating since our recent conversations...

docs/madr/decisions/046-grpc-tcp-probes.md
```diff
@@ -0,0 +1,289 @@
...
+
+The probes support can be implemented in the following steps:
+
+1. Change the `Dataplane` model to contain fields for gRPC and TCP probes (see examples below)
+2. Create a new component in `kuma-dp` (the probe translator) to handle the incoming HTTP probes by performing the actual gRPC and TCP probes and returning the result
+3. Generate a new Envoy cluster `kuma-virtual-probes` and point it to the component
+4. Make sidecar injector to:
+   1. Get the correct port for the probe translator, and generate corresponding command line arguments for `kuma-dp` 
+   2. Identify gRPC and TCP probes and transform them into the updated `Dataplane` object
+5. Change the route of Virtual Probe listener (in `probe_generator`) and forward the user defined gRPC and TCP probes into the new component in `kuma-dp`
```
jakubdyszkiewicz, line 199: What I'm reading here is that the path is
kubelet -> Envoy -> kuma-dp -> app.

Why involve Envoy if the logic is in kuma-dp? We could just do kubelet -> kuma-dp -> app.

This way we get rid of any config in Dataplane and get rid of xds generation

docs/madr/decisions/046-grpc-tcp-probes.md
```diff
@@ -0,0 +1,289 @@
...
+
+In step 2, the new component is an HTTP server and needs to listen on a new port, a different one from the existing Virtual Probe listener port (default `9000`). The new port will be configurable by `runtime.kubernetes.injector.virtualProbesPort` and the default value is `9001`. The user is not expected to change this port unless it is conflicting with an application port.
```
jakubdyszkiewicz, line 201: why do we need to change the port?

docs/madr/decisions/046-grpc-tcp-probes.md
jijiechen: If we don't involve Envoy, we'll need to skip the port when setting up the traffic redirection (using IPtables, etc.)

docs/madr/decisions/046-grpc-tcp-probes.md
jijiechen: I made a mistake on the configuration key, it should be a new and different key.
We need a new port because we'll need to listen to it in `kuma-dp`, and the current one (default `9000`) was taken by Envoy, so we can't listen on that same port.

docs/madr/decisions/046-grpc-tcp-probes.md
jakubdyszkiewicz: which does not sound like a big deal maybe?
Sounds like fewer broken changes than allocating a new port.

jijiechen: Yeah, agreed when we think in this way.
@bartsmykla WDYT?

docs/madr/decisions/046-grpc-tcp-probes.md
jakubdyszkiewicz: technically we can. For example with Envoy Admin we listen on localhost, but we expose parts of this on PodIP network interface. We could do the same here. However, this problem is gone if we consider skipping port with iptables https://github.com/kumahq/kuma/pull/9964#discussion_r1591939749

jijiechen: Yeah, let me add it into the doc.

jijiechen: I'll mark this conversation as resolved since it's now in the proposed approach.

docs/madr/decisions/046-grpc-tcp-probes.md
jijiechen: I'll mark this conversation as resolved since it's now in the proposed approach.

jakubdyszkiewicz (approved):

PR #9965, opened 2024-04-17
chore(GAPI): move from deprecated behavior

bartsmykla:
Specifying `kuma.io/service` tag for `MeshGatewayInstance` is deprecated, so we shouldn't be use it when converting Gateway API resources to our internal ones.

Specifying `kuma.io/mesh` as an annotation is also deprecated, so I changed it to label in conversion to `MeshGatewayInstance`

### Checklist prior to review

  - No relevant issues
  - It won't
  - There is no need
  - There is no need

> Changelog: skip

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/gatewayapi/common/common.go
+++ b/pkg/plugins/runtime/k8s/controllers/gatewayapi/common/common.go
@@ -2,27 +2,18 @@ package common
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


Automaat (approved):

lahabana:
Why do we changelog skip?

bartsmykla:
> Why do we changelog skip?

I thought it didn't bring enough value especially that we deprecated this behavior in this release, but after some thought I unskipped it.

michaelbeaumont:
Can we please use `gatewayapi` as the scope here? GAPI isn't really useful for readers of the changelog

PR #9971, opened 2024-04-17
fix(k8s): virtual probes for sidecar initContainer ports also exposed by a Service

michaelbeaumont:
### Checklist prior to review


pkg/plugins/runtime/k8s/controllers/probes.go
```diff
@@ -30,6 +30,37 @@ func ProbesFor(pod *kube_core.Pod) (*mesh_proto.Dataplane_Probes, error) {
...
 	}
+	var kumaSidecarSeen bool
+	for _, c := range pod.Spec.InitContainers {
+		if c.Name == util.KumaSidecarContainerName {
+			kumaSidecarSeen = true
+			continue
+		}
+		if !kumaSidecarSeen || c.RestartPolicy == nil || *c.RestartPolicy != kube_core.ContainerRestartPolicyAlways {
+			continue
+		}
```
jijiechen, line 41: Here, I see regardless `kumaSidecarSeen` equals  `true` or not, the execution flow will match at least  one `continue`. Will the other statements work as expected?

pkg/plugins/runtime/k8s/webhooks/injector/probes.go
```diff
@@ -29,6 +29,37 @@ func (i *KumaInjector) overrideHTTPProbes(pod *kube_core.Pod) error {
 		return err
 	}
 
+	var kumaSidecarSeen bool
+	for _, c := range pod.Spec.InitContainers {
+		if c.Name == util.KumaSidecarContainerName {
+			kumaSidecarSeen = true
+			continue
+		}
+		if !kumaSidecarSeen || c.RestartPolicy == nil || *c.RestartPolicy != kube_core.ContainerRestartPolicyAlways {
```
jijiechen, line 38: Same question as previous

pkg/plugins/runtime/k8s/controllers/probes.go
michaelbeaumont: Once we have `kumaSidecarSeen` then we won't take either continue if we have a non Kuma sidecar and `RestartPolicy == Always`

jijiechen: OK, understood, you are correct. The logic is correct here, but not intuitive to understand

jijiechen: Can it be splitted into two parts?   The first part is responsible for finding our native sidecar, and there should be only one or none match; The second part is responsible for overriding the probes, if a match is found

michaelbeaumont: I'm not 100% sure what you mean by finding the sidecar but I reduced duplication.

jijiechen: I basically mean splitting the whole block into two stages and we can also convert the logic assertions to the opposite:

```go
var kumaSidecarSeen bool
	var initContainersToOverride []kube_core.Container
	for _, c := range pod.Spec.InitContainers {
		if c.Name == util.KumaSidecarContainerName {
			kumaSidecarSeen = true
			continue
		}

		isNativeSidecar := c.RestartPolicy != nil && *c.RestartPolicy == kube_core.ContainerRestartPolicyAlways
		if kumaSidecarSeen && isNativeSidecar {
			initContainersToOverride = append(initContainersToOverride, c)
		}
	}

	for _, c := range initContainersToOverride {
		if c.LivenessProbe != nil && c.LivenessProbe.HTTPGet != nil {
			if endpoint, err := probeFor(c.LivenessProbe, port); err != nil {
				return nil, err
			} else {
				dpProbes.Endpoints = append(dpProbes.Endpoints, endpoint)
			}
		}
		if c.ReadinessProbe != nil && c.ReadinessProbe.HTTPGet != nil {
			if endpoint, err := probeFor(c.ReadinessProbe, port); err != nil {
				return nil, err
			} else {
				dpProbes.Endpoints = append(dpProbes.Endpoints, endpoint)
			}
		}
		if c.StartupProbe != nil && c.StartupProbe.HTTPGet != nil {
			if endpoint, err := probeFor(c.StartupProbe, port); err != nil {
				return nil, err
			} else {
				dpProbes.Endpoints = append(dpProbes.Endpoints, endpoint)
			}
		}
	}
```

jijiechen (approved):

pkg/plugins/runtime/k8s/controllers/probes.go
michaelbeaumont: OK I did the splitting, I can change the logic assertions, I agree it's easier to read when it's not negated

jakubdyszkiewicz (approved):
make sense to me

PR #10015, opened 2024-04-19
feat(k8s): opt-in to support tls for GAPI in all namespaces

### LINKED ISSUE #9947: A potential risk in kuma that could lead to takeover of the cluster
### Description

Dear Team Members:

Greetings! Our team is very interested in your project and we recently identified a potential RBAC security risk while doing a security assessment of your project. Therefore, we would like to report it to you and provide you with the relevant details so that you can fix and improve it accordingly. I have already sent the relevant details to your private email, but I couldn't confirm if you received it. That's why I raised the issue here. If there is anything inappropriate about it, I hope you can forgive me.

Details:

In this Kubernetes project, there exists a ClusterRole that has been granted list secrets high-risk permissions. These permissions allow the role to list confidential information across the cluster. An attacker could impersonate the ServiceAccount bound to this ClusterRole and use its high-risk permissions to list secrets information across the cluster. By combining the permissions of other roles, an attacker can elevate the privileges and further take over the entire cluster.

we constructed the following attack vectors.

First, you need to get a token for the ServiceAccount that has this high-risk privilege. If you are already in a Pod and have this override, you can directly run the following command to get the token: cat /var/run/secrets/[[kubernetes.io/serviceaccount/](http://kubernetes.io/serviceaccount/)](http://kubernetes.io/serviceaccount/) token. If you are on a node other than a Pod, you can run the following command to get the kubectl describe secret .

Use the obtained token information to authenticate with the API server. By including the token in the request, you can be recognized as a legitimate user with a ServiceAccount and gain all privileges associated with the ServiceAccount. As a result, this ServiceAccount identity can be used to list all secrets in the cluster.

We give two ways to further utilize ServiceAccount Token with other privileges to take over the cluster:

Method 1: Elevation of Privilege by Utilizing ServiceAccount Token Bound to ClusterAdmin

Directly use a Token with the ClusterAdmin role permissions that has the authority to control the entire cluster. By authenticating with this token, you can gain full control of the cluster.

Method 2: Create Privileged Containers with ServiceAccount Token with create pods permission You can use this ServiceAccount Token to create a privileged container that mounts the root directory and schedules it to the master node in a taint-tolerant way, so that you can access and leak the master node's kubeconfig configuration file. In this way you can take over the entire cluster.

For the above attack chain we have developed exploit code and uploaded it to github: https://github.com/HouqiyuA/k8s-rbac-poc

Mitigation methods are explored:

Carefully evaluate the permissions required for each user or service account to ensure that it is following the principle of least privilege and to avoid over-authorization.

If list secrets is a required permission, consider using more granular RBAC rules. Role Binding can be used to grant list secrets permissions instead of ClusterRole, which restricts permissions to specific namespaces or resources rather than the entire cluster.

Isolate different applications into different namespaces and use namespace-level RBAC rules to restrict access. This reduces the risk of privilege leakage across namespaces

Looking forward to hearing from you and discussing this risk in more detail with us, thank you very much for your time and attention.

Best wishes.

HouqiyuA

jakubdyszkiewicz:
### Checklist prior to review

Fix #9947

Provide get permissions for Secrets as opt-in. This comes at the cost that if you want to use gateway API Gateway with TLS in namespace other than system, you need to set `supportAllGatewaySecrets` to true.

We can still list and watch all the secrets in the cluster. We cannot get the content of them though. This is needed for SecretController to run. The only alternative is to try to run another Kubernetes Manager with restricted access to a namespace, but that's hard to manage.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -220,6 +220,11 @@ Previously, these flags were necessary for using the Gateway API feature:
 - `--experimental-gatewayapi` flag for `kumactl install control-plane` and `kumactl install crds`
 - `experimental.gatewayAPI=true` setting in both `kumactl install control-plane` and Helm charts
 
+### TLS Secrets with Gateway API in namespace other than mesh system namespace
+
+If you use TLS secrets with Gateway API for a gateway deployed in any other namespace than mesh system namespace, set `controlPlane.supportGatewaySecretsInAllNamespaces` HELM value to true.
+This change was introduced so that control plane does not have capability to read content of secrets in all namespaces by default.
+
 ## Upgrade to `2.6.x`
 
 ### Policy
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -222,7 +222,7 @@ Previously, these flags were necessary for using the Gateway API feature:
 
 ### TLS Secrets with Gateway API in namespace other than mesh system namespace
 
-If you use TLS secrets with Gateway API for a gateway deployed in any other namespace than mesh system namespace, set `controlPlane.supportGatewaySecretsInAllNamespaces` HELM value to true.
+If you use TLS secrets with Gateway API for a builtin gateway deployed in any other namespace than mesh system namespace, set `controlPlane.supportGatewaySecretsInAllNamespaces` HELM value to true.
 This change was introduced so that control plane does not have capability to read content of secrets in all namespaces by default.
 
 ## Upgrade to `2.6.x`
--- a/pkg/config/plugins/runtime/k8s/config.go
+++ b/pkg/config/plugins/runtime/k8s/config.go
@@ -153,6 +153,9 @@ type KubernetesRuntimeConfig struct {
 	// If this is set to true, deleting a Mesh will not delete resources that belong to that Mesh.
 	// This can be useful when resources are managed in Argo CD where creation/deletion is managed there.
 	SkipMeshOwnerReference bool `json:"skipMeshOwnerReference" envconfig:"kuma_runtime_kubernetes_skip_mesh_owner_reference"`
+	// If true, then control plane can support TLS secrets for builtin gateway outside of mesh system namespace.
+	// The downside is that control plane requires permission to read Secrets in all namespaces.
+	SupportAllGatewaySecrets bool `json:"supportAllGatewaySecrets" envconfig:"KUMA_RUNTIME_KUBERNETES_SUPPORT_ALL_GATEWAY_SECRETS"`
 }
 
 type ControllersConcurrency struct {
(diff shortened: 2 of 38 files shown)


pkg/config/plugins/runtime/k8s/config.go
```diff
@@ -153,6 +153,9 @@ type KubernetesRuntimeConfig struct {
 	// If this is set to true, deleting a Mesh will not delete resources that belong to that Mesh.
 	// This can be useful when resources are managed in Argo CD where creation/deletion is managed there.
 	SkipMeshOwnerReference bool `json:"skipMeshOwnerReference" envconfig:"kuma_runtime_kubernetes_skip_mesh_owner_reference"`
+	// If true, then control plane can support TLS secrets for builtin gateway outside of mesh system namespace.
+	// The downside is that control plane requires permission to read Secrets in all namespaces.
+	SupportAllGatewaySecrets bool `json:"supportAllGatewaySecrets" envconfig:"KUMA_RUNTIME_KUBERNETES_SUPPORT_ALL_GATEWAY_SECRETS"`
```
lobkovilya, line 158: nit: for some reason, we're using lowercase everywhere in this file for `envconfig`, I'd follow the existing style

lobkovilya, line 158: Probably the name should have "namespace" in it. Maybe `SupportGatewaySecretsInAllNamespaces` or something like this

lobkovilya:
Is it a breaking change? Since we didn't have `supportAllGatewaySecrets` in the config, it effectively was similar to `supportAllGatewaySecrets: true` right? Does it mean after the upgrade the CP stops reconciling secrets in non-system namespaces?

jakubdyszkiewicz:
added upgrade md. Technically it was experimental so we can break it, but added just in case.

UPGRADE.md
```diff
@@ -220,6 +220,11 @@ Previously, these flags were necessary for using the Gateway API feature:
 - `--experimental-gatewayapi` flag for `kumactl install control-plane` and `kumactl install crds`
 - `experimental.gatewayAPI=true` setting in both `kumactl install control-plane` and Helm charts
 
+### TLS Secrets with Gateway API in namespace other than mesh system namespace
+
+If you use TLS secrets with Gateway API for a gateway deployed in any other namespace than mesh system namespace, set `controlPlane.supportGatewaySecretsInAllNamespaces` HELM value to true.
```
lobkovilya, line 225: It's only for builtin right?
```suggestion
If you use TLS secrets with Gateway API for a builtin gateway deployed in any other namespace than mesh system namespace, set `controlPlane.supportGatewaySecretsInAllNamespaces` HELM value to true.
```

jakubdyszkiewicz: Gateway API is only for builtin gateway. It's implicit

jakubdyszkiewicz: Well, ok. You can use GAPI to configure delegated gateway technically

lobkovilya (approved):

jakubdyszkiewicz:
this way we can avoid double commits. We want to introduce a job to merge newest release to master automatically. If we go with master and backport, then we've got double commits.

michaelbeaumont (approved):

Issue #3499, opened 2021-12-13
Update references to deprecated protobuf lib


parkanzky:
### What happened?

Kuma code imports `github.com/golang/protobuf/proto` in a number of places. This package is deprecated in favor of `google.golang.org/protobuf/proto`. Should update everything to recommended package.

Unfortunately, the two packages are not compatible, so need to update all dependencies of any changed module at the same time.

jpeach:
I think the main blocker is the generated Envoy protobufs, https://github.com/envoyproxy/go-control-plane/issues/524.

lukidzi:
Looks like there is one more place: https://github.com/kumahq/kuma/blob/master/pkg/plugins/policies/meshloadbalancingstrategy/plugin/xds/lbconfigurer.go#L5

Icarus9913:
Unfortunatlly, the latest v0.12.0 `go-control-plane` still use golang/protobuf.
https://github.com/envoyproxy/go-control-plane/blob/989e83d4a05c74448fdc72e5a67df5529387c021/go.mod#L9

I'll do this migration once our dependency implemented it.

Icarus9913:
> The latest still supports v1 protobuf but the main branch doesn't, see https://github.com/envoyproxy/go-control-plane/blob/main/go.mod

The main branch still include the `github.com/golang/protobuf v1.5.4`

https://github.com/envoyproxy/go-control-plane/blob/108116100bc16c91f49cd50d4b09495471c711c2/go.mod#L24

Icarus9913:
> Yes but that's not a direct dependency 😅.

Oh, you're right. 
I gonna get a try with the next go-control-plane version.

Icarus9913:
still blocked by https://github.com/envoyproxy/go-control-plane/issues/875

Icarus9913:
We can't get rid of the old protobuf dependency due to https://github.com/golang/protobuf/issues/1374
So, we would just only keep the `github.com/golang/protobuf/jsonpb` dependency but update of all others.

PR #10023, opened 2024-04-21
migrate from github.com/golang/protobuf/ptypes to google.golang.org/protobuf/types/known

mmorel-35:
migrate from `github.com/golang/protobuf/ptypes` to `google.golang.org/protobuf/types/known`

### Checklist prior to review

[1]: https://github.com/kumahq/kuma/issues/3499


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/containernetworking/cni v1.1.2
 	github.com/containernetworking/plugins v1.4.1
 	github.com/emicklei/go-restful/v3 v3.12.0
-	github.com/envoyproxy/go-control-plane v0.12.0
+	github.com/envoyproxy/go-control-plane v0.12.1-0.20240419124334-0cebb2f428b3
```
jakubdyszkiewicz, line 16: @lobkovilya can we do it? I think replace won't work

jakubdyszkiewicz (commented):
please don't merge it

go.mod
lobkovilya: yes the replace won't work in this case, I checked locally

lahabana:
@lukidzi is this something that can be unblocked now that we moved to newer go-control-plane?

lahabana:
@Icarus9913 this should be closed now as you've done this in another PR right?

Icarus9913:
I think we can close it and it's finished by https://github.com/kumahq/kuma/pull/12038

Issue #8324, opened 2023-11-13
Add a more systematic way to add transparent-proxy config


lahabana:
### Description

At the moment adding transparent proxy config is very cumborsome and error prone.

You have to:

1. Add to https://github.com/kumahq/kuma/blob/07fcbed400fad5113c3d2140b337bb657a88e091/pkg/transparentproxy/config/config.go#L11-L43

2. Add to https://github.com/kumahq/kuma/blob/07fcbed400fad5113c3d2140b337bb657a88e091/pkg/transparentproxy/config/config.go#L129-L169

3. Add to config in kumactl install:
https://github.com/kumahq/kuma/blob/07fcbed400fad5113c3d2140b337bb657a88e091/app/kumactl/cmd/install/install_transparent_proxy.go#L20-L51

4. Add to args:
https://github.com/kumahq/kuma/blob/07fcbed400fad5113c3d2140b337bb657a88e091/app/kumactl/cmd/install/install_transparent_proxy.go#L190-L222

5. Add to annotations for CNI:
https://github.com/kumahq/kuma/blob/07fcbed400fad5113c3d2140b337bb657a88e091/app/cni/pkg/cni/annotations.go#L20-L31

6. Map correctly: https://github.com/kumahq/kuma/blob/07fcbed400fad5113c3d2140b337bb657a88e091/app/cni/pkg/cni/injector_linux.go#L70-L142

7. Map transparent proxy annotations too: https://github.com/kumahq/kuma/blob/07fcbed400fad5113c3d2140b337bb657a88e091/pkg/transparentproxy/kubernetes/kubernetes.go#L48-L137

No-one can get that right from the first time.
Is there a better way to do this? Seems like transparent-proxy already uses `pkg/config.Config` which enables env var and json config. Should the annotation maybe just: `traffic.kuma.io/transparent-proxy-config-json` ?

lahabana:
Is there use cases where we need to set this twice and we need to merge multiple configs?

PR #10058, opened 2024-04-24
refactor(transparent-proxy): simplify tproxy configuration

bartsmykla:
This is first PR in a series which should simplify transparent-proxy configuration and make it more systematic (xref: https://github.com/kumahq/kuma/issues/8324)

In this PR:
- Get rid of unnecessary intermediate `TransparentProxyConfig` struct. It's a leftover from the time when we had two tproxy engines. Now we can safely use `Config` struct
- There is no V2 or V1 iptables engines. There is only one, so we don't have to have interface representing engine and can just use pure functions, so I could get rid of `TransparentProxyV2` struct and `TransparentProxy` interface
- We can use general `Config` struct, so I removed `transparentProxyArgs` struct for tproxy uninstall
- Replace writes to stdout buffers with `fmt.Fprintln` in tproxy uninsatll, which is simpler and more elegant`

### Checklist prior to review

  - It won't
  - This is refactoring, and all existing unit tests are passing
  - There is no need
  - There is no need


lahabana (approved):
LGTM as I guess it's just moving things around

bartsmykla:
> LGTM as I guess it's just moving things around

Exactly

Icarus9913 (approved):

Icarus9913:
oops, there's a E2E test failed.

Issue #10072, opened 2024-04-25
Kuma control plane pod crashed


Icarus9913:
### What happened?

I followed the Kuma documentation to get a try with builtin gateway feature but found the kuma-control-plane pod crashed.

Reproduce steps:
1. follow the quickstart to deploy kuma on k8s(k3d)
https://kuma.io/docs/2.7.x/quickstart/kubernetes-demo/
2. create MeshGatewayInstance (https://kuma.io/docs/2.7.x/using-mesh/managing-ingress-traffic/builtin/)
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshGatewayInstance
metadata:
  name: edge-gateway
  namespace: default
spec:
  replicas: 1
  serviceType: LoadBalancer
```
3. create MeshGateway (https://kuma.io/docs/2.7.x/using-mesh/managing-ingress-traffic/builtin-listeners/)
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshGateway
mesh: default
metadata:
  name: edge-gateway
spec:
  selectors:
    - match:
        kuma.io/service: edge-gateway_default_svc
  conf:
    listeners:
      - port: 8080
        protocol: HTTP
        tags:
          port: http/8080
```
4. Delete the MeshGateway
```sh
kubectl delete MeshGateway edge-gateway
```

After deleting the `MeshGateway`, the `Kuma-control-plane` pod crashed:
```text
2024-04-25T02:29:33.144Z	INFO	intercp.catalog.writer	replacing a leader in the catalog
2024-04-25T02:29:33.251Z	INFO	controllers.Service	annotating service which is part of the mesh	{"service": {"name":"demo-app","namespace":"kuma-demo"}, "annotation": "ingress.kubernetes.io/service-upstream=true"}
2024-04-25T02:29:33.252Z	INFO	kube-manager	Starting workers	{"controller": "pod", "controllerGroup": "", "controllerKind": "Pod", "worker count": 10}
2024-04-25T02:29:33.256Z	INFO	kube-manager	Observed a panic in reconciler: assignment to entry in nil map	{"controller": "pod", "controllerGroup": "", "controllerKind": "Pod", "Pod": {"name":"edge-gateway-75778f5c87-wbjsd","namespace":"default"}, "namespace": "default", "name": "edge-gateway-75778f5c87-wbjsd", "reconcileID": "8f4f428d-747b-420d-bd45-71b442f50995"}
panic: assignment to entry in nil map [recovered]
[...]

PR #10073, opened 2024-04-25
fix(builtin-gateway): builtin gateway parse null string tags

Icarus9913:
close https://github.com/kumahq/kuma/issues/10072

- [x] [Link to relevant issue][1] as well as docs and UI issues --
- [ ] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [ ] Tests (Unit test, E2E tests, manual test on universal and k8s) --
  - Don't forget `ci/` labels to run additional/fewer tests
- [ ] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [ ] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) --

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #10076, opened 2024-04-25
fix(gateway): handle implicit kuma.io/service in pod annotation

jakubdyszkiewicz:
### Checklist prior to review

Fix https://github.com/kumahq/kuma/issues/10072

As @Icarus9913 noticed in his PR https://github.com/kumahq/kuma/pull/10073 we end up with `kuma.io/tags: "null"` which we then unmarshal to map and try to set value to nil map.

Why do we end up with `kuma.io/tags: "null"`?

It's because for gateway instance we now have "implicit" `kuma.io/service` tag. It should not be set by user. To handle this, we created a method called `resolveGatewayInstanceServiceTag` in `GatewayInstanceReconciler` which resolves this implicit tag. This however was only called in selector func of `SelectGateway`. If there was at least one gateway, we would resolve this, if not then we wouldn't. If gateway was missing, we would then go to `createOrUpdateDeployment` and write tags to an annotation. This is how ended up with "null".
It also has a bug that if the function was called multiple times (multiple gateways) we would emit a warning.

I'm not a fan of modifying Kube object and not storing the state (because of bugs like this), so I changed the logic a bit to separate validation if a user set a tag and to constructing a map of tags.

Additionally, I added a check to not crash a cp if for some reason we ended up in this situation again.

I tested it manually. It's hard to write E2E for this, because even if we crash a cp it would recover. I think we should consider adding restart policy never to our CP in E2E test, but it's out of scope.

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/gateway_instance_controller.go
+++ b/pkg/plugins/runtime/k8s/controllers/gateway_instance_controller.go
@@ -69,6 +69,11 @@ func (r *GatewayInstanceReconciler) Reconcile(ctx context.Context, req kube_ctrl
 		return kube_ctrl.Result{}, errors.Wrap(err, "unable to get Namespace of MeshGatewayInstance")
 	}
 
+	if len(gatewayInstance.Spec.Tags) > 0 && gatewayInstance.Spec.Tags[mesh_proto.ServiceTag] != "" {
+		r.Log.Info("WARNING: Setting kuma.io/service tag is deprecated for this object kind, control-plane creates a name based on the resource name and namespace",
+			"name", gatewayInstance.GetName(), "namespace", gatewayInstance.GetNamespace(), "kind", gatewayInstance.GetObjectKind().GroupVersionKind().Kind)
+	}
+
 	mesh := k8s_util.MeshOfByLabelOrAnnotation(r.Log, gatewayInstance, &ns)
 
 	orig := gatewayInstance.DeepCopy()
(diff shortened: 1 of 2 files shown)


pkg/plugins/runtime/k8s/controllers/gateway_instance_controller.go
```diff
@@ -101,18 +106,15 @@ func k8sSelector(name string) map[string]string {
 	return map[string]string{"app": name}
 }
 
-func (r *GatewayInstanceReconciler) resolveGatewayInstanceServiceTag(gatewayInstance *mesh_k8s.MeshGatewayInstance) {
+func (_ *GatewayInstanceReconciler) gatewayInstanceTags(gatewayInstance *mesh_k8s.MeshGatewayInstance) map[string]string {
```
bartsmykla, line 109: Why not changing it to regular function?

jakubdyszkiewicz: controllers is quite a big package. I wanted to keep the scope of a function on the reconciller only. I can change it if you think it's important

lukidzi (approved):

bartsmykla (approved):

pkg/plugins/runtime/k8s/controllers/gateway_instance_controller.go
bartsmykla: No, I don't think it's necessary

PR #10077, opened 2024-04-25
test(e2e): assert cp did not crash and print previous cp logs

jakubdyszkiewicz:
### Checklist prior to review

We want to print previous logs in case a CP crashed. We also want to assert that CP did not crash

This way we would catch the problem with MeshGateway


jakubdyszkiewicz:
waiting for merge of 2.7 branch to master to pass the test

slonka (approved):
Looks good but we need to figure out why meshaccesslog causes restarts

test/framework/envs/kubernetes/env.go
```diff
@@ -97,3 +97,8 @@ func PrintKubeState(report ginkgo.Report) {
 		}
 	}
 }
+
+func ExpectCpToNotCrash() {
+	restartCount := framework.RestartCount(Cluster.GetKuma().(*framework.K8sControlPlane).GetKumaCPPods())
+	Expect(restartCount).To(Equal(0))
```
lahabana, line 103:
```suggestion
	Expect(restartCount).To(Equal(0), "CP restarted in this suite, this should not happen.")
```

This will make the output easier to understand than: 
```
  [FAILED] Expected
      <int>: 3
  to equal
      <int>: 0
```

lahabana:
This is great! And may explain a flake we've been having!

jakubdyszkiewicz:
> Looks good but we need to figure out why meshaccesslog causes restarts

IMHO It's mesh gateway crash fixed in release. That's why I commented that I'm waiting for master

jakubdyszkiewicz:
> Also it's odd there's nothing: "No files were found with the provided path: /tmp/e2e-debug/. No artifacts will be uploaded."

hmm, where do you see it? This should be ignored. Yeah, debug files are not in all suits (old e2e)

Issue #10074, opened 2024-04-25
Improve verification if Gateway API CRDs are present in a cluster


bartsmykla:
### What happened?

Currently to check if CP should support Gateway API, we are checking if there is `Gateway` CRD with `gateway.networking.k8s.io` group present:
https://github.com/kumahq/kuma/blob/5a2d836dc6e556f636fdf8cdc533876aff08d888/pkg/plugins/runtime/k8s/plugin_gateway.go#L24-L36
This is insufficient in situation when some of the Gateway API CRDs are missing (i.e. package with CRDs was installed before `ReferenceGrant` was added), which causes CP to start throwing lot of errors:
```
controller-runtime.source.EventHandler	if kind is a CRD, it should be installed before calling Start	{"kind": "ReferenceGrant.gateway.networking.k8s.io", "error": "no matches for kind \"ReferenceGrant\" in version \"gateway.networking.k8s.io/v1beta1\""}
```

michaelbeaumont:
Just remember if there's no referencegrant installed we need to keep cross namespace references impossible.

jakubdyszkiewicz:
Should we check if we have a supported version or later? The problem was that a user had 0.6.0-dev version installed.

bartsmykla:
I don't think we have to check exact versions. My reasoning:
- According to Gateway API docs, [changes to the standard channel are backward compatible](https://gateway-api.sigs.k8s.io/guides/implementers/#changes-to-the-standard-channel-crds-are-backwards-compatible)
  - I don't think schema of resources we are using will ever change, even in Experimental channel
- We require `GatewayClass`, `Gateway`, `HTTPRoute` and `ReferenceGrant` to be present, all of which are part of Standard channel `v1`, so we should be quite sure they will work in any newer version of Gateway API

michaelbeaumont:
Yes, though [breaking changes can come on experimental](https://gateway-api.sigs.k8s.io/guides/#install-experimental-channel). ReferenceGrant isn't in v1 and there's a [good chance it moves to a different API group](https://gateway-api.sigs.k8s.io/api-types/referencegrant/?h=referenc#potential-future-api-group-change) so there may be an argument for supporting gateway API without it, though we can evaluate that as needed.

PR #10079, opened 2024-04-25
fix(gatewayapi): validate presence of all required Gateway API resources

bartsmykla:
Existing check of presence for only `Gateway` resource is insufficient in situation when some (not all) of the Gateway API CRDs are missing (i.e. package with CRDs was installed before `ReferenceGrant` was added), which causes CP to start throwing lot of errors:

```
controller-runtime.source.EventHandler	if kind is a CRD, it should be installed before calling Start	{"kind": "ReferenceGrant.gateway.networking.k8s.io", "error": "no matches for kind \"ReferenceGrant\" in version \"gateway.networking.k8s.io/v1beta1\""}
```

This is the simplest solution as we could also verify if `ReferenceGrant` is present, and if not to disable cross-mesh references, but it would be unnecesarily complex, especially in situation when Gateway API is GA and contains `ReferenceGrant` in `v1`.

Closes: https://github.com/kumahq/kuma/issues/10074

### Checklist prior to review

  - https://github.com/kumahq/kuma/issues/10074
  - It won't
  - There is no need

### DIFF
--- a/pkg/plugins/runtime/k8s/plugin_gateway.go
+++ b/pkg/plugins/runtime/k8s/plugin_gateway.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/pkg/errors"
+	"golang.org/x/exp/maps"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
 	kube_ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


jakubdyszkiewicz:
can you make it to release 2.7? so instead of backport we will merge release to master?

michaelbeaumont (approved):

PR #10096, opened 2024-04-26
chore(deps): update go to 1.22.5

jakubdyszkiewicz:
### Checklist prior to review

Update Go. Dependabot PRs are failing because it's trying to add toolchain directive and golangci-lint is missing file. I think we need to add this directive as well, but let's first upgrade Go.


bartsmykla (dismissed):
We can't do it for now:
- https://github.com/kumahq/kuma/pull/9443
- https://github.com/golang/go/issues/65991

slonka:
> We can't do it for now:

Should we put it in the comments in go.mod / CI config that we can't upgrade yet? I forgot about this and if Bart did not review I would've probably approved this.

go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/kumahq/kuma
 
-go 1.21.11
+
+go 1.22.5
```
Icarus9913, line 4: remove the blank line

api/mesh/v1alpha1/traffic_route.proto
```diff
@@ -222,6 +219,5 @@ message TrafficRoute {
   }
 
   // Configuration for the route.
-  Conf conf = 3
-      [ (validate.rules).message.required = true ];
+  Conf conf = 3 [ (validate.rules).message.required = true ];
```
michaelbeaumont, line 222: This is weird right? Why is this changing exactly? I don't think it needs to be in this PR, I don't seem to get this running `make check` locally.

michaelbeaumont (changes requested):

api/mesh/v1alpha1/traffic_route.proto
slonka: yeah, make format is different on CI and my local machine :|

michaelbeaumont: I'm reverting the commit to see if it passes CI.

michaelbeaumont (approved):

api/mesh/v1alpha1/traffic_route.proto
michaelbeaumont: https://github.com/kumahq/kuma/pull/10742

Automaat (approved):

Issue #8216, opened 2023-10-31
rework AddFieldsFromCtx to be less combersome


lahabana:
### What happened?

https://github.com/kumahq/kuma/pull/8153#discussion_r1376639540

The api is not great. logr now supports slog and slog can take `InfoWithContext` maybe that'd be a better way to add these: https://github.com/go-logr/logr#using-a-sloghandler-as-backend-for-logr

slonka:
Closing as rotten - this has been inactive for a long time with no progress. Feel free to reopen if someone wants to pick this up.

PR #10100, opened 2024-04-27
fix(api-server): fix trace/span ID processing in logs

bartsmykla:
## Problem:

Currently, processing trace/span IDs before logging them (e.g., converting them to Datadog format) requires calling [`NewSpanLogValuesProcessorContext`](https://github.com/kumahq/kuma/blob/dd95f22e4e568ebeb899c187b107020a28703db5/pkg/plugins/extensions/logger/context.go#L19-L24) with the `runtime.Extensions()` context. However, this context isn't always available, particularly when [handling errors in REST endpoint handlers through `AddFieldsFromCtx`](https://github.com/kumahq/kuma/blob/dd95f22e4e568ebeb899c187b107020a28703db5/pkg/log/logger.go#L120). This inconsistency leads to the processing function being ignored.

## Proposed Solutions:

- Pass `runtime.Extensions()` Context to [`HandleError`](https://github.com/kumahq/kuma/blob/8d2eaed06c21ad74b6806ce1db837f6610dcaaee/pkg/core/rest/errors/error_handler.go#L27): While effective, this approach adds significant boilerplate code and requires disabling the `contextcheck` linter in multiple locations (see PR #10094).

- Leverage slog's Contextual Logging (ref: https://github.com/kumahq/kuma/issues/8216): This approach offers cleaner code but requires significant changes to our logging infrastructure (ones which I immediately faced when tried to do a quick POC):
  - slog lacks the concept of logger names, necessitating a custom `slog.Handler` for maintaining desired log structure.
  - We'd need to abandon using globally configured/named loggers or potentially introduce deferred logger initialization logic for slog.

## Current Solution (For This PR):

To address the issue without a major logging overhaul, this PR proposes a simpler solution:

- Pass the logger instance through the context when constructing ApiServer.
- Restore the logger from the context in `HandleError` function.

This approach, while not ideal due to the passing logger via context, effectively resolves the problem without introducing complex logging changes.

## Additional changes:

I adjusted the signature of `NewApiServer` function to accept the whole `runtime.Runtime`, which reduced significantly the amount of necessary arguments, which were attained from the `runtime.Runtime`. It makes `contextcheck` linter happy as we don't have to pass `runtime.Extensions()` context as an argument, which linter would flag as passed and not used, without realising it's not the regular context.

### Checklist prior to review

  - No relevant issues
  - It won't
  - Manually tested by running local OTEL collector and then manually triggering handlers to return error with ant without span logging processors
  - There is no need
  - There is no need I think 🤔 

### DIFF
--- a/pkg/api-server/server.go
+++ b/pkg/api-server/server.go
@@ -20,13 +20,13 @@ import (
 
 	"github.com/bakito/go-log-logr-adapter/adapter"
 	"github.com/emicklei/go-restful/v3"
+	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	http_prometheus "github.com/slok/go-http-metrics/metrics/prometheus"
 	"github.com/slok/go-http-metrics/middleware"
 	"go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful"
 
 	"github.com/kumahq/kuma/pkg/api-server/authn"
-	"github.com/kumahq/kuma/pkg/api-server/customization"
 	"github.com/kumahq/kuma/pkg/api-server/filters"
 	api_server "github.com/kumahq/kuma/pkg/config/api-server"
 	kuma_cp "github.com/kumahq/kuma/pkg/config/app/kuma-cp"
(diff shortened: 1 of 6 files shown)


pkg/api-server/server.go
```diff
@@ -227,7 +224,20 @@ func NewApiServer(
...
+				core.Log.WithName("rest"),
+				request.Request.Context(),
+				rt.Extensions(),
+			),
+		))
+
+		chain.ProcessFilter(request, response)
+	})
+
+	rt.APIInstaller().Install(container)
```
slonka, line 240: (just reposting Barts comments on slack): this is where we put the logger in

pkg/core/rest/errors/error_handler.go
```diff
@@ -20,12 +20,12 @@ import (
 	"github.com/kumahq/kuma/pkg/core/validators"
 	"github.com/kumahq/kuma/pkg/intercp/envoyadmin"
 	kds_envoyadmin "github.com/kumahq/kuma/pkg/kds/envoyadmin"
-	kuma_log "github.com/kumahq/kuma/pkg/log"
 	"github.com/kumahq/kuma/pkg/multitenant"
 )
 
 func HandleError(ctx context.Context, response *restful.Response, err error, title string) {
-	log := kuma_log.AddFieldsFromCtx(core.Log.WithName("rest"), ctx, context.Background())
+	log := logr.FromContextOrDiscard(ctx)
```
slonka, line 27: and this is where we get it out

slonka (approved):
I'm approving this but I might be biased, I'll let someone else also review this

michaelbeaumont (approved):

PR #10134, opened 2024-04-29
fix(k8s): use EndpointSlices to determine Service endpoints

michaelbeaumont:
We should completely skip the checking via selectors and make sure this is efficient enough.

### Checklist prior to review


lahabana:
Why this change?

michaelbeaumont:
@lahabana it's a missing feature. The impetus was that there are new Gateway API tests that test forwarding to Services with manually managed EndpointSlices

pkg/plugins/runtime/k8s/util/util_test.go
```diff
@@ -98,15 +98,15 @@ var _ = Describe("Util", func() {
 	DescribeTable("FindServices",
```
bartsmykla, line 98:
```suggestion
	DescribeTable("MatchService",
```

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -382,15 +426,52 @@ func (r *PodReconciler) SetupWithManager(mgr kube_ctrl.Manager, maxConcurrentRec
...
+				return nil
+			}
+			var req []kube_reconcile.Request
+			for _, slice := range endpointSlices.Items {
+				l = l.WithValues(
+					"EndpointSlice",
+					kube_types.NamespacedName{Namespace: slice.Namespace, Name: slice.Name},
+				)
+				for _, endpoint := range slice.Endpoints {
+					l = l.WithValues("Endpoint", endpoint)
```
slonka, line 463: nitpick: we're overwriting this logger but we never use it inside these 2 loops, I guess you used it for debugging and then cleaned up the debug log, right? should we restore it on a debug level?

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -382,15 +426,52 @@ func (r *PodReconciler) SetupWithManager(mgr kube_ctrl.Manager, maxConcurrentRec
...
+							NamespacedName: kube_types.NamespacedName{Name: endpoint.TargetRef.Name, Namespace: endpoint.TargetRef.Namespace},
+						})
+					}
+				}
+			}
+			return req
+		}
+
+		if err := client.List(ctx, pods, kube_client.InNamespace(obj.GetNamespace()), kube_client.MatchingLabels(svc.Spec.Selector)); err != nil {
 			l.WithValues("service", obj.GetName()).Error(err, "failed to fetch Pods")
```
slonka, line 475: this doesn't do anything right? because we never use this new logger right? I know it's not code from this PR but seems just wrong... also it will contain `EndpointSlice` and `Endpoint` from last loop cycles which is not right 🤔

slonka (approved):
logging seems a bit off, otherwise looks fine

pkg/plugins/runtime/k8s/controllers/pod_controller.go
michaelbeaumont: fixed the unused loggers but this calls `Error` so it's important

pkg/plugins/runtime/k8s/controllers/pod_controller.go
michaelbeaumont: I just removed it

slonka (approved):

jakubdyszkiewicz (approved):

Issue #7269, opened 2023-07-17
MADR for applying new policies on different namespaces


jakubdyszkiewicz:
### Description

Right now new policies are restricted to kuma-system namespace.
We want to be able to apply policies on other namespaces so they affect only specific namespace.

PR #10148, opened 2024-04-30
docs(MADR): policy on namespace

lobkovilya:
### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/7269


docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+    Tags map[string]string `json:"tags,omitempty"`
+    // Mesh is reserved for future use to identify cross mesh resources.
+    Mesh string `json:"mesh,omitempty"`
+    // ProxyTypes specifies the data plane types that are subject to the policy. When not specified,
+    // all data plane types are targeted by the policy.
+    // +kubebuilder:validation:MinItems=1
+    ProxyTypes []TargetRefProxyType `json:"proxyTypes,omitempty"`
+    // Namespace of the referenced resource. If not specified, then equal the value is equal to the resource's namespace
+    // where this TargetRef is used.
+    Namespace string `json:"namespace,omitempty"`
```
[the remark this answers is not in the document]
lobkovilya: It wasn't possible to create policies on the namespace before. So I guess we are not constrained by the backward compatibility here

docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+
+Since we didn't make any decisions on how to target MeshService in policies we will leave this out of now. We will
+revisit this while discussing targeting MeshService in policies.
+
+### MeshGateway reference
+
+`MeshGateway` at this moment is not namespace-scoped. It can change in the future with
+the [MeshBuiltinGateway resource](https://github.com/kumahq/kuma/issues/10014).
+
+We've decided that we won't allow targeting `MeshGateway` in namspace scoped polcies except for the
```
jakubdyszkiewicz, line 355: it's fine by me, but I'm just curious. Why only MeshHTTPRoute and MeshTCPRoute. How are those different than policies?

docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+
+## Implementation
+
+### Namespace-scoped policies and multizone
+
+Namespace-scoped policies can be applied to a custom namespace only on zones as there are simply no DPPs and custom
+namespaces on Global.
+
+After a user applied a policy on zone's custom namespace, it automatically gets a `k8s.kuma.io/namespace` label.
+The label is part of the hash suffix when policy is synced to global.
```
jakubdyszkiewicz, line 46: just to double check. There are no plans to convert targetRef implicitly or explicitly for example `targetRef:Mesh` in a namespace x to `targetRef:MeshSubset,namespace:x`. We are planning to add additional logic in policy matching. Is that correct?

docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+    name: backend-route
+  to:
+    - targetRef:
+        kind: Mesh
+      default:
+        http:
+          requestTimeout: 5s
+```
+
+**2. MeshTimeout in "kuma-system", MeshHTTPRoute in "frontend-ns"**
```
jakubdyszkiewicz, line 181: Is there anything in the design that would close up an opportunity to implement this in a future if we were to find use cases for it?

docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+    - targetRef:
+        kind: Mesh
+      default:
+        http:
+          requestTimeout: 3s
+```
+
+This example also demonstrates the scenario when 2 MeshTimeout reference the same route,
+the namespaced MeshTimeout is considered more specific.
+That's why the final request timeout value should be 3s.
```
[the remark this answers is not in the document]
jakubdyszkiewicz: That's the biggest question of this MADR.

It is a bit strange that the service owner can advertise their service using their own namespace, but cannot advertise routes to reference in the policies

docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+    // +kubebuilder:validation:MinItems=1
+    ProxyTypes []TargetRefProxyType `json:"proxyTypes,omitempty"`
+    // Namespace of the referenced resource. If not specified, then equal the value is equal to the resource's namespace
+    // where this TargetRef is used.
+    Namespace string `json:"namespace,omitempty"`
+}
+```
+
+Let's try to review all possible types of references between policies and see if maybe some of them don't make sense.
+We have 3 namespaces: `kuma-system`, `frontend-ns` and `backend-ns`. We have 2 policies: `MeshTimeout`
```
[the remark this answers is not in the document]
lobkovilya: They don't call their namespace as `producer` and `consumer`, these are just terms. From what I see they have examples with `faces` and `fast-clients` namespaces https://gateway-api.sigs.k8s.io/mesh/. TBH I think `frontend-ns` and `backend-ns` are easy enough, no?

docs/madr/decisions/046-policy-on-namespace.md
lobkovilya: Nothing I can think of, we can always add support for this use case

docs/madr/decisions/046-policy-on-namespace.md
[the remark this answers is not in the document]
lobkovilya: Good point!

So if Backend SO can create `MeshService` in `backend-ns` and advertise it to all other namespaces, should this be the case for MeshHTTPRoutes in `backend-ns`?

Reasons why it makes sense:
* we can think of routes as "a more granular destination than MeshService"
* we already want to do this for namespace-scoped MeshHTTPRoutes that target MeshGateaway
* better mapping to GAMMA producer and consumer routes

Reasons why it doesn't make sense:
* all other namespace-scoped policies affects only pods in the namespace they're applied, it won't be the case for MeshHTTPRoute (or MeshTCPRoute)

@Automaat @michaelbeaumont I'm curious to hear your opinion on this

docs/madr/decisions/046-policy-on-namespace.md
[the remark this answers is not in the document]
lahabana: So the comment is a little weird. Because there are currently policies and they have empty namespace and live in the system namespace. This means that they'd match things in the system namespace no?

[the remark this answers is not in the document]
lobkovilya: btw there is a typo in the comment, should be without "equal":
> Namespace of the referenced resource. If not specified, then the value is equal to the resource's namespace where this TargetRef is used.

If today you have a MeshTimeout that references MeshHTTPRoute with an empty namespace like:
```yaml
kind: MeshTimeout
metadata:
  name: timeout-1
  namespace: kuma-system
spec:
  targetRef:
    kind: MeshHTTPRoute
    name: route-1
```
then Kuma searches `route-1` in `kuma-system`. And after the upgrade to 2.8 it won't change, as the default value for `spec.targetRef.namespace` is `metadata.namespace`.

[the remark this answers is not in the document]
lahabana: Yes but what about:

```yaml
kind: MeshTimeout
metadata:
  name: timeout-1
  namespace: kuma-system
spec:
  targetRef:
    kind: Mesh
```
?

This should match all dpps regardless of location no?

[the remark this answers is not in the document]
michaelbeaumont: A Mesh is cluster scoped. The only case we have to worry about not breaking is `namespace: kuma-system` `refers to` `namespace: ""` and either the referred to object is cluster scoped, in which case namespace never makes sense, or it was namespace scoped, in which case it must have been in `kuma-system` so nothing changes.

docs/madr/decisions/046-policy-on-namespace.md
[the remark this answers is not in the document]
michaelbeaumont: > So if Backend SO can create MeshService in backend-ns and advertise it to all other namespaces, should this be the case for MeshHTTPRoutes in backend-ns?

I think so. I see `kuma-system` as a sort of crutch because we didn't have namespaced policies figured out. Ultimately it shouldn't "need" to be used.

> all other namespace-scoped policies affects only pods in the namespace they're applied

I don't think this is an especially meaningful idea. Like Jakub said. Yes, the namespace-scoped policies affect Pods of other namespaces, but doesn't even the creation of Pods in a given namespace affect Pods of other namespaces?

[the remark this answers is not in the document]
michaelbeaumont: Put differently, what the namespapce-scoped policy affects is _requests_ that would be routed to the services of that namespace. I think this is a more useful abstraction than "does this affect the Envoy config of my Pod?"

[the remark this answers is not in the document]
jakubdyszkiewicz: > I don't think this is an especially meaningful idea

I think it is. For me it's easier to think about policies and impact with this model in mind.

Maybe the problem here is that MeshHTTPRoute is not really a policy, just like a Service is not a policy.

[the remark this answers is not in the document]
michaelbeaumont: But how do you define only "affecting Pods" in a namespace exactly? In light of the fact that I can simply create a Service in another namespace in order to affect them.

I mean at this point we're basically just playing a semantic game where we define a policy as a thing that only affects Envoy config of Pods in its namespace. And a bunch of other things affect Envoy config of other namespaces and those are "not policies". Starting to feel like the term isn't especially useful.

docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+        kind: Mesh
+      default:
+        http:
+          requestTimeout: 3s
+```
+
+**5. MeshTimeout in "frontend-ns", MeshHTTPRoute in "backend-ns"**
+
+MeshHTTPRoute in `backend-ns` produces envoy routes in DPPs in `backend-ns`.
+Ability to configure timeouts on these routes from the `frontend-ns` breaks the namespace isolation and seems
```
michaelbeaumont, line 337: I think this is my ideal workflow for implementing 3., where `backend-ns` has some authority over requests to its service from anywhere in the mesh and `frontend-ns` can override that for its own Services.

docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+
+Since we didn't make any decisions on how to target MeshService in policies we will leave this out of now. We will
+revisit this while discussing targeting MeshService in policies.
+
+### MeshGateway reference
+
+`MeshGateway` at this moment is not namespace-scoped. It can change in the future with
+the [MeshBuiltinGateway resource](https://github.com/kumahq/kuma/issues/10014).
+
+We've decided that we won't allow targeting `MeshGateway` in namspace scoped polcies except for the
```
michaelbeaumont, line 355: I would say policies created in the namespace of a Service should affect requests from the MeshGateway to that service.

docs/madr/decisions/046-policy-on-namespace.md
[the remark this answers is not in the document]
jakubdyszkiewicz: It's just we've been here with `sources & destinations`. It was not clear how to affect what and where. Maybe you are right, maybe we need to ditch "all other namespace-scoped policies affects only pods in the namespace they're applied" rule, but I'd be curious to hear what's the alternative so that it is still sane to understand what is matching where.

> I propose that a service owner should never need the mesh operator/access to kuma-system to influence requests to its Services.

That's a _recommendation use-case_.
That's reasonable as long as clients still have a way of overriding this.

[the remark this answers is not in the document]
michaelbeaumont: > That's reasonable as long as clients still have a way of overriding this.

Definitely, the other piece would be if I create a policy in a namespace that targets Services of _different_ namespaces, it affects only requests from the policy's namespace.

docs/madr/decisions/046-policy-on-namespace.md
Automaat: You can target MeshHttpRoute from your namespace, to configure this. 

Allowing to apply policies with MeshGateway as top level targetRef could be problematic. For example, who should be able to configure policy with `to` section that targets Mesh? Example:

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTimeout
metadata:
  name: timeout-global
  namespace: custom-ns
  labels:
    kuma.io/mesh: default
spec:
  targetRef:
    kind: MeshGateway
  to:
  - targetRef:
      kind: Mesh
```

with this you are creating policy that is modifying traffic to services from all namespaces, not only yours

michaelbeaumont: No, you'd be modifying traffic to services from only your namespace.

michaelbeaumont: To be clear, I don't think seeing `kind: Mesh` is ideal there, but that's the semantics I'm suggesting.

The point is, we have Kubernetes RBAC already and it maps very well to service ownership. It'd be a shame not to take advantage of it for policies.

michaelbeaumont: Maybe the better way would be `spec.targetRef.kind: AllMeshServices` and `from[].targetRef.kind: MeshGateway`? Either way, the use case is that as the owner `custom-ns`, I should be able to influence requests from a gateway to my services.

Automaat: I agree that it would be nice to configure these policies in your namespace, so the solution at the moment would be to disallow users setting `to: Mesh` on MeshGateway when policy is configured on custom namespace?

With `to: Mesh` allowed, there could be policies on multiple namespaces. Which policy should be applied? Should we use lexicographical order?

michaelbeaumont: Would it work to translate `kind:Mesh` into `{kind: MeshSubset, tags: {k8s.kuma.io/namespace: custom-ns}}`?

docs/madr/decisions/046-policy-on-namespace.md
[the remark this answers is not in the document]
Automaat: After reading all that comments and revisiting GAMMA docs, I understand why it would be nice to not use system policy at all. 

So the precedence of applying policies will look like: `kuma-system -> different-ns -> my-ns`

I think that there are two different use cases, first with routes and targeting routes, and the second is rest of the policies.

## First use case (routes):
Example producer route with default timeout, created in `backend-ns`:
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshHTTPRoute
metadata:
  name: backend-route
  namespace: backend-ns
spec:
  targetRef:
    kind: MeshService
    name: frontend
  to:
    - targetRef:
        kind: MeshService
        name: backend
        namespace: backend-ns
--- 
apiVersion: kuma.io/v1alpha1
kind: MeshTimeout
metadata:
  name: frontend-to-backend
  namespace: backend-ns
spec:
  targetRef:
    kind: MeshHTTPRoute
    name: backend-route
    namespace: backend-ns 
  to:
    - targetRef:
        kind: Mesh
      default:
        http:
          requestTimeout: 5s
```

Then frontend SO can override this timeout using:
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTimeout
metadata:
  name: frontend-to-backend
  namespace: frontend-ns # I would like to override default set by producer
spec:
  targetRef:
    kind: MeshHTTPRoute
    name: backend-route
    namespace: backend-ns 
  to:
    - targetRef:
        kind: Mesh
      default:
        http:
          requestTimeout: 2s
```

## Policies not targeting routes

As a backend SO, I would like to create default MeshRetry for requests to my service:

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshRetry
metadata:
  name: all-to-backend-retry-http
  namespace: backend-ns
spec:
  targetRef:
    kind: Mesh
    name: default
  to:
    - targetRef:
        kind: MeshService
        name: backend
        namespace: backend-ns
      default:
        http:
          numRetries: 5
``` 
and then frontend SO would like to override this default:
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshRetry
metadata:
  name: to-backend-retry-http
  namespace: frontend-ns
spec:
  targetRef:
    kind: Mesh
    name: default
  to:
    - targetRef:
        kind: MeshService
        name: backend
        namespace: backend-ns
      default:
        http:
          numRetries: 2
```
Should this work? Should first policy be applied to dpps in different namespaces? Or should backend SO create it in `kuma-system` namespace?

@michaelbeaumont Does this make sense? And is it as you would expect it to be?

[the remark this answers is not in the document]
michaelbeaumont: Yeah it does.

> kuma-system -> different-ns -> my-ns

To be clear I was imagining `different-ns` overriding `my-ns` overriding `kuma-system`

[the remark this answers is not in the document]
Automaat: Why should `different-ns` override `my-ns`? Shouldn't my config, that I just created in my namespace, be most important? 

And there is question about second use case, should it work? 

Also, @jakubdyszkiewicz it would be nice to hear your opinion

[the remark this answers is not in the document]
michaelbeaumont: > Why should different-ns override my-ns? Shouldn't my config, that I just created in my namespace, be most important?

I think it depends on whether routes are meant to be defaults or authoritative.

docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+          requestTimeout: 5s
+```
+
+**2. MeshTimeout in "kuma-system", MeshHTTPRoute in "frontend-ns"**
+
+Frontend Service Owner creates a MeshHTTPRoute in the `frontend-ns` namespace.
+What are the reasons for Mesh Operator to create MeshTimeout in `kuma-system` and reference the MeshHTTPRoute policy
+in `frontend-ns`?
+Mesh Operator should work with policies that apply globally, knowing specifics of the `frontend` app seems incorrect.
+We didn't find good use cases why this type of reference would make sense.
```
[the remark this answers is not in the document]
Automaat: it wil result in error, I've updated madr

docs/madr/decisions/046-policy-on-namespace.md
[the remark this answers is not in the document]
lobkovilya: @Automaat when you say:

> So the precedence of applying policies will look like: kuma-system -> different-ns -> my-ns

are you saying "my" from the point of view of producer (Backend SO) or consumer (Frontend SO)?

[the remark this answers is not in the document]
Automaat: Exactly as Mike explained

docs/madr/decisions/046-policy-on-namespace.md
```diff
@@ -0,0 +1,456 @@
...
+3. **[new step]** Policy in a custom namespace is more specific than a policy in the `kuma-system` namespace.
+   When policies applied in the same namespace, go to the next step.
+4. The policy is more specific if it's name is lexicographically less than other policy's name ("aaa" < "bbb" so that "
+   aaa" policy is more specific)
+
+### Cross-policy references
+
+Kuma policies support cross-policy references.
+At this moment, it works only between MeshTimeout and MeshHTTPRoute, but there are plans to support it for other
+policies, i.e. [#6645](https://github.com/kumahq/kuma/issues/6645).
```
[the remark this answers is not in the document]
lobkovilya: Yeah, unfortunately not the case at this moment

docs/madr/decisions/046-policy-on-namespace.md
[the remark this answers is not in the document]
lobkovilya: So if we say that `my-ns` is service owner's ns and `different-ns` is basically consumer's ns, then I agree with Mike – `different-ns` should override `my-ns`. This relative naming is a bit confusing but I feel like we're all on the same page here :)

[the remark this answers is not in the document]
Automaat: There was a mix up in the naming, but in the end I think we all agree, I was thinking from a perspective of consumer, owner of `my-ns` so `different-ns` was namespace where service I want to consume, route and default timeout was configured.

docs/madr/decisions/046-policy-on-namespace.md
Automaat: I've updated MADR as discussed

docs/madr/decisions/046-policy-on-namespace.md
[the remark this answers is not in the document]
lobkovilya: I updated MADR as we discussed

docs/madr/decisions/046-policy-on-namespace.md
lobkovilya: @Automaat added some explanation here https://github.com/kumahq/kuma/pull/10148/files#diff-cb6c27d1ae4efdd51ae3730849991f19c4016f47fb91292fa349d9c0e0ea8a95R318-R320, do it make sense?

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,527 @@
...
+It can be easily replaced with the top-level `targetRef{kind:MeshSubset,tags{"kuma.io/zone":"my-zone"}}` when user wants to opt-out.
+
+##### Pros and Cons of automatically detecting if a policy needs to be exported
+
+Having `spec.export.mode` might be excessive taking into account we already have a top-level targetRef.
+If the top-level targetRef specifies a tag `kuma.io/zone` then it's clear whether we need to export the policy or not.
+
+* Good, because we don't introduce new fields, no need for users to learn new concepts.
+* Good, because all consumers are treated equally by default without favoring the local ones.
+* Bad, because it's not backward compatible.
```
jakubdyszkiewicz, line 264: also bad, because if you want to keep this only to your local zone, you need to specify the name of the zone in the policy

jakubdyszkiewicz (approved):

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+6. The policy is more specific if it's name is lexicographically less than other policy's name ("aaa" < "bbb" so that "
+   aaa" policy is more specific)
+
+### Cross-policy references
+
+Kuma policies support cross-policy references.
+At this moment, it works only between MeshTimeout and MeshHTTPRoute, but there are plans to support it for other
+policies, i.e. [#6645](https://github.com/kumahq/kuma/issues/6645).
+Referencing a policy from another namespace or another zone should be done in a similar way MeshService is referenced.
+Fields `name/namespace` and `labels` are mutually exclusive.
```
lahabana, line 293: I think this: https://github.com/kumahq/kuma/issues/10247 suggests an alternative more straightforward approach in this case that may simplify a lot of things

lahabana (commented):
While thinking about: https://github.com/kumahq/kuma/issues/10247 and thinking about syncing to zones with producer/consumer policies it struck me that you could have multiple `to` entries with different kinds (one could be producer, the other consumer). What would be the syncing behaviour in this case?

docs/madr/decisions/046-policy-on-namespace.md
lobkovilya: It will be like that after we updated MADR

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+* Good, because we have time to work on other features
+* Bad, because cluster operator has to grant write permissions to `kuma-system` namespace to everyone who works with
+  policies
+* Bad, because low isolation between teams (team-a can unintentionally break polices of the team-b)
+
+## Implementation
+
+Applying a policy on the custom namespace **affects requests to the pods in the custom namespace**.
+Since some policies are applied on the client-side, the scope of policy's effect is not limited to the namespace
+where it was applied.
```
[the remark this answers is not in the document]
lobkovilya: Why? The first sentence says policy in the custom namespaces affects **requests** to the custom namespace (not pods in the custom namespace). And the second sentence just elaborates – since some policies are applied on the client-side the policy scope **is no longer limited** to policy's namespace.

lobkovilya:
> While thinking about: https://github.com/kumahq/kuma/issues/10247 and thinking about syncing to zones with producer/consumer policies it struck me that you could have multiple to entries with different kinds (one could be producer, the other consumer). What would be the syncing behaviour in this case?

The new validation rules are going to limit what policies you can create. If top-level targetRef doesn't specify `kuma.io/zone` and `k8s.kuma.io/namespace` then **each** item in `to[]` has to specify both `kuma.io/zone` and `k8s.kuma.io/namespace`. That's kind of a downside we're having with new design – consumer policies must have `kuma.io/zone` and `k8s.kuma.io/namespace` tags in top-level targetRef, i.e Frontend Owner wants to change how frontend communicates with others:

```yaml
metadata:
  namespace: frontend-ns
  labels:
    kuma.io/origin: zone
spec:
  targetRef:
    kind: MeshSubset
    tags:
      kuma.io/zone: i-have-to-know-the-name-of-my-zone
      k8s.kuma.io/namespace: frontend-ns
  to:
    - targetRef:
        kind: Mesh
      default: {}
```

docs/madr/decisions/049-policy-on-namespace.md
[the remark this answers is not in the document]
lahabana: I think I didn't see requests. I had to reread this 3 times, I think it could be made more clear

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+```
+
+Important to note, even though policy in `backend-ns` changes proxy configurations in `frontend-ns` 
+the top-level targetRef still servers the purpose of showing what proxy configurations are going to change.
+Policy's namespace **does not** implicitly turn the top-level targetRef to `MeshSubset{tags:{"kuma.io/namespace":"backend-ns"}}`.
+
+### Namespace-scoped policies and multizone
+
+Namespace-scoped policies can be applied to a custom namespace only on zones as there are simply no DPPs and custom
+namespaces on Global.
```
lahabana, line 68: This will be enforced by validation

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+
+After a user applied a policy on zone's custom namespace, the webhook automatically adds a `k8s.kuma.io/namespace` label to the policy.
+The label is part of the hash suffix when policy is synced to global.
+
+As mentioned previously, a policy applied in producer namespace can affect envoy configs of consumers.
```
lahabana, line 73: This is the first ocurence of the work "producer" so it's not "mentioned previously"

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+Consumers can reside not only in different namespaces, but also in different zones.
+That's why, the policy applied in `backend-ns` in `zone-1` has to be synced to global and then to other zones.
+With the increased blast radius of the policies comes great responsibility, 
+we don't want small mistake in the policy to break traffic on the other side of the world.
+A few things could be done to protect users from impactful mistakes:
+* Enhanced validation. If policy is created in `backend-ns` of the `zone-1` and the top-level targetRef is `Mesh`,
```
lahabana, line 79: Hmm why? I'm confused I think I need an example to understand this

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+
+As mentioned previously, a policy applied in producer namespace can affect envoy configs of consumers.
+Consumers can reside not only in different namespaces, but also in different zones.
+That's why, the policy applied in `backend-ns` in `zone-1` has to be synced to global and then to other zones.
+With the increased blast radius of the policies comes great responsibility, 
+we don't want small mistake in the policy to break traffic on the other side of the world.
+A few things could be done to protect users from impactful mistakes:
+* Enhanced validation. If policy is created in `backend-ns` of the `zone-1` and the top-level targetRef is `Mesh`, 
+then `to[].targetRef` should have both `k8s.kuma.io/namespace: backend-ns` and `kuma.io/zone: zone-1`. 
+* Ability to opt-in or opt-out from syncing the policy to other zones.
```
lahabana, line 81: I don't think that's necessary you can always use a top `spec.tagerRef.kind: MeshSubset` to reduce blast radius no? That's what it is for no?

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+      default:
+        connectionTimeout: 5s # why does the Backend Service Owner try to change the way clients use the 'redis' service?
+```
+
+Such policy should not be allowed, if Backend Service Owner needs to change the way clients use `redis`
+they have to put the policy into `kuma-system` namespace by checking with Mesh Operator first.
+
+The validation rules for namespace-scoped policy are the following:
+* to-policy when placed into the custom namespace must have the namespace and zone either in the top-level targetRef or in each to-item
+* from-policy when placed into the custom namespace must always have the namespace and zone in the top-level targetRef
```
lahabana, line 111: That seems unnecessarily annoying. Again UX is important here!

lahabana: Isn't `from` policy on custom namespace having an "implicit MeshSubset" a better behaviour?  Especially if we go for https://github.com/kumahq/kuma/issues/10247 top level targetRef slowly disappears.

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+          k8s.kuma.io/namespace: redis-ns
+      default:
+        connectionTimeout: 5s # why does the Backend Service Owner try to change the way clients use the 'redis' service?
+```
+
+Such policy should not be allowed, if Backend Service Owner needs to change the way clients use `redis`
+they have to put the policy into `kuma-system` namespace by checking with Mesh Operator first.
+
+The validation rules for namespace-scoped policy are the following:
+* to-policy when placed into the custom namespace must have the namespace and zone either in the top-level targetRef or in each to-item
```
lahabana, line 110: Annoying!

docs/madr/decisions/049-policy-on-namespace.md
lahabana: Also with the current zone originated policies is similar to having an implicit `MeshSubset` no?

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+
+##### Pros and Cons of automatically detecting if a policy needs to be exported
+
+Having `spec.export.mode` might be excessive taking into account we already have a top-level targetRef.
+If the top-level targetRef specifies a tag `kuma.io/zone` then it's clear whether we need to export the policy or not.
+
+Zone-originated policies in `kuma-system` won't be synced to other zones. 
+This eliminated the problem of backwards compatibility. 
+Also, with the introduction of namespace-originated policies, applying policies to `kuma-system` becomes an edge use case.
+The only reason to apply a policy on zone's `kuma-system` is Zone Cluster Operator stepping in to fix the undesired behaviour.
```
lahabana, line 259: Not only `undesired`? Let's say 1 zone is very far away maybe the operator will add increased default timeouts.

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+   When kinds are equal, go to the next step.
+2. Zone originated policy is more specific than global originated policy.
+   When policies originates at the same place, go to the next step.
+3. **[new step]** Policy in any custom namespace is more specific than any policy in the `kuma-system` namespace.
+   When both policies either from custom namespaces or from `kuma-system`, go to the next step.
+4. **[new step]** Policy from local zone is more specific than a policy from another zone. 
+   When policies have the same zone, go to the next step.
+5. **[new step]** Policy from local namespace is more specific than a policy from another namespace.
+   When namespaces are equal, go to the next step.
+6. The policy is more specific if it's name is lexicographically less than other policy's name ("aaa" < "bbb" so that "
```
lahabana, line 284: Can we just write pseudo code? This is hard to process

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+As mentioned previously, applying a policy on the custom namespace affects requests to the pods in the custom namespace.
+This means, no matter where MeshHTTPRoute is created it can be a "producer" route and affect consumers' configurations.
+It's always justifiable to reference any MeshHTTPRoute from any MeshTimeout.
+
+### MeshGateway reference
+
+`MeshGateway` at this moment is not namespace-scoped. It can change in the future with
+the [MeshBuiltinGateway resource](https://github.com/kumahq/kuma/issues/10014).
+
+For policies in custom namespaces, we've decided that we would allow targeting `MeshGateway` in `MeshHTTPRoute` and `MeshTCPRoute`, and policies
```
lahabana, line 320: targeting where? to? top level?

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+      rules:
+        - matches:
+            - path:
+                type: PathPrefix
+                value: /gui
+          default:
+            backendRefs:
+              - kind: MeshService
+                name: frontend
+                namespace: frontend-ns
```
lahabana, line 459: This namespace can be implicit right?
Also, if this namespace were different from `frontend-ns` would this be rejected?

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+name: circuit-breaker
+namespace: frontend-ns
+spec:
+  targetRef:
+    kind: MeshGateway
+    name: edge-gateway
+  to:
+  - targetRef:
+      kind: MeshService # you need to specify MeshService in to[].targetRef.kind
+      name: frontend
```
lahabana, line 479: This is equivalent to having the implicit `namespace: frontend-ns` ? 
What would happen if I had `namespace: backend-ns`

Finally what happens with:

```yaml
type: MeshCircuitBreaker
mesh: default
name: circuit-breaker
namespace: frontend-ns
spec:
  targetRef:
    kind: MeshGateway
    name: edge-gateway
  to:
  - targetRef:
      kind: MeshService # you need to specify MeshService in to[].targetRef.kind
      name: frontend
     default: {}
  - targetRef:
      kind: MeshService
      name: backend
      namespace: backend-ns
    default: {}
```

My gut feeling is that this should rejected (entries in the `to` must have the same namespace). We should likely suggest users to not set multiple `to` entries

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: The next section `#### Enhanced validation` provides more details and also has some examples.

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: It probably wasn't clear that these points:

```yaml
* Enhanced validation...
* Ability to opt-in or opt-out...
```

are followed by the separate sections to elaborate on each point:

```md
#### Enhanced validation
...
#### Considered options on opting-in(out) to cross-zone policy syncing
...
```

and the decision for opting-in(out) is exactly what you said – we can always use a top `spec.targetRef.kind: MeshSubset` to reduce blast radius.

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,529 @@
...
+    tags:
+      k8s.kuma.io/namespace: backend-ns
+      kuma.io/zone: producer-zone
+  to:
+    - targetRef:
+        kind: MeshService
+        name: redis
+        namespace: redis-ns
+    - targetRef:
+        kind: Mesh
```
Icarus9913, line 165: I just get confused with the example, it looks like a cross namespace control? 
So, can we define our top TargetRef `MeshSubset` with another NS `front-end` if our policy is `backend-ns`?

Actually, I feel it's pretty complex in the cross-namespace use cases :(

Icarus9913: OK, I guess it's normal that we have cross namespace application communications use cases. 
What about in these cases, the policy must use `kuma-system` namespace?

Icarus9913: OK, I just get understand with this.
Maybe we should also paste the link here: https://gateway-api.sigs.k8s.io/mesh/#gateway-api-for-service-mesh

For those people who are unfamiliar with the context

docs/madr/decisions/049-policy-on-namespace.md
jakubdyszkiewicz: Alternative?

docs/madr/decisions/049-policy-on-namespace.md
Automaat: updated to be more precise about uing MeshGateway in topLevel targetRef

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,613 @@
...
+  namespace: kuma-system
+  labels:
+    kuma.io/zone: zone-1
+    k8s.kuma.io/namespace: backend-ns
+    kuma.io/display-name: producer-timeout
+spec:
+  to:
+    - targetRef:
+        kind: MeshService
+        name: backend # based on labels we should be able to de-reference this to 'backend-w3r3jwi90l8dyuix'
```
jakubdyszkiewicz, line 390: can't we replace target refs before sync from global to zone?

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: We tried, it doesn't look simpler, we merged 3 steps into 1 so should be easier to process

jakubdyszkiewicz (approved):

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: Right, replacing name with hashed name seems like a better option

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+
+* Good, because it easy to implement
+* Good, because it works similar to zone-originated policies (we implicitly add `kuma.io/zone` tag for top-level targetRef)
+* Bad, because it doesn't remove necessity to create policies in `kuma-system` namespace
+
+## Pros and Cons of adopting producer/consumer concept from GAMMA
+
+* Good, because it's familiar to users who know GAMMA
+* Good, because creating policies in `kuma-system` becomes an edge use-case, the vast majority of use cases is covered by namespace-scoped policies
+* Bad, because it requires syncing policies across zones.
```
lahabana, line 37: Is this actually Bad. I don't think so. I think if anything is bad here is that the logic of which policies can be synced or not can be a little complex

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: I added the link to GAMMA, thanks

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+  from:
+    - targetRef:
+        kind: Mesh
+      default: {}
+```
+
+### Namespace-scoped policies and multizone
+
+Namespace-scoped policies can be applied to a custom namespace only on zones as there are simply no DPPs and custom
+namespaces on Global.
```
lobkovilya, line 282:
```suggestion
namespaces on Global. This will be enforced by validation webhook.
```

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+        kind: MeshService
+        name: backend
+        namespace: finance-ns # this indicates the policy is a producer policy
+    - targetRef:
+        kind: MeshService
+        name: redis
+        namespace: redis-ns # this indicates the policy is a consumer policy
+```
+
+Policies that support both `from` and `to` and specify them at the same time shouldn't be allowed as well.
```
lahabana, line 157: This is a breaking change no? Is it namespaced policies that use from and to or policies using namespaced entities (routes, services, meshexternalServices)?

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+
+### Syntactic sugar
+
+There are few things in policies syntax Kuma can figure out without requiring user to specify them explicitly.
+The syntactic sugar is very important as it allows user to not specify zone's name and namespace explicitly in the policy's yaml. 
+
+We don't want to change `spec` on the fly, because it might be surprising for user to see completely different resource in the store.
+Also, in case of the future changes in the logic, it's preferable to store the less explicit version. 
+But we probably have to implement a new feature in the Inspect API that allows viewing the most explicit version of the resource.
```
lahabana, line 166: Why?

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+
+#### Empty namespace in `to[].targetRef`
+
+If the `spec.to[idx].targetRef.namespace` is not specified then it's equal to `metadata.namespace`.
+
+#### Empty top-level targetRef
+
+* if the top-level targetRef is not specified in **producer policy**
+  then it's assumed to be `Mesh`
+* if the top-level targetRef is not specified in **consumer policy** or **workload-owner policy**
```
lahabana, line 176:
```suggestion
* if the top-level targetRef is not specified  or is `targetRef.kind: Mesh` in **consumer policy** or **workload-owner policy**
```

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+  then it's assumed to be `MeshSubset` with `kuma.io/zone` and `k8s.kuma.io/namespace` tags.
```
lahabana, line 177: Also if you have partial MeshSubset .e.g:

```
targetRef:
  kind: MeshSubset
  tags:
    foo: bar
```

It gets rewritten as:


```
targetRef:
  kind: MeshSubset
  tags:
    foo: bar
    kuma.io/zone: foo
    k8s.kuma.io/namesapce: bar
```

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+      kind: MeshService
+      name: backend
+      namespace: backend-ns
+```
+
+but we have to implement "smart" de-referencing when the policy is synced to another zone.
+
+#### Considered Options
+
+* explicitly de-reference all `to[].targetRef.name` before syncing the resource to another zone
```
lahabana, line 366: So this will be in KDS in the originating zone?

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+        kind: MeshService
+        name: backend
+        namespace: backend-ns
+      default: $conf2 # 'conf2' refines 'conf1' that was already provided by the producer
+```
+
+We can formally define what's producer and consumer Kuma policies:
+
+* if for each `idx` the value of `spec.to[idx].targetRef.namespace` is equal to `metadata.namespace` then it's a producer policy
+* for each `idx` the value of `spec.to[idx].targetRef.namespace` is **not** equal to `metadata.namespace` then it's a consumer policy
```
michaelbeaumont, line 97: Is this 
```suggestion
* if for some `idx` the value of `spec.to[idx].targetRef.namespace` is **not** equal to `metadata.namespace` then it's a consumer policy
```

i.e. the negation of the previous line

docs/madr/decisions/049-policy-on-namespace.md
lahabana: If we do this can't the system namespace be different depending on the zone?

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+      kuma.io/zone: zone-with-backend
+  from:
+    - targetRef:
+        kind: Mesh
+      default: {}
+```
+
+### Namespace-scoped policies and multizone
+
+Namespace-scoped policies can be applied to a custom namespace only on zones as there are simply no DPPs and custom
```
michaelbeaumont, line 281: Because we're syncing policies to other zones now, maybe there should be something in here about where/if egress/ingress play a role.

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+        - matches:
+            - path:
+                type: PathPrefix
+                value: /gui
+          default:
+            backendRefs:
+              - kind: MeshService
+                name: frontend
+                # namespace: frontend-ns - is added automatically
+```
```
lahabana, line 574: Isn't there an access issue here? Anyone, anywhere could break the gateway no?

Feels like we should not allow referencing MeshGateway from namespaced policies ftm.

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+      kind: MeshService
+      name: backend
+      namespace: backend-ns
+```
+
+but we have to implement "smart" de-referencing when the policy is synced to another zone.
+
+#### Considered Options
+
+* explicitly de-reference all `to[].targetRef.name` before syncing the resource to another zone
```
michaelbeaumont, line 366: It's not clear to me from this what explicitly/implicitly dereference means

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+
+We are creating policy in `frontend-ns` namespace, and we are modifying traffic to backend service in different namespace. 
+Right now MeshGateway is cluster scoped, so it is not clear what should we do. I think we have three options here:
+
+1. forbid user from configuring this, `MeshGateway` is cluster scoped so this does not make sense
+2. allow and try to apply config only to `MeshGatewayInstances` in `frontend-ns`, feels wrong since policy is not really applied on MeshGateway but on MeshGatewayInstance, 
+but this can be done now with matching gateway + listener + dataplane tags, and specifying namespace tag 
+3. allow but ignore policy, as there is no namespace scoped MeshGateway present
+
+**We've decided to go with option no 1. We will forbid user from configuring this `MeshGateway` is cluster scoped so this does not make sense**
```
lahabana, line 637: +1

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+                name: frontend
+                # namespace: frontend-ns - is added automatically
+```
+
+##### Targeting MeshGateway in other policies
+
+###### Producer policies
+
+As a **Frontend Service Owner** I want to configure circuit breaker on `MeshGateway` without the access to
+cluster-scoped CRDs so that I could manage my routes independently.
```
lahabana, line 581: This equivalent to a specific MeshSubset right?

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: It shouldn't be possible to mix producer and consumer to-items. That's why the second is not a negation of the first line

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: rephrased both considered options

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: Ingress is never modified by the policies, so I guess it doesn't play role in syncing. Egress can be affected by ExternalServices, so we probably have to add something about the ExternalServices here

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: Depending on the producer/consumer the empty top-level targetRef can be translated either to `Mesh` or `MeshSubset`. It would be nice if user had an ability to see how policy look without the syntactic sugar.

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: What do you mean by breaking the gateway? I see we don't validate `backendRefs` at this moment and potentially you can redirect traffic to any namespace in the cluster. Is it what you meant by the access issue?

docs/madr/decisions/049-policy-on-namespace.md
Automaat: can you elaborate? I don't think I understand the comment?

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: MeshExternalService can be created only in `kuma-system` namespace. This means, all policies that reference MeshExternalService are consumer policies and shouldn't be synced. I'm going to put this into the doc

lobkovilya: I added:

> Policies that reference `MeshExternalService` in `to[].targetRef` won't be synced to other zones. MeshExternalService can be created only in `kuma-system` namespace and that's why the policy it's referenced by is always a consumer policy. That's why the policies that have potential to indirectly change ZoneEgress configuration are never synced to other zones.

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: Added an implem section:

> #### Implementation
> 
> The update of `to[].targetRef` is going to happen in 2 steps:
> 
> 1. Global KDS client updates `to[].targetRef.name` and `to[].targetRef.namespace` when receives policies from the source zone
> 2. KDS client of the destination zone updates `to[].targetRef.namespace` when receives policies from Global
> 
> All clients should update `to[].targetRef.namespace` because this is an environment specific information 
> (i.e. system namespaces can be different, Universal deployments don't have namespaces at all).

docs/madr/decisions/049-policy-on-namespace.md
lahabana: A MeshGateway is very similar to a MeshSubset no? In the end it just identifies a subset of all dataplanes

docs/madr/decisions/049-policy-on-namespace.md
lobkovilya: Yeah, it's right next to the sentence I sent:

> Top-level targetRef in Kuma policies always references a group of sidecars that are subjected to the change. New namespace-scoped policies are no exception, but we're going to implicitly add k8s.kuma.io/namespace and kuma.io/zone for consumer and workload-owner policies.
> 
> ```yaml
> # a valid consumer policy
> metadata:
>   namespace: frontend-ns
>   labels:
>     kuma.io/zone: zone-1
> spec:
>   targetRef:
>     kind: Mesh # despite it targets 'Mesh' the policy will be applied to pods in the 'frontend-ns' of 'zone-1'
>   to:
>     - targetRef:
>         kind: MeshService
>         name: backend
>         namespace: backend-ns
> ---
> # workload-owner policy with 'from'
> metadata:
>   namespace: frontend-ns
>   labels:
>     kuma.io/zone: zone-1
> spec:
>   targetRef:
>     kind: Mesh # despite it targets 'Mesh' the policy will be applied to pods in the 'frontend-ns' of 'zone-1'
>   from:
>     - targetRef:
>         kind: Mesh
> ```

michaelbeaumont (approved):

docs/madr/decisions/049-policy-on-namespace.md
michaelbeaumont: I'm not sure we should have namespaced HTTPRoutes attach to MeshGateway either. Why should frontend be able to decide whether some arbitrary traffic makes it to their service?

For outbound policies though I think there's no issue, since it's already been decided that the traffic is headed to their service.

docs/madr/decisions/049-policy-on-namespace.md
```diff
@@ -0,0 +1,681 @@
...
+        labels:
+          kuma.io/display-name: backend-route
+          k8s.kuma.io/namespace: backend-ns
+          kuma.io/zone: zone-1
+      default:
+        http:
+          requestTimeout: 3s
+```
+
+### Referencing MeshGateway
```
michaelbeaumont, line 524: I admit I overlooked this section, feels like it belongs higher up, before example.

Automaat: Each section has it's own examples to make it easier to understand, thats why MeshGateway is later on with it's own examples

docs/madr/decisions/049-policy-on-namespace.md
Automaat: right, that makes sense, I've changed the MADR accordingly

docs/madr/decisions/049-policy-on-namespace.md
Automaat: That's right, I've added this info to the MADR

Automaat (approved):

lahabana (approved):

PR #10150, opened 2024-05-02
test(e2e): never fail when running debug commands

lahabana:
We want to export any state available and not fail in the middle of an export

### Checklist prior to review

### DIFF
--- a/test/framework/debug.go
+++ b/test/framework/debug.go
@@ -17,26 +17,36 @@ import (
 // * CP logs (although we print this already on failure)
 func DebugUniversal(cluster Cluster, mesh string) {
 	ensureDebugDir()
+	errorSeen := false
 	Logf("printing debug information of cluster %q for mesh %q", cluster.Name(), mesh)
 	// we don't have command to print policies for given mesh, so it's better to print all than none.
 	kumactlOpts := *cluster.GetKumactlOptions()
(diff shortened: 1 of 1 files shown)


test/framework/debug.go
```diff
@@ -17,26 +17,36 @@ import (
...
-	Expect(err).ToNot(HaveOccurred())
+	if err != nil {
+		// We don't want to fail in the middle.
+		out = fmt.Sprintf("kumactl export failed with error: %s", err)
+		errorSeen = true
+	}
 
 	exportFilePath := filepath.Join(Config.DebugDir, fmt.Sprintf("%s-export-%s", cluster.Name(), uuid.New().String()))
 	Expect(os.WriteFile(exportFilePath, []byte(out), 0o600)).To(Succeed())
+	Expect(errorSeen).To(BeTrue(), "some debug commands failed")
```
michaelbeaumont, line 34: Why expect it to be true? Shouldn't we just keep the error and expect it with `Not(HaveOccurred())`

lahabana, line 34:
```suggestion
	Expect(errorSeen).ToNot(BeTrue(), "some debug commands failed")
```

lahabana: You are right fixed!

michaelbeaumont (approved):

PR #10152, opened 2024-05-02
docs(MADR): add policy matching for `MeshService`

michaelbeaumont:
### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- #9707


docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Local `MeshServices`
+
+If the policy ref doesn't set `zone`:
+
+* without `namespace` set on a k8s zone:
+  * if the policy is *not in `kuma-system`*, the ref refers to
+    `MeshServices` in the same namespace
+  * if the policy is in `kuma-system`, the ref refers to
+      the core name of the `MeshService`
```
jakubdyszkiewicz, line 93: core name? do you mean kuma.io/service?

Can I then also use `kuma.io/service` when policy is namespaced?

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,174 @@
...
+```yaml
+kind: MeshService
+name: backend
+labels: {}
+zone: east
+namespace: backend
+port: 80
+```
+
+In general, we try to do the expected thing and refer to `MeshService` via its display name.
```
jakubdyszkiewicz, line 75: I'm tempted not to do this. While it may seem like better UX, I feel like this will introduce ambiguity, because you can use it to match a real name as well as display name.
I feel that labels even though more verbose it's more clear what is going on.

Especially that this way it's easier to then go and introduce those implicit things.

Alternatively I'd consider `displayName` field.

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,174 @@
...
+### Positive Consequences
+
+* Allows users to apply policy to traffic going to `MeshService` and
+  `MeshMultiZoneService`
+* We can match set of `MeshServices`, important for headless services
+* We're flexible with regards to local vs non-local `MeshServices`.
+
+### Negative Consequences
+
+* Relative complexity of the semantics of various combinations
```
jakubdyszkiewicz, line 174: how do we select multiple services? Can MeshService just select multiple services or do we introduce MeshServiceGroup?

michaelbeaumont: It should be covered by using `labels` instead of `name`

docs/madr/decisions/046-meshservice-policy-matching.md
michaelbeaumont: Discussion:

```yaml
kind: Service
metadata:
  namespace: app-ns
  name: backend
---
kind: MeshService
metadata:
  namespace: app-ns
  name: backend
---
targetRef:
  kind: MeshService
  name: backend
  namespace: app-ns
```

```yaml
---
kind: MeshService
metadata:
  namespace: kube-system
  name: backend-a57694fd # hash namespace, zone and mesh
---
targetRef:
  kind: MeshService
  name: backend-a57694d
  namespace: app-ns
---
targetRef:
  kind: MeshService
  displayName: backend
  zone: other-zone
  namespace: app-ns
```

docs/madr/decisions/046-meshservice-policy-matching.md
michaelbeaumont: Assuming above changes, we can look for the Kubernetes MeshService object with this name in `kuma-system`, so the same logic "look in policy namespace".

`zone` only makes sense with `displayName`

```yaml
targetRef:
  kind: MeshService
  zone: other-zone
---
targetRef:
  kind: MeshService
  zone: other-zone
  namespace: app-ns
```

can't have name with labels/zone

docs/madr/decisions/046-meshservice-policy-matching.md
lahabana: It's also not super clear where you can use a set of services vs a single service

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,191 @@
...
+If the policy ref doesn't set `zone`, it refers to a `MeshService` in the local
+zone.
+
+On a k8s zone:
+* without `namespace`, the policy refers to `MeshServices` in the same
+  namespace as the policy.
+* with `namespace`, only that namespace is searched matching `MeshServices`
+
+Note that this means it's possible to refer to synced `MeshServices` via their
+transformed `<service>-<hash-suffix>`.
```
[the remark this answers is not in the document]
michaelbeaumont: We don't have to suggest it, it's just a natural consequence of being able to refer to a Kubernetes object.

docs/madr/decisions/046-meshservice-policy-matching.md
jakubdyszkiewicz: Do we need to differentiate it?

michaelbeaumont: To be clear, I don't mind if we want to introduce another kind for groups, as long as the naming is consistent.

jakubdyszkiewicz (approved):

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,191 @@
...
+* Targeting `MeshMultiZoneService`
+* A balance between correctness and ease of use when configuring policy.
+
+## Considered Options
+
+This MADR adds 2 kinds `to[].targetRef.kind` and Mesh*Route `backendRefs[].kind`:
+
+* `MeshService`
+* `MeshMultiZoneService`
+
```
Icarus9913, line 39: Hi Mike, could you please explain more with this sentence?

In the past, I learned from the [Target-Resources](https://kuma.io/docs/2.7.x/policies/targetref/#target-resources)
The Sidecar targetRef `to[].targetRef.kind` already allows kind `MeshService`, we repeat it again here?
And what's meaning of  "and Mesh*Route `backendRefs[].kind`"?  

Sorry that I did't get it.

slonka (approved):

docs/madr/decisions/046-meshservice-policy-matching.md
michaelbeaumont: Yeah it's badly phrased, it adds new functionality for matching real `MeshServices` to the existing `kind: MeshService`.

> And what's meaning of "and Mesh*Route backendRefs[].kind"?

That refers to references in MeshHTTPRoute/MeshTCPRoute

Icarus9913 (approved):

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,227 @@
...
+object in the zone of the policy.
+
+Exactly one of either `targetRef.name` or `targetRef.labels` is **required**.
+
+#### `sectionName`
+
+If `sectionName` is set in `targetRef.kind` with `kind: MeshService`,
+it refers to an entry in `ports` by name and only traffic to that port is affected.
+
+In `backendRefs`, `sectionName` is required.
```
Automaat, line 103: you say here, that `sectionName` field is required in `backendRef`, but then in example you use `port` field. Which one is correct?

michaelbeaumont: ah yes, forgot to update this. it should be `port` in `backendRef`

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,228 @@
...
+  - port: 80
+    name: http
+---
+targetRef:
+  kind: MeshService
+  namespace: backend
+  name: backend # only name or labels
+  labels: # only name or labels
+    kuma.io/zone: east
+  sectionName: http
```
Automaat, line 90: this field `sectionName` feels weird. I understand that this is taken from Gateway API, but It does not say anything to me. Like, which section are we talking about? Can it be something else then port in the future?  Why can't we just use port like in `backendRef`. Do we already use `sectionName` in other places? 

In my opinion, it adds complexity for regular users, as they need to translate this to port in their head every time they use it

michaelbeaumont: No, we don't use `sectionName` yet.

> In my opinion, it adds complexity for regular users, as they need to translate this to port in their head every time they use it.

I don't disagree with this exact point, but if we ever want to be Gateway API first, it adds more complexity just to do things differently.

The main advantage is that we can reuse the same field if we have similar "subsections" in other resources we want to target.

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,228 @@
...
+     rules:
+       - matches:
+           - path:
+               type: Prefix
+               value: /v1
+         default:
+           backendRefs:
+             - kind: MeshService
+               name: frontend
+               port: http
```
jakubdyszkiewicz, line 200: Is it always by name? should this be `portName`? 🤔

michaelbeaumont: Updated to only allow ints like Gateway API, WDYT?

docs/madr/decisions/046-meshservice-policy-matching.md
Automaat: so maybe we should use `sectionName` in `backendRef` instead of `port` to make it consistent? Is it the same story here, that GatewayAPI uses `port` field?

michaelbeaumont: yeah it does :frowning:

jakubdyszkiewicz: it's unfortunate that the api is not
```
section:
  type: Port
  name: http
```

I'm not happy, but I'm ok with `sectionName`, it is confusing without reading docs, but I see value in being consistent with GAPI.

jakubdyszkiewicz (approved):

docs/madr/decisions/046-meshservice-policy-matching.md
michaelbeaumont: @lahabana would like to hear your opinion before we merge

I agree the decision to have just `port: int` in `backendRef` isn't ideal from an API coherency perspective.
At least this way we keep compatibility with Gateway API but leave our options open to maybe add `sectionName` to `backendRef` at a later point.

Automaat: I see the value in being consistent with Gateway API, so I am ok with this api.

Automaat (approved):

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,228 @@
...
+* `kuma.io/display-name`
+* `kuma.io/zone`
+
+If `kuma.io/zone` is set, a missing `namespace` field means we
+refer to _all_ `MeshServices` with this name.
+If no `kuma.io/zone` is set, the `namespace` is assumed to be the namespace of
+the policy making the reference.
+
+This means we don't infer any similarity between the local namespace
+and the other zone's namespace.
```
michaelbeaumont, line 135: This is not as nice as I thought. So we don't want to assume namespace similarity, that's good. But how do you refer to a MeshService from another zone then?

```
namespace: other-zone-ns
labels:
  kuma.io/zone: other-zone
```

isn't ideal because the real object is not found at this namespace, rather at `kuma-system`.

so do we have

```
namespace: kuma-system
labels:
  kuma.io/zone: other-zone
  k8s.kuma.io/zone: other-zone-ns
```

? This is the cleanest but it's very verbose.

Automaat: Shouldn't it be `k8s.kuma.io/namespace: other-zone-ns` in the second example?

Automaat: So this means that you need to create policy in `kuma-system` namespace in order to target `MeshService` in different zone? While discussing policies on namespace MADR we talked that in best case scenario SO won't need to apply anything on `kuma-system` namespace

lobkovilya: Well, TBH it's not drastically more verbose than the first examples. I'd still with the more verbose version in favor of preserving the rule that we reference real objects

docs/madr/decisions/046-meshservice-policy-matching.md
lobkovilya: But port's name in MeshService is not required right? What happens if there is no name?

docs/madr/decisions/046-meshservice-policy-matching.md
slonka: I agree with Ilya here ☝️

docs/madr/decisions/046-meshservice-policy-matching.md
slonka: I'd also like us to be consistent with GAPI

docs/madr/decisions/046-meshservice-policy-matching.md
michaelbeaumont: OK spoke with @jakubdyszkiewicz and we decided to just disallow `namespace` when using `labels`, and all namespaces will be matched. This makes it more verbose for the local case of a group of MeshServices but :shrug:

michaelbeaumont: > So this means that you need to create policy in kuma-system namespace in order to target MeshService in different zone?

I don't think so, it's just about `namespace` in the target ref itself. Updated the YAML to make it clearer.

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,287 @@
...
+  groups of `MeshService`.
+* Add additional kind `NonLocalMeshService` that explicitly selects non-local
+  `MeshServices` via a structured set of fields `zone`, `port`, `labels`,
+  `namespace`
+* Add an additional kind `MeshServiceId` to support selecting an arbitrary
+  `MeshService` via a kind of UID like `kuma://backend/zone-1/backend-ns/8080`
+
+### Pros and Cons
+
+#### `labels` vs top level fields
```
lahabana, line 68: Need an example here

docs/madr/decisions/046-meshservice-policy-matching.md
lahabana: Can I have examples for both backendRef and targetRef in each case? I tend to lean towards GAPI format but I'm not sure I understand everything

lahabana (commented):
I think this type of matching makes sense to me. Only thing I'd like discussed is using `kind: MeshService` as a top level targetRef similar to `kind: MeshHTTPRoute`. In such case the `to[].targetRef.kind` would always need to be `Mesh`. I'd love to see pros and cons of this.
The fact we're moving to MeshService might an opportunity to iron out issues on the policy matching story that exists nowadays.

docs/madr/decisions/046-meshservice-policy-matching.md
lahabana: I'm concerned with the same targetRef and merging. How do you order policies?

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,287 @@
...
+namespace: ...
+```
+
+Now we have to consider how we attach to a port and select from a zone.
+
+* Add top level field `zone` and `port` and `labels` for selecting groups of
+  `MeshService`
+* Add top level `port` field and use `labels: {"kuma.io/zone": "..."}`
+* Use generic Gateway API targetRef field `sectionName` to specify a port in
+  `targetRef` but `port` in `backendRef` and add `labels` for selecting
```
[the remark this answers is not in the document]
michaelbeaumont: no, `backendRef` has only `port`.

docs/madr/decisions/046-meshservice-policy-matching.md
michaelbeaumont: If you order MeshServiceGroup before MeshService, you order MeshService + labels before MeshService + name.
There's no information lost between the two representations.

docs/madr/decisions/046-meshservice-policy-matching.md
michaelbeaumont: done

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,313 @@
...
+```
+
+```yaml
+spec:
+ to:
+ - targetRef:
+     kind: MeshService
+     labels:
+        kuma.io/display-name: backend
+        kuma.io/zone: other-zone
```
lahabana, line 263: With the producer and consumer model we're building up in: https://github.com/kumahq/kuma/pull/10148
It's not super clear how things work as now zone policies could be synced to other zones.

1. For example I have a backend service:

```yaml
kind: MeshService
metadata:
  name: my-backend
  namespace: my-backend-ns
spec:
  selector:
      dataplaneTags:
        app: my-backend
  ports:
    - port: 80
       targetPort: 8080
       name: http
       appProtocol: http
```

2. Intuitively I'd write this timeout policy that would affect all consumers of my service (in *ALL* zones):

```yaml
kind: MeshTimeout
metadata:
  name: my-backend
  namespace: my-backend-ns
spec:
  targetRef:
     kind: Mesh
  to:
    - targetRef:
          kind: MeshService
          name: my-backend
          port: 80 # (subquestion 1: here the alternative sectionName: http is valid too right?)
      default:
         timeout: 5s
```

3. Then I'd want to write a consumer policy in the same zone as:

```yaml
kind: MeshTimeout
metadata:
   namespace: my-frontend-ns
   name: override-my-backend
spec:
  targetRef:
     kind: Mesh
  to:
    - targetRef:
          kind: MeshService
          name: my-backend
          namespace: my-backend-ns
          port: 80 # (subquestion 1: here the alternative sectionName: http is valid too right?)
      default:
         timeout: 5s
```

4. And I'd write a consumer policy in a different zone as:

```yaml
kind: MeshTimeout
metadata:
   namespace: my-frontend-ns-in-another-zone
   name: override-my-backend
spec:
  targetRef:
     kind: Mesh
  to:
    - targetRef:
          kind: MeshService
          sectionName: http
          labels:
            kuma.io/display-name: my-backend
            k8s.kuma.io/namespace: my-backend-ns
            kuma.io/zone: my-other-k8s-cluster
      default:
         timeout: 10s
```

1 and 2 gets synced to remote zones. 3 and 4 don't. @lobkovilya is this correct?

Multiple questions come up to me:

- Do we find it acceptable that matching MeshService in the same zone vs cross zone is syntactically very different (Using labels for remote vs using name/ns for local)?
- With producer policies these gets synced to remote zones now. Is it still straight-forward to do the matching (do we use the synced labels to enhance the matcher)?

lahabana (commented):
I'm starting to like this. I have a bit of a concern now that we're talking about product vs consumer policies but that's something we can talk about further later

docs/madr/decisions/046-meshservice-policy-matching.md
lobkovilya: I'm talking about [these rules](https://github.com/kumahq/kuma/pull/10148/files#diff-cb6c27d1ae4efdd51ae3730849991f19c4016f47fb91292fa349d9c0e0ea8a95R109-R111):
> The validation rules for namespace-scoped policy are the following:
> * to-policy when placed into the custom namespace must have the namespace and zone either in the top-level targetRef or in each to-item
> * from-policy when placed into the custom namespace must always have the namespace and zone in the top-level targetRef

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,313 @@
...
+
+If the policy ref sets `name`, it refers to one `MeshService` in the *local
+zone*.
+
+On a k8s zone:
+* without `namespace`, the policy refers to `MeshServices` in the same
+  namespace as the policy.
+* with `namespace`, only that namespace is searched for matching `MeshServices`
+
+On a universal zone, `namespace` is not valid.
```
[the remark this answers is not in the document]
michaelbeaumont: yes. our goal was to reserve name/namespace for _real_ objects on this cluster.

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,313 @@
...
+           - path:
+               type: Prefix
+               value: /v1
+         default:
+           backendRefs:
+             - kind: MeshService
+               port: 8080
+               labels:
+                 kuma.io/display-name: frontend
+                 kuma.io/zone: east
```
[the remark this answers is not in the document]
michaelbeaumont: And in particular, what happens to `weight`? Not sure...

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,313 @@
...
+```
+
+#### Status
+
+Similar to Gateway API conditions, we report back as status conditions
+if a route targets a
+`MeshService`/port tuple that doesn't exist.
```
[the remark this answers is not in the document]
michaelbeaumont: On the policy for sure.

docs/madr/decisions/046-meshservice-policy-matching.md
lobkovilya: My messages above are not relevant, as we changed the policy on namespace MADR since then. 

> Do we find it acceptable that matching MeshService in the same zone vs cross zone is syntactically very different (Using labels for remote vs using name/ns for local)?

I think it's acceptable. You can still you name/ns for remote, but the name should be the one with hash suffix.  

> With producer policies these gets synced to remote zones now. Is it still straight-forward to do the matching (do we use the synced labels to enhance the matcher)?

In policy on namespace MADR we described `to[].targetRef.name` should be automatically updated to the name with hash suffix when synced to another zone.

docs/madr/decisions/046-meshservice-policy-matching.md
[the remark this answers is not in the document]
michaelbeaumont: This really should only be able to match one specific service/port combo. I think we need to validate this to be the case.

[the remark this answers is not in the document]
lobkovilya: but we can validate this only when the policy is applied (not when it's created) right?

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,322 @@
...
+
+#### `sectionName`
+
+If `sectionName` is set in `targetRef.kind` with `kind: MeshService`,
+it refers to an entry in `ports` by name and only traffic to that port is affected.
+
+In `backendRefs`, `port` is required. There is no `sectionName`, in accordance
+with Gateway API HTTPRoute.
+
+#### `name`
```
lobkovilya, line 197: The backwards compatibility is not covered, but I think we discussed that 
```yaml
to:
  - targetRef:
      kind: MeshService
      name: backend
```
should try to find the real MeshService resource named `backend` first and if there is none then work as an old-fashioned MeshService?

docs/madr/decisions/046-meshservice-policy-matching.md
```diff
@@ -0,0 +1,322 @@
...
+* use exactly one of `labels` or `name`/`namespace`
+* no new top level fields besides `labels`
+
+### `targetRef`
+
+The policy is applied
+on both the Envoy outbound listeners and clusters
+that correspond to the `MeshService`.
+Note that more than one `MeshService` can be matched in which case it applies to
+all matched `MeshServices`.
```
lahabana, line 163:
```suggestion
each matched `MeshServices`.
```

docs/madr/decisions/046-meshservice-policy-matching.md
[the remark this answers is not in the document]
michaelbeaumont: First of all, we can require `display-name` and `zone` labels.

The only problem is if we have a target Kubernetes zone and don't include `namespace`. In that case, we can treat the ref as invalid no matter how many matching services really exist. Ideally this would be reported as an error condition

Automaat (approved):

docs/madr/decisions/046-meshservice-policy-matching.md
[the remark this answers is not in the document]
Automaat: I've updated MADR with: `backendRef cannot match more than one MeshService. When backendRef matches multiple MeshServices this should result in error, and route should be applied as if no services were matched.`
this should cover the issue

lobkovilya (approved):

PR #10191, opened 2024-05-07
fix(MeshHttpRoute): don't split header value prematurely

### LINKED ISSUE #10156: When modifying the headers, split the value may cause unexpected bug
### What happened?

In https://github.com/kumahq/kuma/blob/5c1f62de50047e0c082ec6d8f92337a12c27ce47/pkg/plugins/policies/meshhttproute/xds/filters/requestheadermodifier.go#L36

the header value is split by `,`, and then added one by one.

In case like `Set-Cookie: id=a3fWa; Expires=Wed, 21 Oct 2015 07:28:00 GMT`, it will be added as:
`Set-Cookie: id=a3fWa; Expires=Wed` and
`Set-Cookie: 21 Oct 2015 07:28:00 GMT`

I wonder why not set the header directly without any splitting?

spacewander:
Fix #10156
Signed-off-by: spacewander <spacewanderlzx@gmail.com>

### Checklist prior to review

- [x] [#10156][1] as well as docs and UI issues --


spacewander:
Will fix the unit test later.

slonka:
from the issue:

> I think Envoy will do the right thing here, we should change this.

@michaelbeaumont did you by any chance had time to check if it does the right thing?

slonka (commented):
from the issue:

> I think Envoy will do the right thing here, we should change this.

@michaelbeaumont did you by any chance had time to check if it does the right thing?

michaelbeaumont (approved):
LGTM, thanks @spacewander !

PR #10192, opened 2024-05-07
fix(gatewayapi): reconcile HTTPRoutes when relevant Services change

michaelbeaumont:
This is important if an HTTPRoute references a Service that didn't initially exist but then is created, we need to reconvert to MeshHTTPRoute and update status.
The PR adds an index on HTTPRoutes to the Services they care about to make it cheaper to queue reconciliation for HTTPRoutes when a Service changes.

### Checklist prior to review

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/gatewayapi/http_route_controller.go
+++ b/pkg/plugins/runtime/k8s/controllers/gatewayapi/http_route_controller.go
@@ -354,7 +354,7 @@ func routesForService(l logr.Logger, client kube_client.Client) kube_handler.Map
 		if err := client.List(ctx, &routes, kube_client.MatchingFields{
 			servicesOfRouteField: kube_client.ObjectKeyFromObject(svc).String(),
 		}); err != nil {
-			l.Error(nil, "unexpected error listing HTTPRoutes")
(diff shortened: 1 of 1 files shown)


pkg/plugins/runtime/k8s/controllers/gatewayapi/http_route_controller.go
```diff
@@ -338,7 +338,76 @@ func routesForGrant(l logr.Logger, client kube_client.Client) kube_handler.MapFu
...
+		if !ok {
+			l.Error(nil, "unexpected error converting object to Service", "typ", reflect.TypeOf(obj))
+			return nil
+		}
+
+		var routes gatewayapi.HTTPRouteList
+		if err := client.List(ctx, &routes, kube_client.MatchingFields{
+			servicesOfRouteField: kube_client.ObjectKeyFromObject(svc).String(),
+		}); err != nil {
+			l.Error(nil, "unexpected error listing HTTPRoutes")
```
bartsmykla, line 357: Don't we want to log an underlying error here?
```suggestion
			l.Error(err, "unexpected error listing HTTPRoutes")
```

bartsmykla (approved):
One question but lgtm overall

PR #10205, opened 2024-05-09
docs(MADR): mesh service universal ux

jakubdyszkiewicz:
### Checklist prior to review


docs/madr/decisions/041-meshservice-universal-ux.md
```diff
@@ -0,0 +1,134 @@
...
+* We keep the same flow without any extra action from the user
+* Flexibility for users to define their own template
+
+**Disadvantages***
+* Complexity of implementing this object
+
+### `MeshService` autogenerated implicitly
+
+Instead of introducing explicit `MeshServiceGenerator` we can just have opinionated process just like Service -> MeshService on Kube.
+Fields would be taken from Dataplane just like described in `MeshServiceGenerator`.
```
michaelbeaumont, line 128: Is this saying the above example is the proposed implicit algo? Maybe worth making the conversion explicit here

docs/madr/decisions/041-meshservice-universal-ux.md
```diff
@@ -0,0 +1,141 @@
...
+**Disadvantages***
+* Complexity of implementing this object
+
+### `MeshService` autogenerated implicitly
+
+Instead of introducing explicit `MeshServiceGenerator` we can just have opinionated process just like Service -> MeshService on Kube.
+
+Fields we would generate:
+* name - value of `kuma.io/service` tag
+* selector.dataplane.tags - `kuma.io/service` key + value
```
michaelbeaumont, line 131: Does a user need to be able to add an additional tag? Like if all of my `backend-api` Dataplanes should have `team: backend`, how can I do this?

jakubdyszkiewicz: I don't think so. Technically we can take all common tags from all DPPs of the same kuma.io/service and put it in selector, but this would select the same set of DPPs. What's the value in doing so?

If they want to manipulate selectors they need to explicitly create MeshService.

michaelbeaumont: Yeah. I guess the more common customization use case would be labeling the generated MeshService.

Automaat (approved):

jijiechen (approved):

PR #10206, opened 2024-05-09
refactor(transparent-proxy): simplify more tproxy configuration

bartsmykla:
I recommend doing review on commit per commit bases

#### Got rid of unnecessary internal `defaultConfig` function

We had `defaultConfig` and `DefaultConfig`, where latter was just calling former function.

#### Simplified configuration handling by removing `MergeConfigWithDefaults`

This commit removes the unnecessary and hard-to-maintain `MergeConfigWithDefaults` function. Instead, we now use `DefaultConfig` directly and explicitly set modified values in commands like `kumactl install transparent-proxy` and CNI.

I've also streamlined the `transparentProxyArgs` structure by removing parameters that can be directly configured via the `Config` structure. This paves the way for potentially removing this intermediate structure altogether in the future, although that would involve more complex changes in flag parsing.

#### Added `StoreFirewalld` parameter to `Config` structure

I think it's the logical place for this value as it's basically part of the configuration.

#### Replaced writes to Writers with pure `fmt.Fprintln` to make code cleaner

We don't have to cast string literals to `[]byte` now and don't have to explicitly ignore errors.

### Checklist prior to review

  - It won't
  - There is no change in logic which means there should be no change in generated iptables rules. Existing unit tests passes
  - There is no need
  - There is no need

### DIFF
--- a/app/kumactl/cmd/install/install_transparent_proxy.go
+++ b/app/kumactl/cmd/install/install_transparent_proxy.go
@@ -30,7 +30,6 @@ type transparentProxyArgs struct {
 	UID                            string
 	User                           string
 	AgentDNSListenerPort           string
-	StoreFirewalld                 bool
 	SkipDNSConntrackZoneSplit      bool
 }
 
@@ -51,7 +50,6 @@ func newInstallTransparentProxy() *cobra.Command {
 		UID:                            "",
 		User:                           "",
 		AgentDNSListenerPort:           "15053",
-		StoreFirewalld:                 false,
 		SkipDNSConntrackZoneSplit:      false,
 	}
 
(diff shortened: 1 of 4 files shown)


app/kumactl/cmd/install/install_transparent_proxy.go
```diff
@@ -259,56 +247,56 @@ func findUidGid(uid, user string) (string, string, error) {
 	return u.Uid, u.Gid, nil
 }
 
-func parseArgs(cmd *cobra.Command, args *transparentProxyArgs) (config.Config, error) {
+func parseArgs(cfg *config.Config, args *transparentProxyArgs) error {
```
michaelbeaumont, line 249: IMO this could just return a new config

michaelbeaumont (approved):

app/kumactl/cmd/install/install_transparent_proxy.go
bartsmykla: To be honest I would like to get rid of `transparentProxyArgs` completely soon, that's why I didn't spent a lot of time on improving it.

PR #10220, opened 2024-05-13
chore(cni): don't redirect inbound for gateway in CNI

bartsmykla:
`Config` has both - inbound and outbound traffic redirection enabled by default (change introduced in https://github.com/kumahq/kuma/pull/10206), which is wrong in case of gateway so `GatewayIPV6CNIV2` tests were failing (ref. https://github.com/kumahq/kuma/actions/runs/9030440854/job/24815380164)

### Checklist prior to review

  - No relevant issues
  - It won't
  - There is no need
  - There is no need

> Changelog: skip


lahabana:
@bartsmykla some description to this PR please

lahabana (approved):

PR #10253, opened 2024-05-16
docs(MADR): meshpassthrough design

lukidzi:
### Checklist prior to review

### DIFF
--- a/docs/madr/decisions/050-mesh-passthrough.md
+++ b/docs/madr/decisions/050-mesh-passthrough.md
@@ -33,7 +33,7 @@ New `MeshPassthrough` policy seems like cleaner, more flexible.
 
 ### New MeshPassthrough policy
 
-We would like to introduce a new policy `MeshPassthrough` which allows exposing domains, IPs, CIDRs through specific sidecars. Policy should be only configured by the MeshOperator and allowed in `kuma-system`.
+We would like to introduce a new policy `MeshPassthrough` which allows exposing domains, IPs, CIDRs through specific sidecars.
 
 
 ```yaml
@@ -57,22 +57,22 @@ spec:
       protocol: tls    
 ```
 
-`MeshPassthrough` should allow targeting specific subset of proxies and apply configuration only on them. We should support following kinds: `Mesh`, `MeshSubset` and in the future `MeshService`.
+`MeshPassthrough` should allow targeting specific subset of proxies and apply configuration only on them. We should support following kinds: `Mesh` and`MeshSubset`.
 
 * **matchAppend**: list of all domains/ips/cidrs supported through the selected sidecars. In case there is many polcies matching the same sidecar, lists are merged.
 * **type**: type of the entry, one of `Domain`, `IP` or `CIDR`
 * **value**: value for the entry
 * **port**: port on which service can communicate
 * **protocol**: defines a protocol of the communication. Possible values:
-  * `tls`: should be used when TLS traffic is originated by the client application in the case the `kuma.io/protocol` would be tcp
-  * `tcp`: WARNING: shouldn't be used when match has only domains. On the TCP level we are not able to disinguish domain, in this case it is going to hijack whole traffic on this port. We are going to validate configuration and do not apply config when protocol is tcp and `type: Domain`.
-  * `grpc`
-  * `http`
-  * `http2`
+  * `Tls`: should be used when TLS traffic is originated by the client application in the case the `kuma.io/protocol` would be tcp
+  * `Tcp`: WARNING: shouldn't be used when match has only domains. On the TCP level we are not able to disinguish domain, in this case it is going to hijack whole traffic on this port. We are going to validate configuration and do not apply config when protocol is tcp and `type: Domain`.
+  * `Grpc`
+  * `Http`
+  * `Http2`
 
 #### Implementation
 
-We can use [Matcher API](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/matching/matching_api.html#matching-api) which got out of a Alpha phase and is stable. We can use it to support different entries on one listener. Together with [passthrough](https://kuma.io/docs/2.7.x/networking/non-mesh-traffic/#outgoing) mode one the mesh, MeshOperator can disable all outgoing traffic except the one provided by `MeshPassthrough`.
+We can use [Matcher API](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/matching/matching_api.html#matching-api) which got out of a Alpha phase and is stable. We can use it to support different entries on one listener. Together with [passthrough](https://kuma.io/docs/2.7.x/networking/non-mesh-traffic/#outgoing) mode one the mesh, MeshOperator can disable all outgoing traffic except the one provided by `MeshPassthrough`. It is worth pointing out that when `passthrough.enabled` is set to `true` on the `Mesh`, `MeshPassthrough` policies have no effect because all outgoing traffic is allowed.
 
 From the envoy configuration point of view, we are going to add filter chain matchers under `outbound:passthrough:ipv4"` listener.
 
@@ -83,6 +83,13 @@ This policy won't apply without transparent proxy.
 #### ZoneEgress
 
 TBA
+
+#### Security
+
+It is advised that the MeshOperator is responsible for the `MeshPassthrough` policy. This policy can introduce traffic outside of the mesh or even the cluster, and the MeshOperator should be aware of this. It is also important to note that once a policy is matched by tags, any service owner can add specific tags that match the policy. Depending on the security requirements, the MeshOperator:
+
+* can forbid service owners from using these tags,
+* should conduct auditing and proceed based on the audit results.
 
 ### Other options
 
--- a/docs/madr/decisions/050-mesh-passthrough.md
+++ b/docs/madr/decisions/050-mesh-passthrough.md
@@ -50,14 +50,14 @@ spec:
     tags:
       chatgpt.io/access: "true"
   default:
-    matchAppend:
+    appendMatch:
     - type: Domain
       value: api.chatgpt.com
       port: 443
       protocol: tls    
 ```
 
-`MeshPassthrough` should allows targeting specific subset of proxies and apply configuration only on them. We should support following kinds: `Mesh`, `MeshSubset` and in the future `MeshService`.
+`MeshPassthrough` should allow targeting specific subset of proxies and apply configuration only on them. We should support following kinds: `Mesh`, `MeshSubset` and in the future `MeshService`.
 
 * **matchAppend**: list of all domains/ips/cidrs supported through the selected sidecars. In case there is many polcies matching the same sidecar, lists are merged.
 * **type**: type of the entry, one of `Domain`, `IP` or `CIDR`
@@ -78,7 +78,7 @@ From the envoy configuration point of view, we are going to add filter chain mat
 
 #### Universal without transparent proxy
 
-This policy won't apply when no transparent proxy.
+This policy won't apply without transparent proxy.
 
 #### ZoneEgress
 
--- a/docs/madr/decisions/050-mesh-passthrough.md
+++ b/docs/madr/decisions/050-mesh-passthrough.md
@@ -51,6 +51,7 @@ spec:
     tags:
       chatgpt.io/access: "true"
   default:
+    enabled: false
     appendMatch:
     - type: Domain
       value: api.chatgpt.com
@@ -60,6 +61,7 @@ spec:
 
 `MeshPassthrough` should allow targeting specific subset of proxies and apply configuration only on them. We should support following kinds: `Mesh` and`MeshSubset`.
 
+* **enabled**: defines if sidecar should be in a passthrough mode and allow all outside traffic. If `true`, `matchAppend` is not used. Default: `false`.
 * **matchAppend**: list of all domains/ips/cidrs supported through the selected sidecars. In case there is many polcies matching the same sidecar, lists are merged.
 * **type**: type of the entry, one of `Domain`, `IP` or `CIDR`
 * **value**: value for the entry
@@ -71,10 +73,16 @@ spec:
   * `Http`
   * `Http2`
 
+#### Mesh with a passthrough
+
+If there is no `MeshPassthrough` policy targetting specific dataplane we are taking the default value from the `Mesh` object. If there is neither `Mesh` setting nor `MeshPassthrough` policy we are using the default for the control-plane which is `enabled` by default.
+
 #### Security
 
-Relying on labels means that it's important to either:
-1. Make sure that service owners can't freely set them (using something like [kyverno](https://kyverno.io/policies/other/allowed-label-changes/allowed-label-changes/))
+It is advised that the MeshOperator is responsible for the `MeshPassthrough` policy. This policy can introduce traffic outside of the mesh or even the cluster, and the MeshOperator should be aware of this
+If you want to secure access to `MeshPassthrough` to specific services, you must choose them manually. 
+If you rely on tags in the top-level `targetRef` you might consider securing them by one of the:
+1. Make sure that service owners can't freely set them (using something like [kyverno](https://kyverno.io/policies/other/allowed-label-changes/allowed-label-changes/), OPA)
 2. Accept the risk of being able to "impersonate" a passthrough label and rely on auditing
 
 #### Implementation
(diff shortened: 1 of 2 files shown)


docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,130 @@
...
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  targetRef:
+    kind: MeshSubset # Mesh, MeshSubset
+    tags:
+      chatgpt.io/access: "true"
+  default:
+    matchAppend:
```
slonka, line 53: I think it needs to start with `append` like `appendModifications`:

https://github.com/kumahq/kuma/blob/fdd9b052240a3c47b51abce675bd30bb3388a8df/pkg/plugins/policies/meshproxypatch/api/v1alpha1/schema.yaml#L23

```suggestion
    appendMatch:
```

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,130 @@
...
+    - type: Domain
+      value: api.chatgpt.com
+      port: 443
+      protocol: tls    
+```
+
+`MeshPassthrough` should allows targeting specific subset of proxies and apply configuration only on them. We should support following kinds: `Mesh`, `MeshSubset` and in the future `MeshService`.
```
slonka, line 60:
```suggestion
`MeshPassthrough` should allow targeting specific subset of proxies and apply configuration only on them. We should support following kinds: `Mesh`, `MeshSubset` and in the future `MeshService`.
```

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,130 @@
...
+* No confusion with Universal: There is no issue with non-transparent proxies.
+
+### Negative Consequences
+
+* Additional complexity: Introduction of another policy.
+* Resource management: The need to manage two separate resources.
+
+### New MeshPassthrough policy
+
+We would like to introduce a new policy `MeshPassthrough` which allows exposing domains, IPs, CIDRs through specific sidecars. Policy should be only configured by the MeshOperator and allowed in `kuma-system`.
```
lobkovilya, line 36: > Policy should be only configured by the MeshOperator and allowed in `kuma-system`.

I'm not sure if we want to limit this on policy levels. I'd allow this in any namespace and then if Mesh Operator needs to restrict access they simply create a k8s RBAC rule. Also, I can imagine some teams to have enough autonomy to manage their own MeshPassthrough policies

lahabana: Yeah I think it's a user defined behaviour. The recommendation is likely to only allow this on global for admins but that's up to the end user.

docs/madr/decisions/050-mesh-passthrough.md
lahabana: I wouldn't allow `MeshService` people should you dataplane tags

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,130 @@
...
+* **matchAppend**: list of all domains/ips/cidrs supported through the selected sidecars. In case there is many polcies matching the same sidecar, lists are merged.
+* **type**: type of the entry, one of `Domain`, `IP` or `CIDR`
+* **value**: value for the entry
+* **port**: port on which service can communicate
+* **protocol**: defines a protocol of the communication. Possible values:
+  * `tls`: should be used when TLS traffic is originated by the client application in the case the `kuma.io/protocol` would be tcp
+  * `tcp`: WARNING: shouldn't be used when match has only domains. On the TCP level we are not able to disinguish domain, in this case it is going to hijack whole traffic on this port. We are going to validate configuration and do not apply config when protocol is tcp and `type: Domain`.
+  * `grpc`
+  * `http`
+  * `http2`
```
lahabana, line 71: I think these should be correct enums (there's no backward compat requirement here) so Tls, Tcp, Grpc, Http, Http2 maybe?

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Implementation
+
+We can use [Matcher API](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/matching/matching_api.html#matching-api) which got out of a Alpha phase and is stable. We can use it to support different entries on one listener. Together with [passthrough](https://kuma.io/docs/2.7.x/networking/non-mesh-traffic/#outgoing) mode one the mesh, MeshOperator can disable all outgoing traffic except the one provided by `MeshPassthrough`.
```
lahabana, line 75: Worth stating that this policy is useless if the Mesh has `passthrough.enabled: true`

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,130 @@
...
+
+From the envoy configuration point of view, we are going to add filter chain matchers under `outbound:passthrough:ipv4"` listener.
+
+#### Universal without transparent proxy
+
+This policy won't apply when no transparent proxy.
+
+#### ZoneEgress
+
+TBA
```
lahabana, line 85: If we were to do this we'd do TCP tunnel to Egress and then configure the Egress (I'm not sure that'd work)? Or would HTTP Connect to Egress work?

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,137 @@
...
+
+`MeshPassthrough` should allow targeting specific subset of proxies and apply configuration only on them. We should support following kinds: `Mesh` and`MeshSubset`.
+
+* **matchAppend**: list of all domains/ips/cidrs supported through the selected sidecars. In case there is many polcies matching the same sidecar, lists are merged.
+* **type**: type of the entry, one of `Domain`, `IP` or `CIDR`
+* **value**: value for the entry
+* **port**: port on which service can communicate
+* **protocol**: defines a protocol of the communication. Possible values:
+  * `Tls`: should be used when TLS traffic is originated by the client application in the case the `kuma.io/protocol` would be tcp
+  * `Tcp`: WARNING: shouldn't be used when match has only domains. On the TCP level we are not able to disinguish domain, in this case it is going to hijack whole traffic on this port. We are going to validate configuration and do not apply config when protocol is tcp and `type: Domain`.
```
lahabana, line 68:
```suggestion
  * `Tcp`: WARNING: can't be used when `match.type == Domain` (at TCP level we are not able to disinguish domain, in this case it is going to hijack whole traffic on this port). This will be validated in the config.
```

Right?

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### ZoneEgress
+
+TBA
+
+#### Security
+
+It is advised that the MeshOperator is responsible for the `MeshPassthrough` policy. This policy can introduce traffic outside of the mesh or even the cluster, and the MeshOperator should be aware of this. It is also important to note that once a policy is matched by tags, any service owner can add specific tags that match the policy. Depending on the security requirements, the MeshOperator:
+
+* can forbid service owners from using these tags,
```
lahabana, line 91:
```suggestion
* control how service owners use these tags (OPA, Kyverno, MeshConstraints...),
```

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,137 @@
...
+* should conduct auditing and proceed based on the audit results.
```
lahabana, line 92:
```suggestion
* should conduct audits and proceed based on the audit results.
```

lahabana (approved):
LGTM just nits

docs/madr/decisions/050-mesh-passthrough.md
slonka: I only thought about TCP tunnel to Egress but we can check other alternatives.

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Positive Consequences
+
+* One policy handling both cases
+
+### Negative Consequences
+
+* Confusing way of targeting `MeshExternalServices` with policies but not affecting `Passthrough` clusters
+* Match with many entries doesn't fit logically with `type: Managed`
+* Not possible to apply only on specific sidecar
```
jakubdyszkiewicz, line 137: Does all matches go to one passthrough cluster or do we have separate clusters?
Or to rephrase, do we have separate stats or not?

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,137 @@
...
+## Decision Outcome
+
+New `MeshPassthrough` policy seems like cleaner, more flexible.
+
+### Positive Consequences
+
+* More flexibility: We can target specific proxies.
+* Clarity: There is no confusion about which fields are involved in the policy configuration.
+* Simplicity: Policies are cleaner and cannot be configured with other policies, unlike MeshExternalService.
+* No confusion with Universal: There is no issue with non-transparent proxies.
```
slonka, line 27:
```suggestion
* No confusion with Universal: There is no issue with non-transparent proxies.
* Can have separate clusters for each "match" and separate metrics for them.
```

docs/madr/decisions/050-mesh-passthrough.md
slonka: as mentioned on slack we have 3 options:
1. one cluster for all policies
2. cluster per policy (with multiple matches)
3. always cluster per match
4. additional field to specify stats name and only then generate separate cluster

1. is of least value
2. is problematic because we don't have that information (maybe we could use inspect API).
3. is a bit problematic because we'd have to use "type" + "value" + "port" + "protocol" - not easiest to read but can be that way
4. mix of 1 and 3

----

Let's start with 1. and go with 4. if we need to

jakubdyszkiewicz (approved):

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,144 @@
...
+  * `Grpc`
+  * `Http`
+  * `Http2`
+
+#### Implementation
+
+We can use [Matcher API](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/matching/matching_api.html#matching-api) which got out of a Alpha phase and is stable. We can use it to support different entries on one listener. Together with [passthrough](https://kuma.io/docs/2.7.x/networking/non-mesh-traffic/#outgoing) mode one the mesh, MeshOperator can disable all outgoing traffic except the one provided by `MeshPassthrough`. It is worth pointing out that when `passthrough.enabled` is set to `true` on the `Mesh`, `MeshPassthrough` policies have no effect because all outgoing traffic is allowed.
+
+From the envoy configuration point of view, we are going to add filter chain matchers under `outbound:passthrough:ipv4"` listener.
+
```
lahabana, line 79: What are the implication with proxy policies like MeshTrace (ones that don't depend on from or to)?

docs/madr/decisions/050-mesh-passthrough.md
lahabana: When you say:

> one cluster for all policies

You mean for all `MeshPassthrough` policy?

lahabana: What would be the envoy cluster name and stat prefix here? Same again the Envoy XDS config should become a bit more "official APIs"

slonka: > You mean for all MeshPassthrough policy?

yes

> What would be the envoy cluster name and stat prefix here?

In which case? Let me try to come up with one for all of them:

1. meshpassthrough
2. meshpassthrough_{policyName}
3. meshpassthrough_{type}_{value}_{port}_{protocol}
4. meshpassthrough_{userSuppliedName} and `meshpassthrough_{type}_{value}_{port}_{protocol}` for the ones that don't have `userSuppliedName`

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,144 @@
...
+* **type**: type of the entry, one of `Domain`, `IP` or `CIDR`
+* **value**: value for the entry
+* **port**: port on which service can communicate
+* **protocol**: defines a protocol of the communication. Possible values:
+  * `Tls`: should be used when TLS traffic is originated by the client application in the case the `kuma.io/protocol` would be tcp
+  * `Tcp`: WARNING: can't be used when `match.type == Domain` (at TCP level we are not able to disinguish domain, in this case it is going to hijack whole traffic on this port). This will be validated in the config.
+  * `Grpc`
+  * `Http`
+  * `Http2`
+
```
slonka, line 73:
```suggestion

#### Security

Relying on labels means that it's important to either:
1. Make sure that service owners can't freely set them (using something like [kyverno](https://kyverno.io/policies/other/allowed-label-changes/allowed-label-changes/))
2. Accept the risk of being able to "impersonate" a passthrough label and rely on auditing

```

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,144 @@
...
+* **matchAppend**: list of all domains/ips/cidrs supported through the selected sidecars. In case there is many polcies matching the same sidecar, lists are merged.
+* **type**: type of the entry, one of `Domain`, `IP` or `CIDR`
+* **value**: value for the entry
+* **port**: port on which service can communicate
+* **protocol**: defines a protocol of the communication. Possible values:
+  * `Tls`: should be used when TLS traffic is originated by the client application in the case the `kuma.io/protocol` would be tcp
+  * `Tcp`: WARNING: can't be used when `match.type == Domain` (at TCP level we are not able to disinguish domain, in this case it is going to hijack whole traffic on this port). This will be validated in the config.
+  * `Grpc`
+  * `Http`
+  * `Http2`
```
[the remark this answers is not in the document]
jakubdyszkiewicz: I don't have strong opinion. I'd go with lowercase. I think it's more intuitive and more consistent with other parts of the system.

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,150 @@
...
+
+#### ZoneEgress
+
+##### Static IP and Domains
+
+Static IP and Domain we can achive the same way we are doing it now. By setting SNI on the cluster and later matching in filter chain match.
+
+##### CIDR and Wildcard domains
+
+This case is a bit more problematic. We can use tunneling over [HTTP2 Connect](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades#tunneling-tcp-over-http). Client's application needs to resolve address and later sidecar sends the requests to egress. On the `egress` we need a cluster which supports routing based on the `original_dst_lb_config` and listener supporting `CONNECT`. In this case we cannot match based on SNI because we are sending this traffic over HTTP. In this case, we would need to terminate TLS. [Example Evoy configuration](https://gist.github.com/lukidzi/34cd94528fe6a3d87dd2f2411ff39018).
```
lukidzi, line 98:
```suggestion
This case is a bit more problematic. We can use tunneling over [HTTP2 Connect](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades#tunneling-tcp-over-http). Client's application needs to resolve address and later sidecar sends the requests to egress. On the `egress` we need a cluster which supports routing based on the `original_dst_lb_config` and listener supporting `CONNECT`. In this case we cannot match based on SNI because we are sending this traffic over HTTP. [Example Evoy configuration](https://gist.github.com/lukidzi/34cd94528fe6a3d87dd2f2411ff39018).
```

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,150 @@
...
+
+#### Implementation
+
+We can use [Matcher API](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/matching/matching_api.html#matching-api) which got out of a Alpha phase and is stable. We can use it to support different entries on one listener. Together with [passthrough](https://kuma.io/docs/2.7.x/networking/non-mesh-traffic/#outgoing) mode one the mesh, MeshOperator can disable all outgoing traffic except the one provided by `MeshPassthrough`. It is worth pointing out that when `passthrough.enabled` is set to `true` on the `Mesh`, `MeshPassthrough` policies have no effect because all outgoing traffic is allowed.
+
+From the envoy configuration point of view, we are going to add filter chain matchers under `outbound:passthrough:ipv4"` listener.
+
+#### Universal without transparent proxy
+
+This policy won't apply without transparent proxy.
```
lukidzi, line 88: It's the first policy that doesn't apply to some specific set of dpps. We should somehow present that to the user. Maybe we need a dataplane 
* status field
* maybe proxyType(`sidecar/gateway/..`)
* another type for transparent/non transparent
* auto generated tags for transparent/non transparent

slonka: we don't care about this case that much

docs/madr/decisions/050-mesh-passthrough.md
```diff
@@ -0,0 +1,150 @@
...
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  targetRef:
+    kind: MeshSubset # Mesh, MeshSubset
+    tags:
+      chatgpt.io/access: "true"
+  default:
+    appendMatch:
```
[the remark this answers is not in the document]
slonka: @lahabana is ok with it

docs/madr/decisions/050-mesh-passthrough.md
slonka: let's cover this as an issue and implement if needed

Automaat (approved):

lobkovilya (approved):

PR #10270, opened 2024-05-17
feat(report): add more info in the report

lahabana:
- Remove `TestRuntime` as it was just unecessary complexity
- Add clusterCreationTime to easily figure when a cluster was first created (roughly)
- Add a zoneHashName to be able to easily uniquely identity the number of zones in a cluster

This should provide better metadata for understand usage

### Checklist prior to review

### DIFF
--- a/pkg/clusterid/reader.go
+++ b/pkg/clusterid/reader.go
@@ -31,13 +31,13 @@ func (c *clusterIDReader) Start(stop <-chan struct{}) error {
 	for {
 		select {
 		case <-ticker.C:
-			clusterID, err := c.read(ctx)
+			clusterID, clusterCreationTime, err := c.read(ctx)
 			if err != nil {
 				log.Error(err, "could not read cluster ID") // just log, do not exit to retry operation
 			}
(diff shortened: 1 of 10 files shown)


app/cni/pkg/cni/utils_linux.go
```diff
@@ -15,15 +15,15 @@ import (
 // copied from https://github.com/kubernetes/kubernetes/blob/v1.24.3/pkg/util/procfs/procfs_linux.go#L99
 // with small modifications
 // and return their pid(s)
-func pidOf(name string) ([]int, error) {
+func pidOf(name string) ([]int, error) { //nolint:unused
```
michaelbeaumont, line 18: why do this? this file is now guaranteed to be forgotten if the calls disappear.

michaelbeaumont (approved):

app/cni/pkg/cni/utils_linux.go
lahabana: You are right!

PR #10306, opened 2024-05-23
feat(MeshExternalService): implement validator

slonka:
checking if everything passes CI

### Checklist prior to review


pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
```diff
@@ -57,44 +68,59 @@ type Endpoint struct {
 	// +kubebuilder:validation:MinLength=1
 	Address string `json:"address"`
 	// Port of the endpoint
-	Port EndpointPort `json:"port"`
+	Port *EndpointPort `json:"port,omitempty"`
```
lukidzi, line 71: Should it be an `EndpointPort` if it's in the match? Maybe just `Port` is enough

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
```diff
@@ -57,44 +68,59 @@ type Endpoint struct {
...
 	// +kubebuilder:default=false
-	AllowRenegotiation bool `json:"allowRenegotiation"`
+	AllowRenegotiation bool `json:"allowRenegotiation,omitempty"`
 	// Verification section for providing TLS verification details.
-	Verification Verification `json:"verification"`
+	Verification *Verification `json:"verification,omitempty"`
 }
 
 // +kubebuilder:validation:Enum=TLSAuto;TLS10;TLS11;TLS12;TLS13
 type TlsMinMaxVersion string
```
lukidzi, line 89: Maybe just TlsVersion

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/validator.go
```diff
@@ -1,10 +1,144 @@
...
+
 	return verr.OrNil()
 }
+
+func validateTls(tls *Tls) validators.ValidationError {
+	var verr validators.ValidationError
+
+	if tls.Version != nil {
+		path := validators.RootedAt("version")
+		if tls.Version.Min != nil {
```
lukidzi, line 49: Should we validate cases:
- min is Auto and max is specific
- max is Auto and min is specific
- min is higher than max

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/validator.go
```diff
@@ -1,10 +1,144 @@
...
+			}
+		}
+		if tls.Version.Max != nil {
+			if !slices.Contains(allTlsVersions, string(*tls.Version.Max)) {
+				verr.AddErrorAt(path.Field("max"), validators.MakeFieldMustBeOneOfErr("max", allTlsVersions...))
+			}
+		}
+	}
+
+	if tls.Verification != nil {
```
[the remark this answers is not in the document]
slonka: why? if you have enabled = false then we just not apply any TLS settings, it's not invalid to have enabled=true and verification != nil, right?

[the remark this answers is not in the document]
slonka: idea was to not do validation if `enabled=false` -> but we decided offline not to do this

lukidzi (approved):

Issue #10300, opened 2024-05-22
Implement MeshPassthrough


lahabana:
### Description

This was designed in: https://github.com/kumahq/kuma/pull/10253

Now it's approved we can implement it

slonka:
closed via https://github.com/kumahq/kuma/pull/10363 - I'll create issues for follow up work

PR #10314, opened 2024-05-24
feat(meshpassthrough): create API and validators

lukidzi:
### Checklist prior to review
API of the policy based on https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/051-mesh-passthrough.md

There are 2 commits:
* 1st adds an API definition based on https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/051-mesh-passthrough.md
* 2nd adds validator

* validator doesn't allow partial wildcard domain example: `*w.example.com` shouldn't be possible
* also we are not allowing different protocols `http/http2/grpc` on the same port

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- xref: https://github.com/kumahq/kuma/issues/10300

### DIFF
--- a/pkg/plugins/policies/meshpassthrough/api/v1alpha1/testdata/full-invalid.output.yaml
+++ b/pkg/plugins/policies/meshpassthrough/api/v1alpha1/testdata/full-invalid.output.yaml
@@ -19,4 +19,4 @@ violations:
   message: 'using the same port in multiple matches requires the same protocol for
     the following protocols: [grpc http http2]'
 - field: spec.default.appendMatch[7].value
-  message: provided DNS has incorrect value, cannot support partial wildcard
+  message: provided DNS has incorrect value, partial wildcard is currently not supported


slonka (commented):
added a couple of suggestions, also need to update test files because new policy makes one assertion fail in api server tests

pkg/plugins/policies/meshpassthrough/api/v1alpha1/testdata/full-invalid.output.yaml
```diff
@@ -0,0 +1,15 @@
+violations:
+- field: spec.default.appendMatch[0].port
+  message: port must be a valid (1-65535)
+- field: spec.default.appendMatch[0].type
+  message: provided type test is not supported, one of Domain, IP, or CIDR is supported
```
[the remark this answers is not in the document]
slonka: added a case and test to cover that

pkg/plugins/policies/meshpassthrough/api/v1alpha1/validator.go
```diff
@@ -0,0 +1,88 @@
...
+			isValid := govalidator.IsIP(match.Value)
+			if !isValid {
+				verr.AddViolationAt(validators.RootedAt("appendMatch").Index(i).Field("value"), "provided IP has incorrect value")
+			}
+		case "Domain":
+			if match.Protocol == "tcp" {
+				verr.AddViolationAt(validators.RootedAt("appendMatch").Index(i).Field("protocol"), "protocol tcp is not supported for a domain")
+			}
+			if wildcardPartialPrefixPattern.MatchString(match.Value) {
+				verr.AddViolationAt(validators.RootedAt("appendMatch").Index(i).Field("value"), "provided DNS has incorrect value, cannot support partial wildcard")
```
slonka, line 75:
```suggestion
				verr.AddViolationAt(validators.RootedAt("appendMatch").Index(i).Field("value"), "provided DNS has incorrect value, partial wildcard is currently not supported")
```

pkg/plugins/policies/meshpassthrough/api/v1alpha1/validator.go
```diff
@@ -0,0 +1,88 @@
...
+func validateDefault(conf Conf) validators.ValidationError {
+	var verr validators.ValidationError
+	portAndProtocol := map[int]ProtocolType{}
+	for i, match := range conf.AppendMatch {
+		if match.Port != nil && pointer.Deref[int](match.Port) == 0 || pointer.Deref[int](match.Port) > math.MaxUint16 {
+			verr.AddViolationAt(validators.RootedAt("appendMatch").Index(i).Field("port"), "port must be a valid (1-65535)")
+		}
+		if match.Port != nil {
+			if value, found := portAndProtocol[pointer.Deref[int](match.Port)]; found && value != match.Protocol && slices.Contains(notAllowedProtocolsOnTheSamePort, match.Protocol) {
+				verr.AddViolationAt(validators.RootedAt("appendMatch").Index(i).Field("port"), fmt.Sprintf("protocol needs to be the same across different port when using: %v", notAllowedProtocolsOnTheSamePort))
```
slonka, line 51: What is the requirement here? Is it that using the same port in multiple matches requires the same protocol? If so, then should this be:

```suggestion
				verr.AddViolationAt(validators.RootedAt("appendMatch").Index(i).Field("port"), fmt.Sprintf("using the same port in multiple matches requires the same protocol for the following protocols: %v", notAllowedProtocolsOnTheSamePort))
```

maybe we should drop the match to "tcp" and have a warning?

slonka (commented):
overall looks good, just 2 sentences not looking 100% right to me

pkg/plugins/policies/meshpassthrough/api/v1alpha1/validator.go
lukidzi: The requirement here: when there are many different protocols on the same port check if we can distinguish the traffic. 
* When there is TCP/TLS  and (one of http/http2/grpc), it's fine we can figure it out.
* When there is 2 of HTTP/HTTP2/GRPC we cannot distinguish traffic easily

slonka (approved):

PR #10319, opened 2024-05-27
chore(MeshGatewayInstance): supplement creating validation for mode

### LINKED ISSUE #10276: AdmissionController should reject MeshGatewayInstance creation if the CP is global
### Description

In multiple zone mode(1 global-cp-k8s, 1 zone-cp-k8s), I thought that I should deploy the MeshGatewayInstance in the global-cp and sync it to the zones.

Improvements:
1. Supplement a validation for global-cp
2. Add docs clarification for multiple-zone builtin-gateway

Icarus9913:
AdmissionController should reject MeshGatewayInstance creation for Global CP.

close #10276 

### Checklist prior to review


Icarus9913:
WIP: Unit Test

michaelbeaumont (commented):
I think better would be to not install the CRD at all

Icarus9913:
> I think better would be to not install the CRD at all

That sounds like feasible and easier. But furthermore, the user may be curious that why is it that no CRD installed in the Global-CP and we need more obvious clue to remind the user of deploying the MeshGatewayInstance in the Zone-CP.
What do you think?

lahabana:
@michaelbeaumont I think @Icarus9913 is right the user is gonna get it wrong. At least here they'll get a nice error message explaining them that they can't do that here.
Without the CRD they might just become suspicious of their install ("Maybe I didn't install CRDs correctly")

pkg/plugins/runtime/k8s/webhooks/gateway_instance_validator.go
```diff
@@ -49,6 +55,11 @@ func (h *GatewayInstanceValidator) ValidateDelete(ctx context.Context, req admis
 }
 
 func (h *GatewayInstanceValidator) ValidateCreate(ctx context.Context, req admission.Request) admission.Response {
+	if h.cpMode == config_core.Global {
+		return admission.Denied(fmt.Sprintf("Operation not allowed. %s resources like %s can be created only from the %s control plane and not from a %s control plane.",
+			version.Product, "MeshGatewayInstance", strings.ToUpper(config_core.Zone), strings.ToUpper(h.cpMode)))
```
lahabana, line 60:
```suggestion
			version.Product, "MeshGatewayInstance", config_core.Zone, h.cpMode))
```

You can emphasise with quotes instead

lahabana (approved):

michaelbeaumont:
>  Without the CRD they might just become suspicious of their install ("Maybe I didn't install CRDs correctly")

I guess? But then they don't understand how Kuma/gateways work, which they could have figured out by once looking at the docs.
I think users assume they can actually use the CRDs that exist in their cluster, the existence of a CRD is an affordance. For example, this would cause problems with tooling that checks whether the CRD exists before making a decision to use it in some way.

To me it doesn't make sense, but we can merge it if you all think it improves things.

jakubdyszkiewicz:
IMHO there's no harm in merging it.

jakubdyszkiewicz (approved):

PR #10320, opened 2024-05-27
feat(meshservice): ipam

jakubdyszkiewicz:
### Checklist prior to review

IPAM part of the implementation decided in [MADR 046](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/046-meshservice-hostname-vips.md).

Still unit tests. Will create e2e tests once we have hostname generator to actually check that we can communicate over DNS -> VIP.

### DIFF
--- a/pkg/core/resources/apis/meshservice/vip/allocator.go
+++ b/pkg/core/resources/apis/meshservice/vip/allocator.go
@@ -10,7 +10,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 
-	"github.com/kumahq/kuma/pkg/core"
 	meshservice_api "github.com/kumahq/kuma/pkg/core/resources/apis/meshservice/api/v1alpha1"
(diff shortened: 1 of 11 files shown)


pkg/core/resources/apis/meshservice/vip/allocator.go
```diff
@@ -0,0 +1,146 @@
...
+
+	kumaIpam, err := a.initIpam(ctx)
+	if err != nil {
+		return err
+	}
+
+	for {
+		select {
+		case <-ticker.C:
+			start := core.Now()
```
michaelbeaumont, line 80: Why do we have this mocked time again? Can we avoid this unless we know we need it, though I don't think we ever actually need it?

michaelbeaumont (approved):

pkg/core/resources/apis/meshservice/vip/allocator.go
jakubdyszkiewicz: technically we don't need it. Removed.

michaelbeaumont (approved):

PR #10335, opened 2024-05-28
feat(store): update does not wipe out labels

### LINKED ISSUE #10334: .Update() should not wipe out labels
### Description

Calling `.Update(ctx, resource)` wipes out labels from the resource. This is error-prone.
We should preserve existing labels. At the same time, it should be possible to remove labels.

jakubdyszkiewicz:
### Checklist prior to review

Fix #10334

### DIFF
--- a/pkg/core/resources/store/options.go
+++ b/pkg/core/resources/store/options.go
@@ -57,6 +57,7 @@ func CreateWithLabels(labels map[string]string) CreateOptionsFunc {
 type UpdateOptions struct {
 	ModificationTime time.Time
 	Labels           map[string]string
+	ModifyLabels     bool
 }
 
 func ModifiedAt(modificationTime time.Time) UpdateOptionsFunc {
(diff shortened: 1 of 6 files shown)


pkg/core/resources/store/options.go
```diff
@@ -57,6 +57,7 @@ func CreateWithLabels(labels map[string]string) CreateOptionsFunc {
 type UpdateOptions struct {
 	ModificationTime time.Time
 	Labels           map[string]string
+	ModifyLabels     bool
```
lobkovilya, line 60: Why do we need this? It's always true no?

pkg/plugins/resources/k8s/store.go
```diff
@@ -99,7 +99,11 @@ func (s *KubernetesStore) Update(ctx context.Context, r core_model.Resource, fs
 		return errors.Wrapf(err, "failed to convert core model of type %s into k8s counterpart", r.Descriptor().Name)
 	}
 
-	labels, annotations := splitLabelsAndAnnotations(opts.Labels, obj.GetAnnotations())
+	updateLabels := r.GetMeta().GetLabels()
+	if opts.ModifyLabels {
```
michaelbeaumont, line 103: this could probably just be `if opts.Labels != nil` though right?

michaelbeaumont (approved):

pkg/core/resources/store/options.go
michaelbeaumont: Seems like it's always true if `Labels != nil`?

slonka (approved):

Issue #10238, opened 2024-05-15
Implement MeshExternalService


slonka:
### Description

follow up from https://github.com/kumahq/kuma/pull/10229

PR #10336, opened 2024-05-28
feat(MeshExternalService): resource implementation

slonka:
closes https://github.com/kumahq/kuma/issues/10238

~future(?) todos:~
* ~policy matching~ figured out not implemented here
* ~when single domain is present in endpoints just use that for SNI~
* ~add validation that requires serverName when multiple domains are present~
* ~try get rid of deprecated functions~
* ~e2e test on universal~

[manually tested configurations](https://github.com/kumahq/kuma/pull/10336#issuecomment-2152401418)

### Checklist prior to review


pkg/xds/generator/meshexternalservice/generator.go
```diff
@@ -0,0 +1,186 @@
...
+				Origin:   OriginMeshExternalService,
+				Resource: listener,
+			})
+		}
+	}
+
+	return resources, nil
+}
+
+func createCluster(mes *v1alpha1.MeshExternalServiceResource, proxy *core_xds.Proxy, xdsCtx xds_context.Context) (envoy_common.NamedResource, error) {
```
michaelbeaumont, line 102: How does this fit together with MeshHTTPRoute, which at the moment is responsible for listeners and clusters?

lahabana: Yes maybe that code should fall in MeshHTTPRoute and MeshTCPRoute instead which are sort of the Meta policies that create clusters (including subsets)

pkg/xds/topology/outbound.go
```diff
@@ -469,6 +469,9 @@ func createMeshExternalServiceEndpoint(
 	tls := mes.Spec.Tls
 	meshName := mesh.GetMeta().GetName()
 	service := mes.Meta.GetName()
+	if tls == nil {
+		return nil
```
lukidzi, line 473: probably we shouldn't return when tls is nil just don't configure it

slonka: this is just temporary so my test doesn't NPE :D

slonka:
re-checked the test cases from https://github.com/kumahq/kuma/pull/10336#issuecomment-2152401418 on [fc6eed6](https://github.com/kumahq/kuma/pull/10336/commits/fc6eed665c9de4088d3f8d6abc8fae4d5c9521cb)

pkg/xds/topology/outbound.go
```diff
@@ -458,19 +466,171 @@ func fillExternalServicesReachableFromZone(
...
+	meshName := mesh.GetMeta().GetName()
+	name := mes.Meta.GetName()
+	tls := mes.Spec.Tls
+	if tls != nil && tls.Enabled {
+		var caCert, clientCert, clientKey []byte
+		es.TLSEnabled = tls.Enabled
+		es.FallbackToSystemCa = true
+		es.AllowRenegotiation = tls.AllowRenegotiation
+
+		// TODO: if cacert is empty use path - this can't be configured here
```
slonka, line 525: this is done elsewhere

pkg/xds/context/mesh_context_builder.go
```diff
@@ -161,6 +162,11 @@ func (m *meshContextBuilder) BuildIfChanged(ctx context.Context, meshName string
 		dataplanesByName[dp.Meta.GetName()] = dp
 	}
 	meshServices := resources.MeshServices().Items
+	meshExternalServices := resources.MeshExternalServices().Items
+	meshExternalServicesByName := make(map[string]*meshextenralservice_api.MeshExternalServiceResource, len(dataplanes))
```
Automaat, line 166: Isn't `len(dataplanes)` too much, or simply wrong? I don't think that number of mes will correlate with number of ddps

pkg/xds/topology/outbound.go
```diff
@@ -527,19 +535,170 @@ func fillExternalServicesReachableFromZone(
...
+					return errors.Wrap(err, "could not load clientCert")
+				}
+				clientKey, err = loadBytes(ctx, tls.Verification.ClientKey.ConvertToProto(), meshName, loader)
+				if err != nil {
+					return errors.Wrap(err, "could not load clientKey")
+				}
+				es.ClientCert = clientCert
+				es.ClientKey = clientKey
+			}
+			if tls.Verification.ServerName != nil && pointer.Deref(tls.Verification.ServerName) != "" {
```
Automaat, line 615: couldn't it be just: `pointer.Deref(tls.Verification.ServerName) != ""` instead of: tls.Verification.ServerName != nil && pointer.Deref(tls.Verification.ServerName) != ""` Deref will handle nil value

pkg/xds/topology/outbound.go
```diff
@@ -527,19 +535,170 @@ func fillExternalServicesReachableFromZone(
...
+				es.ServerName = pointer.Deref(tls.Verification.ServerName)
+			}
+			if tls.Verification.SubjectAltNames != nil {
```
Automaat, line 618: same here, you can just iterate over nil slice, without checking if this is not nil

pkg/xds/envoy/clusters/configurers.go
```diff
@@ -86,6 +86,16 @@ func ClientSideTLS(endpoints []core_xds.Endpoint) ClusterBuilderOpt {
 	})
 }
 
+func MesClientSideTLS(endpoints []core_xds.Endpoint, systemCaPath string, useCommonTlsContext bool) ClusterBuilderOpt {
```
Automaat, line 89: `Mes` like MeshExternalService? Maybe we could change this name, because this one looks like a typo, I am think someone will rename it sooner or later thinking it is a typo

pkg/plugins/policies/meshtrafficpermission/api/v1alpha1/validator.go
```diff
@@ -39,6 +39,7 @@ func validateFrom(from []From) validators.ValidationError {
 				common_api.MeshSubset,
 				common_api.MeshService,
 				common_api.MeshServiceSubset,
+				common_api.MeshExternalService,
```
Automaat, line 42: Why do we allow setting `MeshExternalService` in from section for MTP?

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go
```diff
@@ -139,6 +139,10 @@ func prepareRoutes(
 ) []api.Route {
 	// policy matching for real MeshService is not yet ready
 	conf := rules.ComputeConf[api.PolicyDefault](toRules, core_rules.MeshService(svc.ServiceName))
+	switch svc.BackendRef.Kind {
+	case common_api.MeshExternalService:
+		conf = rules.ComputeConf[api.PolicyDefault](toRules, core_rules.MeshExternalService(svc.ServiceName))
```
Automaat, line 144: do we need this specific `MeshExternalService` subset? It basicaly is the same as MeshService. Do we expect to match on anything else than service tag in the future?

pkg/xds/context/mesh_context_builder.go
slonka: I think @lukidzi just copied it from a couple of lines above. Will remove it.

pkg/xds/topology/outbound.go
slonka: yeah, Deref handles nils, I'll change it

pkg/xds/envoy/clusters/configurers.go
slonka: we use MES as a short name for this, but I'll rename.

pkg/plugins/policies/meshtrafficpermission/api/v1alpha1/validator.go
slonka: @lukidzi help me out here but I think otherwise egress won't work

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go
slonka: @lukidzi - WDYT? I think it was just safer so that if MeshService changes then we won't change and we can change without the need to change MS.

pkg/plugins/policies/meshtrafficpermission/api/v1alpha1/validator.go
lukidzi: Yes, it's for an egress.

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go
lukidzi: I think it's a good way to distinguish between 2 different types. We shouldn't mix meshservice and meshexternalservice

Automaat (approved):

lukidzi (approved):

PR #1807, opened 2021-04-13
feat(kuma-cp) generate DNS filter

jakubdyszkiewicz:
### Summary

This PR introduces support to configure built-in DNS filter in Kuma.
This is meant to be used only for transparent proxy.

The flow will be as follow:
* IPtables will intercept traffic to DNS on 53 and redirect it to CoreDNS started by Kuma DP
* CoreDNS will be configured to first send a DNS request to Envoy DNS and then if the domain is not found it the fallbacks to /etc/resolv.conf

Current implementation contains one workaround:
* Right now we cannot disable upstream resolvers in DNSFilter therefore we need to configure upstream resolver to dummy resolver that always responds in empty response. CoreDNS will be configured to open extra dns listener with dummy responses.

One question that might occurs is why not use Envoy DNS directly. The reasons are
* DNSFilter does not support SRV records (and probably more since it was introduced just recently)
* We want to start DNS server before Envoy even receives the config. Otherwise, we could not use the hostname as an address for kuma cp

With this change, VIPs do not have to be stable anymore between instances of the control plane. We may drop persisting it but we are keeping it for now for simplicity of implementation.

Other PRs will follow to achieve full functionality.

### Documentation

- Docs will arrive when we have full feature

PR #10338, opened 2024-05-29
chore(dns): remove empty dns port completely

### LINKED ISSUE #8639: Remove completely EmptyDNSPort

Icarus9913:
Refer to #8640, we have drop usage of empty dns port carefully and left some properties for `rollback/downgrade`. Nowadays, we gonna remove it completely.

close #8639 

### Elaboration

- #### What does `EmptyDNSPort` do?
Refer to the first PR [generate DNS filter](https://github.com/kumahq/kuma/pull/1807), we introduced Envoy DNS filter. But at that time, there was no way to disable upstream resolvers in DNSFilter thus we used the coreDNS Empty DNS port as a dummy resolver to respond the empty response.  
Later we found `Users can disable external DNS resolution by omitting the client configuration object.` from https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/udp_filters/dns_filter , which corresponds to the previous PR source codes change https://github.com/kumahq/kuma/pull/8640/files#diff-4188e1209c75b5064ff4c5f6b33180cb84e164a99021efa1fa01d2938e06e1f1L55-L80

- #### Why didn't we drop it completely by #8640 ?
Due to  the first PR codes https://github.com/kumahq/kuma/pull/1807/files#diff-4cf14047d8fdc54a2debb5bfa2b95f2f5cccc7d48d7803cf3700ff2934ad86cbR23-R25 , it would not generate the configurations
```text
	if dnsPort == 0 || emptyDnsPort == 0 {
		return nil, nil
	}
```

If we totally removed the `EmptyDNSPort` from the previous PR #8640 and the user downgrade to a earlier CP version, the user must meet the `if condition` and return nil empty configuration which leads to interruption in the traffic.

### Checklist prior to review


Icarus9913:
cc @jakubdyszkiewicz

slonka (approved):

jakubdyszkiewicz (approved):

Issue #10163, opened 2024-05-06
Add ability to apply policies on the custom namespaces


lobkovilya:
### Description

* it should be possible to apply policy on Zone in a non-system namespace
* the policy is synced to global with `k8s.kuma.io/namespace` label
* the policy affects only DPPs in the namespace it's applied to
* the order while merging should be as specified in [madr](https://github.com/kumahq/kuma/pull/10148) 
* e2e test to check all of the above

note: cross-policy references or referencing MeshGateway is out of the scope of this issue

Issue #10247, opened 2024-05-16
Rework Mesh*Route matching 


lahabana:
### Description

With `MeshService` policy matching (https://github.com/kumahq/kuma/pull/10152) and policy on namespace (https://github.com/kumahq/kuma/pull/10148) we're getting a good story for producer/consumer policies and how to "attach a policy to a service"

The choice was made for `spec.to[].targetRef.kind: MeshService` this goes a little against how MeshHTTPRoute matches which was decided in [MADR-028](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/028-meshhttproute-as-targetref.md).
The strong argument for going with top level targetRef in MADR-28 was matching dataplanes that don't have the route. In this case it's simple enough we just ignore it.

This current implementation has made things a little tricky because policy matching is convoluted where the top level MeshHTTPRoute gets converted to a "fake `to`":

https://github.com/kumahq/kuma/blob/55fa983fc12bdd1a5ea7c1cfe75901320870372b/pkg/plugins/policies/core/rules/rules.go#L346-L401

## suggested changes

1. Allow `spec.to[].targetRef.kind: Mesh*Route` working like `MeshService` (If the service has this route then we apply the matcher on it). It's also great because we can add validation like (disallow TCP parameters on MeshHTTPRoute matches).
2. Deprecate `MeshService` and `MeshHTTPRoute` as a toplevel targetRef (top level targetRef should identify entire proxies) [^1]
3. Make topLevel targetRef optional (in practice people will most often use `targetRef: kind: Mesh`). Having it optional creates less confusion to users about "where they should match stuff".

## benefits

- top level targetRef is something most people can ignore so this makes policies easier to use (this was already discussed in [MADR-005](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/005-policy-matching.md) but I think experience has changed our mind).
- The policy matching algorithm and rules api becomes simpler and there are no indirections anymore top level targetRef identifies proxies that are matched and then `to` and `from` identify parts of the config (namely: inbounds, outbounds and routes).

## open questions

- How does merging work? (#9151 was already asking questions there)
- What happens when we have a timeout on a service and on a route? (multiple filters?, routes is always more specific to service?) 

## xref

- https://github.com/kumahq/kuma/issues/9589
- https://github.com/kumahq/kuma/issues/8681
- https://github.com/kumahq/kuma/issues/9151

### A producer attached policy (all consumers everywhere use this):

```yaml
[...]

PR #10361, opened 2024-05-31
feat(kuma-cp): consumer policies on app's namespace

lobkovilya:
The PR partially implements [madr-52](https://github.com/kumahq/kuma/blob/180976f81dcb6a9b86c4c3fa534e11534060171d/docs/madr/decisions/052-policy-on-namespace.md) and allows applying policies on the custom namespace. 

Based on how policy looks Kuma CP adds a new label:
```yaml
kuma.io/policy-role: producer | consumer | workload-owner | system
```

Producer policies are not supported at this moment, because we can't reference `MeshService` in `to[].targetRef.kind`.

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/10163 


pkg/core/resources/model/resource.go
```diff
@@ -435,6 +436,66 @@ func ResourceOrigin(rm ResourceMeta) (mesh_proto.ResourceOrigin, bool) {
...
+		hasFrom = true
+	}
+
+	if hasTo && !hasFrom {
+		// todo(lobkovilya): detect if the policy is a producer policy when they're supported
+		return mesh_proto.ConsumerPolicyRole
+	} else {
+		return mesh_proto.WorkloadOwnerPolicyRole
+	}
+}
```
Icarus9913, line 490: Do we need to define a `default` return? 
It seems like `if hasTo == false && hasFrom == false`, we'll return an empty string "" ?

pkg/plugins/policies/meshtimeout/api/v1alpha1/validator.go
```diff
@@ -18,19 +20,30 @@ func (r *MeshTimeoutResource) validate() error {
...
+				common_api.MeshService,
+				common_api.MeshServiceSubset,
+				common_api.MeshHTTPRoute,
+			},
+			GatewayListenerTagsAllowed: true,
+		})
+	default:
+		return mesh.ValidateTargetRef(targetRef, &mesh.ValidateTargetRefOpts{
+			SupportedKinds: []common_api.TargetRefKind{
+				common_api.Mesh,
```
Automaat, line 40: shouldn't we allow targeting `MeshHttpRoute`? Or will it be changed in the future with full support for policies on namespace?

pkg/core/resources/model/resource.go
Automaat: why? If I read it correctly, we will return either `mesh_proto.ConsumerPolicyRole` or `mesh_proto.WorkloadOwnerPolicyRole` there are no other return statements in this function

Automaat (approved):

pkg/core/resources/model/resource.go
Icarus9913: Oh, you're right. 
My fault, my head was foggy.

pkg/plugins/policies/meshtimeout/api/v1alpha1/validator.go
lobkovilya: TBH I don't want to allow MeshHTTPRoute in top-level targetRef of namespace-scoped policies to deprecated in the near future with https://github.com/kumahq/kuma/issues/10247.

PR #10362, opened 2024-05-31
feat(HostnameGenerator): apply templates to MeshServices

michaelbeaumont:
- [x] interval config

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- https://github.com/kumahq/kuma/issues/10034


pkg/dns/hostname_generator.go
```diff
@@ -0,0 +1,39 @@
...
+	if !slices.Contains(rt.Config().CoreResources.Enabled, "hostnamegenerators") {
+		logger.Info("HostnameGenerator is not enabled, not starting generator")
+		return nil
+	}
+	generator, err := hostname.NewGenerator(
+		logger,
+		rt.Metrics(),
+		rt.ResourceManager(),
+		100*time.Millisecond,
+		// rt.Config().IPAM.MeshService.AllocationInterval.Duration, // TODO
```
lukidzi, line 28: Question: Should this interval be the same for different resources? `MeshExternalService`needs hostname generator and I think the only needed thing is to add support [here](https://github.com/kumahq/kuma/pull/10362/files#diff-4de19adee6a0c1d08e3f861fc716e75a2b540a5809e65b0a840659c1cf96b996) so it seems that we don't need a separate component for a MeshExternalService

lukidzi (approved):

pkg/dns/hostname_generator.go
michaelbeaumont: No you're right, we should reuse it and we shouldn't need separate configs per resource type.

pkg/core/resources/apis/meshservice/api/v1alpha1/meshservice.go
```diff
@@ -53,8 +63,70 @@ type TLS struct {
...
+const (
+	GeneratedCondition string = "Generated"
+)
+
+const (
+	GeneratedReason     string = "Generated"
+	TemplateErrorReason string = "TemplateError"
+)
+
+type Condition struct {
```
lukidzi, line 79: Should this be a common object? I think `MeshExternalService` could have the same structure

pkg/dns/hostname_generator.go
slonka: Why would we need to pass the interval of IPAM to HostnameGenerator? Do they need to run in sync?

pkg/core/resources/apis/hostnamegenerator/hostname/generator.go
```diff
@@ -0,0 +1,224 @@
...
+			"label": func(key string) (string, error) {
+				val, ok := service.GetMeta().GetLabels()[key]
+				if !ok {
+					return "", errors.Errorf("label %s not found", key)
+				}
+				return val, nil
+			},
+		},
+	)
+	tmpl, err := tmpl.Parse(generator.Spec.Template)
```
jakubdyszkiewicz, line 94: Do we validate this statically on resource apply as well?

pkg/dns/hostname_generator.go
michaelbeaumont: No, we just need _some_ interval

pkg/core/resources/apis/hostnamegenerator/hostname/generator.go
michaelbeaumont: added

pkg/core/resources/apis/meshservice/api/v1alpha1/meshservice.go
michaelbeaumont: I'll leave that to the PR that adds MeshExternalService

pkg/core/resources/apis/hostnamegenerator/hostname/generator.go
```diff
@@ -0,0 +1,224 @@
...
+			return errors.Wrap(err, "couldn't update MeshService status")
+		}
+	}
+
+	return nil
+}
+
+func (g *Generator) NeedLeaderElection() bool {
+	return true
+}
```
jakubdyszkiewicz, line 224: do we have conflict resolution here somewhere?

pkg/core/resources/apis/hostnamegenerator/hostname/generator.go
```diff
@@ -155,8 +178,16 @@ func (g *Generator) generateHostnames(ctx context.Context) error {
 					message = err.Error()
 				}
 				if generated != "" {
-					generationConditionStatus = kube_meta.ConditionTrue
-					reason = meshservice_api.GeneratedReason
+					if key, ok := generatedHostnames[generated]; ok && key != serviceKey {
+						generationConditionStatus = kube_meta.ConditionFalse
+						reason = meshservice_api.CollisionReason
+						message = "Hostname collision"
```
jakubdyszkiewicz, line 184: can we add "Hostname collision with service XYZ"?

jakubdyszkiewicz (approved):

PR #7218, opened 2023-07-10
fix(kuma-dp): pass sockets in metadata from dp to cp

lahabana:
Currently socket paths where generated from both the cp and dp. These relied on `os.TempDir()` which may return different results as they run on 2 different machines.

With this change we make the paths configurable from the kuma-dp process.
The information is then sent in the metadata and can be used during config generation.

From a backward compatibility POV there are 2 things to take care of:

- We may update the CP and have an old DP still running (the metadata won't have the entries for sockets).
- We may have an old DP bootstrap on a newer CP (this bootstrap request won't have the DP set).

Fix #5228

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues --
- [x] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [x] Tests (Unit test, E2E tests, manual test on universal and k8s) --
- [x] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [x] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? --
- [x] Do you need to explicitly set a [`> Changelog:` entry here](../blob/master/CONTRIBUTING.md#submitting-a-patch) or add a `ci/` label to run fewer/more tests?

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

Issue #5228, opened 2022-10-28
access-log-streamer has different TMPDIR than Envoy


slonka:
### What happened?

If `TMPDIR` env variable is different between `kuma-dp` and Envoy that will cause the listener configuration failure. This issue was reported via Kong slack.

Trimmed logs (ping me for full):

```
2022-10-21T10:57:22.767+0200	INFO	access-log-streamer	creating access log pipe	{"file": "/home/USER_X/tmp/kuma-al-a-b.sock"} # line 16
...
[2022-10-21 10:57:26.427][77617][warning][config] [source/common/config/grpc_subscription_impl.cc:126] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) outbound:127.0.0.1:50013: unable to open file '/home/USER_Y/tmp/kuma-al-a-b.sock': No such file or directory # line 134
```

This is because `AccessLogSocketName` is based on `os.TempDir` that is based on `TMPDIR`:

https://github.com/kumahq/kuma/blob/e00d8d886d87a476c350b51a695a2b84e2c2377b/pkg/xds/envoy/sockets.go#L12

Possible solutions:
- introduce an explicit working dir of Kuma DP and hold files there

slonka:
Update: same thing happens to the `metrics-hijacker`

lahabana:
@StuAtKong

StuAtKong:
Got it

PR #10364, opened 2024-06-01
chore(kuma-dp): remove backward compability for tempDir generating for access log and metrics sockets

### LINKED ISSUE #7220: Remove access and metrics socket path backward compatibility
### What happened?

https://github.com/kumahq/kuma/pull/7218 ensured that socket files are created in one place and then passed in metadata.
Things were kept for backward compatibility that we should be able to clean in the future.

Probably can be done from 2.6 (2 versions after 2.4).

Icarus9913:
Refer to #7218, we kept something to meet the backward compability with the access log socket and metrics socket in the past. Now we need to remove them.

close #7220 

### What happened?

As #5228 report, we found that the socket paths where generated from both the CP and DP, which may cause an unmatched path. Just like:
```
/home/USER_X/tmp/kuma-al-a-b.sock
/home/USER_Y/tmp/kuma-al-a-b.sock
```
 

So the #7218 solved it but kept something to meet the backward. This PR gonna remove the legacy.

### Checklist prior to review


pkg/core/xds/sockets.go
```diff
@@ -5,8 +5,6 @@ import (
 	"path/filepath"
 )
 
-// TODO inline these in kuma-dp once backward compatibility is removed. https://github.com/kumahq/kuma/issues/7220
-
 // AccessLogSocketName generates a socket path that will fit the Unix socket path limitation of 104 chars
 func AccessLogSocketName(tmpDir, name, mesh string) string {
 	return socketName(filepath.Join(tmpDir, fmt.Sprintf("kuma-al-%s-%s", name, mesh)))
```
Icarus9913, line 10: Since this function relies on a private function `socketName` and there are also some other functions in this file, maybe we could just keep `AccessLogSocketName` here?

jijiechen: Had an offline discussion with @Icarus9913 , I think if the `socketName` is also a dependent of other functions and those can not be inlined togher, we may only move this single function `AccessLogSocketName` to kuma-dp, and make `socketName` public.

slonka: **All of these** should be moved to kuma-dp and inlined if that makes sense. The core problem is that before some of the paths were generated in the CP that can have different paths than the DP. The correct way of doing things is generating these on `DP` and sending it in metadata.

slonka: ok, looking at this again, the first two can actually be moved to kuma-dp, the third one is used both on kuma-dp (to scrape config) and kuma-cp (to generate Envoy listener) and the forth one only in kuma-dp. So I guess only the third one needs to be shared and there is no workaround. So I think we can make the function public and move whatever we can.

app/kuma-dp/cmd/run.go
```diff
@@ -185,8 +186,8 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
 				DNSPort:             cfg.DNS.EnvoyDNSPort,
 				EnvoyVersion:        *envoyVersion,
 				Workdir:             cfg.DataplaneRuntime.SocketDir,
-				AccessLogSocketPath: core_xds.AccessLogSocketName(cfg.DataplaneRuntime.SocketDir, cfg.Dataplane.Name, cfg.Dataplane.Mesh),
-				MetricsSocketPath:   core_xds.MetricsHijackerSocketName(cfg.DataplaneRuntime.SocketDir, cfg.Dataplane.Name, cfg.Dataplane.Mesh),
+				AccessLogSocketPath: accessLogSocketName(cfg.DataplaneRuntime.SocketDir, cfg.Dataplane.Name, cfg.Dataplane.Mesh),
```
Automaat, line 189: If we are changing this we should do more than just changing how we build the name here. We already sent `Workdir` with bootstrap request. We should build the name in the [policy](https://github.com/kumahq/kuma/blob/af83aed08e98d7526f711d6dc8197d4d0cbf62ca/pkg/plugins/policies/meshaccesslog/plugin/v1alpha1/plugin.go#L47) using workdir. Just like we do it in [MeshMetric plugin](https://github.com/kumahq/kuma/blob/5e7b4f86e2b454131e707c3cb35ba740b5715573/pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go#L159). This applies to access log and old metrics socket

pkg/core/xds/metadata_test.go
```diff
@@ -98,14 +91,12 @@ var _ = Describe("DataplaneMetadataFromXdsMetadata", func() {
...
-				AccessLogSocketPath: "/tmp/kuma-al-dp-1-mesh.sock",
-				MetricsSocketPath:   "/tmp/kuma-mh-dp-1-mesh.sock",
-				DynamicMetadata:     map[string]string{},
+				DynamicMetadata: map[string]string{},
 			},
 		}),
 	)
 
-	It("should fallback to service side generated paths", func() { // remove with https://github.com/kumahq/kuma/issues/7220
+	It("should fallback to service side generated paths", func() {
```
jakubdyszkiewicz, line 99: test names do not make sense anymore. We just removed fallback to have service generated paths. Now we always take paths provided from dpp, right?

I think `should fallback to service side generated paths` is redundant because it's already covered by `Entry("from non-empty node"` and `should fallback to service side generated paths without dpp in metadata` is redundant because it's covered by `Entry("from empty node"`

jakubdyszkiewicz (changes requested):

app/kuma-dp/cmd/run.go
Icarus9913: I totally got your point.
There's a concern that actually we just wanna keep the `accessLogSocketName` function in **private** within kuma-dp rather than keep it public in our pkg dir.. If we do the thing like 
```go
SocketName:   openTelemetrySocketName(proxy.Metadata.WorkDir, backendName),
```

We may have to keep `accessLogSocketName` to public and locate it in our `pkg/core/sockets.go` file again because both `kuma-dp` and `meshaccesslog policy` and other places do call the function. It might looks a little bit weird? 


My original intention is to follow the previous PR  [TODO](https://github.com/lahabana/kuma/blob/fd6ee1218f399f4e7d92e6681411d2810df8f83b/pkg/core/xds/sockets.go#L8-L18)

pkg/core/xds/metadata_test.go
Icarus9913: Yeah, you're correct. Thanks for pointing out the deficiency :)

jakubdyszkiewicz (approved):

PR #10371, opened 2024-06-03
fix(kuma-cp): validate the bandwidth strictly

spacewander:
### Checklist prior to review

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -10,7 +10,7 @@ does not have any particular instructions.
 
 ### MeshFaultInjection responseBandwidth.limit
 
-With [#10371](https://github.com/kumahq/kuma/pull/10371) we strict the validation of `responseBandwidth.limit` field in policies `MeshFaultInjection`. Policies with bad value like `-10kbps` will be rejected.
+With [#10371](https://github.com/kumahq/kuma/pull/10371) we have tightened the validation of the `responseBandwidth.limit` field in `MeshFaultInjection` policy. Policies with invalid values, such as `-10kbps`, will be rejected.
 
 ### MeshRetry tcp.MaxConnectAttempt
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,10 @@ does not have any particular instructions.
 
 ## Upgrade to `2.8.x`
 
+### MeshFaultInjection responseBandwidth.limit
+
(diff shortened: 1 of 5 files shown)


spacewander:
The test failed because unit `mbps` is used:
https://github.com/kumahq/kuma/blob/536ea8c0f79420fd5ab8e7bde8ec1d56f991d947/pkg/plugins/policies/meshfaultinjection/api/v1alpha1/validator_test.go#L37

But according to the code in
https://github.com/kumahq/kuma/blob/536ea8c0f79420fd5ab8e7bde8ec1d56f991d947/pkg/xds/envoy/listeners/v3/util.go#L123-L134

The unit should be case-sensitive.

Which one needs to be changed?

slonka:
> Which one needs to be changed?

the test in https://github.com/kumahq/kuma/blob/536ea8c0f79420fd5ab8e7bde8ec1d56f991d947/pkg/plugins/policies/meshfaultinjection/api/v1alpha1/validator_test.go#L37

according to https://en.wikipedia.org/wiki/Metric_prefix

spacewander:
> 100mbps

So 100mbps means mill-bps. I will vote for changing the test as Envoy itself does not support this unit.

slonka:
> I will vote for changing the test as Envoy itself does not support this unit.

you can just change it

slonka (dismissed):

spacewander:
@slonka 
Updated.

UPGRADE.md
```diff
@@ -8,6 +8,10 @@ does not have any particular instructions.
 
 ## Upgrade to `2.8.x`
 
+### MeshFaultInjection responseBandwidth.limit
+
+With [#10371](https://github.com/kumahq/kuma/pull/10371) we strict the validation of `responseBandwidth.limit` field in policies `MeshFaultInjection`. Policies with bad value like `-10kbps` will be rejected.
```
slonka, line 13:
```suggestion
With [#10371](https://github.com/kumahq/kuma/pull/10371) we have tightened the validation of the `responseBandwidth.limit` field in `MeshFaultInjection` policy. Policies with invalid values, such as `-10kbps`, will be rejected.
```

slonka (approved):

slonka:
Need to fix the examples as well https://kuma.io/docs/2.7.x/policies/meshfaultinjection/#configuration

PR #10378, opened 2024-06-04
feat(kuma-cp): add policy matching api for meshservice

Automaat:
API for policy matching of MeshService. As described in [MADR](https://github.com/kumahq/kuma/pull/10152)

### Checklist prior to review


api/common/v1alpha1/ref.go
```diff
@@ -71,6 +71,17 @@ type TargetRef struct {
 	// all data plane types are targeted by the policy.
 	// +kubebuilder:validation:MinItems=1
 	ProxyTypes []TargetRefProxyType `json:"proxyTypes,omitempty"`
+	// Namespace specifies the namespace of target resource. If empty only resources in policy namespace
+	// will be targeted.
+	// Namespace can be used only with MeshService kind, and when targeting resources in Kubernetes zone.
```
lobkovilya, line 76: I would not mention the specific kind here. Technically `Namespace`, `Labels` and `SectionName` are not limited to `MeshService` and potentially can be used for referencing other kinds.

api/common/v1alpha1/ref.go
```diff
@@ -71,6 +71,17 @@ type TargetRef struct {
...
+	Namespace string `json:"namespace,omitempty"`
+	// Labels are used to select group of MeshServices that match labels. Either Labels or
+	// Name and Namespace can be used.
+	// Labels can be used only with MeshService kind.
+	Labels map[string]string `json:"labels,omitempty"`
+	// SectionName is used to target specific port from MeshService.ports[]. Only traffic to this port will be affected.
```
lobkovilya, line 82: In GAPI docs [SectionName](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.SectionName) is not limited to port, it can select section of various resources. I'd give it a generic comment and added MeshService's port as an example. 

Also, we probably should mention it's not a port value but actually a port name

api/common/v1alpha1/ref.go
Automaat: changed

api/common/v1alpha1/ref.go
Automaat: changed

pkg/core/resources/apis/mesh/validators.go
```diff
@@ -364,16 +364,24 @@ func ValidateTargetRef(
...
+		err.Add(disallowedField("labels", ref.Labels, ref.Kind))
+		err.Add(disallowedField("namespace", ref.Namespace, ref.Kind))
+		err.Add(disallowedField("sectionName", ref.SectionName, ref.Kind))
 	case common_api.MeshService, common_api.MeshHTTPRoute:
-		err.Add(requiredField("name", ref.Name, ref.Kind))
 		err.Add(validateName(ref.Name, opts.AllowedInvalidNames))
 		err.Add(disallowedField("mesh", ref.Mesh, ref.Kind))
 		err.Add(disallowedField("tags", ref.Tags, ref.Kind))
 		err.Add(disallowedField("proxyTypes", ref.ProxyTypes, ref.Kind))
+		if len(ref.Labels) > 0 && (ref.Name != "" || ref.Namespace != "") {
```
Automaat, line 382: things I am not 100% sure:
- should we be able to combine `sectionName` with labels? (from implementation point of view section name will be just another tag, so maybe you can specify it as label?)
- should we be able to specify targetRef with just `sectionName`?
- should we have at least name or namespace if labels are empty? having empty targetRef makes no sense, so we should probably have at least one of thm

michaelbeaumont: 1. Probably we can just restrict this for now and open it up if there's a use case
2. You mean without a name? No IMO 
3. Yes I think if there's no labels it requires a name and if there's no namespace it's the current namespace

Automaat: I've updated validation accordingly to the comment

lobkovilya (approved):

PR #10388, opened 2024-06-05
docs(MADR): multizone sni

jakubdyszkiewicz:
### Checklist prior to review

### DIFF
--- a/docs/madr/decisions/053-sni.md
+++ b/docs/madr/decisions/053-sni.md
@@ -40,7 +40,7 @@ The new format could look like this then:
   Trimmed to 64 chars. If trimmed we add `x` at the end.
 * `<resource-type>` - (required) `ms` (MeshService) or `mes` (MeshExternalService). Abbreviation to at most 3 chars. 
 
-The proposed format has at most 1+16+129+65+3 = 214 chars.
+The proposed format has at most 1+16+129+65+3+3(dots) = 217 chars.
 
 Examples:
 * `ae3b0c44298fc1c14.backend.demo.ms` - MeshService named `backend` in `demo` mesh
--- a/docs/madr/decisions/053-sni.md
+++ b/docs/madr/decisions/053-sni.md
@@ -32,19 +32,22 @@ It means that we can hash the data as long as it's consistent hashing.
 We cannot put tags in the original form because they can violate hostname requirements. For example `/` in `kuma.io/service` is not ok.
 The new format could look like this then:
 `<resource-name>.<mesh-name>.<extra-hash-if-needed>.<resource-type>`
-* `<resource-name>` - name of the resource. Name of MeshService or MeshExternalService
-* `<mesh-name>` - name of the mesh
-* `<extra-hash-if-needed>` - 16 characters of SHA256. In case of MeshService tags it's `SHA256(<tag1>=<value1>,<tag2>=<value2>)` sorted by keys.
-* `<resource-type>` - `meshservice` or `meshexternalservice`
+* `<resource-name>` - (required) name of the resource. Name of MeshService or MeshExternalService
+* `<mesh-name>` - (required) name of the mesh
+* `<extra-hash-if-needed>` - (optional) 16 characters of SHA256. In case of MeshService tags it's `SHA256(<tag1>=<value1>,<tag2>=<value2>)` sorted by keys.
+* `<resource-type>` - (required) `ms` (MeshService) or `mes` (MeshExternalService)
 
 Examples:
-* `backend.demo.meshservice` - MeshService named `backend` in `demo` mesh
-* `redis.kuma-demo.default.meshservice` - MeshService named `redis` in `kuma-demo` namespace in `default` mesh
-* `google.default.meshexternalservice` - MeshExternalService named `google` in `default` mesh
-* `backend.demo.e3b0c44298fc1c.meshservice` - subset of MeshService named `backend` in `demo` mesh for tags `version=v2`
+* `backend.demo.ms` - MeshService named `backend` in `demo` mesh
+* `redis.kuma-demo.default.ms` - MeshService named `redis` in `kuma-demo` namespace in `default` mesh
+* `google.default.mes` - MeshExternalService named `google` in `default` mesh
+* `backend.demo.e3b0c44298fc1c.ms` - subset of MeshService named `backend` in `demo` mesh for tags `version=v2`
 
 Resource name can be at most 255 chars long. Mesh should not be longer than 64 chars.
-If we go over the limit we compute SHA256 hash `<resource-name>.<mesh-name>.<extra-hash-if-needed>` and use 128 chars of it.
+If we go over the limit we compute SHA256 hash `<resource-name>.<mesh-name>.<extra-hash-if-needed>`.
+For example:
+* `super-long-name-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.default.e3b0c44298fc1c14.ms`
+  Changes to `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.ms`
 
 The format is only relevant for the new resources (MeshService, MeshExternalService, MeshMultizoneService),
 we don't want to modify existing format to avoid breaking changes.
--- a/docs/madr/decisions/053-sni.md
+++ b/docs/madr/decisions/053-sni.md
@@ -1,0 +1,57 @@
+# SNI for MeshService, MeshExternalService etc.
+
+* Status: accepted
+
+## Context and Problem Statement
+
+Currently, the SNI format is `<kuma.io/service>{mesh=<mesh>,tag1=value1}`. For example:
+* `backend{mesh=demo}`
+* `backend{mesh=demo,version=2}`
+
+We need to introduce SNI format for `MeshService` and `MeshExternalService` etc.
+Current format does not work, because we need to differentiate between `kuma.io/service` and MeshService.
+Current format also has other problems:
+* It violates [SNI spec](https://datatracker.ietf.org/doc/html/rfc3546#page-8). It should be a valid hostname. Using characters like `,{/` is not valid.
+  Envoy does not complain about it right now, but it can bite us in the future.
+* It can be too long. See https://github.com/kumahq/kuma/issues/8915
+
+## Considered Options
+
+* New format
+
+## Decision Outcome
+
+Chosen option: "New format", because it solves existing problem and lets us differentiate MeshService from old service.
+
+## Pros and Cons of the Options
+
+### New format
+
+It's important to note that we never parse SNI. We only use it for multizone communication on zone proxies to associate to which service pass the data.
+It means that we can hash the data as long as it's consistent hashing.
+We cannot put tags in the original form because they can violate hostname requirements. For example `/` in `kuma.io/service` is not ok.
+The new format could look like this then:
+`<resource-name>.<mesh-name>.<extra-hash-if-needed>.<resource-type>`
+* `<resource-name>` - name of the resource. Name of MeshService or MeshExternalService
+* `<mesh-name>` - name of the mesh
+* `<extra-hash-if-needed>` - 16 characters of SHA256. In case of MeshService tags it's `SHA256(<tag1>=<value1>,<tag2>=<value2>)` sorted by keys.
+* `<resource-type>` - `meshservice` or `meshexternalservice`
+
+Examples:
+* `backend.demo.meshservice` - MeshService named `backend` in `demo` mesh
+* `redis.kuma-demo.default.meshservice` - MeshService named `redis` in `kuma-demo` namespace in `default` mesh
+* `google.default.meshexternalservice` - MeshExternalService named `google` in `default` mesh
+* `backend.demo.e3b0c44298fc1c.meshservice` - subset of MeshService named `backend` in `demo` mesh for tags `version=v2`
+
+Resource name can be at most 255 chars long. Mesh should not be longer than 64 chars.
+If we go over the limit we compute SHA256 hash `<resource-name>.<mesh-name>.<extra-hash-if-needed>` and use 128 chars of it.
+
+The format is only relevant for the new resources (MeshService, MeshExternalService, MeshMultizoneService),
+we don't want to modify existing format to avoid breaking changes.
+
+Advantages:
+* It solves existing problems
+
+Disadvantages:
+* We can lose a bit of visibility if we just look at SNI, because we always hash tags (or even everything).
+  However, it's still clear what specific filter chain of the SNI points to by cluster name and LB subset. 
--- a/docs/madr/decisions/053-sni.md
+++ b/docs/madr/decisions/053-sni.md
@@ -30,23 +30,26 @@ Chosen option: "New format", because it solves existing problem and lets us diff
 It's important to note that we never parse SNI. We only use it for multizone communication on zone proxies to associate to which service pass the data.
 It means that we can hash the data as long as it's consistent hashing.
 We cannot put tags in the original form because they can violate hostname requirements. For example `/` in `kuma.io/service` is not ok.
+
 The new format could look like this then:
-`<format-version><hash>.<resource-name>.<mesh-name>.<resource-type>`
+`<format-version><hash>.<resource-name>.<port>.<mesh-name>.<resource-type>`
 * `<format-version>` - (required) `a`. 1 character. It's so we can evolve format in the future.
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/053-sni.md
```diff
@@ -0,0 +1,57 @@
...
+
+It's important to note that we never parse SNI. We only use it for multizone communication on zone proxies to associate to which service pass the data.
+It means that we can hash the data as long as it's consistent hashing.
+We cannot put tags in the original form because they can violate hostname requirements. For example `/` in `kuma.io/service` is not ok.
+The new format could look like this then:
+`<resource-name>.<mesh-name>.<extra-hash-if-needed>.<resource-type>`
+* `<resource-name>` - name of the resource. Name of MeshService or MeshExternalService
+* `<mesh-name>` - name of the mesh
+* `<extra-hash-if-needed>` - 16 characters of SHA256. In case of MeshService tags it's `SHA256(<tag1>=<value1>,<tag2>=<value2>)` sorted by keys.
+* `<resource-type>` - `meshservice` or `meshexternalservice`
```
lahabana, line 38: This sounds good but the order confuses me.

It should least to most specific or vice versa. Now it's a mixed bad.

docs/madr/decisions/053-sni.md
```diff
@@ -0,0 +1,57 @@
...
+### New format
+
+It's important to note that we never parse SNI. We only use it for multizone communication on zone proxies to associate to which service pass the data.
+It means that we can hash the data as long as it's consistent hashing.
+We cannot put tags in the original form because they can violate hostname requirements. For example `/` in `kuma.io/service` is not ok.
+The new format could look like this then:
+`<resource-name>.<mesh-name>.<extra-hash-if-needed>.<resource-type>`
+* `<resource-name>` - name of the resource. Name of MeshService or MeshExternalService
+* `<mesh-name>` - name of the mesh
+* `<extra-hash-if-needed>` - 16 characters of SHA256. In case of MeshService tags it's `SHA256(<tag1>=<value1>,<tag2>=<value2>)` sorted by keys.
```
lahabana, line 37: Should we fit a version inside to easily change this in the future (.e.g not convinced sha256 is the best hash here)?

docs/madr/decisions/053-sni.md
lahabana: Should we just use less characters for these? (ms, mes, mmzs)

docs/madr/decisions/053-sni.md
```diff
@@ -0,0 +1,57 @@
...
+* `<resource-type>` - `meshservice` or `meshexternalservice`
+
+Examples:
+* `backend.demo.meshservice` - MeshService named `backend` in `demo` mesh
+* `redis.kuma-demo.default.meshservice` - MeshService named `redis` in `kuma-demo` namespace in `default` mesh
+* `google.default.meshexternalservice` - MeshExternalService named `google` in `default` mesh
+* `backend.demo.e3b0c44298fc1c.meshservice` - subset of MeshService named `backend` in `demo` mesh for tags `version=v2`
+
+Resource name can be at most 255 chars long. Mesh should not be longer than 64 chars.
+If we go over the limit we compute SHA256 hash `<resource-name>.<mesh-name>.<extra-hash-if-needed>` and use 128 chars of it.
```
lahabana, line 47: I'm confused here. I think I need an example

docs/madr/decisions/053-sni.md
michaelbeaumont, line 37: what are the odds of collisions?

docs/madr/decisions/053-sni.md
jijiechen: I feel like it's better to put hash at the end, so readable information comes first. In some cases, maybe useful for matching.

docs/madr/decisions/053-sni.md
jijiechen: My understanding is that the hash is part of the SNI, it should always be used with other parts of this SNI. 
So the hash should never be used as the identify of this service. So hash collision is not a problem across multiple services.

docs/madr/decisions/053-sni.md
jakubdyszkiewicz: > This sounds good but the order confuses me.
> It should least to most specific or vice versa. Now it's a mixed bad.

Can you elaborate? I don't understand. Resource type sounds good but other place holder confuses you? What do you mean about specificity?

> Should we just use less characters for these? (ms, mes, mmzs)

We can

docs/madr/decisions/053-sni.md
jakubdyszkiewicz: The same odds as with clusters for splits that we create.

jakubdyszkiewicz:
```
const base62Charset = "0123456789abcdefghijklmnopqrstuvwxyz"

func base62Encode(input []byte) string {
	var result []byte
	num := new(big.Int).SetBytes(input)
	base := big.NewInt(int64(len(base62Charset)))

	for num.Sign() > 0 {
		mod := new(big.Int)
		num.DivMod(num, base, mod)
		result = append([]byte{base62Charset[mod.Int64()]}, result...)
	}

	return string(result)
}

func main() {
	h := sha256.New()
	bytes := h.Sum(nil)
	x := fmt.Sprintf("%x", bytes)
	base62 := base62Encode(bytes)
	fmt.Printf("%s %d %s %d", x, len(x), base62, len(base62))
}
```

```
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 64 5oaq0bjhj6un82wg98mgigso5q7qlhc63je4gw7ivixqqhkd3p 50
```

There is a little win with this at the cost of this custom implem (there seems no to be base62 lib using which we could specify charset. I'd lean towards sticking to hexadecimal, but if you think it's worth it. We can roll this implem.

docs/madr/decisions/053-sni.md
jakubdyszkiewicz: added

docs/madr/decisions/053-sni.md
jakubdyszkiewicz: changed resoruce-type to shorter versions

docs/madr/decisions/053-sni.md
michaelbeaumont: The source will use TLS to verify the identity of the destination anyway so there's no risk here.

lahabana: @lobkovilya used something from k8s I think for policies hash

michaelbeaumont: I added https://github.com/kumahq/kuma/blob/4ad47f000116872bf8b25676cd5ccc3cb115eed6/pkg/util/k8s/name_converter.go#L34 to copy how pod template hash is calculated, but its only 10 characters.

docs/madr/decisions/053-sni.md
```diff
@@ -0,0 +1,60 @@
...
+* `backend.demo.ms` - MeshService named `backend` in `demo` mesh
+* `redis.kuma-demo.default.ms` - MeshService named `redis` in `kuma-demo` namespace in `default` mesh
+* `google.default.mes` - MeshExternalService named `google` in `default` mesh
+* `backend.demo.e3b0c44298fc1c.ms` - subset of MeshService named `backend` in `demo` mesh for tags `version=v2`
+
+Resource name can be at most 255 chars long. Mesh should not be longer than 64 chars.
+If we go over the limit we compute SHA256 hash `<resource-name>.<mesh-name>.<extra-hash-if-needed>`.
+For example:
+* `super-long-name-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.default.e3b0c44298fc1c14.ms`
+  Changes to `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.ms`
```
lahabana, line 50: I think @lobkovilya for resource sync did something that kept the beginning of the name. I think it's useful for troubleshooting as it keeps most of the info available.

Something like:

`<resource-name>.<mesh-name>.<extra-hash-if-needed>.<resource-type>`

If resource-name is too long we switch to: `<resource-name>[:16]hash(<resource-name)[:16]`

Thinking about it I think it would be beneficial to just define this format with a character capacity for each element.

.e.g:

Format is `<format>.<resource-name>.<resourceHash>.<mesh-name>.<hash>.<resourceType>`

We have 256-5 (separators) = 251 chars available.

- format is 1 character (helps us with differnt versions).
- resourceNameHash is at 16 charts
- resourceName is at most 151 chars if it's larger it's truncated
- meshName is at most 64 chars
- hash is 16 chars
- resourceType is at most 3 chars

Otherwise we route just by hash and the resourceName is only an indicator.
Format is: `<format>.<serviceHash>.<routeHash>.<resourceType>.<resourceName>.<meshName>`

- format is 1 char (versioning)
- serviceHash 16 first chars of  `sha256(<resourceName>,<meshName>)`
- resourceType is at most 3 chars
- route hash 16 first chars of  `sha256(labels...)`
- resourceName is at most 64 chars (if too long it's truncated)
- meshName is at most 64 chars

This way jsut the format and serviceHash and resourceType is useful, the rest is for debugging no?

docs/madr/decisions/053-sni.md
jakubdyszkiewicz: > @lobkovilya used something from k8s I think for policies hash

It's `fnv.New64a()` + encoding from kube which is 
```
	// We omit vowels from the set of available characters to reduce the chances
	// of "bad words" being formed.
	alphanums = "bcdfghjklmnpqrstvwxz2456789"
```
fnv64 produces 16 chars in hex. Then some of them are exchanged to alphanums listed above.

We could use it for tags I think instead of SHA256 + trimming.

docs/madr/decisions/053-sni.md
jakubdyszkiewicz: that works I think. The only question is do we include serviceHash always or only if we go over char limit. I think including it always might be a bit redundant, no?

lobkovilya: Is there a benefit of having separate hashes for each component? Can we have a single hash at the end but separate limits:
`format(3-bytes).resource-name(N-bytes).mesh-name(64-bytes).resource-type(K-bytes).hash(resource-name, mesh-name, resource-type, tags)`?

slonka: > kept the beginning of the name. I think it's useful for troubleshooting as it keeps most of the info available.

➕

jakubdyszkiewicz: Yes, we can have a single hash. Again - the question is. Do we always add it or on demand?
```
v1.backend.demo.ms
```
is cleaner and has less data than
```
v1.backend.demo.e3b0c44298fc1c14.ms
```

btw. I'd put resource type at the end, it's safer to get valid domain. hash can technically produce only numbers which is not valid TLD.

Any opinions?

slonka: > Any opinions?

```
v1.backend.demo.e3b0c44298fc1c14.ms
```

visually I like

```
e3b0c44298fc1c14.v1.backend.demo.ms
```

more, because it looks more similar to `pkc-41231jofsda.us-east-2.aws-cloud` and other dynamically generated domains.

> Do we always add it or on demand?

I'd add it always - I value consistency and there won't be any confusion on "why does this have a hash and that does not?"

lobkovilya: I don't really have preferences between including or not including hash if components are not trimmed. I don't see pros or cons of either approach.

Though I'd avoid having format version as obvious as `v1.` as it might be easily assumed as version v1 of `backend`. Maybe version should be just a first character in the hash:
```
backend.demo.aXXXXXXXXXXXXXXXX.ms
```
26 lowercase letters + 10 digits should be enough to encode versions for many years ahead :)

lahabana: So what's the conclusion here?

jakubdyszkiewicz: updated. Let me know if that's good.

I put version "into" hash, but I moved this to the beginning because I don't want to parse everything to guess which format it is.

docs/madr/decisions/053-sni.md
```diff
@@ -31,23 +31,22 @@ It's important to note that we never parse SNI. We only use it for multizone com
...
-* `<mesh-name>` - (required) name of the mesh
-* `<extra-hash-if-needed>` - (optional) 16 characters of SHA256. In case of MeshService tags it's `SHA256(<tag1>=<value1>,<tag2>=<value2>)` sorted by keys.
-* `<resource-type>` - (required) `ms` (MeshService) or `mes` (MeshExternalService)
+`<format-version><hash>.<resource-name>.<mesh-name>.<resource-type>`
+* `<format-version>` - (required) `a`. 1 character. It's so we can evolve format in the future.
+* `<hash>` - (required) 16 characters of fnv64. It's `FNV64(resourceName,mesh,type,optional-data)`. Optional data can be for example tags.
+* `<resource-name>` - (required) name of the resource. Name of MeshService or MeshExternalService.
+  Trimmed to 128 chars. If trimmed we add `x` at the end, so we don't end up in invalid hostname `verylongname-.default.ms` 
+* `<mesh-name>` - (required) name of the mesh.
+  Trimmed to 64 chars. If trimmed we add `x` at the end.
```
lobkovilya, line 40: I think mesh name already limited to 64 chars since we're using `kuma.io/mesh` label to specify it?

docs/madr/decisions/053-sni.md
```diff
@@ -31,23 +31,22 @@ It's important to note that we never parse SNI. We only use it for multizone com
...
+* `<resource-type>` - (required) `ms` (MeshService) or `mes` (MeshExternalService). Abbreviation to at most 3 chars. 
 
-Examples:
-* `backend.demo.ms` - MeshService named `backend` in `demo` mesh
-* `redis.kuma-demo.default.ms` - MeshService named `redis` in `kuma-demo` namespace in `default` mesh
-* `google.default.mes` - MeshExternalService named `google` in `default` mesh
-* `backend.demo.e3b0c44298fc1c.ms` - subset of MeshService named `backend` in `demo` mesh for tags `version=v2`
+The proposed format has at most 1+16+129+65+3 = 214 chars.
```
lobkovilya, line 43: 129 and 65 here because of the `x` at the end? if so, then we also need to count dots no?

docs/madr/decisions/053-sni.md
jakubdyszkiewicz: I'm not sure we have strict limit when we create mesh, so I added it just in case.

docs/madr/decisions/053-sni.md
jakubdyszkiewicz: yes, updated

docs/madr/decisions/053-sni.md
```diff
@@ -30,23 +30,26 @@ Chosen option: "New format", because it solves existing problem and lets us diff
...
+  It's trimmed to 64 chars in total because DNS labels (subdomains) cannot exceed 64 chars. 
 * `<mesh-name>` - (required) name of the mesh.
-  Trimmed to 64 chars. If trimmed we add `x` at the end.
-* `<resource-type>` - (required) `ms` (MeshService) or `mes` (MeshExternalService). Abbreviation to at most 3 chars. 
+  Trimmed to 63 chars. If trimmed we add `x` at the end.
+* `<port>` - port of a service. If not relevant in the context of resource then `0`.
+* `<resource-type>` - (required) `ms` (MeshService) or `mes` (MeshExternalService). Abbreviation to at most 3 chars.
 
-The proposed format has at most 1+16+129+65+3+3(dots) = 217 chars.
+The proposed format has at most 1(format)+16(hash)+64(name)+64(mesh)+4(port)+3(type)+3(dots) = 155 chars.
```
lobkovilya, line 46: Shouldn't the port be 5? Also the number of dots is 4 now

michaelbeaumont (approved):

docs/madr/decisions/053-sni.md
```diff
@@ -0,0 +1,62 @@
...
+### New format
+
+It's important to note that we never parse SNI. We only use it for multizone communication on zone proxies to associate to which service pass the data.
+It means that we can hash the data as long as it's consistent hashing.
+We cannot put tags in the original form because they can violate hostname requirements. For example `/` in `kuma.io/service` is not ok.
+
+The new format could look like this then:
+`<format-version><hash>.<resource-name>.<port>.<mesh-name>.<resource-type>`
+* `<format-version>` - (required) `a`. 1 character. It's so we can evolve format in the future.
+* `<hash>` - (required) 16 characters of fnv64. It's `FNV64(resourceName,mesh,type,optional-data)`. Optional data can be for example tags.
```
lahabana, line 37: I'm uneasy about this in the end. Let's use sha256 especially now that we have versions in worst case we can change the version and change the hashing algorithm if we find this too slow...

lahabana (approved):
LGTM but I'm changing my mind on the hashing algo

docs/madr/decisions/053-sni.md
jakubdyszkiewicz: why? not enough chars for a hash = hash collisions?

PR #10402, opened 2024-06-06
chore(distribution): remove envoy version from info, allow custom envoy version 

lukidzi:
### Checklist before review

For some specific architectures, we could allow defining a custom version of Envoy. Added a change that will enable setting the version of envoy during distribution.

### DIFF
--- a/mk/build.mk
+++ b/mk/build.mk
@@ -2,8 +2,7 @@ build_info_fields = \
 	version=$(BUILD_INFO_VERSION) \
 	gitTag=$(GIT_TAG) \
 	gitCommit=$(GIT_COMMIT) \
-	buildDate=$(BUILD_DATE) \
-	Envoy=$(ENVOY_VERSION)
+	buildDate=$(BUILD_DATE)
 
 build_info_ld_flags := $(foreach entry,$(build_info_fields), -X github.com/kumahq/kuma/pkg/version.$(entry))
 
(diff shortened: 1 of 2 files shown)


slonka (approved):

mk/build.mk
```diff
@@ -2,8 +2,7 @@ build_info_fields = \
 	version=$(BUILD_INFO_VERSION) \
 	gitTag=$(GIT_TAG) \
 	gitCommit=$(GIT_COMMIT) \
-	buildDate=$(BUILD_DATE) \
-	Envoy=$(ENVOY_VERSION)
```
lukidzi, line 6: It is used in compatibility check. Do you think we need it?

lukidzi:
Still working on it, once distribution/test/build works, somehow test/e2e doesn't work 😕

tools/releases/version.sh
```diff
@@ -1,31 +1,5 @@
...
-  if [[ "${ENVOY_TAG}" =~ ^v[0-9]*\.[0-9]*\.[0-9]*$ ]]; then
-    echo "${ENVOY_VERSION}"
-  else
-    echo "${ENVOY_VERSION}-${ENVOY_TAG:0:8}"
-  fi
-  set +o errexit
-  set +o pipefail
-  set +o nounset
-}
-
```
slonka, line 28: are we sure we don't need any of this? the only thing that boggles my mind is pulling `https://raw.githubusercontent.com/envoyproxy/envoy/${ENVOY_TAG}/VERSION.txt` - it sometimes has `-dev` (on main) otherwise it always is just the tag 🤔 did we ever pull `-dev` versions?

slonka (approved):

tools/releases/version.sh
lukidzi: No, we never use dev envoy version

PR #10426, opened 2024-06-10
fix(kuma-cp): consistently update ZoneIngress available services

michaelbeaumont:
Previously this was happening during config generation, so ZoneIngresses without a connected proxy weren't being updated.

### Checklist prior to review

  - [x] manual
  - [x] e2e


pkg/zone/available_services.go
```diff
@@ -0,0 +1,166 @@
...
+func (t *ZoneAvailableServicesTracker) updateZoneIngresses(ctx context.Context) error {
+	meshes := core_mesh.MeshResourceList{}
+	if err := t.resManager.List(ctx, &meshes); err != nil {
+		return err
+	}
+
+	dps := core_mesh.DataplaneResourceList{}
+	if err := t.resManager.List(ctx, &dps); err != nil {
+		return err
+	}
```
jakubdyszkiewicz, line 129: This call concerns me when it comes to perf. What I read here is listing DPPs using uncached resource manager every second. What we could do here is

1) Use cached resource manager and bump default interval to maybe something like 5s
2) Rewrite this to event based approach but I don't think I'd bother since long term this will be gone
3) Use `MeshCache.GetMesh()`, this way we can leverage resources that were already fetched for the mesh. We get all the resources for free in O(1) assuming that there is at least one DPP connected of mesh X connected to this CP

michaelbeaumont: Went with option 3.

pkg/zone/available_services.go
```diff
@@ -0,0 +1,149 @@
...
+
+	availableServices := ingress.GetAvailableServices(
+		aggregatedMeshCtxs.AllDataplanes(),
+		aggregatedMeshCtxs.AllMeshGateways(),
+		t.getIngressExternalServices(aggregatedMeshCtxs),
+		t.ingressTagFilters,
+	)
+
+	for _, zi := range zis.Items {
+		// Initially zone is empty
```
jakubdyszkiewicz, line 136: where is it set then?

pkg/zone/available_services.go
```diff
@@ -0,0 +1,149 @@
...
+		if zi.Spec.Zone != "" && zi.Spec.Zone != t.zone {
+			continue
+		}
+		if availableServicesEqual(availableServices, zi.Spec.GetAvailableServices()) {
+			continue
+		}
+		zi.Spec.AvailableServices = availableServices
+		if err := t.resManager.Update(ctx, zi); err != nil {
+			t.logger.Error(err, "couldn't update ZoneIngress", "name", zi.GetMeta().GetName())
```
jakubdyszkiewicz, line 145: should we add something like "It will be updated on next reconciliation"? So when user read logs they can see error but then "update etc. etc."

pkg/zone/available_services.go
```diff
@@ -0,0 +1,149 @@
...
+
+	for _, zi := range zis.Items {
+		// Initially zone is empty
+		if zi.Spec.Zone != "" && zi.Spec.Zone != t.zone {
+			continue
+		}
+		if availableServicesEqual(availableServices, zi.Spec.GetAvailableServices()) {
+			continue
+		}
+		zi.Spec.AvailableServices = availableServices
```
jakubdyszkiewicz, line 143: I think I'd add here a log "updating zone ingress available services"

pkg/zone/available_services.go
```diff
@@ -0,0 +1,149 @@
...
+	return true
+}
+
+func (t *ZoneAvailableServicesTracker) getIngressExternalServices(
+	aggregatedMeshCtxs xds_context.AggregatedMeshContexts,
+) []*core_mesh.ExternalServiceResource {
+	var externalServices []*core_mesh.ExternalServiceResource
+
+	for _, mesh := range aggregatedMeshCtxs.Meshes {
+		if !mesh.ZoneEgressEnabled() {
```
jakubdyszkiewicz, line 101: I think it's worth to drop a comment why we skipping. I know why but it might not be obvious for the reader

pkg/zone/available_services.go
```diff
@@ -0,0 +1,149 @@
...
+		if availableServicesEqual(availableServices, zi.Spec.GetAvailableServices()) {
+			continue
+		}
+		zi.Spec.AvailableServices = availableServices
+		if err := t.resManager.Update(ctx, zi); err != nil {
+			t.logger.Error(err, "couldn't update ZoneIngress", "name", zi.GetMeta().GetName())
+		}
+	}
+	return nil
+}
```
jakubdyszkiewicz, line 149: Please add the test

pkg/zone/available_services.go
michaelbeaumont: I added one line that lists all the ingresses updated WDYT?

jakubdyszkiewicz: that's great. Just one nit, specify what was updated so something like `updated available services for ZoneIngresses`. Or alternatively change the logger name so it's more clear that it's a logger for available services

pkg/zone/available_services.go
michaelbeaumont: added

pkg/zone/available_services_test.go
```diff
@@ -0,0 +1,218 @@
...
+						Name: "builtin",
+						Type: "builtin",
+					},
+				},
+			},
+			Routing: &mesh_proto.Routing{
+				ZoneEgress: true,
+			},
+		}
+		Expect(resManager.Create(context.Background(), mesh, store.CreateByKey("mesh1", ""))).To(Succeed())
```
jakubdyszkiewicz, line 89: Can we use at least existing samples & builders?
```
samples.MeshMTLSBuilder().
  WithName("mesh1").
  WithEgressRoutingEnabled().
  Create(resManager)
```

same with Dataplane

pkg/zone/available_services_test.go
```diff
@@ -0,0 +1,218 @@
...
+		Expect(err).ToNot(HaveOccurred())
+
+		stop := make(chan struct{})
+		done = make(chan struct{})
+		go func() {
+			defer GinkgoRecover()
+			Expect(tracker.Start(stop)).To(Succeed())
+			close(done)
+		}()
+		defer close(stop)
```
jakubdyszkiewicz, line 180: nit: why do we not start this in Before like in any other component? Do we need to create those resources before we start the component?

jakubdyszkiewicz (approved):

test/e2e_env/multizone/connectivity/available_services.go
```diff
@@ -0,0 +1,116 @@
...
+	AfterEachFailure(func() {
+		DebugUniversal(multizone.Global, meshName)
+		DebugUniversal(statefulCluster, meshName)
+		DebugUniversal(multizone.UniZone1, meshName)
+	})
+
+	E2EAfterAll(func() {
+		Expect(multizone.UniZone1.DeleteMeshApps(meshName)).To(Succeed())
+		Expect(statefulCluster.DeleteMeshApps(meshName)).To(Succeed())
+		Expect(multizone.Global.DeleteMesh(meshName)).To(Succeed())
```
jakubdyszkiewicz, line 63: I think we also need to shutdown this new cluster that you created

jakubdyszkiewicz (approved):

PR #10428, opened 2024-06-11
feat(api-server): update policies api response structure

### LINKED ISSUE #9851: consider adding display-name/namespace labels to 'nested' policies in the HTTP API
### Description

We have display/name/namespace labels in the HTTP API for the policy listing and policy detail response, but for places where policies aren't the "top-level" resource such as affected policies, we don't have this information.

Therefore in the GUI we sometimes display the `display-name` as the "name of the thing" and sometimes we show the full `display-name.namespace` as the "name of the thing", which would be confusing.

Icarus9913:
Add `lables` map property for the kuma api-server `Meta` property which will improve 2 apis to get more details about the dataplanes and rules.

The following 2 http routes will be afftected:
* `/meshes/{mesh}/{policyType}/{policyName}/_resources/dataplanes`
* `/meshes/{mesh}/{resourceType}/{resourceName}/_rules`

Close #9851 

### Checklist prior to review

### DIFF
--- a/api/openapi/specs/common/resource.yaml
+++ b/api/openapi/specs/common/resource.yaml
@@ -41,7 +41,7 @@ components:
           type: boolean
     Meta:
       type: object
-      required: [type, mesh, name]
+      required: [type, mesh, name, labels]
       properties:
         type:
           type: string
@@ -55,14 +55,16 @@ components:
           type: string
           example: my-resource
           description: the name of the resource
-        displayName:
-          type: string
-          example: mtp
-          description: the original name of the resource
-        namespace:
-          type: string
-          example: kuma-system
-          description: the namespace of this resource
+        labels:
+          type: object
+          additionalProperties:
+            type: string
+          example:
+            "k8s.kuma.io/namespace": "kuma-system"
+            "kuma.io/display-name": "mtp"
+            "kuma.io/mesh": "default"
+            "kuma.io/origin": "zone"
+          description: the label of the resource
     InspectRule:
       type: object
       required: [type]
--- a/api/openapi/specs/common/resource.yaml
+++ b/api/openapi/specs/common/resource.yaml
@@ -55,6 +55,14 @@ components:
           type: string
           example: my-resource
           description: the name of the resource
+        displayName:
+          type: string
+          example: mtp
+          description: the original name of the resource
+        namespace:
+          type: string
+          example: kuma-system
+          description: the namespace of this resource
     InspectRule:
       type: object
       required: [type]
--- a/pkg/api-server/oapi-helpers/helpers.go
+++ b/pkg/api-server/oapi-helpers/helpers.go
@@ -12,9 +12,10 @@ func ResourceToMeta(m core_model.Resource) api_common.Meta {
 
 func ResourceMetaToMeta(resType core_model.ResourceType, m core_model.ResourceMeta) api_common.Meta {
 	return api_common.Meta{
-		Type: string(resType),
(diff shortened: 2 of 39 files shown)


Icarus9913:
Considerations:
 - Do these properties need to mark as `required` ?
 - What is the difference of `displayName` between `k8s` and `universal` mode ?
 - todo...

api/openapi/specs/common/resource.yaml
```diff
@@ -55,6 +55,14 @@ components:
           type: string
           example: my-resource
           description: the name of the resource
+        displayName:
```
lahabana, line 58: This should be `labels` of type object and be required

johncowen:
Draft PR for the GUI side of this: https://github.com/kumahq/kuma-gui/pull/2620

Icarus9913:
goldenfiles

slonka:
> goldenfiles

you need `/golden_files` I think

/golden_files

pkg/api-server/oapi-helpers/helpers.go
```diff
@@ -12,9 +12,10 @@ func ResourceToMeta(m core_model.Resource) api_common.Meta {
 
 func ResourceMetaToMeta(resType core_model.ResourceType, m core_model.ResourceMeta) api_common.Meta {
 	return api_common.Meta{
-		Type: string(resType),
-		Mesh: m.GetMesh(),
-		Name: m.GetName(),
+		Type:   string(resType),
+		Mesh:   m.GetMesh(),
+		Name:   m.GetName(),
+		Labels: m.GetLabels(),
```
lahabana, line 18: @johncowen  can confirm but I think we agreed on empty objects instead of nil when entries is empty

pkg/api-server/resource_endpoints_test.go
```diff
@@ -223,4 +224,71 @@ var _ = Describe("Resource Endpoints on Zone, label origin", func() {
...
+		respBytes, err := io.ReadAll(resp.Body)
+		Expect(err).ToNot(HaveOccurred())
+
+		// validate the response labels
+		var data api_types.InspectDataplanesForPolicyResponse
+		err = json.Unmarshal(respBytes, &data)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(data.Items).To(HaveLen(1))
+		Expect(data.Items[0].Labels).To(Equal(dpLabels), fmt.Sprintf("unmatched api-server policy dataplanes response label data: %+v", data.Items[0].Labels))
+	})
```
jakubdyszkiewicz, line 293: please take a look at other tests in this file or in other files. We try to follow the given when then convention, where in
* given - you prepare the data to test
* when - you execute the action that you are testing
* then - you check the result of the action

let me know if you need more clarification with this one

lahabana: Actually should you just add a test in: https://github.com/kumahq/kuma/blob/d023fb7bbc0fad6c0bf5b80b957b66b36b83b7b3/pkg/api-server/testdata/resources/inspect/dataplanes/_rules/overriding_meshtrace.input.yaml#L27

As long as you add labels in there they should appear. That's avoids starting a server etc

pkg/api-server/oapi-helpers/helpers.go
Icarus9913: Got it.

lahabana (approved):
LGTM. It wasn't necessary to add it to ALL test cases but it's fine anyway :) 👍

PR #10441, opened 2024-06-11
docs(MADR): mesh service identity sync

jakubdyszkiewicz:
### Checklist prior to review

MADR for sync of mesh service identity


michaelbeaumont (approved):

docs/madr/decisions/054-meshservice-identity-sync.md
```diff
@@ -0,0 +1,75 @@
...
+Let's assume this situation:
+* Service A is synced from zone X to global
+* Service A is synced from global to zone Y
+* Zone Y updates status with VIP and a hostname
+* There is update to identities in zone X
+* Without any extra logic, KDS on Zone Y side would receive a resource, replace the MeshService including previously computed status. 
+  It would wipe out VIP and a hostname which will result in interruption of traffic.
+
+A simpler solution would be to put this info in `spec`. This way we have a single writer of `spec` (source zone) and single writer of `status` (destination zone).
+We need a new KDS flag `PreserveStatusOnZone`, which would indicate that whenever a Zone receives a resource it should preserve the existing `status`.
```
lahabana, line 75: This implies we need to be careful in the MeshService reconciler to not lose these right?

docs/madr/decisions/054-meshservice-identity-sync.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+Currently, the only type is `ServiceTag` but we want this API to be extensible to future changes of identity.
+
+This is a list because one MeshService can select multiple Dataplane proxies of multiple different `kuma.io/service`, so for mTLS we need to require either one of them.
+
+We need a component that updates this value in `MeshService` object.
+The component would go over all Dataplane objects, match proxies for MeshService and add all `kuma.io/service` tags to this list.
+Additionally unrelated to this MADR, we have to introduce a component that is equivalent of `ServiceInsight` data (count of offline, online dpps etc.) in Service status.
+This data also requires all matched Dataplane objects therefore it can be the same component.
```
lahabana, line 55: Can we mention the future where these might just be correct spiffe ids?

docs/madr/decisions/054-meshservice-identity-sync.md
jakubdyszkiewicz: yes, but we are already quite careful in reconciler (I think current implem would not break this)

lobkovilya (approved):
Looks good. I agree that `spec` makes more sense than `status`

docs/madr/decisions/054-meshservice-identity-sync.md
jakubdyszkiewicz: added

lukidzi (approved):

lahabana (approved):

PR #10451, opened 2024-06-12
feat(meshservice): sync identity cross zones

jakubdyszkiewicz:
### Checklist prior to review

Sync mesh service identity over zones so we can pass cross zone traffic.


pkg/kds/context/context.go
```diff
@@ -225,8 +225,9 @@ func HashSuffixMapper(checkKDSFeature bool, labelsToUse ...string) reconcile.Res
 
 func UpdateResourceMeta(fs ...util.CloneResourceMetaOpt) reconcile.ResourceMapper {
 	return func(_ kds.Features, r core_model.Resource) (core_model.Resource, error) {
-		r.SetMeta(util.CloneResourceMeta(r.GetMeta(), fs...))
-		return r, nil
+		newRes := util.CloneResource(r)
+		newRes.SetMeta(util.CloneResourceMeta(r.GetMeta(), fs...))
```
michaelbeaumont, line 229: it always feels like a nit but this should really go in as a separate PR where it's visible

michaelbeaumont (approved):

pkg/kds/context/context.go
jakubdyszkiewicz: I'll put it in separate PR

Issue #10433, opened 2024-06-11
Add Zone and Namespace to hash when syncing resource from Global to Zone


jakubdyszkiewicz:
### What happened?

As I was writing E2E test I noticed that if I define MeshService call “backend” in Zone Uni1 I can access this using `hash.HashedName(meshName, "backend", "kuma-4")` in Global. However, to access this in Zone Uni2 I have to just do `hash.HashedName(meshName, "backend")`.

As far as I understand. That means that If we have two MeshServices called “backend” in two clusters, they both will be hashed to the same string.
This is only a problem if you have at least 3 zones, and two of them has a resource named in the same way. This is possible right now with policies applied on two Zone CPs.
I expect this to be a bigger problem with MeshService and policies on namespaces.

I think the problem is that here https://github.com/kumahq/kuma/blob/master/pkg/kds/context/context.go#L91 we don’t use ZoneTag and NamespaceTag as well.

PR #10452, opened 2024-06-13
fix(kds): fix resource name hashing on global

Automaat:
Fixes: #10433

### Checklist prior to review

### DIFF
--- a/pkg/kds/context/context.go
+++ b/pkg/kds/context/context.go
@@ -88,7 +88,7 @@ func DefaultContext(
 				// particular format to be able to reference them
 				reconcile.Not(reconcile.TypeIs(system.SecretType)),
 			),
-			HashSuffixMapper(true)),
+			HashSuffixMapper(true, mesh_proto.ZoneTag, mesh_proto.KubeNamespaceTag)),
 	}
 
 	zoneMappers := []reconcile.ResourceMapper{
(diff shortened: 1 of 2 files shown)


pkg/kds/context/context.go
```diff
@@ -88,7 +88,7 @@ func DefaultContext(
 				// particular format to be able to reference them
 				reconcile.Not(reconcile.TypeIs(system.SecretType)),
 			),
-			HashSuffixMapper(true)),
+			HashSuffixMapper(true, mesh_proto.ZoneTag, mesh_proto.KubeNamespaceTag)),
```
slonka, line 91: Is this in any way a breaking change?

slonka (approved):
Looks like it does what the issue describing it says so 👍 - only Q about backwards compatiblity

pkg/kds/context/context.go
Automaat: I am in the middle of testing backward compatiblity

lobkovilya:
Could you please add a test [here](https://github.com/kumahq/kuma/blob/e4c9f4da4360f10360e98be19beea6346d6a715f/pkg/kds/global/components_test.go) or [here](https://github.com/kumahq/kuma/blob/e4c9f4da4360f10360e98be19beea6346d6a715f/pkg/kds/zone/components_test.go) to check that MeshServices from 2 zones are synced to the 3rd zone without collisions?

lukidzi (approved):

lobkovilya (approved):
LGTM, please check the upgrade manually especially given the fact we disabled kuma_helm_upgrade_multizone test.

PR #10455, opened 2024-06-13
ci(policies): include labels and status in openapi spec

lahabana:
We were not handling these new field correctly

Still missing syncing resources to the general openapispec

### Checklist prior to review


tools/policy-gen/templates/schema.yaml
```diff
@@ -11,4 +11,7 @@ properties:
   name:
     description: 'Name of the Kuma resource'
     type: string
+  labels:
```
lobkovilya, line 14: it's probably a good idea to add
```yaml
additionalProperties:
  type: string
```
to limit key/value of the object to be `string`. Also, that's how k8s defines `labels` https://github.com/kubernetes/kubernetes/blob/master/api/openapi-spec/v3/apis__policy__v1_openapi.json#L876-L883

lukidzi (approved):

PR #10457, opened 2024-06-13
feat(kuma-cp): always add kuma.io/zone label to resource

Automaat:
This change is needed to simplify targeting new MeshService in policies. It will simplify how we build subset from MeshService: https://github.com/kumahq/kuma/pull/10422/files#diff-aa3f3b4f30ce7a99336aefc7674797950e7e9c608e6a1c04699452820a8c5041R216

### Checklist prior to review


lobkovilya:
We also need a validation (and unfortunately I missed that for `kuma.io/policy-role` as well), it shouldn't be possible for user to set arbitrary values for `kuma.io/zone` and `kuma.io/policy-role`

Automaat:
> We also need a validation (and unfortunately I missed that for `kuma.io/policy-role` as well), it shouldn't be possible for user to set arbitrary values for `kuma.io/zone` and `kuma.io/policy-role`

I've added validation

slonka (approved):
LGTM

Automaat:
I had to remove k8s validation. On Kubernetes mutating webhook runs before validating, so we were adding labels and then checking if they are absent

lobkovilya (changes requested):
Please check how we implemented `kuma.io/origin` validation. Because mutating webhook runs before the validating we have to call validation inside the mutating webhook https://github.com/kumahq/kuma/blob/d89e3e53960da72f261229f699b22c3e2d3273e8/pkg/plugins/runtime/k8s/webhooks/defaulter.go#L70

Automaat:
@lobkovilya I've added validation as you suggested

pkg/api-server/resource_endpoints.go
```diff
@@ -532,6 +532,13 @@ func (r *resourceEndpoints) validateLabels(rm model.ResourceMeta) validators.Val
 		}
 	}
 
+	if _, ok := rm.GetLabels()[mesh_proto.PolicyRoleLabel]; ok {
+		err.AddViolationAt(validators.Root().Key(mesh_proto.PolicyRoleLabel), fmt.Sprintf("%s label should not be set manually", mesh_proto.PolicyRoleLabel))
```
lobkovilya, line 536: should be the same as on k8s, we allow these labels to be set, but validate the value to be correct

Automaat: right, I've forgot to update it after fixing k8s validation

pkg/api-server/resource_endpoints.go
```diff
@@ -532,17 +532,54 @@ func (r *resourceEndpoints) validateLabels(rm model.ResourceMeta) validators.Val
...
 
-	for _, k := range maps.SortedKeys(rm.GetLabels()) {
+	if r.mode != config_core.Global {
+		zoneTag, ok := resource.GetMeta().GetLabels()[mesh_proto.ZoneTag]
+		if ok && zoneTag != r.zoneName {
+			err.AddViolationAt(validators.Root().Key(mesh_proto.ZoneTag), fmt.Sprintf("%s label should have %s value", mesh_proto.ZoneTag, r.zoneName))
+		}
+	}
+
+	r.validatePolicyRole(resource)
```
lobkovilya, line 542: please validate `r.descriptor` to have `.IsPolicy` and `.IsPluginOriginated`, you can avoid handling cast errors in this case

pkg/api-server/resource_endpoints.go
```diff
@@ -532,17 +532,54 @@ func (r *resourceEndpoints) validateLabels(rm model.ResourceMeta) validators.Val
...
+		for _, msg := range validation.IsValidLabelValue(resource.GetMeta().GetLabels()[k]) {
 			err.AddViolationAt(validators.Root().Key(k), msg)
 		}
 	}
 	return err
 }
 
+func (r *resourceEndpoints) validatePolicyRole(resource rest.Resource) validators.ValidationError {
+	var err validators.ValidationError
+	ns, ok := resource.GetMeta().GetLabels()[mesh_proto.KubeNamespaceTag]
```
lobkovilya, line 557: the code runs only on Universal no? why do we expect `k8s.kuma.io/namespace`?

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -78,6 +83,42 @@ func (c *ResourceAdmissionChecker) isPrivilegedUser(allowedUsers []string, userI
...
+	return c.validatePolicyRole(r)
+}
+
+func (c *ResourceAdmissionChecker) validatePolicyRole(r core_model.Resource) *admission.Response {
+	ns, ok := r.GetMeta().GetLabels()[mesh_proto.KubeNamespaceTag]
+	if !ok {
+		return nil
+	}
+	policy, ok := r.GetSpec().(core_model.Policy)
+	if !ok {
```
lobkovilya, line 103: we can check r.Descriptor somewhere earlier in the code, to avoid handling cast errors

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -41,6 +42,10 @@ func (c *ResourceAdmissionChecker) IsOperationAllowed(userInfo authenticationv1.
 		return c.resourceIsNotAllowedResponse()
 	}
 
+	if errResponse := c.validateLabels(r); errResponse != nil {
```
lobkovilya, line 45: The method `IsOperationAllowed` calls `isResourceTypeAllowed` to check if "TypeMeta" (what we call Descriptor) is allowed in the environment, then it calls `isResourceAllowed` to check that ResourceMeta+ResourceSpec are allowed in the environment. Your call to `validateLabels` should be moved inside `isResourceAllowed` as it checks ResourceMeta

pkg/plugins/runtime/k8s/webhooks/testdata/validation/cli-user_create_dpp.input.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
   creationTimestamp: null
   annotations:
     k8s.kuma.io/synced: "true"
+  labels:
```
lobkovilya, line 11: why did you add `kuma.io/zone` label here? does it work without the zone label?

Automaat: it does not work, without zone label now

pkg/api-server/resource_endpoints.go
Automaat: added

pkg/api-server/resource_endpoints.go
Automaat: right, removed

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
Automaat: moved

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
Automaat: changed

pkg/plugins/runtime/k8s/webhooks/testdata/validation/cli-user_create_dpp.input.yaml
lobkovilya: why? it's a breaking change then no? Can't we just add the `kuma.io/zone` label if it's missing?

pkg/api-server/resource_endpoints.go
```diff
@@ -532,17 +532,45 @@ func (r *resourceEndpoints) validateLabels(rm model.ResourceMeta) validators.Val
...
 		}
 	}
 	return err
 }
 
+func (r *resourceEndpoints) validatePolicyRole(resource rest.Resource) validators.ValidationError {
+	var err validators.ValidationError
+	policyObject, ok := resource.GetSpec().(core_model.Policy)
+	if !ok {
+		return err
```
lobkovilya, line 561: you're returning an empty error here, I think it's easier to not handle cast because you're checking the descriptor

pkg/api-server/resource_endpoints.go
```diff
@@ -532,17 +532,45 @@ func (r *resourceEndpoints) validateLabels(rm model.ResourceMeta) validators.Val
...
+	}
+	policyRole, ok := resource.GetMeta().GetLabels()[mesh_proto.PolicyRoleLabel]
+	if !ok {
+		return err
```
lobkovilya, line 565: same, you're returning an empty error, just use [PolicyRole](https://github.com/kumahq/kuma/blob/d89e3e53960da72f261229f699b22c3e2d3273e8/pkg/core/resources/model/resource.go#L493) it always returns a value

Automaat: I am returning empty error since missing policy role is not a problem here. I cannot use `PolicyRole` method here, as this the resource types are not compatible. This is a `rest.Resource`

pkg/api-server/resource_endpoints.go
Automaat: changed

pkg/plugins/runtime/k8s/webhooks/testdata/validation/cli-user_create_dpp.input.yaml
Automaat: I've check again and it is not needed, I don't remember why I've added it

pkg/api-server/resource_endpoints.go
lobkovilya: Just replace `PolicyRole` method with:
```go
func PolicyRole(rm ResourceMeta) mesh_proto.PolicyRole {
	if rm.GetLabels() == nil || rm.GetLabels()[mesh_proto.PolicyRoleLabel] == "" {
		return mesh_proto.SystemPolicyRole
	}
	return mesh_proto.PolicyRole(rm.GetLabels()[mesh_proto.PolicyRoleLabel])
}
```
it doesn't actually need the whole Resource

Automaat: changed

pkg/plugins/runtime/k8s/webhooks/testdata/validation/cli-user_create_zone-ns-mt.input.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
   labels:
     kuma.io/origin: zone
     kuma.io/mesh: default
+    kuma.io/zone: zone-1
```
lobkovilya, line 9: Should we remove these in a couple of places? Maybe only in old test cases. I think it's important to check `kuma.io/zone` label is still optional

lobkovilya (approved):
LGTM, just a small comment regarding tests

pkg/plugins/runtime/k8s/webhooks/testdata/validation/cli-user_create_zone-ns-mt.input.yaml
Automaat: done

PR #10478, opened 2024-06-14
docs(MADR): default HostnameGenerators

jakubdyszkiewicz:
### Checklist prior to review


docs/madr/decisions/055-default-hostname-generators.md
```diff
@@ -0,0 +1,218 @@
...
+#### Local Kubernetes MeshService
+
+It's rather unlikely for user to use custom hostnames to consume MeshService that is local in the Kubernetes cluster.
+Therefore, no default HostnameGenerator is needed for it.
+
+#### Synced Kubernetes MeshService
+
+```yaml
+spec:
+  template: '{{ label "k8s.kuma.io/service-name" }}.{{ label "k8s.kuma.io/namespace" }}.svc.mesh.{{ label "kuma.io/zone" }}'
```
michaelbeaumont, line 58: Why don't we want `.namespace`?

jakubdyszkiewicz: why do we need namespace? I think it's easier to just rely on existing mechanism, no?

michaelbeaumont: Easier for who though? For UX, readability, optimizing for most common use cases I think `.Namespace` is better. Labels are completely arbitrary and so using a label to get a namespace, which is guaranteed to be there for a Kubernetes service, is excessively noisy. It adds friction between Kuma and Kubernetes. IMO if namespace doesn't deserve its own field then nothing deserves its own field.

michaelbeaumont: Yeah they would all be fields then IMO.

> But for synced resources Namespace is system namespace. Isn't it more clear that label is actually the original namespace?

I highly suspect the intuition is to think "the namespace" of a synced service as its original namespace, rather than the object with a hashed name in `kuma-system`. This synced object is essentially just Kuma internal state. As evidenced by the unusable name.

Unfortunately, I realize this kind of clashes with the decision on `targetRef.namespace`, which doesn't refer to the original namespace for synced services, and you have to use labels. It's really kind of a different scenario because the targetRef semantics are not under our control. I suppose it's arguably a valid reason to use labels here too. But it'd be unfortunate IMO.

docs/madr/decisions/055-default-hostname-generators.md
```diff
@@ -0,0 +1,218 @@
...
+* `backend.west.mesh.west`
+* `httpbin.meshext.local`
+* `google.meshext.global`
+* `auth.meshmz.global`
+
+Advantages:
+* It's familiar to Kubernetes pattern of `svc.cluster.local`
+
+Disadvantages:
+* `{{ zone }}` might violate TLD. Although we can validate zone name to prevent this.
```
slonka, line 197: do we validate zone name in any way? would this be painful to migrate?

slonka (commented):
Good to put all of this together. I like the fact that now I don't have to care about if HostnameGenerator/MeshExternalService is synced from global or is zone originated and the template will just work. I agree that we should start with documenting and then create by default. For the suffix I think it largely depends on how hard would it be to migrate from z non tld valid zone name to a valid one, if hard then I'd go with `.{{ zone }}.{{ type }}`.

docs/madr/decisions/055-default-hostname-generators.md
jakubdyszkiewicz:
```
func (r *ZoneConfig) Validate() error {
	if r.Name == "" {
		return errors.Errorf("Name is mandatory")
	}
	if !govalidator.IsDNSName(r.Name) {
		return errors.Errorf("Zone name %s has to be a valid DNS name", r.Name)
	}
	if len(r.Name) > 63 {
		return errors.New("Zone name cannot be longer than 63 characters")
	}
```
We do, although I'm not sure yet if `govalidator.IsDNSName` == `valid TLD`

docs/madr/decisions/055-default-hostname-generators.md
```diff
@@ -0,0 +1,218 @@
...
+
+This HostnameGenerator could be automatically created by Global CP so that it's synced down to every Zone CP.
+
+### MeshExternalService
+
+#### MeshExternalService applied on Zone CP
+
+```yaml
+spec:
+  template: '{{ .DisplayName }}.svc.meshext.local'
```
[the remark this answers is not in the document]
jakubdyszkiewicz: It's an interesting idea, but IMHO not worth it. I'm happy to hear other opinions.

docs/madr/decisions/055-default-hostname-generators.md
slonka: It's just a regex 

```
DNSName           string = `^([a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62}){1}(\.[a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62})*[\._]?$
```` 

so I wouldn't say so

File: docs/madr/decisions/055-default-hostname-generators.md (hunk 1 above)
Comment (jakubdyszkiewicz) [reply docs/madr/decisions/055-default-hostname-generators.md:58]: We did namespace.

File: docs/madr/decisions/055-default-hostname-generators.md [hunk 4]
```diff
@@ -0,0 +1,218 @@
...
+  namespace: kuma-system
+```
+
+Would be `httpbin.svc.meshext.local`
+
+#### MeshExternalService applied on Global CP
+
+```yaml
+spec:
+  template: '{{ .DisplayName }}.svc.meshext.global'
```
jakubdyszkiewicz, line 132: other alternative would be
```
{{ .DisplayName }}.extsvc.mesh.global
```
and
```
'{{ .DisplayName }}.mzsvc.mesh.global'
```
for meshmultizoneservice

docs/madr/decisions/055-default-hostname-generators.md
```diff
@@ -0,0 +1,218 @@
...
+
+### MeshService
+
+#### Local Kubernetes MeshService
+
+It's rather unlikely for user to use custom hostnames to consume MeshService that is local in the Kubernetes cluster.
+Therefore, no default HostnameGenerator is needed for it.
+
+#### Synced Kubernetes MeshService
+
```
michaelbeaumont, line 55: Wondering whether we need a separate generator for headless services here. Since the name for headless is sort of useless

jakubdyszkiewicz: We do, let me update MADR

lahabana: Random question, how do we manage names of MES for headless svcs? How do we ensure adding the suffix with the pod offset doesn't lead to too long names?

docs/madr/decisions/055-default-hostname-generators.md
```diff
@@ -0,0 +1,218 @@
...
+* Create
+
+Suffix:
+* `.svc.{{ type }}.{{ zone }}`
+* `.{{ zone }}.{{ type }}`
+
+## Decision Outcome
+
+Template and matching - Proposed options.
+Shipping - short term document. Once we get all prerequisites and MeshService/MeshExternalService/MeshMultizoneService goes GA, we create by default.
```
lahabana, line 26: When we say shipping is this just resources added to the helm chart?
If yes:
- what's the story for universal?
If no:
- if it's code generated we must make sure people can start in prod with nothing and define everything with IaaC

jakubdyszkiewicz: I think it should be the same mechanism as for creating any other "global-scoped" resources so:
* Default Mesh
* Default RBAC in Kong Mesh

so not in HELM but in the code with option to opt-out

docs/madr/decisions/055-default-hostname-generators.md
jakubdyszkiewicz: > MES for headless svcs?

MeshExternalService for Headless svcs?

> How do we ensure adding the suffix with the pod offset doesn't lead to too long names

Do you mean to not go over 63 char of DNS label (=subdomain)?

michaelbeaumont: https://github.com/kumahq/kuma/blob/16c7cb0c137ebc084d960b81e4d8bce21588a359/pkg/plugins/runtime/k8s/controllers/meshservice_controller.go#L206

docs/madr/decisions/055-default-hostname-generators.md
slonka: I like these two more 👍

docs/madr/decisions/055-default-hostname-generators.md
```diff
@@ -0,0 +1,234 @@
...
+Suffix - ?
```
slonka, line 27: we need decision on this, right?

docs/madr/decisions/055-default-hostname-generators.md
slonka: how hard is renaming a zone? is it just chaning `KUMA_MULTIZONE_ZONE_NAME`?

slonka (approved):
Generally I don't think anything is missing from here. We just need to come up with the suffix. I personally would go with the more similar one `.svc.{{ type }}.{{ zone }}` but both options are fine.

docs/madr/decisions/055-default-hostname-generators.md
lahabana: So we truncate ok. Is this something k8s does as well? I'm just scared of creating new issues (This is actually parallel to this MADR and it might be better to track as its own issue)

docs/madr/decisions/055-default-hostname-generators.md
lahabana: OK

We need to have a better way to start empty and have customers create these in the future but that's parallel to this MADR

docs/madr/decisions/055-default-hostname-generators.md
michaelbeaumont: Yeah: https://github.com/kubernetes/kubernetes/blob/29e4f5a893bf47c608aa8593ae6e67d20b0fd775/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L49

docs/madr/decisions/055-default-hostname-generators.md
jakubdyszkiewicz: In theory yes, but in practice I suspect you can get some turbulence, because for example DPPs has `kuma.io/zone` tag as well as some policies `kuma.io/zone` label

docs/madr/decisions/055-default-hostname-generators.md
```diff
@@ -0,0 +1,234 @@
...
+
+### `.svc.{{ type }}.{{ zone }}`
+
+Types:
+* `mesh` for MeshService
+* `meshext` - MeshExternalService
+* `meshmz` - MeshMultizoneService
+
+Examples:
+* `redis.kuma-demo.mesh.east`
```
lukidzi, line 203:
```suggestion
* `redis.kuma-demo.svc.mesh.east`
```

docs/madr/decisions/055-default-hostname-generators.md
slonka: oh, that's right, so let's go with the other one - `.{{ zone }}.{{ type }}`

lukidzi (approved):

docs/madr/decisions/055-default-hostname-generators.md
jakubdyszkiewicz: talked offline. I updated MADR

michaelbeaumont (approved):

Issue #10463, opened 2024-06-13
KDS updates stuck after webhook rejects the changes with NACK


lobkovilya:
### What happened?

Visible as an e2e test failure:

```
• [FAILED] [148.788 seconds]
Upgrade Multizone with Helm upgrade helm multizone [It] from version: 2.6.7 [job-2]
github.com/kumahq/kuma/test/framework/ginkgo.go:105

  [FAILED] Timed out after 30.000s.
  meshtimeouts are not synced to zone
  Expected
      <int>: 1
  to equal
      <int>: 3
Error: It 06/12/24 20:09:49.466
```

It's possible to reproduce locally after running the test in the loop for a few minutes.

  
[zone.txt](https://github.com/user-attachments/files/15822695/zone.txt)
[global.txt](https://github.com/user-attachments/files/15822696/global.txt)

lukidzi:
Action items:
* Don’t set owner ref on remote resources
* Distinguish validation errors and other errors and only return NACK on validation error
* In other cases restart KDS sync (crash the component)

PR #10480, opened 2024-06-14
fix(kds): send NACK only when resource is invalid and do not retry

lukidzi:
### Checklist prior to review

Based on the investigation we noticed that our force retrier might cause some issues when connected with the sleep. We retried once the resource was NACKed, but it's incorrect xDS behavior. If the client returns NACK it means that the resource is incorrect and shouldn't be synced until is fixed. Now instead of retrying NACK we decided to distinguish validation errors from storage errors. Once there is a storage error we restart a whole component and on the initial stream sync all the resources, but for the validation error, we wait until the resource is fixed, and once it's changed we try to apply it again.

* Do not set owner reference for synced resource has been already done https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/webhooks/owner_reference_mutator.go#L88 and should be backported to 2.6 and 2.7(probably)
* remove force retrier and add schema validation
* added test that validate that once resource is fixed, it's synced to the client 
* restart component on the db error, NACK on validation

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/10463


lukidzi:
Should be backported to 2.6 and 2.7

pkg/kds/zone/components_test.go
```diff
@@ -303,6 +303,48 @@ var _ = Describe("Zone Sync", func() {
...
+			err := globalStore.Create(context.Background(), &mesh.MeshResource{Spec: invalidMesh}, store.CreateByKey("mesh-1", model.NoMesh))
+			Expect(err).ToNot(HaveOccurred())
+
+			// should not be synchronized
+			Consistently(func(g Gomega) {
+				actual := mesh.MeshResourceList{}
+				err := zoneStore.List(context.Background(), &actual)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(actual.Items).To(BeEmpty())
+			}, "5s", "100ms").Should(Succeed())
```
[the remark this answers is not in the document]
lukidzi: Currently, we are unable to, we would need to rework the unit test setup to add stats callback.

pkg/kds/v2/client/kds_client.go
```diff
@@ -75,6 +87,16 @@ func (s *kdsSyncClient) Receive() error {
 			}
 			return errors.Wrap(err, "failed to receive a discovery response")
 		}
+
+		if err := received.Validate(); err != nil {
+			if err := s.kdsStream.NACK(received.Type, err); err != nil {
```
lobkovilya, line 92: please restore the log `s.log.Info("error during callback received, sending NACK", "err", err)`, it's useful for debugging

lobkovilya: ah but the text should be different

pkg/kds/zone/components_test.go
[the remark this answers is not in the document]
lobkovilya: I see, given the fact we reconcile the snapshot every 100ms in unit tests, probably 1s will be enough

pkg/kds/global/components_test.go
```diff
@@ -70,6 +71,38 @@ var _ = Describe("Global Sync", func() {
...
+							mesh_proto.ServiceTag:  "web",
+							mesh_proto.ProtocolTag: "http",
+						},
+					},
+				},
+			},
+		}
+		err := zoneStores[0].Create(context.Background(), &mesh.DataplaneResource{Spec: dp}, store.CreateByKey("dp-1", "mesh-1"))
+		Expect(err).ToNot(HaveOccurred())
+		Eventually(func() int {
```
jakubdyszkiewicz, line 97: Does Eventually make sense in this context? Shouldn't this be Consistentnly?

pkg/kds/v2/client/kds_client.go
```diff
@@ -75,6 +87,17 @@ func (s *kdsSyncClient) Receive() error {
...
+			s.log.Info("received resource is invalid, sending NACK", "err", err)
+			if err := s.kdsStream.NACK(received.Type, err); err != nil {
+				if err == io.EOF {
+					return nil
+				}
+				return errors.Wrap(err, "failed to NACK a discovery response")
+			}
+			continue
+		}
 		s.log.V(1).Info("DeltaDiscoveryResponse received", "response", received)
```
jakubdyszkiewicz, line 101: I think this log should go above the change with validation that you added

pkg/kds/global/components_test.go
```diff
@@ -70,6 +71,38 @@ var _ = Describe("Global Sync", func() {
...
+			actual := mesh.DataplaneResourceList{}
+			err := globalStore.List(context.Background(), &actual)
+			Expect(err).ToNot(HaveOccurred())
```
jakubdyszkiewicz, line 100: nit: I'd inject Gomega to eventually if you use Expect. Just like you did for components_test

jakubdyszkiewicz (approved):
a couple of nits, but looks good for me 👍

Issue #10534, opened 2024-06-18
Multizone Headless service + MeshService E2E test


jakubdyszkiewicz:
### Description

We need to add E2E test for MeshService + Headless Service

PR #10535, opened 2024-06-18
fix(meshservice): tags and selector

jakubdyszkiewicz:
### Checklist prior to review

`k8s.kuma.io/namespace` was missing from the selector. We have the following option
1) Add this using controller
2) Add this using webhook.
3) Infer when retrieving from the store
4) Handle this every time when using selecotr

Since it's not really something users would apply manually on kube, I think adding this in controller is fine. 4) option is error prone.

Additionally, DataplaneRef does not contain a namespace. I don't think it's even handleded anywhere. Added an issue for e2e test https://github.com/kumahq/kuma/issues/10534


michaelbeaumont:
> Additionally, DataplaneRef does not contain a namespace. I don't think it's even handleded anywhere

It shouldn't though, the mistake is that we didn't convert the endpoint i.e. Pod name to name.namespace

michaelbeaumont (approved):

Issue #10560, opened 2024-06-19
MeshExternalService: cluster is marked as TCP when there are two endpoints.


slonka:
### What happened?

e.g.

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshExternalService
metadata:
  name: mes-http
  namespace: kong-mesh-system
  labels:
    kuma.io/mesh: default
spec:
  match:
    type: HostnameGenerator
    port: 80
    protocol: http
  endpoints:
    - address: httpbin.org
      port: 80
    - address: 1.1.1.1
      port: 443
```

Issue #10559, opened 2024-06-19
MeshExternalService through egress should have HTTP conn manager and HTTP RBAC for HTTP matches


slonka:
### What happened?

Now it has TCP proxy and network rbac.

PR #10568, opened 2024-06-19
fix(meshexternalservice): correctly set tls context and protocol

lukidzi:
### Checklist prior to review

* fix setting tls context for MeshExternalService
* take the correct protocol for MeshExternalService
* do not set tls context when tls disabled
* use correct way of checking protocol for egress

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/10560 fix: https://github.com/kumahq/kuma/issues/10559

### DIFF
--- a/pkg/xds/context/mesh_context_builder.go
+++ b/pkg/xds/context/mesh_context_builder.go
@@ -553,6 +553,9 @@ func inferServiceProtocol(endpoints []xds.Endpoint) core_mesh.Protocol {
 	}
 	for _, endpoint := range endpoints[1:] {
 		endpointProtocol := core_mesh.ParseProtocol(endpoint.Tags[mesh_proto.ProtocolTag])
+		if endpoints[0].ExternalService != nil && endpoints[0].ExternalService.Protocol != "" {
+			endpointProtocol = endpoints[0].ExternalService.Protocol
+		}
 		serviceProtocol = util_protocol.GetCommonProtocol(serviceProtocol, endpointProtocol)
 	}
 	return serviceProtocol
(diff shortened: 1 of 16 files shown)


pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/zone.log
slonka, line 1: not needed probably

pkg/xds/context/mesh_context_builder.go
```diff
@@ -553,6 +553,9 @@ func inferServiceProtocol(endpoints []xds.Endpoint) core_mesh.Protocol {
 	}
 	for _, endpoint := range endpoints[1:] {
 		endpointProtocol := core_mesh.ParseProtocol(endpoint.Tags[mesh_proto.ProtocolTag])
+		if endpoints[0].ExternalService != nil && endpoints[0].ExternalService.Protocol != "" {
+			endpointProtocol = endpoints[0].ExternalService.Protocol
+		}
```
slonka, line 558: It doesn't make sense to loop if we're always checking index [0].

```suggestion
```

shouldn't we just return the following in line 552

```
return endpoints[0].ExternalService.Protocol
```

?

why does it make sense to first assign `serviceProtocol = endpoints[0].ExternalService.Protocol` and then assign `endpointProtocol = endpoints[0].ExternalService.Protocol` (which is the same) and do `GetCommonProtocol` on it?

slonka (changes requested):
In general ok, one thing needs changing IMO

pkg/xds/context/mesh_context_builder.go
lukidzi: My bad it supposed to be `endpoint.ExternalService != nil && endpoint.ExternalService.Protocol`

slonka (approved):

PR #10595, opened 2024-06-21
fix(meshpassthrough): return 503 when no route found

lukidzi:
### Checklist prior to review

When MeshPassthrough doesn't expose traffic, but the traffic is HTTP and there is other HTTP traffic exposed, we used to return a 404 error indicating that there is no route. I have changed this to return a 503 error when there is no route because the traffic is forbidden.

> Changelog: feat(meshpassthrough): implement new policy

### DIFF
--- a/pkg/plugins/policies/meshpassthrough/plugin/xds/listeners_filter_chain.go
+++ b/pkg/plugins/policies/meshpassthrough/plugin/xds/listeners_filter_chain.go
@@ -2,6 +2,7 @@ package xds
 
 import (
 	"fmt"
+	"net/http"
 
 	envoy_listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 
@@ -12,6 +13,8 @@ import (
 	xds_routes "github.com/kumahq/kuma/pkg/xds/envoy/routes"
 	xds_virtual_hosts "github.com/kumahq/kuma/pkg/xds/envoy/virtualhosts"
 )
+
+const NoMatchMsg = "This is a Kuma Sidecar. No routes match this Domain!\n"
 
 type FilterChainConfigurer struct {
 	APIVersion core_xds.APIVersion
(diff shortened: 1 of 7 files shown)


pkg/plugins/policies/meshpassthrough/plugin/xds/listeners_filter_chain.go
```diff
@@ -13,6 +14,8 @@ import (
 	xds_virtual_hosts "github.com/kumahq/kuma/pkg/xds/envoy/virtualhosts"
 )
 
+const NoMatchMsg = "This is a Kuma Sidecar. No routes match this Domain!\n"
```
slonka, line 17: I'm not sure about this message. Maybe @lahabana or @jakubdyszkiewicz could give some input here.

pkg/xds/envoy/virtualhosts/route_configurer.go
```diff
@@ -77,3 +78,29 @@ func (c VirtualHostBasicRouteConfigurer) Configure(virtualHost *envoy_config_rou
...
+
+type VirtualHostDirectResponseRouteConfigurer struct {
+	status      uint32
+	responseMsg string
+}
+
+func (c VirtualHostDirectResponseRouteConfigurer) Configure(virtualHost *envoy_config_route_v3.VirtualHost) error {
+	virtualHost.Routes = append(virtualHost.Routes, &envoy_config_route_v3.Route{
+		Match: &envoy_config_route_v3.RouteMatch{
+			PathSpecifier: &envoy_config_route_v3.RouteMatch_Path{
```
slonka, line 90: shouldn't this be prefix (`RouteMatch_Prefix`)?

jijiechen (approved):

pkg/plugins/policies/meshpassthrough/plugin/xds/listeners_filter_chain.go
jijiechen: Yeah, the wording here needs to be improved.

PR #10624, opened 2024-06-25
feat(kuma-dp): rework on the virtual probes to support probing tcp and grpc ports

jijiechen:
### Checklist prior to review

Implements this MADR: [gRPC and TCP probes](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/047-grpc-tcp-probes.md)

  - fixes https://github.com/kumahq/kuma/issues/3875
  - tested with unit and manual tests
  - No, it's compatible
  - No

> Changelog: Supported TCP and gRPC probes for data planes running on Kubernetes


api/mesh/v1alpha1/dataplane.proto
```diff
@@ -310,33 +310,5 @@ message Dataplane {
   // defined at a Mesh level.
   MetricsBackend metrics = 2;
 
-  message Probes {
```
lukidzi, line 313: I don't think we can remove it like this. What about universal?

jijiechen: Checked the code, this field only serves Kubernetes scenarios. There is another field called `serviceProbe`. Docs at [here](https://kuma.io/docs/2.8.x/policies/service-health-probes/#universal-probes)

jijiechen:
Hi reviewers, sorry that this is a large PR. To help you review the code easily, let me explain what are included in this PR.
This PR basicly refactors the existing HTTP-only "virtual probes" and implements the new model of "virtual probe" that support all three kinds of pod probes: TCP, gRPC and HTTP.

Here are the break down tasks:
1. introduce a new component into `kuma-dp` to start the virtual probe server, it's an HTTP server. This is placed at directory `app/kuma-dp/pkg/dataplane/probes`
2. configure `kuma-dp` from the `injector.go` to run the virtual probe server according to user's auto-injector configuration
3. in the function of `OverridePodProbes` that is called by `injector.go`, change the existing virtual probe overriding to support all the three kinds of probes
4. exclude the virtual probe server port on the ways of settings of transparent proxy:
    1. kuma-sidecar: `PodRedirect` in file `pkg/transparentproxy/kubernetes/kubernetes.go`
    2. kuma-cni: `NewIntermediateConfig` in file `app/cni/pkg/cni/annotations_linux.go`

jijiechen:
The PR also includes some minor/revelant refactorings. 

Here are some of them:
1. Removed `Probes` to be on `dataplane` objects sinec we don't generate an insecure listener for the virtual probes server any more. So I removed this field and its `message` definition.
2. Removed grpc health check proto files from the repo and use the package from GitHub since it was included in the `testserver` package previously and if I introduce the same proto into a package under `kuma-dp`, it will generate a conflict panic.
3. Added the capabilities of supporting `scheme` and `httpHeaders` for HTTP probes, supporting of `timeout` is added for all three kinds of probes.

api/mesh/v1alpha1/dataplane.proto
lahabana: I'm pretty sure that's not going to be fine from a backward compat pov

jijiechen: Hi @lahabana I don't quite understand why removing this field will break backward compatibility?  
Without this field, we can still parse the existing dataplane objects stored in the DB when people upgrade their versions. So I think this is not a break change.

lobkovilya: There're might be users who rely on this functionality on Universal. Backward compatibility means not only being able to parse the format, but also preserve the functionality. Usually we mark the field as deprecated and remove it after the 2 minor releases

lobkovilya:
@jijiechen please also think about the case when old kuma-dp connects to the new kuma-cp, I believe it won't work because you removed `probe_generator.go`

api/mesh/v1alpha1/dataplane.proto
jijiechen: Thanks for the explanation of the "Backward compatibility", I agree with your view of "not only being able to parse the format, but also preserve the functionality". I meant that you guys may be confused this field with the other one `serviceProbe` which serves Universal scenarios. The field "probes" here was only serving Kubernetes scenarios.

lobkovilya: 1. It makes the most sense on k8s, but the usage is not limited to k8s. You can use it on Universal if you want to expose probe without mTLS.
2. Even on k8s I believe you can't delete it, think about the moment when old kuma-dp connects to the new kuma-cp

jijiechen: OK, I'm sold. I'll get it back.

jijiechen:
If we want to maintain "Backward compatibility" by restoring field `probes` on the `Dataplane` object and the associated implementation logic, I'll need to use a seperated port as default port for the new transformation based virtual probe functionality.

jijiechen:
Hi @lobkovilya @lahabana  thanks for reviewing this PR and the current comments are focusing one topic:
* We should bring back field `probes` on the `Dataplane` object and associated implmenetation. The reason is to keep "backward compatibility", especially when an old kuma-dp connects to the new kuma-cp

I think there is some ideas I want to align with you, please take a look at your convenience.

###  Background

1. In the previous implementation, the user defined HTTP probes are handled by a virtual listener built from xDS synced from the kuma-dp. It relies on the transparent proxy mechanism to work since the listener will not bind to port and user traffic will be redirected to Envoy anyway.

2. The new way is, transform user defined probes to a new HTTP probe that will be handled by kuma-dp. and this port now has a real listener so it'll be excluded from the transparent proxy automatically.

### Ideas

* __Q:__ Do we differenciate scenarios that should use the new way or not? 
        When we see a user defined HTTP probe on the pod, should we transform it using the old way or new way?
  __A:__ We always use the new way of transforming as it's compatibile with the old way. (The port and path are the same, and the behaviour is compatibile)

* __Q:__ Should we still generate xDS resource that builds the virtual listener on port 9000?
        This will impact the online dataplanes during an upgrade. If an old DPP is connecting to a new CP, and no xDS for the virtual listener received, the listener will be dropped, so the probes on the old DPP would fail.
__A:__ Will still generate xDS resources for the virtual listener when we see `probes` data on `Dataplane` objects

* __Q:__ How should we maintain "probes" data on Dataplane objects? 
__A:__ For exsiting Dataplane objects, we don't touch them, they are stored as is and handled identically with previous logic. 
    For new Pods, we'll not genearte "probes" data on the Dataplane object, so no virutal probe listener will generate for them
    When Pods who had exisitng Dataplane object is being updated, existing "probes" data will be kept.


Do you agree with the above ideas?

api/mesh/v1alpha1/dataplane.proto
jijiechen: Please see this comment:
https://github.com/kumahq/kuma/pull/10624#issuecomment-2252162823

app/cni/pkg/cni/main_linux.go
```diff
@@ -208,6 +208,11 @@ func prepareResult(conf *PluginConf, logger logr.Logger) error {
 	return types.PrintResult(result, conf.CNIVersion)
 }
 
+func errorLogged(logger logr.Logger, err error, message string) error {
```
lahabana, line 211: Why is the a change in the CNI for something that's not related?

pkg/core/resources/apis/mesh/dataplane_validator.go
```diff
@@ -105,26 +98,6 @@ func validateNetworking(networking *mesh_proto.Dataplane_Networking) validators.
...
-		err.Add(ValidatePort(indexPath.Field("inboundPort"), endpoint.GetInboundPort()))
-		if _, URIErr := url.ParseRequestURI(endpoint.InboundPath); URIErr != nil {
-			err.AddViolationAt(indexPath.Field("inboundPath"), `should be a valid URL Path`)
-		}
-		if _, URIErr := url.ParseRequestURI(endpoint.Path); URIErr != nil {
-			err.AddViolationAt(indexPath.Field("path"), `should be a valid URL Path`)
-		}
-	}
-	return err
-}
```
lahabana, line 126: Why have we lost validation? It was optional to start with right?

test/server/cmd/grpc_server.go
```diff
@@ -16,16 +17,16 @@ var grpcServerLog = grpcLog.WithName("server")
 
 type grpcServer struct {
 	api.UnimplementedGreeterServer
-	api.UnimplementedHealthServer
+	grpchealth.UnimplementedHealthServer
```
lahabana, line 20: Should this just be a different PR?

app/cni/pkg/cni/annotations_linux.go
```diff
@@ -36,6 +38,8 @@ var annotationRegistry = map[string]*annotationParam{
 	"iptablesLogs":                {"traffic.kuma.io/iptables-logs", "false", alwaysValidFunc},
 	"excludeInboundIPs":           {"traffic.kuma.io/exclude-inbound-ips", "", validateIPs},
 	"excludeOutboundIPs":          {"traffic.kuma.io/exclude-outbound-ips", "", validateIPs},
+	"virtualProbesEnabled":        {"kuma.io/virtual-probes", defaultVirtualProbeEnabled, alwaysValidFunc},
+	"virtualProbesPort":           {"kuma.io/virtual-probes-port", defaultVirtualProbePorts, validateSinglePort},
```
lahabana, line 42: Do we need both enabled and a port? Can't we just set the port to 0 to disable?

pkg/plugins/runtime/k8s/probes/probe_converter_test.go
```diff
@@ -0,0 +1,191 @@
...
+			podProbeYaml := `
+                timeoutSeconds: 15
+                httpGet:
+                  scheme: HTTPS
+                  path: /c1/healthz
+                  port: 8080
+                  httpHeaders:
+                  - name: Host
+                    value: example.com
+                  - name: X-Custom-Header
```
lukidzi, line 86: We don't copy custom headers?

.github/workflows/_e2e.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
   e2e:
     timeout-minutes: 60
     runs-on: ${{ fromJSON(inputs.runnersByArch)[fromJSON(inputs.matrix).arch] }}
+    if: ${{ fromJSON(inputs.runnersByArch)[fromJSON(inputs.matrix).arch] != '' }}
```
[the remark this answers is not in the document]
jijiechen: It will not be needed after the latest master is merged, otherwise it was needed because I tried to enable `ci/run-full-matrix` and it failed on ARM64 tasks without this because no runners available.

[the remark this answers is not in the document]
jijiechen: See the reply above.

app/cni/pkg/cni/annotations_linux.go
jijiechen: These are existing annotations and I'm only introducing the support into `kuma-cni`. If we want to merge them, I think it's better to use a separated PR.

app/cni/pkg/cni/injector_linux.go
```diff
@@ -161,11 +161,11 @@ func mapToConfig(intermediateConfig *IntermediateConfig, logWriter *bufio.Writer
 
 func GetEnabled(value string) (bool, error) {
 	switch strings.ToLower(value) {
-	case "enabled", "true":
+	case "enabled", "true", "yes":
```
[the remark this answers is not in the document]
jijiechen: These values are supported in the injector at [annotations.go](https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/metadata/annotations.go#L222), so I'm trying to align the values here.

app/cni/pkg/cni/main_linux.go
jijiechen: It's related. When I was debugging this feature, I found these log entries were only in the pod descriptions, but not in kuma-cni pod logs. There are other warnings printed in kuma-cni pod logs, so they are not consistant and it confused me and costed more time to find out these logs.

jijiechen: I agree it's good to make it an independent change. I made it here just because it was a quick and small change.

pkg/config/app/kuma-dp/config.go
```diff
@@ -384,3 +389,20 @@ func (d *DNS) Validate() error {
 	}
 	return nil
 }
+
+type VirtualProbesServer struct {
+	config.BaseConfig
+
+	Enabled bool   `json:"enabled,omitempty" envconfig:"virtual_probes_server_enabled"`
```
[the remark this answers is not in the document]
jijiechen: Thanks for the note, it helps a lot!

pkg/core/resources/apis/mesh/dataplane_validator.go
jijiechen: Ah, thanks for reminding. I'll bring it back.
I removed this feature on the `Dataplane` object so this validation was removed as well. But after the first round of review, I restored that field but forgot to restore this validation.

pkg/plugins/runtime/k8s/probes/probe_converter_test.go
jijiechen: We do. I forgot to assert it. I'll add it.
Thanks.

test/server/cmd/grpc_server.go
jijiechen: I had it put it here because `kuma-dp` inplicitly use this pb and it crashes and complained that the same proto was found from two packages.

pkg/transparentproxy/kubernetes/kubernetes.go
```diff
@@ -87,6 +87,7 @@ func NewPodRedirectForPod(pod *kube_core.Pod) (*PodRedirect, error) {
 	}
 
 	podRedirect.ExcludeInboundPorts, _ = metadata.Annotations(pod.Annotations).GetString(metadata.KumaTrafficExcludeInboundPorts)
```
lukidzi, line 89: nit: maybe we can merge these 2 lines in something like a method `getExcludedInboundPorts` and just one assigning

app/kuma-dp/pkg/dataplane/probes/http.go
```diff
@@ -0,0 +1,141 @@
...
+	// from [200,400)
+	body := string(b)
+	if res.StatusCode >= http.StatusOK && res.StatusCode < http.StatusBadRequest {
+		logger.V(1).Info(fmt.Sprintf("probe succeeded for %s", upstreamReq.URL.Path),
+			"headers", upstreamReq.Header, "body", body)
+		writeProbeResult(writer, Healthy)
+		return
+	}
+
+	logger.V(1).Info(fmt.Sprintf("probe failed for %s", upstreamReq.URL.Path),
```
lukidzi, line 72: nit: do we need format, or can we just add `"path", upstreamReq.URL.Path` to the log

app/kuma-dp/pkg/dataplane/probes/grpc.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+		} else {
+			logger.V(1).Info("the upstream check request failed")
+			writeProbeResult(writer, Unhealthy)
+			return
+		}
+	}
+
+	if resp.GetStatus() != grpchealth.HealthCheckResponse_SERVING {
+		logger.V(1).Info(fmt.Sprintf("the upstream server returned status %s", resp.GetStatus()))
```
lukidzi, line 81: nit: maybe just put info in the log

app/kuma-dp/pkg/dataplane/probes/tcp.go
```diff
@@ -0,0 +1,38 @@
...
+	hostPort := net.JoinHostPort(p.podAddress, strconv.Itoa(port))
+	protocol := "tcp"
+	if p.isPodAddrIPV6 {
+		protocol = "tcp6"
+	}
+	conn, err := d.Dial(protocol, hostPort)
+	if err != nil {
+		logger.V(1).Info(fmt.Sprintf("unable to establish TCP connection to %s", hostPort), "error", err)
+		writeProbeResult(writer, Unhealthy)
+		return
```
lukidzi, line 30: should we don't return here, we might leak the connection (but might be wrong)

app/kuma-dp/pkg/dataplane/probes/grpc.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/kumahq/kuma/pkg/version"
+)
+
+func (p *Prober) probeGRPC(writer http.ResponseWriter, req *http.Request) {
+	// /grpc/<port>
+
+	opts := []grpc.DialOption{
+		grpc.WithUserAgent(fmt.Sprintf("kube-probe/%s", version.Build.Version)),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
```
lukidzi, line 25: Should we add timeout?

pkg/transparentproxy/kubernetes/kubernetes.go
jijiechen: Yeah, `excludeVirtualProbePort` is that function

pkg/plugins/runtime/k8s/probes/probe_converter.go
```diff
@@ -0,0 +1,164 @@
...
+			"be set to %d. It is reserved for the dataplane that will serve pods without mTLS.", virtualPort)
+	}
+
+	probePath := p.Path()
+	if !strings.HasPrefix(p.Path(), "/") {
+		probePath = fmt.Sprintf("/%s", p.Path())
+	}
+
+	var headers []kube_core.HTTPHeader
+	headerIdx := slices.IndexFunc(p.HTTPGet.HTTPHeaders, func(header kube_core.HTTPHeader) bool {
```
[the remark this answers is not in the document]
jijiechen: It's not putting `Host` to the first one, it's trying to find possible `Host` header that already in existing headers and rename it to `X-Kuma-Probes-Host` to make sure it's not overriden by kubelet when it's sending request to the virtual probe address.
In some cases, applications want a custom `Host` header to serve probe responses.

app/kuma-dp/pkg/dataplane/probes/grpc.go
jijiechen: You mean using `Info` level?

app/kuma-dp/pkg/dataplane/probes/grpc.go
```diff
@@ -0,0 +1,87 @@
...
+
+	addr := net.JoinHostPort(p.podAddress, fmt.Sprintf("%d", port))
+	conn, err := grpc.NewClient(addr, opts...)
+	if err != nil {
+		logger.V(1).Info("unable to connect to upstream server", "error", err)
+		writeProbeResult(writer, Unhealthy)
+		return
+	}
+
+	defer func() {
```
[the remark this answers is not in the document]
jijiechen: Checked the current implementation of `NewClient`, all  return statements are returning `nil` connection when there is an error. I believe it's common practice that we should not return actual data, along with an error -- especially something needs to be disposed manually.

app/kuma-dp/pkg/dataplane/probes/tcp.go
jijiechen: https://github.com/kubernetes/kubernetes/blob/master/pkg/probe/tcp/tcp.go#L53-L58
The same logic is in Kubernetes code base, so I guess it OK to do it this way.

app/kuma-dp/pkg/dataplane/probes/grpc.go
jijiechen: `grpc.WithTimeout` is now deprecated and the recommended way is to attach timeout in the context object when invoking the RPC. We used timeout settings in `context` underneath.

app/kuma-dp/pkg/dataplane/probes/grpc.go
lukidzi:
```
logger.V(1).Info("the upstream server status", "status", resp.GetStatus())
```

lukidzi (approved):

app/cni/pkg/cni/injector_linux.go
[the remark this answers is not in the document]
lahabana: Can we open an issue to deprecate `yes`/`no` ?

app/cni/pkg/cni/main_linux.go
lahabana: Right, but the fact that you are including a bunch of "small" changes in this one PR makes it bigger which possibly explains why you are struggling to get reviews.

test/server/cmd/grpc_server.go
lahabana: I understand this. My point was just that this should be a different PR like: https://github.com/kumahq/kuma/pull/10624#discussion_r1713449313

pkg/config/app/kuma-dp/config.go
```diff
@@ -47,6 +47,10 @@ var DefaultConfig = func() Config {
 			PrometheusPort:            19153,
 			CoreDNSLogging:            false,
 		},
+		VirtualProbesServer: VirtualProbesServer{
+			Enabled: false,
+			Port:    9000,
+		},
```
lahabana, line 53: What was the motivation for not having this in the `dataplane.yaml` ?

pkg/core/resources/apis/mesh/dataplane_validator.go
```diff
@@ -50,7 +50,6 @@ func (d *DataplaneResource) Validate() error {
 
 		err.AddErrorAt(net.Field("gateway"), validateGateway(d.Spec.GetNetworking().GetGateway()))
 		err.Add(validateNetworking(d.Spec.GetNetworking()))
-		err.Add(validateProbes(d.Spec.GetProbes()))
```
lahabana, line 53: To me you still validate this. You only make it optional

pkg/plugins/runtime/k8s/webhooks/injector/injector.go
```diff
@@ -215,7 +218,7 @@ func (i *KumaInjector) InjectKuma(ctx context.Context, pod *kube_core.Pod) error
 
 	pod.Spec.InitContainers = append(append(prependInitContainers, pod.Spec.InitContainers...), appendInitContainers...)
 
-	if err := i.overrideHTTPProbes(pod); err != nil {
+	if err := probes.OverridePodProbes(pod, log); err != nil {
```
lahabana, line 221: I think the behaviour here should be safer.

It should be:

1. virtualProbesEnabled by default
2. if virtualProbesEnabled is disabled fallback to the old way

This enables users to fallback to what was done before if switching to this new option doesn't work as expected.

pkg/transparentproxy/kubernetes/kubernetes.go
```diff
@@ -158,6 +157,22 @@ func NewPodRedirectForPod(pod *kube_core.Pod) (*PodRedirect, error) {
...
+		return inboundPortsToExclude
+	}
+
+	vpPort, _ := metadata.Annotations(annotations).GetString(metadata.KumaVirtualProbesPortAnnotation)
+	if inboundPortsToExclude == "" {
+		return vpPort
+	}
+
+	return fmt.Sprintf("%s,%s", inboundPortsToExclude, vpPort)
+}
```
lahabana, line 174: I'm very uneasy with this... Why do we not have something similar to this with the current probes?

app/cni/pkg/cni/annotations_linux.go
lahabana: virtualProxyProbesPort

test/server/cmd/grpc_server.go
jijiechen: Sorry I can't agree with it. I did not encounter this issue before I made this change and when I ran into the problem, the most straitforward way is to include the fix in as part of this atomic PR.

app/cni/pkg/cni/injector_linux.go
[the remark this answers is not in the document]
jijiechen: Issue created: https://github.com/kumahq/kuma/issues/11108

app/cni/pkg/cni/main_linux.go
jijiechen: Thanks for the suggestion, I'll improve the practice in future PRs.

test/server/cmd/grpc_server.go
jijiechen: Thanks for the suggestion, I'll improve the practice in future PRs.

pkg/config/app/kuma-dp/config.go
jijiechen: "Virtual Probes" is an existing concepts and it was configured using either a pod annotation or the kuma dp injector configuration with env `KUMA_RUNTIME_KUBERNETES_VIRTUAL_PROBES_PORT`.
Previously, we captured pod probe info on `Dataplane` objects because probe paths can be different on multiple pods. The new way does not care about probe paths defined by users and all user defined information is supported nearly "as-is". So we don't need this information from the `dataplane.yaml` any more.

pkg/core/resources/apis/mesh/dataplane_validator.go
jijiechen: I forgot to bring it back when I restore the `Probes` field on the `Dataplane` object. I'll restore it now.

pkg/transparentproxy/kubernetes/kubernetes.go
jijiechen: Discussed in Slack.

pkg/config/app/kuma-dp/config.go
jijiechen: In addition, as you suggested, we should keep the capability of falling back to the old way. So I'm going to keep the path information on pods in new commits.

jijiechen:
According to recent review suggestions, here are the latest changes:
1. Introduced a new concept of "application probe proxy", and corresponding annotations/configurations
1. Brought back previous versions of "virtual probes"
1. Built the compatibility of using "virtual probes" by disabling "application probe proxy"
1. Marked "virtual probes" related annotations as Deprecated

Basically, the idea is, "application probe proxy" will be enabled defaultly and it's successor of "virtual probes".
User can still use "virtual probes" by disabling "application probe proxy" (set the port to 0)
When virtual probes are previously disabled using CP configuration or pod annotations, "application probe proxy" will also be disabled defaultly.

lukidzi (approved):

Issue #5552, opened 2022-12-23
panic in rules api


lahabana:
### What happened?

Seems to be an error in the merge algorithm


```
panic: expected struct or pointer to a struct got ptr                                                                                                                                                   
                                                                                                                                                                                                       
goroutine 1663 [running]:                                                                                                                                                                               
github.com/kumahq/kuma/pkg/core/xds.mustUnwrapStruct({0x27adb20?, 0xc000bde468?, 0x8?})                                                                                                                 
    /home/circleci/project/pkg/core/xds/merge.go:115 +0xb4                                                                                                                                              
github.com/kumahq/kuma/pkg/core/xds.appendSlices({0x27adb20?, 0xc000bdedb0?, 0x0?}, {0x27adb20?, 0xc000bde468?, 0x27adb20?})                                                                            
    /home/circleci/project/pkg/core/xds/merge.go:87 +0x9b                                                                                                                                               
github.com/kumahq/kuma/pkg/core/xds.appendSlices({0x23bbc00?, 0xc000bdedb0?, 0xb0?}, {0x2311a60?, 0xc000e4d250?, 0x30?})                                                                                
    /home/circleci/project/pkg/core/xds/merge.go:103 +0x31f                                                                                                                                             
github.com/kumahq/kuma/pkg/core/xds.MergeConfs({0xc002038160, 0x2, 0x0?})                                                                                                                               
    /home/circleci/project/pkg/core/xds/merge.go:49 +0x4af                                                                                                                                              
github.com/kumahq/kuma/pkg/core/xds.BuildRules({0xc00206d940, 0x2, 0xc000ed4e70?})                                                                                                                      
    /home/circleci/project/pkg/core/xds/rules.go:162 +0x570                                                                                                                                             
github.com/kumahq/kuma/pkg/plugins/policies/matchers.fromRules(0xc00155b328?)                                                                                                                           
    /home/circleci/project/pkg/plugins/policies/matchers/match.go:98 +0xff                                                                                                                              
github.com/kumahq/kuma/pkg/plugins/policies/matchers.MatchedPolicies({0x282755a, 0xb}, 0xc001563710, {0xc001941fb0?, 0xc000f66180?})                                                                    │
    /home/circleci/project/pkg/plugins/policies/matchers/match.go:55 +0x2fd                                                                                                                             
github.com/kumahq/kuma/pkg/plugins/policies/meshtimeout/plugin/v1alpha1.plugin.MatchedPolicies(...)                                                                                                     
[...]

PR #10631, opened 2024-06-25
test(e2e): add assertions to not panic in e2e tests

slonka:
so we know earlier about: https://github.com/kumahq/kuma/actions/runs/9603592697/job/26488848273#step:13:25165 

### Checklist prior to review

### DIFF
--- a/test/e2e_env/kubernetes/kubernetes_suite_test.go
+++ b/test/e2e_env/kubernetes/kubernetes_suite_test.go
@@ -50,8 +50,10 @@ var _ = E2ESynchronizedBeforeSuite(kubernetes.SetupAndGetState, kubernetes.Resto
 // SynchronizedAfterSuite keeps the main process alive until all other processes finish.
 // Otherwise, we would close port-forward to the CP and remaining tests executed in different processes may fail.
 var (
-	_ = SynchronizedAfterSuite(func() {}, kubernetes.ExpectCpToNotCrash)
-	_ = SynchronizedAfterSuite(func() {}, kubernetes.ExpectCpToNotPanic)
+	_ = SynchronizedAfterSuite(func() {}, func() {
+		kubernetes.ExpectCpToNotCrash()
+		kubernetes.ExpectCpToNotPanic()
+	})
 )
 
 var (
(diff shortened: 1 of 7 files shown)


test/framework/envs/universal/env.go
```diff
@@ -67,3 +68,12 @@ func PrintCPLogsOnFailure(report ginkgo.Report) {
 		}
 	}
 }
+
+func ExceptCpToNotPanic() {
+	logs, err := Cluster.GetKumaCPLogs()
+	if err != nil {
+		framework.Logf("could not retrieve cp logs")
+	} else {
+		Expect(strings.Contains(logs, "runtime.gopanic")).To(BeFalse())
```
lukidzi, line 77: I don't know if that catches all panics. This one happened in the goroutine, what if there is other panic? Example issue from the past: https://github.com/kumahq/kuma/issues/5552

```
panic: expected struct or pointer to a struct got ptr                                                                                                                                                   
                                                                                                                                                                                                       
goroutine 1663 [running]:   
```

test/e2e_env/kubernetes/kubernetes_suite_test.go
```diff
@@ -50,6 +50,7 @@ var _ = E2ESynchronizedBeforeSuite(kubernetes.SetupAndGetState, kubernetes.Resto
 // SynchronizedAfterSuite keeps the main process alive until all other processes finish.
 // Otherwise, we would close port-forward to the CP and remaining tests executed in different processes may fail.
 var _ = SynchronizedAfterSuite(func() {}, kubernetes.ExpectCpToNotCrash)
+var _ = SynchronizedAfterSuite(func() {}, kubernetes.ExceptCpToNotPanic)
```
jakubdyszkiewicz, line 53: nit typo: Expect

test/framework/envs/universal/env.go
slonka: I was afraid of just using "panic" because it might catch "lb_healthy_panic_threshold` or other Envoy strings with "panic" - changed that to include also "panic:"

test/framework/envs/multizone/env.go
```diff
@@ -4,13 +4,15 @@ import (
 	"encoding/json"
 	"sync"
 
+
```
slonka, line 7:
```suggestion
```

test/framework/envs/universal/env.go
```diff
@@ -3,11 +3,13 @@ package universal
 import (
 	"encoding/json"
 
+
```
slonka, line 6:
```suggestion
```

jakubdyszkiewicz (approved):

lukidzi (approved):

Issue #10507, opened 2024-06-17
`MeshMultiZoneService`


michaelbeaumont:
### Description

Implement [MADR](https://github.com/kumahq/kuma/pull/9979) that's already merged

- [ ] XDS
  - Envoy resource naming
- [ ] policy matching & inspect
- [ ] e2e tests
- [ ] docs

jakubdyszkiewicz:
Done by linked issues

PR #10643, opened 2024-06-26
feat(meshmultizoneservice): api

jakubdyszkiewicz:
### Checklist prior to review

MeshMultiZoneService api and validation.

Name is still under consideration, but I'd like to move forward with the rest of implem. We can still rename this before the next minor release.

The only divergence from the MADR is that instead of list of zones in status I have a list of matched mesh service. I think this will help XDS generation and can be useful in the GUI.


pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/meshmultizoneservice.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+type MeshServiceSelector struct {
+	// MatchLabels matches multiple MeshServices by labels
+	MatchLabels map[string]string `json:"matchLabels"`
+}
+
+type MeshMultiZoneServiceStatus struct {
+	// Ports is an aggregated distinct list of ports from selected MeshServices
+	Ports []meshservice_api.Port `json:"ports,omitempty"`
```
lukidzi, line 30: How useful will aggregated information be? I can imagine a situation where you can see a few ports, but in one zone, there is only one. In this case, it might be confusing for me. Do we need information indicating which zone/meshservice is the origin of this entry, or do we want to base it on the MeshServices list?

jakubdyszkiewicz: > I can imagine a situation where you can see a few ports, but in one zone, there is only one

What would be that scenario in practice?

I think in 99,9% cases you aggregate the same service over different clusters.

lukidzi: If you integrate K8s and Universal, it could happen. But true, most of the time they should aggregate the same service

lukidzi (approved):

PR #10648, opened 2024-06-26
feat(meshmultizoneservice): update status

jakubdyszkiewicz:
### Checklist prior to review

Status updater for MeshMultiZoneService. We collect ports and MeshService that matched.


pkg/config/core/resources/apis/config.go
```diff
@@ -1,7 +1,33 @@
...
 type Config struct {
+	// List of enabled core resources
 	Enabled []string `json:"enabled" envconfig:"KUMA_CORE_RESOURCES_ENABLED" default:""`
+	// Status of core resources
+	Status ConfigStatus `json:"status"`
+}
+
+type ConfigStatus struct {
+	// How often we compute status of MeshMultiZoneService
+	MeshMultiZoneServiceInterval config_types.Duration `json:"meshMultiZoneServiceInterval" envconfig:"CORE_RESOURCES_STATUS_MESH_MULTI_ZONE_SERVICE_INTERVAL"`
```
Automaat, line 18: shouldn't `envconfig` start with `KUMA_CORE_...` as `KUMA_CORE_RESOURCES_ENABLED`?

pkg/core/resources/apis/meshmultizoneservice/status_updater_test.go
```diff
@@ -0,0 +1,95 @@
...
+		}()
+
+		Expect(samples.MeshDefaultBuilder().Create(resManager)).To(Succeed())
+	})
+
+	AfterEach(func() {
+		close(stopCh)
+	})
+
+	It("should add mesh services and port to the status of multizone service", func() {
```
Automaat, line 49: Should we also add test for updating mesh service or ports? There is some logic with sorts etc, and we are not checking any of this by starting from scratch and adding service

pkg/core/resources/apis/meshmultizoneservice/status_updater.go
```diff
@@ -0,0 +1,139 @@
...
+	ctx := user.Ctx(context.Background(), user.ControlPlane)
+
+	for {
+		select {
+		case <-ticker.C:
+			start := time.Now()
+			if err := s.updateStatus(ctx); err != nil {
+				s.logger.Error(err, "could not update status of mesh multizone services")
+			}
+			s.metric.Observe(float64(time.Since(start).Milliseconds()))
```
Automaat, line 68: don't we want to have metric with fails? It is easier to monitor and add alerting on metric than on logs

pkg/config/core/resources/apis/config.go
jakubdyszkiewicz: uhhh, it definitely should! thx

pkg/core/resources/apis/meshmultizoneservice/status_updater.go
jakubdyszkiewicz: Hm, we don't do it usually. We could but at this point maybe it make sense to do this as "generic measurable ticker component". Adding this metric and writing ticker component is already a boilerplate. Let me create an issue. I'd like not to add it to this PR

jakubdyszkiewicz: https://github.com/kumahq/kuma/issues/10671

Automaat (approved):

PR #10653, opened 2024-06-27
test(transparentproxy): add transparent proxy tests

bartsmykla:
This pull request aims to enhance the tests for transparent proxy installation by introducing a set of test for various Docker images. These changes ensure that the installation process is tested across different environments and configurations, providing greater confidence in the stability and reliability of the transparent proxy installation.

Tests Configuration Enhancements:
- Created `config_transparentproxy.go` to define the `TransparentProxyConfig` struct, holding configuration parameters for testing Docker images, install flags, and IPv6 support, which in the future can be used to dynamically configure what Docker images should be tested (we could automate updating the images by the latest releases or removing ones which went out of support).
- Auto-configuration and validation methods added to ensure proper setup of configuration parameters.

Test Scenarios:
- Tests include various scenarios such as base installation on different Docker images (Ubuntu, Debian, Alpine, RHEL variants, Amazon Linux) and installation with additional flags specified in the configuration.
- Special handling for image-specific commands to ensure compatibility with different package management systems and environment-specific requirements.
- Created `EnsureInstallSuccessful` and `EnsureGoldenFiles` functions to validate installation success and verify iptables rules against golden files.
- Introduction of `genEntriesForImages` and `genEntriesForImage` functions to generate Ginkgo test entries dynamically based on the provided Docker images and additional flags.
- Defined `EntriesForImages` and `XEntriesForImages` functions to create regular and paused test entries, respectively, incorporating retry mechanisms for robustness.

Creation of Utility Functions:
  - Introduced file containing utility functions to clean and standardize `iptables-save` output, sanitize file names, and build golden file names for storing expected iptables rules.
  - Utility functions to construct container hooks for executing commands within test containers, enhancing test setup flexibility.

Instructions for Reviewers:
- Please review the added utility functions in `utils.go` for correctness and completeness.
- Validate the configuration enhancements in `config_transparentproxy.go` to ensure proper setup and handling of test parameters.
- Verify the test scenarios and dynamic generation of test entries in `install.go` for accuracy and coverage.
- Provide feedback on the overall structure, documentation, and any areas for improvement.

### Checklist prior to review

  - No relevant issues
  - It won't
  - There is no need
  - It doesn't


pkg/test/matchers/golden/update_files.go
```diff
@@ -16,5 +16,5 @@ func RerunMsg(path string) string {
 	if err != nil {
 		absPath = path + " Failed to retrieve cwd"
 	}
-	return fmt.Sprintf("Rerun the test with UPDATE_GOLDEN_FILES=true flag to update file: %s. Example: make test UPDATE_GOLDEN_FILES=true", absPath)
+	return fmt.Sprintf("Rerun the test with UPDATE_GOLDEN_FILES=true flag to update file: %s. Example: make test UPDATE_GOLDEN_FILES=true or make test/transparentproxy UPDATE_GOLDEN_FILES=true", absPath)
```
Automaat, line 19: I think we should add new entry in https://github.com/kumahq/kuma/blob/738b743bc365f49d9af6cdbc012ffb4b76dd1ba1/.github/workflows/pr-comments.yaml#L63 to utilize github action for this

test/transparentproxy/install/install.go
```diff
@@ -0,0 +1,396 @@
...
+					err,
+					"%s: copying output reader to buffer failed",
+					cmd,
+				))
+				continue
+			}
+
+			outputMap[suffix] = append(
+				outputMap[suffix],
+				fmt.Sprintf("# %s", cmd),
```
Automaat, line 170: why do we put result of all running all executables into a single golden file? Wouldn't it be easier to have them separate, like: `ubuntu-22-04-ipv4.iptables.golden`, `ubuntu-22-04-ipv4.iptables-legacy.golden`, `ubuntu-22-04-ipv4.iptables-nft.golden`. In my opinion it will be easier to immediately spot which one failed

test/transparentproxy/install/install.go
```diff
@@ -0,0 +1,396 @@
...
+
+			EnsureInstallSuccessful(container, tc.additionalFlags)
+			EnsureGoldenFiles(container, tc)
+		},
+		EntriesForImages(TProxyConfig.DockerImagesToTest),
+		XEntriesForImages(TProxyConfig.DockerImagesToTestPaused),
+	)
+}
+
+func EnsureInstallSuccessful(container testcontainers.Container, flags []string) {
```
Automaat, line 97: I think that `EnsureInstallSuccessful` and `EnsureGoldenFiles` could be moved to the `framework` directory.

test/transparentproxy/install/install.go
```diff
@@ -0,0 +1,396 @@
...
+	additionalFlags []string
+}
+
+func Install() {
+	DescribeTable(
+		"kumactl install transparent-proxy inside Docker container",
+		func(tc testCase) {
+			Expect(TProxyConfig.KumactlLinuxBin).NotTo(BeEmpty())
+
+			container, err := testcontainers.GenericContainer(
```
Automaat, line 55: It would be nice to wrap this into a builder as we do in other e2e tests so we can do something like:

```go
NewContainerSetup().
  WIthImage(tc.image).
  WithKumaExecutable(TProxyConfig.KumactlLinuxBin).
  WithPostStartHook(tc.postStart).
  WaitingForCmd([]string{"kumactl", "version"})
  Start()
```

test/transparentproxy/install/install.go
bartsmykla: I did it this way as there is a huge amount of files generated when putting it in separate files, but now when I think about it I think you are right and there are more benefits than disadvantages doing it like that

test/transparentproxy/install/install.go
bartsmykla: I don't think so to be honest as they are specific to these particular tests (install transparent proxy) and in current form there is extremely low probability these functions could be reused anywhere else.

test/transparentproxy/utils/utils.go
```diff
@@ -0,0 +1,174 @@
+package utils
```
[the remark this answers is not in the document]
bartsmykla: These function are specific to transparent proxy testing that why I placed it here, but I don't mind moving them

test/transparentproxy/install/install.go
```diff
@@ -0,0 +1,396 @@
...
+//     approach ensures robustness without significantly increasing test
+//     complexity. It is extremaly unlikely that tproxy installation would be
+//     flaky, so this method should provide a reliable solution.
+func EntriesForImages(images map[string]string) []TableEntry {
+	return genEntriesForImages(images, Entry, FlakeAttempts(3))
+}
+
+// XEntriesForImages generates paused ginkgo test entries (XEntry|PEntry) for
+// various Transparent Proxy installation scenarios on a given Docker image.
+func XEntriesForImages(images map[string]string) []TableEntry {
```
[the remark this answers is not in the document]
bartsmykla: My idea was to see information about skipped tests in the suite result, but I agree it introduces unnecessary complexity with not much value.

test/transparentproxy/install/install.go
bartsmykla: Addressed

test/transparentproxy/utils/utils.go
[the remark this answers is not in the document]
bartsmykla: Addressed

test/transparentproxy/install/install.go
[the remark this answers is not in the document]
bartsmykla: Addressed

test/transparentproxy/install/install.go
bartsmykla: Addressed

pkg/test/matchers/golden/update_files.go
bartsmykla: Addressed

Automaat:
Nice work! I hope this will address our recent iptables issues

PR #10667, opened 2024-06-27
feat(meshmultizoneservice): vip allocation

jakubdyszkiewicz:
### Checklist prior to review

VIP allocation for MeshMultiZoneService. Instead of copy-pasting it third time I refactored it so we only pass CIDRs and resource types to the allocator.


pkg/core/resources/apis/core/vip/allocator_test.go
```diff
@@ -29,9 +30,18 @@ var _ = Describe("VIP Allocator", func() {
...
-		Expect(err).ToNot(HaveOccurred())
-		allocator, err := vip.NewAllocator(logr.Discard(), 50*time.Millisecond, []vip.VIPAllocator{meshServiceAllocator})
+
+		allocator, err := vip.NewAllocator(
+			logr.Discard(),
+			50*time.Millisecond,
+			map[string]model.ResourceTypeDescriptor{
+				"241.0.0.0/8": meshservice_api.MeshServiceResourceTypeDescriptor,
+				"242.0.0.0/8": meshextenralservice_api.MeshExternalServiceResourceTypeDescriptor,
+			},
```
slonka, line 40: are we missing an entry for MeshMultiZoneService?

slonka (approved):
LGTM

pkg/core/resources/apis/core/vip/allocator_test.go
jakubdyszkiewicz: I decided not to add it.

This is _generic_ VIP allocator, so it should just differentiate one type from another.

Issue #10513, opened 2024-06-17
MeshPassthrough policy support for Delegated MeshGateway


lobkovilya:
### Description

We need to support MeshGateways

lukidzi:
Is it possible? MeshPassthrough works only thanks to a transparent proxy. Based on the original destination, we can match traffic and route it through the sidecar if there is no matching route. The Gateway defines routes and listeners and does not use a transparent proxy, making it impossible to match on the packet level. If the request comes from the outside, the Gateway needs to have a defined route to handle the request. In this case, it is more the role of MeshHTTPRoute/MeshTCPRoute to handle it. Additionally, if communication with an external endpoint is needed, the user can create a MeshExternalService.

lukidzi:
After the discussion, it only makes sense for a delegated gateway so we decided to implement it.

PR #10675, opened 2024-06-28
feat(meshpassthrough): add support for delegated gateway

lukidzi:
### Checklist prior to review

* Added support for MeshGateway in a delegated mode
* Added E2E test for delegated gateway
* Added appending domains with a port and adding them to the domains. Based on the [RFC](https://www.rfc-editor.org/rfc/rfc9110.html#section-7.2) host header might have a port at the end if it's not 80/443. 

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/10513

### DIFF
--- a/pkg/plugins/policies/meshpassthrough/plugin/v1alpha1/plugin.go
+++ b/pkg/plugins/policies/meshpassthrough/plugin/v1alpha1/plugin.go
@@ -30,18 +30,21 @@ func (p plugin) MatchedPolicies(dataplane *core_mesh.DataplaneResource, resource
 }
 
 func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *core_xds.Proxy) error {
-	if proxy.Dataplane == nil || (proxy.Dataplane.Spec.Networking.Gateway != nil && proxy.Dataplane.Spec.Networking.Gateway.Type == v1alpha1.Dataplane_Networking_Gateway_BUILTIN) {
+	if proxy.Dataplane == nil {
 		return nil
 	}
 	policies, ok := proxy.Policies.Dynamic[api.MeshPassthroughType]
 	if !ok {
 		return nil
 	}
+	if proxy.Dataplane.Spec.Networking.Gateway != nil && proxy.Dataplane.Spec.Networking.Gateway.Type == v1alpha1.Dataplane_Networking_Gateway_BUILTIN {
+		policies.Warnings = append(policies.Warnings, "policy doesn't support builtin gateway")
+		return nil
+	}
 	if proxy.Dataplane != nil && proxy.Dataplane.Spec.Networking.TransparentProxying == nil {
 		policies.Warnings = append(policies.Warnings, "policy doesn't support proxy running without transparent-proxy")
 		return nil
 	}
-
 	listeners := policies_xds.GatherListeners(rs)
 	if err := applyToOutboundPassthrough(ctx, rs, policies.SingleItemRules, listeners, proxy); err != nil {
 		return err
--- a/pkg/plugins/policies/meshpassthrough/plugin/v1alpha1/plugin.go
+++ b/pkg/plugins/policies/meshpassthrough/plugin/v1alpha1/plugin.go
@@ -3,6 +3,7 @@ package v1alpha1
 import (
 	envoy_resource "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 
+	"github.com/kumahq/kuma/api/mesh/v1alpha1"
 	core_plugins "github.com/kumahq/kuma/pkg/core/plugins"
 	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
 	core_xds "github.com/kumahq/kuma/pkg/core/xds"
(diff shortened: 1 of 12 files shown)


test/e2e_env/kubernetes/gateway/delegated/meshpassthrough.go
```diff
@@ -0,0 +1,113 @@
...
+`, config.CpNamespace, config.Mesh)
+
+			// when
+			err = kubernetes.Cluster.Install(framework.YamlK8s(meshPassthrough))
+
+			// we need to wait for a config to arrive because once request is done, connection is estabilished and it won't return 502
+			time.Sleep(5 * time.Second)
+
+			// then
+			Expect(err).ToNot(HaveOccurred())
```
lobkovilya, line 54: this assertion probably should be right after `kubernetes.Cluster.Install`

test/e2e_env/kubernetes/gateway/delegated/meshpassthrough.go
```diff
@@ -0,0 +1,113 @@
...
+    tags:
+      kuma.io/service: delegated-gateway-admin_delegated-gateway_svc_8444
+  default:
+    passthroughMode: None
+`, config.CpNamespace, config.Mesh)
+
+			// when
+			err = kubernetes.Cluster.Install(framework.YamlK8s(meshPassthrough))
+
+			// we need to wait for a config to arrive because once request is done, connection is estabilished and it won't return 502
```
lobkovilya, line 50: Can we set a small idle timeout for connections, so we don't reuse it between requests?

lukidzi: I think in this case we could try with `MeshProxyPatch` because it's on transparent proxy listener

lukidzi: Changed to proxy patch, lets see if it's stable

test/e2e_env/kubernetes/gateway/delegated/meshpassthrough.go
```diff
@@ -0,0 +1,143 @@
...
+		// we cannot test in a stable way first doing a successful request, because once connection is estabilished we don't break it after applying the policy (similar case to egress) the connection exists and works.
+		It("should control traffic to domains and IPs", func() {
+			esIP, err := framework.ServiceIP(kubernetes.Cluster, "external-service", config.NamespaceOutsideMesh)
+			Expect(err).ToNot(HaveOccurred())
+
+			// given
+			// we want the connection to close quickly, so after configuration change requests will start to fail.
+			meshProxyPatch := fmt.Sprintf(`
+apiVersion: kuma.io/v1alpha1
+kind: MeshProxyPatch
```
slonka, line 34: do we even need this if we first check that the request fails https://github.com/kumahq/kuma/pull/10675/files#diff-db8aa9c3fe244af0e07dc4a1c3d74aa18821196a4c3957b36ce775c6ba07d451R25 and only then allow it?

slonka (commented):
Some Qs

pkg/plugins/policies/meshpassthrough/plugin/v1alpha1/plugin.go
```diff
@@ -29,7 +30,7 @@ func (p plugin) MatchedPolicies(dataplane *core_mesh.DataplaneResource, resource
 }
 
 func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *core_xds.Proxy) error {
-	if proxy.Dataplane == nil {
+	if proxy.Dataplane == nil || (proxy.Dataplane.Spec.Networking.Gateway != nil && proxy.Dataplane.Spec.Networking.Gateway.Type == v1alpha1.Dataplane_Networking_Gateway_BUILTIN) {
```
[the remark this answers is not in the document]
lukidzi: We would probably produce logs that might not be important. I think we can say only in docs that it's not supported

test/e2e_env/kubernetes/gateway/delegated/meshpassthrough.go
lukidzi: The problem is the delay in configuration delivery. Once we apply the policy that disables traffic through passthrough, requests go through the gateway, but the configuration might not be delivered. In this case, the connection is established and will not break.

pkg/plugins/policies/meshpassthrough/plugin/xds/listeners_filter_chain.go
```diff
@@ -65,6 +65,9 @@ func (c FilterChainConfigurer) addFilterChainConfiguration(listener *envoy_liste
 			switch route.MatchType {
 			case Domain, WildcardDomain, IP, IPV6:
 				domains := []string{route.Value}
+				if route.MatchType == IP {
+					domains = append(domains, fmt.Sprintf("%s:%d", route.Value, c.Port))
+				}
```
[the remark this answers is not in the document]
lukidzi: It's not related I've discovered that sometimes clients add a port to the domain and might not match a host header. I can create a separate PR if you think we need it

pkg/plugins/policies/meshpassthrough/plugin/v1alpha1/plugin.go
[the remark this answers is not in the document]
lahabana: You can return policy matching warnings in MatchedPolicies.

pkg/plugins/policies/meshpassthrough/plugin/xds/listeners_filter_chain.go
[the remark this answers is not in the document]
lahabana: I think a comment to explain this would be useful...

test/e2e_env/kubernetes/gateway/delegated/meshpassthrough.go
slonka: let's leave it like this, in a perfect world we could have something like "wait until applied" (the cp would return a version of the configuration and we could check on the dp if that version was applied)

slonka (approved):
LGTM now

PR #10706, opened 2024-07-02
test(framework): init TProxyConfig only in tproxy tests

bartsmykla:
TProxyConfig is used only in tproxy tests, and when it was initialized in `init()` function, it was run even for e2e tests, which in case of legacy e2e tests was failing because of missing kumactl binary.

### Checklist prior to review

  - No relevant issues
  - It won't
  - There is no need
  - There is no need

> Changelog: skip

### DIFF
--- a/test/framework/config_transparentproxy.go
+++ b/test/framework/config_transparentproxy.go
@@ -80,7 +80,7 @@ var defaultTProxyConf = TransparentProxyConfig{
 	IPV6: false,
 }
 
-func init() {
+func InitTproxyConfig() {
 	TProxyConfig = defaultTProxyConf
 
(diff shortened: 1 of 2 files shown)


jijiechen (approved):

test/framework/config_transparentproxy.go
```diff
@@ -80,7 +80,7 @@ var defaultTProxyConf = TransparentProxyConfig{
 	IPV6: false,
 }
 
-func init() {
+func InitTproxyConfig() {
```
lahabana, line 83: If it's used in a single place what's the point of abstracting this in the framework far from where it's actually used?

bartsmykla: You are right, I'm gonna followup with the PR removing it from the framework (it's blocking other PRs, so I'm not gonna include this in this PR to merge it faster)

PR #10715, opened 2024-07-02
test(framework): move tproxy tests config from test framework

bartsmykla:
This Config is and will be used only in transparent proxy tests, so we can safely move it there. As now there is no conflict in names, I renamed it from `TProxyConfig` to just `Config`.

re. https://github.com/kumahq/kuma/pull/10706#discussion_r1661984816

### Checklist prior to review

  - No relevant issues
  - It won't
  - There is no need
  - There is no need

> Changelog: skip

### DIFF
--- a/test/transparentproxy/config.go
+++ b/test/transparentproxy/config.go
@@ -1,4 +1,4 @@
-package framework
+package transparentproxy
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


test/transparentproxy/config.go
```diff
@@ -80,14 +80,14 @@ var defaultTProxyConf = TransparentProxyConfig{
 	IPV6: false,
 }
 
-func InitTproxyConfig() {
-	TProxyConfig = defaultTProxyConf
+func init() {
```
lahabana, line 83: Don't use init it's an anti pattern that burnt you once already.

lahabana (changes requested):

test/transparentproxy/config.go
bartsmykla: addressed

lahabana (approved):

Issue #10679, opened 2024-07-01
Referencing real resource in `to[]` requires a new type of `Rule`


lobkovilya:
### Description

`TargetRefs` in `to[]` will reference real resources in the cluster. We can stop using `Rule.Subset` and start using `ResourceKey`.

We need MADR to cover changes in Inspect API.

Issue #9173, opened 2024-02-08
origin order in _rules api should reflect apply order


lahabana:
### Description

This would be very useful for people to reason about policies overriding one another.

At the moment their lexicographic:

<img width="1174" alt="image" src="https://github.com/kumahq/kuma/assets/702355/22ccc4bb-fb33-4538-9762-377425a27736">

PR #10726, opened 2024-07-02
docs(MADR): policy matching algorithm with the real resource kinds

lobkovilya:
### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/10679


docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+This structure is used by both "from" and "to" policies plugins.
+Policy plugins use the `Rules` list to compute `Conf` for the known destination.
+In example, if we want to know the `Conf` for outbound with `kuma.io/service: bar` tags,
+we're going to iterate over `Rules` and check if our outbound belongs to the `Rules[].Subset`.
+
+The approach has a number of disadvantages:
+
+* even though it works well for "from" policies, using it for "to" policies doesn't feel natural
```
slonka, line 41: Can you elaborate? I don't share the same feeling - maybe I'm missing some context.

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,185 @@
...
+
+Policy plugins should implement a fallback logic when trying to get a conf for the given `MeshService{name: backend}`:
+
+1. Lookup `MeshService{name: backend}` in the new `ToRules.ResourceRules`
+2. If it exists then the real MeshService `backend` exists in the cluster and we can use `ResourceRule.Conf`
+3. If it's absent then fallback to legacy `ToRules.Rules` and get conf for `kuma.io/service: backend` tag
+
+Eventually we're going to delete `ToRules.Rules`.
+
+When it comes to the Inspect API, we're going to add a new field `to `InspectRule`
```
slonka, line 108:
```suggestion
When it comes to the Inspect API, we're going to add a new field to `InspectRule`
```

slonka (commented):
Does this change in any way with MeshExternalService? I feel like it does not but wanted to get your opinion

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,185 @@
...
+      description: the type of the policy
+    proxyRule:
+      description: a rule that affects the entire proxy
+      $ref: '#/components/schemas/ProxyRule'
+    toRules: # should be removed eventually
+      type: array
+      description: a set of rules for the outbounds of this proxy
+      items:
+        $ref: '#/components/schemas/Rule'
+    resourceRules: # new field
```
jakubdyszkiewicz, line 127: shouldn't this be `toResourceRules`? Do you foresee this being used for anything other than `to` section?

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,185 @@
...
+    conf: conf-1
+    origin: [...]
+  - resource:
+      name: finance-db.finance
+      mesh: mesh-1
+    conf: conf-1
+    origin: [...]
+```
+
+Policy plugins should implement a fallback logic when trying to get a conf for the given `MeshService{name: backend}`:
```
jakubdyszkiewicz, line 100: How do we do this right now?

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,185 @@
...
+
+1. Lookup `MeshService{name: backend}` in the new `ToRules.ResourceRules`
+2. If it exists then the real MeshService `backend` exists in the cluster and we can use `ResourceRule.Conf`
+3. If it's absent then fallback to legacy `ToRules.Rules` and get conf for `kuma.io/service: backend` tag
```
jakubdyszkiewicz, line 104: Why do we need to fallback to legacy `ToRules.Rules`? How would this work? MeshService do not map directly to `kuma.io/service`, technically we have `.Status.Identities`, but I'd avoid using it for anything else than mTLS

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,185 @@
...
+## Considered Options and Decision Outcome
+
+No changes for "from" policies, both policy matching algorithm and Inspect API stays the same.
+
+For "to" policies the algorithm is going to return the additional structure:
+
+```go
+type ToRules struct {
+    Rules         Rules 
+    ResourceRules map[UniqueResourceKey]ResourceRule // new field
```
jakubdyszkiewicz, line 57: What is UniqueResourceKey? Is it ResourceMeta + ResourceType?

docs/madr/decisions/056-policy-matching-with-real-resources.md
Automaat: I don't think we neefd that fallback, when applying policy we know if we are looking for new ToRules or old subset, I think what is important is that we need to build ToRules and old Subset from policy, to handle old policies targeting old MeshService by name

lobkovilya: Ah right, it's not a fallback, we always know in the code when to use old `ToRules.Rules` and when to use new `ToRules.ResourceRules`

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,181 @@
...
+
+When targetRef has kind `MeshService` and uses `labels` we're going to produce multiple `ResourceRule` entities
+for each real `MeshService` that satisfies `labels`, i.e for the given `targetRef`:
+
+```yaml
+to:
+  - targetRef:
+      kind: MeshService
+      labels:
+        k8s.kuma.io/namespace: finance
```
slonka, line 75: are we missing be an example with `name/namespace` like here - https://github.com/kumahq/kuma/pull/10422/files#diff-a9aa670d35a03c1b3324191b8be01d743be8714c951ea7d4f35d8ec9e9084305R10-R12 ?

lobkovilya:
> Does this change in any way with MeshExternalService? I feel like it does not but wanted to get your opinion

@slonka it's the same approach for MeshExternalServices. In the policy plugin you'll have access to `ToRules.ResourceRules` and by getting `UniqueResourceKey` (ResourceType + ResourceKey) you'll get `Conf`.

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: I changed this part, it was never about fallback

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: added another example

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: Yeah, it's ResourceType + Mesh + Name

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,208 @@
...
+      namespace: finance
+    conf: conf1
+```
+
+we're going to produce a single `ResourceRule` entity:
+
+```yaml
+resourceRules:
+  - resource:
+      name: backend.finance
```
Automaat, line 88: One question, we were talking on Thursday meeting that you need 4 things to identify MeshService: `name`, `namespace`, `mesh` and `zone`. We have here name, namespace and mesh, what about zone? Does it even matter in this case?

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,208 @@
...
+      kind: MeshService
+      labels:
+        k8s.kuma.io/namespace: finance
+    conf: conf1
+```
+
+we are going to fetch all `MeshService` resources, filter them by `k8s.kuma.io/namespace: finance` and produce N rules:
+
+```yaml
+resourceRules:
```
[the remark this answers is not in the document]
lobkovilya: > Won't it be unexpected that the response changes based on what services are running at the moment?

Not "running", but on "what services are defined". Response change depending on "what services are running" wouldn't be nice I guess – you lost connection between DPP and CP and the Inspect API response changed. But response change depending on "what services are defined" I think is more than acceptable, no?

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: Zone will be encoded in name's hash. We might do it better with ResourceIdentity in the future, but it depends on the PoC

docs/madr/decisions/056-policy-matching-with-real-resources.md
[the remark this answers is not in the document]
michaelbeaumont: yeah definitely, by running i mean in the sense of there's a Service that exists.

I don't know. the comparison with the example I gave for a specific service still stands I think.

maybe it depends on what you're using the API for. I mean in a sense you kind of lose the information that the labels give you. Which is "all services that fulfill these labels" vs "this list of specific services"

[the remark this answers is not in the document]
lobkovilya: yeah I guess would be nicer to show both labels and the actual MeshServices that were selected by these labels. So the ResourceRule might looks something like:
```go
type ResourceRule struct {
    Resource         core_model.ResourceMeta
    Conf             interface{}
    Origin           []core_model.ResourceMeta
    SelectedByLabels map[string]string // optional field that's set when `targetRef.Labels` were specified
}
```
wdyt?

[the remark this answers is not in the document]
lobkovilya: Imagine user created a policy:
```yaml
kind: MeshTimeout
metadata:
  namespace: finance
spec:
  to:
    - targetRef:
        kind: MeshService
        labels:
          k8s.kuma.io/namespace: finance
        default: conf1
    - targetRef:
        kind: MeshService
        labels:
          kuma.io/zone: zone-1
        default: conf2
```
how the list of Rules should look like in the Inspect API in case we don't want to use specific MeshServices?

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,208 @@
...
+## Considered Options and Decision Outcome
+
+No changes for "from" policies, both policy matching algorithm and Inspect API stays the same.
+
+For "to" policies the algorithm is going to return the additional structure:
+
+```go
+type ToRules struct {
+    Rules         Rules 
+    ResourceRules map[UniqueResourceKey]ResourceRule // new field
```
Automaat, line 59: I think there is something wrong with this example, here `ResourceRules` is a map `map[UniqueResourceKey]ResourceRule` but later on it is a list without `UniqueResourceKey`: https://github.com/kumahq/kuma/pull/10726/files#diff-abf82870f4f836d209c8e63911d38b9e53732eac8d8fd88b66708ed38f9e9cdbR108-R125

Should the examples be updated?

slonka:
> I am not sure if this is related to the MADR, because you've covered here the part of creating rules from policy, and matching real resource. What would be nice is to find a way to easily identify Envoy resources that was generated for given `MeshService` or `Mesh*Route`.
> 
...
> 
> It would be nice to identify these resources by the real resource for which it was generated, for example using `UniqueResourceKey` you've introduced in this MADR. With this, applying policy could be as simple as:
> ...
> 
> What do you think?

is it always 1:1 ?

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: yes, updated

docs/madr/decisions/056-policy-matching-with-real-resources.md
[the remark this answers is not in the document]
lobkovilya: `SelectedByLabels` is kind of similar to Origin. So I decided to extend the Origin type with the index of to-item in the `to[]` array that contributed to the conf. This also solves the problem when the final conf for `MeshService` is affected by to-item with `Mesh`:

```yaml
kind: MeshTimeout
metadata:
  namespace: finance
  name: mt1
spec:
  to:
    - targetRef:
        kind: Mesh
        default: conf1
    - targetRef:
        kind: MeshService
        name: backend
        default: conf2
```
the rule will look like:

```yaml
resourceRules:
  meshservice.mesh-1.backend:
    resource:
      name: backend
      namespace: finance
    conf: merge(conf2, conf1)
    origin:
      - resource:
          name: mt1
          namespace: finance
        toIdx: 0
      - resource:
          name: mt1
          namespace: finance
        toIdx: 1
```
I updated the madr according to this.

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,226 @@
...
+      connectionTimeout: 2s
+      http:
+        requestTimeout: 20s
+      idleTimeout: 20s
+    Origin: [ ... ]
+```
+
+This structure is used by both "from" and "to" policies plugins.
+Policy plugins use the `Rules` list to compute `Conf` for the known destination.
+In example, if we want to know the `Conf` for outbound with `kuma.io/service: bar` tags,
```
lahabana, line 36:
```suggestion
For example, if we want to know the `Conf` for outbound with `kuma.io/service: bar` tags,
```

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,226 @@
...
+      type: array
+      items:
+        type: object
+        properties:
+          resourceMeta:
+            $ref: '#/components/schemas/Meta'
+          toIdx:
+            description: index of the to-item in the policy
+            type: integer
+```
```
lahabana, line 213: Need an actual example here. Computing these schema is hard.

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,226 @@
...
+resourceRules:
+  meshservice.mesh-1.backend.finance:
+    resource:
+      name: backend.finance
+      mesh: mesh-1
+    conf: conf-1
+    origin: [...]
+```
+
+When targetRef has kind `MeshService` and uses `labels` we're going to produce multiple `ResourceRule` entities
```
lahabana, line 103: Is Mesh just a special case of this (Equivalent to matching all services)?

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,226 @@
...
+```yaml
+to:
+  - targetRef:
+      kind: MeshService
+      labels:
+        k8s.kuma.io/namespace: finance
+    conf: conf1
+```
+
+we are going to fetch all `MeshService` resources, filter them by `k8s.kuma.io/namespace: finance` and produce N rules:
```
lahabana, line 115: What's the merging story?


What happens when I have:

```yaml
type: MeshService
name: my-service
labels:
  a-common-label: foo
---
type: MeshService
name: labeled-service
labels:
  a-common-label: foo
---
type: MeshService
name: other-service
labels: {}
---
kind: Policy
spec:
  targetRef:
    kind: Mesh
  to:
    - targetRef:
         kind: Mesh
       conf: $conf1
---
kind: Policy
spec:
  targetRef:
    kind: Mesh
  to:
    - targetRef:
         kind: MeshService
         labels:
             a-common-label: foo
       conf: $conf2
---
kind: Policy
spec:
  targetRef:
    kind: Mesh
  to:
    - targetRef:
         kind: MeshService
         name: my-service
       conf: $conf3
```
I'd expect:

```
resourceRules:
   my-service:
      conf: merge($conf1, $conf2, conf3)
   labeled-service:
     conf: merge($conf1, $conf2)
   other-service:
     conf: merge($conf1)
```

Is this a correct understanding? If it is it should be explicitly stated.

Also what about Mesh*Route on top of this?

lahabana (commented):
I like the route this is taking. I think we need to be explicit about merging behaviour here

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: That's correct understanding with one caveat – at this moment the order of confs inside `merge()` is defined by the alphabetical order of `metadata.name`. So depending on the alphabetical order you can get either:
```
my-service:
  conf: merge($conf1, $conf2, $conf3)
```
or 
```
my-service:
  conf: merge($conf2, $conf3, $conf1)
```
and etc.

That's the same thing you mentioned here https://github.com/kumahq/kuma/issues/9151.

lobkovilya: > Also what about Mesh*Route on top of this?

I'm planning to implement function `includes(r1, r2 core_model.Resource) bool` that returns true if 1 resource is included into another:
```
* Mesh includes every other resource
* MeshService includes MeshHTTPRoutes and MeshTCPRoute
```
it's already implemented here for simple cases https://github.com/kumahq/kuma/pull/10886/files#diff-b78b017a9fd32d0715c7c584478b3836fd22ae71c3bc2b49eb06480e3866d5b8R80.

I'll update MADR to reflect that.

lobkovilya: Added information about merging

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: It's not a special case, targetRef with Mesh is going to reference the real Mesh resource. This means Mesh is present in the ResourceRules structure.

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,524 @@
...
+But the problem is less visible due to the fact `from[].targetRef.kind` is `Mesh` in the overwhelming majority of cases.
+For MeshTrafficPermission it can be `Mesh` and `MeshSubset`, and technically the behaviour of the following policies is going to change:
+
+```yaml
+kind: MeshTrafficPermission
+name: aaa
+spec:
+  targetRef:
+    kind: Mesh
+  to:
```
jakubdyszkiewicz, line 283: nit: `from`

jakubdyszkiewicz (approved):

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya:
```suggestion
  from:
```

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,508 @@
...
+  targetRef:
+    kind: Mesh
+  from:
+    - targetRef:
+        kind: MeshSubset
+        tags:
+          kuma.io/service: backend
+      default: 
+        action: Allow
+# resulting behaviour is 'action: Deny` for all
```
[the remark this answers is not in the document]
lobkovilya: Yes, I meant the current resulting behaviour. I rephrased it to make it more clear.

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,508 @@
...
+  meshservice:mesh/mesh-1:name/my-service:
+    conf: merge($conf1, $conf2, $conf3)
+  meshservice:mesh/mesh-1:name/labeled-service:
+    conf: merge($conf1, $conf2)
+  mesh:name/mesh-1:
+    conf: $conf1
+```
+
+Note that the structure doesn't contain `other-service`, but instead it contains entry for `mesh-1`.
+The code that computes confs should handle the absence of `other-service` and replace it with conf of the `mesh-1`.
```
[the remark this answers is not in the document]
lobkovilya: Added 4th policy and added `another-common-label: foo` to the `labeled-service`, otherwise it wouldn't affect the ResourceRules. 

> Also finally, will this solve the origin order .i.e: https://github.com/kumahq/kuma/issues/9173

Good point, yes we can solve it in this MADR. I added `description` to the OpenAPI `origin` field saying the order matters.

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,543 @@
...
+        tags:
+          kuma.io/service: backend
+      default: 
+        action: Allow
+```
+
+Current behaviour is 'action: Deny' for all, because `aaa` happens to be more specific than `bbb`.
+The new behaviour is 'action: Deny' for all except `backend` and 'action: Allow' for `backend`.
+If such change in behaviour happens it probably means the user already had policies in the obscure state. 
+The change is harmless, and we don't need to provide a migration path for `from[]`.
```
slonka, line 306: I would say we need to document this. The current behaviour just seems a bit broken. Might make sense to think about any "sane" cases where this change could break something for the end user.

slonka (commented):
Apart from one worry that I have that there is a "sane" case where `from` behaviour change would cause some traffic to fail it seems 👍 from me.

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: Yeah, there is actually one kind of "sane" case, when mesh operator decided to step in and override the behaviour configured by the application developers. I'll update the MADR with this info

slonka (approved):

Automaat (approved):

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,548 @@
...
+      labels:
+        k8s.kuma.io/namespace: finance
+    conf: conf1
+```
+
+we are going to fetch all `MeshService` resources, filter them by `k8s.kuma.io/namespace: finance` and produce N rules:
+
+```yaml
+resourceRules:
+  meshservice:mesh/mesh-1:name/finance-backend:ns/finance:
```
lahabana, line 120: While the internal structure can be a dictionnary it's should be transformed into an array in the API

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,548 @@
...
+    origin:
+      type: array
+      description: |
+        The list of policies that contributed to the 'conf'. The order is important as it reflects 
+        in what order confs were merged to get the resulting 'conf'. 
+      items:
+        type: object
+        properties:
+          resourceMeta:
+            $ref: '#/components/schemas/Meta'
```
lahabana, line 473: Are we missing the sectionName in there?

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,548 @@
...
+      items:
+        $ref: '#/components/schemas/Meta'
+ResourceRule: # new type
+  type: object
+  required: [resourceMeta, conf, origin]
+  properties:
+    resourceMeta:
+      $ref: '#/components/schemas/Meta'
+    resourceSectionName:
+      type: string
```
lahabana, line 459: Should the sectionName be included in the resourceMeta?

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,548 @@
...
+      type: array
+      description: |
+        The list of policies that contributed to the 'conf'. The order is important as it reflects 
+        in what order confs were merged to get the resulting 'conf'. 
+      items:
+        type: object
+        properties:
+          resourceMeta:
+            $ref: '#/components/schemas/Meta'
+          toIdx:
```
lahabana, line 474: Why `toIdx`?  should it be just `ruleIndex` ?

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,548 @@
...
+  to:
+    - targetRef:
+        kind: Mesh
+      conf: conf1
+    - targetRef:
+        kind: MeshService
+        name: backend
+        namespace: finance
+      conf: conf2
+```
```
lahabana, line 498: Hmm one thing I've just realised it that the one above is equivalent to:

```yaml
kind: MeshTimeout
metadata:
  name: mt-1
  namespace: client-ns
  labels:
    kuma.io/mesh: mesh-1
spec:
  to:
    - targetRef:
        kind: Mesh
      conf: conf1
---
kind: MeshTimeout
metadata:
  name: mt-1
  namespace: client-ns
  labels:
    kuma.io/mesh: mesh-1
spec:
  to:
    - targetRef:
        kind: MeshService
        name: backend
        namespace: finance
      conf: conf2
```

Is this intentional? Did we play it out? I guess it makes sense but I want to make sure we've thourougly thought about this

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: ResourceMeta identifies the entire resource, we can't put `sectionName` inside it. I could've created a new type like:
```go
type ResourceSection struct {
    ResourceMeta
    SectionName string
}
```
but the usage at this moment is scoped to ResourceRules (and Inspect API), so I don't think introducing new type makes much sense

docs/madr/decisions/056-policy-matching-with-real-resources.md
lobkovilya: Yes, it's intentional, the reason why users would prefer a single policy with multiple to-items is to have configurations for different services in the same consumer policy.

lobkovilya: Yes, it's intentional, the reason why users would prefer a single policy with multiple to-items is to have configurations for different services in the same consumer policy.

docs/madr/decisions/056-policy-matching-with-real-resources.md
```diff
@@ -0,0 +1,548 @@
...
+
+we're going to produce the following `ResourceRule`:
+
+```yaml
+resourceRules:
+  meshservice:mesh/mesh-1:name/backend:ns/finance:section/http-port:
+    resource:
+      name: backend.finance
+      mesh: mesh-1
+    resourceSectionName: http-port
```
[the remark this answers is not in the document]
lobkovilya: answered in the [next comment](https://github.com/kumahq/kuma/pull/10726#discussion_r1689494301)

Issue #10602, opened 2024-06-24
Reachable services with `MeshExternalService`


slonka:
### Description

We should support them. xref https://github.com/kumahq/kuma/issues/10508

PR #10737, opened 2024-07-02
docs(MADR): reachable/autoreachable services with Mesh*Service

lukidzi:
### Checklist prior to review

We aim to provide feature parity for MeshService and MeshExternalService. One of the missing features is reachable services, and the goal is to provide some ideas and determine the best way to achieve that.
 
xrel: https://github.com/kumahq/kuma/issues/10602


docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,239 @@
...
+Another thing that is required to do is add another fields to the `BackendRef` in Dataplane object: namespace and labels
+```proto
+      message BackendRef {
+        // Kind is a type of the object to target. Allowed: MeshService, MeshExternalService
+        string kind = 1;
+        // Name of the targeted object
+        string name = 2;
+        // Port of the targeted object. Required when kind is MeshService, MeshExternalService
+        uint32 port = 3;
+        // Namespace of the targeted object.
```
lukidzi, line 147: I am not yet fully certain if this is necessary, as the name might already include this functionality. The name is a bit tricky because, once it's synced, it has a hash and namespace appended, which can cause a lot of confusion.

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Reachable Services
+The functionality allows the user to define a static list of services they want to communicate with. Currently, on Kubernetes, this is possible by setting the annotation `kuma.io/transparent-proxying-reachable-services` with the values of kuma.io/service
+
+Example:
+
+```yaml
+...
+  annotations:
+    kuma.io/transparent-proxying-reachable-services: ["demo-app_kuma-demo_svc_5000", "redis_kuma-demo_svc_6379"]
```
jakubdyszkiewicz, line 17: nit: it's `kuma.io/transparent-proxying-reachable-services: "demo-app_kuma-demo_svc_5000,redis_kuma-demo_svc_6379"`

it's a comma separated string. Not that it matters for the rest of madr

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,239 @@
...
+          kuma.io/servic: "demo-app_kuma-demo_svc_5000"
+      default:
+        action: Allow
+```
+
+More details in (MADR #036)
+
+## Considered Options
+
+* Support reachable and autoreachable services for `MeshService` and `MeshExternalService`
```
jakubdyszkiewicz, line 65: ...and MeshMultiZoneService as well?

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,239 @@
...
+
+In the case of Kubernetes, we can introduce a new annotation that has a more structured model and better flexibility.
+
+```yaml
+kuma.io/reachable-backend-refs: |
+  refs:
+  - kind: MeshService
+    name: demo-app
+    namespace: kuma-demo # if missing, then my namespace
+    port: 5000
```
jakubdyszkiewicz, line 81: `# if missing then all ports of a given service`

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,239 @@
...
+---
+kind: MeshTrafficPermission
+spec:
+  targetRef:
+    kind: MeshSubset
+    tags:
+      app: backend
+```
+
+In the case of `MeshExternalService`, we can introduce a top-level TargetRef for this type, which selects a specific `MeshExternalService`. The problem is that once the resource is synced between zones and a hash is added, the name needs to be the hashed one.
```
jakubdyszkiewicz, line 181: I'd try to avoid making any decision when it comes to MeshExternalService + Autoreachable services here. We need to first figure out how does MeshExternalService + MeshTrafficPermission work (probably a separate madr). Then we can build autoreachable services on top of that

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,239 @@
...
+        // Labels of the targeted object
+        map<string, string> labels = 5;
+      }
+```
+
+This enhancement should make it easier to match the reachable services provided by the user with the generated or provided outbounds.
+
+### Autoreachable services
+
+From a code perspective, we used to build a map of supported tags: `kuma.io/service`, `k8s.kuma.io/namespace`, `k8s.kuma.io/service-name`, and `k8s.kuma.io/service-port`, and later match `MeshTrafficPermission` to dataplanes with these tags. Since the introduction of `MeshService` and `MeshExternalService`, we might not need to iterate over all dataplanes. Instead, we can match all `MeshService` and `MeshExternalService` entries. Based on this information, we know which dataplanes need to be configured. The only limitation is that `MeshService` doesn't carry all tags that a dataplane might have, which means we can support a limited scope of tags, similar to how we do with `kuma.io/service`.
```
jakubdyszkiewicz, line 159: It's not clear from the description that to avoid iterating over ddps we can allow MeshSubet but *only* with `dpTags` selected by MeshService.

This seems ok, as it's very likely that if Service Owner has a "main application tag" like `app`, they will use the same thing in `MeshTrafficPermission`

docs/madr/decisions/056-reachable-svc-mesh-service.md
michaelbeaumont: I think this should align with the backendRef of routes ultimately, right?

docs/madr/decisions/056-reachable-svc-mesh-service.md
lukidzi: We should support them too. Added

docs/madr/decisions/056-reachable-svc-mesh-service.md
lukidzi: Changed and added a note that we need to decide how to do this. So far we should be fine with `MeshSubset`

docs/madr/decisions/056-reachable-svc-mesh-service.md
lukidzi: Tried to clarify this

docs/madr/decisions/056-reachable-svc-mesh-service.md
lukidzi: It should, but I feel it might not. Currently, backendRef in a dataplane's outbound has a name in the format `name-hash.namespace` (hash when synced), while backendRef in a route has `name-hash` with the namespace in a separate field. That's why I am considering doing the same for outbounds.

docs/madr/decisions/056-reachable-svc-mesh-service.md
jakubdyszkiewicz: I think a simple algo definition would be helpful here to clarify how we want to do it. It's a bit handwavy right now.

It's something like
1) Take each MeshService
2) For each MeshService, take tags from dpTags (+kuma.io/zone, +kuma.io/origin) and build a fake "DPP" with one inbound
3) Execute MatchedPolicies on this "DDP" with a list of MeshTrafficPermission to get rules
4) Store rules for each MeshService.
5) When we build XDS config for client DPP and check if we should add a cluster/listener for given MeshService, execute rules for MeshService using client DPP.

Writing this clarifies why we can only support dpTags defined in MeshService.

Also for each MeshService we need to transform list of MeshTrafficPermission using `trimNotSupportedTags`. This time we need to pass the tags of fake DPP instead of static list of SupportedTags. I would not get into details of this func, just reference it.

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,242 @@
...
+---
+kind: MeshTrafficPermission
+spec:
+  targetRef:
+    kind: MeshSubset
+    tags:
+      app: backend
+```
+
+In the case of `MeshExternalService`, we need to decide how we want to target them (this needs to be clarified in a separate MADR). For now, we can provide `MeshSubset` support.
```
jakubdyszkiewicz, line 181: Add that as for now, we will just add all MeshExternalServices until we decide how to handle MTP

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,254 @@
...
+
+Instead we can retrive `MeshService` and match based on its dpTags.
+
+### Autoreachable services
+
+From a code perspective, we used to build a map of supported tags: `kuma.io/service`, `k8s.kuma.io/namespace`, `k8s.kuma.io/service-name`, and `k8s.kuma.io/service-port`, and later match `MeshTrafficPermission` to dataplanes with these tags. Since the introduction of `MeshService` and `MeshExternalService`, we might not need to iterate over all dataplanes. Instead, we can match all `MeshService` and `MeshExternalService` entries. Based on this information, we know which dataplanes need to be configured. The only limitation is that `MeshService` doesn't carry all tags that a dataplane might have, which means that when using `MeshSubset`, users should be allowed to use the same tags as `dpTags` defined in `MeshService`.
+
+Algorithm:
+
+1. Take each`MeshService`
```
slonka, line 167: nitpick:

```suggestion
1. Take each `MeshService`
```

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,254 @@
...
+spec:
+  destinations:
+  - address: 192.168.0.1
+    port: 9090 
+---
+kind: MeshTrafficPermission
+spec:
+  targetRef:
+    kind: MeshExternalService
+    name: mydomain-HASH
```
slonka, line 216: why would we use a name with the hash?

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,254 @@
...
+  name: mydomain
+  namespace: kuma-system
+  labels:
+    my.label.io/service: my-service 
+    kuma.io/mesh: default
+    kuma.io/zone: east-1
+    kuma.io/origin: zone
+    access.io/domain: "httpbin"
+spec:
+  destinations:
```
slonka, line 235: nitpick:

```suggestion
  endpoints:
```

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,254 @@
...
+---
+kind: MeshTrafficPermission
+spec:
+  targetRef:
+    kind: MeshSubset
+    tags:
+      app: backend
+```
+
+In the case of `MeshExternalService`, we need to decide how we want to target them (this needs to be clarified in a separate MADR). For now, we are going to provide all `MeshExternalService` entries until we figure out targeting.
```
slonka, line 193: Let's start working on this madr since we're already doing some targeting PRs

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,254 @@
...
+    // MeshExternalService and MeshService. Setting an explicit list
+    // can dramatically improve the performance of the mesh. If not specified,
+    // all services in the mesh are reachable.
+    repeated ReachableBackendRef reachable_backends_ref = 7;
+    }
+```
+
+Most of the fields are optional and may not be used simultaneously. We need to support the following cases:
+* When `name`, `namespace`, and `port` are defined(combination of them, all are not requried), you cannot use `labels`.
+* When `labels` are defined, you cannot use `name`, `namespace`, and `port`.
```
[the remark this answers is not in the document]
lukidzi: Once the resource is synced, you need to use the name with the hash and the hash says from which zone it is.

docs/madr/decisions/056-reachable-svc-mesh-service.md
lukidzi: If the MeshExternalService is synced from global we need to use a resource name in the zone, if we create it on zone. I might write this example better

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,254 @@
...
+        //  +optional
+        string namespace = 4;
+        // Labels of the targeted object
+        map<string, string> labels = 5;
+      }
+```
+
+This enhancement should make it easier to match the reachable services provided by the user with the generated or provided outbounds. There is a disadventage that makes it maybe not the best option. If we expose it to the user, it might configure one `BackendRef` that reference many MeshServices. 
+
+Instead we can retrive `MeshService` and match based on its dpTags.
```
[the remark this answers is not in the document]
lukidzi: We won't add this to the backendRef of data plane's outbound. Instead, we can retrieve a MeshService and validate that there

docs/madr/decisions/056-reachable-svc-mesh-service.md
lukidzi: It should be separate, and might be related to the #10726

jakubdyszkiewicz: It is a MADR of how to handle MeshTrafficPermission with MeshExternalService.
I don't think it's related with Ilya's design, because it's about handling MTP with MES on Egress.

docs/madr/decisions/056-reachable-svc-mesh-service.md
```diff
@@ -0,0 +1,254 @@
...
+spec:
+  targetRef:
+    kind: MeshSubset
+    tags:
+      app: backend
+```
+
+In the case of `MeshExternalService`, we need to decide how we want to target them (this needs to be clarified in a separate MADR). For now, we are going to provide all `MeshExternalService` entries until we figure out targeting.
+
+Option 1
```
jakubdyszkiewicz, line 195: Let's not go into options here and leave it for MTP + MES madr

docs/madr/decisions/056-reachable-svc-mesh-service.md
slonka: [#10868 Support for a MeshExternalService targeting in MeshTrafficPermission](https://github.com/kumahq/kuma/issues/10868)

slonka (approved):

jakubdyszkiewicz (approved):

Automaat (approved):

PR #10745, opened 2024-07-03
ci(.github): enable self hosted runners for AMD64 E2E tasks

jijiechen:
### Checklist prior to review

Set the runner dynamically to be used before invoking `_e2e.yaml` from `_test.yaml` to support switching runners based on whether a build is trusted or not.

  - fixes https://github.com/kumahq/kuma/issues/10585
  - Confirmed
  - Manually tested
  - No need
- Added


jijiechen:
Here is a test run with self hosted runners enabled on e2e:
https://github.com/kumahq/kuma/actions/runs/9775105506

jakubdyszkiewicz:
PR looks good to me, but I see that sample run fails :(

jijiechen:
Yeah, some tasks failed at `make dev/tools`, I'll take a look.

jakubdyszkiewicz (approved):

.github/workflows/_test.yaml
```diff
@@ -95,6 +96,48 @@ jobs:
...
+              ALL_RUNNERS=$(echo $ALL_RUNNERS | jq -r ".$ARCH=\"\"")
+            else
+              ALL_RUNNERS=$(echo $ALL_RUNNERS | jq -r ".$ARCH=.$ARCH[0]")
+            fi
+          done
+
+          echo "final runners: $ALL_RUNNERS"
+          echo "runners<<EOF" >> $GITHUB_OUTPUT
+          echo "$ALL_RUNNERS" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
```
lahabana, line 140: Isn't that just:

```shell
ALL_RUNNERS='{"amd64": "ubuntu-latest", "arm64": "ubuntu-latest-arm64"}'
if [[ "${{ evv.IS_TRUSTED_BUILD}}" == "true" ]]; then
  ALL_RUNNERS='{"amd64": "ubuntu-latest-kong", "arm64": "ubuntu-latest-arm64-kong"}'
fi

echo "final runners: $ALL_RUNNERS"
echo "runners<<EOF" >> $GITHUB_OUTPUT
echo "$ALL_RUNNERS" >> $GITHUB_OUTPUT
echo "EOF" >> $GITHUB_OUTPUT
```

?

Then in KM we can simply override `IS_TRUSTED_BUILD` to always be true.

PR #10748, opened 2024-07-03
chore(zoneingress): refactor building endpoints

### LINKED ISSUE #10747: Refactor Zone Proxies endpoints
### Description

Zone Ingress builds separate endpoint map for each ingress. First of all, it should reuse one endpoint map for all ingresses. Second is that ideally it should construct the map from existing EDS endpoint maps

jakubdyszkiewicz:
### Checklist prior to review

Fix #10747

Refactor building endpoints for ZoneIngress.

We had a code duplication when it comes to building endpoints for regular DPPs and ZoneIngress.
After analyzing the code, it seems we can just
* Reuse the same `BuildEdsEndpointMap` function, but without passing a list of ZoneIngresses, because ZoneIngress should not pass connections to other zone ingresses
* Add local cross mesh outbound

We were building an endpoint map for each ZoneIngress. This time we build it in MeshContextBuilder so it's shared for all ZoneIngresses. This should improve the performance if we run multiple ZoneIngresses.

We could also mark Endpoints that target ingresses.
Then take the results of BuildEdsEndpointMap and BuildCrossMeshEndpointMap and filter out those endpoints. I didn't want to do this now, I worry this would be error-prone, because we also need to manipulate weights.

I moved the test to outbound package. Why the test result is a bit different?
In the existing implementation, we were also building something called `DestinationMap`, so a simple map of possible destinations based on AvailableServices. I got rid of this concept since it's unnecessary. Everything is filtered by AvailableServices on later stage anyways and in the long term we want to get rid of AvailableServices. Some of the tests did not set destinations correctly, therefore some endpoint map entries were missing.


Automaat (approved):

lobkovilya (approved):
LGTM!

Issue #8411, opened 2023-11-20
[Transparent Proxy] Add comments to `iptables` rules


bartsmykla:
### Description

When I was floating around other service meshes I learned something new about iptables.
More specifically that you can add comments to rules using `comment` extension

I think it could be a good improvements so users could better understand the purpose of each rule. It might be beneficial even for us, when we are be debugging issues related to transparent proxy.

---

```sh
comment
    Allows you to add comments (up to 256 characters) to any rule.

    --comment comment

    Example:
          iptables -A INPUT -i eth1 -m comment --comment "my local LAN"
```
ref. [man iptables-extensions(8)](https://man7.org/linux/man-pages/man8/iptables-extensions.8.html)

---

[Example from Linkerd](https://github.com/linkerd/linkerd2-proxy-init/blob/f934c857ca25181e7e95d90212f133b5442a3a8d/internal/iptables/iptables_test.go#L27-L41)

```sh
# Generated by iptables-save v1.6.1 on Fri Jan  6 23:00:00 2023
*nat
:PREROUTING ACCEPT [0:0]
:INPUT ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
:POSTROUTING ACCEPT [0:0]
:PROXY_INIT_OUTPUT - [0:0]
:PROXY_INIT_REDIRECT - [0:0]
-A PREROUTING -m comment --comment "proxy-init/install-proxy-init-prerouting/testExecutionTraceID" -j PROXY_INIT_REDIRECT
-A OUTPUT -m comment --comment "proxy-init/install-proxy-init-output/testExecutionTraceID" -j PROXY_INIT_OUTPUT
-A PROXY_INIT_OUTPUT -o lo -m comment --comment "proxy-init/ignore-loopback/testExecutionTraceID" -j RETURN
-A PROXY_INIT_OUTPUT -p tcp -m comment --comment "proxy-init/redirect-all-outgoing-to-proxy-port/testExecutionTraceID" -j REDIRECT --to-ports 1234
-A PROXY_INIT_REDIRECT -p tcp -m multiport --dports 1234 -m comment --comment "proxy-init/ignore-port-1234/testExecutionTraceID" -j RETURN
COMMIT
# Completed on Fri Jan  6 23:00:00 2023
```

jakubdyszkiewicz:
> Triage: if we cannot squeeze full explanation we should link to docs done by https://github.com/kumahq/kuma-website/issues/1287


[...]

PR #10809, opened 2024-07-05
feat(transparent-proxy): add comments to tproxy iptables rules

bartsmykla:
This PR adds functionality and comments itself to every iptables rule we are generating. This has 2 purposes:
- helps explaining what each rule is for 
- in the near future it'll allow us to add transparent proxy uninstallation functionality

Closes https://github.com/kumahq/kuma/issues/8411

### Checklist prior to review

  - https://github.com/kumahq/kuma/issues/8411
  - It won't
  - Updated existing tests and manually tested
  - There is no need
  - There is no need

### DIFF
--- a/test/transparentproxy/install/testdata/alpine-3-redirect-all-dns-traffic.ip6tables-nft.golden
+++ b/test/transparentproxy/install/testdata/alpine-3-redirect-all-dns-traffic.ip6tables-nft.golden
@@ -1,10 +1,10 @@
 *raw
 :PREROUTING ACCEPT
 :OUTPUT ACCEPT
--A PREROUTING -p udp -m comment --comment "kuma/mesh/tproxy/assign connection tracking zone 1 to all DNS responses" -m udp --sport 53 -j CT --zone 1
--A OUTPUT -p udp -m comment --comment "kuma/mesh/tproxy/assign connection tracking zone 1 to DNS traffic from the kuma-dp user (UID 5678)" -m udp --dport 53 -m owner --uid-owner 5678 -j CT --zone 1
--A OUTPUT -p udp -m comment --comment "kuma/mesh/tproxy/assign connection tracking zone 2 to DNS responses from the kuma-dp DNS proxy" -m udp --sport 15053 -m owner --uid-owner 5678 -j CT --zone 2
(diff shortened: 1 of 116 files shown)


pkg/transparentproxy/iptables/builder/builder_table_nat.go
```diff
@@ -127,82 +155,137 @@ func buildMeshOutbound(
...
-			Jump(Return()),
+		AddRules(
+			rules.
+				NewRule(
+					Source(Address(inboundPassthroughSourceAddress)),
+					OutInterface(cfg.LoopbackInterfaceName),
+					Jump(Return()),
+				).
+				WithCommentf("prevent traffic loops by ensuring traffic from the sidecar proxy (using %s) to loopback interface is not redirected again", inboundPassthroughSourceAddress),
+			// Redirect outbound TCP traffic (except DNS port 53) from the
```
michaelbeaumont, line 166: Is it necessary to have the comment here in the code and in the rules?

pkg/transparentproxy/iptables/consts/consts.go
```diff
@@ -139,3 +139,8 @@ var FallbackExecutablesSearchLocations = []string{
 // Debug log level for iptables LOG jump target
 // ref. https://git.netfilter.org/iptables/tree/extensions/libebt_log.c#n27
 const LogLevelDebug uint16 = 7
+
+// IptablesRuleCommentPrefix defines a consistent prefix used in iptables rule
+// comments. This prefix helps identify and distinguish rules that were
+// specifically generated by our transparent proxy.
+const IptablesRuleCommentPrefix = "kuma/mesh/tproxy"
```
michaelbeaumont, line 146: Maybe not "tproxy", doesn't `tproxy` exist as a separate feature/concept at the kernel level?

pkg/transparentproxy/iptables/builder/builder_table_nat.go
bartsmykla: Addressed

pkg/transparentproxy/iptables/consts/consts.go
bartsmykla: Addressed

michaelbeaumont (approved):
Nice this will be really helpful for understand

I guess the comment prefix doesn't go over the limit

bartsmykla:
> I guess the comment prefix doesn't go over the limit

It doesn't (checked)

Issue #3250, opened 2021-11-22
Http support in TrafficPermissions


lahabana:
### Description

Extend TrafficPermission resource to support HTTP traffic

lahabana:
Probably blocked by policy matching rewrite.

bartsmykla:
> We need to spend some time to figure out how what what we want to achieve here. It seems that `from` should be differently implemented than currently

lahabana:
Is this related to: https://github.com/kumahq/kuma/issues/10247 ?

PR #10858, opened 2024-07-10
docs(MADR): targeting mesh*routes in to section of policy

Automaat:
MADR for issue: https://github.com/kumahq/kuma/issues/10247

### Checklist prior to review


docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,379 @@
...
+      default:
+        http:
+          requestTimeout: 5s
+          streamIdleTimeout: 1h
+```
+
+To introduce this change we need to deprecate specifying Mesh*Route in topLevel targetRef.
+
+Should we allow selecting MeshGateway in topLevel targetRef? probably this is obsolete as Route will select gateway, and we
+want to apply policy on a route. Is there a need to make it more specific than route?
```
slonka, line 80: Can we have the same route attached to multiple gateways? Does it make sense to have multiple gateways? 🤔

slonka (commented):
just a comment now, will do another round of this

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,379 @@
...
+  to:
+  - targetRef:
+      kind: MeshService
+      name: backend
+      namespace: backend-ns
+    default:
+      http:
+        streamIdleTimeout: 2h
+---
+# MeshTimeout targeting producer route
```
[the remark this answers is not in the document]
Automaat: right, fixed

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,379 @@
...
+After applying this two policies timeout won't be applied to any proxy, since Mesh*Route and MeshTimeout is applied
+on different subset of proxies. We should detect this situation and add warning in Inspect API.
+
+#### Overriding configuration
+
+For policies on namespace this is as simple as creating consumer policy that targets MeshService or Mesh*Route:
+(This can be useful when we consume MeshService from other zone)
+
+```yaml
+# Producer MeshTimeout targeting whole MeshService
```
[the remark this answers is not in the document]
Automaat: right, I've copied example and then forgot to change comment on second reading of madr

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,379 @@
...
+- add possibility to put Mesh*Routes in `spec.to[].targetRef`
+- leave Mesh*Routes in top level targetRef as it is now
+
+## Decision Outcome
+
+- add possibility to put Mesh*Routes in `spec.to[].targetRef`
+
+### Positive consequences
+
+- improved policy validation, we can enforce applying only http config on MeshHttpRoute
```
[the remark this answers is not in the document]
Automaat: I got this suggestion from original issue, but now when I think about it, I don't see any valid example of this

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,379 @@
...
+- simplifying topLevel targetRef as it will now match only whole proxies
+- no need for specifying topLevel targetRef when targeting `Mesh`
+- easier policy matching code
```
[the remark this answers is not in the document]
Automaat: added

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,379 @@
...
+real resource in `to` section we will create confs by resources
+
+```go
+map[UniqueResourceKey]ResourceRule{
+    "backend.backend-ns": {  // this is just a placeholder name for visualisation. UniqueResourceKey is build from ResourceType, Name and Mesh, I will update it after real resource matching madr is merged
+        requestTimeout: "2s"
+        streamIdleTimeout: "1h"
+    },
+    "backend-route.backend-ns": {
+		requestTimeout: "10s"
```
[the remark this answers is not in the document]
Automaat: right, updated examples

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,379 @@
...
+        name: route-to-backend
+      default:
+        http:
+          requestTimeout: 5s
+          streamIdleTimeout: 1h
+```
+
+To introduce this change we need to deprecate specifying Mesh*Route in topLevel targetRef.
+
+Should we allow selecting MeshGateway in topLevel targetRef? probably this is obsolete as Route will select gateway, and we
```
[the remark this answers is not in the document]
Automaat: I think it is simple, you create route on MeshGateway:
```yaml
kind: MeshHTTPRoute
metadata:
  name: demo-app-route
  namespace: kuma-system
  labels:
    kuma.io/origin: zone
    kuma.io/mesh: default
spec:
  targetRef:
    kind: MeshGateway
    name: demo-app
  to:
    - targetRef:
        kind: Mesh
      rules:
        - matches:
            - path:
                type: PathPrefix
                value: "/"
          default:
            backendRefs:
              - kind: MeshService
                name: demo-app_kuma-demo_svc_5000
```

This route is applied on MeshGateway and then you create MeshTimeout policy like: 

```yaml
kind: MeshTimeout
metadata:
  name: demo-app-timeout
  namespace: kuma-system
  labels:
    kuma.io/origin: zone
    kuma.io/mesh: default
spec:
  to:
    - targetRef:
        kind: MeshHTTPRoute
        name: demo-app-route
      default:
        requestTimeout: 5s
```

This should be applied on Gateway on route since you target Mesh in topLevel targetRef on MeshTimeout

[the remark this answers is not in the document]
michaelbeaumont: Of course we need MeshGateway in the top level. There is just as much a need for MeshGateway top level as there is for MeshSubset.

1. How do I apply configuration for all routes?
2. Are we even sure that we can say whether _all_ Envoy configuration belongs to a route or not?

docs/madr/decisions/057-rework-mesh-routes-matching.md
michaelbeaumont: Yes and yes.

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,379 @@
...
+
+At the moment when targeting Mesh*Route you need to put it in topLevel targetRef. 
+We want to iterate over policy design to fully utilize the fact that we are referencing real objects.
+The current implementation has made things a little tricky because policy matching is convoluted where the top level 
+MeshHTTPRoute gets converted to a fake `to`.
+
+## Considered options
+
+- add possibility to put Mesh*Routes in `spec.to[].targetRef`
+- leave Mesh*Routes in top level targetRef as it is now
```
michaelbeaumont, line 17: I hope there's an evaluation of this option and its advantages/disadvantages. In particular from a UX point of view.

> because policy matching is convoluted where the top level 
MeshHTTPRoute gets converted to a fake `to`.

this isn't convincing to me. The _main_ interface here is the policy structure, what does it matter that it gets converted to a fake `to`? Why don't we just change the implementation?

michaelbeaumont: There's a very strong intuition I have that it is _much_ clearer what is happening for a user when they simply see _one_ `spec.targetRef`, at the top level, pointing to a route. The policy applies to this route, done. No noise, no ambiguity.

As compared to when there's `spec.targetRef`, there's the policy namespace, there's multiple `spec.to[].targetRef`. How do these things interact?

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
Automaat: This point is not about removing MeshGateway from topLevel targetRef entirely, but to not use it with Mesh*Route in to section. 

Ad 1. in this case, you will have topLevel targetRef it to you will most likely have `Mesh` or `MeshService`
Ad 2. Only part of Envoy config belongs to a route, at the moment it is requestTimeout and requestHeaderTimeout

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,382 @@
...
+      kind: MeshHTTPRoute
+      name: route-to-backend
+    default:
+      http:
+        requestTimeout: 10s
+```
+
+As described in targeting real resources MADR, while merging policies we will create `ResourceRules map[UniqueResourceKey]ResourceRule` 
+field that holds configuration for a given resource identified by `UniqueResourceKey`. So when user creates policy that has 
+real resource in `to` section we will create confs by resources
```
michaelbeaumont, line 148: This explanation feels kind of incomplete or has missing context, what Dataplane does this apply to? It might be a lot clearer in fact to just show the resulting config for a given Dataplane.

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
michaelbeaumont: OK, understood. Well as far as I understand the whole point of having `spec.targetRef` is to select which proxies the policy applies to. From that point of view I don't see why you'd want to forbid it. But maybe we can evaluate in general which top level targetRefs need to be available.

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: I've added more context to this section

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: I've updated the MADR with more context and concrete advantages/disadvantages. If'you have any more questions, talking points please ask

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,379 @@
...
+## Context and Problem Statement
+
+At the moment when targeting Mesh*Route you need to put it in topLevel targetRef. 
+We want to iterate over policy design to fully utilize the fact that we are referencing real objects.
+The current implementation has made things a little tricky because policy matching is convoluted where the top level 
+MeshHTTPRoute gets converted to a fake `to`.
+
+## Considered options
+
+- add possibility to put Mesh*Routes in `spec.to[].targetRef`
```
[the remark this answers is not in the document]
Automaat: maybe I am missing something here, but the more I think about it, that the two options are similar. Mesh*Route in `spec.to[].targetRef` gives more flexibility with overriding conf but `spec.targetRef` could be simpler to understand

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+#### Advantages
+
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will pick outbounds
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies)
+- no need for specifying `spec.targetRef` when targeting `Mesh`
+- easier policy matching code
+
+#### Disadvantages
+
+- can be harder to understand because `spec.targetRef` can be MeshSubset
```
lahabana, line 32: One suggestion here was to make `spec.targetRef` optional so most people wouldn't even know it exists

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -105,9 +105,15 @@ At the moment we can only target `Mesh*Routes` in MeshTimeout policy. This MADR
 Looking at what can be configured at Envoy route we can also apply this to our MeshRetry policy. Rate limit is also configurable on route,
 but we are configuring rate limit on inbound traffic, which does not apply to this MADR. 
 
+#### Producer/consumer model for targeting Mesh*Route in other policies
+
 It is worth to mention that rate limit would not work with our producer/consumer model. In our model consumer always has more priority
```
michaelbeaumont, line 110: I think we only support rate limit + routes if they're attached to a gateway, right? Otherwise rate limit is an inbound policy

Automaat: at the moment we can target whole `MeshGateway` not the route in MeshRateLimit, but since MeshGateway is cluster scoped, producer/consumer model does not apply to it

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: And it is treated as optional in next sections of MADR, I've updated this point to make it clear that this would be rather rare case

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+  to:
+  - targetRef:
+      kind: MeshHTTPRoute
+      name: route-to-backend
+    default:
+      http:
+        requestTimeout: 10s
+```
+
+All of these policies have `spec.targetRef` `Mesh` and these are producer policies so this configuration will be applied
```
lobkovilya, line 181: In this specific example the order of to-items in the resulting `MeshTimeout` policy will be decided by the alphabetical order of `metadata.name`. The order is important, because:
```yaml
to:
  - targetRef:
      kind: MeshService
      name: backend
    default:
      http:
        requestTimeout: 2s
        streamIdleTimeout: 1h
  - targetRef:
      kind: MeshHTTPRoute
      name: route-to-backend
    default:
      http:
        requestTimeout: 10s
```
and
```yaml
to:
  - targetRef:
      kind: MeshHTTPRoute
      name: route-to-backend
    default:
      http:
        requestTimeout: 10s
  - targetRef:
      kind: MeshService
      name: backend
    default:
      http:
        requestTimeout: 2s
        streamIdleTimeout: 1h
```
produce different Rules.

docs/madr/decisions/057-rework-mesh-routes-matching.md
michaelbeaumont: Right I just mean I don't think it's relevant, MeshRateLimit targetting MeshHTTPRoute doesn't make sense independent of this MADR AFAICT

docs/madr/decisions/057-rework-mesh-routes-matching.md
michaelbeaumont: > One suggestion here was to make spec.targetRef optional so most people wouldn't even know it exists

This is kind of the problem I have with the whole thing. We're basically admitting `spec.targetRef` is almost always going to be empty. So why does it have the top level place? Why not just have `spec.workloadTargetRef` that's optional and keep the most important info at `spec.targetRef`?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+
+To introduce this change we need to deprecate specifying `Mesh*Route` in `spec.targetRef`.
+
+Should we allow selecting MeshGateway in `spec.targetRef` when `Mesh*Route` is configured in `spec.to[].targetRef`? probably this 
+is obsolete as Route will select gateway, and we want to apply policy on a route. Is there a need to make it more specific than route? 
+
+### Affected policies
+
+At the moment we can only target `Mesh*Routes` in MeshTimeout policy. This MADR is based on MeshTimeout policy.
+Looking at what can be configured at Envoy route we can also apply this to our MeshRetry policy. Rate limit is also configurable on route,
```
lobkovilya, line 105: There is also MeshLoadBalancingStrategy that configures hashPolicy on routes. Should we add support for MeshHTTPRoute?

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
michaelbeaumont: Right I was thinking more like:

> The spec.to[].targetRef field defines rules that applies to outgoing traffic of proxies selected by spec.targetRef. The spec.from[].targetRef field defines rules that applies to incoming traffic of proxies selected by spec.targetRef.

Clearly this hasn't been updated since we introduced MeshHTTPRoute top level. But can we update it now?

Previously `spec.to[].targetRef` defined which traffic was affected by selecting which destinations are affected. This isn't _quite_ true anymore, at least not directly.
How do we explain this? Should we define this in terms of Envoy resources?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+
+- add possibility to put `Mesh*Routes` in `spec.to[].targetRef`
+- leave `Mesh*Routes` in top level targetRef as it is now
+
+### Add possibility to put Mesh*Routes in `spec.to[].targetRef`
+
+#### Advantages
+
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will pick outbounds
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies)
```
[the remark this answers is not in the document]
Automaat: on universal zone

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+- can be harder to understand because `spec.targetRef` can be MeshSubset (which is optional and probably will be rarely used)
+- yet another migration for users
+- unable to apply config per routes backendRef 
+
+### Leave Mesh*Routes in top level targetRef as it is now
+
+#### Advantages
+
+- can be simpler as you are always targeting all proxies to which the route applies
+- can target specific backendRef from route
```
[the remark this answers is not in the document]
Automaat: Oh, I didn't find this information in docs, so with this in mind this whole argument is invalid

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will pick outbounds
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies)
+- no need for specifying `spec.targetRef` when targeting `Mesh`
+- easier policy matching code
+
+#### Disadvantages
+
+- can be harder to understand because `spec.targetRef` can be MeshSubset (which is optional and probably will be rarely used)
+- yet another migration for users
+- unable to apply config per routes backendRef
```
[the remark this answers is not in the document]
Automaat: you are right this whole point does not make sense

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: why does it work like this? Shouldn't the two be the same? they target different resources, so why naming should be more important that resource they select? This is counter intuitive

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will pick outbounds
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies)
+- no need for specifying `spec.targetRef` when targeting `Mesh`
+- easier policy matching code
+
+#### Disadvantages
+
+- can be harder to understand because `spec.targetRef` can be MeshSubset (which is optional and probably will be rarely used)
+- yet another migration for users
```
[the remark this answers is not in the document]
Automaat: added do madr

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
Automaat: added to madr

docs/madr/decisions/057-rework-mesh-routes-matching.md
lobkovilya: I agree, the issue is covered here https://github.com/kumahq/kuma/issues/9151

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: so you want `spec.workloadTargetRef` to select entire proxies, like now `spec.targetRef` does in this madr?

But what about MeshService? should this be still in `spec.to[].targetRef` or as route in `spec.targetRef`? What is the difference here?

michaelbeaumont: In my ideal world it "should" be in `spec.targetRef`. The `spec.workloadTargetRef` is also 99% of the time going to be either Mesh or namespace. The most interesting part of the policy is the outbound it affects.

Automaat: Got it, I will wrap this up and present this as another option to consider, also I will try to present some use cases with examples

michaelbeaumont: >  it doesn't change the way MeshHTTPRoute is used

it would make this MADR unnecessary right? MeshHTTPRoute is already in top level

> just syntactical changes

so this MADR is also just syntactical changes then

lobkovilya: The change in this MADR is not syntactical. We change the semantic of `Policy <-> MeshHTTPRoute` reference from:
```
Policy targets MeshHTTPRoute and because of that targets all the places when the route is applied.
```
to
```
Policy targets the place where MeshHTTPRoute is applied (same route can have different confs on different DPPs).
```

> I don't think it's syntactical. This is about making the semantics of spec.targetRef and by extension API the most useful and user friendly it can be

I don't know, you can kind of convert one to another by using a template:
```yaml
targetRef: ${{ policy.to[0].targetRef }}
conf: ${{ policy.to[0].conf }}
workloadTargetRef: ${{ policy.targetRef }}
```

michaelbeaumont: First of all I think you're forgetting about what would happen to `from`.

> Policy targets MeshHTTPRoute and because of that targets all the places when the route is applied.
> Policy targets the place where MeshHTTPRoute is applied (same route can have different confs on different DPPs).

I think this is just playing with "targeting". Let's be very honest. We're always "targeting" two things. That's why we have two `targetRefs`.

We're always "targeting the place where MeshHTTPRoute is applied" because we're targeting the MeshHTTPRoute, it's just in `to`.

You can also mechanically translate the two options in this MADR, can't you?
`spec.to[0].targetRef: ${{ spec.targetRef }}`

All you're doing is essentially adding the ability to define `workloadTargetRef`, except it's called `targetRef`.

Anyway I don't think any of that matters either way. We're playing around with the structure of policies already. We're literally defining the API. I am unconvinced that this change actually improves anything when it comes to UX.

michaelbeaumont: Put a different way, I think it would be more of a UX improvement if we put important things in `spec.targetRef` everywhere than if we put MeshHTTPRoute in `spec.to[].targetRef`

michaelbeaumont: I see

```
## Context and Problem Statement

At the moment when targeting `Mesh*Route` you need to put it in topLevel targetRef. 
We want to iterate over policy design to fully utilize the fact that we are referencing real objects.
```

We're already referencing real objects as far as I can tell?
What is the actual problem we're solving?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,417 @@
...
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will pick outbounds
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies, on universal)
+- no need for specifying `spec.targetRef` when targeting `Mesh`
+- easier policy matching code
+
+#### Disadvantages
+
+- can be harder to understand because `spec.targetRef` can be MeshSubset (which is optional and probably will rarely be used)
+- yet another migration for users (only for system policies as for namespaced policies this is already not allowed) 
+
```
michaelbeaumont, line 34:
```suggestion
- mixing outbounds and routes in `spec.to[].targetRef`

```

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,741 @@
...
+- can be harder to understand because `spec.targetRef` can be MeshSubset (which is optional and probably will rarely be used)
+- yet another migration for users (only for system policies as for namespaced policies this is already not allowed) 
+- mixing outbounds and routes in `spec.to[].targetRef`
+- hides the most often most important information under `to[]` (only in certain cases as most of the time policy will target Mesh and will have only `spec.to[].targetRef`)  
+
+### Leave Mesh*Routes in top level targetRef as it is now and introduce `spec.workloadTargetRef`
+
+#### Advantages
+
+- can be simpler with clear distinction of selecting proxies in `spec.workloadTargetRef` and other resources in `spec.targetRef`
```
lobkovilya, line 44: I'd say we already have a clear distinction. I guess the advantage is having the most interesting part of the policy on the top instead of being nested in `to[].targetRef`

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,741 @@
...
+#### Producer/consumer model for targeting Mesh*Route in other policies
+
+It is worth to mention that rate limit would not work with our producer/consumer model. In our model consumer always has more priority
+but consumer should not override producer rate limit as owner of the service should know best, the amount of traffic it could handle.
+
+For MeshRetry our producer/consumer makes sense since producer config is a suggestion from owner of the service. If consumer
+wants to override it because they don't want to retry, they just want to fail fast or the want to retry on bigger number of errors
+they can configure it by applying consumer policy.
+
+### Merging and applying configurations
```
lobkovilya, line 439: What about the case when MeshHTTPRoute has multiple to-items to different MSs?
```yaml
kind: MeshHTTPRoute
metadata:
  name: route-1
spec:
  to:
    - targetRef:
        kind: MeshService
        name: backend
      rules: [...]
    - targetRef:
        kind: MeshService
        name: frontend
      rules: [...]
```

Can I reference `route-1` from the policy in this case? Does it apply to both outbounds?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,741 @@
...
+- gives more flexibility as you can target multiple outbounds in single policy. Reducing number of consumer policies
+- more consistent with current approach
+- easier policy matching code
+
+#### Disadvantages
+
+- can be harder to understand because `spec.targetRef` can be MeshSubset (which is optional and probably will rarely be used)
+- yet another migration for users (only for system policies as for namespaced policies this is already not allowed) 
+- mixing outbounds and routes in `spec.to[].targetRef`
+- hides the most often most important information under `to[]` (only in certain cases as most of the time policy will target Mesh and will have only `spec.to[].targetRef`)
```
michaelbeaumont, line 38: it's _always_ hiding it in the sense that there's a "useless" `to[]` in front of it. it's noise. the fact that most of the time the policy will target Mesh is what makes it worse

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,741 @@
...
+1. add possibility to put `Mesh*Routes` in `spec.to[].targetRef` -> `spec.targerRef` selects proxies, `spec.from[].targetRef` selects inbounds, `spec.to[].targetRef` selects outbounds.
+2. leave `Mesh*Routes` in top level targetRef as it is now, and introduce `spec.workloadTargetRef` to match whole proxies
+
+### Add possibility to put Mesh*Routes in `spec.to[].targetRef`
+
+#### Advantages
+
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will select outgoing traffic
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies, on universal)
+- no need for specifying `spec.targetRef` when targeting `Mesh`, simplifies policy
```
michaelbeaumont, line 28: Is this referring to specifying `spec.to[].targetRef.kind: Mesh`?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,741 @@
...
+
+### Leave Mesh*Routes in top level targetRef as it is now and introduce `spec.workloadTargetRef`
+
+#### Advantages
+
+- can be simpler with clear distinction of selecting proxies in `spec.workloadTargetRef` and other resources in `spec.targetRef`
+
+#### Disadvantages
+
+- mixing proxies, outbounds and routes in `spec.targetRef`
```
michaelbeaumont, line 48: Hmm when does it mix proxies?

Also can we explain how this is a disadvantage? it's the entire point. This is using the rule "`spec.targetRef` refers to a proxy" to justify not changing _that_ rule

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,741 @@
...
+- when `Mesh*Route` is in `spec.targetRef` targetRef we still have `spec.to[].targetRef` section which can be misleading as route already points to some outbound
```
michaelbeaumont, line 49: In fact this would make `spec.to` obsolete. Otherwise we can just make it optional like the other option suggests doing for `spec.targetRef`, so it's not really a disadvantage.

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,741 @@
...
+- cannot configure multiple outbounds in the same policy, forcing users to create and later manage multiple policies, which can be problematic with consumer policies
+- MeshService in topLevel can be misleading. Is it targeting all proxies belonging to MeshService or outbound to MeshService?
```
michaelbeaumont, line 51: Hmm I don't think it's misleading anymore. `MeshService` (the new resource) represents a destination. These semantics are why `HTTPRoute` uses `Service` as `spec.parentRef`.

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,741 @@
...
+- need to introduce entirely new `spec.workloadTargetRef` section
```
michaelbeaumont, line 52: In what sense is this a disadvantage? or is this referring to the migration being difficult?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,741 @@
...
+
+#### Advantages
+
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will select outgoing traffic
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies, on universal)
+- no need for specifying `spec.targetRef` when targeting `Mesh`, simplifies policy
+- gives more flexibility as you can target multiple outbounds in single policy. Reducing number of consumer policies
+- more consistent with current approach
+- easier policy matching code
+
```
michaelbeaumont, line 32:
```suggestion
- least problematic migration

```

docs/madr/decisions/057-rework-mesh-routes-matching.md
michaelbeaumont: What we should also be asking is "how does `spec.from` work in this option"? The answer being that we get rid of `from` as well. And we have `MeshTrafficPermission` for example target `spec.targetRef.kind: WorkloadIdentity`.

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: this point does not make sense, as in both cases you can omit either `spec.targetRef` or `spec.workloadRef` when targeting Mesh

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: this also does not apply any more if we add `workloadTargetRef`

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: right, removed this point as `spec.to[]` can be omitted in that case

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: I guess it is not really a disadvantage for users, I was looking at it from work needed to introduce it perspective, removed

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: rewritten this point

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: rewritten this point, it should be ok now

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: I've updated this point to make it clear that this could be problematic for existing users.

Automaat: As for the second point I am not sure if we want to cover this here, like at the begging I thought that decision was already made, and I am just writing MADR on how to introduce this change, now this MADR is changing into fundamental redesign of policy API. I don't know what to focus on any more

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,738 @@
...
+- more consistent with current approach
+- easier policy matching code
+- least problematic migration
+
+#### Disadvantages
+
+- can be harder to understand because `spec.targetRef` can be MeshSubset (which is optional and probably will rarely be used)
+- yet another migration for users (only for system policies as for namespaced policies this is already not allowed) 
+- mixing outbounds and routes in `spec.to[].targetRef`
+- hides the most often most important information under `to[]` adding noise when targeting Mesh in `spec.targetRef`
```
michaelbeaumont, line 38:
```suggestion
- hides the most often most important information under `to[]` adding noise when targeting Mesh in `spec.targetRef`
- is it consistent in the case of MeshGateways? MeshHTTPRoutes aren't on outbound listeners in the case of MeshGateways, they're on inbound listeners
```

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: added section about MeshLoadBalancingStrategy

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will pick outbounds
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies)
+- no need for specifying `spec.targetRef` when targeting `Mesh`
+- easier policy matching code
+
+#### Disadvantages
+
+- can be harder to understand because `spec.targetRef` can be MeshSubset 
+- yet another migration for users
```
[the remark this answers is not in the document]
Automaat: it's only for system policies, it is already in the MADR, I think you were looking at the old commit

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+      default:
+        http:
+          requestTimeout: 5s
+          streamIdleTimeout: 1h
+```
+
+To introduce this change we need to deprecate specifying `Mesh*Route` in `spec.targetRef`.
+
+Should we allow selecting MeshGateway in `spec.targetRef` when `Mesh*Route` is configured in `spec.to[].targetRef`? probably this 
+is obsolete as Route will select gateway, and we want to apply policy on a route. Is there a need to make it more specific than route?
```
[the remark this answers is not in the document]
Automaat: added example

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+### Affected policies
+
+At the moment we can only target `Mesh*Routes` in MeshTimeout policy. This MADR is based on MeshTimeout policy.
+Looking at what can be configured at Envoy route we can also apply this to our MeshRetry policy. Rate limit is also configurable on route,
+but we are configuring rate limit on inbound traffic, which does not apply to this MADR. 
+
+#### Producer/consumer model for targeting Mesh*Route in other policies
+
+It is worth to mention that rate limit would not work with our producer/consumer model. In our model consumer always has more priority
+but consumer should not override producer rate limit as owner of the service should know best, the amount of traffic it could handle.
```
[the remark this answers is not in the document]
Automaat: updated

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+
+For MeshRetry our producer/consumer makes sense since producer config is a suggestion from owner of the service. If consumer
+wants to override it because they don't want to retry, they just want to fail fast or the want to retry on bigger number of errors
+they can configure it by applying consumer policy.
```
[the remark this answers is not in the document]
Automaat: added example

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+  labels:
+    kuma.io/mesh: default
+spec:
+  to:
+  - targetRef:
+      kind: MeshHTTPRoute
+      name: route-to-backend
+    default:
+      http:
+        requestTimeout: 10s
```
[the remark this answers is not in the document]
Automaat: changed all examples as requested

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+    tags:
+      k8s.kuma.io/namespace: some-other-ns
+  to:
+    - targetRef:
+        kind: MeshHTTPRoute
+        name: route-to-backend
+```
+
+After applying this two policies timeout won't be applied to any proxy, since `Mesh*Route` and MeshTimeout is applied
+on different subset of proxies. We should detect this situation and add warning in Inspect API.
```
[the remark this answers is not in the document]
Automaat: updated

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,533 @@
...
+  to:
+  - targetRef:
+      kind: MeshHTTPRoute
+      name: route-to-backend
+    default:
+      http:
+        requestTimeout: 10s
+```
+
+will take precedence and will override timeouts for specific MeshService.
```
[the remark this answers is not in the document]
Automaat: I think in ideal case when configuring route we should be able only to configure stuff from Envoy route. And when merging policies, we should take this into account.

docs/madr/decisions/057-rework-mesh-routes-matching.md
michaelbeaumont: this has been integrated into the MADR

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
michaelbeaumont: updated

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: So this is problematic, I think that in perfect world we should merge confs for selected resource since different resources correspond with different Envoy config. So I think when merging we should not treat Mesh
*Route as subset of MeshService and merge conf with MeshService so for policy:

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTimeout
metadata:
  name: timeout-on-backend
  namespace: backend-ns
  labels:
    kuma.io/mesh: default
spec:
  to:
  - targetRef:
      kind: MeshService
      name: backend
    default:
      http:
        streamIdleTimeout: 1h
        requestTimeout: 5s
   - targetRef:
        kind: MeshService
        name: frontend
     default:
       http:
         streamIdleTimeout: 2h
         requestTimeout: 10s
  - targetRef:
      kind: MeshHTTPRoute
      name: slow-backend-route
      namespace: backend-ns
    default:
      http:
        requestTimeout: 15s
```

after merging, we should have:

```yaml
resourceRules:
  backend.backend-ns:
    conf:
      http:
        streamIdleTimeout: 1h
        requestTimeout: 5s
  frontend.backend-ns:
    conf:
      http:
        streamIdleTimeout: 2h
        requestTimeout: 10s
  slow-backend-route.backend-ns:
    conf:
      http:
        requestTimeout: 15s
```

And when applying configuration, we should take this into account. So we will have two outbounds for each MeshService, and they will share a route. So take all of this information into account, and combine config for MeshService with Mesh*Route for given outbound pointing to specific service

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
Automaat: I think this is related to: https://github.com/kumahq/kuma/pull/10858#discussion_r1680742225

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,826 @@
...
+- more consistent with current approach
+- easier policy matching code
+- least problematic migration
+
+#### Disadvantages
+
+- can be harder to understand because `spec.targetRef` can be MeshSubset (which is optional and probably will rarely be used)
+- yet another migration for users (only for system policies as for namespaced policies this is already not allowed) 
+- mixing outbounds and routes in `spec.to[].targetRef`
+- hides the most often most important information under `to[]` adding noise when targeting Mesh in `spec.targetRef`
```
slonka, line 38:
```suggestion
- often hides the most important information under `to[]`, adding noise when targeting Mesh in `spec.targetRef`
```

what is the noise in question?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,826 @@
...
+- is it consistent in the case of MeshGateways? MeshHTTPRoutes aren't on outbound listeners in the case of MeshGateways, they're on inbound listeners
```
slonka, line 39: Do we care? `To` / `from` don't have to always map to `outbound` / `inbound`. A route always routes "to" something right?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,826 @@
...
+
+### Leave Mesh\*Routes in top level targetRef as it is now and refine semantics of `spec.targetRef`
+
+Instead of having `targetRef` always refer to a workload,
+instead `targetRef` selects some part of the Envoy configuration of a workload,
```
slonka, line 44:
```suggestion
`targetRef` selects some part of the Envoy configuration of a workload,
```

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,826 @@
...
+	- `MeshService` _is_ a destination so `spec.targetRef.kind: MeshService` targets the configuration for a given destination
+	- `WorkloadIdentity` (obviously doesn't exist yet but we want to change how we manage identity)
+      would represent an identity so fits as `spec.targetRef` for e.g. `MeshTrafficPermission`
+
+#### Disadvantages:
+
+- `targetRef` is a single field, so it's not easy to share configuration
+  in a single resource to avoid duplication.
+  Though the very WIP policy attachment GEPs do propose a `targetRefs` field
+- requires a huge migration
```
slonka, line 70: can you add more details? in top level targetRef we only have `Mesh*Route` right? It's not like we have every policy there, right?

slonka (commented):
It feels like this https://github.com/kumahq/kuma/pull/10858/files#diff-95c559a0f6155d986d45f294106f1a5a789e99bd0c0c93dae4405724efc76f3cR47-R49 is the most compelling for me.

Also added a couple of nitpicks.

docs/madr/decisions/057-rework-mesh-routes-matching.md
michaelbeaumont: The noise is the `.to[].targetRef` structure, as opposed to `spec.targetRef`

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,826 @@
...
+
+### Leave Mesh\*Routes in top level targetRef as it is now and refine semantics of `spec.targetRef`
+
+Instead of having `targetRef` always refer to a workload,
+instead `targetRef` selects some part of the Envoy configuration of a workload,
+where that part is represented by a real Kuma resource.
+
+In fact `targetRef` is the wrong concept for selecting workloads,
+since "targetRef refers to a workload" means we aren't targeting
+an arbitrary `Kind` of resource but instead always a `Dataplane`.
```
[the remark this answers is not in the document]
michaelbeaumont: Yea I liked this but in fact I kind of glossed over the gateway aspect of this. I had the `MeshHTTPRoute` case too far in the front of my mind and in general we would need a place to select whether it applies to gateways or mesh services, so it can't really be as simple as labels I guess.

docs/madr/decisions/057-rework-mesh-routes-matching.md
michaelbeaumont: > in top level targetRef we only have Mesh*Route right? 

Yes, and the problem is thus all the outbound policies would need to change, where we have `spec.to[].targetRef.kind: MeshService`.

docs/madr/decisions/057-rework-mesh-routes-matching.md
michaelbeaumont: Sure the route routes "to" something but that's not what `to` is referring to, it's targetting traffic "to" the route. I.e. for a normal mesh service, when we have `.to[].targetRef.kind: MeshService` a policy targets traffic "to" a service, when we have `.to[].targetRef.kind: MeshHTTPRoute` it targets traffic "to" the route. But it doesn't make sense for a gateway to say a policy targets traffic "to" a route, the traffic is going "to" a gateway listener but the routes determine where traffic goes "to". A gateway listener is an _inbound_ listener.

The point is that it's not really useful or meaningful to talk about "to" or "from" a route, it's not useful to _also_ have this on the policy. The route resource itself determines whether it's "from" or "to". We're just targeting the route.

And my claim and motivation for this variant is that it's not even useful _ever_ to have "to" and "from". Given a policy kind `SomePolicy` and a targeted kind `Target`, you can never have both `to[].kind.targetRef.kind: Target` and `from[].kind.targetRef.kind: Target`. The semantics of `SomePolicy` and `Target` uniquely determine whether you use `to` or `from`, so you might as well have `spec.targetRef.kind: Target`.

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,902 @@
...
+  - targetRef:
+      kind: MeshService
+      name: backend
+      namespace: backend-ns
+    default:
+      http:
+        requestTimeout: 10s
+```
+
+```yaml
```
lukidzi, line 140: I know that the MADR is about targeting `to` but would be nice to have an example that shows how to configure a Timeout on the inbound of the specific dataplane since there is no from section. Is there any change or should it work as it is?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,902 @@
...
+      default:
+        http:
+          numRetries: 3
+          backOff:
+            baseInterval: 10ms
+            maxInterval: 1s
+          retryOn:
+            - "500"
+---
+# Consumer policy with override to retry on all 5xx errors
```
lukidzi, line 535: Is it a valid example? It looks to me like a producer policy,

Automaat: righ, the example was incorrect, I've updated it

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: I've updated mard as discussed

docs/madr/decisions/057-rework-mesh-routes-matching.md
jakubdyszkiewicz: Is it that big of the noise to justify yet another big migration?
I'm not even talking about code, I'm talking about "docs", "explaining it to users", "deprecating", asking for a change etc.

We are talking about
```
spec:
  targetRef:
    kind: MeshService
    name: backend
```
vs
```
spec:
  to:
  - targetRef:
      kind: MeshService
      name: backend
```

How is it really better than the second option? What's the value here?

I'm reading advantages

> - the workloads targeted by the policy is often "all workloads" or can be
  inferred from the namespace so we kind of waste the top level, "special" `spec.targetRef`

what does "waste" mean here? I don't think I get this argument

> tends to be less noisy as we move most interesting part of the policy from `to` /`from`section

are users confused by `to`? Is it difficult to construct the policies with `to` and it will be much easier if we were to go with the first option?

> - Doesn't force every potential resource into a "to" or a "from", instead using the semantics of the `kind` itself
	- `HTTPRoute` is a route, regardless of whether it's inbound or outbound

but HTTPRoute is "outbound" route, no? Wouldn't we have a separate resource for inbound route?

> WorkloadIdentity

left the comment in another place about it.

I miss inbound policies in this discussion.

If we were starting this from scratch maybe that would be a _slightly_ better way, but is it worth the effort now?

michaelbeaumont: > Wouldn't we have a separate resource for inbound route?

We have the resource on gateways where it definitely doesn't fit the definition of "outbound route".

No I don't think it's worth the effort, I just want it documented and the drawbacks and implications of the current method understood.

michaelbeaumont: > what does "waste" mean here? I don't think I get this argument

In the context of gateway API spec.targetRef has a special meaning. Our API relegates this to "almost never used" and adds a _different_ targetRef to "100% always used". Thus we have wasted spec.targetRef

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,897 @@
...
+#### Advantages:
+
+- the workloads targeted by the policy is often "all workloads" or can be
+  inferred from the namespace so we kind of waste the top level, "special" `spec.targetRef`
+- tends to be less noisy as we move most interesting part of the policy from `to` /`from`section
+- Doesn't force every potential resource into a "to" or a "from", instead using the semantics of the `kind` itself
+	- `HTTPRoute` is a route, regardless of whether it's inbound or outbound
+	- `MeshService` _is_ a destination so `spec.targetRef.kind: MeshService` targets the configuration for a given destination
+	- `WorkloadIdentity` (obviously doesn't exist yet but we want to change how we manage identity)
+      would represent an identity so fits as `spec.targetRef` for e.g. `MeshTrafficPermission`
```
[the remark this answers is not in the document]
michaelbeaumont: No, `from` would disappear. You'd have the mesh traffic permission apply to traffic to the namespace and the WorkloadIdentity would define which sources are allowed.

docs/madr/decisions/057-rework-mesh-routes-matching.md
michaelbeaumont: This would require a different resource

michaelbeaumont: I added a note about this to the option, we'd need a different policy `InboundTimeout` or `OutboundTimeout`

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,897 @@
...
+2. leave `Mesh*Routes` in top level targetRef as it is now, improve the semantics of `spec.targetRef` and introduce `spec.workloadSelector` to select whole proxies
+
+### Add possibility to put Mesh*Routes in `spec.to[].targetRef`
+
+#### Advantages
+
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will select outgoing traffic
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies, on universal)
+- gives more flexibility as you can target multiple outbounds in single policy. Reducing number of consumer policies
+- more consistent with current approach
```
michaelbeaumont, line 29: > - more consistent with current approach

First of all, the current approach is to have MeshHTTPRoute in `spec.targetRef`. The current approach is the most consistent with the current approach.

And otherwise how exactly is it more consistent? I go back and forth on this. `MeshHTTPRoute` is the _first_ `kind` that is itself a policy. So what exactly is it more consistent with?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,897 @@
...
+## Considered options
+
+1. add possibility to put `Mesh*Routes` in `spec.to[].targetRef` -> `spec.targetRef` selects proxies, `spec.from[].targetRef` selects inbounds, `spec.to[].targetRef` selects outbounds.
+2. leave `Mesh*Routes` in top level targetRef as it is now, improve the semantics of `spec.targetRef` and introduce `spec.workloadSelector` to select whole proxies
+
+### Add possibility to put Mesh*Routes in `spec.to[].targetRef`
+
+#### Advantages
+
+- simplifying policy api as `spec.targetRef` will always pick proxy and `spec.to[].targetRef` section will select outgoing traffic
```
michaelbeaumont, line 26: > `spec.to[].targetRef` section will select outgoing traffic

This isn't really the whole picture. The `to` section would also pick the proxy.

Current:
- `spec.targetRef` picks both the proxy and the outgoing traffic
- `spec.to[].targetRef` is always `Mesh`

Proposal:
- `spec.targetRef` picks the proxy
- `spec.to[].targetRef` picks both the proxy and the outbound

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
michaelbeaumont: added

michaelbeaumont (commented):
Added the positive/negative of the decided approach

michaelbeaumont (commented):
Can we please address https://github.com/kumahq/kuma/issues/3250#issuecomment-2188491271 in this MADR? Maybe I missed it

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: right, this does not make sense. I've removed it

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: removed the point

Automaat:
> Can we please address [#3250 (comment)](https://github.com/kumahq/kuma/issues/3250#issuecomment-2188491271) in this MADR? Maybe I missed it

@michaelbeaumont If if get the issue correctly, this is about configuring MTP per route. I think this is a topic for separate MADR, as here we are discussing outbound routes, and MTP should be configured on inbounds, and can use different approach

michaelbeaumont:
> @michaelbeaumont If if get the issue correctly, this is about configuring MTP per route. I think this is a topic for separate MADR, as here we are discussing outbound routes, and MTP should be configured on inbounds, and can use different approach

OK then we can comment in that issue I guess. Or ask @lahabana what he was thinking.

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
jakubdyszkiewicz: I see. However
1) I though we want to diverge to using real resources in targetRef. WorkloadIdentity won't be a real resource AFAIK (correct me if I'm wrong).
2) Creating X resources for each service is kind of annoying. I'd like to have list of services that I allow/deny

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+
+Technical story: https://github.com/kumahq/kuma/issues/10247
+
+## Context and Problem Statement
+
+This MADR is focused to targeting outbound routes configured on client proxy. 
+
+At the moment when targeting `Mesh*Route` you need to put it in topLevel targetRef. 
+We want to iterate over policy design and move `Mesh*Route` to `spec.to[].targetRef` section follow model where `spec.targetRef`
+selects whole proxies, `spec.from[].targetRef` selects inbounds and `spec.to[].targetRef` selects outbounds.
```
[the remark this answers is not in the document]
Automaat: changed

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+- ability to set different confs on different proxy for the same route (overriding default conf for route without producer/consumer policies, on universal)
+- gives more flexibility as you can target multiple outbounds in single policy. Reducing number of consumer policies
+- easier policy matching code
+- least problematic migration
+
+#### Disadvantages
+
+- can be harder to understand because `spec.targetRef` can be MeshSubset (which is optional and probably will rarely be used)
+- which sidecars are actually configured depends on both `spec.targetRef` of the policy _and_ the `spec.targetRef` of the `MeshHTTPRoute`.
+- yet another migration for users (only for system policies as for namespaced policies this is already not allowed)
```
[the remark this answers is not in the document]
Automaat: updated

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+metadata:
+  name: producer-retry
+  namespace: backend-ns
+  labels:
+    kuma.io/mesh: default
+spec:
+  to:
+    - targetRef:
+        kind: MeshHTTPRoute
+        name: backend-route
```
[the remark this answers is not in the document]
Automaat: added comment

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
michaelbeaumont: 1) I suppose whether it's real or not remains to be seen. Maybe it'd be ServiceAccount if that's our how identity works

2) maybe it's enough to use Namespace. Or maybe the structure of WorkloadIdentity might allow for some kind of multiple selection or maybe we just need the upstream proposed `spec.targetRefs`

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+
+When referencing Mesh*Route we should validate policy configuration to make sure user is applying only configuration 
+that can be applied on route.
+
+### Affected policies
+
+At the moment we can only target `Mesh*Routes` in MeshTimeout policy. This MADR is based on MeshTimeout policy.
+Looking at what can be configured at Envoy route we can also apply this to our MeshRetry policy. Rate limit is also configurable on route,
+but we are configuring rate limit on inbound traffic, which does not apply to this MADR. `MeshLoadBalancingStrategy` also
+configures hashing per route and we should probably add support for routes in the future
```
lahabana, line 491: This paragraph is hard to follow. I'd go with something more like a list.

Here are the policies that may be configurable on routes:

- `MeshTimeout`: can be configured per route. However, not all parameters make sense (.e.g: tcp timeouts). Validation will be added to reject policies with `spec.to[].targetRef.kind: MeshHTTPRoute` which configs that can't be configured. What about inspect policies and merging? Will we see TCP timeouts?
- `MeshLoadBalancingStrategy`: can be configured per route. Like `MeshTimeout` all configurations do not make sense and will need to be validated.
- `MeshRetry` : can be configured per route. Like `MeshTimeout` all configurations do not make sense and will need to be validated.
- `MeshRateLimit`: while it's possible to set rate limit on a route. We currently only set rate limits on the inbound so supporting `spec.to[].target.kind: MeshHTTPRoute` doesn't make sense.

What about:
- `MeshAccessLog`: Seems like we should be able to support it just for a route no?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+      default:
+        http:
+          retryOn:
+            - "5xx"
+```
+
+For MeshLoadBalancingStrategy adding producer/consumer model can be problematic, for load balancing and hashing. Since this
+is configured on outbound it falls into producer/consumer policy model. But who should have more power on deciding hashing rules?
+I think it should be producer who is an owner of service who decides how to apply this configuration which is not compliant
+with our consumer/producer model.
```
lahabana, line 547: I don't see why the consumer can't control everything?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+        namespace: backend-ns
+      default:
+        http:
+          retryOn:
+            - "5xx"
+```
+
+For MeshLoadBalancingStrategy adding producer/consumer model can be problematic, for load balancing and hashing. Since this
+is configured on outbound it falls into producer/consumer policy model. But who should have more power on deciding hashing rules?
+I think it should be producer who is an owner of service who decides how to apply this configuration which is not compliant
```
lahabana, line 546: If you think you need to justify why you think so

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+When merging we will merge `$conf1` and `$conf2` for MeshService with `$conf3` for `Mesh*Route`. 
+
+```yaml
+resourceRules:
+  backend.backend-ns: 
+    conf: $conf1
+  frontend.backend-ns:
+    conf: $conf2
+  backend-route.backend-ns:
+    conf: merge($conf1, $conf2, $conf3)
```
lahabana, line 718: Isn't there an index thing for this @lobkovilya ?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+```
+
+Configuration priority (from most to least important): 
+1. consumer policy targeting `Mesh*Route`
+2. producer policy targeting `Mesh*Route`
+3. consumer policy targeting MeshService
+4. producer policy targeting MeshService
+
+We always prioritize policies targeting `Mesh*Route` than the one targeting MeshService because route can have more specific
+requirements than a whole MeshService
```
lahabana, line 794: How do you prioritise with a route on the consumer side vs on the producer side?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+```
+
+### Backward compatibility and migration
+
+First we need to start by deprecating putting `Mesh*Route` in `spec.targetRef`.
+At the moment we are building `Rules` object with Subset for matching. This will change with implementation of 
+`policy matching with real resources` MADR. For real resources we will build `ResourceRule` object. To smoothen the transition
+we can still build old `Rules` for old policies and new `ResourceRule` for newly created policies with `Mesh*Route` in `spec.to[].targetRef`.
+At the policy level we can prioritize new `ResourceRule` over old `Rules`. Thanks to this new way of targeting `Mesh*Routes`
+will take precedence, so that users can easily switch to new targeting mechanism without any downtime.
```
lahabana, line 895: What about the deprecation around number of entries in `to` for `Mesh*Route` ?

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+    kuma.io/mesh: default
+spec:
+  targetRef:
+    kind: MeshHTTPRoute
+    name: slow-backend-route
+    namespace: backend-ns
+  default:
+    http:
+      requestTimeout: 15s
+```
```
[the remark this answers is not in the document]
Automaat: added info about translation in madr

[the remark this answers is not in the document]
Automaat: > Btw how does that solve workloadSelector on the MeshHTTPRoute and on the MeshTimeout?

@lahabana can you extend your question? I don't think I get it

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,895 @@
...
+
+* We have a relatively small migration
+* a policy can target a subset of the sidecars that a `MeshHTTPRoute` targets
+  * this is especially important on universal where the zone isn't granular enough and there are no namespaces to implicitly limit the effects of a policy
+* easy to target multiple `MeshHTTPRoutes` with a policy
+
+### Negative Consequences
+
+* `spec.targetRef` **must be combined** with `spec.to[].targetRef` to figure out which sidecars are actually affected since policies referenced in `spec.to[].targetRef` also select sidecars.
+* `spec.targetRef`, given that it's likely to be empty, is relegated to not being very useful on its own for figuring out what a policy affects in the policy attachment sense
```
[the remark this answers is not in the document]
Automaat: I am not sure, issue is referencing this MADR, and most information is here, so I don't think we need this

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: rewritten this section as suggested. As for:
> What about inspect policies and merging? Will we see TCP timeouts?

@lobkovilya correctly me if I am wrong but with current merging we will get all the information not only route specific, right?

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: Added example to explain this

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: Added info about deprecating, and using consumer/producer policcies to smoothen it out

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: I've updated this information and updated merging order with consumer/producer routes

docs/madr/decisions/057-rework-mesh-routes-matching.md
lobkovilya: Yes, with the current implementation of ResourceRules we consider MeshHTTPRoutes included in MeshServices, so the final configuration for route is a product of merging:
```
meshservice:mesh/mesh-1:name/backend:
  conf: $ms_specific_conf
meshhttproute:mesh/mesh-1:name/route-to-backend:
  conf: merge($ms_specific_conf, $route_specific_conf)
```

docs/madr/decisions/057-rework-mesh-routes-matching.md
lobkovilya: You mean the [ruleIndex](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/060-policy-matching-with-real-resources.md?plain=1#L77) in origin?

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: After discussion with @lobkovilya we've decided that we will change merging mechanism to not merge confs for different resources types, as it was now. So we won't include Mesh or MeshService conf in Mesh*Route conf. So in case of inspecting policies we will see only config for a route without defaults from Mesh or MeshService, as they could not be applicable on route itself

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,922 @@
...
+
+It's clear that it's very important when applying a policy, that we often need
+to apply the policy based on its source _and_ its destination.
+We always end up having to choose one of these to put in `spec.targetRef`. In
+the `MeshTimeout.to` case it's the source of the traffic. With
+`MeshTrafficPermission` it's the destination.
+
+But it could be argued that this distinction is an implementation detail. The user doesn't need
+to know which sidecar is used to apply the policy, as long as its effects are
+as desired.
```
lahabana, line 110: I think this isn't true. I think we departed from `source` `destination` for this reason, assuming things were transparent was wrong.

docs/madr/decisions/057-rework-mesh-routes-matching.md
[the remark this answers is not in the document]
lahabana: If you have workloadSelector on both MeshHTTPRoute and MeshTimeout how do you check these are not exclusive (I remain convinced it doesn't matter, just like services selector can select no pods)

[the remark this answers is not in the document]
Automaat: I also think that it does not matter, and will be same as with MeshSubset now, we ignore it

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,922 @@
...
+```
+
+For MeshLoadBalancingStrategy adding producer/consumer model can be problematic, for load balancing and hashing. Since this
+is configured on outbound it falls into producer/consumer policy model. But who should have more power on deciding hashing rules?
+It should be producer who is an owner of service who decides how to apply this configuration which is not compliant
+with our consumer/producer model. 
+
+For example, we could have a service that each instance is responsible for handling specific clients. We can use `RingHash`
+to make sure that specific clients requests are processed by the same instance. This should be configured in producer policy. 
+Changing this at consumer level can influence correctness of computations in our service.
```
lahabana, line 558: I think it's fine. Consumers can do stupid things but it doesn't matter becasue they only impact themselves.

lahabana (approved):

docs/madr/decisions/057-rework-mesh-routes-matching.md
```diff
@@ -0,0 +1,922 @@
...
+As described in targeting real resources MADR, while merging policies we will create `ResourceRules map[UniqueResourceKey]ResourceRule` 
+field that holds configuration for a given resource identified by `UniqueResourceKey`. So when user creates policy that has 
+real resource in `spec.to[].targetRef` we will create confs by resources: (this is configuration that will be applied to selected dataplane)
+
+```yaml
+resourceRules:
+  backend.backend-ns:
+    conf: $conf1
+  backend-route.backend-ns:
+    conf: merge($producer_conf, $consumer_conf)
```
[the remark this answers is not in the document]
Automaat: I won't be taken into account when merging. policy plugin should make that decision based on which outbound it is configuring

docs/madr/decisions/057-rework-mesh-routes-matching.md
Automaat: updated

PR #10860, opened 2024-07-10
docs(MADR): add MeshService migration

### LINKED ISSUE #9708: MADR for migration to MeshService
### Description

Create MADR for migration to MeshService

michaelbeaumont:
Closes #9708 

### Checklist prior to review


docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,137 @@
...
+If for example we have a policy targeted at:
+
+```
+spec:
+  targetRef:
+    kind: MeshService
+    name: demo-app_kuma-demo_svc_5000
+```
+
+when switching over to MeshService, this policy will no longer apply.
```
Icarus9913, line 25: Do we need to put the right use case here?
And also could we add webhook validating here to forbid the old use case and give us the warning to use the latest one?

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,137 @@
...
+### Cross-zone traffic
+
+One of the fundamental changes with `MeshService` is that traffic to a service
+is no longer load balanced across all zones.
+
+We need to take measures to prevent this change from happening on upgrade.
+
+In particular with Kubernetes, it's important to keep in mind that after
+MeshService is enabled, _any_ requests to the Kubernetes IP/via Kube DNS
+go to the same outbound as to MeshService hostnames.
```
Icarus9913, line 43: a little bit unclear with `Kubernetes IP/via Kube DNS`

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### Status quo
+
+- On Kubernetes there are two outbound listeners:
+  - Traffic to the Kube DNS name/Cluster IP is load-balanced cross-zone
```
Icarus9913, line 48: How about `KubeDNS-name/ClusterIP` ?

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,137 @@
...
+    Users would need to manually switch over to local zone load-balancing after having
+    potentially migrated cross-zone services to `MeshMultiZoneService`.
+  * instead of changing the behavior of MeshService, instead also generate
+    MeshMultiZoneService by default, which would reintroduce the cross-zone load balancing behavior.
+
+## Decision Outcome
+
+* Policies applied to `demo-app_kuma-demo_svc_5000` should also apply to
+  the `MeshService `{name: demo-app, namespace: kuma-demo, spec.ports[port == 5000]}`.
+
```
Icarus9913, line 74: Do we have the decision for the `cross-zone`?

michaelbeaumont: not yet, i'll add a tentative decision

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: What do you mean by use case? I don't think there's anything to validate. Policies targeting `kuma.io/service` and those targeting `MeshService` objects are both valid

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+when switching over to MeshService, this policy will no longer apply.
+
+### Local traffic
+
+Because the `MeshServices` we generate have selectors equivalent to the `Service`,
+they will select the same `Dataplanes` that previously had the `kuma.io/service`
+corresponding to the `Service` and thus, absent any policies, traffic flows to
+the same endpoints and Envoy has the same cluster configuration.
```
slonka, line 32: So I have an app that now talks to `httpbin-0_kuma-demo_svc_8000.mesh:80` - I turn on MeshService generation - does the traffic still work? If so, what do I do next? Do I switch the app to a hostname generated by hostname generator? Same question goes for a service that uses k8s dns.

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### `MeshService`
+
+- We now have one outbound listener:
+  - Traffic to the generated hostname of the `MeshService` goes through the Cluster IP
+  - This cluster IP is no longer cross-zone load balanced
+
+## Considered Options
+
+* Policies applied to `demo-app_kuma-demo_svc_5000` should also apply to
```
slonka, line 60: indefinitely or for some transition period and then it's going to get removed?

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,140 @@
...
+
+One of the fundamental changes with `MeshService` is that traffic to a service
+is no longer load balanced across all zones.
+
+We need to take measures to prevent this change from happening on upgrade.
+
+In particular with Kubernetes, it's important to keep in mind that after
+MeshService is enabled we no longer generate our own VIP.
+So any requests to MeshService hostnames go to the same
+outbound listener as with requests to Kubernetes DNS names, the ClusterIP.
```
slonka, line 44: Is `MeshLoadBalancingStrategy` changing anything here? How will it work with MeshService / MeshMultizoneService?

slonka (commented):
Maybe this is all figured out but what would be the easiest for me to understand is a step by step guide on how to migrate from the user's perspective. Starting from: I'm running Kuma, I'm using both `.mesh` and `k8s` domains with local / cross zone traffic and I want to end up using MeshService / MeshMultizone service.

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,140 @@
...
+If for example we have a policy targeted at:
+
+```
+spec:
+  targetRef:
+    kind: MeshService
+    name: demo-app_kuma-demo_svc_5000
+```
+
+when switching over to MeshService, this policy will no longer apply.
```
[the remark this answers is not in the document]
michaelbeaumont: Yeah. We should have this in the docs

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,140 @@
...
+
+* One of the easiest, most straightforward option as far as stability and UX goes.
+
+#### Cons
+
+* Will be harder to motivate users to switch at all
+* None of the advantages of MeshService & HostnameGenerator "for free"
+* Odd that MeshService objects exist but aren't used
+
+### Make cross-zone the default behavior
```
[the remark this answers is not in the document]
michaelbeaumont: it doesn't really. It's nothing but a temporary fix

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: It will work because the old VIP will be generated, that DNS name will point to it and an outbound listener will be added. But this listener will point to a cluster generated "by MeshService" and so is subject to the behavior of MeshService i.e. routes to local zones only atm. The intention is yes, you would switch to the MeshService native name. Same goes for k8s. The listener for the cluster IP also points to the cluster.

I'll add the above to the MADR

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: as long as we have `kuma.io/service` I suppose

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,165 @@
...
+```
+spec:
+  targetRef:
+    kind: MeshService
+    name: demo-app_kuma-demo_svc_5000
+```
+
+when switching over to MeshService, this policy will no longer apply.
+
+### Traffic
```
lukidzi, line 34: I think it can also affect `reachableServices`. If we have old `kuma.io/services` and `MeshService` outbounds, a user will now have outbounds included by reachable services annotation and MeshService one.

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,165 @@
...
+In particular with Kubernetes, it's important to keep in mind that after
+MeshService is enabled we no longer generate our own VIP.
+So any requests to MeshService hostnames go to the same
+outbound listener as with requests to Kubernetes DNS names, the ClusterIP.
+
+##### Status quo
+
+- On Kubernetes there are two outbound listeners:
+  - Traffic to the Kubernetes DNS name/ClusterIP is load-balanced cross-zone
+  - Traffic to the `<kuma.io/service>.mesh`/VIP is load-balanced cross-zone
```
jakubdyszkiewicz, line 74: nit: I think it's now one outbound listener with `additional_addresses` or something like that, no?

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,165 @@
...
+
+#### Cons
+
+* Is it coherent?
+  - the set of endpoints is shared between MeshServices local
+    and synced from other zones
+  - HostnameGenerator can't/shouldn't by default generate with a zone suffix
+* Need to keep using ZoneIngress availableServices
+
+### Generate MeshMultiZoneService
```
jakubdyszkiewicz, line 151: There are a couple of open questions here.

1) How would we preserve policy matching in this case?

policy matching
```
kind: MeshService
name: redis_kuma-demo_svc_5678
```
would not match MeshMultiZoneService, right?

2) How would we switch to using MeshMultiZoneService outbound?
so let say someone is using ClusterIP, we now have to switch so it's using MeshMultiZoneService instead of kuma.io/service. How do we then go to the desired behavior that uses MeshService?

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,165 @@
...
+If for example we have a policy targeted at:
+
+```
+spec:
+  targetRef:
+    kind: MeshService
+    name: demo-app_kuma-demo_svc_5000
+```
+
+when switching over to MeshService, this policy will no longer apply.
```
lobkovilya, line 32: With new MeshServices, what does it mean to reference MeshService in the top-level targetRef?

docs/madr/decisions/056-meshservice-migration.md
lukidzi: Let's say that you have reachableServices defined, and you enable MeshServices. Your dp is going to observe all services instead of the one that you provided in rechableServices

jakubdyszkiewicz:
It might be implem detail.

The assumption is that the ServiceTag cluster name is the same as MeshService cluster name. This is the case on Kube with name_namespace_svc_port. This is not the case on Universal
```
func (m *MeshServiceResource) DestinationName(port uint32) string {
	return fmt.Sprintf("%s_svc_%d", strings.ReplaceAll(m.GetMeta().GetName(), ".", "_"), port)
}
```
However, we could detect that MeshService is autogenerated by the universal generator and just assign it the service tag destination name.

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: It should be a to

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,104 @@
...
+Users would need to enable MeshService generation by default.
+
+#### Pros
+
+* Easiest, most straightforward option as far as stability and UX goes.
+
+#### Cons
+
+* Will be harder to motivate users to switch at all
+* None of the advantages of MeshService & HostnameGenerator "for free"
```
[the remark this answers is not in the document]
michaelbeaumont: Just the advantages of the resources in general. Custom hostnames, better insight with services enumerated as resources `.status`.

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: 1.
> would not match MeshMultiZoneService, right?

I mean it could, right?

2. You mean how to go from cross-zone to local zone behavior and switch to MeshService right? I could imagine:
  - Disable autogeneration
  - Delete MeshMultiZoneServices for services they don't want to do cross-zone

It does feel trickier than a switch to change the behavior for one resource.

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: I don't think it changes much except that it would just no longer work once the local zone behavior comes into effect. Though this sort of follows from the fact that a MeshService is one zone only, it should be heavily called out. And the policy would then need to be applied to a MeshMultiZoneService: https://github.com/kumahq/kuma/issues/10935

michaelbeaumont:
> Starting from: I'm running Kuma, I'm using both .mesh and k8s domains with local / cross zone traffic

I tried to lay this out in the `### Traffic` section @slonka can you take a look?

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,180 @@
...
+But MeshService behaves differently in several different ways.
+
+In general the migration process for MeshService consists of no longer using
+`kuma.io/service` so:
+
+- targeting policies to the real `MeshService` resource
+- switching over to the `MeshService` i.e. `HostnameGenerator` hostnames, if not
+  using Kubernetes DNS
+
+### Policy
```
Automaat, line 21: I have a question about policy API, are we going to deprecate and later remove `MeshService` from `spec.targetRef`? I am asking because we will most likely move Mesh*Routes to `spec.to[].targetRef` and since `MeshService` is an outbound, it does not make sense to put it in `spec.targetRef`

michaelbeaumont: Yes that's the goal. I don't think it affects this MADR though, it's basically just a rule: "wherever kuma.io/service is targeted, also target MeshService generated for that kuma.io/service"

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### Cons
+
+* Will be harder to motivate users to switch at all
+* None of the advantages of MeshService & HostnameGenerator "for free"
+* Odd that MeshService objects exist but aren't used
+
+### Make cross-zone the default behavior
+
+The cross-zone behavior would be kept for the old, `<kuma.io/service>.mesh` VIP
```
lahabana, line 147: I don't understand here, AFAIK the clusterIP clusters are not cross-zone so why to we need to change behaviour?

lahabana (dismissed):
i think this doesn't think enough what a user would want to do.

What do you think it's the hardest for the mesh operator? (1) Adding extra policies with matching targetRefs? (2)Asking all users to change a hostname/port?

IMO 2 is much harder as it requires coordination, yet this PR doesn't mention adding a HostnameGenerator to have hostname parity with older versions

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: Why wouldn't they be cross-zone? They have the same clusters

https://kuma.io/docs/2.8.x/production/cp-deployment/multi-zone/#cross-zone-communication-details

michaelbeaumont:
> IMO 2 is much harder as it requires coordination, yet this PR doesn't mention adding a HostnameGenerator to have hostname parity with older versions

True, that's missing from the MADR

docs/madr/decisions/056-meshservice-migration.md
jakubdyszkiewicz: why? I think it will just stay as is. DPP will still be able to connect to the service using servicetag vip

docs/madr/decisions/056-meshservice-migration.md
jakubdyszkiewicz: Yeah. I'm asking those questions to lay out complexity. On paper, it sounds best to "just generate" MeshMultiZoneService, but it's a lot of moving parts.

michaelbeaumont: Yeah I don't disagree and probably wouldn't go with this option at this point

slonka:
> Wait we still generate this because we have the old hostname & VIP generation

Will we continue to do this indefinitely or this will only happen for old users? Do we plan to get people off `<kuma.io/service>.mesh`?

michaelbeaumont:
> Will we continue to do this indefinitely or this will only happen for old users? Do we plan to get people off <kuma.io/service>.mesh?

yeah this is in the MADR I think. The goal is of course to stop using `kuma.io/service` entirely, which includes the `.mesh` addresses.

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,301 @@
...
+On Kubernetes we also have
+
+- `name.namespace.svc.cluster.local` -> ClusterIP -> same cross-zone cluster
+
+#### Once `MeshService` is created
+
+`MeshService` takes over _the cluster generation_ but does
+not affect the VIP generation. So we now have:
+
+- `<kuma.io/service>.mesh` -> Kuma VIP -> local-zone cluster
```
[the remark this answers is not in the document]
michaelbeaumont: Yes, unless `SKIP_PERSISTED_VIPS` is enabled. But this is just a description of how things are right now.

lukidzi (approved):

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,343 @@
...
+### Don't generate MeshServices by default
+
+Users would need to enable MeshService generation and MeshService-controlled XDS
+generation.
+
+#### Migration
+
+1. Upgrade to new version
+1. Create MeshMultiZoneServices for cross-zone traffic.
+  - Potentially create MeshService manually to gradually switch over
```
lukidzi, line 204: We could have a tool that generates MeshServices out of Services/Dataplanes.

lukidzi (approved):

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,343 @@
...
+redeploying on a new `Mesh` or control plane.
+
+We have to make a choice between:
+
+- new behavior by default, `UPGRADE.md` tells users to set a variable to keep
+  old behavior
+- old behavior by default, new users have to be convinced to set a variable to
+  get new behavior
+
+### `MeshService`-targeted policy
```
lobkovilya, line 35: Probably worth mentioning `backendRefs` inside Mesh*Route policies. Do we want to convert existing one to targetRef with labels as well?

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,343 @@
...
+
+- `name.namespace.svc.cluster.local` -> ClusterIP -> same cross-zone cluster
+
+#### Once `MeshService` is created
+
+`MeshService` takes over _the cluster generation_ but does
+not affect the VIP generation. So we now have:
+
+- `<kuma.io/service>.mesh` -> Kuma VIP -> local-zone cluster
+- `name_namespace_svc_port` -> ClusterIP -> local-zone cluster
```
lobkovilya, line 127: I don't really understand this, what do you mean by `name_namespace_svc_port` here? it's not hostname right?

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,343 @@
...
+### Generate new clusters
+
+Instead of `MeshService` taking over cluster generation, generate separate clusters so that:
+
+- `<kuma.io/service>.mesh` -> Old Kuma VIP -> old cluster
+- `HostnameGenerator` DNS name for the `MeshService` -> New Kuma VIP
+  - this includes synced `MeshServices`
+- `name.namespace.svc.cluster.local` -> ClusterIP -> depends
+
+For example, we could have two formats like `name_namespace_svc_port` and `name_namespace_msvc_port`.
```
lobkovilya, line 285: Probably just `name` and `namespace` might be not ideal, as the name for synced resources contains hash suffix and namespace for these resources is `kuma-system`. Should we use the same 4 fields that we have in ResourceIdentifier – name, namespace, zone, mesh?

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: yeah this is meant to be `name.namespace.cluster.local`

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -9,132 +9,335 @@ being automatically created, either from `Services` in Kubernetes or from
...
+  - `Disabled` is the default
+  - `Everywhere` syncs `MeshService` outbounds to all data plane proxies
+  - `Exclusive` disables `kuma.io/service` outbounds
+  - `ReachableBackendRefs` to enable on a case-by-case basis by setting
+     `reachableBackendRefs`:
+        `reachableBackendRefs: { kind: MeshService, labels: {} }`
+  - **NOTE**: On k8s if the consumer can reach a given local `MeshService`,
+    the Kubernetes IP has the behavior of `MeshService`
+  - **NOTE**: This change makes the *experimental* options `GENERATE_MESH_SERVICES` & `SKIP_PERSISTED_VIPS`
+    *obsolete*, these options *now do nothing*
```
slonka, line 188: What's the sequence of going through the options in relation to other points? 

I assume we start with `disabled` then we go with `ReachableBackendRefs` then we do point 2, then we go with `Everywhere` and `Exclusive` (which is point 3 right?) after the migration is done?

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: @lobkovilya take another look

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: I cleaned this up a bit @slonka

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,349 @@
...
+      kind: MeshService
+      name: backend
+```
+
+is meant to target `kuma.io/service` or a `MeshService` named `backend` and thus
+we can't determine whether all instances of `kuma.io/service: backend` are meant
+or only the instances matched by `MeshService`.
+
+Thus we have to require `sectionName` to a port, in order to refer to
+`MeshServices` unambiguously.
```
jakubdyszkiewicz, line 84: so is it sectionName or port? In the example above there is `port`

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,349 @@
...
+
+#### Port name
+
+We can't require users to add a `name` field to their existing `Services` or
+annotate their `Dataplanes` for universal, so we should
+set a canonical name for every port assuming one isn't set.
+
+Because the `port` field must be unique anyway, this could be a suitable,
+automatic `name`.
```
jakubdyszkiewicz, line 93: Do we add this explicitly when we convert Service -> MeshService and Dataplane -> MeshService?

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,349 @@
...
+
+#### Migration
+
+1. Upgrade to new version
+  If the user doesn't change any config, nothing changes. No `MeshServices` are generated.
+1. On `Mesh`, users have the option `meshServices`:
+  - `Disabled` is the default
+  - `Everywhere` syncs `MeshService` outbounds to all data plane proxies
+  - `Exclusive` disables `kuma.io/service` outbounds and enables `MeshService`
+  - `ReachableBackendRefs` to generate all `MeshServices` but sync
```
jakubdyszkiewicz, line 330: Does it disable kuma.io/service as well?

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: yeah

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: No. The idea being you'd still choose what to sync via `reachable*`

docs/madr/decisions/056-meshservice-migration.md
```diff
@@ -0,0 +1,349 @@
...
+
+- `<kuma.io/service>.mesh` -> Old Kuma VIP -> old cluster
+- `HostnameGenerator` DNS name for the `MeshService` -> New Kuma VIP
+  - this includes synced `MeshServices`
+- `name.namespace.svc.cluster.local` -> ClusterIP -> depends
+
+The new format is:
+
+```
+<name>_<namespace>_<zone>_<mesh>_msvc_<port>
```
lobkovilya, line 288: what if namespace and zone are empty? is it `name_mesh_msvc_port` or `name___mesh_msvc_port`?

michaelbeaumont: yeah i mentioned namespace in the next sentence. should we just use the local zone name so it's never empty?

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: Yeah we do.

michaelbeaumont: Actually this doesn't quite make sense like it does for `targetRef`. It's important that these are manually updated because after MeshServices are created, it isn't clear whether traffic should go to that specific MeshService or maybe a MeshMultiZoneService.

michaelbeaumont: If we accept the above I'll add it to the MADR as part of the migration.

lobkovilya: But if you have a TCP route like:
```yaml
to:
  - targetRef:
      kind: MeshService
      name: backend_demo-ns_svc_8080
    rules:
      - default:
          backendRefs:
            - targetRef:
                kind: MeshService
                name: backend-v2_demo-ns_svc_8080
```
don't you expect after MeshServices are generated the route still works?

jakubdyszkiewicz (approved):

docs/madr/decisions/056-meshservice-migration.md
michaelbeaumont: OK let's use the empty string

michaelbeaumont: zone is empty when the resource comes from global

michaelbeaumont: let's triple check that underscores won't show up in any names

lobkovilya (approved):

Issue #10395, opened 2024-06-06
Add GOMEMLIMIT to cp deployment based on memory resources settings from user specified values


jijiechen:
### Description

`GOMEMLIMIT` is a game changer for limiting the memory consumption for Go programs; especially when it's running in a containerized environment, it will reduce the chances of unexpected OOM-killings because of memory stavation.
Let's add this environment variable onto Kuma control plane deployment to get these benefits.

ref: https://weaviate.io/blog/gomemlimit-a-game-changer-for-high-memory-applications

PR #10879, opened 2024-07-11
chore(helm): add GOMEMLIMIT and GOMAXPROCS environment variables

lukidzi:
### Checklist prior to review

As explained in the issue and article, setting `GOMEMLIMIT` and `GOMAXPROCS` is highly beneficial.

https://www.ardanlabs.com/blog/2024/02/kubernetes-cpu-limits-go.html
https://weaviate.io/blog/gomemlimit-a-game-changer-for-high-memory-applications

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/10395

### DIFF
--- a/app/kumactl/cmd/install/testdata/install-control-plane.cni-enabled.golden.yaml
+++ b/app/kumactl/cmd/install/testdata/install-control-plane.cni-enabled.golden.yaml
@@ -685,6 +685,16 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  containerName: control-plane
+                  resource: limits.memory
+            - name: GOMAXPROCS
+              valueFrom:
+                resourceFieldRef:
+                  containerName: control-plane
+                  resource: limits.cpu
           args:
             - run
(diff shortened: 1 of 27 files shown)


Automaat (approved):

jakubdyszkiewicz:
For the GOMEMLIMIT. Can we actually just take a number from limit? I think we need to go a little bit lower? @jijiechen 

For GOMAXPROCS. How does it relate to https://home.robusta.dev/blog/stop-using-cpu-limits ? Is it better to have hard limit + GOMAXPROCS or is it better to not go with a limit at all?

lukidzi:
I would say not setting a CPU limit, but if the user sets it, we can pass it in the environment variables.

`GOMEMLIMIT` we can adjust it probably. Should we use 80% of the limit?

jakubdyszkiewicz (approved):

jijiechen:
Yes, it's best to have a 80% or 90% percentage applied.
However, if we want to use the downward API, there is not an easy way to apply this percentage.

jijiechen:
We may use a Helm snippet to implement the percentage, it's not officially supported by the downward API: https://github.com/kubernetes/kubernetes/issues/91514

PR #10886, opened 2024-07-12
feat(kuma-cp): introduce ResourceRules

lobkovilya:
Implements the [MADR](https://github.com/kumahq/kuma/pull/10726).
Don't let the number of lines to scare you, those are just golden files 😉

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- https://github.com/kumahq/kuma/issues/10991


pkg/plugins/policies/core/rules/resourcerules.go
```diff
@@ -0,0 +1,172 @@
...
+	rules := ResourceRules{}
+
+	SortByTargetRefV2(list)
+
+	var resolvedItems []*resolvedPolicyItem
+	for _, item := range list {
+		resolvedItems = append(resolvedItems, resolveTargetRef(item, l)...)
+	}
+
+	for _, i := range resolvedItems {
```
Automaat, line 67: can we change naming of this variables form i, j to something more concrete that will be easier to understand?

pkg/plugins/policies/core/rules/resourcerules.go
```diff
@@ -0,0 +1,172 @@
...
+type ResourceRules map[UniqueResourceKey]ResourceRule
+
+func (rr ResourceRules) Compute(r core_model.Resource, l ResourceLister) *ResourceRule {
+	key := uniqueKey(r)
+	if rule, ok := rr[key]; ok {
+		return &rule
+	}
+
+	findMesh := func(meshName string) core_model.Resource {
+		meshes := l.ListOrEmpty(core_mesh.MeshType).GetItems()
```
Automaat, line 30: Can't we just get mesh by key from store? We always use `ResourceLister` in this and we end up iterating over all resources, it does not look optimal from performance point of view

Automaat: This will be especially problematic when we start adding support for routes and we end up iterating over every MeshService each time we run Compute and other code here

pkg/plugins/policies/core/rules/testdata/sort/by-origin.input.yaml
```diff
@@ -0,0 +1,45 @@
...
+  to:
+    - targetRef:
+        kind: Mesh
+      default:
+        connectionTimeout: 111s
+        idleTimeout: 109s
+---
+type: MeshTimeout
+name: matched-for-rules-bbb
+mesh: mesh-1
```
Automaat, line 19: can we have a policy without `kuma.io/origin` label?

pkg/plugins/policies/core/rules/resourcerules.go
lobkovilya: I rewrote this function to be more readable, no more i,j in the code

pkg/plugins/policies/core/rules/testdata/sort/by-origin.input.yaml
lobkovilya: I don't remember how exactly you can have it, but I'm pretty sure it's possible. Maybe if you upgrade from the version that doesn't add `kuma.io/origin` automatically and you had the origin validation disabled then I guess after the Kuma CP upgrade you'll have resources without origin

pkg/plugins/policies/core/rules/resourcerules.go
lobkovilya: Under the hood it's using `MeshContext.Resources` that stores resources as `map[core_model.ResourceType]core_model.ResourceList`. Changing this to map by ResourceIdentifier can potentially affect multiple places and I'd prefer doing this in a separate PR. 

Though in this PR I created a method `Get()` that's used everywhere in `resourcerule.go` so we don't have linear searches all-over the code.

pkg/plugins/policies/core/rules/resourcerules.go
```diff
@@ -0,0 +1,257 @@
...
+
+type UniqueResourceIdentifier struct {
+	core_model.ResourceIdentifier
+
+	ResourceType core_model.ResourceType
+	SectionName  string
+}
+
+func (ri UniqueResourceIdentifier) String() string {
+	var pairs []string
```
Automaat, line 37: would we need to create ResourceIdentifier from this string? When I started implementing InspectAPI I needed to extract resource type, so some part of it. Now resource type can be empty, should we always add something?

pkg/plugins/policies/core/rules/resourcerules.go
Automaat: I didn't check what we use under the Lister interface, but with MeshContext this makes sense.

pkg/plugins/policies/core/rules/testdata/sort/by-origin.input.yaml
Automaat: this sounds like a possible use case

pkg/plugins/policies/core/rules/resourcerules.go
```diff
@@ -0,0 +1,257 @@
...
+
+// isRelevant returns true if the policyItem is relevant to the resource or section of the resource
+func isRelevant(policyItem *resolvedPolicyItem, r core_model.Resource, sectionName string) bool {
+	switch policyItem.resource.Descriptor().Name {
+	case core_mesh.MeshType:
+		switch r.Descriptor().Name {
+		case core_mesh.MeshType:
+			return policyItem.resource.GetMeta().GetName() == r.GetMeta().GetName()
+		default:
+			return policyItem.resource.GetMeta().GetName() == r.GetMeta().GetMesh()
```
Automaat, line 195: `policyItem.resource.GetMeta().GetName() == r.GetMeta().GetMesh()` should we even check this? wouldn't it end up in merge all other types with Mesh? I think that we've discussed that we won't merge MeshService conf with Mesh conf at this level?

pkg/plugins/policies/core/rules/resourcerules.go
lobkovilya: I actually changed the `ResourceRules` map to have `UniqueResourceIdentifier` as a key. So there is no need now to parse it

Automaat (approved):

PR #10888, opened 2024-07-13
test(transparent-proxy): flatten test structure and prepare for uninstal

bartsmykla:
- Moved `install` tests from the `install` directory to the root suite directory
- Renamed tests to facilitate future integration of `uninstall` tests
- Extracted `iptables-save` commands to global variables for reuse in  uninstall tests
- Made small improvements in the codebase:
  - Renamed some variables
  - Added comments to `Ensure*` functions
  - Added comments in test functions
  - Passed `context.Context` to all functions instead of using hardcoded `context.Background()`

### Checklist prior to review

  - No relevant issues
  - It won't
  - There is no need
  - There is no need


test/transparentproxy/transparentproxy_test.go
```diff
@@ -44,90 +61,127 @@ type testCase struct {
 	additionalFlags  []string
 }
 
-func Install() {
+var _ = Describe("Transparent Proxy", func() {
 	DescribeTable(
-		"kumactl install transparent-proxy inside Docker container",
+		"install in container",
 		func(tc testCase) {
+			ctx := context.Background()
```
Icarus9913, line 68: Nice change!

Icarus9913 (approved):

PR #10889, opened 2024-07-13
fix(utils): enhance IsClosed

sjmshsh:
### Checklist prior to review

- The current IsClosed method is not rigorous enough. When the channel is not closed, the IsClosed function is called. However, if a message is sent to the channel by the producer at this time, the channel that has not been closed will be mistakenly considered to be closed.


jakubdyszkiewicz:
Outside of tests we use this method only on "close channels", channels that we only close, we don't send anything. But technically you are right 👍 can you sign your commit so the DCO passes? Thanks

sjmshsh:
> Outside of tests we use this method only on "close channels", channels that we only close, we don't send anything. But technically you are right 👍 can you sign your commit so the DCO passes? Thanks

Ok, I did it in this PR https://github.com/kumahq/kuma/pull/10894

PR #10894, opened 2024-07-15
fix(utils): enhance IsClosed

sjmshsh:
Checklist prior to review
- The current IsClosed method is not rigorous enough. When the channel is not closed, the IsClosed function is called. However, if a message is sent to the channel by the producer at this time, the channel that has not been closed will be mistakenly considered to be closed.

xrel: https://github.com/kumahq/kuma/pull/10889


slonka:
Hi :) I think it's going to be easier if you brach off on your fork and create a PR from that branch instead of "master". That way the PR won't include some unnecessary commits.

slonka (approved):

Issue #10882, opened 2024-07-11
Kuma CP on k8s adds `k8s.kuma.io/namespace` label for synced resources


lobkovilya:
### What happened?

When global-k8s receives resource from zone-uni it's going to add `k8s.kuma.io/namespace: kuma-system` label.

It's bad for a few reasons:
* `kuma.io/zone` and `k8s.kuma.io/namespace` labels are part of the hash, so the invariant for synced resources:
    ```
    name = hashedName(mesh, name, labels[kuma.io/zone], labels[k8s.kuma.io/namespace])
    ```
    is not preserved.
* we should be able to tell if the resource us synced from k8s or uni zone by looking at `k8s.kuma.io/namespace` label. We need this for validation of labels in `backendRef` when referencing MeshService.

lukidzi:
> we shouldn't add namespace labels for resources created on universal and synced, lets look into it

PR #10913, opened 2024-07-16
fix(kuma-cp): don't add namespace labels when resource was synced from universal zone

Automaat:
Fixes: #10882

### Checklist prior to review


pkg/plugins/resources/k8s/store_test.go
```diff
@@ -560,6 +560,37 @@ var _ = Describe("KubernetesStore", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(actual.Meta.GetLabels()[mesh_proto.DisplayName]).To(Equal("dn"))
 		})
+
+		It("should not add namespace label when resource is synced from zone", func() {
+			// setup
+			expected := backend.ParseYAML(fmt.Sprintf(`
+            apiVersion: kuma.io/v1alpha1
+            kind: TrafficRoute
```
lobkovilya, line 568: please use a new policy for the test, so we have one less test to rewrite when we remove old policies

pkg/plugins/resources/k8s/store.go
```diff
@@ -298,8 +298,12 @@ func (m *KubernetesMetaAdapter) GetLabels() map[string]string {
 	} else {
 		labels[v1alpha1.DisplayName] = m.GetObjectMeta().GetName()
 	}
-	if _, ok := labels[v1alpha1.KubeNamespaceTag]; !ok && m.Namespace != "" {
-		labels[v1alpha1.KubeNamespaceTag] = m.Namespace
+	origin := labels[v1alpha1.ResourceOriginLabel]
+	_, zoneLabelPresent := labels[v1alpha1.ZoneTag]
+	if !(origin == string(v1alpha1.ZoneResourceOrigin) && zoneLabelPresent) {
```
lobkovilya, line 303: If this code runs on zone k8s and checks the local resource then I `!(origin == zone && zoneLabelPresent)` is `false` and we won't add ns label while we actually need it.

You should either call (but it requires CpMode): https://github.com/kumahq/kuma/blob/2896dd48d541c8b0dca1356ef9e85dbb47951299/pkg/core/resources/model/resource.go#L411

Or you can do the same way as I did in my PR (passing CpMode would affect a lot of function signatures) https://github.com/kumahq/kuma/pull/10886/files#diff-1779d56feb27e63cfeffdc2ddf598e93ba7a95977d3227e8414955cdd77e4fd4R450-R456 

If you go with the second option please make it a reusable function in `pkg/core/resources/model/resource.go`

pkg/plugins/resources/k8s/store_test.go
Automaat: changed

pkg/plugins/resources/k8s/store.go
Automaat: changed to using function, unfortunately passing cpMode is impossible in this case

Automaat: So this approach does not work, since introducing function in `pkg/core/resources/model/resource.go` with nice api, makes it use methods from which I was calling it, resulting in stack overflow. I've moved it to GetLabel method, but this approach is hacky :/  

Maybe the most clean option would  be to introduce new label like `kuma.io/synced-resource` to simply know that this resource was not created on current zone?

pkg/plugins/resources/k8s/store.go
```diff
@@ -59,6 +62,14 @@ func (s *KubernetesStore) Create(ctx context.Context, r core_model.Resource, fs
 	}
 
 	labels, annotations := SplitLabelsAndAnnotations(opts.Labels, obj.GetAnnotations())
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	if core_model.IsLocallyOriginated(s.CpMode, r) {
```
lobkovilya, line 68: On Kubernetes, the resource can be created in 2 ways:
* applied by user, we catch that in webhook `defaulter.go`
* created by CP, we catch that in this `store.go`

Should we share a call to `ComputeLabels` in these two places?

pkg/plugins/resources/k8s/store.go
```diff
@@ -59,6 +62,14 @@ func (s *KubernetesStore) Create(ctx context.Context, r core_model.Resource, fs
...
+		if _, ok := labels[v1alpha1.KubeNamespaceTag]; !ok && namespace != "" {
+			labels[v1alpha1.KubeNamespaceTag] = namespace
+		}
+	}
```
lobkovilya, line 72: We probably should set labels on the update as well in case they were deleted

pkg/core/resources/model/resource.go
```diff
@@ -409,12 +409,19 @@ func IsReferenced(refMeta ResourceMeta, refName string, resourceMeta ResourceMet
 }
 
 func IsLocallyOriginated(mode config_core.CpMode, r Resource) bool {
+	if r.GetMeta() == nil {
```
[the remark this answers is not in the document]
Automaat: we are checking this because lots of tests panicked without this check.

It feels wrong to return `false` in this case, but when I started from `true` the whole functionality didn't work and we were still setting namespace labels on synced resources. I checked it now and defaulter is running on every store operation as this is kubernetes webhook. So probably we don't need to add namespace in store, it would be enough to update defaulter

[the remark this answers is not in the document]
lobkovilya: It was probably panic when called from the Create, because ResourceMeta is not set. If you call this only from defaulter now, maybe we don't need this `nil` check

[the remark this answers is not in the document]
Automaat: removed check

pkg/core/resources/model/resource.go
```diff
@@ -409,12 +409,16 @@ func IsReferenced(refMeta ResourceMeta, refName string, resourceMeta ResourceMet
 }
 
 func IsLocallyOriginated(mode config_core.CpMode, r Resource) bool {
+	return IsLocallyOriginatedByLabels(mode, r.GetMeta().GetLabels())
+}
+
+func IsLocallyOriginatedByLabels(mode config_core.CpMode, labels map[string]string) bool {
```
lobkovilya, line 415: I think we should keep only one of the methods `IsLocallyOriginated` or `IsLocallyOriginatedByLabels`. Having both just makes no sense in my opinion

Automaat: removed

lobkovilya (approved):

Issue #10915, opened 2024-07-16
Exclude unhealthy zone in MeshMultiZoneService


jakubdyszkiewicz:
### Description

Right now we don't have status of MeshService so we don't know if there is any healthy endpoint in the zone

PR #10916, opened 2024-07-16
docs(MADR): proxies stats of MeshService

jakubdyszkiewicz:
### Checklist prior to review

xref #10915

### DIFF
--- a/docs/madr/decisions/056-meshservice-proxies-stats.md
+++ b/docs/madr/decisions/056-meshservice-proxies-stats.md
@@ -46,7 +46,7 @@ This information is computed by the source zone and synced to global for visibil
 
 Fields:
 * `dataplaneProxies.connected` - number of DPPs connected to the Zone CP
-* `dataplaneProxies.healthy` - number of DPPs connected with all healthy inbounds
+* `dataplaneProxies.healthy` - number of DPPs connected with all healthy inbounds selected by the MeshService
 
 Fields that are in `ServiceInsights`, but won't be included in status of `MeshService`
 * `status` - If we want to present in the GUI if `MeshService` is available to consume, we can use `spec.state` described in the next section
--- a/docs/madr/decisions/056-meshservice-proxies-stats.md
+++ b/docs/madr/decisions/056-meshservice-proxies-stats.md
@@ -1,0 +1,80 @@
+# MeshService proxies stats
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/10915
+
+## Context and Problem Statement
+
+[ServiceInsight](https://github.com/kumahq/kuma/blob/master/api/mesh/v1alpha1/service_insight.proto) has detailed information about the Service.
+We need an equivalent of it in MeshService.
+
+ServiceInsight includes the Status field
+* Online - all dataplane proxies are connected to CP and have healthy inbounds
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/056-meshservice-proxies-stats.md
```diff
@@ -0,0 +1,80 @@
...
+Because `status` of MeshService is synced from Zone CP to Global CP only (it's not synced cross zone),
+we need to put this in `spec` just like we did with `spec.identities`.
+Client's zone do not care how many endpoints are in the zone, it only cares if there is at least one endpoint available.
+Not sending the exact number of endpoints cross zone helps with excessive KDS updates of remote MeshServices.
+`state` gets two states:
+* `Available` - at least one DPP selected by `MeshService` is healthy.
+  While technically health status can be different for each port, assuming each port is exposed from different container,
+  we want to avoid this complexity, and we want to use just one status. 
+* `Unavailable` - no DPPs selected by `MeshService` are healthy
+If `state` field is missing, we assume that it is `Unavailable`.
```
johncowen, line 76: I think I see this happen a lot (as a GUI engineer consuming the HTTP API). I'm guessing there is a reason for this, so excuse this question if its non-trivial:

How come the initial state can't be `Unavailable` and then only mark the MeshService as `Available` if it is actually available? That way we don't have things like `status: Available | Unavailable | undefined` and then I don't need to check whether the `state` property exists before checking its value.

Having to check that `state` is there first is almost like saying it has a third state called 'stateless'.

Again I'm assuming this is more complex than it sounds or there are more things in the way that make it more difficult than it sounds, but lemme know.

johncowen: (apart from this question, the MADR itself sounds great to me! 👍 )

jakubdyszkiewicz: That's a fair question. It's because status is computed async and updated by the CP, user should not set this. However, we technically could leverage a "defaulter" mechanism to always set this to Unavailable if it's empty. I think that would work.

johncowen: Gotcha, some kind of defaulting mechanism sounds good to me!

slonka (approved):

slonka:
nit: title can't have scope `madr`

docs/madr/decisions/056-meshservice-proxies-stats.md
```diff
@@ -0,0 +1,80 @@
...
+    healthy: 8
+    total: 10
+```
+
+We now separate connected proxies to the zone control plane from proxies that have healthy inbounds.
+This information is computed by the source zone and synced to global for visibility.
+
+Fields:
+* `dataplaneProxies.connected` - number of DPPs connected to the Zone CP
+* `dataplaneProxies.healthy` - number of DPPs connected with all healthy inbounds
```
lahabana, line 49: Is it that the dpp has all healthy inbounds or inbounds for this service healthy?

jakubdyszkiewicz: It's all inbounds selected by this service. At least for now.

lukidzi (approved):

lahabana (approved):

PR #10918, opened 2024-07-16
docs(MADR): SNI in the resource

jakubdyszkiewicz:
### Checklist prior to review

SNI in MeshService and MeshMultiZoneService


michaelbeaumont (approved):

docs/madr/decisions/057-sni-in-the-resource.md
```diff
@@ -0,0 +1,130 @@
...
+
+SNI is unique for each port, therefore we need to define it for each port
+
+```yaml
+kind: MeshService
+spec:
+  ports:
+    - port: 80
+      targetPort: 8080
+      sni: ae10a8071b8a8eeb8.backend.8080.demo.ms # new field
```
lahabana, line 40: I guess migration is still not possible like this (we'd need to have possibility to have multiple SNIs to support migration). But I guess this is a start at least.

If we wanted to support migration I guess we would need:

```yaml
sni:
   server: [a, b] # zoneIngress
   client: a
```

So migration would happen:

1. Migrate all servers to support both
```yaml
sni:
   server: [old, new]
   client: old
```

2. Migrate clients to new one:
```yaml
sni:
   server: [old, new]
   client: new
```

3. Remove old from server:
```yaml
sni:
  server: [new]
  client: new
```

For sure this is too cumbersome to expose from the start. But should we at least make the model flexible for the future?

.e.g:

```yaml
sni:
  client: new
```

Maybe client is not the right field name. But the ability to extend SNI in the future seems required

docs/madr/decisions/057-sni-in-the-resource.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+The `sni` field will be filled out automatically be the original Zone CP and synced cross zone.
```
lahabana, line 43:
```suggestion
The `sni` field will be filled out automatically by the original Zone CP and synced cross zone.
```

docs/madr/decisions/057-sni-in-the-resource.md
lahabana: Ha or by migration you mean we wouldn't update existing SNIs?

docs/madr/decisions/057-sni-in-the-resource.md
```diff
@@ -0,0 +1,130 @@
...
+
+Advantage:
+* We see ports in the GUI
+* Asking less things from the user
+
+Disadvantage:
+* Putting more compute operations on Global CP.
+  However, we may need to have a component to compute `status` on global anyway to give better GUI visibility of aggregated services. 
+
+### Option 2 - Keep it implicit
```
lahabana, line 123:
```suggestion
### Option 3 - Keep it implicit
```

docs/madr/decisions/057-sni-in-the-resource.md
```diff
@@ -0,0 +1,130 @@
...
+    - ip: 241.0.0.1
+  meshServices:
+    - name: test-server.mzmsconnectivity
+    - name: test-server-234jhg34yt34.kuma-system
+  ports:
+    - port: 80
+      targetPort: 8080
+```
+
+##### Option 1 - require ports from user
```
lahabana, line 82: What's you suggested option recommendation?

I lean towards 1 just because all the advantages you mention are things that we hit with the implicit `service` we have right now.

So I tend to think if we hit them now, why wouldn't we hit them in a very similar case.

Also it's easier I can imagine people wanting to restrict exposed ports and if we go the future route of auto generating in the future this would not matter.

WDYT?

docs/madr/decisions/057-sni-in-the-resource.md
jakubdyszkiewicz: You are right. Instead of this server client etc, we could do this
```
snis:
- new
- old
```
and just recommend the client to always use the first one on the list

docs/madr/decisions/057-sni-in-the-resource.md
jakubdyszkiewicz: Option 1 is the best for us = engineers because it removes ambiguity.
I worry that it requires more from the user, but maybe it's not that bad?

It's easier to get rid of it in the future than introduce it, so maybe we should start with it

slonka: I like option 2 more

docs/madr/decisions/057-sni-in-the-resource.md
slonka: Please remember to update the madr with this new `snis` array

docs/madr/decisions/057-sni-in-the-resource.md
```diff
@@ -0,0 +1,131 @@
...
+The SNI is constructed on the server and the client side. This approach has a couple of downsides
+* We hit multiple edge cases of constructing it on the client side, because of resource hashing.
+  This logic is not as simple as it should be.
+* Any migration of this format would be hard to execute
+* If we introduce a functionality to enter the mesh directly through zone ingress (assuming that cert is acquired) it would be hard for the client to construct this.
+
+## Considered Options
+
+* SNI in the resource definition
+* Keep it implicit
```
slonka, line 21: There should be 2 subsections right?

```suggestion
MeshService:
* SNI in the resource definition
* Keep it implicit

MeshMultiZoneService:
* Require ports from user
* Compute ports on global cp
* Keep it implicit
```

slonka (approved):

michaelbeaumont (approved):

docs/madr/decisions/057-sni-in-the-resource.md
```diff
@@ -0,0 +1,171 @@
...
+      - value: "ae10a8071b8a8eeb8.backend.80.demo.mzms"
+```
+
+SNI is then applied the same way as MeshService (via webhook/ResourceManager) and is synced down to zones.
+
+Advantage:
+* We see ports in the GUI
+* We avoid mistakes when ports are different between services. What to do when
+  * One service has port 80 and other has port 8080
+  * One service has appProtocol http on port 80 and other has tcp on port 80
```
lahabana, line 146: MeshMZService is similar to MeshService. This is something that confused @johncowen today

lukidzi (approved):

jakubdyszkiewicz:
talked offline. Settled on what is in MADR

Issue #10897, opened 2024-07-15
Decide on `MeshExternalService` and `ZoneEgress`


slonka:
### Description

Should we make MeshExternalService a `ZoneEgress` only resource? 
Where does it make sense on ZoneEgress?
Where does it make sense on the sidecar?
Which targetRef level do we use in which policies? (topLevel / to / from)

PR #10921, opened 2024-07-17
docs(MADR): MeshExternalService traffic through egress/sidecar and configuration placement

slonka:
one section will still need work

closes https://github.com/kumahq/kuma/issues/10897

### Checklist prior to review


docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,239 @@
...
+
+and if it makes sense for the user to pick where.
+
+| Policy                    | On sidecar   | On egress    | On both | User configurable |
+|---------------------------|--------------|--------------|---------|-------------------|
+| MeshAccessLog             | Yes          | Yes          | Yes     | Yes               |
+| MeshCircuitBreaker        | Yes          | Maybe        | No      | No                |
+| MeshFaultInjection        | Maybe        | Yes          | No      | No                |
+| MeshHealthCheck           | Yes          | Maybe        | No      | No                |
+| MeshHealthCheck           | Yes          | Maybe        | No      | No                |
```
lukidzi, line 49: duplicate

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,239 @@
...
+| MeshMetric                | Yes          | Yes          | Yes     | Maybe             |
+| MeshProxyPatch            | Yes          | Yes          | Yes     | Yes               |
+| MeshRateLimit             | Probably not | Yes          | No      | No                |
+| MeshRetry                 | Yes          | Maybe        | No      | Maybe             |
+| MeshTimeout               | Yes          | Probably not | No      | No                |
+| MeshTrace                 | Yes          | Yes          | Yes     | Maybe             |
+| MeshTrafficPermission     | Maybe        | Yes          | Maybe   | Maybe             |
```
lukidzi, line 56: MeshTrafficPermission currently works for ExternalService to support AutoreachableServices

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka:
```suggestion
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,239 @@
...
+The table below shows a summary of policies and if it makes sense to configure them on:
+- sidecar
+- egress
+- on both
+
+and if it makes sense for the user to pick where.
+
+| Policy                    | On sidecar   | On egress    | On both | User configurable |
+|---------------------------|--------------|--------------|---------|-------------------|
+| MeshAccessLog             | Yes          | Yes          | Yes     | Yes               |
```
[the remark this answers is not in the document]
slonka: > As in if you use a sidecar (non egress). Does it make sense to have both to and from ?

I agree that on the sidecar only `to` makes sense and on the `egress`... do both make sense? You receive request X on the inbound but it fails and is retried so we should see X twice on the outbound.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,276 @@
...
+* predefined mixed approach on a case-by-case basis
+* only on the egress
+* only on the sidecar
+
+## Decision Outcome
+
+Chosen option: ?
+
+Chosen option for configuration placement: "Configurable where it makes sense on both sidecar and egress, predefined where only one makes sense"
+because it fits the decision drivers (flexibility / restiveness).
```
slonka, line 41:
```suggestion
Chosen option for configuration placement: ?
```

docs/madr/decisions/notes.md
```diff
@@ -0,0 +1,22 @@
...
+  - it's a bit annoying to set up on universal
+- we've got a bit of a gap in other functionalities on egress
+- can we hit scalability issues with egress?
+- you need mTLS enabled (but only for services) to run egress, but we know if some traffic is MES, so we can relax that 
+
+
+Should we make egress required for MeshServices if we make it required for MeshExternalServices?
+
+Conclusion from the meeting
+**It seems like we're not cutting any serious use case**
```
slonka, line 22: mistakenly added 

```suggestion
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,275 @@
...
+#### Table summary
+
+The table below shows a summary of policies and if it makes sense to configure them on:
+- sidecar
+- egress
+- on both
+
+and if it makes sense for the user to pick where.
+
+| Policy                    | On sidecar    | On egress              | On both | User configurable |
```
jakubdyszkiewicz, line 79: what is user configurable? I don't get it

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,275 @@
...
+| MeshMetric                | Yes           | Yes                    | Yes     | Maybe             |
+| MeshProxyPatch            | Yes           | Yes                    | Yes     | Yes               |
+| MeshRateLimit             | Probably not  | Yes                    | No      | No                |
+| MeshRetry                 | Yes           | Maybe                  | No      | Maybe             |
+| MeshTimeout               | Yes           | Maybe                  | No      | No                |
+| MeshTrace                 | Yes           | Yes                    | Yes     | Maybe             |
+| MeshTrafficPermission     | Maybe         | Yes                    | Maybe   | Maybe             |
+| MeshLoadBalancingStrategy | Yes           | Yes                    | No      | Maybe             |
+| MeshTCPRoute              | Yes           | Yes                    | No      | No                |
+| MeshHTTPRoute             | Yes           | Yes                    | No      | No                |
```
jakubdyszkiewicz, line 94: how would routes work on egress only? What's the use case?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,275 @@
...
+#### MeshCircuitBreaker
+
+##### Old ExternalService behaviour
+
+Circuit Breaker is configured on the sidecar always.
+
+##### New behaviour
+
+If on egress
+- services from one zone will act as an aggregate - if a CB triggers by service-1 then service-2 will also see that result immediately
```
jakubdyszkiewicz, line 115: can you elaborate on this one? Are you saying that CB configured on envoy cluster "service-1" is carried over to envoy cluster "service-2?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: user can pick where it's placed (something like `proxyType: egress` and `proxyType: sidecar`)

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,275 @@
...
+Chosen option for configuration placement: ?
+
+## Pros and Cons of the Options
+
+### Running traffic through egress only
+
+- there is a bit of operational cost to running egress
+- it's a bit tricky to set up on universal
+- we've got a bit of a gap in other functionalities on egress (configuration, observability)
+- it's more probable to hit scalability issues than with sidecar (egress needs to know about all services / external services)
```
slonka, line 49:
```suggestion
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,275 @@
...
+- we'd need to relax the requirement of mTLS (it's only needed for services) to run egress
```
slonka, line 50:
```suggestion
- mTLS is required to run egress
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,275 @@
...
+- there is additional network hop
+- there is a load balancing issue with TCP traffic using zone egress when we make a connection and change the load
+  balancing algorithm we have to wait for the connection to be re-created for the reconfiguration to take place
```
slonka, line 53:
```suggestion
- for HTTP this wouldn't apply but for TCP there is a load balancing issue with TCP traffic using zone egress when we make a connection and change the load
  balancing algorithm we have to wait for the connection to be re-created for the reconfiguration to take place
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: can you elaborate on the problems there?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,274 @@
...
+
+The table below shows a summary of policies and if it makes sense to configure them on:
+- sidecar
+- egress
+- on both
+
+and if it makes sense for the user to pick where.
+
+| Policy                    | On sidecar    | On egress              | On both | User configurable |
+|---------------------------|---------------|------------------------|---------|-------------------|
```
slonka, line 79: @slonka to rework this table to reflect to / from policies being applied on `sidecar` / `egress`

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,274 @@
...
+- there is additional network hop
+- with the introduction of MeshPassthrough the barrier for external traffic is lower and some cases can be covered by that
+
+Conclusion: **it seems like we're not cutting any serious use case**.
+
+### Running traffic through sidecar only
+
+- all the problems of running through egress are not applicable (operational cost, extra networking hop, scalability etc.)
+- no single exit point for all the external traffic
+- if we put credentials in the ExternalService then these credentials are accessible to the app owners (need to check Envoy secrets)
```
slonka, line 59: after the meeting I checked and I think we can just use `generic_secret` for this

```suggestion
- if we put credentials in the ExternalService then these credentials are accessible to the app owners if not delivered via generic_secret
```

slonka:
reverted back to draft until I make changes.

lahabana:
The title is incorrect for the PR

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: added a note on this

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: what I'm thinking about is a scenario like this:

`serivce-a` consumes `es-1` through egress
`service-b` consumes `es-1` through egress

there is a bug in `service-a` that causes it to create lots of unnecessary connections, the circuitbreaker trips and starts rejecting connections for both `service-a` and `service-b` because on egress there is only one `es-1` cluster. We could theoretically work around this by having a cluster per service but that wouldn't scale for very long.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,371 @@
...
+| MeshLoadBalancingStrategy | Yes           | Yes                    | No      | Maybe             |
+| MeshTCPRoute              | Yes           | Yes                    | No      | No                |
+| MeshHTTPRoute             | Yes           | Yes                    | No      | No                |
+
+In paragraphs below I go deeper into each policy.
+
+##### MeshAccessLog
+
+MeshAccessLog is not applied on egress at all, and we don't provide a way to configure it.
+I don't see any reason not to do this though.
```
lahabana, line 107: I guess one issue is that zoneEgress is multi-mesh whereas other components usually are not.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
lahabana: Is this just a `to` vs `from` case?

In the `to` the CB is applied on the envoy cluster of `service-X`'s sidecar so A and B won't conflate.
In the `from` the CB is applied on the envoy cluster of `es-1` on each ZoneEgress therefore it applies to all clients.

I don't see how this is any different than regular dataplanes.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,371 @@
...
+##### MeshFaultInjection
+
+###### Old ExternalService behaviour
+
+FaultInjection and MeshFaultInjection is configured for ExternalService on egress inbound.
+
+###### New behaviour
+
+If on sidecar:
+- less traffic in the cluster (it won't even exit the pod)
```
lahabana, line 136: It's a `from` policy so if supported it should be applied on the Egress no?

I thought the whole point of making Egress compulsory was to make MES a lot closer to regular services.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,371 @@
...
+  to:
+    - targetRef:
+        kind: MeshExternalService
+        name: httpbin
+      default:
+        backends:
+          - type: File
+            file:
+              path: /tmp/access.log
+```
```
lahabana, line 366: I think this is not acknowledging that there are multiple problems and we've decided not to address some of it:

Regular app:
```mermaid
flowchart LR
  SrcApp --A--> SrcSidecar --B--> Network --C--> DestSidecar --D--> DestApp 
```
MES with Egress:
```mermaid
flowchart LR
  SrcApp --A--> SrcSidecar --B--> Network --C--> Egress --D--> ExternalService
```

We see there are 4 locations where Envoy plays a role.

B is done with `to`
C is done with `from`

The targetRef inside to and from only identify the type of traffic.

There is nowhere the ability to apply policies to A and D we never really talked about it because in the regular use case this is all on localhost so it's not that useful. Here you are pointing the 1 place where maybe it would be useful, I'd vote for judging out of scope.

I think that's really all there is with policy matching.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,371 @@
...
+
+Another option is to use `MeshExternalService` as a top level target + `proxyType: egress` + `name`.
+
+```yaml
+kind: MeshAccessLog
+spec:
+  targetRef:
+    kind: MeshExternalService
+    proxyType: egress
+    name: httpbin
```
slonka, line 325: This a bit depends on https://github.com/kumahq/kuma/pull/10858/files and  https://github.com/kumahq/kuma/pull/10858/files#diff-95c559a0f6155d986d45f294106f1a5a789e99bd0c0c93dae4405724efc76f3cR47-R49 . I guess we could say we're selecting an egress sidecar.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: > Here you are pointing the 1 place where maybe it would be useful, I'd vote for judging out of scope.

👍

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: > It's a from policy so if supported it should be applied on the Egress no?
> 
> I thought the whole point of making Egress compulsory was to make MES a lot closer to regular services.

It is currently an inbound policy but this could be also placed on an outbound. So while brainstorming different ideas I've put it here just so it's documented. And to me it makes sense, I don't see too many differences on putting this on the destination inbound vs source outbound. You get the same result minus network hop.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: > I don't see how this is any different than regular dataplanes.

I don't think it is and I don't think I claimed anywhere that it is.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: does multi-mesh play a role here?

lobkovilya: MeshAccessLog is configured per TCPProxy (or HTTPConnectionManager), so we can set
```
kind: MeshAccessLog
metadata:
  labels:
    kuma.io/mesh: mesh-1
spec:
  targetRef:
    kind: ZoneEgress
    name: zoneegress-1
```
only for FilterChains related to `mesh-1`.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
lukidzi: I believe we shouldn't apply it to the outbound and should stick to `MeshFaultInjection` as an inbound policy. My argument is that you won't be able to validate if your policies (`MeshRetry`, `MeshTimeout`, etc.) work correctly because they won't be used.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: Good to note this but the table is not about what currently works for ES. Can I resolve this conv?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,347 @@
...
+##### MeshFaultInjection
+
+###### Old ExternalService behaviour
+
+FaultInjection and MeshFaultInjection is configured for ExternalService on egress inbound.
+
+###### New behaviour
+
+If on sidecar:
+- less traffic in the cluster (it won't even exit the pod)
```
slonka, line 136:
```suggestion
- less traffic in the cluster (it won't even exit the pod)
- you won't be able to validate if your policies (MeshRetry, MeshTimeout, etc.) work correctly because they won't be triggered
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: Just a reminder: this is not a proposed change, just evaluation. I added this concern as a note.

lukidzi (approved):

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,348 @@
...
+    sectionName: httpbin
+  from:
+    - targetRef:
+        kind: Mesh
+      default:
+        backends:
+          - type: File
+            file:
+              path: /tmp/access.log
+```
```
jakubdyszkiewicz, line 316: Isn't just
```
kind: MeshAccessLog
spec:
  targetRef:
    kind: Mesh
    proxyType: egress # or just kind: ZoneEgress
  to:
    - targetRef:
        kind: MeshExternalService
        name: httpbin
      default:
        backends:
          - type: File
            file:
              path: /tmp/access.log
```
?

It's configured on one filter chain anyways. Basically what Ilya suggested

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
jakubdyszkiewicz: > there is a bug in service-a that causes it to create lots of unnecessary connections, the circuitbreaker trips and starts rejecting connections for both service-a and service-b because on egress there is only one es-1 cluster

I think that's acceptable. That's the cost of centralized access.

btw. in this example it would be very hard to hit this, because we do http2 between envoys

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
jakubdyszkiewicz: 1) How do you target egress and specific MeshExternalService
2) Can you redirect just to other MeshExternalService or to MeshServices as well?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: I think MAL is not a great example here - MeshTrafficPermission is better. There is no `to` in MTP and you still want to select which services can reach which MES.

slonka:
```
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  namespace: kuma-system
  name: allow-orders
spec:
  targetRef:
    kind: Mesh
    proxyType: egress
    sectionName: httpbin # we need this field to know which MES to allow traffic from proxies in `from` section
  from:
    - targetRef:
        kind: MeshSubset
        tags:
           kuma.io/service: orders
      default:
        action: Allow
```

If we wanted to use "to" in MTP:

```
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  namespace: kuma-system
  name: allow-orders
spec:
  targetRef:
    kind: Mesh
    proxyType: egress
  to:
    - targetRef:
        kind: MeshExternalService
        name: httpbin
      default:
        action: Allow
```

we have a problem that we don't know which services are allowed to talk to this MES (`orders` in previous). This config does not make it possible.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: > 1. How do you target egress and specific MeshExternalService

it's the same question as in https://github.com/kumahq/kuma/pull/10921/files#diff-be8076fceda3a49a1fa26af0d8cab5bfb664c6e87fd1a0dccbb4fac043629484R270 section. It's either: `kind: Mesh`, `proxyType` + `sectionName` or use `MeshExternalService` as a top level target + `proxyType: egress` + `name`.

> 2. Can you redirect just to other MeshExternalService or to MeshServices as well?

I don't think you should be able to redirect to other Kinds.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,348 @@
...
+| MeshFaultInjection        | Maybe         | Yes                    | No      | No                |
+| MeshHealthCheck           | Yes           | Maybe                  | No      | No                |
+| MeshMetric                | Yes           | Yes                    | Yes     | Maybe             |
+| MeshProxyPatch            | Yes           | Yes                    | Yes     | Yes               |
+| MeshRateLimit             | Probably not  | Yes                    | No      | No                |
+| MeshRetry                 | Yes           | Maybe                  | No      | Maybe             |
+| MeshTimeout               | Yes           | Maybe                  | No      | No                |
+| MeshTrace                 | Yes           | Yes                    | Yes     | Maybe             |
+| MeshTrafficPermission     | Maybe         | Yes                    | Maybe   | Maybe             |
+| MeshLoadBalancingStrategy | Yes           | Yes                    | No      | Maybe             |
```
slonka, line 98: ideally this should be on egerss, but we could have validation to only have 'from' if using `proxyType: egress` (similar story with the gateway for fault injection)

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: it's `to` policy so it's on the sidecar

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,407 @@
...
+Configuration (like timeouts, retries, circuit breakers) of MeshExternalService can also be applied both on a sidecar and on an egress.
+
+This MADR will go into the details of:
+- limitations / features that occur / are possible when running the traffic directly from sidecar vs through egress
+- each policy type and the placement to figure out what is the best option
+
+## Decision Drivers
+
+* not restricting any useful use-cases
+* flexible enough that we don't limit users in scenarios where it makes sense to configuration in both/one place
```
lobkovilya, line 20: nit: "not restricting useful use-cases" and "don't limit users in scenarios" sound like the same decision driver :)

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,407 @@
...
+- MeshGlobalRateLimit
+- MeshFaultInjection
+
+```yaml
+kind: MeshTrafficPermission
+spec:
+  targetRef:
+    kind: MeshExternalService
+    name: httpbin
+    proxyType: egress
```
lobkovilya, line 303: nit: proxyType is an array

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,407 @@
...
+
+##### Empty `proxyType`
+
+Currently, egress is not selected by `kind: Mesh` targetRef, but missing `proxyType` means both `gateway` and `sidecar`.
+We don't want to assume empty `proxyType` selecting `egress` because we want greater control over what is configured and to make sure user is aware that egress is affected.
+To avoid breaking changes now and in the future we can extend this more easily.
+
+We will require `proxyType: egress` in validation:
+- on top level `MeshExternalService`
+- on to egress policies
```
lobkovilya, line 394: Not sure I understand this, what happens if I apply:
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshAccessLog
spec:
  targetRef:
    kind: Mesh
  to:
  - targetRef:
      kind: MeshExternalService
      name: backend
    default: ...
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,407 @@
...
+
+Configuration (like timeouts, retries, circuit breakers) of MeshExternalService can also be applied both on a sidecar and on an egress.
+
+This MADR will go into the details of:
+- limitations / features that occur / are possible when running the traffic directly from sidecar vs through egress
+- each policy type and the placement to figure out what is the best option
+
+## Decision Drivers
+
+* not restricting any useful use-cases
```
slonka, line 19:
```suggestion
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: removed one

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: talked about offline

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+
+If on egress
+- services from one zone will act as an aggregate - if a CB triggers by service-1 then service-2 will also see that result immediately
+  - this makes the configuration more sensitive to drastic changes
+  - if a MES starts failing on one endpoint /example1 that is only consumed by service-1 it will also trigger for service-2 that might be happily consuming endpoint /example2
+  - this could be mitigated by re-defining a MeshExternalService under a different name (a bit awkward)
+
+###### Verdict
+
+There might be situations where it makes sense on egress but seems like it's more natural on the sidecar.
```
jakubdyszkiewicz, line 141: "so more natural on the sidecar", but in decision outcome we say `to + proxyType: egress`

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+###### Old ExternalService behaviour
+
+Both Retry and MeshRetry are configured on the sidecar
+
+###### New behaviour
+
+If on egress:
+- when an ExternalService is unavailable a retry will only occur `numRetries` tries - not `numRetries` * instances (in sidecar case) - which is good because we're not overwhelming the service
+- same story as circuit breaker in terms of acting as an aggregate / sensitivity
+- we definitely shouldn't configure it both on sidecar and egress, that way we'll have squared retries
```
jakubdyszkiewicz, line 215: there is no verdict as with other policies

jakubdyszkiewicz: same with other below. Is this intentional?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: MeshAccessLog is `to&from` policy, here we're talking about `MeshHealthCheck`, `MeshLoadBalancingStrategy`, `MeshCircuitBreaker`. Right @lukidzi ?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: Yes, the verdicts got out of date and I forgot to change some. Now we have the table so is it ok to remove the ones under each policy?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
jakubdyszkiewicz: I think Ilya's question is where config is applied with the policy he proposed:
1) It configures MAL only on sidecar
2) It configures MAL on sidecar and egress

Which one is it?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: we were still figuring this out and the verdicts got outdated as we went along - should I remove them now that we have a table sumarizing them all?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+
+If on egress
+- services from one zone will act as an aggregate - if a CB triggers by service-1 then service-2 will also see that result immediately
+  - this makes the configuration more sensitive to drastic changes
+  - if a MES starts failing on one endpoint /example1 that is only consumed by service-1 it will also trigger for service-2 that might be happily consuming endpoint /example2
+  - this could be mitigated by re-defining a MeshExternalService under a different name (a bit awkward)
+
+###### Verdict
+
+There might be situations where it makes sense on egress but seems like it's more natural on the sidecar.
```
slonka, line 141:
```suggestion
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: I think we elaborated on the topic, feel free to re-open if you think it needs more info.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+| MeshLoadBalancingStrategy | Yes           | Yes                    | No      | Maybe             |
+| MeshTCPRoute              | Yes           | Yes                    | No      | No                |
+| MeshHTTPRoute             | Yes           | Yes                    | No      | No                |
+
+In paragraphs below I go deeper into each policy.
+
+##### MeshAccessLog
+
+MeshAccessLog is not applied on egress at all, and we don't provide a way to configure it.
+I don't see any reason not to do this though.
```
slonka, line 123:
```suggestion
I don't see any reason not to do this though.
To define a specific mesh we can use `kuma.io/mesh` label.
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+
+###### New behaviour
+
+If on sidecar:
+- less traffic in the cluster (it won't even exit the pod)
+- you won't be able to validate if your policies (MeshRetry, MeshTimeout, etc.) work correctly because they won't be triggered
+
+###### Verdict
+
+Probably should stay on egress.
```
slonka, line 157:
```suggestion
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+
+###### New behaviour
+
+If on egress:
+- Seems to be the same story as circuit breaker in terms of acting as an aggregate / sensitivity
+- Doing it on the egress could cause less traffic to MES (assuming there is fewer instances of egress than services HC-ing)
+
+###### Verdict
+
+To me, it makes a little bit more sense (than circuit breaker) to have HC be configurable on where it's placed
```
slonka, line 173:
```suggestion
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+
+###### New behaviour
+
+If on sidecar:
+- limit is multiplied by the number of instances (which is bad)
+- less traffic in the mesh (traffic does )
+
+###### Verdict
+
+It seems that it only makes sense on egress, we don't want the ratelimit to be changing with number of instances.
```
slonka, line 202:
```suggestion
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: I would say both. A better example is what @lukidzi mentioned with timeouts. If you apply a timeout you probably don't want different timeout on the sidecar vs the egress right?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: Added note about this.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
lobkovilya: But then it contradicts the statement, no?
> We don't want to assume empty `proxyType` selecting `egress` because we want greater control over what is configured and to make sure user is aware that egress is affected.

For policies like MeshRetry we probably don't want to have both `sidecar` and `egress` by default.

slonka: You're right :| I think we need to always require `proxyType` when `MeshExternalService` is referenced and document the Timeout case.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+#### For outbound policies on the sidecar, for inbound policies on egress
+
+It naturally fits our targetRef mechanism.
+It does not make sense to apply inbound policies on the sidecar since there is no inbound for it.
+Egress plays a role of a "sidecar" that is inside the mesh, and we can target it's inbound.
+Configuring outbound policies on the sidecar avoids the problem with different clients influencing each other (see [MeshCircuitBreaker](#meshcircuitbreaker)).
+
+##### How to configure it
+
+###### From only policy (policies that do not have to)
```
michaelbeaumont, line 289:
```suggestion
###### `from` only policy (policies that do not have `to`)
```

I find it quite exhausting to read to and from everywhere if they aren't formatted, they're too common as normal words

actually

```suggestion
###### `from` only policy
```

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshRetry
+metadata:
+  name: backend
+spec:
+  targetRef:
+    kind: Mesh
+    proxyType: ["sidecar"]
```
michaelbeaumont, line 331: why should this be necessary?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+
+We will require `proxyType` in validation whenever `MeshExternalService` is referenced.
+
+#### Only on the egress
+
+There are policies that make sense to be in both places and the ones like MeshTimeout that only have drawbacks on egress.
+
+#### Only on sidecar
+
+There are policies that make more sense on egress and in both places.
```
michaelbeaumont, line 401: What do these lines mean?

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,406 @@
...
+##### MeshRetry
+
+###### Old ExternalService behaviour
+
+Both Retry and MeshRetry are configured on the sidecar
+
+###### New behaviour
+
+If on egress:
+- when an ExternalService is unavailable a retry will only occur `numRetries` tries - not `numRetries` * instances (in sidecar case) - which is good because we're not overwhelming the service
```
michaelbeaumont, line 214: > - when an ExternalService is unavailable a retry will only occur `numRetries` tries - not `numRetries` * instances

Is this true? I can't quite find where this s documented/implied in the docs. Aren't retries tracked per request?

slonka: > Is this true? I can't quite find where this s documented/implied in the docs. Aren't retries tracked per request?

I think you're right. I confused that with HC / outlier detection which will behave differently

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: Just wanted to document that we considered configuring everything on egress or on sidecar.

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
slonka: I think it's easier for the user to understand where the policy is applied. I think there are a two options:
- explicit (always state `proxyType`)
- implicit (based on `kind: MeshExternalService` and policy type we implicitly place the configuration on the appropriate component)

I think implicit can get us in trouble fast. Currently "[Empty list: Defaults to targeting all data plane proxies](https://kuma.io/docs/2.8.x/policies/targetref/#understanding-targetref-policies)" so that would mean without `proxyType` we would place the configuration on both `sidecar` and `egress` - that's not good.

On the other hand maybe it's fine to be implicit for MES and we just document it.

lukidzi: We decide to configure things implicit

lukidzi (approved):

lobkovilya (approved):

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -0,0 +1,405 @@
...
+- MeshTrafficPermission
+- MeshRateLimit
+- MeshGlobalRateLimit
+- MeshFaultInjection
+
+```yaml
+kind: MeshTrafficPermission
+spec:
+  targetRef:
+    kind: MeshExternalService
```
michaelbeaumont, line 297: Shouldn't `spec.targetRef` only target the proxies the config affects? How does this fit in with all the ideas about "consistency" and https://github.com/kumahq/kuma/pull/10858?
`MeshExternalService` is just a destination, why don't we allow `spec.targetRef.kind: MeshService`?
We changed this already and it was apparently a big mistake, what exactly does `spec.targetRef` mean?

michaelbeaumont: How do we integrate this with https://kuma.io/docs/2.8.x/policies/targetref/#understanding-targetref-policies for example

lukidzi: Once we agreed to select dataplane but it seems we don't have a way of targeting ZoneEgress and a specific part of it which is required here. It requires a design but it's out of the scope of this MADR

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
michaelbeaumont: Ah ok then maybe like

```suggestion
#### Only configuring policies on the egress

This doesn't work since there are policies that make sense to be in both places and the ones like MeshTimeout that only have drawbacks on egress.

#### Only configuring policies on sidecar

This doesn't work because there are policies that make more sense on egress and in both places.
```

because as is it just reads like a non sequitur

michaelbeaumont (approved):

docs/madr/decisions/056-meshexternalservice-and-zoneegress.md
```diff
@@ -377,18 +364,20 @@ spec:
...
+
+When `mTLS` is enabled (which is required with `Egress`), we require users to explicitly create a `MeshTrafficPolicy` to enable traffic. In the case of `MeshExternalService` and the current lack of targeting `Egress`, we have decided to, by default, allow all traffic through `Egress` for `MeshExternalService`.
+
+However, users will be able to disable this behavior on the `Mesh` object.
+```proto
+// Routing defines configuration for the routing in the mesh
+message Routing {
+  ...
+  // Changes the default behavior to not include the allow RBAC filter.
+  bool forbidMeshExternalServiceAccess = 3;
```
michaelbeaumont, line 379: maybe

```suggestion
  bool defaultForbidMeshExternalServiceAccess = 3;
```

for a future when it can be set to true and controlled via policy?

Issue #10896, opened 2024-07-15
Seamless permissive mTLS with MeshService


michaelbeaumont:
### Description

- TLS readiness in status probably

jakubdyszkiewicz:
Done by #10924

PR #10924, opened 2024-07-17
docs(MADR): permissive mtls and mesh service

jakubdyszkiewicz:
### Checklist prior to review

xref https://github.com/kumahq/kuma/issues/10896

### DIFF
--- a/docs/madr/decisions/058-meshservice-permissive-tls.md
+++ b/docs/madr/decisions/058-meshservice-permissive-tls.md
@@ -1,0 +1,81 @@
+# Permissive mTLS and MeshService
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/10896
+
+## Context and Problem Statement
+
+Kuma supports permissive mTLS, which is an mTLS mode in which applications can accept mTLS and plaintext traffic.
+This is useful for introducing mTLS in the cluster without any downtime, because distributing mTLS certs takes more than 0ms,
+the client needs to know if it can send mTLS traffic to a server.
+
+They way we solved this is we store `issuedBackend` that indicates if mTLS cert was issued for DPP.
+Then, we take a list of `DataplaneInsights` and aggregate this information in `ServiceInsight`.
+Then we determine if the service is ready to accept mTLS traffic with the condition of
+```
+issued certificates == data plane proxies offline + data plane proxies online
+```
+
+This approach has one problem. When we introduce new Dataplane of the service, then we might get into situation where `ServiceInsight` is updated before Dataplane is even online or was issued the cert.
+In this case the condition described above is false, and we stop sending mTLS traffic until DPP is online, cert was issued and new reconciliation of `ServiceInsight` happens.
+
+With the introduction of MeshService we cannot rely on `ServiceInsight`, because it's an aggregation of `kuma.io/service`.
+
+## Decision Drivers
+
+Feature parity for MeshService
+
+## Considered Options
+
+* Option 1 - Use DataplaneInsight
+* Option 2 - Store it in the Dataplane object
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/058-meshservice-permissive-tls.md
```diff
@@ -0,0 +1,81 @@
...
+```
+status:
+  tls:
+    status: Ready
+```
+
+However, we can be a bit smarter how we set it.
+We know that when you enable mTLS all new data plane proxies always start with a certificate.
+Envoy is only ready when Secret with cert is delivered.
+Assuming we don't disable mTLS, the status can only go one way from `NotReady` to `Ready`.
```
slonka, line 70: so what happens when we disable mTLS? is there a period where we wait for all the clients to go back to sending plaintext?

docs/madr/decisions/058-meshservice-permissive-tls.md
```diff
@@ -0,0 +1,81 @@
...
+
+MeshService status updater can then have a following logic
+* If mTLS is disabled, set `tls.status` to `NotReady`
+* If mTLS is enabled and `tls.status` is `NotReady`, check `issued certificates == data plane proxies offline + data plane proxies online`. If true, set `tls.status` to `Ready`
+
+We can store this is `tls.status`. It's synced to global for visibility.
+It's not synced cross-zone, because it's only useful in the context of local zone. For cross zone traffic we require mTLS anyway.
+
+**How would MeshMTLS policy affect this?**
+It's not decided if MeshMTLS will expose selective mTLS (mTLS only for a subset of proxies).
```
slonka, line 80: I think that was the core idea behind it - @lahabana can you confirm?

docs/madr/decisions/058-meshservice-permissive-tls.md
```diff
@@ -0,0 +1,81 @@
...
+If so, we would need to do policy matching of proxies selected in MeshService selector to check if mTLS is enabled or not.
```
slonka, line 81: and we could just feed that into the algorithm so it would be `mTLS is disabled for this service` right?

lukidzi (approved):

docs/madr/decisions/058-meshservice-permissive-tls.md
jakubdyszkiewicz: Isn't it more about setting permissive mTLS for individual services?
I don't think I get the use case for enabling mTLS for a subset of the mesh

docs/madr/decisions/058-meshservice-permissive-tls.md
jakubdyszkiewicz: yes

docs/madr/decisions/058-meshservice-permissive-tls.md
jakubdyszkiewicz: We don't handle this at the moment.

jakubdyszkiewicz: Or rather, that would be still Ready / NotReady. It would probably be a setting on MeshMTLS, we would still need to keep the backend but mark it as "draining".

michaelbeaumont (approved):

docs/madr/decisions/058-meshservice-permissive-tls.md
slonka: > permissive mTLS for individual services

yes, I think this it what it was.

> I don't think I get the use case for enabling mTLS for a subset of the mesh

In the previous company I think that the financial services had a strict mTLS policy and other services could've been on permissive + audit logs.

slonka (approved):

PR #10927, opened 2024-07-18
fix(kuma-cp): pass future meta to Validate when creating a resource

michaelbeaumont:
At the moment the [MeshService validator](https://github.com/kumahq/kuma/blob/master/pkg/core/resources/apis/meshservice/api/v1alpha1/validator.go) fails when Create is called because the meta is `nil` and the future name of the resource is missing.

needed to get rid of [this commit in #10917](https://github.com/michaelbeaumont/kuma/commit/7bcee67d58d0deb7e2db192d2a3c3133a5aacbc4)

### Checklist prior to review

### DIFF
--- a/pkg/core/resources/manager/manager.go
+++ b/pkg/core/resources/manager/manager.go
@@ -59,9 +59,7 @@ func (r *resourcesManager) Create(ctx context.Context, resource model.Resource, 
 	if err := model.Validate(resource); err != nil {
 		return err
 	}
-	if existingMeta == nil {
-		resource.SetMeta(existingMeta)
-	}
+	resource.SetMeta(existingMeta)
 
 	var owner model.Resource
 	if resource.Descriptor().Scope == model.ScopeMesh {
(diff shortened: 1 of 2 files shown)


pkg/core/resources/manager/manager.go
```diff
@@ -49,10 +49,19 @@ func (r *resourcesManager) List(ctx context.Context, list model.ResourceList, fs
...
+	// Create temporary meta so validation can see the meta that the store will set
+	existingMeta := resource.GetMeta()
+	if existingMeta == nil {
+		resource.SetMeta(metaFromCreateOpts(resource.Descriptor(), *opts))
+	}
 	if err := model.Validate(resource); err != nil {
 		return err
 	}
-	opts := store.NewCreateOptions(fs...)
+	if existingMeta == nil {
```
jakubdyszkiewicz, line 62: nit: technically you could drop this if and call it always

jakubdyszkiewicz (approved):

PR #10945, opened 2024-07-23
ci(.github): enable self hosted runners for AMD64 E2E tasks

jijiechen:
### Checklist prior to review

We disabled these runners after we ran into a few E2E flakinesses. 

Those issues turned out to be caused by "docker hub pull rate limiting". So here I'm supporting using the provided docker credential in both kind/k3d clusters to prevent pulls from being limited. 

This relies on an environment variable `DOCKERHUB_PULL_CREDENTIAL` (in the format of `username:password`) to be set on the runner. If it's not set, the script will still run, but without the credential attached.

  - fixes https://github.com/kumahq/kuma/issues/10585
  - Confirmed
  - Manual tested
    - Added
  - No
  - No


jijiechen:
The script tries to provide credential to kind/k3d. Here are successful runs with the changes in this PR applied:
1. No credentials provided on the CI: it does nothing on existing CI: https://github.com/kumahq/kuma/pull/10945/files#diff-034cf00afd1a7b01eab427da6af7a51278aa40f4a74e1da9079a5680646360e7R4
2. Invalid credential: https://github.com/kumahq/kuma/actions/runs/10038367883/job/27740475708#step:13:183
3. Correct credential: https://github.com/kumahq/kuma/actions/runs/10038637341/job/27781485596#step:13:56

tools/ci/prepare-pull-credentials.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -e
+
+if [[ "$DOCKERHUB_PULL_CREDENTIAL" == "" ]]; then
+    >&2 echo "No docker pull credential information specified"
+else
+    DOCKER_USER=$(echo "$DOCKERHUB_PULL_CREDENTIAL" | cut -d ':' -f 1)
+    DOCKER_PWD=$(echo "$DOCKERHUB_PULL_CREDENTIAL" | cut -d ':' -f 2)
+    echo -n "$DOCKER_PWD" | docker login -u "$DOCKER_USER" --password-stdin > /dev/null
+    cat <<EOF > "$HOME/.docker/k3d-registry.yaml"
```
jijiechen, line 10: Generate a registry file to be used by `k3d`
docs: https://k3d.io/v5.2.0/usage/registries/#authenticated-registries

tools/ci/prepare-pull-credentials.sh
```diff
@@ -0,0 +1,28 @@
...
+  registry-1.docker.io:
+    auth:
+      username: $DOCKER_USER
+      password: $DOCKER_PWD
+EOF
+    echo "$HOME/.docker/k3d-registry.yaml"
+
+    echo "{\"auths\":{\"https://index.docker.io/v1/\":{\"auth\":\"$CRED\"}}}" > "$HOME/.docker/kind-config.json"
+    EXTRA_MOUNTS="[ {\"containerPath\": \"/var/lib/kubelet/config.json\", \"hostPath\": \"$HOME/.docker/kind-config.json\"}]"
+    for FILE in test/kind/*.yaml; do
```
jijiechen, line 21: Add `extraMounts` to kind cluster configuration files so that it mounts the `kind-config.json` generated above.
docs: https://kind.sigs.k8s.io/docs/user/private-registries/#add-credentials-to-the-nodes

.github/workflows/_test.yaml
```diff
@@ -103,18 +103,12 @@ jobs:
           ALL_RUNNERS: |-
             {
               "trusted_builds": {
```
jijiechen, line 105: Simplifying the existing runners expressions. This does not break existing code using this logic.

tools/ci/prepare-pull-credentials.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -e
+
+if [[ "$DOCKERHUB_PULL_CREDENTIAL" == "" ]]; then
```
lahabana, line 4: Why can't this just be invoked as part of the makefile?

jijiechen: You mean embeding this whole script file to the makefile `docker.mk` or `e2e.new.mk`? 
I think the yq expressions underneath are too complicated to be embeded.

lahabana (changes requested):
I think hiding the complexity inside the makefile is problem much better here.
Also it seems like a lot of the yq stuff we're doing has no logic. Why can't we just have a config file?

We're also modifying files that are actually checkedin which is likely to cause failures to things like `make check`

tools/ci/prepare-pull-credentials.sh
jijiechen: If you mean we can put the `if` check into the makefile, then yes, we can do it.

.github/workflows/_test.yaml
```diff
@@ -103,18 +103,12 @@ jobs:
...
+                "amd64": "ubuntu-latest-kong",
+                "arm64": "ubuntu-latest-arm64-kong"
               },
               "untrusted_builds": {
-                "amd64": [
-                  "ubuntu-latest"
-                ],
-                "arm64": []
+                "amd64": "ubuntu-latest",
+                "arm64": ""
```
lahabana, line 111: Wasn't there a way to do this with much simpler syntax? No yq just github expressions?

jijiechen: Let me think about it

.github/workflows/_e2e.yaml
```diff
@@ -103,6 +108,17 @@ jobs:
...
+          
+          function on_exit()
+          {
+            docker logout docker.io
+          }
+          K3D_REGISTRY_FILE=$(tools/ci/prepare-pull-credentials.sh)
+          if [[ "$K3D_REGISTRY_FILE" != "" ]]; then
+            export K3D_REGISTRY_FILE
+            trap "on_exit" EXIT
+          fi
```
lahabana, line 120: What I mean is that instead of calling this here explicitly it could just be part of the Makefile. With targets just like `kind/load/images` and `k3d/load/images`. For .e.g: `k3d/setup-docker-credentials` and `kind/setup-docker-credentials` it's closer to what's done everywhere

.github/workflows/_test.yaml
lahabana: https://github.com/kumahq/kuma/pull/10945#discussion_r1687751862

tools/ci/prepare-pull-credentials.sh
lahabana: If you feel the need to add a comment to a PR to explain your code then it should probably just be a comment in the code in the first place

tools/ci/prepare-pull-credentials.sh
```diff
@@ -0,0 +1,28 @@
...
+configs:
+  registry-1.docker.io:
+    auth:
+      username: $DOCKER_USER
+      password: $DOCKER_PWD
+EOF
+    echo "$HOME/.docker/k3d-registry.yaml"
+
+    echo "{\"auths\":{\"https://index.docker.io/v1/\":{\"auth\":\"$CRED\"}}}" > "$HOME/.docker/kind-config.json"
+    EXTRA_MOUNTS="[ {\"containerPath\": \"/var/lib/kubelet/config.json\", \"hostPath\": \"$HOME/.docker/kind-config.json\"}]"
```
lahabana, line 20: Could we just have always the same config with an ignored file? Then your make target would create the file it needs systematically (with or without creds). This way using creds is not more complex and there are a lot less conditionals

.github/workflows/_test.yaml
jijiechen: For runners, I'll use GHA expressions.

mk/k3d.mk
```diff
@@ -79,8 +80,25 @@ k3d/network/create:
 		else docker network create -d=bridge $(KIND_NETWORK_OPTS) kind || true; fi && \
 		rm -f $(BUILD_DIR)/k3d_network.lock
 
+K3D_PULL_CREDENTIAL ?=
```
lahabana, line 83: Why not Always put the registry config? Just sometimes it hasn't got creds inside it...

.github/workflows/_e2e.yaml
```diff
@@ -103,6 +104,23 @@ jobs:
...
+          function on_exit()
+          {
+            docker logout docker.io
+          }
+          
+          # we pull a few images during the E2E run, sometimes we get rate-limited by docker hub
+          # to prevent this, we support specifying a pull credential here
+          if [[ "${{ secrets.DOCKERHUB_PULL_CREDENTIAL }}" != "" ]]; then
+            export K3D_PULL_CREDENTIAL="${{ secrets.DOCKERHUB_PULL_CREDENTIAL }}"
+            export KIND_PULL_CREDENTIAL="${{ secrets.DOCKERHUB_PULL_CREDENTIAL }}"
```
lahabana, line 117: You don't need to differentiate these because either you are K3D or KIND you are never both.

mk/k3d.mk
```diff
@@ -79,8 +80,25 @@ k3d/network/create:
...
+ifneq ($(strip $(K3D_PULL_CREDENTIAL)),)
```
lahabana, line 84: ifdef is likely what you want here

lahabana (approved):
If it passes LGTM!

Issue #9075, opened 2024-01-30
Remove KDS v1


lahabana:
### What happened?

It's been default for 2 versions! Remove in 2.7 🥳

Icarus9913:
on me

PR #6278, opened 2023-03-14
feat(kuma-cp): introduce KDS sync from Global to Zone service with de…

lukidzi:
…lta xDS sync

### Checklist prior to review

- [ ] [Link to relevant issue][1] as well as docs and UI issues --
- [ ] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [ ] Tests (Unit test, E2E tests, manual test on universal and k8s) --
- [ ] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [ ] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? --
- [ ] Do you need to explicitly set a [`> Changelog:` entry here](../blob/master/CONTRIBUTING.md#submitting-a-patch) or add a `ci/` label to run fewer/more tests?

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #5373, opened 2022-11-24
fix(kuma-cp): kds deadlock

jakubdyszkiewicz:
## The Context

KDS uses one gRPC bi-directional stream initiated from the client (zone CP) to the server (global CP). Then, under the hood, it creates two logical streams similar to xDS (DiscoveryRequest/Response etc.). We call this multiplexing. As a result, we can reuse parts of go-control-plane to have similar functionality to xDS.

gRPC is based on HTTP/2 therefore it has a functionality called flow control.
The sender sends the data to the receiver and then waits for `WINDOW_UPDATE` to send more.
With regular HTTP/2, there is a new stream for every request/response message exchange. Only one side is expected to send `DATA` frames (only frames affected by flow controls) at any given time.

With gRPC streaming it's different. Multiple messages are streamed on one HTTP/2 stream. Moreover, with bi-directional streaming, there are messages exchanged both ways at the same time.

## The Problem

When Zone CP connects to Global CP it sends DiscoveryRequest for resources synced from Global to Zone CP (policies mostly). In parallel, it waits for DiscoveryRequest in order to send DiscoveryResponse of resources synced from Zone to Global (Dataplane, DataplaneInsight etc.).

Global CP may currently sending [mux messages](https://github.com/kumahq/kuma/blob/master/api/mesh/v1alpha1/mux.proto) while not actively receiving mux messages from the Zone CP. Zone CP may be in the same situation.

If both are stuck on send operation, we are in deadlock. We cannot have a pattern of: send and wait until it's done, then receive a message.

When we look at a high-level logical KDS stream, we have such a pattern on the [client side](https://github.com/kumahq/kuma/blob/master/pkg/kds/client/sink.go#L45) and there is a similar pattern on the [server side](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/sotw/v3/server.go).

However when we operate on a low-level multiplexed gRPC stream we [do have separate goroutines](https://github.com/envoyproxy/go-control-plane/blob/main/pkg/server/sotw/v3/server.go) that in parallel try to send and receive mux messages.

The problem is that we can send and receive [at most two messages](https://github.com/kumahq/kuma/blob/master/pkg/kds/mux/session.go#L146) before we are stuck on processing those messages.

In the case of a high burst of big messages, we can actually end up in a deadlock. We do have gRPC keep alive, but it uses `PING` frames which are not throttled by flow control.

## The solution

The solution is to break the "send and wait until it's done, then receive a message." pattern.

[...]

Issue #4926, opened 2022-08-29
Remove vendored code from `go-control-plane`


michaelbeaumont:
### What happened?

We [vendored some code from `go-control-plane`](https://github.com/kumahq/kuma/issues/2719) for using non-Envoy resources and [we depend on concrete upstream structs instead of new interfaces](https://github.com/kumahq/kuma/pull/4877/files#diff-30f173d4378b9c8ba6e3de09febda533fc84c0d16c561d7d5f4ec5fff923c060R208). It will likely require some upstream changes to remove this code. See #2719 and https://github.com/kumahq/kuma/pull/4877#discussion_r954132160 for more context.

lahabana:
We now have a fork so there shouldn't be any vendored code left. Let's check and either remove this code or just close this issue if things are already removed

lahabana:
What's the state here? Can we remove the code?

lobkovilya:
> triage: we definitely should be able to either remove or move the code to fork

lukidzi:
Checked the tree — there is no vendored `go-control-plane` code left:

- `pkg/util/xds/v3/cache.go` (the forked `SnapshotCache`) was removed in #11590, `versioner.go` in #11516, and `resource_warming_forcer.go` in #11579. No file in the repo carries the `Copyright 2018 Envoyproxy Authors` header any more.
- The upstream ask from #2719 (envoyproxy/go-control-plane#489, type-oblivious cache) landed upstream as the `cache.ResourceSnapshot` interface, which we now implement instead of forking.

What is left is Kuma-specific glue over upstream interfaces, not vendored code: `util_xds_v3.SingleTypeSnapshot` (HDS/MADS) and `pkg/kds/v2/cache.Snapshot` (KDS) implement `cache.ResourceSnapshot`. Neither can move to `LinearCache` as suggested in #2719, because HDS keys snapshots per node and MADS per client id, while `LinearCache` holds a single global set per type. Neither belongs in the fork either — both are built on Kuma types (`KumaResource`, `core_model.ResourceType`).

Two leftovers did shadow upstream code and are cleaned up in #17834: `resourceTypeURL` in the xDS reconciler reimplemented `envoy_cache.GetResponseTypeURL`, and `SingleTypeSnapshotEqual` took the upstream interface only to type-assert back and panic.

[...]

PR #10946, opened 2024-07-23
chore(kds): remove kds v1

Icarus9913:
This PR aims to remove the deprecated KDS v1 source codes.

# Elaboration

## What is KDS?

KDS is Kuma Discovery Service which works for resource synchronization(Policies/Ingress..) between `Global` and `Zone`

#### v1

- With the [first kds server PR](https://github.com/kumahq/kuma/pull/839), we implemented something like the go-control-plane `SotW` protocol.

- With the [upgrade go-control-plane PR](https://github.com/kumahq/kuma/pull/1290), we use the go-control-plane `SotW` SDK directly.

#### v2

- KDS redesign: https://github.com/kumahq/kuma/issues/4354

- KDS v2 implementation: https://github.com/kumahq/kuma/pull/6278

#### Why did we migrate kds to v2 ?

1. The v1 KDS uses GRPC bidirectional stream which is hard to implement. (We also met a bug with that: https://github.com/kumahq/kuma/pull/5373)
2. The v1 KDS uses go-control-plane `SotW` protocol which can lead to a low efficiency when there are many data that need to be exchanged.

For more details you can read [KDS Redesign MADR](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/021-kds-redesign.md)  

## Why can't we just remove v1 since v2 merged?

At the v2 time, the `Delta` is something new and might have bugs. Also we should follow the compatibility between Global-CP and Zone-CP (Like a higher version Global-CP could work well with low version Zone-CP which follow the Kuma Upgrade rules `Since Kuma 1.4.x upgrades can be performed up to two minor versions`)

- [x] Normal Test to verify if the new building cluster works well.

```text
1. Create a Multizone ENV with k8s-global, k8s-zone, universal-zone with this branch version
2. Create a `MeshTrafficPermission` policy on global and check if the resource synced to zones
3. Create a `MeshFaultInjection` policy on k8s-zone and check if the resource synced to global
4. Create a `MeshCircuitBreaker ` policy on universal-zone and check if the resource synced to global
5. Make sure no error logs in each `control-plane`
```

- [x] Upgrade Test to verify if the 2 versions previous upgrading cluster works well.

```text
1. Create a Multizone ENV with k8s-global, k8s-zone, universal-zone with Kuma `2.7.6` version
2. Upgrade global to this branch version and create a `MeshTrafficPermission` policy on global then check if the resource synced to zones
3. Upgrade k8s-zone to this branch version and create a `MeshFaultInjection` policy on k8s-zone then check if the resource synced to global
4. Upgrade universal-zone to this branch version and create a `MeshCircuitBreaker` policy on universal-zone then check if the resource synced to global
5. Stop and wait a moment to restart universal Ingress to check if the `zoningress` resources disappeared and re-created on global
6. Make sure no error logs in each `control-plane`
```

## Further thinking
Remove vendored code from `go-control-plane`: https://github.com/kumahq/kuma/issues/4926

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- https://github.com/kumahq/kuma/issues/9075
- [x] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? -- Updated
- [x] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) -- No


Icarus9913:
Considering:
1. remove kds v1 proto file
2. remove snapshot reconciler which syncs the resources(policies/ingress/egress...)
3. remove some configuration flags
4. better add pprof to check how much memories we could save?
5. TODO...

pkg/kds/context/context.go
```diff
@@ -246,6 +245,8 @@ func GlobalProvidedFilter(rm manager.ResourceManager, configs map[string]bool) r
 				zone_tokens.SigningKeyPrefix,
 			}...)
 		case r.Descriptor().KDSFlags.Has(core_model.GlobalToAllButOriginalZoneFlag):
+			// TODO (Icarus9913): replace the function by model.ZoneOfResource(r)
+			// Reference: https://github.com/kumahq/kuma/issues/10952
 			zoneTag := util.ZoneTag(r)
```
Icarus9913, line 250: @lobkovilya Can we put `ZoneTag` function refactor in another PR? For this PR, I just removed the KDS v1 source codes without any break or fix.

In another PR, I want to do such things:
1. Replace the `ZoneTag` function by `model.ZoneOfResource`
2. Disalow the ZoneIngress resource creation in Global
3. At present, if I have a multizone ENV with 2 zones and the zone-1 ZoneIngress resource would be synced to global and futhur synced to zone-2 due to the `ZoneTag` function. And this behavior would be reverted in the next PR but make it clear with the compatibility with those synced `zone-1 ZoneIngress resource` in zone-2

lukidzi: Maybe we can get some refactor and move everything from v2 to normal package

lukidzi (approved):

pkg/kds/context/context.go
lobkovilya: Sure, I don't think removing KDS v1 has something to do with zone validation of the resources on Global. So yes, we definitely should work on this in another PR.

Issue #2313, opened 2021-07-06
Make logger names consistent in the xDS server package


jpeach:
### Summary

```
 git grep Log.*WithName pkg/xds/
pkg/xds/bootstrap/handler.go:var log = core.Log.WithName("bootstrap")
pkg/xds/cache/cla/cache.go:     claCacheLog = core.Log.WithName("cla-cache")
pkg/xds/cache/mesh/snapshot.go:var meshCacheLog = core.Log.WithName("mesh-cache")
pkg/xds/envoy/clusters/v3/health_check_configurer.go:           core.Log.WithName("health-check-configurer").Error(
pkg/xds/generator/outbound_proxy_generator.go:var outboundLog = core.Log.WithName("outbound-proxy-generator")
pkg/xds/server/callbacks/dataplane_lifecycle.go:var lifecycleLog = core.Log.WithName("xds").WithName("dp-lifecycle")
pkg/xds/server/callbacks/dataplane_status_sink.go:var sinkLog = core.Log.WithName("xds").WithName("sink")
pkg/xds/server/callbacks/dataplane_status_tracker.go:var statusTrackerLog = core.Log.WithName("xds").WithName("statusTracker")
pkg/xds/server/callbacks/dataplane_sync_tracker.go:     dataplaneSyncTrackerLog = core.Log.WithName("xds-server").WithName("dataplane-sync-tracker")
pkg/xds/server/callbacks/nack_backoff.go:var nackLog = core.Log.WithName("xds").WithName("nack-backoff")
pkg/xds/server/v3/components.go:var xdsServerLog = core.Log.WithName("xds-server")
pkg/xds/server/v3/reconcile.go: reconcileLog = core.Log.WithName("xds-server").WithName("reconcile")
pkg/xds/server/v3/resource_warming_forcer.go:var warmingForcerLog = xdsServerLog.WithName("warming-forcer")
pkg/xds/sync/components.go:     xdsServerLog = core.Log.WithName("xds-server")
pkg/xds/sync/dataplane_proxy_builder.go:var syncLog = core.Log.WithName("sync")
pkg/xds/sync/dataplane_watchdog_factory.go:     log := xdsServerLog.WithName("dataplane-sync-watchdog").WithValues("dataplaneKey", proxyId.ToResourceKey())
pkg/xds/template/proxy_template_resolver.go:    templateResolverLog = core.Log.WithName("proxy-template-resolver")
```

[...]

PR #10965, opened 2024-07-24
feat(kuma-cp): make loggers naming from xds package consistent

Automaat:
Fixes: #2313

### Checklist prior to review

### DIFF
--- a/pkg/xds/topology/outbound.go
+++ b/pkg/xds/topology/outbound.go
@@ -26,7 +26,7 @@ import (
 	envoy_tags "github.com/kumahq/kuma/pkg/xds/envoy/tags"
 )
 
-var log = core.Log.WithName("xds")
+var log = core.Log.WithName("xds").WithName("outbound")
 
 func BuildExternalServicesEndpointMap(
 	ctx context.Context,
--- a/pkg/xds/topology/outbound.go
+++ b/pkg/xds/topology/outbound.go
@@ -26,7 +26,7 @@ import (
 	envoy_tags "github.com/kumahq/kuma/pkg/xds/envoy/tags"
 )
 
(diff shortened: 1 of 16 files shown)


pkg/xds/topology/outbound.go
```diff
@@ -26,6 +26,8 @@ import (
 	envoy_tags "github.com/kumahq/kuma/pkg/xds/envoy/tags"
 )
 
+var log = core.Log.WithName("xds")
```
lukidzi, line 29: Should we add some more informations here, `core.Log.WithName("xds").WithName("outbound")`

michaelbeaumont: just be careful with this since it's a package level logger, we might have to use `outboundLog` i.e. one per file or create them in the components

Automaat: we can, updated

pkg/xds/topology/outbound.go
```diff
@@ -26,6 +26,8 @@ import (
 	envoy_tags "github.com/kumahq/kuma/pkg/xds/envoy/tags"
 )
 
+var log = core.Log.WithName("xds").WithName("outbound")
```
michaelbeaumont, line 29:
```suggestion
var outboundLog = core.Log.WithName("xds").WithName("outbound")
```

right?

michaelbeaumont (changes requested):

pkg/xds/topology/outbound.go
Automaat: updated

lukidzi (approved):

michaelbeaumont (approved):

Issue #9635, opened 2024-03-18
Prohibit manual configuration of `kuma.io/service` within the `MeshGatewayInstance`


lukidzi:
### Description

We decided to remove the ability to set `kuma.io/service` in `MeshGatewayInstance` because it is a breaking change we can't remove it now but we should be able to remove it in 2.9.

PR #10968, opened 2024-07-24
chore(meshgateway): forbid possibility of setting kuma.io/service for MGI

lukidzi:
### Checklist prior to review

In version 2.7.x we deprecate setting `kuma.io/service` tag for `MeshGatewayInstance`. In 2.9x we should be able to forbid setting it manually.
* added validation is kuma.io/service is set, and return error once is present 
* fixed e2e test after the change
* remove the warning message
* added test that validates if kuma.io/service is set
* added UPGRADE.md note

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/9635


pkg/plugins/runtime/k8s/webhooks/gateway_instance_validator_test.go
```diff
@@ -100,19 +95,43 @@ var _ = Describe("MeshGatewayInstanceValidator", func() {
                   labels:
                     kuma.io/mesh: default
                 spec:
-                  selectors:
-                    - match:
-                        kuma.io/service: edge-gateway_kuma-demo_svc
-                  conf:
-                    listeners:
```
michaelbeaumont, line 107: this was just nonsense right, that's why you removed it?

michaelbeaumont (approved):

pkg/plugins/runtime/k8s/webhooks/gateway_instance_validator_test.go
lukidzi: Yes, it wasn't MGI definition so I've replaced it with the real resource

Issue #9070, opened 2024-01-30
refactor(e2e): instead of `RunKubectlAndGetOutputE` we could use `GetServiceEndpoint` or `GetServiceE`


michaelbeaumont:
Worth doing all over the e2e tests

_Originally posted by @bartsmykla in https://github.com/kumahq/kuma/pull/9054#discussion_r1470640608_

PR #10972, opened 2024-07-24
test(e2e): use GetServiceE instead of RunKubectl

lukidzi:
### Checklist prior to review

Instead of running `RunKubectlAndGetOutputE` to retrieve service ip, I've added a method that uses `GetServiceE` and later from the service object retrieves IP

Another option is to retrieve Service and later call `GetServiceEndpoint` but in this case, we need to know the port, which sometimes we don't need (for curl resolve). That would require splitting a returned address so I think just getting service is enough.

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/9070

### DIFF
--- a/test/e2e_env/kubernetes/gateway/gateway.go
+++ b/test/e2e_env/kubernetes/gateway/gateway.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/gruntwork-io/terratest/modules/k8s"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
@@ -94,7 +93,7 @@ data:
 type: system.kuma.io/secret
 `, Config.KumaNamespace, secretData)
 	}
-	var clusterIP string
+	var gwIP string
 
 	BeforeAll(func() {
 		err := NewClusterSetup().
@@ -120,13 +119,9 @@ type: system.kuma.io/secret
 
 		Eventually(func(g Gomega) {
 			var err error
-			clusterIP, err = k8s.RunKubectlAndGetOutputE(
-				kubernetes.Cluster.GetTesting(),
-				kubernetes.Cluster.GetKubectlOptions(namespace),
-				"get", "service", "simple-gateway", "-ojsonpath={.spec.clusterIP}",
-			)
+			gwIP, err = kubernetes.Cluster.GetServiceIP("simple-gateway", namespace)
 			g.Expect(err).ToNot(HaveOccurred())
-			g.Expect(clusterIP).ToNot(BeEmpty())
+			g.Expect(gwIP).ToNot(BeEmpty())
 		}, "30s", "1s").Should(Succeed())
 	})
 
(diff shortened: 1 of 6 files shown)


test/e2e_env/kubernetes/gateway/gateway.go
```diff
@@ -120,13 +119,9 @@ type: system.kuma.io/secret
 
 		Eventually(func(g Gomega) {
 			var err error
-			clusterIP, err = k8s.RunKubectlAndGetOutputE(
-				kubernetes.Cluster.GetTesting(),
-				kubernetes.Cluster.GetKubectlOptions(namespace),
-				"get", "service", "simple-gateway", "-ojsonpath={.spec.clusterIP}",
-			)
+			gwIP, err = kubernetes.Cluster.GetServiceIP("simple-gateway", namespace)
```
michaelbeaumont, line 122: This might actually give the "wrong" IP, right? Since it might be a load balancer, and a load balancer also has a ClusterIP.

lukidzi: True, you are right, maybe I can add a `type` so you can choose which address you want

michaelbeaumont (approved):

test/framework/k8s_cluster.go
```diff
@@ -1027,6 +1027,34 @@ func (c *K8sCluster) DeleteMesh(mesh string) error {
...
+		c.t,
+		c.GetKubectlOptions(namespace),
+		serviceName,
+	)
+	if err != nil {
+		return "", err
+	}
+	ingress := service.Status.LoadBalancer.Ingress
+	if len(ingress) == 0 {
+		return "", errors.New("ingress not found")
```
jijiechen, line 1053: The wording here can be improved since it's easy to be confused since "ingress" has multiple meanings in many scenarios. At here we can't find any ingress for the load balancer service, so it's better to make it `fmt.Sprintf("ingress information not found on the load balancer service '%s'", serviceName)`

jijiechen, line 1053:
```suggestion
		return "", errors.New(fmt.Sprintf("ingress information not found on the load balancer service '%s'", serviceName))
```

test/framework/k8s_cluster.go
```diff
@@ -1027,6 +1027,34 @@ func (c *K8sCluster) DeleteMesh(mesh string) error {
...
+		c.t,
+		c.GetKubectlOptions(namespace),
+		serviceName,
+	)
+	if err != nil {
+		return "", err
+	}
+	ingress := service.Status.LoadBalancer.Ingress
+	if len(ingress) == 0 {
+		return "", errors.New(fmt.Sprintf("ingress information not found on the load balancer service '%s'", serviceName))
```
lukidzi, line 1053:
```suggestion
		return "", errors.Errorf("ingress information not found on the load balancer service '%s'", serviceName)
```

michaelbeaumont (approved):
:rocket:

PR #11010, opened 2024-07-29
fix(HostnameGenerator): consistently order resources

### LINKED ISSUE #10989: HostnameGenerator is flaky when there are collisions
### What happened?

1. Force HostnameGenerator to have collisions
2. Observe the hostname in `meshservice.status` is constantly changing from the real hostname to the error message 

We noticed this in the multizone scenario with statefulsets, but the repro could be way simpler:
1. Apply statefulset in 2 zones:
    ```yaml
    apiVersion: apps/v1
    kind: StatefulSet
    metadata:
      name: redis
      namespace: kuma-demo
    spec:
      selector:
        matchLabels:
          app: redis
      serviceName: redis
      replicas: 0
      template:
        metadata:
          labels:
            app: redis
        spec:
          containers:
            - name: redis
              image: "redis"
              ports:
                - name: tcp
                  containerPort: 6379
              env:
                - name: POD_NAME
                  valueFrom:
                    fieldRef:
                      fieldPath: metadata.name
              lifecycle:
                preStop: # delay shutdown to support graceful mesh leave
                  exec:
                    command: ["/bin/sleep", "30"]
                postStart:
                  exec:
                    command: ["sh", "-c", "/usr/local/bin/redis-cli set zone ${POD_NAME}"]
    ```
2. In one zone apply the hostname generator (without `kuma.io/origin` label):
    ```yaml
    kind: HostnameGenerator
    apiVersion: kuma.io/v1alpha1
    metadata:
      name: local-headless-kube-mesh-service
      namespace: kuma-system
      labels:
        kuma.io/origin: zone
    spec:
      template: '{{ label "statefulset.kubernetes.io/pod-name" }}.{{ label "k8s.kuma.io/service-name" }}.{{ .Namespace }}.svc.mesh.local'
      selector:
        meshService:
          matchLabels:
            kuma.io/managed-by: k8s-controller
            k8s.kuma.io/is-headless-service: "true"
    ```
3. Observe hostnames appearing and disappearing 
image
4. Zone CP is constantly reconciling resources [zone-cp.log](https://github.com/user-attachments/files/16390351/zone-cp.log)

michaelbeaumont:
- https://github.com/kumahq/kuma/pull/11009

We weren't sorting the resources before running the HostnameGenerators

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- Closes #10989 

### DIFF
--- a/pkg/core/resources/apis/meshservice/hostname/generator_test.go
+++ b/pkg/core/resources/apis/meshservice/hostname/generator_test.go
@@ -15,7 +15,7 @@ import (
 	meshservice_hostname "github.com/kumahq/kuma/pkg/core/resources/apis/meshservice/hostname"
 	"github.com/kumahq/kuma/pkg/core/resources/manager"
 	"github.com/kumahq/kuma/pkg/core/resources/model"
-	"github.com/kumahq/kuma/pkg/core/resources/store"
+	core_store "github.com/kumahq/kuma/pkg/core/resources/store"
 	core_metrics "github.com/kumahq/kuma/pkg/metrics"
 	"github.com/kumahq/kuma/pkg/plugins/resources/memory"
 	"github.com/kumahq/kuma/pkg/test/resources/builders"
@@ -24,12 +24,14 @@ import (
 
 var _ = Describe("MeshService Hostname Generator", func() {
 	var stopChSend chan<- struct{}
+	var store core_store.ResourceStore
 	var resManager manager.ResourceManager
 
 	BeforeEach(func() {
 		m, err := core_metrics.NewMetrics("")
 		Expect(err).ToNot(HaveOccurred())
-		resManager = manager.NewResourceManager(memory.NewStore())
+		store = memory.NewStore()
+		resManager = manager.NewResourceManager(store)
 		allocator, err := hostname.NewGenerator(
 			logr.Discard(), m, resManager, 50*time.Millisecond,
 			[]hostname.HostnameGenerator{meshservice_hostname.NewMeshServiceHostnameGenerator(resManager)},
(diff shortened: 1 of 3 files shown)


pkg/core/resources/apis/meshservice/hostname/generator_test.go
```diff
@@ -143,19 +152,114 @@ var _ = Describe("MeshService Hostname Generator", func() {
...
+						Status: kube_meta.ConditionTrue,
+						Reason: hostnamegenerator_api.GeneratedReason,
+					}},
+				},
+			))
+		}, "2s", "100ms").Should(Succeed())
+
+		// when
+		Expect(resManager.DeleteAll(context.Background(), &meshservice_api.MeshServiceResourceList{}, core_store.DeleteAllByMesh(model.DefaultMesh))).To(Succeed())
+		Expect(resManager.DeleteAll(context.Background(), &hostnamegenerator_api.HostnameGeneratorResourceList{})).To(Succeed())
```
jakubdyszkiewicz, line 208: Everything from this line looks like the rest of the test aside of the order of how we create other and backend.

Wouldn't this be enough to just keep
```
		Expect(other.Create(resManager, core_store.CreatedAt(createdSecond))).To(Succeed())
		Expect(backend.Create(resManager, core_store.CreatedAt(createdFirst))).To(Succeed())
```
?

This way we check that
1) we preserve the order, otherwise the test will be flaky
2) we respect the CreatedAt

pkg/core/resources/apis/meshservice/hostname/generator_test.go
```diff
@@ -24,12 +24,14 @@ import (
 
 var _ = Describe("MeshService Hostname Generator", func() {
 	var stopChSend chan<- struct{}
+	var store core_store.ResourceStore
```
jakubdyszkiewicz, line 27: nit: do we need it?

pkg/core/resources/apis/meshservice/hostname/generator_test.go
jakubdyszkiewicz: We also could keep the hostname generators in before each I think?

pkg/core/resources/apis/meshservice/hostname/generator_test.go
michaelbeaumont: Ah no, I had this from when I tried with `memory.ClearStore`

pkg/core/resources/apis/meshservice/hostname/generator_test.go
michaelbeaumont: > Everything from this line looks like the rest of the test aside of the order of how we create other and backend.

that's the point I think. that the order doesn't affect the outcome, even if `List` returns the resources in a different order

michaelbeaumont: > We also could keep the hostname generators in before each I think?

yeah, I added this back.

PR #11020, opened 2024-07-31
fix(kuma-cp): don't add namespace labels when resource was synced from universal zone

Automaat:
Fixes: #10882

Previous pr has a problem because it was only adding proper label on defaulter. Not all resources are updated by defaulter, and not all resources are created using store, some like Dataplane are created and updated in controllers. These need to add labels explicitly.

### Checklist prior to review


jakubdyszkiewicz:
so what's different between this and the previous PR?

lobkovilya:
So basically we have 3 ways resources are created on k8s:
* store.go
* directly applied using kubectl (we catch that in webhook)
* controllers

It would be ideal all these places to use the same code that sets labels. But if it requires a lot of code changes we at least should try to have all controller to share the code

pkg/core/resources/model/resource.go
```diff
@@ -461,7 +468,14 @@ func ComputeLabels(r Resource, mode config_core.CpMode, isK8s bool, systemNamesp
 		}
 	}
 
-	if ns, ok := labels[mesh_proto.KubeNamespaceTag]; ok && r.Descriptor().IsPolicy && r.Descriptor().IsPluginOriginated {
```
Automaat, line 464: we should refactor computeLabals, to be more specific and probably use some form of builder

lobkovilya:
Resources such as Dataplane, MeshGatewayInstance and etc should be added to the `defaulter` webhook. In this case the `ComputeLabels` method will be called automatically for them.

lobkovilya (approved):

Issue #11023, opened 2024-08-01
Don't set initial FetchTimeout


lahabana:
### What happened?

We shouldn't set it as it's useless to timeout, we should just wait for the conf.

Here's some pointers:

https://github.com/kumahq/kuma/blob/32d6e52f08955fd9b8824e3ce2b0c9fc8a693b46/pkg/xds/envoy/clusters/v3/eds_cluster_configurer.go#L15

PR #11024, opened 2024-08-01
fix(xds): explicitly set initial fetch timeout to zero to keep Envoy wait for xds resources

jijiechen:
### Checklist prior to review

  - fixes https://github.com/kumahq/kuma/issues/11023
  - Confirmed
  - Tested
  - No
  - yes

### DIFF
--- a/pkg/xds/bootstrap/template_v3.go
+++ b/pkg/xds/bootstrap/template_v3.go
@@ -1,6 +1,7 @@
 package bootstrap
 
 import (
+	"google.golang.org/protobuf/types/known/durationpb"
 	"net"
 	"strconv"
 	"time"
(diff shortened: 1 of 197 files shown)


pkg/xds/bootstrap/template_v3.go
```diff
@@ -158,10 +159,12 @@ func genConfig(parameters configParameters, proxyConfig xds.Proxy, enableReloada
 		DynamicResources: &envoy_bootstrap_v3.Bootstrap_DynamicResources{
 			LdsConfig: &envoy_core_v3.ConfigSource{
 				ConfigSourceSpecifier: &envoy_core_v3.ConfigSource_Ads{Ads: &envoy_core_v3.AggregatedConfigSource{}},
+				InitialFetchTimeout:   durationpb.New(0),
 				ResourceApiVersion:    envoy_core_v3.ApiVersion_V3,
 			},
 			CdsConfig: &envoy_core_v3.ConfigSource{
 				ConfigSourceSpecifier: &envoy_core_v3.ConfigSource_Ads{Ads: &envoy_core_v3.AggregatedConfigSource{}},
+				InitialFetchTimeout:   durationpb.New(0),
```
michaelbeaumont, line 167: Sanity check: is this also going to apply for ClusterLoadAssignment?

jijiechen: ClusterLoadAssignment is covered in EDS, located at `pkg/xds/envoy/clusters/v3/eds_cluster_configurer.go`

michaelbeaumont (commented):
I think we need to set https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-eds-cluster-config

EDIT: right we're doing that

pkg/xds/bootstrap/template_v3.go
michaelbeaumont: Yeah we need to set that as well I think

michaelbeaumont (approved):

Issue #11036, opened 2024-08-02
`HostnameGenerator`: select a specific mesh


michaelbeaumont:
### Description

Let's make sure it's possible to select one mesh on both Universal and Kubernetes

- Explicitly match mesh (not as label)?

PR #11037, opened 2024-08-02
feat(kuma-cp): set `kuma.io/mesh` on universal resource labels

michaelbeaumont:
### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- Closes https://github.com/kumahq/kuma/issues/11036


Icarus9913 (approved):

Icarus9913:
Just one considering:
Do we need to supplement the KDS mapper function to update this label?


https://github.com/kumahq/kuma/blob/e5d62a7b9180a244a27a80421e567ebb78f2a6ae/pkg/kds/context/context.go#L70-L71

https://github.com/kumahq/kuma/blob/e5d62a7b9180a244a27a80421e567ebb78f2a6ae/pkg/kds/context/context.go#L94-L98

jakubdyszkiewicz (approved):
I'm slightly anxious about this change.
It means that it's duplicated (field + label), not sure if that won't confuse users. For example they will try to set mesh by label on universal which is not what they should do. 

Maybe we need a validation that if they put a resource, mesh label cannot be different than mesh field.

michaelbeaumont:
> Maybe we need a validation that if they put a resource, mesh label cannot be different than mesh field.

I wonder if maybe the label shouldn't be settable at all? So that there's no confusion around the fact that the label isn't the source of truth

jakubdyszkiewicz:
> I wonder if maybe the label shouldn't be settable at all? So that there's no confusion around the fact that the label isn't the source of truth

Maybe, but someone could take a YAML from the GUI (or export!), try to apply it and it would fail with validation, because label exist

jakubdyszkiewicz (approved):

PR #11043, opened 2024-08-05
feat(kuma-cp): add resource owner to resources in ResourceSet

Automaat:
They should simplify policy matching with ResourceRules. As we can easily extract conf from ResourceRules by UniqueResourceIdentifier

### Checklist prior to review


pkg/plugins/policies/core/rules/resourcerules.go
```diff
@@ -71,27 +71,27 @@ func WithSectionName(sectionName string) ComputeOptsFn {
...
 	}
 
-	switch r.Descriptor().Name {
+	switch uri.ResourceType {
 	case meshservice_api.MeshServiceType:
 		// find MeshService without the sectionName and compute rules for it
-		if opts.sectionName != "" {
-			return rr.Compute(r, reader)
+		if uri.SectionName != "" {
+			uriWithoutSection := copyIdentifier(uri)
```
lobkovilya, line 83: This is a struct with primitive field types, simple `uriWithoutSection := uri` already makes a copy

pkg/core/xds/resource.go
```diff
@@ -7,16 +7,19 @@ import (
...
 
 // Resource represents a generic xDS resource with name and version.
 type Resource struct {
-	Name     string
-	Origin   string
-	Resource ResourcePayload
+	Name          string
+	Origin        string
+	Resource      ResourcePayload
+	ResourceOwner *rules.UniqueResourceIdentifier
```
lobkovilya, line 22: Maybe it's better to avoid using `Owner` in this context as we already have an ownership concept on the store level. Probably something with `Origin`, like `OriginalResource` or `ResourceOrigin` suits better here

pkg/core/xds/resource.go
```diff
@@ -167,3 +170,20 @@ func (s *ResourceSet) List() ResourceList {
 
 	return list
 }
+
+func (s *ResourceSet) GetByOwner() map[rules.UniqueResourceIdentifier]map[string][]*Resource {
```
lobkovilya, line 174: nit: the name sort of means that `owner` is passed as an argument and we get values for the `owner` (i.e. we have `GetByKey(key)` or `GetByMesh(mesh)` in the code). This method should be `IndexByOwner()`

Automaat:
With this changes policy plugin code is much simpler:

```go
func applyToRealResources(rs *core_xds.ResourceSet, rules core_rules.ResourceRules, meshCtx xds_context.MeshContext) error {
	for uri, resType := range rs.IndexByOrigin() {
		conf := rules.Compute(uri, meshCtx.Resources)
		if conf == nil {
			continue
		}

		for typ, resources := range resType {
			switch typ {
			case envoy_resource.ListenerType:
				err := configureListeners(resources, conf.Conf[0].(api.Conf))
				if err != nil {
					return err
				}
			case envoy_resource.ClusterType:
				err := configureClusters(resources, conf.Conf[0].(api.Conf))
				if err != nil {
					return err
				}
			}
		}
	}
	return nil
}

func configureListeners(resources []*core_xds.Resource, conf api.Conf) error {
	for _, resource := range resources {
		configurer := plugin_xds.RealResourceListenerConfigurer{
			Conf:     conf,
			Protocol: resource.Protocol,
		}
		err := configurer.ConfigureListener(resource.Resource.(*envoy_listener.Listener))
		if err != nil {
			return err
		}
	}
	return nil
}

func configureClusters(resources []*core_xds.Resource, conf api.Conf) error {
	for _, resource := range resources {
		configurer := plugin_xds.ClusterConfigurerFromConf(conf, resource.Protocol)
		err := configurer.Configure(resource.Resource.(*envoy_cluster.Cluster))
		if err != nil {
			return err
		}
	}
	return nil
}
```

pkg/core/xds/resource.go
Automaat: changed

pkg/plugins/policies/core/rules/resourcerules.go
Automaat: changed

pkg/core/xds/resource.go
Automaat: changed

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -117,16 +119,30 @@ func GenerateClusters(
...
+				ResourceOrigin: createOwner(service.BackendRef(), meshCtx),
+				Protocol:       protocol,
 			})
 		}
 	}
 
 	return resources, nil
 }
 
+func createOwner(ref common_api.BackendRef, meshCtx xds_context.MeshContext) *core_rules.UniqueResourceIdentifier {
```
lobkovilya, line 134: nit: createResourceOrigin

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -117,16 +119,30 @@ func GenerateClusters(
...
+	if ref.Kind == common_api.MeshService && ref.ReferencesRealObject() {
```
lobkovilya, line 135: something is odd here, `ref.ReferencesRealObjects()` checks if `ref.Kind` is equal to `MeshService`

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -117,16 +119,30 @@ func GenerateClusters(
...
+		ms := meshCtx.MeshServiceByName[ref.Name]
```
lobkovilya, line 136: At this moment, `meshCtx.MeshServiceByName` contains `ms.GetMeta().GetName()` as keys which means on k8s it's equal to `name.namespace`. And `ref.Name` is probably just a name right? 

I think `meshCtx.MeshServiceByName` map should contain `ResourceIdentifier` as keys. And when you want to lookup `ref` in this map you'll need to call `TargetRefToResourceIdentifier(meta, ref)`

pkg/plugins/policies/core/xds/meshroute/clusters.go
Automaat: yes, but I don't want to implement this for other kinds for now, I just wanted to handle MeshService, also other resource are in different parts of meshCtx

pkg/plugins/policies/core/xds/meshroute/clusters.go
Automaat: changed

pkg/plugins/policies/core/xds/meshroute/clusters.go
lobkovilya: That's odd, does it work on k8s?

pkg/plugins/policies/core/xds/meshroute/clusters.go
lobkovilya: But if you know that `ref.Kind` is `MeshService` do you need to call `ref.ReferencesRealObjects()`?

Automaat: I think so, we do this in other planes route: https://github.com/kumahq/kuma/blob/982dbdc6819c899f0c4854b5e5dac07942ee8a91/pkg/plugins/policies/core/xds/meshroute/listeners.go#L241

pkg/plugins/policies/core/xds/meshroute/clusters.go
Automaat: Yes, I tested it on k8s, and it worked in e2e test for mesh timeout and resource rules

pkg/plugins/policies/core/xds/meshroute/clusters.go
Automaat: I think there could be old backendRef with MeshService kind, but old one

pkg/plugins/policies/core/xds/meshroute/clusters.go
Automaat: backendRef for real MeshService have `name.namespace` as name, that's why this work

lobkovilya (approved):

Issue #11038, opened 2024-08-05
Pod start delayed by 30-60s on GKE when oslogin enabled with CNI


bartsmykla:
### What happened?

We are basically susceptible to the same issue as described in https://github.com/istio/istio/issues/48416

> With this setup, all pods will take ~60s to startup.
>
> The root cause is a bad interaction between these causing a circular dependency. When a pod starts, istio-cni will be invoked. At this point, the Pod is created but not containers are running yet. istio-cni more or less calls nsenter <POD> -- iptables-restore. Our iptables commands use the xt_owner module (-m owner) which in turn calls getpwnam (https://git.netfilter.org/iptables/tree/extensions/libxt_owner.c#n150). This triggers PAM for authentication. When OSLogin is enabled on the GCE machine, a PAM module will be loaded: https://github.com/GoogleCloudPlatform/guest-oslogin. When triggered, this will call the metadata server (request like /computeMetadata/v1/oslogin/users?username=...).
>
> The GKE Metadata Server will detect this as a request coming from the pod. The pod, however, has not yet started, so the request is denied. The PAM module will [retry](https://github.com/GoogleCloudPlatform/guest-oslogin/blob/a2a250b02af56e161d57fbdae1941b21250c8eb9/src/oslogin_utils.cc#L429) this request a number of times before giving up. Once it gives up, execution can continue as usual.
>
> This impacts all Istio versions, but only recent GKE versions (later patch releases in GKE 1.25+). It does not impact Autopilot, which cannot use oslogin.

slonka:
@bartsmykla did you check kuma on GKE autopilot?

PR #11050, opened 2024-08-06
fix(transparent-proxy): fix pod delay when CNI on GKE with OS Login

bartsmykla:
A bug resulting in pod start being delayed by 30-60s occurs in the following situation:

- Kuma CNI plugin is used
- GKE is new enough to have a recent GKE Metadata Server, which is more restrictive
- OS Login is enabled in GCE

The issue arises due to a circular dependency during the pod initialization process. When a pod starts, `kuma-cni` gets invoked. At this moment, the pod exists, but its containers are not yet running. `kuma-cni` essentially runs a command similar to `nsenter <POD> -- iptables-restore`. This command uses the `xt_owner` module with the `-m owner` flag, which makes a call to `getpwnam` (reference here: https://git.netfilter.org/iptables/tree/extensions/libxt_owner.c#n150). This call triggers the PAM authentication process.

In environments where OS Login is enabled on a GCE machine, a specific PAM module from Google's guest-oslogin: https://github.com/GoogleCloudPlatform/guest-oslogin gets loaded. This module attempts to contact the metadata server with a request like `/computeMetadata/v1/oslogin/users?username=...`.

The GKE Metadata Server treats this as a request originating from the pod. Since the pod's containers are not yet running, the server denies the request. The PAM module retries the request several times before it eventually times out and allows the initial process to continue.

This commit fixes the issue by running iptables commands in a sandbox with the root filesystem remounted (as in CNI context, network namespaces for pods are separate, but the filesystem is the same) and mounting `/etc/nsswitch.conf` to `/dev/null`. This prevents the PAM authentication process from being triggered, thereby stopping requests from being sent to the Metadata server.

Closes https://github.com/kumahq/kuma/issues/11038

### Checklist prior to review

  - https://github.com/kumahq/kuma/issues/11038
  - It won't
  - Manually tested on
    - [x] Local k3d cluster
      - [x] kuma-cni
      - [x] kuma-init
    - [x] GKE (COS) with OS Login
      - [x] kuma-cni
      - [x] kuma-init
    - [x] GKE (Ubuntu) with OS Login
      - [x] kuma-cni
      - [x] kuma-init
    - [x] EKS
      - [x] kuma-cni
      - [x] kuma-init
    - [x] AKS (Azure CNI Overlay)
      - [x] kuma-cni
      - [x] kuma-init
    - [x] AKS (Azure CNI)
      - [x] kuma-cni
      - [x] kuma-init
    - [x] OpenShift 4 - local
      - [x] kuma-cni
  - There is no need
  - There is no need

### DIFF
--- a/pkg/transparentproxy/config/config_executables.go
+++ b/pkg/transparentproxy/config/config_executables.go
@@ -31,8 +31,79 @@ type Executable struct {
 	prefix string
 }
 
+// verifyIptablesMode checks if the provided 'path' corresponds to an iptables
+// executable operating in the expected mode.
+//
+// This function verifies the mode by:
+//  1. Executing the iptables command specified by 'path' with the `--version`
+//     argument to obtain the version output.
+//  2. Parsing the standard output using the `consts.IptablesModeRegex`.
+//     - The regex is designed to extract the mode string from the output (e.g.,
+//     "legacy" or "nf_tables").
+//     - If a match is found, the extracted mode is compared with the expected
+//     mode (`mode`) using the `consts.IptablesModeMap`.
+//  3. Returning:
+//     - `true` if the extracted mode matches the expected mode.
+//     - `false` if the command execution fails, parsing fails, or the extracted
+//     mode doesn't match the expected mode.
+//
+// Special Considerations:
+// Older iptables versions (e.g., 1.4.21, 1.6.1) may not support the `--version`
+// flag and exhibit the following behaviors:
+//   - The command exits with a non-zero code and a warning is written to
+//     stderr.
+//   - A warning is written to stderr but the command exits with code 0.
+//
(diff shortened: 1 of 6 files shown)


pkg/transparentproxy/config/config_executables.go
```diff
@@ -31,8 +31,79 @@ type Executable struct {
...
+		return c.mode == IptablesModeLegacy
+	}
+
+	matched := IptablesModeRegex.FindStringSubmatch(stdout.String())
+	if len(matched) == 2 {
+		return slices.Contains(IptablesModeMap[c.mode], matched[1])
+	}
+
+	return false
+}
```
bartsmykla, line 82: logic of this function didn't change, it just became a method and have more parameters which `execCmd` function expects now

michaelbeaumont (approved):

Issue #10607, opened 2024-06-24
MeshTimeout to support MeshService in `to[].targetRef.kind`


lobkovilya:
### Description

We want to reference real MeshService resources in the `to[]` section of MeshTimeout.

PR #11060, opened 2024-08-07
feat(policy): add possibility to target real MeshService in MeshTimeout policy

Automaat:
Fixes: #10607 

### Checklist prior to review


pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go
```diff
@@ -218,3 +218,65 @@ func toProtoDurationOrDefault(d *kube_meta.Duration, defaultDuration time.Durati
 	}
 	return util_proto.Duration(d.Duration)
 }
+
+type RealResourceListenerConfigurer struct {
```
lobkovilya, line 222: I don't think we need to bring "RealResource" naming to this area of the code. It's kind of irrelevant from the Configurer point of view if the `api.Conf` originates from the real resource or somewhere else.

Also, there is a lot of code duplication with `ListenerConfigurer`, that'd be acceptable if we were planning to delete `ListenerConfigurer` in the future, but we can't since we need it for inbound listeners. 

What we probably should do is:
1. Mark old `ListenerConfigurer` as deprecated and use it only for legacy ToRules so we could delete it in the future. 
2. Create a new `ListenerConfigurer` (without `Rules` and `Subsets`) and use it for inbound, outbound and in the future for routes.

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
```diff
@@ -272,3 +279,50 @@ func createInboundClusterName(servicePort uint32, listenerPort uint32) string {
...
+				}
+			case envoy_resource.ClusterType:
+				err := configureClusters(resources, conf.Conf[0].(api.Conf))
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
```
lobkovilya, line 305: This looks nice 👍

lukidzi:
Should this pr https://github.com/kumahq/kuma/pull/10422 be closed?

pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go
Automaat: you are right, changed configurer as suggested

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
```diff
@@ -83,9 +90,12 @@ func applyToInbounds(fromRules core_rules.FromRules, inboundListeners map[core_r
 		}
 
 		protocol := core_mesh.ParseProtocol(inbound.GetProtocol())
+		conf := getConf(fromRules.Rules[listenerKey], core_rules.MeshSubset())
+		if conf == nil {
```
lobkovilya, line 94: I'm pretty sure if the `conf` is nil we still have to let the ListenerConfigurer to work. For example, the idleTimeout here is set to 0 regardless of `conf`: https://github.com/kumahq/kuma/blob/dd29be83bc55f7620a6f9aec407f0f0dbbb5ccd7/pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go#L47

Automaat: changed to using deafult conf

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
```diff
@@ -65,6 +66,12 @@ func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *
 			return err
 		}
 	}
+
+	err := applyToRealResources(rs, policies.ToRules.ResourceRules, ctx.Mesh)
```
lukidzi, line 70: Does support of existing policies (kind: MeshService and name == kuma.io/service) is a part of this task?(Old policies apply to new listeners)

Automaat: No, this is only to add support for real MeshService. 

Also when you have old policy targeting `MeshService` and there is real MeshService with the same name this will work. I think this will work for universal, but on k8s we have a different naming scheme. 

For this to work on k8s, we would need to update how we build ResourceRules to seek for real MeshService based on `kuma.io/service` tag

lukidzi: Ok, it seems like after MADR related to migration we can add this

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
lobkovilya: why haven't we caught that in unit tests? are we missing some cases when conf is nil?

Automaat: we have unit test for it, but it was working without this. I think that this is because by default on resource it is 0, and we are setting it to 0 if there is no conf. Looking at the commit history it was needed to override old Timeout policy. And don't think we should add unit test for overriding old policies as they should not be mixed anyway

pkg/xds/envoy/types.go
```diff
@@ -227,6 +227,10 @@ func (sa ServicesAccumulator) AddBackendRef(backendRef common_api.BackendRef, cl
 			backendRef: backendRef,
 		}
 	}
+	// prioritize backendRef pointing to real resource
+	if backendRef.ReferencesRealObject() && !sa.services[cluster.Service()].backendRef.ReferencesRealObject() {
```
lukidzi, line 231: Can backendRef here be `nil` and we will get a panic?

pkg/plugins/runtime/k8s/webhooks/validation.go
```diff
@@ -43,6 +44,10 @@ type validatingHandler struct {
 	decoder      admission.Decoder
 }
 
+var meshServiceSupportImplemented = map[core_model.ResourceType]bool{
```
lukidzi, line 47: We can easily forget to add it here (other prs). Should we have some flag saying that it supports, or. a method returning supported policies?

pkg/xds/envoy/types.go
michaelbeaumont: The `backendRef` argument isn't a pointer

pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go
```diff
@@ -218,3 +221,77 @@ func toProtoDurationOrDefault(d *kube_meta.Duration, defaultDuration time.Durati
...
+		rc.configureRequestHeadersTimeout(hcm)
+		// old Timeout policy configures idleTimeout on listener while MeshTimeout sets this in cluster
+		if hcm.CommonHttpProtocolOptions == nil {
+			hcm.CommonHttpProtocolOptions = &envoy_core.HttpProtocolOptions{}
+		}
+
+		hcm.CommonHttpProtocolOptions.IdleTimeout = util_proto.Duration(0)
+		return nil
+	}
+	tcpTimeouts := func(proxy *envoy_tcp.TcpProxy) error {
```
michaelbeaumont, line 258: do we need to duplicate all of this?

pkg/xds/envoy/types.go
michaelbeaumont: Although it _does_ seem like if `sa.services[cluster.Service()].backendRef` is the empty value, we'll return `true`, is that ok?

Automaat: you are right, we will return true, which does not fit here

pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go
Automaat: I tried to reuse some of the code, but with MeshService, Mesh*Routes and how we target real resources code will have small differences. But we will remove the old configurer after we migrate to MeshService

pkg/xds/envoy/types.go
Automaat: I've updated `ReferencesRealObject` to take this into account

lukidzi (approved):

Issue #11061, opened 2024-08-07
Switch default labels logic to ResourceManager


jakubdyszkiewicz:
### Description

Consider switching default labels logic to ResourceManager.
For example we add kuma.io/zone, but it's added on the API layer on universal. If we create this resource in the CP itself, it's not applied. Such logic should be in the ResourceManager

slonka:
> triage: There are multiple places in k8s that don't go through resource manager. All kuma resources should go through kuma webhook. @lobkovilya will look into this. We don't think this logic will be in ResourceManager. It might get fixed in this PR https://github.com/kumahq/kuma/pull/11020.

slonka:
@lobkovilya - have you [had time look](https://github.com/kumahq/kuma/issues/11061#issuecomment-2284157335) into this?

PR #11062, opened 2024-08-07
fix(hostnamegenerator): multizone syncing

### LINKED ISSUE #11041: Backwards compatibility of default HostnameGenerators
### Description

We now have `meshMultiZoneSelector`, a HostnameGenerator with this field might exist on global because of default HostnameGenerators. We have to make sure it doesn't cause problems when connecting a zone CP from the previous version

jakubdyszkiewicz:
### Checklist prior to review

Fix multizone syncing
* HG applied on global were not hashed when synced to Zone
* `kuma.io/zone` was missing on the default. xref https://github.com/kumahq/kuma/issues/11061
* Old zones were NACKing new HostnameGenerator. It would NACK all HostnameGenerators. I introduced a KDS flag to detect if we can sync HostnameGenerator with MeshMultiZoneService selector. Fix #11041


pkg/kds/context/context_test.go
```diff
@@ -472,7 +472,7 @@ var _ = Describe("Context", func() {
 			var r model.Resource = core_mesh.NewCircuitBreakerResource()
 			switch given.scope {
 			case model.ScopeGlobal:
-				r = core_mesh.NewZoneIngressResource()
+				r = core_system.NewGlobalSecretResource()
```
jakubdyszkiewicz, line 475: ZoneIngress is never provided by global

pkg/kds/context/context.go
```diff
@@ -83,10 +84,16 @@ func DefaultContext(
 			RemoveStatus()),
 		reconcile.If(
 			reconcile.And(
-				reconcile.ScopeIs(core_model.ScopeMesh),
```
lobkovilya, line 86: Why do we want to hash global-scoped resources now?

jakubdyszkiewicz: To avoid conflicts. If a Global scoped resource can be applied on both global and zone CP, why wouldn't we hash it?

So far Global resources were
* GlobalSecret - we sync only signing key from global to zone
* Mesh - we sync as is without hashing
* Config - we sync selectively
* RBAC resources - we don't sync them

pkg/defaults/hostname_generator.go
```diff
@@ -98,12 +98,15 @@ func EnsureHostnameGeneratorExists(ctx context.Context, resManager core_manager.
...
 		}
-		if err := ensureHostnameGeneratorExists(ctx, resManager, log, "synced-mesh-external-service", namespace, spec); err != nil {
+		if err := ensureHostnameGeneratorExists(ctx, resManager, log, "synced-mesh-external-service", namespace, nil, spec); err != nil {
 			return err
 		}
 	}
 
 	if cfg.Mode == core.Zone {
+		labels := map[string]string{
+			mesh_proto.ResourceOriginLabel: string(mesh_proto.ZoneResourceOrigin),
```
lobkovilya, line 108: @Automaat don't we already set this automatically?

jakubdyszkiewicz: yes, but on api level :/ See https://github.com/kumahq/kuma/issues/11061

lobkovilya: I see, thanks for creating an issue

lobkovilya (approved):

Issue #10997, opened 2024-07-29
Add support for new `MeshService` in `spec.to[].targetRef` in MeshRetry policy


Automaat:
### Description

Implement support using new `ResourceRules`

PR #11063, opened 2024-08-07
feat(policy): add possibility to target real MeshService in MeshRetry policy

Automaat:
Fixes: #10997

### Checklist prior to review


lukidzi (commented):
In the code at [this link](https://github.com/kumahq/kuma/blob/b303034932aaff88d309623eeaa971371e491e04/pkg/plugins/runtime/k8s/webhooks/validation.go#L47), a map with supported types is added, but we haven't added it here. Is there a better way to automate this process? In cases like this, it's easy to forget to add it, as happened here.

Automaat:
> In the code at [this link](https://github.com/kumahq/kuma/blob/b303034932aaff88d309623eeaa971371e491e04/pkg/plugins/runtime/k8s/webhooks/validation.go#L47), a map with supported types is added, but we haven't added it here. Is there a better way to automate this process? In cases like this, it's easy to forget to add it, as happened here.

@lukidzi we will remove this validation after we finish implementing this for all policies that support `MeshService` so this won't be needed in the future

lukidzi (approved):

PR #11076, opened 2024-08-08
perf(xds): `add x-kuma-tags` conditionally

jakubdyszkiewicz:
### Checklist prior to review

Do not merge for release-2.9. I'd rather put it in 2.10

We use `x-kuma-tags` internally. We add this to outbound HTTP listeners and we remove them on the inbound listeners. Currently, we use this header for 3 things
1. Match sources in `FaultInjection` policy
2. Match from in `MeshFaultInjection` policy
3. Match sources in `RateLimit` policy

I'd argue that in 99% cases you don't need it then.
Just a simple app on k8s looks like this:
```json
"request_headers_to_add": [
  {
    "header": {
      "key": "x-kuma-tags",
      "value": "&app=redis&&k8s.kuma.io/namespace=kuma-demo&&k8s.kuma.io/service-name=redis&&k8s.kuma.io/service-port=6379&&kubernetes.io/hostname=k3d-kuma-server-0&&kuma.io/protocol=tcp&&kuma.io/service=redis_kuma-demo_svc_6379&&kuma.io/zone=default&&pod-template-hash=686bbbb5c8&"
    }
  }
]
```
This is quite a lot of data to put on every single request between Envoys.
Additionally, that's quite a lot of data to put on every single outbound in XDS config. Removing this slims down the XDS config.

Long term this header should be deleted and we should extract this from the certificate on the inbound side.

- [X] This will not break child repos: it doesn't hardcode values (i.e. "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --


michaelbeaumont:
> Long term this header should be deleted and we should extract this from the certificate on the inbound side.

isn't this for the non-mtls case?

jakubdyszkiewicz:
We don't have extracting tags from cert (aside MTP validation ofc), so listed cases are for both mtls and non-mtls

bartsmykla (approved):

Issue #11079, opened 2024-08-09
Route MeshExternalService traffic through egress only


slonka:
### Description

In [madr](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/056-meshexternalservice-and-zoneegress.md?plain=1#L38) we decided to route through egress only.

Follow up from https://github.com/kumahq/kuma/issues/10897

PR #11080, opened 2024-08-09
feat(meshexternalservice): route traffic through egress only

lukidzi:
### Checklist prior to review
Requires https://github.com/kumahq/kuma/pull/11075

As outlined in [this decision document](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/056-meshexternalservice-and-zoneegress.md), we want to support MeshExternalService only when ZoneEgress is enabled. This PR introduces a change that adds MeshExternalServices only when ZoneEgress is enabled.

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/11079


pkg/xds/topology/outbound_test.go
```diff
@@ -1434,74 +1434,118 @@ var _ = Describe("TrafficRoute", func() {
...
-								ServerName:               "example.com",
-								SANs:                     []core_xds.SAN{},
+								Protocol:   core_mesh.ProtocolTCP,
+								TLSEnabled: false,
 							},
 						},
 					},
 					"no-tls-mes": []core_xds.Endpoint{
 						{
-							UnixDomainPath: "unix://no-tls-mes",
```
slonka, line 1461: if it always goes through egress does unix domain socket make sense still? Should we just remove the support for it altogether?

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/testdata/default-meshexternalservice.clusters.golden.yaml
```diff
@@ -2,19 +2,37 @@ resources:
...
+      typedConfig:
+        '@type': type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
+        commonTlsContext:
+          alpnProtocols:
+          - kuma
+          combinedValidationContext:
+            defaultValidationContext:
+              matchTypedSubjectAltNames:
+              - matcher:
+                  exact: spiffe://default/zone-egress
```
slonka, line 22: Do we have a way to add egress config to a golden file and assert on that? With the current config we know that egress is configured but we don't know if it's configured correctly 🤔

slonka (commented):
2 things I wanted to know.

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/testdata/default-meshexternalservice.clusters.golden.yaml
lukidzi: It would require a lot of changes. We would need to configure Egress generators and probably everything around the context :/

pkg/xds/topology/outbound_test.go
lukidzi: Maybe it doesn't make sense, It would require egress to run on the same machine. We can deprecate it in a separate pr 👍

slonka: I'll open up an issue about this.

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/testdata/default-meshexternalservice.clusters.golden.yaml
slonka: I still think it makes sense to do it - let's open up an issue about this and let's talke it in 2.10.x

PR #11089, opened 2024-08-12
feat(transparent-proxy): allow configure transparent proxy from config file

bartsmykla:
This PR implements MADR: [Streamlining Transparent Proxy Configuration (Phase 1): Enable Configuration via Config Files and Environment Variables](https://docs.google.com/document/d/1auMWZDEEbuu_kBfrpP5VfxbxJqN9FTxXysZRLz4g4Wg)

> [!NOTE]  
> Since this PR includes several changes that were necessary to introduce the described functionality, I recommend reviewing it commit by commit

### Checklist prior to review

  - [MADR 061: Streamlining Transparent Proxy Configuration (Phase 1): Enable Configuration via Config Files and Environment Variables](https://docs.google.com/document/d/1auMWZDEEbuu_kBfrpP5VfxbxJqN9FTxXysZRLz4g4Wg)
  - It won't
  - Tests updated as well as manually tested
  - There is no need
  - There is no need


pkg/config/loader.go
```diff
@@ -1,56 +1,141 @@
...
+func (l *Loader) Load(stdin io.Reader) error {
+	switch l.Filename {
+	case "-":
+		return l.LoadReader(stdin)
+	case "":
+		return l.LoadBytes([]byte(l.Content))
+	default:
+		return l.LoadFile(l.Filename)
 	}
+}
```
Icarus9913, line 60: Just curious of when we would call this method? Also we expose `Filename` and `Content` properties as Public but we don't have these two parameters in the NewLoader functions.

So, is this a way to allow us to use nonstandard `Loader`? Could you please add some comments for these case conditions meaning?

pkg/transparentproxy/config/config.go
```diff
@@ -57,6 +72,27 @@ func (c *Owner) Set(s string) error {
...
+	var jsonValue interface{}
+
+	if err := json.Unmarshal(bs, &jsonValue); err != nil {
+		return err
+	}
+
+	switch typedValue := jsonValue.(type) {
+	case string, float64:
+		return c.Set(fmt.Sprint(jsonValue))
+	case map[string]interface{}:
```
jijiechen, line 85: This looks like a magic. Why do we need to handle map and key "uid" specially? 
May be a comment is helpful if it's not a big deal.

pkg/transparentproxy/config/config.go
```diff
@@ -89,6 +125,98 @@ func NewValueOrRangeList[T ~[]uint16 | ~uint16 | ~string](v T) ValueOrRangeList
...
+		*p = append(*p, parsedPort)
+	}
+
+	return nil
+}
+
+func parsePort(s string) (Port, error) {
+	u, err := parseUint16(s)
+
+	if err != nil || u == 0 {
```
jijiechen, line 189: Do we support specifying a `0` value to disable something?

bartsmykla: No, I don't think we should as it would duplicate the existing logic and is unnecessary

jijiechen (approved):

Icarus9913:
Could you please add a precedence explanation for the flags, ENV, config-files in the `install_transparent_proxy.go` command `Long` message?

Icarus9913 (approved):

bartsmykla:
> Could you please add a precedence explanation for the flags, ENV, config-files in the `install_transparent_proxy.go` command `Long` message?

Unfortunately it's currently not trivial to do because of the library we are using to parse this stuff. We'll have to update our docs at the beginning. Maybe I'll have some time later to figure this out

pkg/transparentproxy/config/config.go
bartsmykla: That was to allow user to provide a value with an object with a `uid` key as basically it's how the underlying structure looks like, but I got rid of it

pkg/config/loader.go
bartsmykla: I actually changed this logic and got rid of `Filename` and `Content` properties. We are calling this function here: https://github.com/kumahq/kuma/blob/c15478c28567dc156cabbf7a6e60d1b345787ff1/app/kumactl/cmd/install/install_transparent_proxy.go#L140-L144

PR #11095, opened 2024-08-12
fix(kuma-dp): don't fail if internal version is not semver

### LINKED ISSUE #11086: Dataplane fails to run if Envoy version is mismatched
### What happened?

# The issue

The Kuma dataplane currently fails to run if the version of Envoy used at runtime doesn't precisely align with the version used at buildtime, down to the patch level. 

# Why is this a problem?

If I were to build Kuma from source, currently there's a limitation in place where the version of Envoy defined in `mk/dev.mk` must match exactly with the version provided at runtime.

We end up with this interdependent behavior where if we wanted to upgrade to a newly patched version of Envoy, we'd need to rebuild Kuma, even though there really isn't a reason to do so.

Validating the correct version is in use is important, but it shouldn't stop someone from being able to patch Envoy without needing to do another rebuild for Kuma.

# The solution

I suggest only validating the major and minor version of Envoy align with the version used at buildtime in the version check here: https://github.com/kumahq/kuma/blob/master/app/kuma-dp/pkg/dataplane/envoy/compatibility.go

I do not foresee compatibility issues being a problem for Kuma between patched releases of Envoy, and as far as I can tell, there is not a direct dependency anywhere in the source code on an explicit patched version of Envoy

lahabana:
In compatibility checks we checked that the version matched our version constraints.
We used to panic if the build envoy version isn't a valid semver. We now allow this and an error log will show up

Also if the version isn't semver we check lexicographic equality.

fix #11086

### Checklist prior to review

### DIFF
--- a/app/kuma-dp/cmd/run.go
+++ b/app/kuma-dp/cmd/run.go
@@ -171,7 +171,7 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
 				return errors.Wrap(err, "failed to get Envoy version")
 			}
 
-			if envoyVersion.KumaDpCompatible, err = envoy.VersionCompatible("~"+kuma_version.Envoy, envoyVersion.Version); err != nil {
+			if envoyVersion.KumaDpCompatible, err = envoy.VersionCompatible(kuma_version.Envoy, envoyVersion.Version); err != nil {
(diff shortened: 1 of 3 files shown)


lukidzi (approved):

slonka:
@lahabana unit tests are failing

Icarus9913 (approved):

Issue #11090, opened 2024-08-12
Losing leader election leading to a restart that causes CI failure


slonka:
### Description

Some time ago we introduced a restart counter in our E2E tests to check if the CP does not restart (theoretically if everything is fine it shouldn't and a restart could indicate some problem with the CP like OOM). This works fine in general but leader election on k8s kills the CP and that causes the CI to "fail". 

One idea (from @michaelbeaumont) is to distinguish these restarts by exit code and filter them out.

slonka:
> Triage: this assumes that exit code is different for leader election: needs checking

michaelbeaumont:
I think the best we can do here, if we want to be safe, is if `restartCount: 1` and `lastState.terminated` has `exitCode: <leader election lost code>`, after my change to exit with 0 on leader lost, then we can ignore it. Unfortunately because only the last termination is kept, if `restartCount > 1` then we don't know if it may have exited with an error.

Automaat:
closing for now as we didnt see this recently. We can reopen if needed

PR #11106, opened 2024-08-12
feat(kuma-cp): exit with 0 when kubernetes leader election is lost

michaelbeaumont:
So we can distinguish this from an error case. The `Pod` has `restartPolicy: Always` as default

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- Part of https://github.com/kumahq/kuma/issues/11090

### DIFF
--- a/app/kuma-cp/cmd/run.go
+++ b/app/kuma-cp/cmd/run.go
@@ -35,7 +35,7 @@ import (
 
 var runLog = controlPlaneLog.WithName("run")
 
-const gracefullyShutdownDuration = 3 * time.Second
+const gracefulShutdownDuration = 3 * time.Second
 
 // This is the open file limit below which the control plane may not
 // reasonably have enough descriptors to accept all its clients.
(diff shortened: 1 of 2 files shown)


Icarus9913:
Just one question: will this Pod get one `event` to tell us the restart caused by `leader lost` ?

michaelbeaumont:
> Just one question: will this Pod get one event to tell us the restart caused by leader lost ?

Not sure exactly what you mean @Icarus9913, the Pod _should_ exit with exit code 0 now instead of with an error.

Icarus9913:
> > Just one question: will this Pod get one event to tell us the restart caused by leader lost ?
> 
> Not sure exactly what you mean @Icarus9913, the Pod _should_ exit with exit code 0 now instead of with an error.

Yeah, the container would exit with status code 0. I mean once if we see there's a `RESTARTS` and we try to use `kubectl describe po ***` to check its events, do we need add some extra events there to let people know the previous container exited with `leader lost`. Or it's fine that the previous exit code 0 and previous container logs could prove that?

Icarus9913 (approved):

Issue #11034, opened 2024-08-02
Improve kuma-dp probe


jakubdyszkiewicz:
### Description

Kuma DP probe is on `<POD-IP>:9901/ready` which is listener that we generate. For plaintext traffic it responds only on ready. The request is passed to Envoy Admin. Envoy Admin is executed in the same `main_thread` as xDS processing. When we generate many xDS changes, `main_thread` may not respond on the probe even though Envoy is fine.

Of the solution is to change the probe to give static response. Here is a MeshProxyPatch
```
apiVersion: kuma.io/v1alpha1
kind: MeshProxyPatch
metadata:
  name: admin-static-response
  namespace: kuma-system
spec:
  targetRef:
    kind: Mesh
  default:
    appendModifications:
      - listener:
          operation: Patch
          match:
            name: "kuma:envoy:admin"
          jsonPatches:
            - op: remove
              path: /filterChains/0/filters/0/typedConfig/routeConfig/virtualHosts/0/routes/0/route
            - op: add
              path: /filterChains/0/filters/0/typedConfig/routeConfig/virtualHosts/0/routes/0/direct_response
              value: {"status": 200, "body": {"inline_string": "LIVE"}}
```

Envoy will receive this listener only when it successfully connects to the CP and receives config. Additionally, we can change the cluster to EDS so that we will only make it healthy when we actually receive LDS/CDS/EDS.
The response for `/ready` endpoint is then handled on the worker threads.

It seems ok to always respond with LIVE. When Pod is terminating, we remove the endpoint from all the clients.

Things to check
1) Currently, when kuma-dp receives SIGSTOP it goes into DRAINING state. Check what HC responds in this scenario. We might need to handle it with stats probe so that outside of Kubernetes, the scheduler can also immediately remove endpoint by checking HC.
2) Check when other values can occur. It's hot restart which we don't support, but is there anything else?

Icarus9913:
relevant issue: https://github.com/kumahq/kuma/issues/2571

jijiechen:
Let me clarify the idea:

## Overall idea

After some discussion, our proposed solution is to implement an endpoint within the kuma-dp process to report a static "READY" status for now.

## Implementation

* In the `kuma-dp` process, we use a unix domain socket to serve this endpoint to prevent potential port conflicts with application ports. This HTTP server respond "READY" with status 200 constantly for now.
[...]

PR #11107, opened 2024-08-13
feat(kuma-dp): add a separate component to handle kuma-sidecar readiness probes

jijiechen:
### Checklist prior to review

  -  fixes https://github.com/kumahq/kuma/issues/11034
  - Confirmed
  - Added
  - No

> Changelog: respond probes of kuma-sidecar from kuma-dp process instead of Envoy


pkg/xds/generator/admin_proxy_generator.go
```diff
@@ -84,10 +90,31 @@ func (g AdminProxyGenerator) Generate(ctx context.Context, _ *core_xds.ResourceS
 		return nil, err
 	}
 
+	readinessCluster, err := envoy_clusters.NewClusterBuilder(proxy.APIVersion, dppReadinessClusterName).
+		Configure(envoy_clusters.ProvidedEndpointCluster(
+			govalidator.IsIPv6(adminAddress),
+			core_xds.Endpoint{
+				UnixDomainPath: core_xds.DppReadinessSocketName(proxy.Metadata.WorkDir, proxy.Id.ToResourceKey().Name),
```
lukidzi, line 97: Do you know if it works correctly on Windows? It's changing current behavior and I wonder if that won't break the flow (if anyone is doing healthcheck)?

jijiechen: I think it does not. NamedPipe is the similar technical on Windows, but the DP is not listening named pipe. Do you think we need to support Windows? We are using UnixDomain listeners on MeshMetrics and many other features. When considering backporting, I agree we need to support Windows.

jijiechen: When we do backporting, we can just respond a literal string "READY" on the `admin_proxy_generator.go` and don't need this component at all. Let me create a new PR to do that and let's discuss if that's more easy to be backported.
The reason why I created this PR introducing the component was because we can include more factors when calculating the readiness of `kuma-sidecar`. 
However, they will only be introduced in the future, I think. And in those version, we may have dropped Windows support. So for now, it's better to just keep it a more simple and "stupid" static response?

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -0,0 +1,120 @@
...
+func NewReporter(socketPath string) *Reporter {
+	return &Reporter{
+		socketPath: socketPath,
+	}
+}
+
+func (r *Reporter) Start(stop <-chan struct{}) error {
+	_, err := os.Stat(r.socketPath)
+	if err == nil {
+		// File is accessible try to rename it to verify it is not open
```
slonka, line 42: we already do similar things in

https://github.com/kumahq/kuma/blob/5e7b4f86e2b454131e707c3cb35ba740b5715573/app/kuma-dp/pkg/dataplane/metrics/server.go#L142 

should we dry it up?

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -0,0 +1,120 @@
...
+	if r.isTerminating.Load() {
+		state = stateTerminating
+		stateHTTPStatus = http.StatusServiceUnavailable
+	}
+
+	stateBytes := []byte(state)
+	writer.Header().Set("Content-Type", "text/plain")
+	writer.Header().Set("Content-Length", fmt.Sprintf("%d", len(stateBytes)))
+	writer.Header().Set("Cache-Control", "no-cache, max-age=0")
+	writer.Header().Set("X-Readiness-Server", "kuma-dp")
```
slonka, line 107: I think all these should be lowercase to avoid confusing users (Envoy lowercases, kuma-dp does not)

```suggestion
	writer.Header().Set("content-type", "text/plain")
	writer.Header().Set("content-length", fmt.Sprintf("%d", len(stateBytes)))
	writer.Header().Set("cache-control", "no-cache, max-age=0")
	writer.Header().Set("x-readiness-server", "kuma-dp")
```

also aren't `x-*` headers deprecated? https://datatracker.ietf.org/doc/html/rfc6648 https://tonyxu.io/blog/http-deprecate-x-prefix/ - I'm not sure we follow that RFC but might be good to figure out.

pkg/xds/generator/admin_proxy_generator.go
slonka: Since which version we stopped support for Win?

jijiechen: > Since which version we stopped support for Win?
I don't know actually. I'm hearing about it many times but I didn't find any official reference.

jijiechen: @lukidzi @slonka 
Just created this new replacement PR, that one does not rely on the listener in the `kuma-dp` process:
https://github.com/kumahq/kuma/pull/11145

jijiechen: I got two options:
* use named pipe
* listen a port and bind to `127.0.0.1`
Which one do you prefer?

slonka: I prefer `listen a port and bind to 127.0.0.1` - if we ever need to get back to windows this will work there as well. @lukidzi - WDYT?

jijiechen: Yeah, a port that binds to `127.0.0.1` makes sense to me. However, I don't like to make a new annotation to let the users specify this port (to prevent potential conflicts), which makes things complicated.

jijiechen: I'll use `9902` for now, and please feel free to raise your ideas.

app/kuma-dp/pkg/dataplane/readiness/component.go
jijiechen: Right, this practice is deprecated, and it looks like many `X-*` headhers are now becoming official. One example is "x-powered-by" and it looks this one fit our scenario well. I think it's beneficial to clearly tell the client that this response is generated from `kuma-dp`.
That been said, I'm going to change it to `x-powered-by`, what do you think?

pkg/xds/generator/admin_proxy_generator.go
lukidzi: We could bind to localhost and define the port. But it has to be configurable. If you run 2 instances of DP on the same VM you are going to hit conflict

jijiechen: Thanks for the reply, I'm providing a configuration key to change the port.

pkg/xds/generator/testdata/admin/05.envoy-config.golden.yaml
```diff
@@ -62,6 +62,11 @@ resources:
               - '*'
               name: kuma:envoy:admin
               routes:
+              - match:
+                  prefix: /ready
+                route:
+                  cluster: kuma:envoy:admin
```
lukidzi, line 68: Should it be admin or `kuma_readiness`?

pkg/xds/generator/testdata/admin/07.envoy-config.golden.yaml
```diff
@@ -0,0 +1,115 @@
...
+    altStatName: kuma_readiness
+    connectTimeout: 5s
+    loadAssignment:
+      clusterName: kuma:readiness
+      endpoints:
+      - lbEndpoints:
+        - endpoint:
+            address:
+              socketAddress:
+                address: 127.0.0.1
```
lukidzi, line 30: I think the address needs to be configurable because it might not work for ipv6. Lets verify it

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Reporter reports the health status of this Kuma Dataplane Proxy
+type Reporter struct {
+	localListenPort uint32
+	isTerminating   atomic.Bool
+}
+
+var (
+	logger      = core.Log.WithName("readiness")
+	localIPAddr = "127.0.0.1"
```
lukidzi, line 31: Here we need to support ipv4 and ipv6 (I believe)

pkg/xds/generator/admin_proxy_generator_test.go
```diff
@@ -93,16 +96,29 @@ var _ = Describe("AdminProxyGenerator", func() {
...
+			dataplaneFile: "04.dataplane.input.yaml",
+			expected:      "06.envoy-config.golden.yaml",
+			adminAddress:  "127.0.0.1",
+			readinessPort: 9902,
+		}),
+		Entry("should generate admin resources, IPv6 with readiness port 9400", testCase{
+			dataplaneFile: "05.dataplane.input.yaml",
+			expected:      "07.envoy-config.golden.yaml",
+			adminAddress:  "::1",
+			readinessPort: 9400,
```
lukidzi, line 120: Is this enabled by default? The above examples don't have a defined value, and it seems to use the old way or is it just a test?

jijiechen: This is a test case and I'm adding a value to assert this value should be used.

app/kuma-dp/pkg/dataplane/readiness/component.go
jijiechen: We don't support IPv6 only as of now, so either IPv4 or DualStack. So both of them include ipv4 address.

pkg/xds/generator/testdata/admin/07.envoy-config.golden.yaml
jijiechen: We don't support "IPv6 only" as of now, so either IPv4 or DualStack. So both of them include ipv4 address.

app/kuma-dp/pkg/dataplane/readiness/component.go
jijiechen: Well, I can see the `AdminAddress`, we'd built such support. If I reuse the `AdminAddress` to be the `localIPAddr` here for the readiness reporter to listen, do you think that make sense?

app/kuma-dp/pkg/dataplane/readiness/component.go
slonka: yeah, we can do it that way 👍

app/kuma-dp/pkg/dataplane/readiness/component.go
jijiechen: This is done now, please take a look at the lastest commit. Thank you.

pkg/xds/generator/testdata/admin/05.envoy-config.golden.yaml
jijiechen: Case 05 does not specify a value for `readinessPort`, so this is the case of an existing/old DPP is connecting to a new CP. We'll still forward traffic to the admin API in this scenario because there is no readiness reporter component running in the DP.

pkg/xds/generator/testdata/admin/07.envoy-config.golden.yaml
jijiechen: It's supported now, so resolving.

pkg/xds/generator/admin_proxy_generator_test.go
jijiechen: btw, it's enabled by default from the DP. But from CP's perspective, it does not have a default value and relies on the port value passed form the DP to decide whether it should generate xds config to forward to the readiness reporter component.

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -24,34 +23,28 @@ const (
...
-		}
-		err = os.Remove(newName)
-		if err != nil {
-			return errors.Errorf("not able the delete the backup file %s", newName)
-		}
+	protocol := "tcp"
+	addr := r.localListenAddr
+	if govalidator.IsIPv6(addr) {
+		protocol = "tcp6"
+		addr = fmt.Sprintf("[%s]", addr)
```
slonka, line 45: nitpick: just FYI you could use `netip.AddrPort` and have it handle formatting, but it probably won't blow up anywhere so we can leave it like that

slonka (approved):
Just one nitpick. Once you check universal + 2 dpps feel free to merge it

jijiechen:
I believe the CI failures are caused by image pulling issues that can recover in a different time.
It seems there is something with docker hub, they are not able to serve the image pulling at this time.

Issue #11094, opened 2024-08-12
Watchdog waits an entire period before first tick


lahabana:
### What happened?

From the code:

https://github.com/kumahq/kuma/blob/ae945b01999633d1e7bd68ad74ae3979ad99b511/pkg/util/watchdog/watchdog.go#L21-L53

We probably should not wait an entire tick before triggering first call. This can delay things unecessarilly (for example sending the configuration).

slonka:
> triage: accepted

PR #11110, opened 2024-08-13
fix(kuma-cp): use contexts instead of channels in watchdog

### LINKED ISSUE #7349: SimpleWatchdog.Start should take a ctx
### Description

After working on:
 - https://github.com/kumahq/kuma/pull/7348

We realised that SimpleWatchdog takes in a channel it would probably just use a context as it would avoid needing a goroutine and complex logic to handle closed channel.
If we actually go up the levels we can see that go-control-plane passes the ctx of the stream already. Should we just use this one or would that cause issues?
We have the stream ctx stored already in dataplane_callbacks.

lahabana:
This makes the code simpler and easier to track

xref #11094
fix #7349

### Checklist prior to review


michaelbeaumont:
I think you want to link https://github.com/kumahq/kuma/issues/7349

pkg/util/watchdog/watchdog.go
```diff
@@ -7,66 +7,42 @@ import (
...
-		case <-stop:
+		case <-ctx.Done():
 			if w.OnStop != nil {
 				w.OnStop()
 			}
-			// cancel will be called by the above goroutine
 			return
 		}
 	}
 }
```
Icarus9913, line 36: Nice clean up with the goroutines generation in every loop!

pkg/util/watchdog/watchdog.go
```diff
@@ -7,66 +7,42 @@ import (
...
 		select {
 		case <-ticker.C:
-			continue
-		case <-stop:
+		case <-ctx.Done():
 			if w.OnStop != nil {
 				w.OnStop()
 			}
-			// cancel will be called by the above goroutine
 			return
```
Icarus9913, line 33: Just one consideration here:
The golang select runs in random rather than rust marco `select!`, if the `ticker.C` and `ctx.Done` occurred at the same time, it might goes into `<-ticker.C` and runs the for loop again. You can add another `ctx.Done` check inner the `ticker.C` but it looks ugly.

I would prefer to make it in order with priority just like this:
```text
select {
  case <-ctx.Done():
    if w.OnStop != nil {
	    w.OnStop()
    }
    return
  default:
}

select {
  case <-ticker.C:
}
```

If there's something wrong with this, please point it out.

lahabana: Correct. It's a little trickier as you still need to select on the ctx in the second select but I'll improve a bit.

lahabana: PTAL @Icarus9913

Icarus9913: The new one looks good to me.

michaelbeaumont (commented):
This kind of mixes a refactor and some fixes, right?

pkg/kds/v2/server/components.go
```diff
@@ -128,7 +127,7 @@ func newSyncTracker(
 			NewTicker: func() *time.Ticker {
 				return time.NewTicker(refresh)
 			},
-			OnTick: func(context.Context) error {
+			OnTick: func(ctx context.Context) error {
```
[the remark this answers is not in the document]
lahabana: it's actually the same ctx now so I thought it was just clearer this way. Also in any case the ctx for reconcile should be the one of the OnTick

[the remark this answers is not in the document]
michaelbeaumont: do we still need the outer one?

michaelbeaumont (approved):

PR #11113, opened 2024-08-14
fix(transparent-proxy): avoid mounting xtables.lock for newer versions of legacy iptables

bartsmykla:
- Address warnings about missing `/run/xtables.lock` in CNI by introducing logic to parse the iptables version. The lock file is now only mounted for legacy iptables versions lower than 1.8.6, where `XTABLES_LOCKFILE` environment variable handling was not yet introduced
- Avoid running unnecessary iptables commands in a sandbox environment, particularly when executing `iptables --version`. This change reduces complications during config initialization, where multiple iptables commands are executed to verify functionality and parse versions

This solves situations like on k3d:
```sh
2024-08-14T09:55:06.233Z    INFO    install-cni    chaining Kuma CNI config. Updating CNI config file    {"file": "/host/etc/cni/net.d/10-flannel.conflist"}
# [iptables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [iptables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [iptables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [iptables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [iptables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [iptables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [iptables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [iptables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [ip6tables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [ip6tables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [ip6tables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [ip6tables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [ip6tables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [ip6tables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [ip6tables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
# [ip6tables] [WARNING] failed to set up the sandbox environment. This may be due to restrictive access controls (e.g., SELinux). Attempting to continue without the sandbox: failed to mount /proc/88645/task/88652/ns/net to /run/xtables.lock: no such file or directory
2024-08-14T09:55:12.695Z    INFO    kuma-cni    iptables rules applied    {"pod": "redis-c44756dff-q5mct", "namespace": "kuma-demo", "podInfraContainerId": "64d3810a547f991e73b0e7302db03309d3a7903ae44626952b8ff626d0c4b50d", "ip": "", "containerId": "64d3810a547f991e73b0e7302db03309d3a7903ae44626952b8ff626d0c4b50d", "args": "K8S_POD_NAMESPACE=kuma-
demo;K8S_POD_NAME=redis-c44756dff-q5mct;K8S_POD_INFRA_CONTAINER_ID=64d3810a547f991e73b0e7302db03309d3a7903ae44626952b8ff626d0c4b50d;K8S_POD_UID=892ba9ac-71fd-4977-b8d4-c057a0954c83;IgnoreUnknown=1"}
...
```

### Checklist prior to review

  - no relevant issues
  - It won't
  - [x] manually tested on local k3d cluster and GKE cluster with metadata server enabled
  - there is no need
  - there is no need


pkg/transparentproxy/config/config_executables.go
```diff
@@ -148,11 +133,23 @@ type InitializedExecutable struct {
...
+	// Only iptables and iptables-restore executables need a lock because they
+	// perform write operations
+	switch c.name {
+	case "", "restore":
+		return true
+	}
+
+	// For all other cases (such as iptables-save), no lock is needed as they
+	// don't perform write operations
+	return false
```
bartsmykla, line 152: This method is exactly the same as before, but was just moved from `Executable` to `InitializedExecutable`

pkg/transparentproxy/config/config_executables.go
```diff
@@ -14,90 +14,73 @@ import (
...
+	name string
+	mode IptablesMode
+	// prefix represents the prefix used for iptables executables, which varies
+	// based on whether it's for IPv4 or IPv6 operations. For IPv4, it can be
+	// `iptables` (for binaries such as `iptables`, `iptables-restore`, or
+	// `iptables-save`). For IPv6, it can be `ip6tables` (for binaries such as
+	// `ip6tables`, `ip6tables-restore`, or `ip6tables-save`). This property
+	// facilitates constructing the full executable names, generating temporary
+	// file names for iptables rules, and other related operations.
+	prefix string
```
bartsmykla, line 83: Nothing changed in this struct, it was just moved below the `Version`

pkg/transparentproxy/config/config_executables_linux.go
```diff
@@ -39,12 +39,9 @@ func setupSandbox(netns ns.NetNS, needLock bool) error {
 		return errors.Wrap(err, "failed to remount root filesystem as private")
 	}
 
-	if needLock {
+	if c.NeedLock() && c.version.LessThan(IptablesVersionWithLockfileEnv) {
```
jijiechen, line 42: A random question, when `XTABLES_LOCKFILE` is not available and the container has a readonly filesystem, what will happen? 

We are using this environment variable on `kuma-init` too, I'm thinking if we need to do a similar detection to decide whether we should enable the `readonlyFilesystem` security flag on the container.
https://github.com/kumahq/kuma/blob/dc1daadf0a8ee930bb9d1412a56e10c5a92d4d65/pkg/plugins/runtime/k8s/webhooks/injector/injector.go

jijiechen (approved):

pkg/transparentproxy/config/config_executables_linux.go
bartsmykla: > A random question, when XTABLES_LOCKFILE is not available and the container has a readonly filesystem, what will happen?

This code runs only in CNI, so as far as I know there shouldn't be issues with read-only filesystems

> Further, is it correct to just specify a new path by using this environment variable on newer iptable versions?

Yes

> Are there other steps steps we need to do on this file, like is there any specified content/format this file should be?

If you are asking about xtables.lock file, then we don't touch the file, we just provide path and kernel is responsible to handle (create, delete etc.) it

> We are using this environment variable on kuma-init too, I'm thinking if we need to do a similar detection to decide whether we should enable the readonlyFilesystem security flag on the container.
https://github.com/kumahq/kuma/blob/dc1daadf0a8ee930bb9d1412a56e10c5a92d4d65/pkg/plugins/runtime/k8s/webhooks/injector/injector.go

For init containers it's different, because it's we who are responsible for installing iptables, so we know what version will be installed: https://github.com/kumahq/kuma/blob/31832e83137324b1b42b504bb20e5871f7fffb29/tools/releases/dockerfiles/kuma-init.Dockerfile#L6

jijiechen: Thank you for your detailed answers. They really help resolve my concerns!

Issue #11042, opened 2024-08-05
`MeshMultiZoneService` policy matching


michaelbeaumont:
### Description

Handle as `to.targetRef`

michaelbeaumont:
Going to close this as the above PRs cover everything

PR #11121, opened 2024-08-14
fix(MeshMultiZoneService): handle protocol with routes

michaelbeaumont:
We weren't properly computing the conf and handling the ports as HTTP

I won't close the linked issue yet but after this PR, policies applied to `MeshMultiZoneService` should just work after loosening validation. Tested with https://github.com/kumahq/kuma/pull/11060

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- Part of #11042 

> Changelog: skip


pkg/plugins/policies/core/rules/resourcerules_test.go
```diff
@@ -307,6 +308,44 @@ var _ = Describe("Compute", func() {
 		Expect(rule.Conf).To(Equal([]interface{}{"conf-2"}))
 	})
 
+	It("should return Mesh rule if MeshMultiZoneService is not found", func() {
```
slonka, line 311: Why is this related?

slonka (commented):
Please help me understand why is this mesh fallback needed 🤔

pkg/plugins/policies/core/rules/resourcerules_test.go
michaelbeaumont: Because `Compute` wasn't working. It's just the fact that a `kind: Mesh` rule applies if there's no `kind: MeshMultiZoneService`

slonka (approved):

PR #11145, opened 2024-08-20
feat(kuma-dp): statically respond readiness of Envoy as READY

jijiechen:
### Checklist prior to review

This is an alternative solution and replaces https://github.com/kumahq/kuma/pull/11107

  -  fixes https://github.com/kumahq/kuma/issues/11034
  - Confirmed
  - Added
  - No

> Changelog: statically respond readiness of Envoy as `READY` instead of using Envoy admin API

### DIFF
--- a/pkg/xds/envoy/listeners/v3/static_endpoints_configurer.go
+++ b/pkg/xds/envoy/listeners/v3/static_endpoints_configurer.go
@@ -1,6 +1,7 @@
 package v3
 
 import (
+	envoy_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	envoy_route "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	envoy_hcm "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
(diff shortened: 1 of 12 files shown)


pkg/xds/envoy/listeners/v3/static_endpoints_configurer.go
```diff
@@ -28,14 +29,26 @@ func (c *StaticEndpointsConfigurer) Configure(filterChain *envoy_listener.Filter
...
+			route.Action = &envoy_route.Route_Route{
 				Route: &envoy_route.RouteAction{
 					ClusterSpecifier: &envoy_route.RouteAction_Cluster{
 						Cluster: p.ClusterName,
 					},
 					PrefixRewrite: p.RewritePath,
 				},
-			},
+			}
+		} else if p.DirectResponse != nil {
```
jijiechen, line 43: We have an existing `DirectResponseConfigurer`, but that one only receives `DirectResponseEndpoints`, and we need `StaticEndpointPath` along with DirectResponse in `admin_proxy_generator.go`. 
So I just added DirectResponse support into this `StaticEndpointsConfigurer`. I think we can also consider redirect the existing `DirectResponseConfigurer` to this `StaticEndpointsConfigurer`, and still keep the API of `envoy_listeners.DirectResponse` to enjoy its intuitiveness.

jijiechen:
From my testing, this one will never return `503` even it's in the process of terminating.

PR #11151, opened 2024-08-20
fix(meshgatewayinstance): remove required since we generate serviceName

lukidzi:
### Checklist prior to review

While debugging https://github.com/kumahq/kuma/pull/11137, I discovered that after upgrading the Kubernetes libraries, the gateway tests were failing. The issue was due to the `kuma.io/service` label being required, even though it's autogenerated and we don't allow it to be set manually since 2.9.x(https://github.com/kumahq/kuma/pull/10968). To resolve this, I changed the label's status from required to optional.

### DIFF
--- a/pkg/plugins/resources/k8s/native/api/v1alpha1/mesh_gateway_instance.go
+++ b/pkg/plugins/resources/k8s/native/api/v1alpha1/mesh_gateway_instance.go
@@ -30,8 +30,7 @@ type MeshGatewayInstanceSpec struct {
 	// since is auto-generated, and should match exactly one Gateway
 	// resource.
 	//
-	// +required
-	// +kubebuilder:validation:MinLen=1
+	// +optional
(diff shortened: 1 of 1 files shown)


pkg/plugins/resources/k8s/native/api/v1alpha1/mesh_gateway_instance.go
```diff
@@ -30,8 +30,7 @@ type MeshGatewayInstanceSpec struct {
 	// since is auto-generated, and should match exactly one Gateway
 	// resource.
 	//
-	// +required
-	// +kubebuilder:validation:MinLen=1
+	// +optional
```
michaelbeaumont, line 33: Do you need to remove the minlen validation?

lukidzi: We don't need, reverted

slonka (approved):

PR #11155, opened 2024-08-21
fix(policy): don't fail once cannot map MES to tags rules

lukidzi:
### Checklist prior to review

`BuildToRules` calls two methods: `BuildRules` and `BuildResourceRules`. `BuildRules` maps kinds to tags, but for `MeshExternalService`, it cannot be mapped to a specific tag because it's a different type. Previously, we were returning an error and didn't attempt to run `BuildResourceRules`, which caused the policy to not match the dataplane (DP).

Note: I will add a test once MeshExternalService is supported by one of the policies, as it currently does not pass the validator.

### DIFF
--- a/api/common/v1alpha1/ref.go
+++ b/api/common/v1alpha1/ref.go
@@ -55,6 +55,8 @@ func (k TargetRefKind) IsRealResource() bool {
 	}
 }
 
+// These are the kinds that can be used in Kuma policies before support for 
+// actual resources (e.g., MeshExternalService, MeshMultiZoneService, and MeshService) was introduced.
 func (k TargetRefKind) IsOldKind() bool {
 	switch k {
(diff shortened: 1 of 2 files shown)


api/common/v1alpha1/ref.go
```diff
@@ -55,6 +55,15 @@ func (k TargetRefKind) IsRealResource() bool {
 	}
 }
 
+func (k TargetRefKind) IsOldKind() bool {
```
lobkovilya, line 58: nit: please leave a doc comment for this method. Like these are kinds that's possible to use in Kuma policies before introducing support for real resources (i.e. MeshExternalService, MeshMutliZoneService and MeshService)

lobkovilya (approved):

Issue #10724, opened 2024-07-02
MeshRetry should be able to configured for MeshExternalService


slonka:
### Description

To avoid having squared number of retries we're implementing this on the sidecar.

PR #11162, opened 2024-08-21
feat(policy): add possibility to target real MeshExternalService in MeshRetry policy

lukidzi:
### Checklist prior to review

* added e2e test, because we don't support FaultInjection I am using one HTTP and HTTPS service and the https fails
* added resource rules set
* added validator for MeshExternalService

- [x] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/10724


pkg/plugins/policies/meshretry/plugin/v1alpha1/plugin_test.go
```diff
@@ -587,6 +598,102 @@ var _ = Describe("MeshRetry", func() {
 			},
 			goldenFilePrefix: "http-real-mesh-service",
 		}),
+		Entry("http retry mesh external service", testCase{
```
slonka, line 601: should we have one for tcp and grpc as well?

pkg/plugins/policies/core/rules/testdata/resourcerules/mretry_with_mes.zone-uni.golden.yaml
```diff
@@ -0,0 +1,73 @@
+mesh:name/mesh-1:
```
slonka, line 1: do all of these differ in any way other than names? (I guess from syncing right?)

slonka (commented):
The actual code required to run this is just a couple of lines, right?

pkg/plugins/policies/meshretry/plugin/v1alpha1/plugin_test.go
lukidzi: I’m not entirely convinced, but we can add it. These tests will require some rewriting (similar to the ones you started in MeshHTTPRoute), as we mock a lot of data. Even if the test passes, it might not work in E2E. I think these tests are useful for checking the output Envoy config, but they don't test the entire flow.

pkg/plugins/policies/core/rules/testdata/resourcerules/mretry_with_mes.zone-uni.golden.yaml
lukidzi: Yes, it's mostly to check if everything still works after syncing and adding the hash. I've added another use case to the existing test base.

pkg/plugins/policies/meshretry/plugin/v1alpha1/plugin_test.go
slonka: Is there any reason for this not to work for grpc, they are configured in a similar place right?

lukidzi: I don’t think it shouldn’t. We can add another case with different protocol

slonka (approved):

PR #11175, opened 2024-08-22
feat(kuma-cp): add backendRef indexes to rules

lobkovilya:
Both `Rule` and `ResourceRule` contain the final `Conf`. This `Conf` is a result of merging of multiple policies (their metas are in `Origin`). But in general, for any given field in `Conf` we can't tell what policies contributed the value for the field. 

So far, we didn't need this information, but it changes when `backendRefs` is inside `Conf` and point to real resources and we have to resolve it. Resolving requires knowing the meta of the policy where `backendRefs` came from.

This PR creates additional maps both in `Rule` and `ResourceRule` so we could get this information.

Also, the PR fixes https://github.com/kumahq/kuma/issues/9173 since I reused `origin(...)` method from `resourcerule.go` in `rules.go`

### Checklist prior to review


pkg/api-server/testdata/inspect_dataplane_rules_subset.golden.json
```diff
@@ -135,11 +135,11 @@
    "origins": [
     {
      "mesh": "default",
-     "name": "mtp-1"
```
lukidzi, line 138: Why has the order changed?

lobkovilya: That's on purpose, since I reused the new `origin(...)` method from `resourcerules.go` in `rules.go` I fixed the issue we had https://github.com/kumahq/kuma/issues/9173.

lukidzi (approved):

pkg/plugins/policies/core/rules/testdata/resourcerules/mhr_backendrefs.global-k8s-2-zone-k8s.golden.yaml
```diff
@@ -0,0 +1,72 @@
+meshservice:mesh/mesh-1:namespace/ns-k8s:name/backend:
+  BackendRefOriginIndex:
+    INA9CySpV6H/9i+YojV8runPQRKlTJorpk2wpmoIjeE=: 0
```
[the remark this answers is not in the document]
lobkovilya: You can see how it'll be used here https://github.com/kumahq/kuma/pull/11203/files#diff-e1922605e03ea1f5b0d00abbf2d30f53dfb69c1854bc307b12160242db68c8abR47. We don't need to include this hash, you can always compute this hash based on `Matches` array

Issue #11135, opened 2024-08-19
The universal DP is delayed to generate the XDS configuration


Icarus9913:
### What happened?

When I try to figure out how the xds Endpoints data were generated, I found that there's delay for the universal DP XDS configuration generation.

Refer to the [XDS callbackchain](https://github.com/kumahq/kuma/blob/ffb0a135b8320cd474c503b785b43627d77a1338/pkg/xds/server/v3/components.go#L54-L56), the universal DP XDS generation happens in the grpc stream `OnStreamRequest`.

To put it shortly, the callback functions were call one by one in order

```text
callback {
...
   function dataplane_metadata_tracker                    // store the DP metadata in cache
   
   function goroutine dataplane_sync_tracker           // watchdog ontick reconciler to generate XDS configs
   
   function dataplane_lifecycle                                    // create the dataplane object
...
}
```

With the pseudocode, you can see the second step was called in goroutine and it relies on the third step dataplane object creation. There's a case that the second step try to generate the XDS configs but no dataplane exists, it has to wait for the **next time** reconciler to do that.(PS. the interval duration is defined by ENV `KUMA_XDS_SERVER_DATAPLANE_CONFIGURATION_REFRESH_INTERVAL`)

Actually, we can use the `metadata.Resource` to generate the XDS configurations which is used to create the `Dataplane` resource at the third step. Also, you can see the function [GetDataplaneResource](https://github.com/kumahq/kuma/blob/ffb0a135b8320cd474c503b785b43627d77a1338/pkg/core/xds/metadata.go#L68) is never called.

Icarus9913:
I also tested it in my local environment.

```text
2024-08-19T01:23:46.141+0800	INFO	xds.dataplaneSyncTracker	sync/dataplane_watchdog.go:160	Dataplane object not found. Can't regenerate XDS configuration. It's expected during Kubernetes namespace termination. If it persists it's a bug.	{"key": {"Mesh":"default","Name":"dp-echo-1"}}
```

This log was output by the second step, once the third step callback function was called then it went to well.

lukidzi:
> Triage: Let's revert the order and create the Dataplane before the sync call

PR #11180, opened 2024-08-23
fix(xds): accelerate universal dp XDS generation

Icarus9913:
In the XDS callback chain, the dataplane sync tracker callback relies on the dataplane object to generate the xds configurations. However, the dataplane object creation happens in dataplane liftcycle callback which is latter than the sync tracker callback. Consequetenly, just revert these 2 order and create the dataplane object before the sync call.

Reference issue: #11135

### Checklist prior to review


lukidzi (approved):
@jakubdyszkiewicz Do you know if that order was important? I think it was visible after this fix: https://github.com/kumahq/kuma/issues/11094

jakubdyszkiewicz:
It makes sense, but it does not fully fix the problem.

We retrieve DPP for which we generate XDS from the MeshContext to avoid additional queries. MeshContext is cached (by default for 1s). This means that between creation of DPP and running XDS watchdog you need to invalidate the cache to pick up new DPP. This may not be the case very often.
If you were testing this on a single DPP you always had invalid cache, because no DPPs connected = no cached MeshContext.

Icarus9913:
> It makes sense, but it does not fully fix the problem.
> 
> We retrieve DPP for which we generate XDS from the MeshContext to avoid additional queries. MeshContext is cached (by default for 1s). This means that between creation of DPP and running XDS watchdog you need to invalidate the cache to pick up new DPP. This may not be the case very often. If you were testing this on a single DPP you always had invalid cache, because no DPPs connected = no cached MeshContext.

Yeah, you're right.
Once the `Dataplane` resource was created, the xds `Sync` would try to retrieve this `Dataplane` object by `ReadOnlyResourceManager` which uses the cache to get the `MeshContext` first.

https://github.com/kumahq/kuma/blob/e21c751d782bc06eb7e8c779dacf71097bfe5732/pkg/core/resources/manager/cache.go#L93-L117

The default cache expiration is `1s`: https://github.com/kumahq/kuma/blob/e21c751d782bc06eb7e8c779dacf71097bfe5732/pkg/config/app/kuma-cp/kuma-cp.defaults.yaml#L71

So, this PR doesn't fix the issue totally.

Icarus9913:
Maybe we should set an interval between `goroutine dataplane_sync_tracker` and `dataplane_lifecycle`

jakubdyszkiewicz:
Yeah, I think sleeping after creating Universal DP for cache expiration time makes sense.

Btw. it's important to do this only when we create the resource in lifecycle. On Kube, Dataplane object is created by the controller, not by the callback

pkg/xds/server/callbacks/dataplane_lifecycle.go
```diff
@@ -142,6 +145,10 @@ func (d *DataplaneLifecycle) register(
 		return errors.Wrap(err, "could not register proxy passed in kuma-dp run")
 	}
 
+	// We should wait for Cache ExpirationTime to let MeshContext sync the latest data
+	// in which the latter DataplaneSyncTracker callback relies on it to generate the XDS configuration.
+	// This only happens on Universal Dataplane because the k8s Dataplane object is created by the controller.
+	time.Sleep(d.cacheExpirationTime)
```
jakubdyszkiewicz, line 151: the problem is that this sleep will block
```
	defer info.mtx.Unlock()
```
so ALL dpps that are trying to connect will be waiting

jakubdyszkiewicz (changes requested):

pkg/xds/server/callbacks/dataplane_lifecycle.go
Icarus9913: IIUC, the variable `info` is corresponding to a unique `proxyKey`.

```golang
info, loaded := d.proxyInfos.LoadOrStore(proxyKey, &proxyInfo{
  proxyType: md.GetProxyType(),
})
```

And the proxyKey is generated by the envoy nodeID:

```golang
proxyId, err := core_xds.ParseProxyIdFromString(request.NodeId())
if err != nil {
  return errors.Wrap(err, "invalid node ID")
}
dpKey := proxyId.ToResourceKey()
```

So, this lock would only block the unique dataplane object rather than all of them right?
Here's the envoy bootstrap configuration that I downloaded, the `dp-echo-1` is the universal dp name that I defined:

```text
node:
  cluster: backend
  id: default.dp-echo-1
  metadata:
    dataplane.admin.address: 127.0.0.1
```

jakubdyszkiewicz: I read this wrong, you are right with this one

jakubdyszkiewicz (approved):

PR #11203, opened 2024-08-23
feat(kuma-cp): use ResourceIdentifier in MeshContext structs

lobkovilya:
* use ResourceIdentifier in mesh context and proxy objects
* change in the public API for MeshMultiZoneService status. Storing only `name` of the MeshService is not enough to uniquely identify the resource
* resolve BackendRefs in MeshTCPRoute and MeshHTTPRoute

### Checklist prior to review


pkg/core/resources/model/resource.go
```diff
@@ -704,6 +704,26 @@ func TargetRefToResourceIdentifier(meta ResourceMeta, tr common_api.TargetRef) R
...
+	case br.Kind == common_api.MeshExternalService:
+	case br.Kind == common_api.MeshMultiZoneService:
+	default:
+		return resolved
+	}
+
+	resolved.Resource = &TypedResourceIdentifier{
+		ResourceIdentifier: TargetRefToResourceIdentifier(meta, br.TargetRef),
+		ResourceType:       ResourceType(br.Kind),
+		SectionName:        fmt.Sprintf("%d", br.Port),
```
lukidzi, line 721: We should check if the port is `nil`. I think it's not required for `MeshExternalService`, which might cause a panic.

pkg/xds/context/context.go
```diff
@@ -65,8 +67,7 @@ type LabelValue struct {
 	Value string
 }
 type (
-	ServiceName                  string
-	LabelsToValuesToServiceNames map[LabelValue]map[ServiceName]bool
+	LabelsToValuesToServiceNames map[LabelValue]map[core_model.ResourceIdentifier]bool
```
lukidzi, line 70: Should we rename? `LabelValuesToResourceIdentifiers`

pkg/xds/context/context.go
```diff
@@ -78,11 +79,11 @@ type MeshContext struct {
 	Resource                                *core_mesh.MeshResource
 	Resources                               Resources
 	DataplanesByName                        map[string]*core_mesh.DataplaneResource
-	MeshServiceByName                       map[string]*meshservice_api.MeshServiceResource
+	MeshServiceByName                       map[core_model.ResourceIdentifier]*meshservice_api.MeshServiceResource
```
lukidzi, line 82: Maybe this one too, `MeshServiceByIdentifiers`

pkg/xds/sync/dataplane_proxy_builder.go
```diff
@@ -111,7 +114,25 @@ func (p *DataplaneProxyBuilder) resolveVIPOutbounds(meshContext xds_context.Mesh
 	if dataplane.Spec.Networking.GetTransparentProxying() == nil {
 		newOutbounds := []*core_xds.Outbound{}
 		for _, o := range dataplane.Spec.Networking.Outbound {
-			newOutbounds = append(newOutbounds, &core_xds.Outbound{LegacyOutbound: o})
+			if o.BackendRef != nil {
+				newOutbounds = append(newOutbounds, &core_xds.Outbound{
```
lukidzi, line 118: Maybe we can extract this into a new method?

pkg/test/xds/builders/context_builder.go
```diff
@@ -22,8 +23,8 @@ func Context() *ContextBuilder {
 				Resource:                  samples.MeshDefault(),
 				EndpointMap:               map[core_xds.ServiceName][]core_xds.Endpoint{},
 				ServicesInformation:       map[string]*xds_context.ServiceInformation{},
-				MeshExternalServiceByName: map[string]*meshexternalservice_api.MeshExternalServiceResource{},
-				MeshServiceByName:         map[string]*meshservice_api.MeshServiceResource{},
+				MeshExternalServiceByName: map[model.ResourceIdentifier]*meshexternalservice_api.MeshExternalServiceResource{},
```
jakubdyszkiewicz, line 26: Is it still by the name then?

pkg/xds/context/context.go
```diff
@@ -101,70 +102,66 @@ type ServiceInformation struct {
...
+		} else {
+			key := core_model.TypedResourceIdentifier{
+				ResourceType: core_model.ResourceType(reachableBackend.Kind),
+				ResourceIdentifier: core_model.TargetRefToResourceIdentifier(dataplane.GetMeta(), common_api.TargetRef{
+					Name:      reachableBackend.Name,
+					Namespace: reachableBackend.Namespace,
+				}),
+			}
+			if key.Zone == "" {
+				fmt.Println("DEBUG", "zone is not set", dataplane.GetMeta().GetLabels(), dataplane.GetMeta().GetMesh(), dataplane.GetMeta().GetName())
```
jakubdyszkiewicz, line 133: remove and one more a couple of lines below

pkg/xds/envoy/listeners/listener_builder.go
```diff
@@ -47,6 +47,9 @@ func NewOutboundListenerBuilder(
 	port uint32,
 	protocol core_xds.SocketAddressProtocol,
 ) *ListenerBuilder {
+	if address == "" {
+		address = "127.0.0.1"
```
jakubdyszkiewicz, line 51: where is this used? I wonder about IPV6

lobkovilya: This is just the logic we have here https://github.com/kumahq/kuma/blob/master/api/mesh/v1alpha1/dataplane_helpers.go#L239. I can open an issue to figure this out

jakubdyszkiewicz: I see, let's leave it as is if this results in the same behavior

jakubdyszkiewicz (approved):
I'm having a bit of a hard time to wrap my head around those changes. Aside of some nits it looks ok

pkg/xds/sync/dataplane_proxy_builder.go
lobkovilya: I created `AsOutbounds()` method for MeshService, MeshExternalService, MeshMultiZoneService, Dataplane and it simplified the code for mesh and dpp context builders.

pkg/core/resources/model/resource.go
lobkovilya: Good catch, fixed

lukidzi (approved):

PR #11207, opened 2024-08-26
fix(transparent-proxy): prioritize IPv4 executables initialization

bartsmykla:
Improve IP-related initialization by first setting up IPv4 executables to determine the iptables type (nft or legacy). IPv6 initialization now occurs only if IPFamilyMode is set to DualStack, an IPv6 local address exists, and adding the `::6` address succeeds. This approach prevents failures in environments like Minikube, where nft is unavailable and IPv6 is unsupported, ensuring the correct selection of executables without errors.

### Checklist prior to review

  - No relevant links
  - It won't
  - Manually tested + all transparent proxy tests are passing
  - There is no need
  - There is no need


michaelbeaumont (approved):
Works for me

PR #11245, opened 2024-08-28
test(e2e/universal): copy container logs during e2e universal debug

bartsmykla:
So far we didn't have access to logs of workloads in docker containers when e2e universal test fails. This commit copies files with logs to appropriate debug directory, which then is uploaded as an artifact via our CI.

### Checklist prior to review

  - No relevant issues
  - It won't
  - Locally tested
  - There is no need
  - Only to 2.8

> Changelog: skip

### DIFF
--- a/test/framework/debug.go
+++ b/test/framework/debug.go
@@ -2,17 +2,21 @@ package framework
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/google/uuid"
 	"github.com/gruntwork-io/terratest/modules/k8s"
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
 
 	"github.com/kumahq/kuma/test/framework/kumactl"
+	"github.com/kumahq/kuma/test/framework/universal_logs"
 )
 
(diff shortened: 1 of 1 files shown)


test/framework/debug.go
```diff
@@ -205,3 +233,114 @@ func CpRestarted(cluster Cluster) bool {
...
+	for _, entry := range entries {
+		sourcePath := filepath.Join(scrDir, entry.Name())
+		destPath := filepath.Join(dest, entry.Name())
+
+		fileInfo, err := os.Stat(sourcePath)
+		if err != nil {
+			return err
+		}
+
+		stat, ok := fileInfo.Sys().(*syscall.Stat_t)
```
michaelbeaumont, line 256: I guess we don't run tests on Darwin/Windows so this shouldn't matter?

michaelbeaumont (approved):

test/framework/debug.go
bartsmykla: It works on darwin: https://github.com/golang/go/blob/master/src/syscall/ztypes_darwin_arm64.go#L65-L85

jakubdyszkiewicz (approved):

test/framework/debug.go
bartsmykla: I would love to get rid of these functions, but we can't currently smoothly upgrade to go 1.23 because of the version of golangci-linter. Tried to quickly update it, but it's more complex

Issue #11134, opened 2024-08-19
Implement MeshTLS policy


slonka:
### What happened?

continuation from https://github.com/kumahq/kuma/issues/4622

details TBD.

PR #11254, opened 2024-08-28
feat(meshtls): policy implementation

slonka:
### Checklist prior to review

* added e2e multizone and universal
* added validation
* added plugin implementation, which has the same logic as `inbound_proxy_generator.go` for strict/permissive
* configuration of TLS version and cipher is only available when using Mesh/Mesh 
* because we don't have a `to` policy we take the first `from` policy to configure tls version and cipher 
* when MeshTLS policy is defined and mode is not provided we set is as strict

- [x] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/11134


pkg/plugins/policies/meshtls/plugin/v1alpha1/testdata/permissive-no-mtls.policy.yaml
```diff
@@ -0,0 +1,12 @@
...
+  kind: Mesh
+from:
+  - targetRef:
+      kind: Mesh
+    default:
+      tlsVersion:
+        min: TLS12
+        max: TLS13
+      tlsCiphers:
+        - "ECDHE-ECDSA-AES128-GCM-SHA256"
```
michaelbeaumont, line 11: This doesn't show up in the golden file I dont think

pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin.go
```diff
@@ -26,6 +45,207 @@ func (p plugin) MatchedPolicies(dataplane *core_mesh.DataplaneResource, resource
...
+			continue
+		}
+		l, err := configure(proxy, listener, iface, inbound, conf, ctx)
+		if err != nil {
+			return err
+		}
+		if l != nil {
+			rs.Remove(resource.ListenerType, listener.GetName())
+			rs.Add(&core_xds.Resource{
+				Name:     listener.Name,
```
michaelbeaumont, line 98:
```suggestion
				Name:     listener.GetName(),
```

pkg/plugins/policies/meshtls/plugin/v1alpha1/testdata/permissive-no-mtls.policy.yaml
lukidzi: Because this test has mTLS disabled

michaelbeaumont (approved):

pkg/plugins/policies/meshtls/plugin/v1alpha1/testdata/permissive-no-mtls.policy.yaml
michaelbeaumont: ah ok, disabled on the Mesh level

PR #11296, opened 2024-09-03
feat(kuma-cp): implement delta xDS for envoy config exchange

lukidzi:
## Motivation

Envoy supports incremental xDS, which sends only changes rather than the entire state.

## Implementation information

* Run a Kuma 2-cluster multizone setup with incremental xDS. While https://github.com/kumahq/go-control-plane/pull/4 is still needed, it’s not required for Envoy 1.32+ since it supports the Endpoints cache when initial_fetch_timeout is set.
* Extracted common logic in each callback to support both communication types.
* Each callback now uses a separate map to store connection information, depending on mode, preventing stream conflicts.
* Added a global and dataplane-only flag to toggle incremental xDS for all DPPs or a single DPP.
* 
## Supporting documentation

https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#incremental-xds

Fix #XX


pkg/config/app/kuma-cp/config.go
```diff
@@ -464,6 +464,8 @@ type ExperimentalConfig struct {
 	// If true skips persisted VIPs. Change to true only if generateMeshServices is enabled.
 	// Do not enable on production.
 	SkipPersistedVIPs bool `json:"skipPersistedVIPs" envconfig:"KUMA_EXPERIMENTAL_SKIP_PERSISTED_VIPS"`
+	// If true uses Delta xDS to deliver changes to sidecars.
+	UseDeltaXDS bool `json:"useDeltaXDS" envconfig:"KUMA_EXPERIMENTAL_USE_DELTA_XDS"`
```
Icarus9913, line 468: Do we still need this switch ? 
Or just enable the `rest, sotw, delta` simultaneously?

lukidzi: It aims to force all DPPs to use delta instead of sotw. So it's a global switch

Icarus9913:
Is it possible with this case?
2.9.0(CP|DP)  --> 2.10.0(CP|DP) --> 2.9.0(CP), 2.10.0(DP)

We just allow **hight version** CP wth **low version** DP right?

pkg/xds/bootstrap/generator.go
```diff
@@ -124,6 +127,14 @@ func (b *bootstrapGenerator) Generate(ctx context.Context, request types.Bootstr
 			params.AdminPort = b.defaultAdminPort
 		}
 	}
+	switch request.XdsConfigMode {
+	case types.DELTA:
+		params.UseDelta = true
```
Icarus9913, line 132: If the CP disabled `UseDeltaXds` and the DP request DeltaXDSMode, should we return the DP an error?

lukidzi: No, we should allow individual DPPs to run with Delta. This approach enables a progressive migration. UseDeltaXds enforces Delta mode for all DPPs but still allows individual DPPs to revert to the previous mode if needed.

Icarus9913: OK, I understand it right now. With the file `pkg/xds/server/v3/components.go` we enable the deltaServer by default. I think maybe we shold rename the variable `Experimental.UseDeltaXds` because we may thought it was a swicth to enable or disable the whole `delta XDS` feature.

So we got these configurations:
1. run the delta server by default
2. CP `Experimental.UseDeltaXds` force enable deltaXDS for all DPPs
3. DP `ENV KUMA_EXPERIMENTAL_USE_DELTA_XDS` to choose use stow or delta

lukidzi: I think the DPP environment variable is clear, and it shouldn't be a boolean. If it were, we wouldn’t be able to distinguish between using SOTW and "Not Defined." For the control plane configuration, you're right—I need to find a better name for it

pkg/config/app/kuma-cp/config.go
```diff
@@ -464,6 +464,8 @@ type ExperimentalConfig struct {
 	// If true skips persisted VIPs. Change to true only if generateMeshServices is enabled.
 	// Do not enable on production.
 	SkipPersistedVIPs bool `json:"skipPersistedVIPs" envconfig:"KUMA_EXPERIMENTAL_SKIP_PERSISTED_VIPS"`
+	// If true uses Delta xDS to deliver changes to sidecars.
+	UseDeltaXds bool `json:"useDeltaXds" envconfig:"KUMA_EXPERIMENTAL_USE_DELTA_XDS"`
```
lahabana, line 468:
```suggestion
	DeltaXds bool `json:"deltaXds" envconfig:"KUMA_EXPERIMENTAL_DELTA_XDS"`
```

It's simpler and just as expresive

pkg/xds/bootstrap/generator.go
lahabana: The docs talk about XDSTransportProtocolVariant: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#variants-of-the-xds-transport-protocol

lahabana: Or Just APIType:

```

// APIs may be fetched via either REST or gRPC.
type ApiConfigSource_ApiType int32

const (
	// Ideally this would be 'reserved 0' but one can't reserve the default
	// value. Instead we throw an exception if this is ever used.
	//
	// Deprecated: Marked as deprecated in envoy/config/core/v3/config_source.proto.
	ApiConfigSource_DEPRECATED_AND_UNAVAILABLE_DO_NOT_USE ApiConfigSource_ApiType = 0
	// REST-JSON v2 API. The `canonical JSON encoding
	// <https://developers.google.com/protocol-buffers/docs/proto3#json>`_ for
	// the v2 protos is used.
	ApiConfigSource_REST ApiConfigSource_ApiType = 1
	// SotW gRPC service.
	ApiConfigSource_GRPC ApiConfigSource_ApiType = 2
	// Using the delta xDS gRPC service, i.e. DeltaDiscovery{Request,Response}
	// rather than Discovery{Request,Response}. Rather than sending Envoy the entire state
	// with every update, the xDS server only sends what has changed since the last update.
	ApiConfigSource_DELTA_GRPC ApiConfigSource_ApiType = 3
	// SotW xDS gRPC with ADS. All resources which resolve to this configuration source will be
	// multiplexed on a single connection to an ADS endpoint.
	// [#not-implemented-hide:]
	ApiConfigSource_AGGREGATED_GRPC ApiConfigSource_ApiType = 5
	// Delta xDS gRPC with ADS. All resources which resolve to this configuration source will be
	// multiplexed on a single connection to an ADS endpoint.
	// [#not-implemented-hide:]
	ApiConfigSource_AGGREGATED_DELTA_GRPC ApiConfigSource_ApiType = 6
)
```
In which I'd use the values: `DELTA_GRPC` and `GRPC` btw how come we're not using `AGGREGATED_GRPC` ?

pkg/xds/bootstrap/types/bootstrap_request.go
```diff
@@ -21,8 +21,17 @@ type BootstrapRequest struct {
...
+	XdsConfigMode        XdsMode           `json:"xdsConfigMode,omitempty"`
 }
 
+type XdsMode int
+
+const (
+	NOT_DEFINED XdsMode = iota
+	SOTW
+	DELTA
+)
```
lahabana, line 33: This is a json request let's use string for enum please

lahabana (commented):
Didn't look at implem only the API.

This is looking interesting! One thing that isn't clear to me is how you do the opt-in on Kubernetes. How do I set the dataplane env var to just opt-in 1 deployment?

pkg/config/app/kuma-cp/config.go
```diff
@@ -459,6 +459,8 @@ type ExperimentalConfig struct {
 	// Enables sidecar containers in Kubernetes if supported by the Kubernetes
 	// environment.
 	SidecarContainers bool `json:"sidecarContainers" envconfig:"KUMA_EXPERIMENTAL_SIDECAR_CONTAINERS"`
+	// If true uses Delta xDS to deliver changes to sidecars.
+	DeltaXds bool `json:"deltaXds" envconfig:"KUMA_EXPERIMENTAL_DELTA_XDS"`
```
Icarus9913, line 463: Also one further question for the ENV variable.
In the future we'll get rid of `EXPERIMENTAL` right? 
At that time, do we support both `KUMA_EXPERIMENTAL_DELTA_XDS ` and `KUMA_DELTA_XDS` or just support `KUMA_DELTA_XDS` ?

pkg/plugins/runtime/k8s/metadata/annotations.go
```diff
@@ -134,6 +134,10 @@ const (
 
 	// HeadlessService is "true" when the Service had ClusterIP: None, otherwise "false"
 	HeadlessService = "k8s.kuma.io/is-headless-service"
+
+	// KumaXdsTransportProtocolVariant allows to specify mode in which control-plane exchange configuration with the sidecar.
+	// Available values are: DELTA_GRPC and GRPC
+	KumaXdsTransportProtocolVariant = "kuma.io/xds-transport-protocol-variant"
```
Icarus9913, line 140: Did you miss the webhook validation for this annotation's value?

api/mesh/v1alpha1/envoy.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package kuma.mesh.v1alpha1;
+
+option go_package = "github.com/kumahq/kuma/api/mesh/v1alpha1";
+
+message EnvoyConfiguration {
+  enum XdsTransportProtocolVariant {
+    // This value defines default control-plane variant.
```
Icarus9913, line 9: Can we enrich this sentence?
I mean `This value defines default control-plane variant, it depends on the CP DeltaXds switch to enable incremental xDS or just enable state of the world xDS` some stuff like this, otherwise we don't know the default behavior at first glance.

Icarus9913: All right, maybe in the future once we'll remove the `KUMA_EXPERIMENTAL_DELTA_XDS` and just enable the deltaXDS by default, so there's no options for the CP but just configured by the DP's exclusive options.

pkg/plugins/runtime/k8s/metadata/annotations.go
lukidzi: I was reviewing the code and noticed that we don't currently validate all annotation values. I've updated the code to throw an error during Dataplane conversion if an invalid value is found

Icarus9913 (approved):

api/mesh/v1alpha1/zone_ingress.proto
```diff
@@ -68,4 +69,7 @@ message ZoneIngress {
   // AvailableService contains tags that represent unique subset of
   // endpoints
   repeated AvailableService availableServices = 3;
+
+  // EnvoyConfiguration provides additional configuration for the Envoy sidecar.
+  EnvoyConfiguration envoy = 4;
```
slonka, line 74: Why do we need that in zone ingress? (I might answer that question myself later)

pkg/hds/tracker/callbacks.go
```diff
@@ -57,10 +58,11 @@ func NewCallbacks(
 	return &tracker{
 		resourceManager:    resourceManager,
 		streamsAssociation: map[xds.StreamID]core_model.ResourceKey{},
-		dpStreams:          map[core_model.ResourceKey]streams{},
-		config:             config,
-		log:                log,
-		metrics:            metrics,
+		// deltaStreamsAssociation: map[xds.StreamID]core_model.ResourceKey{},
```
slonka, line 61: why do we need this commented out?

pkg/xds/bootstrap/generator.go
```diff
@@ -38,6 +38,7 @@ func NewDefaultBootstrapGenerator(
 	enableReloadableTokens bool,
 	hdsEnabled bool,
 	defaultAdminPort uint32,
+	deltaXdsEnabled bool,
```
slonka, line 41: do we need a builder for this? it's getting quite packed with parameters

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -104,14 +146,27 @@ func (d *xdsCallbacks) OnStreamRequest(streamID core_xds.StreamID, request util_
...
+	} else {
+		dpStream = d.dpStreams[streamID]
+	}
 	dpStream.dp = &dpKey
-	d.dpStreams[streamID] = dpStream
+	if isDelta {
+		d.dpDeltaStreams[streamID] = dpStream
+	} else {
+		d.dpStreams[streamID] = dpStream
+	}
```
slonka, line 169: maybe instead of this constant if/else you could have a generic function that takes either `dpStreams` or `dpDeltaStreams`? this applies to all functions with `isDelta`

slonka (commented):
A couple of Qs but really happy that we're going to be shipping this 🚢

lukidzi:
> A couple of Qs but really happy that we're going to be shipping this 🚢 . Also does go-control-plane handle the stuff like hashing individual resources instead of whole arrays?

Yes, it should verify individual resources so only one endpoint will be sent. I don't know if that is the question.

pkg/xds/bootstrap/generator.go
lukidzi: We could change it, but probably in a separate PR

api/mesh/v1alpha1/zone_ingress.proto
lukidzi: It's because Egress and Ingress also can use delta for configuration exchange

lukidzi:
I think there is still some edge case causing flake, need to investigate

pkg/xds/bootstrap/types/bootstrap_request.go
```diff
@@ -23,6 +23,14 @@ type BootstrapRequest struct {
 	SystemCaPath         string            `json:"systemCaPath"`
 }
 
+type XdsMode string
```
jijiechen, line 26: We are using "XdsTransportProtocolVariant" at many places, so why not use "Variant" here to replace "Mode"?

lukidzi:
Closing in favor of https://github.com/kumahq/kuma/pull/13356 since i can run more ci stability there

PR #11300, opened 2024-09-04
fix(kuma-cp): consumer scoped policies should be applied only on dpps from the same namespace

Automaat:
Without this change, consumer policy is matched to all dpps in zone no matter the namespace, which is wrong as it should be namespace scoped

### Checklist prior to review


pkg/plugins/policies/core/matchers/dataplane.go
```diff
@@ -171,6 +174,15 @@ func dppSelectedByPolicy(
 	}
 }
 
+func ddpSelectedByNamespace(meta core_model.ResourceMeta, dpp *core_mesh.DataplaneResource) bool {
+	if meta.GetLabels()[mesh_proto.PolicyRoleLabel] == string(mesh_proto.ConsumerPolicyRole) {
```
lobkovilya, line 178: should be both consumer and workload-owner https://github.com/kumahq/kuma/blob/180976f81dcb6a9b86c4c3fa534e11534060171d/docs/madr/decisions/052-policy-on-namespace.md?plain=1#L103-L102

pkg/plugins/policies/core/matchers/dataplane.go
```diff
@@ -129,6 +129,9 @@ func dppSelectedByPolicy(
 ) ([]core_rules.InboundListener, []core_rules.InboundListenerHostname, bool, error) {
 	switch ref.Kind {
 	case common_api.Mesh:
+		if !ddpSelectedByNamespace(meta, dpp) {
```
lobkovilya, line 132: it's not just for kind Mesh, it should be checked for all kinds no?

pkg/plugins/policies/core/matchers/dataplane.go
Automaat: changed

pkg/plugins/policies/core/matchers/dataplane.go
Automaat: you are right it applies to all types, changed

lobkovilya (approved):

Issue #10588, opened 2024-06-21
add dnsConfig to egress/ingress deployments in kuma helm chart


jijiechen:
### Description

It's asked by some kuma users that they want kuma helm chart to support configuring `dnsConfig` for egress and ingress deployments.

They are currently manually adding the configuration to their deployments after kuma installation, and they need to redo this each time they upgrade kuma.

slonka:
> Triage: this is passed through to the Kube DNS, it's just plumming through the DNS config

PR #11307, opened 2024-09-04
chore(helm): added option to configure controlplane/ingress/egress dns

lukidzi:
### Checklist prior to review

Added an option in a helm values to configure DNS of controlPlane, ingress and egress pods

- [X] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/10588


jijiechen (approved):

app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
```diff
@@ -331,6 +331,16 @@ controlPlane:
   # -- If true, then control plane can support TLS secrets for builtin gateway outside of mesh system namespace.
   # The downside is that control plane requires permission to read Secrets in all namespaces.
   supportGatewaySecretsInAllNamespaces: false
+  # -- DNS configuration for the pod.
```
slonka, line 334: for the CP pod, right? should we make it clear?

app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
```diff
@@ -331,6 +331,16 @@ controlPlane:
...
+  dns:
+    # -- Defines how DNS resolution is configured for that Pod.
```
slonka, line 336: should we state that this is equivalent to https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy and what values are available?

deployments/charts/kuma/templates/cp-deployment.yaml
```diff
@@ -409,3 +409,21 @@ spec:
...
+        - {{ $nameserver }}
+        {{- end }}
+        {{- end }}
+        {{- if gt (len .Values.controlPlane.dns.config.searches) 0 }}
+        searches:
+        {{- range $search := .Values.controlPlane.dns.config.searches }}
+        - {{ $search }}
+        {{- end }}
+        {{- end }}
+      {{- end }}
```
slonka, line 429: Could this be a function that we would call so we don't need to define this 3 times?

slonka (commented):
Maybe we could improve the comments and dry up the implem a bit - otherwise 👍

jijiechen (approved):

slonka (approved):

PR #11316, opened 2024-09-05
refactor(transparent-proxy): move `contsts` package and remove dot imports

bartsmykla:
The constants in the `consts` package are no longer specific to iptables, so moving the package to a higher-level directory improves organization and better reflects its broader usage across the codebase. Dot aliases for imports can cause confusion and disrupt IDE functionality, especially when refactoring or moving packages. In the case of transparent proxy constants, the value of using dot imports doesn't justify the potential downsides, so this commit replaces them with explicit imports for clarity and maintainability.

I'm skipping tests as this is transparent change and have completely no change in any logic

### Checklist prior to review

  - No relevant issues
  - It won't
  - Unit tests and transparent tests ran locally
  - There i no need
  - There i no need

> Changelog: skip


jijiechen (approved):
All are simple renaming the reference. Looks good to me.

PR #11327, opened 2024-09-06
perf(k8s): do not update resource on control-plane restart

lukidzi:
### Context

While running on Kubernetes, Kuma has different controllers responsible for translating Kubernetes objects into Kuma-specific objects, such as Dataplane, ZoneIngress, or ZoneEgress. Each change to a pod or service that affects the internal model is subsequently stored in the database.

I've noticed that when running Kuma on Kubernetes, after each restart of the control-plane, the Dataplane, ZoneIngress, or ZoneEgress resources are updated in the store, even when there is no change in the actual proto object.

### Root cause

Kuma stores a resource's spec as [JSON in a Kubernetes store](https://github.com/kumahq/kuma/blob/master/pkg/plugins/resources/k8s/native/api/v1alpha1/zz_generated.mesh.go#L143). When there is a control-plane restart, the controller responsible for the pod-to-Dataplane mapping performs the following steps:

1. Verify if the event requires reconciliation.
2. Call [CreateOrUpdate](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controllerutil/controllerutil.go#L292) in the controller-runtime: 
  a) Check if the resource exists. If it doesn't exist, go to step c). 
  b) Create a copy of the existing Dataplane object. 
  c) Mutate the object by calling a method that generates a new Dataplane object based on the pod, service, and namespace (and other Dataplanes if KUMA_EXPERIMENTAL_KUBE_OUTBOUNDS_AS_VIPS is false). 
  d) Check if the existing object is the same as the new one.
At step d), we were detecting a change even when the Protobuf objects were identical. This issue was caused by Kubernetes returning the existing JSON with fields sorted differently than in the newly generated resource, which follows the order defined by the Protobuf schema.

For example, the JSON returned for the existing Dataplane:

```json
"{\"networking\":{\"address\":\"10.42.0.8\",\"admin\":{\"port\":9901},\"inbound\":[{\"health\":{\"ready\":true},\"port\":5000,\"tags\":{\"app\":\"demo-app\",\"k8s.kuma.io/namespace\":\"kuma-demo\",\"k8s.kuma.io/service-name\":\"demo-app\",\"k8s.kuma.io/service-port\":\"5000\",\"kubernetes.io/hostname\":\"k3d-kuma-server-0\",\"kuma.io/protocol\":\"http\",\"kuma.io/service\":\"demo-app_kuma-demo_svc_5000\",\"kuma.io/zone\":\"default\",\"pod-template-hash\":\"d789d49c\"}}],\"transparentProxying\":{\"ipFamilyMode\":\"DualStack\",\"redirectPortInbound\":15006,\"redirectPortOutbound\":15001}},\"probes\":{\"port\":9000}}"}
```

Whereas the newly generated JSON might look like this:
```json
"{\"networking\":{\"address\":\"10.42.0.8\",\"inbound\":[{\"port\":5000,\"tags\":{\"app\":\"demo-app\",\"k8s.kuma.io/namespace\":\"kuma-demo\",\"k8s.kuma.io/service-name\":\"demo-app\",\"k8s.kuma.io/service-port\":\"5000\",\"kubernetes.io/hostname\":\"k3d-kuma-server-0\",\"kuma.io/protocol\":\"http\",\"kuma.io/service\":\"demo-app_kuma-demo_svc_5000\",\"kuma.io/zone\":\"default\",\"pod-template-hash\":\"d789d49c\"},\"health\":{\"ready\":true}}],\"transparentProxying\":{\"redirectPortInbound\":15006,\"redirectPortOutbound\":15001,\"ipFamilyMode\":\"DualStack\"},\"admin\":{\"port\":9901}},\"probes\":{\"port\":9000}}"
```

Since this data is stored as a byte array, DeepEqual detects a change in the order and causes all Dataplanes to be updated on each control-plane leader restart.

### Current solution

Since the [JSON RFC](https://www.rfc-editor.org/rfc/rfc7159) states:
> An object is an unordered collection of zero or more name/value
   pairs, where a name is a string and a value is a string, number,
   boolean, null, object, or array.

we cannot rely on the order of fields in a JSON object. Additionally, byte comparison is error-prone, especially when empty objects start being included or when sorting changes.

We have two options:

1. Figure out how to make all JSON maps sorted (although this is challenging since JSON does not guarantee order).
2. **Verify if the Protobuf objects are the same before performing an update.**

I have updated the logic, and the flow is now as follows:

1. Create a Dataplane object from the existing resources.
2. Retrieve the existing Dataplane from the database.
3. Unmarshal the existing resource into a Protobuf object (to avoid comparing byte arrays and encountering the same issue).
4. Compare the existing Protobuf object with the new one. If there is no change, return; otherwise, continue.
5. Use the generated Dataplane to call `CreateOrUpdate`.

Since the Kubernetes client uses a cache, the second call to get the existing resource in `CreateOrUpdate` shouldn't impact performance. This change will help us avoid unnecessary updates.

### Checklist prior to review


lukidzi:
Please also take a look at this https://github.com/kumahq/kuma/pull/11328 lets choose the best. I feel like 2nd approach solves it globally but might have some impact on the perf

Icarus9913:
I found something interesting: Firstly I agree that the json is an unordered collection, however the golang `json.Marshal` would make it ordered by alphabetical key and it's also [documented](https://pkg.go.dev/encoding/json#Marshal) in the standard package: 
> The map keys are sorted and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above

And I think that's why I can't reproduce it in my local ENV programmed codes with `json.Marshal` function to handle with `map[string]any`.

I gonna check the `proto` docs about the `marshal`.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -352,6 +366,36 @@ func (r *PodReconciler) createOrUpdateDataplane(
...
+func (r *PodReconciler) hasResourceChange(ctx context.Context, resource kube_client.Object, newResourceSpec model.ResourceSpec) (bool, error) {
+	key := kube_client.ObjectKeyFromObject(resource)
+	// This call shouldn't affect performance because if we don't call it here `createOrUpdate` does this call.
+	// If the resource change, `createOrUpdate` calls it 2nd time but in this case it should use the value
+	// from the cache.
+	if err := r.Client.Get(ctx, key, resource); err != nil {
+		if !kube_apierrs.IsNotFound(err) {
+			return false, err
+		}
+	}
```
Icarus9913, line 378: Maybe we can just return as if `NotFound` occurred which could avoid the later `model.Equal` to be called to save performance cost.

```text
	if err := r.Client.Get(ctx, key, resource); err != nil {
		if kube_apierrs.IsNotFound(err) {
		        log.debug(...)
			return false, nil
		}else{
		   return false, err
		}
	}
```

PR #11328, opened 2024-09-06
perf(k8s): spec in k8s resource should be in deterministic order

lukidzi:
### Checklist prior to review

This is a second approach for [this PR](https://github.com/kumahq/kuma/pull/11327). Instead of checking the Protobuf object directly, I created a new method called `MustMarshalJSONSorted` and replaced every call to `SetSpec` with it. Now, each call to `SetSpec` should produce the same order as the one returned by Kubernetes, allowing us to detect only actual changes.

The downside of this solution is that it involves:

* Marshalling Protobuf to JSON,
* Unmarshalling JSON to a map,
* Marshalling the map back to JSON.
`json.Marshal` function sorts the values, as described [here](https://pkg.go.dev/encoding/json#Marshal).

### DIFF
--- a/pkg/plugins/resources/k8s/native/api/v1alpha1/zz_generated.system.go
+++ b/pkg/plugins/resources/k8s/native/api/v1alpha1/zz_generated.system.go
@@ -85,7 +85,7 @@ func (cb *Zone) SetSpec(spec core_model.ResourceSpec) {
 		panic(fmt.Sprintf("unexpected protobuf message type %T", spec))
 	}
 
-	cb.Spec = &apiextensionsv1.JSON{Raw: util_proto.MustMarshalJSON(s)}
+	cb.Spec = &apiextensionsv1.JSON{Raw: util_proto.MustMarshalJSONSorted(s)}
 }
 
 func (cb *Zone) GetStatus() (core_model.ResourceStatus, error) {
@@ -190,7 +190,7 @@ func (cb *ZoneInsight) SetSpec(spec core_model.ResourceSpec) {
 		panic(fmt.Sprintf("unexpected protobuf message type %T", spec))
 	}
 
-	cb.Spec = &apiextensionsv1.JSON{Raw: util_proto.MustMarshalJSON(s)}
+	cb.Spec = &apiextensionsv1.JSON{Raw: util_proto.MustMarshalJSONSorted(s)}
 }
 
 func (cb *ZoneInsight) GetStatus() (core_model.ResourceStatus, error) {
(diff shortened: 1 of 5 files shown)


Icarus9913:
The ordering will be [removed in v2 json package](https://github.com/golang/go/discussions/63397)
I suggest to go futher with the option 1 PR

pkg/plugins/resources/k8s/native/api/v1alpha1/zz_generated.system.go
```diff
@@ -190,7 +190,7 @@ func (cb *ZoneInsight) SetSpec(spec core_model.ResourceSpec) {
 		panic(fmt.Sprintf("unexpected protobuf message type %T", spec))
 	}
 
-	cb.Spec = &apiextensionsv1.JSON{Raw: util_proto.MustMarshalJSON(s)}
+	cb.Spec = &apiextensionsv1.JSON{Raw: util_proto.MustMarshalJSONSorted(s)}
```
lahabana, line 193: Dumb question but using `json.Marshal(s)` just doesn't work here?

lukidzi: I've tested it but unfortunately, order is not correct, it preservers proto order

lahabana: Would it be less hacky to do roughly what #11327 does but instead of doing the compare in the main controller code doing it in [PodToDataplane](https://github.com/kumahq/kuma/blob/c82b1190303912344b29b20051d09946238ec304/pkg/plugins/runtime/k8s/controllers/pod_converter.go#L36-L51) ? 
Doing something like:

```
	dataplane.Mesh = util_k8s.MeshOfByAnnotation(pod, ns)
	dataplaneProto, err := p.dataplaneFor(ctx, pod, services, others)
	if err != nil {
		return err
	}
        curSpec := datapane.GetSpec()
        if proto.DeepEqual(curSpec, datapalneProto) {
             return UnchangedError // Or something different.
        }
        dataplane.SetSpec(dataplaneProto)
	return nil
```

lukidzi: Yes, I feel that handling this in `PodToDataplane` might be simpler, but it may not address all the cases. I've noticed the issue affecting `Dataplane`, `ZoneIngress`, and `ZoneEgress`. However, given how the data is stored and how the checks are performed, other components like `MeshService` and `MeshGateway` might also be impacted.

lahabana: I don't think so as MeshService doesn't use protobufs. MeshGateway might be have the problem yes

lukidzi: Ok, so I will update a previous PR with adjustment to validate if the proto has changed and check if the same is for gateway

lukidzi: Closing this in favor of [#11327](https://github.com/kumahq/kuma/pull/11327).

PR #11336, opened 2024-09-09
refactor(kuma-cp): use `ResolvedBackendRef.LegacyBackendRef` only when required

lobkovilya:
https://github.com/kumahq/kuma/pull/11203 introduced ResolvedBackendRef with ResourceIdentifier:
```go
type ResolvedBackendRef struct {
	LegacyBackendRef *common_api.BackendRef
	Resource         *TypedResourceIdentifier
}
```
but it was always setting `LegacyBackendRef` (for old and new resource kinds) to not overcomplicate the PR.

This PR changes the code so `LegacyBackendRef` is set only when BackendRef is pointing to legacy service tag services and `RealResourceBackendRef` when referencing real resources:

```go
type ResolvedBackendRef struct {
	// Ref is either LegacyBackendRef or RealResourceBackendRef
	Ref IsResolvedBackendRef
}
```

The PR is a refactoring and doesn't change the behaviour.

### Checklist prior to review


pkg/core/resources/model/resource.go
```diff
@@ -777,7 +779,9 @@ type TypedResourceIdentifier struct {
 
 type ResolvedBackendRef struct {
 	LegacyBackendRef *common_api.BackendRef
```
michaelbeaumont, line 781: If these are mutually exclusively set, is it possible to have a private interface like 

```
type IsResolvedBackendRef interface {
  isResolvedBackendRef()
}
```

and then have

```
type ResolvedBackendRef struct {
  Weight uint
  Ref IsResolvedBackendRef
}

...

var ref ResolvedBackendRef

switch t := ref.Ref {
case *LegacyBackendRef:
case *RealResourceBackendRef:
}
```

so that we always have exactly one of these?

michaelbeaumont: or otherwise make it clear via comments what it means when one thing is set/not set and whether both can be set

lobkovilya: I did like you suggested with a single `Ref` field for these mutually exclusive values. It's actually nice that now we have several functions like `SniForBackendRef` and `ServiceTagIdentities` that accept only `RealResourceBackendRef` and don't require checking for legacy stuff.

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go
```diff
@@ -32,10 +32,13 @@ func generateFromService(
 	rules rules.ToRules,
 	svc meshroute_xds.DestinationService,
 ) (*core_xds.ResourceSet, error) {
-	tags := svc.BackendRef.Tags
-	listenerBuilder := envoy_listeners.NewOutboundListenerBuilder(proxy.APIVersion, svc.OutboundInterface.DataplaneIP, svc.OutboundInterface.DataplanePort, core_xds.SocketAddressProtocolTCP).
+	var tags envoy_tags.Tags
+	if svc.Outbound.LegacyOutbound != nil {
```
michaelbeaumont, line 36: the one thing that scares me about this is that one has to remember to always check LegacyOutbound for tags. Isn't `svc.Outbound.Tags` or similar safer?

Or like `envoy_listeners.ConfigureFromOutbound(outbound xds_types.Outbound)` or `xds_types.Outbound.Configure(listener envoy_listener.Listener)` ? The tags thing is so "leaky" here.

lobkovilya: Updated the PR with `TagsOrNil` method

michaelbeaumont (approved):

PR #11348, opened 2024-09-09
fix(MeshService): add kuma.io/display-name label to generated MeshService on universal

michaelbeaumont:
These aren't added otherwise. `ComputeLabels` isn't called when using the store directly.

### Checklist prior to review

> Changelog: skip


slonka (commented):
Can you add a short description why wy need to do that?

michaelbeaumont:
> Can you add a short description why wy need to do that?

Because it doesn't happen otherwise

lobkovilya:
It's similar to https://github.com/kumahq/kuma/pull/11334 but now about `kuma.io/display-name` label right? Do we need it for MES and MMZS?

lukidzi (approved):

michaelbeaumont:
> Do we need it for MES and MMZS?

I don't know, how are they being created? the problem here was that we're creating them directly with the store and the labels aren't added in that case

lobkovilya:
> It's similar to https://github.com/kumahq/kuma/pull/11334 but now about kuma.io/display-name label right? Do we need it for MES and MMZS?

nvm it's just for autogenerated

PR #11380, opened 2024-09-11
Revert "refactor(xds): reduce size of meshroute.DestinationService type"

lobkovilya:
Reverts kumahq/kuma#11375

The idea is to get rid of `DestinationService` altogether and use `xds_types.Outbound` instead. https://github.com/kumahq/kuma/pull/11336 already does a step in this direction by removing `BackendRef` field.


michaelbeaumont:
Will `xds_types.Outbound` also be reduced? The point of the PR was that we didn't need all of `xds_types.Outbound`.

lobkovilya:
What fields from `xds_types.Outbound` we don't need?

PR #11385, opened 2024-09-11
feat(meshservice): mitigate and handle resource conflicts

### LINKED ISSUE #10511: Avoid conflicts on `MeshService` updates
### Description

- Some kind of locking between new components that update MeshService?
- Batching of updates?
- What issues do conflicts cause?

jakubdyszkiewicz:
### Checklist prior to review

Fix #10511

I think the best strategy for now is to
* Introduce jitter so not all components start at once
* Log on info, not on ERROR

Lock might be too aggressive and batching updates are complicated to implement for now.
I noticed that in E2E jitter already solves a lot of such issues. We can revisit other strategies if needed

### DIFF
--- a/pkg/core/resources/apis/core/vip/allocator.go
+++ b/pkg/core/resources/apis/core/vip/allocator.go
@@ -16,6 +16,7 @@ import (
 	"github.com/kumahq/kuma/pkg/core/runtime/component"
 	"github.com/kumahq/kuma/pkg/core/user"
 	core_metrics "github.com/kumahq/kuma/pkg/metrics"
+	util_time "github.com/kumahq/kuma/pkg/util/time"
 )
 
 type ResourceHoldingVIPs interface {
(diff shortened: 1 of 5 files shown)


bartsmykla (approved):

Icarus9913 (approved):

Icarus9913:
Further thought:
Maybe we could also use the kubernetes `Retry` function to handle with our `update` operation in the future?

https://github.com/kubernetes/apimachinery/blob/4e174c5e3655ac47294239904b796671c5d9dfd6/pkg/util/wait/backoff.go#L459

jakubdyszkiewicz:
@Icarus9913 we already have `.Upsert()` function with retry backoff and jitter. The problem is that you need to first retrieve the resource on retry so writing this logic is a bit meh when you don't want to pay the cost upfront for retrieving a resource because you already fetched the list

PR #11397, opened 2024-09-12
test(framework): print original stdout from cURL when collecting failures form test requests

jijiechen:
More information, like `errormsg` is available in cURL generated JSON, so let's print it to help diagnose isues.

### Checklist prior to review

  - N/A
  - Manually tested
  - Don't

### DIFF
--- a/test/e2e_env/multizone/reachablebackends/reachablebackends_meshservices.go
+++ b/test/e2e_env/multizone/reachablebackends/reachablebackends_meshservices.go
@@ -137,10 +137,11 @@ spec:
 	It("should not be able to connect to cross-zone MeshService if reachable backends isn't set", func() {
 		Consistently(func(g Gomega) {
 			// when
-			resp, err := client.CollectFailure(
+			resp, stdout, err := client.CollectFailureRaw(
 				multizone.KubeZone1, "client-without-reachable", "other-zone-test-server.mesh-service-reachable-backends.svc.kuma-2.mesh.local",
 				client.FromKubernetesPod(namespace, "client-without-reachable"),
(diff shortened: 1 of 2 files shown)


test/e2e_env/multizone/reachablebackends/reachablebackends_meshservices.go
```diff
@@ -137,10 +137,11 @@ spec:
 	It("should not be able to connect to cross-zone MeshService if reachable backends isn't set", func() {
 		Consistently(func(g Gomega) {
 			// when
-			resp, err := client.CollectFailure(
+			resp, stdout, err := client.CollectFailureRaw(
 				multizone.KubeZone1, "client-without-reachable", "other-zone-test-server.mesh-service-reachable-backends.svc.kuma-2.mesh.local",
 				client.FromKubernetesPod(namespace, "client-without-reachable"),
 			)
+			Logf("cURL stdout: %s", stdout)
```
jakubdyszkiewicz, line 144: I'd rather have something like `client.CollectFailure(..., client.Verbose())`
and then inside CollectFailure Log the stdout if we have this flag.

Or maybe even log it always? Why do we want to have it for this specific test only?

jijiechen: Because we want to know the verbose error logged out rather than just get the JSON result.
When we were debugging the flakiness, we can't see the underlying error using existing method.

jijiechen:
Closing this as the original flaky had been fixed by https://github.com/kumahq/kuma/pull/11398

PR #11409, opened 2024-09-13
refactor(k8s): don't handle resource change detection based on error

lukidzi:
### Checklist prior to review

Refactor of [#11327](https://github.com/kumahq/kuma/pull/11327). We no longer need to handle this with an error because when the spec is not set, it gets compared with the same spec in `CreateOrUpdate` and doesn't update a resource.


pkg/plugins/runtime/k8s/controllers/pod_converter.go
```diff
@@ -44,6 +44,7 @@ func (p *PodConverter) PodToDataplane(
 	services []*kube_core.Service,
 	others []*mesh_k8s.Dataplane,
 ) error {
+	logger := converterLog.WithValues("Dataplne.name", dataplane.Name, "Pod.name", pod.Name)
```
lahabana, line 47:
```suggestion
	logger := converterLog.WithValues("Dataplane.name", dataplane.Name, "Pod.name", pod.Name)
```

lahabana: It seems unusual to me to put `.` in log keys. @jakubdyszkiewicz opinions?

lukidzi: I wanted to preserve the same format as for ZoneIngress and ZoneEgress https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/controllers/pod_converter.go#L66

Icarus9913 (approved):

pkg/plugins/runtime/k8s/controllers/pod_converter.go
jakubdyszkiewicz: yeah, I'd try to avoid `.`

jakubdyszkiewicz (approved):

PR #11412, opened 2024-09-13
fix(kuma-cp): replace maps.Clone until upstream fix

slonka:
we suspect because of https://github.com/golang/go/issues/69110 

can be reverted when fixed

not 100% sure it will work because that's just the same code that it's in go but I think this directive: https://github.com/golang/go/blob/8a4b439ee6e0f5d41cb6ae30146b901f3c0ad4b7/src/maps/maps.go#L45 makes it so that the linker uses some sort of optimised version here https://github.com/golang/go/blob/8a4b439ee6e0f5d41cb6ae30146b901f3c0ad4b7/src/runtime/map_swiss.go#L1504 which actually has the bug. Otherwise if an NPE can be triggered just in a simple loop then 🤷 - we need to wait for the fix.

### Checklist prior to review


jijiechen (approved):

michaelbeaumont (changes requested):
Hold on

Issue #11410, opened 2024-09-13
Inspect API returns wrong response for DPPs affected by consumer policy


lobkovilya:
### What happened?

It takes into account only top-level TargetRef, but should also filter by namespace and zone.

PR #11425, opened 2024-09-16
fix(inspect-api): added check if dpp is affected by zone policy

lukidzi:
### Checklist prior to review

* Renamed the function from ddpSelectedByNamespace to dppSelectedByNamespace.
* Added a check to verify if the resource has the origin zone, and if so, ensure the zone of the DPP matches the zone of the policy.
* Added corresponding tests.
 
- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/11410


jakubdyszkiewicz (approved):

pkg/plugins/policies/core/matchers/dataplane.go
```diff
@@ -185,6 +188,20 @@ func ddpSelectedByNamespace(meta core_model.ResourceMeta, dpp *core_mesh.Datapla
 	}
 }
 
+func dppSelectedByZone(meta core_model.ResourceMeta, dpp *core_mesh.DataplaneResource) bool {
+	switch meta.GetLabels()[mesh_proto.PolicyRoleLabel] {
+	case string(mesh_proto.ConsumerPolicyRole), string(mesh_proto.WorkloadOwnerPolicyRole):
```
lobkovilya, line 193: `system` policies that originate on zone should also be prefiltered by the `zone`. Probably it's easier to check for `producer` here and do the filter for all other types:
```go
switch core_model.PolicyRole(meta) {
case mesh_proto.ProducerPolicyRole:
    return true
default:
    // check the zone label
    // ...
}
```

I think it's better to use `core_model.PolicyRole` because policies created before the upgrade don't have `kuma.io/policy-role` label.

pkg/plugins/policies/core/matchers/testdata/matchedpolicies/torules/04.golden.yaml
```diff
@@ -3,21 +3,14 @@ Rules:
 - BackendRefOriginIndex: {}
   Conf:
     http:
-      requestTimeout: 3s
+      requestTimeout: 1s
```
lobkovilya, line 6: That's weird... I'm not sure if it's correct that it changed

lukidzi: https://github.com/kumahq/kuma/blob/master/pkg/plugins/policies/core/matchers/testdata/matchedpolicies/torules/04.policies.yaml#L17

lukidzi: Should the `kuma.io/zone` be in labels or `inbound.[].tags` ?

lukidzi: seems like on both, so should be ok

lukidzi: It seems that the change broke something with the test on universal

lukidzi: I think it's because of this https://github.com/kumahq/kuma/pull/11449. Could you take a look if it makes sense?

Issue #11424, opened 2024-09-16
Check MeshTLS with GW (builtin and delegated)


slonka:
### Description

I don't think we tested this. We might also add an E2E test for it.

PR #11437, opened 2024-09-16
feat(meshtls): added builtin gateway implementation

lukidzi:
### Checklist prior to review

While testing, I noticed that the built-in gateway isn't supported. This PR adds support for configuring Gateway clusters with TLS versions and ciphers.

d8b489d4e304bfcc4e9dac9bb3ccadbc02dc41f0 implementation 
fdc419a824a97961dfbb6843fbc1ceea600f2a2b unit test + golden files 
4b2c978538d052c1b62f3a3da59868bbe9e30123 e2e test

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/kumahq/kuma/issues/11424

### DIFF
--- a/test/e2e_env/kubernetes/gateway/gateway.go
+++ b/test/e2e_env/kubernetes/gateway/gateway.go
@@ -1039,6 +1039,101 @@ spec:
 		})
 	})
 
+	Context("MeshTLS", func() {
+		httpRoute := fmt.Sprintf(`
+apiVersion: kuma.io/v1alpha1
+kind: MeshHTTPRoute
+metadata:
+  name: http-route-1-mtls
+  namespace: %s
+  labels:
+    kuma.io/mesh: %s
+spec:
+  targetRef:
+    kind: MeshGateway
+    name: simple-gateway
+  to:
+  - targetRef:
+      kind: Mesh
+    rules:
+    - matches:
+      - path:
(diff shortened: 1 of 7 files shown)


test/e2e_env/kubernetes/gateway/gateway.go
```diff
@@ -1039,6 +1039,101 @@ spec:
...
+			// still can access test-server from the gateway
+			Eventually(func(g Gomega) {
+				response, err := client.CollectEchoResponse(
+					kubernetes.Cluster, "demo-client",
+					"http://simple-gateway.simple-gateway:8080/test",
+					client.WithHeader("host", "example.kuma.io"),
+					client.FromKubernetesPod(clientNamespace, "demo-client"),
+				)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(response.Received.Headers["Host"]).To(HaveLen(1))
```
slonka, line 1131: should we also check stats that there is TLS13 handshake?

slonka (approved):
One question, otherwise seems good

test/e2e_env/kubernetes/gateway/gateway.go
lukidzi: We could but it's not required. Without the implementation test was failing

PR #11439, opened 2024-09-16
fix(policy-gen): check lowercase name for suffix when generating plural names

lukidzi:
### Checklist prior to review

The policy generation script was comparing an uppercase name with a lowercase ending, which caused plural names to not generate correctly. I’ve updated it to lowercase the resource name before comparison.


jakubdyszkiewicz (approved):

pkg/plugins/policies/meshtls/api/v1alpha1/zz_generated.resource.go
```diff
@@ -145,7 +145,7 @@ var MeshTLSResourceTypeDescriptor = model.ResourceTypeDescriptor{
 	IsPolicy:                     true,
 	IsExperimental:               false,
 	SingularDisplayName:          "Mesh T L S",
-	PluralDisplayName:            "Mesh T L Ss",
+	PluralDisplayName:            "Mesh T L Ses",
```
lahabana, line 148: The way this is spaced out also seems like a bug to me no?

lukidzi: It looks strange, I am not sure why.

lukidzi: Ok seems related to pkg/core/resources/model/display_name.go
```
if unicode.IsUpper(c) && i != 0 {
   displayName += " "
}
```

PR #11445, opened 2024-09-17
feat(meshexternalservice): do not require zone egress setting on mesh

### LINKED ISSUE #11443: MeshExternalService uses egress without enabling it
### Description

MeshExternalService requires egress setting on the Mesh, but this also enables cross zone internal services traffic through extress. MES does not with egress anyways so we should just do best effort to enable the traffic if the egress is just present.

jakubdyszkiewicz:
### Checklist prior to review

Fix #11443

We now do not require egress option on the mesh to pass the traffic of mesh external service through egress.


pkg/xds/generator/egress/external_services_generator.go
```diff
@@ -146,14 +145,11 @@ func (*ExternalServicesGenerator) generateCDS(
...
 func (*ExternalServicesGenerator) buildServices(
 	endpointMap core_xds.EndpointMap,
-	localZone string,
-	meshResources *core_xds.MeshResources,
 ) map[string]bool {
 	services := map[string]bool{}
 
 	for serviceName, endpoints := range endpointMap {
-		if len(endpoints) > 0 && endpoints[0].IsExternalService() &&
-			(!meshResources.Mesh.ZoneEgressEnabled() || endpoints[0].IsReachableFromZone(localZone)) {
```
jakubdyszkiewicz, line 156: This part
```
!meshResources.Mesh.ZoneEgressEnabled() || endpoints[0].IsReachableFromZone(localZone)
```
was
1) wrong, because it should be `meshResources.Mesh.ZoneEgressEnabled()`
but also
2) unnecessary, because endpoints in endpoint map are prefiltered with `endpoints[0].IsReachableFromZone(localZone)` anyways, so this part was always true for external services. We build this endpoint map per single egress.

pkg/xds/generator/egress/internal_services_generator.go
```diff
@@ -23,6 +23,11 @@ func (g *InternalServicesGenerator) Generate(
 	meshResources *core_xds.MeshResources,
 ) (*core_xds.ResourceSet, error) {
 	resources := core_xds.NewResourceSet()
+
+	if !meshResources.Mesh.ZoneEgressEnabled() {
+		return resources, nil
```
jakubdyszkiewicz, line 28: because in egress proxy builder I had to change `ZoneEgressEnabled` to just `MTLSEnabled` I had to add a condition here for only internal services so that we won't waste CP resources on reconfiguring egress when it's not used

pkg/xds/generator/egress/testdata/input/02.internalservice-only.yaml
```diff
@@ -5,6 +5,8 @@ mtls:
   backends:
   - name: ca-1
     type: builtin
+routing:
+  zoneEgress: true
```
jakubdyszkiewicz, line 9: because of wrong condition https://github.com/kumahq/kuma/pull/11445/files#r1763279132
I had to add it now to the test

It's not necessarily a bug in the real product. It seems that we were configuring configure egress with external services regardless of zoneEgress setting, which is fine. A bit excessive but should not have any significant impact

pkg/xds/topology/outbound_test.go
```diff
@@ -1421,7 +1421,7 @@ var _ = Describe("TrafficRoute", func() {
 						},
 					},
 				},
-				mesh: defaultMeshWithMTLSAndZoneEgress,
+				mesh: defaultMeshWithMTLS,
```
jakubdyszkiewicz, line 1424: This is so we test that we add MES clusters even without zone egress setting now

michaelbeaumont (approved):

PR #11449, opened 2024-09-17
fix(kuma-cp): add labels to dataplane object on universal

lukidzi:
### Checklist prior to review

While investigating the failure of the universal test (https://github.com/kumahq/kuma/pull/11425), I discovered that labels were not being set for dataplanes on the universal environment. In the universal setup, users can send the Dataplane object in the metadata, which is later processed by the control plane. On Kubernetes, we call ComputeLabels in the defaulter, which sets the labels, but this was not happening for the universal environment. In the universal setup, tags were only being set on the inbounds and not on the entire dataplane object. I've added a flow to apply tags to the whole dataplane.

### DIFF
--- a/pkg/core/resources/model/resource.go
+++ b/pkg/core/resources/model/resource.go
@@ -457,9 +457,9 @@ func resourceOrigin(labels map[string]string) (mesh_proto.ResourceOrigin, bool) 
 }
 
 func ComputeLabels(r Resource, mode config_core.CpMode, isK8s bool, systemNamespace string, localZone string) (map[string]string, error) {
-	labels := r.GetMeta().GetLabels()
-	if len(labels) == 0 {
-		labels = map[string]string{}
+	labels := map[string]string{}
+	if r.GetMeta() != nil && len(r.GetMeta().GetLabels()) > 0 {
+		labels = r.GetMeta().GetLabels()
 	}
 
 	setIfNotExist := func(k, v string) {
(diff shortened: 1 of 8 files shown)


pkg/core/resources/model/resource.go
```diff
@@ -469,8 +469,10 @@ func ComputeLabels(r Resource, mode config_core.CpMode, isK8s bool, systemNamesp
 	}
 
 	getMeshOrDefault := func() string {
-		if mesh := r.GetMeta().GetMesh(); mesh != "" {
-			return mesh
+		if r.GetMeta() != nil {
```
lobkovilya, line 472: I think it's a bit error prone that we have to check `r.GetMeta()` to not be nil, it's easy to forget when adding new labels to `ComputeLabels`. 

For example, it's kind of just a coincidence we don't use use `dataplaneManager` on Kubernetes, otherwise we wouldn't get `k8s.kuma.io/namespace` (because it requires to check `r.GetMeta().GetNameExtensions()`)

Probably we should change `ComputeLabel` signature to not require `ResourceMeta` at all, i.e 
```go
func ComputeLabels(rd ResourceTypeDescriptor, namespace, mesh string, mode config_core.CpMode, isK8s bool, systemNamespace string, localZone string)
```

It's not ideal it's that long, but we can refactor this in the future.

WDYT?

Icarus9913 (approved):

lobkovilya (approved):

Issue #11426, opened 2024-09-16
Check new Global with old Zone to handle top-level targetRef syntactic sugar properly


lobkovilya:
### What happened?

1. Create policy on Global that doesn't have top-level targetRef
2. Let the policy to be synced to old zone (2.8.x)

Automaat:
Zone will send NACK on this policy and we will see log: 
```
kds-delta-zone    received resource is invalid, sending NACK    {"kds-version": "v2", "err": "spec.targetRef.kind: must be defined"}
```

Automaat:
> Does the issue fix itself when Zone CP is updated?
yes, it is synced properly after upgrading zone

> Does the KDS sync work for other valid policies of the same type?
yes, after applying another meshtimeout with explicit topLevel targetRef, it is synced correctly

One thing with default topLevel targetRef is that when running `kubectl get meshtimeouts -n kuma-system` we have missing targetRef kind in output:

![Image](https://github.com/user-attachments/assets/3a83d4a3-af35-46bc-a88c-fa2eae17c8f2)

lobkovilya:
I tried:
1. Create 2 policies on Global with and without top-level targetRef
2. Connect Zone CP 2.8.3
3. None of the policies is synced to Zone CP (even the valid one)

Moreover, when I deleted the policy without top-level targetRef from Global, the valid policy is still not synced until I restart the Zone CP. 

I think we should put this syntactic sugar behind the KDS feature flag. If we have an option to avoid NACKs I'd rather go for it.

Automaat:
I guess the KDS behaviour is not consistent, I had the same setup as you, and it worked properly for me (only correct policy was synced)

lobkovilya:
It probably depends on the order, because the Sync() just iterates over the list and tries to apply resource one by one until it sees the error.

PR #11457, opened 2024-09-18
fix(kuma-cp): do not sync policies with empty topLevel targetRef to zones that does not support it

Automaat:
Fixes: #11426

Since `Policy` interface exposes only function `GetTargetRef` which returns default targetRef, we don't have an easy way of checking if policy had targetRef set. To deal with this, we can introduce `UsesSyntacticSugar` to make sure we are not syncing policies not supported by older version. This can be easily removed in two versions

### Checklist prior to review


jakubdyszkiewicz (approved):
I'm slightly worried this bool will show in the API schema etc. but doesn't seem like it 🤔

Automaat:
> I'm slightly worried this bool will show in the API schema etc. but doesn't seem like it 🤔

Non of the golden files or schema files were updated automatically, so it should not, also there is `json:"-"` annotation, which should ignore this field

Issue #11456, opened 2024-09-18
In MeshHTTPRoute with MeshMultizoneService XDS config is invalid


lahabana:
### What happened?

on version: 0.0.0-preview.vacb7899a5

With the following MeshHTTPRoute:

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshHTTPRoute
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"kuma.io/v1alpha1","kind":"MeshHTTPRoute","metadata":{"annotations":{},"labels":{"kuma.io/mesh":"default","kuma.io/origin":"zone"},"name":"demo-app","namespace":"kong-mesh-system"},"spec":{"targetRef":{"kind":"MeshGateway","name":"demo-app"},"to":[{"rules":[{"default":{"backendRefs":[{"kind":"MeshService","name":"demo-app_kuma-demo_svc_5000"}]},"matches":[{"path":{"type":"PathPrefix","value":"/"}}]}],"targetRef":{"kind":"Mesh"}}]}}
    kuma.io/display-name: demo-app
  creationTimestamp: "2024-09-18T12:42:36Z"
  generation: 3
  labels:
    k8s.kuma.io/namespace: kong-mesh-system
    kuma.io/env: kubernetes
    kuma.io/mesh: default
    kuma.io/origin: zone
    kuma.io/policy-role: system
    kuma.io/zone: west
  name: demo-app
  namespace: kong-mesh-system
  ownerReferences:
  - apiVersion: kuma.io/v1alpha1
    kind: Mesh
    name: default
    uid: c048cf62-1adb-4047-b6f4-f89e02aee21b
  resourceVersion: "61213"
  uid: 8fac6858-66ea-4b56-9f56-c24fc4ca9a57
spec:
  targetRef:
    kind: MeshGateway
    name: demo-app
  to:
  - rules:
    - default:
        backendRefs:
        - kind: MeshMultiZoneService
          labels:
            kuma.io/display-name: demo-app-everywhere
          weight: 1
      matches:
      - path:
          type: PathPrefix
          value: /
    targetRef:
      kind: Mesh
```

And meshMZService:

```yaml
type: MeshMultiZoneService
mesh: default
name: demo-app-everywhere
spec:
  selector:
    meshService:
      matchLabels:
        kuma.io/display-name: demo-app
  ports:
  - name: '5000'
    port: 5000
```

```
[...]

PR #11461, opened 2024-09-18
fix(MeshGateway): handle unresolved real backendRefs

michaelbeaumont:
### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- #11456 

### DIFF
--- a/pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin_test.go
+++ b/pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin_test.go
@@ -152,6 +152,7 @@ var _ = Describe("MeshTLS", func() {
 				Items: []*core_mesh.MeshGatewayResource{samples.GatewayResource()},
 			}
 			backendRefOriginIndex := map[common_api.MatchesHash]int{}
+			var backendRef common_api.BackendRef
 			if given.meshService {
 				backendRefOriginIndex = map[common_api.MatchesHash]int{
 					meshhttproute_api.HashMatches([]meshhttproute_api.Match{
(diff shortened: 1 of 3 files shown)


pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin_test.go
```diff
@@ -217,11 +228,7 @@ var _ = Describe("MeshTLS", func() {
 														},
 													}},
 													Default: meshhttproute_api.RuleConf{
-														BackendRefs: &[]common_api.BackendRef{{
-															TargetRef: builders.TargetRefMeshService("backend", "", "test-port"),
```
michaelbeaumont, line 221: a backendRef can't have a sectionName and it needs to _not_ have a port to refer to a kuma.io/service

michaelbeaumont: @lukidzi does the test still make sense as is? Probably it should have a Dataplane with inbound `kuma.io/service: backend` to really be complete but since we don't check endpoints here it seems to be ok without

test/e2e_env/multizone/meshservice/connectivity.go
```diff
@@ -264,6 +275,29 @@ spec:
...
+					client.FromKubernetesPod(clientNamespace, "demo-client"),
+				)
+
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(response.ResponseCode).To(Equal(given.responseCode))
+			}, "30s", "1s").Should(Succeed())
+		},
+		Entry("no matching backend should return 500", failureCase{
+			path:         "no-matching-backend",
+			responseCode: 500,
```
slonka, line 297: why 500? shouldn't it be 404? 🤔 or at least 503...

slonka (commented):
One Q

test/e2e_env/multizone/meshservice/connectivity.go
michaelbeaumont: That's existing behavior that gateway API prescribes https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPRouteRule

jakubdyszkiewicz (approved):

PR #11467, opened 2024-09-19
test(e2e): improve traffic route test and delete ES app

lahabana:
- ES apps in universal were not torn down
- Traffic route tests rely on a more predictible patterns where we first unsure that policy propagation happened and then check policy effect

### Checklist prior to review


test/e2e_env/universal/trafficroute/traffic_route.go
```diff
@@ -677,40 +687,41 @@ destinations:
...
-				),
-			)
+			// Check and retry until the config got propagated to the client
+			Eventually(func() ([]server_types.EchoResponse, error) {
+				return client.CollectResponses(universal.Cluster, "demo-client", "test-server.mesh/i-am-here")
+			}, "1m", "500ms").MustPassRepeatedly(5).Should(HaveEach(HaveField("Received.Headers", HaveKeyWithValue("X-I-Am-Here", []string{"route-internal-external"}))))
+
+			Expect(client.CollectResponsesByInstance(universal.Cluster, "demo-client", "test-server.mesh", client.WithNumberOfRequests(100))).
+				Should(And(
+					HaveKeyWithValue("echo-v1", BeNumerically("~", 50, 10)),
```
michaelbeaumont, line 722: I dunno, is `10` enough? Why not `20`, `30`? Like why should there be ever an error where the weights are distorted by say 50%? The only error I cna imagine would be one where one of the weights is set to `0`

lahabana: We could do just HaveKey but why not do the check too? Envoy is still supposed to respect some level of Weighted routing

lukidzi (approved):

PR #11474, opened 2024-09-19
fix(kuma-cp): allow specifying namespace when targeting MeshExternalService in policies

Automaat:
We are not always configuring policies in a system namespace when MES is created. We want users to be able to select MES in policies on their namespaces, but in order to do it they need to be able to specify namespace in `to.targetRef` section

### Checklist prior to review


slonka (commented):
Aren't we missing tests? Also can we get some description of this?

Automaat:
> Aren't we missing tests? Also can we get some description of this?

@slonka added description and test

slonka (approved):

Issue #11455, opened 2024-09-18
origins on gateway seems to not work well with multiple zones


lahabana:
### What happened?

I have the demo app deployed in 2 zones (with gateway). This shows this which just seems invalid as each route only exists in one zone.

<img width="1630" alt="image" src="https://github.com/user-attachments/assets/42359d04-cca0-46b0-bc00-87724b0be98e">

lahabana:
Actually this even creates invalid config (here the rules should show a MeshMZService)

PR #11504, opened 2024-09-20
fix(kuma-cp): properly match policies to gateway when calling _rules endpoint

Automaat:
When we call inspect API `_rules` endpoint for MeshGateway resource, we create fake dpp without meta that is only used to pick proper gateway for matching. With this change, we are adding gateway meta to it, and we are filtering gateways by zone to pick only gateways from the same zone. Moreover, we are now matching policies by zone not only for dpp but also for MeshGateway

Fixes: #11455

### Checklist prior to review


Automaat:
I think there might be a bigger issue here. When we are [selecting gateway](https://github.com/kumahq/kuma/blob/8340f846aa0eaa491c14e13b8c948f2cb8c51ae0/pkg/plugins/policies/core/matchers/dataplane.go#L51) we are looking only at selector, and when we run this on global we ignore the fact that we could have synced gateways from different zones with the same selector. I think that this only affects inspect api on global, but I am not sure

pkg/plugins/policies/core/matchers/dataplane_test.go
```diff
@@ -206,4 +206,21 @@ var _ = Describe("MatchedPolicies", func() {
...
+
+			policies, err := matchers.MatchedPolicies(meshaccesslog_api.MeshAccessLogType, dpp, resources)
+			Expect(err).ToNot(HaveOccurred())
+
+			bytes, err := yaml.Marshal(policies.GatewayRules)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(bytes).To(test_matchers.MatchGoldenYAML(given.goldenFile))
+		},
+		generateTableEntries(filepath.Join("testdata", "matchedpolicies", "dpp_without_meta")),
+	)
```
lahabana, line 225: I think you don't need to add this new test case. The point of the generateTableEntries is that you can simply add tests cases by sending: input, policies --> golden.

Automaat: yes, but I needed to clean dpp meta, and you cannot create dpp without meta from file

pkg/api-server/testdata/resources/inspect/dataplanes/_rules/meshgateway_from_global.input.yaml
```diff
@@ -0,0 +1,77 @@
+#/meshes/default/meshgateways/the-gateway/_rules 200
```
lahabana, line 1: Looking at this path this test is in the wrong folder. It should likely be there no: https://github.com/kumahq/kuma/tree/513a70e65cae9f47b69435b4594a02029fdc42c2/pkg/api-server/testdata/resources/inspect/meshgateways ?

pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/03.policies.yaml
```diff
@@ -0,0 +1,90 @@
...
+     kind: Mesh
+    default:
+      backends:
+        - type: File
+          file:
+            path: /servicesubset
+---
+type: MeshAccessLog
+mesh: mesh-1
+name: gateway
```
lahabana, line 43:
```suggestion
name: in-zone-2
```

Makes it easier to check the golden file because you are looking for zone-2 in something that runs in zone-1

Also the tests can have any name. So naming it something like "policies-in-other-zone" might be better than `03`

pkg/api-server/testdata/resources/inspect/dataplanes/_rules/meshgateway_from_global.input.yaml
Automaat: right, moved

pkg/api-server/testdata/resources/inspect/dataplanes/_rules/meshgateway_endpoint.input.yaml
```diff
@@ -0,0 +1,52 @@
+#/meshes/default/meshgateways/the-gateway/_rules 200
```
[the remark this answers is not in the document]
Automaat: moved

pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/03.policies.yaml
Automaat: renamed

jakubdyszkiewicz (approved):

PR #11506, opened 2024-09-20
test(framework): dont print CP logs to stdout on Universal as they can be too long

jijiechen:
When the debug level logs are enbled,too many log entries make the GitHub log view truncate.
We now export the CP logs into artifacts, so don't need to print to stdout any more.

### Checklist prior to review

  - N/A
  - Manually tested
  - No
  - No

> Changelog: skip

### DIFF
--- a/test/framework/envs/universal/env.go
+++ b/test/framework/envs/universal/env.go
@@ -58,18 +58,6 @@ func RestoreState(bytes []byte) {
 	Cluster.SetCp(cp)
 }
 
-func PrintCPLogsOnFailure(report ginkgo.Report) {
-	if !report.SuiteSucceeded {
-		logs, err := Cluster.GetKumaCPLogs()
-		if err != nil {
-			framework.Logf("could not retrieve cp logs")
-		} else {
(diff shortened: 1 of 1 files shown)


test/framework/envs/universal/env.go
```diff
@@ -89,5 +77,8 @@ func AfterSuite(report ginkgo.Report) {
 	if framework.Config.CleanupLogsOnSuccess {
 		universal_logs.CleanupIfSuccess(framework.Config.UniversalE2ELogsPath, report)
 	}
-	PrintCPLogsOnFailure(report)
+	if !report.SuiteSucceeded {
+		framework.Logf("Please see full CP logs by downloading the debug artifacts")
```
lahabana, line 81: I guess we can't create a link?

jijiechen: It's not easy to create the link without importing the GH SDK, or by referencing some environment variables which indedicate the run id and repo name, etc.

Icarus9913 (approved):

PR #11508, opened 2024-09-22
fix(api-server): skip display-name label for service insight

### LINKED ISSUE #9729: Skip display name ServiceInsight
### Description

We should not show display name label in ServiceInsight

Icarus9913:
This API would list the SeviceInsight resources with k8s mode, in which the common method [GetLabels](https://github.com/kumahq/kuma/blob/e21c751d782bc06eb7e8c779dacf71097bfe5732/pkg/plugins/resources/k8s/store.go#L299) would be called and supplement the `display-name` for the `Meta`.

So, let's just add a mapper function to remove that label in the API-Server component.

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues -- close #9729 

### DIFF
--- a/pkg/api-server/service_insight_endpoints.go
+++ b/pkg/api-server/service_insight_endpoints.go
@@ -108,7 +108,6 @@ func (s *serviceInsightEndpoints) listResources(request *restful.Request, respon
 		_, exists := filterMap[service.ServiceType]
 		return exists
 	})
-
 	restList := rest.ResourceList{
 		Total: uint32(len(items)),
 		Items: items,
--- a/pkg/api-server/service_insight_endpoints.go
+++ b/pkg/api-server/service_insight_endpoints.go
@@ -2,6 +2,7 @@ package api_server
 
 import (
 	"fmt"
+	"maps"
 	"sort"
 	"strconv"
 	"strings"
@@ -55,6 +56,10 @@ func (s *serviceInsightEndpoints) findResource(request *restful.Request, respons
 		res := out.(*rest_unversioned.Resource)
 		res.Meta.Name = service
 		res.Spec = stat
+
+		mapperFn := removeDisplayNameLabel()
+		mapperFn(res)
+
 		if err := response.WriteAsJson(res); err != nil {
 			core.Log.Error(err, "Could not write the response")
 		}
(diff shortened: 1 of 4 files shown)


pkg/api-server/service_insight_endpoints.go
```diff
@@ -135,7 +144,12 @@ func (s *serviceInsightEndpoints) fillStaticInfo(name string, stat *v1alpha1.Ser
 // 2) Mesh+Name is a key on Universal, but not on Kubernetes, so if there are two services of the same name in different Meshes we would have problems with naming.
 // From the API perspective it's better to provide ServiceInsight per Service, not per Mesh.
 // For this reason, this method expand the one ServiceInsight resource for the mesh to resource per service
-func (s *serviceInsightEndpoints) expandInsights(serviceInsightList *mesh.ServiceInsightResourceList, nameContains string, filterFn func(service *v1alpha1.ServiceInsight_Service) bool) []rest.Resource {
+func (s *serviceInsightEndpoints) expandInsights(
+	serviceInsightList *mesh.ServiceInsightResourceList,
+	nameContains string,
+	filterFn func(service *v1alpha1.ServiceInsight_Service) bool,
+	mapperFn unversionedResourceMapper,
```
jakubdyszkiewicz, line 151: why do we need a mapper function here? Do we expect other implementations? If not, can't we just call removeDisplayNameLabel directly?

jakubdyszkiewicz (commented):
please add a test

pkg/api-server/service_insight_endpoints.go
Icarus9913: I was just wondering if we need other mapper implementations in the further, tha's why I write a `mapperFn` here. Surely we can just call `removeDisplayNameLabel` directly, let me do that.

Icarus9913:
I'm adding the test

Icarus9913:
Since our unit tests are written in universal memory cache store mode rather than k8s store mode, it's a little bit trivial to add/rewrite that framewrok. So, let's just add this label manually to the object that we store for the test.

pkg/api-server/testdata/service-insights/get.input.yaml
```diff
@@ -5,6 +5,8 @@ name: mesh-1
 type: ServiceInsight
 mesh: mesh-1
 name: all-services-mesh-1
+labels:
+  kuma.io/display-name: all-services-mesh-1 # refer to https://github.com/kumahq/kuma/issues/9729
```
jakubdyszkiewicz, line 9: I mean, I'd just write
```
# add display name manually to test if it's removed in the response
```
We can always refer to issue by git blame

jakubdyszkiewicz (approved):

PR #11510, opened 2024-09-23
test(framework): smarter logs on universal, multizone and gatewayapi

jijiechen:
Solving two issues:
1. When `ENABLE_E2E_DEBUG_GLOBALLY` is turned on, there is too much logs flooding the GHA output view and it's truncated (fixing again, a further step following [a previous PR](https://github.com/kumahq/kuma/pull/11506))
2. We can't collect logs from pods from GatewayAPI conformance for the moment as they are cleaned up after running. So turn off cleaning up when debug is enabled.

### Checklist prior to review

  - A helper effort when fixing flakiness: https://github.com/kumahq/kuma/actions/runs/10986369217/job/30500941118
  - Confirmed
  - Tested
  - No need
  - No

> Changelog: skip

### DIFF
--- a/test/e2e_env/gatewayapi/conformance_test.go
+++ b/test/e2e_env/gatewayapi/conformance_test.go
@@ -106,8 +106,8 @@ func TestConformance(t *testing.T) {
 		RestConfig:           clientConfig,
 		Clientset:            clientset,
 		GatewayClassName:     "kuma",
-		CleanupBaseResources: !Config.Debug,
-		Debug:                Config.Debug,
+		CleanupBaseResources: !Config.Debug, // we need to keep the resources to collect logs when Debug is enabled
+		Debug:                Config.Debug,  // controls if request details should be printed to stdout
 		NamespaceLabels: map[string]string{
 			metadata.KumaSidecarInjectionAnnotation: metadata.AnnotationEnabled,
 		},
--- a/test/e2e_env/gatewayapi/conformance_test.go
+++ b/test/e2e_env/gatewayapi/conformance_test.go
@@ -51,7 +51,7 @@ func TestConformance(t *testing.T) {
 	opts := cluster.GetKubectlOptions()
 
 	t.Cleanup(func() {
-		if t.Failed() || Config.Debug {
+		if t.Failed() {
 			var namespaces []string
 			clientset, err := k8s.GetKubernetesClientFromOptionsE(t, opts)
 			if err == nil {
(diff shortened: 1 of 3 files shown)


jijiechen:
An example lenghy output:
https://github.com/kumahq/kuma/actions/runs/10986369217/job/30500941118

test/e2e_env/gatewayapi/conformance_test.go
```diff
@@ -106,7 +106,7 @@ func TestConformance(t *testing.T) {
 		RestConfig:           clientConfig,
 		Clientset:            clientset,
 		GatewayClassName:     "kuma",
-		CleanupBaseResources: true,
+		CleanupBaseResources: !Config.Debug,
```
jijiechen, line 109: When the resources are cleaned up, we can't collect logs from these containers. We disable cleaning up when debug is enabled so that the logs/state are dumped to files.

test/e2e_env/gatewayapi/conformance_test.go
```diff
@@ -51,7 +51,7 @@ func TestConformance(t *testing.T) {
 	opts := cluster.GetKubectlOptions()
 
 	t.Cleanup(func() {
-		if t.Failed() || Config.Debug {
+		if t.Failed() {
```
jijiechen, line 54: We only dump details when failed.

test/framework/envs/multizone/env.go
```diff
@@ -261,14 +261,9 @@ func AfterSuite(report ginkgo.Report) {
...
+		framework.Logf("Please see full CP logs by downloading the debug artifacts")
 		for _, cluster := range append(Zones(), Global) {
-			Logf("\n\n\n\n\nCP logs of: " + cluster.Name())
-			logs, err := cluster.GetKumaCPLogs()
-			if err != nil {
-				Logf("could not retrieve cp logs")
-			} else {
-				Logf(logs)
-			}
+			framework.DebugUniversalCPLogs(cluster)
```
jijiechen, line 266: Same logic as https://github.com/kumahq/kuma/pull/11506/files for multizone.

test/e2e_env/gatewayapi/conformance_test.go
```diff
@@ -106,8 +106,8 @@ func TestConformance(t *testing.T) {
 		RestConfig:           clientConfig,
 		Clientset:            clientset,
 		GatewayClassName:     "kuma",
-		CleanupBaseResources: true,
-		Debug:                Config.Debug,
+		CleanupBaseResources: !Config.Debug, // we need to keep the resources to collect logs when Debug is enabled
```
jijiechen, line 109: gatewayapi cases are run in a dedicated CI job, so keeping these resources in the k8s cluster will not impact other cases.

slonka (approved):

test/e2e_env/gatewayapi/conformance_test.go
michaelbeaumont, line 109: Note that this each test case will cleanup its own resources no matter what, so don't expect too much from this

michaelbeaumont (commented):
```
		CleanupBaseResources: !Config.Debug, // we need to keep the resources to collect logs when Debug is enabled
```

Note that each test case will still clean up all of its resources, so don't expect too much from this

PR #11515, opened 2024-09-23
fix(inspect-api): amend openapi types for arbitrary objects

johncowen:
This PR alters instances of `type: object` (with no additional information) to use:

1.  `additionalProperties: true` in the case where we are saying an arbitrary object/`interface {}` (i.e. TS `Record<string, unknown>`)
2. `additionalProperties.type: string` where we are saying a `map[string]string` (i.e. TS `Record<string, string>`)

Generated files generated using `make generate/oas`.

Previous to this PR these properties were using `never` in the produced frontend TS which was causing issues.

---

**Please note:** I will probably add more amends to this PR once I've had an initial review, so initially I'm looking for comment so I know I'm doing things correctly so I can go ahead and do more before getting a final review ~(see inline question)~

---

### Checklist prior to review

### DIFF
--- a/api/openapi/specs/common/resource.yaml
+++ b/api/openapi/specs/common/resource.yaml
@@ -113,6 +113,8 @@ components:
       properties:
         tags:
           type: object
+          additionalProperties: 
+            type: string
           x-go-type: 'map[string]string'
         port:
           type: integer
@@ -123,7 +125,7 @@ components:
         conf:
           description: The actual conf generated
           type: object
-          x-go-type: 'interface{}'
+          additionalProperties: true
         origin:
           type: array
           items:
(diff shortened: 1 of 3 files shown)


api/openapi/specs/common/resource.yaml
```diff
@@ -123,7 +125,7 @@ components:
         conf:
           description: The actual conf generated
           type: object
-          x-go-type: 'interface{}'
```
johncowen, line 126: Can I remove these? If so should I also remove the `x-go-type: 'map[string]string'` form the `tags` property?

johncowen: Looks like go needs these, resolving.

slonka (approved):

lahabana (approved):

Issue #11512, opened 2024-09-23
`to[].targetRef: Mesh` doesn't work, even though it shows up in Inspect API


lobkovilya:
### What happened?

I noticed in the Inspect API we build `BaseMeshContext` and it contains its own `Mesh`. But the real `MeshContext` we build for proxy doesn't contain its own mesh (we exclude it):
https://github.com/kumahq/kuma/blob/17fe0560f9b09e4049dbef4a42cd2b6bfb8a124b/pkg/xds/context/mesh_context_builder.go#L116

New ResourceRule algo treats `kind: Mesh` as a real resource and tries to look it up in the MeshContext. If it doesn't find it, there will be no rule.

Repro:
1. Apply MeshTimeout:
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTimeout
metadata:
  name: timeout-all
  namespace: kuma-system
  labels:
    kuma.io/mesh: default
spec:
  targetRef:
    kind: Mesh
  to:
  - targetRef:
      kind: Mesh
    default:
      connectionTimeout: 29s
```
2. See the value "29s" is not present in the config dump

PR #11525, opened 2024-09-23
fix(resourcerules): add own mesh to the MeshContext, so it could be successfully resolved

lobkovilya:
### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- https://github.com/kumahq/kuma/issues/11512

### DIFF
--- a/pkg/xds/context/mesh_context_builder.go
+++ b/pkg/xds/context/mesh_context_builder.go
@@ -112,20 +112,6 @@ func (m *meshContextBuilder) BuildIfChanged(ctx context.Context, meshName string
 	for resType := range m.typeSet {
 		rl, ok := globalContext.ResourceMap[resType]
 		if ok { // Exists in global context take it from there
-			switch resType {
-			case core_mesh.MeshType: // Remove our own mesh from the list
-				otherMeshes := rl.NewItem().Descriptor().NewList()
-				for _, rentry := range rl.GetItems() {
-					if rentry.GetMeta().GetName() != meshName {
-						err := otherMeshes.AddItem(rentry)
-						if err != nil {
-							return nil, err
-						}
-					}
-				}
-				otherMeshes.GetPagination().SetTotal(uint32(len(otherMeshes.GetItems())))
-				rl = otherMeshes
-			}
 			resources.MeshLocalResources[resType] = rl
 		} else {
 			rl, ok = baseMeshContext.ResourceMap[resType]
@@ -142,7 +128,7 @@ func (m *meshContextBuilder) BuildIfChanged(ctx context.Context, meshName string
 		}
 	}
 
-	if err := m.decorateWithCrossMeshResources(ctx, resources); err != nil {
+	if err := m.decorateWithCrossMeshResources(ctx, meshName, resources); err != nil {
 		return nil, errors.Wrap(err, "failed to retrieve cross mesh resources")
 	}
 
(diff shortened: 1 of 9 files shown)


jakubdyszkiewicz (approved):
The code looks good but if I understand https://github.com/kumahq/kuma/issues/11512 the really basic case seems to be broken. We've got here because we missed the test. Can we add it?

pkg/xds/context/mesh_context_builder.go
```diff
@@ -482,10 +468,13 @@ func (m *meshContextBuilder) resolveTLSReadiness(
 	}
 }
 
-func (m *meshContextBuilder) decorateWithCrossMeshResources(ctx context.Context, resources Resources) error {
+func (m *meshContextBuilder) decorateWithCrossMeshResources(ctx context.Context, meshName string, resources Resources) error {
 	// Expand with crossMesh info
 	otherMeshesByName := map[string]*core_mesh.MeshResource{}
-	for _, m := range resources.OtherMeshes().GetItems() {
+	for _, m := range resources.Meshes().GetItems() {
+		if m.GetMeta().GetName() == meshName {
```
lahabana, line 475: Comment to explain we're skipping self

pkg/xds/context/resources.go
```diff
@@ -146,7 +146,7 @@ func (r Resources) VirtualOutbounds() *core_mesh.VirtualOutboundResourceList {
 	return r.ListOrEmpty(core_mesh.VirtualOutboundType).(*core_mesh.VirtualOutboundResourceList)
 }
 
-func (r Resources) OtherMeshes() *core_mesh.MeshResourceList {
+func (r Resources) Meshes() *core_mesh.MeshResourceList {
```
lahabana, line 149: Why not keep the `OtherMeshes` method? That could make code less error prone.

pkg/xds/generator/secrets/generator.go
```diff
@@ -113,13 +119,13 @@ func (g Generator) Generate(
 	}
 
 	if usedIdentity || len(usedCas) > 0 {
```
lahabana, line 121: Can you correct the variable names to be `usedCAs` I struggled to understand this code because I was misreading the variable

lahabana (commented):
I'd strongly argue for keeping the `OtherMeshes` method and adding a new `Meshes` method as it would make the change simpler but also makes the API less error prone

pkg/xds/context/resources.go
lobkovilya: That'd require adding another field like `LocalMeshName string` to the Resources struct:
```go
type Resources struct {
    MeshLocalResources ResourceMap
    CrossMeshResources map[xds.MeshName]ResourceMap
}
```
Given the fact at this moment we don't have constructor for `Resources` it increases the scope of this PR.

I can maybe add method:
```go
func (r Resources) OtherMeshes(localMesh string) *core_mesh.MeshResourceList {
    ...
}
```

@lahabana @michaelbeaumont wdyt?

pkg/xds/context/mesh_context_builder.go
lobkovilya: Changed this code to use `OtherMeshes(localMesh)`, probably the comment now is less needed.

pkg/xds/context/resources.go
lahabana: I think this is great!

lahabana (approved):

PR #11528, opened 2024-09-24
fix(inspect-api): `conf` should be an array of unknown structs

johncowen:
1. Amends `toResourceRules[].conf` to be an array of structs rather than a struct
2. Adds a makefile target for easier full generation of all OpenAPI related code (i.e. _both_ the golang and the `docs` YAMLs which typescript uses)

### Checklist prior to review

### DIFF
--- a/api/openapi/specs/common/resource.yaml
+++ b/api/openapi/specs/common/resource.yaml
@@ -174,9 +174,11 @@ components:
           type: string
         conf:
           description: The actual conf generated
-          type: object
-          additionalProperties: true
-          x-go-type: 'interface{}'
+          type: array
+          items:
+            type: object
+            additionalProperties: true
+            x-go-type: 'interface{}'
         origin:
           type: array
           description: The list of policies that contributed to the 'conf'. The order is important as it reflects in what order confs were merged to get the resulting 'conf'.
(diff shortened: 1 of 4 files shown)


api/openapi/specs/common/resource.yaml
```diff
@@ -174,9 +174,11 @@ components:
           type: string
         conf:
           description: The actual conf generated
-          type: object
-          additionalProperties: true
-          x-go-type: 'interface{}'
+          type: array
```
michaelbeaumont, line 177: This is the source of truth for this type, right? It's not generated from go types or something?

michaelbeaumont (approved):

api/openapi/specs/common/resource.yaml
johncowen: Yeah, this the single source of truth. I found out this generates the go types by running `make generate/oas`, but then you need to generate the complete bundled OpenAPI (the one in `/docs`) which we use for the Typescript generation, hence the new make target.

mk/generate.mk
```diff
@@ -109,6 +109,10 @@ generate/oas: $(GENERATE_OAS_PREREQUISITES)
 		PATH=$(CI_TOOLS_BIN_DIR):$$PATH oapi-codegen -config api/openapi/openapi.cfg.yaml -o api/openapi/types/$$(dirname $${DEST}})/zz_generated.$$(basename $${DEST}).go $${endpoint}.yaml; \
 	done
 
+.PHONY: generate/oas-for-ts
```
michaelbeaumont, line 112: should this be a dependency of some other target?

johncowen: This is what I'm using as my command `make generate/oas-for-ts`

Probably a bad name, and tbh I was tempted to change things a bit more, but I figured I'd be best not changing too much.

Automaat (approved):

PR #11537, opened 2024-09-24
chore(dev): refactor openapi generation make targets

johncowen:
Following https://github.com/kumahq/kuma/pull/11528 we had a chat about how ideally the make targets would be configured.

This PR:

1. Moves the old `docs/generated/openapi.yaml` target to be a named `generate/openapi` target, and moves it into the generate file.
2. Adds a dependency on `generate/oas` to the above `generate/openapi` so that everything necessary is done following an update of the YAML specs before generating the final 'bundle'/combined spec.
3. I also changed the only place I found `docs/generated/openapi.yaml` being used to use the new named rule.

### Checklist prior to review

### DIFF
--- a/mk/docs.mk
+++ b/mk/docs.mk
@@ -7,7 +7,7 @@ clean/docs:
 	rm -rf docs/generated
 
 .PHONY: docs
-docs: helm-docs docs/generated/raw docs/generated/openapi.yaml $(DOCS_EXTRA_TARGETS) ## Dev: Generate local documentation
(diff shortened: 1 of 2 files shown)


lahabana (changes requested):
`docs/generated/openapi.yaml` target is also the name of the file the spec is written too. 

So now after make generate the spec is no longer in `docs/generated/openapi.yaml but it's in `generate/openapi` which is very broken.

What's the motivation for moving the target around and changing its name?

Why do we need generate/oas as a dep btw?

johncowen:
Just to note, these questions are being answered offline. We'll be back here to either close or amend

PR #11539, opened 2024-09-24
fix(e2e): don't reuse namespace for helm upgrade test

Automaat:
### Checklist prior to review

### DIFF
--- a/test/e2e/helm/kuma_helm_upgrade_multizone.go
+++ b/test/e2e/helm/kuma_helm_upgrade_multizone.go
@@ -20,6 +20,7 @@ import (
 )
 
 func UpgradingWithHelmChartMultizone() {
+	namespace := "helm-upgrade-ns"
 	var global, zoneK8s, zoneUniversal Cluster
 	var globalCP ControlPlane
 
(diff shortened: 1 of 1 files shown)


test/e2e/helm/kuma_helm_upgrade_multizone.go
```diff
@@ -20,6 +20,7 @@ import (
 )
 
 func UpgradingWithHelmChartMultizone() {
+	namespace := "helm-upgrade-ns"
```
slonka, line 23: shouldn't we randomize this? so that we don't have to wait for ns deletion?

slonka (approved):
LGTM - one q

test/e2e/helm/kuma_helm_upgrade_multizone.go
lahabana: The k8s cluster is the same so we do need to wait for ns deletion anyway

lahabana:
Why do we think this would help anything?

Automaat:
> Why do we think this would help anything?

https://github.com/kumahq/kuma/actions/runs/11011538712/job/30576411597#step:12:3356 We failed few times on removing namespace, and since this namespace is shared in multiple tests there might be resources from other tests. Also in new tests we always create new namespace to separate test resources

jijiechen (approved):

PR #11543, opened 2024-09-24
ci(workflow): skip cis docker scan

jakubdyszkiewicz:
### Checklist prior to review

We hit rate limiting a lot in Kong Mesh.
Temporarily skip docker scan until sec team figure out trivy image cache / registry

### DIFF
--- a/.github/workflows/_build_publish.yaml
+++ b/.github/workflows/_build_publish.yaml
@@ -133,6 +133,7 @@ jobs:
           asset_prefix: image_${{ matrix.image }}-amd64
           image: ./build/docker/${{ matrix.image }}-amd64.tar
           upload-sbom-release-assets: true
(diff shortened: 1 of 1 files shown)


.github/workflows/_build_publish.yaml
```diff
@@ -133,6 +133,7 @@ jobs:
           asset_prefix: image_${{ matrix.image }}-amd64
           image: ./build/docker/${{ matrix.image }}-amd64.tar
           upload-sbom-release-assets: true
+          skip_cis_scan: true
```
jijiechen, line 136: Hi, what's the different between this property and just putting `if: false` onto the step?
I'll approve this change ftm, but if `if: false` applies, I'd rather use it.

jijiechen (approved):

PR #11548, opened 2024-09-25
fix(MeshTrace): invalid sampling default values

michaelbeaumont:
### Checklist prior to review

### DIFF
--- a/app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
+++ b/app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
@@ -6913,7 +6913,7 @@ spec:
                         anyOf:
                         - type: integer
                         - type: string
-                        default: 100%
+                        default: 100
                         description: |-
                           Target percentage of requests that will be force traced if the
                           'x-client-trace-id' header is set. Mirror of client_sampling in Envoy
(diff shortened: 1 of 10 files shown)


slonka (commented):
Don't we need to add a test to prevent re-introducing this? How did this pass `validation` tests the first time?

ok so the original PR never changed the tests to check if it work 👀 https://github.com/kumahq/kuma/blob/5d0f0648c22ea1b9e49da469fd31b9b8386a0455/pkg/plugins/policies/meshtrace/api/v1alpha1/validator_test.go#L49

michaelbeaumont:
> Don't we need to add a test to prevent re-introducing this? How did this pass validation tests the first time?

So would we have validation that sets every object field to the empty value to test if the default values of that field's fields are correct?

michaelbeaumont:
> ok so the original PR never changed the tests to check if it work 👀 

In particular this only shows up if `sampling:` is set but not all the fields are set.

slonka:
> So would we have validation that sets every object field to the empty value to test if the default values of that field's fields are correct?

That will also be beneficial. It's just that if we change the default from "100" to "100%" then we **NEED** to change the unit tests as well.

michaelbeaumont:
> It's just that if we change the default from "100" to "100%" then we NEED to change the unit tests as well.

What exactly would the test test, just that the default values we enter aren't invalid?

Note this also only shows up with Kubernetes because of #6070

slonka:
@michaelbeaumont can you add an error case with explicit "80%"?

michaelbeaumont:
> @michaelbeaumont can you add an error case with explicit "80%"?

What would that test?

slonka:
> What would that test?

well exactly the problem that we hit 🤔

slonka (approved):

Issue #11500, opened 2024-09-20
Inspect API `/_rules` to include the list of outbounds


lobkovilya:
### Description

Kuma GUI needs to create cross links to MS, MES, MMZS even when there are no specific rules for them.

We need to include the list of outbounds, like:
```yaml
outbounds:
- name: demo-app-2955c88v9z48dww6.kuma-demo # core name
  envoy: default_demo-app_kuma-demo_west_msvc_5000
  resourceIdentifier:
    type: MeshService
    mesh: default
    displayName: demo-app
    namespace: kuma-demo
    zone: west
    sectionName: "5000"
```

and change `toResourceRules[].resourceMeta` to `resourceIdentifier`:

```yaml
toResourceRules:
  - conf: {...}
    origins: [...]
    resourceIdentifier:
      type: MeshService
      mesh: default
      displayName: demo-app
      namespace: kuma-demo
      zone: west
      sectionName: "5000"
```

Matching between item in `outbounds` and item in `toResourceRules` should be done by using the `resourceIdentifier`. 

The only place where we have name with hash is `outbounds[].name` and it only should be used for creating cross links (until Kuma API gets endpoints that accept `resourceIdentifier`).

lahabana:
I'm curious also if we could add in the future internal and external inbounds/outbounds

lobkovilya:
> What happens when one of the field is empty (.e.g: zone)?

Empty fields in ResourceIdentifier are meaningful, i.e. resources from global will have their zone empty, resources from universal will have their namespace empty.

> If we put outbounds should we put inbounds too?

Yeah, I don't see why not, it's basically exactly the same format as outbounds, being able to make cross links to inbound's MeshService is useful as well. 

> Does this only exist in the _rules api or it's part of a datastructure that's more internal and that we could use in the XDS config generation?

We already have a list of outbounds in [the Proxy struct](https://github.com/kumahq/kuma/blob/749aed7d633c79cf179344936146ae76dfdf766e/pkg/core/xds/types.go#L173), but we don't call `dataplane_proxy_builder.go` for `_rules` at this moment. Though we do it in [other Inspect API endpoints](https://github.com/kumahq/kuma/blob/749aed7d633c79cf179344936146ae76dfdf766e/pkg/core/xds/types.go#L173).

> I'm curious also if we could add in the future internal and external inbounds/outbounds

by internal you mean like envoy admin or prometheus listeners or what exactly?

johncowen:
[...]

PR #11556, opened 2024-09-25
feat(inspect-api): return the list of outbounds

lobkovilya:
### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/11500

### DIFF
--- a/docs/generated/openapi.yaml
+++ b/docs/generated/openapi.yaml
@@ -2142,6 +2142,7 @@ components:
         - rules
         - resource
         - httpMatches
+        - outbounds
       properties:
         resource:
           $ref: '#/components/schemas/Meta'
@@ -2153,6 +2154,10 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/HttpMatch'
+        outbounds:
+          type: array
+          items:
+            $ref: '#/components/schemas/Outbound'
     BaseStatus:
       type: object
       title: Status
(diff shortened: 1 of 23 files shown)


docs/generated/openapi.yaml
```diff
@@ -2632,6 +2637,67 @@ components:
         match:
           type: object
           x-go-type: interface{}
+    ResourceIdentifier:
```
lahabana, line 2640: It's technically a `TypedResourceIdentifier` no?

pkg/api-server/resource_endpoints.go
```diff
@@ -981,13 +984,67 @@ func (r *resourceEndpoints) rulesForResource() restful.RouteFunction {
 		sort.Slice(httpMatches, func(i, j int) bool {
 			return httpMatches[i].Hash < httpMatches[j].Hash
 		})
+		outbounds, err := r.outboundsForDPP(request.Request.Context(), dp, meshName)
```
lahabana, line 987: For `MeshGateways/_rules` we create a fake dpp. Do you know how this works out?

pkg/api-server/resource_endpoints.go
```diff
@@ -981,13 +984,67 @@ func (r *resourceEndpoints) rulesForResource() restful.RouteFunction {
...
 		}
 	}
 }
+
+func (r *resourceEndpoints) outboundsForDPP(
+	ctx context.Context,
+	dpp *core_mesh.DataplaneResource,
+	meshName string,
+) ([]api_common.Outbound, error) {
+	meshContext, err := r.meshContextBuilder.BuildIfChanged(ctx, meshName, nil)
```
lahabana, line 1009: Do we get the full MeshContext again right? That's problematic then because the whole point of the `_rules` endpoint is that it didn't require getting all DPPs because this didn't scale at all in global as each of this calls would be in `O(#Dpp)`.

If we need to break this then we'd need to rework the MeshContextBuillder to make this more workable, in which case we should punt this to 2.10.x

PR #11560, opened 2024-09-25
fix(MeshAccessLog): strengthen validation for MeshAccessLog and MeshGateway

michaelbeaumont:
This is a breaking change. The access logging is configured at the gateway listener level, which makes it impossible to configure depending on the traffic destination.

### Checklist prior to review

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -11,7 +11,7 @@ does not have any particular instructions.
 ### MeshAccessLog
 
 Policies targeting `spec.targetRef.kind: MeshGateway` can now only target `kind: Mesh` in
-`to[].targetRef`. Previously this was allowed but the resulting configuration
+`to[].targetRef`. Previously MeshService, MeshExternalService, MeshMultiZoneService were allowed but the resulting configuration
 was ambiguous and nondeterministic.
 
 ### MeshExternalService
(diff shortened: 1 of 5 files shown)


lahabana:
The Changelog still needs to be a conventional commit. Isn't it more a change to `UPGRADE.md` that you want?

michaelbeaumont:
> The Changelog still needs to be a conventional commit. Isn't it more a change to UPGRADE.md that you want?

Yeah you're right of course

UPGRADE.md
```diff
@@ -8,6 +8,12 @@ does not have any particular instructions.
 
 ## Upgrade to `2.9.x`
 
+### MeshAccessLog
+
+Policies targeting `spec.targetRef.kind: MeshGateway` can now only target `kind: Mesh` in
+`to[].targetRef`. Previously this was allowed but the resulting configuration
+was ambiguous and nondeterministic.
```
slonka, line 15: I don't understand it, what do you mean by `this`? Shouldn't it be:

```suggestion
Policies targeting `spec.targetRef.kind: MeshGateway` can now only target `kind: Mesh` in
`to[].targetRef`. Previously MeshService, MeshExternalService, MeshMultiZoneService were allowed but the resulting configuration
was ambiguous and nondeterministic.
```

slonka (commented):
One thing from changelog seems off to me.

slonka (approved):

PR #11578, opened 2024-09-26
fix(mads): add mutex when checking if reconcile is needed and reconciling

lobkovilya:
This is a straightforward fix for the MADS problem we see in unit tests. But we probably don't need `NeedsReconciliation` at all since we fixed watchdogs to reconcile before the first tick https://github.com/kumahq/kuma/pull/11105

CC @Automaat @michaelbeaumont 

### Checklist prior to review


pkg/mads/v1/reconcile/reconciler.go
```diff
@@ -18,12 +18,14 @@ func NewReconciler(hasher envoy_cache.NodeHash, cache util_xds_v3.SnapshotCache,
...
 		knownClientIds:      map[string]bool{},
 		knownClientIdsMutex: &sync.Mutex{},
+		cacheMutex:          &sync.Mutex{},
 	}
 }
 
 type reconciler struct {
 	hasher              envoy_cache.NodeHash
 	cache               util_xds_v3.SnapshotCache
+	cacheMutex          *sync.Mutex
```
lobkovilya, line 28: ah probably should rename the mutex to something else

Automaat:
> But we probably don't need NeedsReconciliation at all 
 
I am not sure about this, MADS works in a weird way since it is responding to http requests not relying entirely on watchdogs.

Automaat (approved):

lahabana (approved):
This is simple fix. The more complex fix is trickier because of these reconcilers running concurrently makes 0 sense. So this will work for now

slonka (approved):

Icarus9913:
> I am not sure about this, MADS works in a weird way since it is responding to http requests not relying entirely on watchdogs.

Because the watchdog runs in goroutine and we need every interval to generate the XDS. And each http request gets response from go-control-plane `fetch` method.

Although we got that first tick reconcile from #11105, but you can imagine that the request happens between the first and second tick to fetch xds configurations then it would not found something new.

PR #11583, opened 2024-09-27
feat(kubectl): include eds when inspecting config-dump

jijiechen:
We do support the "include eds" feature in the API, but our `kumactl` does not use it. So let's include eds by default when trying to dump the dataplane xds config.

### Checklist prior to review

  - N/A
  - Manual tested


Automaat:
shouldn't this be a flag for a command? I am not sure if we always need eds when getting config dump

lahabana:
+1 to @Automaat 's point. I think a flag would be nice

jijiechen:
I've made a new flag. Could you take a second look? @lahabana @Automaat

jakubdyszkiewicz (approved):

PR #11590, opened 2024-09-27
chore(mads/hds): remove custom cache implementation

lahabana:
In mads and hds we were using a custom implementation of a cache. This was unfortunate as go-control-plane has a perfectly working one.

The only feature on this cache was the ability to long poll. This was only required in mads and we've changed the handler of mads to make this handling outside of the cache.

### Checklist prior to review


pkg/mads/v1/service/mads.go
```diff
@@ -16,17 +17,44 @@ type Server interface {
...
+	resChan := make(chan envoy_cache.Response, 1)
+	streamState := stream.NewStreamState(false, nil)
+	// Because we want to do long polling we use the watch system if there's a deadline on the context
+	if _, hasDeadline := ctx.Deadline(); hasDeadline {
+		cancelWatch := s.cache.CreateWatch(req, streamState, resChan)
+		defer cancelWatch()
+		select { // Wait until either we timeout or the watch triggers
+		case <-ctx.Done():
+		case <-resChan:
+		}
```
Icarus9913, line 46: > CreateWatch returns a watch for an xDS request.  A nil function may be returned if an error occurs.

https://github.com/envoyproxy/go-control-plane/blob/bd58d40e8ff97f5da06c9a7b979142b403d0ff26/pkg/cache/v3/simple.go#L386-L388

pkg/mads/v1/service/mads.go
```diff
@@ -16,17 +17,44 @@ type Server interface {
...
+	}
+	resChan := make(chan envoy_cache.Response, 1)
+	streamState := stream.NewStreamState(false, nil)
+	// Because we want to do long polling we use the watch system if there's a deadline on the context
+	if _, hasDeadline := ctx.Deadline(); hasDeadline {
+		cancelWatch := s.cache.CreateWatch(req, streamState, resChan)
+		defer cancelWatch()
+		select { // Wait until either we timeout or the watch triggers
+		case <-ctx.Done():
```
jakubdyszkiewicz, line 44: should we return here?

pkg/mads/v1/service/mads.go
```diff
@@ -16,17 +17,44 @@ type Server interface {
...
+		case <-resChan:
+		}
+	}
+	resp, err := s.cache.Fetch(context.WithoutCancel(ctx), req)
```
jakubdyszkiewicz, line 48: why without cancel?

pkg/mads/v1/service/service.go
```diff
@@ -1,46 +1,65 @@
...
+		server:   srv,
+		config:   config,
+		log:      log,
+		watchdog: watchdog,
 	}
 }
+
+func (s *service) Start(ctx context.Context) {
+	go s.watchdog.Start(ctx)
+	s.watchdog.HasTicked(true)
```
jakubdyszkiewicz, line 64: couldn't you just call `s.watchdog.OnTick()` instead of those tick checks?

lahabana: No because `OnTick` is rarely threadSafe, I find `HasTicked` something that could be quite useful anyway as a construct...

jakubdyszkiewicz: I meant before starting the watchdog, then the thread safety is not a concern

pkg/mads/v1/service/mads.go
jakubdyszkiewicz: it helps, I get it now. Thanks

jakubdyszkiewicz (approved):

PR #8796, opened 2024-01-10
feat(xds): do not generate independent listener for vips, use additional_addresses instead

jijiechen:
fixes https://github.com/kumahq/kuma/issues/6717

I haven't run a full CI on local env yet, so creating this PR as a draft. 

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues 
  - https://github.com/kumahq/kuma/issues/6717
- [x] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS 
  - Confirmed
- [x] Tests (Unit test, E2E tests, manual test on universal and k8s) 
  - Done
  - Don't forget `ci/` labels to run additional/fewer tests
  - No need to add
- [x] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? 
  - No
- [x] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) 
  - No

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #11601, opened 2024-09-30
revert(kuma-cp): do not use additional addresses

lukidzi:
### Checklist prior to review

When using TrafficRoute, a user might encounter an issue with address conflicts in the listeners. If the vips-config-map (which holds internal VIPs) is removed, the control plane attempts to rebuild it. However, the order in which the map is rebuilt may differ from the addresses previously allocated and delivered by the control plane to the sidecars. If a sidecar runs with listeners that have additional addresses defined, and the rebuilt map assigns a different IP to a service, this IP mismatch can cause a conflict. As a result, the listener may be rejected by Envoy.

revert: https://github.com/kumahq/kuma/pull/8796

- [ ] [Link to relevant issue][1] as well as docs and UI issues -- fix: https://github.com/Kong/kong-mesh/issues/6679


pkg/xds/generator/outbound_proxy_generator.go
```diff
@@ -50,11 +47,10 @@ func (g OutboundProxyGenerator) Generate(ctx context.Context, _ *model.ResourceS
 	// If we have same split in many HTTP matches we can use the same cluster with different weight
 	clusterCache := map[string]string{}
 
-	outboundsMultipleIPs := buildOutboundsWithMultipleIPs(proxy.Dataplane, outbounds, xdsCtx.Mesh.VIPDomains)
```
slonka, line 53: Is this still used anywhere?

slonka (approved):
LGTM, one Q

pkg/xds/generator/outbound_proxy_generator.go
lukidzi: It was only used here

Issue #11473, opened 2024-09-19
Inspect API `_rules` don't set `toRules` field when `meshServices.enabled: Exclusive`


lobkovilya:
### Description

When MeshServices are enabled exclusively, we don't need `toRules` field as all the relevant confs are in `toResourceRules`.

PR #11623, opened 2024-10-02
fix(inspect-api): don't set 'toRules' when `meshServices.mode: Exclusive`

lobkovilya:
### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues -- Fix https://github.com/kumahq/kuma/issues/11473

### DIFF
--- a/docs/generated/openapi.yaml
+++ b/docs/generated/openapi.yaml
@@ -2601,12 +2601,16 @@ components:
           $ref: '#/components/schemas/ProxyRule'
         toRules:
           type: array
-          description: a set of rules for the outbounds of this proxy
+          description: >-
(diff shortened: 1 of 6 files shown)


docs/generated/openapi.yaml
```diff
@@ -2601,12 +2601,16 @@ components:
           $ref: '#/components/schemas/ProxyRule'
         toRules:
           type: array
-          description: a set of rules for the outbounds of this proxy
+          description: >-
+            a set of rules for the outbounds of this proxy. The field is not set
+            when 'meshService.mode' is 'Exclusive'.
```
Automaat, line 2606: can make this description more specific? this resource does not have `meshService` param so it might be confusing

Automaat (approved):

PR #11667, opened 2024-10-04
fix(egress): same external service tag in multiple meshes

jakubdyszkiewicz:
## Motivation

We had a problem when you used Zone Egress with `ExternalService` with the same `kuma.io/service` in multiple meshes.

## Implementation information

Whenever we added a Resource to ResourceSet, we added it with `serviceName` instead of `clusterName` (which contains mesh information as well). This way if you added both Clusters from multiple meshes, one replaced another.

## Supporting documentation

No issue. I noticed the flake in E2E test.


lahabana (approved):
Good find!

PR #11674, opened 2024-10-07
ci(github): fix checklist PR comment

lahabana:
## Supporting documentation

xrel: #11654 #11666


lukidzi (approved):

.github/workflows/check.yaml
```diff
@@ -1,4 +1,3 @@
-name: "PR health"
```
michaelbeaumont, line 1: Why this change?

lahabana: Because the title makes debugging just harder:

michaelbeaumont: @lahabana i think your image is missing

michaelbeaumont: Does it? That's so noisy to me. the awful prefix in front of everything and then `.yaml`. Can't you always jump to Workflow file on the left of the run summary and even see the name at the top of the list of runs for that workflow?

Issue #11634, opened 2024-10-02
Remove MeshServices when `meshService.mode` is switched from `Exclusive/Everywhere` to `Disabled`


lobkovilya:
### What happened?

At this moment switching mode back to `Disabled` changes nothing – meshService clusters still exist and referenced by listeners.

PR #11675, opened 2024-10-07
fix(kuma-cp): remove automatically created MeshServices when mode is switched to `Disabled`

lobkovilya:
## Motivation

Users should have a safe way to rollback the MeshService migration. Today switching from Disabled to Exclusive to Disabled leads to a situation when MeshServices are not deleted.

## Implementation information

There are 2 changes for Kubernetes and Universal. When mode is Disabled we remove automatically created MeshServices.

The alternative would be to keep generated MeshServices when mode is switched to Disabled, but check the mode on mesh every time we generate Listeners and Clusters. That creates another potential state of the cluster that we should take into account. The cleanest approach is to rollback to the exact state user had before the migration.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/11634

### DIFF
--- a/pkg/core/resources/apis/meshservice/generate/generator.go
+++ b/pkg/core/resources/apis/meshservice/generate/generator.go
@@ -23,7 +23,7 @@ import (
 	"github.com/kumahq/kuma/pkg/core/user"
 	core_metrics "github.com/kumahq/kuma/pkg/metrics"
 	"github.com/kumahq/kuma/pkg/plugins/runtime/k8s/metadata"
-	"github.com/kumahq/kuma/pkg/xds/cache/mesh"
+	mesh_cache "github.com/kumahq/kuma/pkg/xds/cache/mesh"
 	xds_context "github.com/kumahq/kuma/pkg/xds/context"
 )
 
@@ -39,7 +39,7 @@ type Generator struct {
 	deletionGracePeriod time.Duration
 	metric              prometheus.Summary
 	resManager          manager.ResourceManager
-	meshCache           *mesh.Cache
+	meshCache           *mesh_cache.Cache
 }
 
 var _ component.Component = &Generator{}
@@ -50,7 +50,7 @@ func New(
 	deletionGracePeriod time.Duration,
 	metrics core_metrics.Metrics,
 	resManager manager.ResourceManager,
-	meshCache *mesh.Cache,
+	meshCache *mesh_cache.Cache,
 ) (*Generator, error) {
 	metric := prometheus.NewSummary(prometheus.SummaryOpts{
 		Name:       "component_meshservice_generator",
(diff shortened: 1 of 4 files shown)


pkg/core/resources/apis/meshservice/generate/generator.go
```diff
@@ -290,6 +290,19 @@ func (g *Generator) Start(stop <-chan struct{}) error {
 					dataplanes := meshCtx.Resources.Dataplanes()
 					meshServices := meshCtx.Resources.MeshServices()
 					g.generate(ctx, mesh, dataplanes.Items, meshServices.Items)
+				} else {
+					log := g.logger.WithValues("mesh", mesh)
```
lukidzi, line 294: Do we need this? 
Can't we do at 299? 

```go
log := g.logger.WithValues("mesh", mesh, "meshService", meshService.GetMeta().GetName())
```

lukidzi (changes requested):
I think you need to specify `meshServices.mode` in [this line](https://github.com/kumahq/kuma/blob/master/test/e2e_env/multizone/meshtimeout/meshtimeout.go#L27), as it is set to `Disabled` by default. Also, could we use a `samples.MeshMTLSBuilder()...`

lukidzi:
Another falke in [this line](https://github.com/kumahq/kuma/blob/master/test/e2e_env/kubernetes/meshcircuitbreaker/meshcircuitbreaker.go#L199). We manually create a `MeshService`, and it has the `kuma.io/managed-by: k8s-controller` label, which was likely removed because the mode was set to `Disabled`.

lukidzi:
It also might fail here https://github.com/kumahq/kuma/blob/master/test/e2e_env/multizone/reachablebackends/reachablebackends.go#L188 (places where we create a `MeshService` manually - worth checking all tests)

test/e2e_env/kubernetes/meshcircuitbreaker/meshcircuitbreaker.go
```diff
@@ -19,36 +21,18 @@ func MeshCircuitBreaker() {
...
-`, namespace, mesh, namespace)
-
 	BeforeAll(func() {
 		err := NewClusterSetup().
-			Install(MeshKubernetes(mesh)).
+			Install(
+				Yaml(
+					builders.Mesh().
+						WithName(mesh).
+						WithMeshServicesEnabled(mesh_proto.Mesh_MeshServices_Exclusive),
```
lukidzi, line 30: I am not sure if that should be `Mesh_MeshServices_Exclusive` since the tests are for the `kuma.io/service` maybe `Mesh_MeshServices_Everywhere`?

lobkovilya: ah right I missed this one

lukidzi (approved):

Issue #11546, opened 2024-09-24
Solve conntrack issue for Universal E2E tests


jakubdyszkiewicz:
### Description

For some reason we did not enable conntrack fix for e2e tests https://github.com/kumahq/kuma/pull/3459#issuecomment-1016276550

We now see DNS errors because of it

PR #11684, opened 2024-10-07
fix(transparent-proxy): conntrack zone splitting in docker containers with custom network

bartsmykla:
> [!TIP]
> Please review on a per-commit basis as I included in this PR few small refactorings

## Motivation

In e2e universal tests, we were encountering issues where DNS queries bypassed our iptables rules and were sent directly to upstream DNS servers. This typically occurred when a high volume of DNS queries overwhelmed conntrack. The existing mechanism to address this was ineffective in Docker containers within custom networks, as Docker applies DNAT rules that modify the destination port for DNS traffic. This PR introduces support for conntrack zone splitting in such scenarios, ensuring proper handling of DNS traffic in our e2e tests

## Implementation information

These commits solve the actual issue:
- https://github.com/kumahq/kuma/pull/11684/commits/1d0e04c4632966a3c76d1f31a1d3ae4d9118bdf0 - when setting up DNS capture for specific DNS servers listed in `/etc/resolv.conf` (without capturing all DNS traffic), we add a rule to the raw table, such as `-A PREROUTING -d 10.12.1.3/32 -p udp -m udp --sport 53 -j CT --zone 1`. This rule is meant to match response packets from the upstream DNS server and move them to conntrack zone 1, associating them with the flows created by outgoing requests. However, this caused pairs of `UNREPLIED` conntrack flows when communicating with the upstream DNS server, where there should only be a single established flow in zone 1. This commit resolves that issue
- https://github.com/kumahq/kuma/pull/11684/commits/f88ec2a3794d8bc5941882a1443dd33ffde78c87 - this commit adds a rule to handle DNS responses in scenarios where the transparent proxy is installed in Docker containers within custom networks. In such cases, Docker may NAT the destination port to a random one. The rule ensures that DNS responses from local DNS servers (e.g., 127.0.0.11) to localhost are assigned to conntrack zone 1. This prevents issues  where the query port has been altered by a DNAT iptables rule, ensuring proper flow matching and preventing unexpected behavior when dealing with local DNS responses
- https://github.com/kumahq/kuma/pull/11684/commits/c0d1e01670dc19eaec389558a7dcc70869865a07

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/11546


lukidzi (changes requested):
Thanks for the explanation, it's great. 🎉 
Could we only remove `FlakeAttempts` here: https://github.com/kumahq/kuma/blob/master/test/e2e_env/universal/universal_suite_test.go#L95

bartsmykla:
> Could we only remove `FlakeAttempts` here: https://github.com/kumahq/kuma/blob/master/test/e2e_env/universal/universal_suite_test.go#L95

done

lukidzi (approved):

Issue #11124, opened 2024-08-15
`kuma.io/mesh` does not work on labels of namespaces or pods


jijiechen:
### What happened?

According to #10580 We are now recommending using labels instead of annoations to set mesh name for pods on Kubernetes. 

However, it is found by some users that label `kuma.io/mesh` does not work as expected. Only annotations work as documented.

Reproduced version: `2.8.2`

lukidzi:
> Triage: Let's test it with master branch and fix it if the case

lahabana:
@jakubdyszkiewicz is this expected?

jakubdyszkiewicz:
@lahabana no

PR #11690, opened 2024-10-07
fix(kuma-cp): deprecate use kuma.io/mesh annotation and use label instead

lukidzi:
## Motivation

We changed the logic to retrieve the mesh name from the namespace's labels instead of annotations https://github.com/kumahq/kuma/pull/10580. However, the change wasn't applied everywhere, so the `kuma.io/mesh` label didn't work properly on the namespace.

## Implementation information

Instead of using `MeshOfByAnnotation`, I've updated the code to use `MeshOfByLabelOrAnnotation` for consistency. This ensures that the deprecation message will be displayed correctly and the label will be used. Also, we now set `labels` with `kuma.io/mesh` instead of setting the annotations.

## Supporting documentation

Suggests to use a label
https://kuma.io/docs/2.8.x/reference/kubernetes-annotations/#kumaiomesh

Fix https://github.com/kumahq/kuma/issues/11124

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/configmap_controller_test.go
+++ b/pkg/plugins/runtime/k8s/controllers/configmap_controller_test.go
@@ -201,7 +201,7 @@ var _ = Describe("ServiceToConfigMapMapper", func() {
 				},
 			},
 			[]kube_core.Pod{
-				podFn("pod1", map[string]string{"app": "app1"}, map[string]string{metadata.KumaMeshLabel: "mesh1"}),
+				podFn("pod1", map[string]string{"app": "app1"}, map[string]string{metadata.KumaMeshAnnotation: "mesh1"}),
 				podFn("pod2", map[string]string{"app": "app1"}, nil),
 			},
 			[]string{"kuma-mesh1-dns-vips"},
--- a/pkg/plugins/runtime/k8s/controllers/configmap_controller_test.go
+++ b/pkg/plugins/runtime/k8s/controllers/configmap_controller_test.go
@@ -115,7 +115,7 @@ var _ = Describe("ServiceToConfigMapMapper", func() {
 			defaultNs,
 			[]kube_core.Pod{
 				podFn("pod1", map[string]string{"app": "app2"}, nil),
-				podFn("pod2", map[string]string{"app": "app2"}, map[string]string{metadata.KumaMeshAnnotation: "mesh2"}),
+				podFn("pod2", map[string]string{"app": "app2", metadata.KumaMeshLabel: "mesh2"}, map[string]string{}),
 			},
 			[]string{},
 		),
@@ -124,7 +124,7 @@ var _ = Describe("ServiceToConfigMapMapper", func() {
 			defaultNs,
 			[]kube_core.Pod{
 				podFn("pod1", map[string]string{"app": "app1"}, nil),
-				podFn("pod2", map[string]string{"app": "app2"}, map[string]string{metadata.KumaMeshAnnotation: "mesh2"}),
+				podFn("pod2", map[string]string{"app": "app2", metadata.KumaMeshLabel: "mesh2"}, map[string]string{}),
 			},
 			[]string{"kuma-default-dns-vips"},
 		),
@@ -132,8 +132,8 @@ var _ = Describe("ServiceToConfigMapMapper", func() {
 			serviceFn(map[string]string{"app": "app1"}),
 			defaultNs,
 			[]kube_core.Pod{
-				podFn("pod1", map[string]string{"app": "app1"}, map[string]string{metadata.KumaMeshAnnotation: "mesh1"}),
-				podFn("pod2", map[string]string{"app": "app2"}, map[string]string{metadata.KumaMeshAnnotation: "mesh2"}),
+				podFn("pod1", map[string]string{"app": "app1", metadata.KumaMeshLabel: "mesh1"}, map[string]string{}),
+				podFn("pod2", map[string]string{"app": "app2", metadata.KumaMeshLabel: "mesh2"}, map[string]string{}),
 			},
 			[]string{"kuma-mesh1-dns-vips"},
 		),
@@ -141,8 +141,8 @@ var _ = Describe("ServiceToConfigMapMapper", func() {
 			serviceFn(map[string]string{"app": "app1"}),
 			defaultNs,
 			[]kube_core.Pod{
-				podFn("pod1", map[string]string{"app": "app1"}, map[string]string{metadata.KumaMeshAnnotation: "mesh1"}),
-				podFn("pod2", map[string]string{"app": "app1"}, map[string]string{metadata.KumaMeshAnnotation: "mesh2"}),
+				podFn("pod1", map[string]string{"app": "app1", metadata.KumaMeshLabel: "mesh1"}, map[string]string{}),
+				podFn("pod2", map[string]string{"app": "app1", metadata.KumaMeshLabel: "mesh2"}, map[string]string{}),
 			},
 			[]string{"kuma-mesh1-dns-vips", "kuma-mesh2-dns-vips"},
 		),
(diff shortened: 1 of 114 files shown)


michaelbeaumont (dismissed):
Doesn't the injector need to apply a label now then?

pkg/plugins/runtime/k8s/metadata/annotations.go
```diff
@@ -11,11 +11,6 @@ import (
 
 // Annotations that can be used by the end users.
 const (
-	// KumaMeshAnnotation defines a Pod annotation that
-	// associates a given Pod with a particular Mesh.
-	// Annotation value must be the name of a Mesh resource.
```
michaelbeaumont, line 16: i kind of think it'd be nice to keep this, mark it deprecated and then use it at the one place where we check Annotations

michaelbeaumont: and also in the tests. this way it's clear that we have something deprecated still

pkg/plugins/runtime/k8s/controllers/configmap_controller_test.go
```diff
@@ -201,7 +201,7 @@ var _ = Describe("ServiceToConfigMapMapper", func() {
 				},
 			},
 			[]kube_core.Pod{
-				podFn("pod1", map[string]string{"app": "app1"}, map[string]string{metadata.KumaMeshLabel: "mesh1"}),
+				podFn("pod1", map[string]string{"app": "app1"}, map[string]string{metadata.KumaMeshAnnotation: "mesh1"}),
```
michaelbeaumont, line 204:
```suggestion
				podFn("pod1", map[string]string{"app": "app1"}, map[string]string{metadata.KumaMeshAnnotation: "mesh1"}), // nolint:staticcheck
```

right? instead of blanket ignoring it

lukidzi:
I've changed the log message from `"ignoring internal labels when converting labels to tags"` to debug, as we have a label `kuma.io/mesh`, and this message was printed for each inbound request. Do you think we should keep it as an info log and simply ignore the mesh label? Is this log useful?

pkg/plugins/runtime/k8s/metadata/annotations.go
```diff
@@ -11,6 +11,8 @@ import (
 
 // Annotations that can be used by the end users.
 const (
+	// Deprecated: use KumaMeshLabel from pkg/plugins/runtime/k8s/metadata/labels.go instead.
```
michaelbeaumont, line 14: nit pick: 

```suggestion
	// Deprecated: use KumaMeshLabel as a label from pkg/plugins/runtime/k8s/metadata/labels.go instead of this annotation.
```

pkg/plugins/runtime/k8s/util/util.go
```diff
@@ -137,37 +136,24 @@ func CopyStringMap(in map[string]string) map[string]string {
...
 
 	// Label wasn't found on the object, let's look on the namespace instead
-	if mesh, exists := metadata.Annotations(namespace.GetLabels()).GetString(metadata.KumaMeshAnnotation); exists && mesh != "" {
+	if mesh, exists := metadata.Annotations(namespace.GetLabels()).GetString(metadata.KumaMeshLabel); exists && mesh != "" {
 		return mesh
 	}
 
-	if mesh, exists := metadata.Annotations(namespace.GetAnnotations()).GetString(metadata.KumaMeshAnnotation); exists && mesh != "" {
+	if mesh, exists := metadata.Annotations(namespace.GetAnnotations()).GetString(metadata.KumaMeshAnnotation); exists && mesh != "" { // nolint:staticcheck
 		log.Info("WARNING: The kuma.io/mesh annotation is deprecated for this object kind. Use label instead", "name", obj.GetName(), "namespace", obj.GetNamespace(), "kind", obj.GetObjectKind().GroupVersionKind().Kind)
```
lahabana, line 157:
```suggestion
		log.Info("WARNING: The kuma.io/mesh annotation is no longer supported. Use label instead", "name", obj.GetName(), "namespace", obj.GetNamespace(), "kind", obj.GetObjectKind().GroupVersionKind().Kind)
```

pkg/plugins/runtime/k8s/webhooks/injector/injector.go
```diff
@@ -189,14 +196,22 @@ func (i *KumaInjector) InjectKuma(ctx context.Context, pod *kube_core.Pod) error
...
 			pod.Annotations[key] = value
 		}
 
+		if pod.Labels != nil {
+			pod.Labels[metadata.KumaMeshLabel] = meshName
+		} else {
+			pod.Labels = map[string]string{
+				metadata.KumaMeshLabel: meshName,
+			}
+		}
```
michaelbeaumont, line 213: also nit pick but I think

```suggestion
		if pod.Labels == nil {
			pod.Labels = map[string]string{}
		}
		pod.Labels[metadata.KumaMeshLabel] = meshName
```

is way less noisy

pkg/transparentproxy/kubernetes/kubernetes_config.go
```diff
@@ -484,13 +484,11 @@ func ConfigToAnnotations(
 	cfg tproxy_config.Config,
 	runtimeCfg k8s.Injector,
 	annotations map[string]string,
-	mesh string,
 	defaultAdminPort uint32,
 ) (map[string]string, error) {
 	result := map[string]string{
 		k8s_metadata.KumaSidecarInjectedAnnotation:                 k8s_metadata.AnnotationTrue,
 		k8s_metadata.KumaTransparentProxyingAnnotation:             k8s_metadata.AnnotationEnabled,
-		k8s_metadata.KumaMeshAnnotation:                            mesh, // either user-defined value or default
```
michaelbeaumont, line 493: does this need to be now be put in labels by transparentproxy? :thinking:

michaelbeaumont (approved):

pkg/plugins/runtime/k8s/util/util.go
```diff
@@ -137,37 +136,24 @@ func CopyStringMap(in map[string]string) map[string]string {
...
 // MeshOfByLabelOrAnnotation returns the mesh of the given object according to its own
 // annotations or labels or the annotations of its namespace. It treats the annotation
 // directly on the object as deprecated.
 func MeshOfByLabelOrAnnotation(log logr.Logger, obj kube_client.Object, namespace *kube_core.Namespace) string {
 	if mesh, exists := metadata.Annotations(obj.GetLabels()).GetString(metadata.KumaMeshLabel); exists && mesh != "" {
 		return mesh
 	}
-	if mesh, exists := metadata.Annotations(obj.GetAnnotations()).GetString(metadata.KumaMeshAnnotation); exists && mesh != "" {
+	if mesh, exists := metadata.Annotations(obj.GetAnnotations()).GetString(metadata.KumaMeshAnnotation); exists && mesh != "" { // nolint:staticcheck
 		log.Info("WARNING: The kuma.io/mesh annotation is deprecated for this object kind. Use label instead", "name", obj.GetName(), "namespace", obj.GetNamespace(), "kind", obj.GetObjectKind().GroupVersionKind().Kind)
```
lukidzi, line 147:
```suggestion
		log.Info("WARNING: The kuma.io/mesh annotation is no longer supported. Use label instead", "name", obj.GetName(), "namespace", obj.GetNamespace(), "kind", obj.GetObjectKind().GroupVersionKind().Kind)
```

pkg/transparentproxy/kubernetes/kubernetes_config.go
lukidzi: Yes, It was called by `injector.go` and now we add it there

michaelbeaumont: OK got it

Icarus9913 (approved):

jijiechen (approved):

PR #11691, opened 2024-10-07
chore(gha): cancel in progress on prs to not clog up CI

slonka:
## Motivation

We're sometimes seeing long queue times and that can be partially due to people pushing to PRs often.

## Implementation information

I'm adding a concurrency group to cancel jobs in progress on PRs only.

## Supporting documentation

xrel https://github.com/kumahq/kuma/pull/10146/files

### DIFF
--- a/.github/workflows/build-test-distribute.yaml
+++ b/.github/workflows/build-test-distribute.yaml
@@ -14,8 +14,8 @@ env:
   KUMA_DIR: "."
   CI_TOOLS_DIR: "/home/runner/work/kuma/kuma/.ci_tools"
 concurrency:
-  group: ${{github.workflow}}-${{ github.ref == 'refs/heads/master' && github.sha || github.event.pull_request.number }}
-  cancel-in-progress: ${{ github.ref == 'refs/heads/master' && false || true }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-distribute.yaml
```diff
@@ -13,6 +13,9 @@ permissions:
 env:
   KUMA_DIR: "."
   CI_TOOLS_DIR: "/home/runner/work/kuma/kuma/.ci_tools"
+concurrency:
+  group: ${{github.workflow}}-${{ github.ref == 'refs/heads/master' && github.sha || github.event.pull_request.number }}
```
michaelbeaumont, line 17: any reason not to just use `github.event_name == 'pull_request'`?

michaelbeaumont: i just mean why use `github.ref == 'refs/heads/master'`? don't we really mean `event_name == 'push'`? I find that makes it easier to understand what's happening

slonka: you're right

michaelbeaumont (approved):
LGTM let's try it

PR #11716, opened 2024-10-08
test(e2e): check higher number of procs

jakubdyszkiewicz:
## Supporting documentation

Fix #XX


Automaat:
Have you seen this in docs: 
```
Note that it is not recommended that you run specs in parallel with -p on CI. Some CI services run on shared machines that will report (e.g.) 32 cores but will not actually give an individual account access to all those compute resources!
```
Also didn't we use this somewhere in the past and removed it because of some issues?

lahabana:
The idea is that we mostly sleep in our e2e tests so we can over provision the number of parallel tests.
This is completely the opposite for unit tests which never sleep and would use all the CPU they can get :)

lahabana:
Also should we look at improve unit tests too: https://github.com/kumahq/kuma/blob/9917a3c9d522955ea327511fa82a5e3eb785fcef/mk/test.mk#L12 

We seem to be pinning but I feel like we have 4 cores ...

PR #11719, opened 2024-10-09
chore(ci): kick ci

slonka:
check CI stability

## Supporting documentation

Fix #XX


lukidzi:
I think you need to create a branch from the repo not a fork to run arm runners

PR #11732, opened 2024-10-09
chore(transparent-proxy): deprecated code cleanup and small improvements

bartsmykla:
description in progress

## Supporting documentation

Fix #XX


pkg/transparentproxy/config/config.go
```diff
@@ -486,13 +483,25 @@ type Log struct {
 type Retry struct {
 	// MaxRetries specifies the number of retries after the initial attempt.
 	// A value of 0 means no retries, and only the initial attempt will be made.
-	MaxRetries int `json:"maxRetries" split_words:"true"` // KUMA_TRANSPARENT_PROXY_RETRY_MAX_RETRIES
+	MaxRetries uint `json:"maxRetries" split_words:"true"` // KUMA_TRANSPARENT_PROXY_RETRY_MAX_RETRIES
```
Icarus9913, line 486: May I know why do we need to set it `uint` type?

Icarus9913 (approved):

pkg/transparentproxy/config/config.go
bartsmykla: Basically, we don't have to, but it's more syntactically correct

PR #11672, opened 2024-10-06
test(e2e): ci stability

jakubdyszkiewicz:
## Motivation

Check CI stability

## Implementation information

Commits triggered by GHA

## Supporting documentation

No issue

PR #11734, opened 2024-10-09
test(e2e): ci stability app token

jakubdyszkiewicz:
## Motivation

We need to pass PAT to trigger a CI on commits we push

tested from a branch here https://github.com/kumahq/kuma/pull/11672

## Supporting documentation

Noticed CI not being triggered

### DIFF
--- a/.github/workflows/ci-stability.yaml
+++ b/.github/workflows/ci-stability.yaml
@@ -24,6 +24,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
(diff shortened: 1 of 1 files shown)


michaelbeaumont (commented):
@jakubdyszkiewicz this is happening because the commit is triggered by a commit made by the Github actions account right?

jakubdyszkiewicz:
correct https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

PR #11736, opened 2024-10-09
fix(kuma-cp): map port to section name for reachable backends

lukidzi:
## Motivation

While testing reachable backends, I noticed that when a port is specified but the port name is not a string (i.e., it's not the actual port name), I got 0 clusters.

## Implementation information

Once we retrieve a ResourceIdentifier and the port is provided, we look up the MeshService or MeshMultiZoneService from the map of resources and attempt to find a matching port name, if one exists. If no port name is found, we treat the port as a string.

## Supporting documentation

Noticed once testing

### DIFF
--- a/pkg/core/resources/apis/mesh/dataplane_validator.go
+++ b/pkg/core/resources/apis/mesh/dataplane_validator.go
@@ -280,9 +280,6 @@ func validateTransparentProxying(tp *mesh_proto.Dataplane_Networking_Transparent
 			if (backendRef.Name != "" || backendRef.Namespace != "") && len(backendRef.Labels) > 0 {
 				result.AddViolationAt(path.Index(i).Field("labels"), "labels cannot be defined when name is specified")
 			}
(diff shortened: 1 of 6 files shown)


pkg/core/resources/apis/mesh/dataplane_validator.go
```diff
@@ -280,6 +280,9 @@ func validateTransparentProxying(tp *mesh_proto.Dataplane_Networking_Transparent
 			if (backendRef.Name != "" || backendRef.Namespace != "") && len(backendRef.Labels) > 0 {
 				result.AddViolationAt(path.Index(i).Field("labels"), "labels cannot be defined when name is specified")
 			}
+			if len(backendRef.Labels) > 0 && backendRef.Port != nil {
```
lobkovilya, line 283: why can't we use port and labels at the same time?

lukidzi: My thought was to do not filter out some services by port but I think we can do it

lobkovilya (approved):

PR #11737, opened 2024-10-09
fix(e2e): port forward to service instead of pod of kuma-cp to improve test stability

Automaat:
## Motivation

From time to time we were experiencing issues in upgrade test because we were starting port forward to wrong pod of kuma-cp, pod that was already terminating. This fix should mitigate this issue

## Implementation information

port forward to service instead of pod

### DIFF
--- a/test/framework/k8s_controlplane.go
+++ b/test/framework/k8s_controlplane.go
@@ -81,7 +81,7 @@ func (c *K8sControlPlane) GetKubectlOptions(namespace ...string) *k8s.KubectlOpt
 func (c *K8sControlPlane) PortForwardKumaCP() {
 	kumaCpSvc := c.GetKumaCPSvc()
 	if k8s.IsServiceAvailable(&kumaCpSvc) {
-		k8s.NewTunnel(c.GetKubectlOptions(Config.KumaNamespace), k8s.ResourceTypePod, kumaCpSvc.Name, 0, 5681)
+		c.portFwd.apiServerTunnel = k8s.NewTunnel(c.GetKubectlOptions(Config.KumaNamespace), k8s.ResourceTypePod, kumaCpSvc.Name, 0, 5681)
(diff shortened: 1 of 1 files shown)


test/framework/k8s_controlplane.go
```diff
@@ -79,21 +79,19 @@ func (c *K8sControlPlane) GetKubectlOptions(namespace ...string) *k8s.KubectlOpt
...
-			c.portFwd.ApiServerEndpoint = c.portFwd.apiServerTunnel.Endpoint()
-
-			c.madsFwd.apiServerTunnel = k8s.NewTunnel(c.GetKubectlOptions(Config.KumaNamespace), k8s.ResourceTypePod, kumaCpPods[i].Name, 0, 5676)
-			c.madsFwd.apiServerTunnel.ForwardPort(c.t)
-			c.madsFwd.ApiServerEndpoint = c.madsFwd.apiServerTunnel.Endpoint()
-			return
-		}
+	kumaCpSvc := c.GetKumaCPSvc()
+	if k8s.IsServiceAvailable(&kumaCpSvc) {
+		c.portFwd.apiServerTunnel = k8s.NewTunnel(c.GetKubectlOptions(Config.KumaNamespace), k8s.ResourceTypePod, kumaCpSvc.Name, 0, 5681)
```
michaelbeaumont, line 84: something like

```suggestion
		c.portFwd.apiServerTunnel = k8s.NewTunnel(c.GetKubectlOptions(Config.KumaNamespace), k8s.ResourceTypeService, kumaCpSvc.Name, 0, 5681)
```

right?

Automaat: right, I was looking at it but I forgot to change it in a hurry

slonka (approved):

lukidzi (approved):

PR #11746, opened 2024-10-10
test(framework): inspect clusters, stats and policies when DebugKube on E2E failures

jijiechen:
## Motivation

Also dump the clusters/stats and policies in the `DebugKube` helper function when an E2E test case fail. This is a supplementary feature to the existing helper.

## Implementation information

Invoking the `kumactl` command line, nothing fancy.

## Supporting documentation

N/A

> Changelog: skip


test/framework/debug.go
```diff
@@ -186,61 +186,10 @@ func DebugKube(cluster Cluster, mesh string, namespaces ...string) {
 
 	Logf("printing debug information of cluster %q for mesh %q and namespaces %q", cluster.Name(), mesh, namespaces)
 	for _, namespace := range namespaces {
-		nsDir := getNsDirPath(debugPath, cluster.Name(), namespace)
```
jijiechen, line 189: The code here is extracted to function `debugKubeNamespace` underneath.

test/framework/debug.go
```diff
@@ -186,61 +186,10 @@ func DebugKube(cluster Cluster, mesh string, namespaces ...string) {
...
-		out, err := k8s.RunKubectlAndGetOutputE(cluster.GetTesting(), &kubeOptions, "get", "all,kuma", "-oyaml")
-		if err != nil {
-			out = fmt.Sprintf("kubectl get for namespace %s failed with error: %s", namespace, err.Error())
+		nsErr := debugKubeNamespace(cluster, namespace, debugPath)
+		if nsErr {
 			errorSeen = true
 		}
-
-		// Ignore it if we don't have Gateway API resources installed
-		gatewayAPIOut, err := k8s.RunKubectlAndGetOutputE(cluster.GetTesting(), &kubeOptions, "get", "gateway-api", "-oyaml")
```
jijiechen, line 201: Same as above, code here is extracted to function `debugKubeNamespace` underneath.

test/framework/debug.go
```diff
@@ -291,64 +301,97 @@ func configDump(kumactlOpts kumactl.KumactlOptions, debugPath string, cluster Cl
...
-				if dpNS == "" {
-					continue
-				}
-				if !namespaceExported(debugPath, cluster.Name(), dpNS) {
-					continue
-				}
+		}
+	}
 
-				configDumpResp := ""
```
jijiechen, line 313: Code here is extracted into the new function `doInspect` underneath. In that function, I implementd the ability to inspect different types info.

slonka (commented):
Shouldn't this be aimed at release-2.9 branch?

slonka:
In general looks good, nice refactoring as well. Just let me know if we should point this at 2.9 branch so we get this for this release.

jijiechen:
Yeah, I think we can make it target `release-2.9`. Otherwise, we'll need to backport in the future.

test/framework/debug.go
```diff
@@ -272,8 +283,7 @@ const (
 	zoneingressType dpType = "zoneingress"
 )
 
-func configDump(kumactlOpts kumactl.KumactlOptions, debugPath string, cluster Cluster, mesh string, dpType dpType) {
-	errorSeen := false
+func inspectDataplane(kumactlOpts *kumactl.KumactlOptions, debugPath string, cluster Cluster, mesh string, dpType dpType) {
```
jakubdyszkiewicz, line 286: question. Could we have the same implementation for inspecting for both universal and kube? It should be the same in the end, right?

jakubdyszkiewicz, line 286: question. Could we have the same implementation for inspecting for both universal and kube? It should be the same in the end, right?

jakubdyszkiewicz (commented):
this is great. We really need it

test/framework/debug.go
jijiechen: Yeah, I also guess so. Let me try to refactor and consolidate the logic afterwards.

slonka (approved):

test/framework/debug.go
```diff
@@ -291,64 +301,84 @@ func configDump(kumactlOpts kumactl.KumactlOptions, debugPath string, cluster Cl
...
+		return ""
+	}
+
+	var err error
+	inspectResp := ""
+
+	switch dpType {
+	case dataplaneType:
+		inspectResp, err = kumactlOpts.RunKumactlAndGetOutput("inspect", string(dataplaneType), dpName, "--mesh", mesh, "--type", inspectType)
+	case zoneegressType:
```
jakubdyszkiewicz, line 359: why Egress is missing?

jakubdyszkiewicz (approved):
ok, we can merge it so it can give us more info, but please - let's consolidate this logic soon between universal and kubernetes

jijiechen:
@jakubdyszkiewicz Actually there was an issue with the switch/case on line 359. I just fixed it. Thanks.

Issue #11077, opened 2024-08-09
AutoReachable services with MeshExternalService


slonka:
### Description

we scoped down https://github.com/kumahq/kuma/issues/10602 to reachable services because of https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/056-meshexternalservice-and-zoneegress.md#include-default-allow-rbac-configuration-for-meshexternalservice

we still need support for auto-rechable services after figuring out https://github.com/kumahq/kuma/issues/10868

PR #11747, opened 2024-10-10
fix(autoreachableservices): do not filter out MeshMultiZoneService

lukidzi:
## Motivation

When auto reachable services are enabled, we had no MeshMultiZoneServices.

## Implementation information

Support for MeshMultiZoneService for auto reachable services is not there and we shouldn't filter out MeshMultiZoneService.

Auto reachable services therefore works this way:

- For MeshService it works
- For MeshExternalService and MeshMultiZoneService are just ignored and therefore always present regardless of MTP

## Supporting documentation

Reported during tests
Proper fix to support auto reachable services tracked in:
- https://github.com/kumahq/kuma/issues/11748
- https://github.com/kumahq/kuma/issues/11077


Automaat (approved):

lahabana:
Is the behaviour then that when using auto reachable services it doesn't work for MeshService and MeshExternalService?
In other words auto-reachable services has no impact on MS and MES?

lukidzi:
Autoreachable services works for MeshService, doesn't work for MeshExternalService (policy support) and doesn't work for MeshMultiZoneService

PR #11750, opened 2024-10-10
fix(kuma-cp): don't allow namespace-scoped policies with 'to' and 'from' arrays at the same time

lobkovilya:
## Motivation

That what MADR says https://github.com/kumahq/kuma/blob/180976f81dcb6a9b86c4c3fa534e11534060171d/docs/madr/decisions/052-policy-on-namespace.md?plain=1#L155

## Implementation information

Return error when policy has both `from` and `to` arrays.


lahabana (approved):

lahabana (changes requested):
Isn't this going to disallow all to and from policies? Even existing ones? Seems like a breaking change no?

Should this be enforced only when there are "producer" entries.

lobkovilya:
You're right, the expectation was `ComputePolicyRole` is called only for policies on custom namespace. But as I learned it wasn't always the case and overall that's a sign of a poor design to have such expectations from the function. 

I updated the code so that now `ComputePolicyRole` is called for all types policies (both system and custom).

pkg/core/resources/model/resource.go
```diff
@@ -458,15 +458,42 @@ func resourceOrigin(labels map[string]string) (mesh_proto.ResourceOrigin, bool)
...
+func NewNamespace(value string, system bool) Namespace {
+	return Namespace{
+		value:  value,
+		system: system,
+	}
+}
+
+func UnknownNamespace() Namespace {
+	return Namespace{}
+}
```
lahabana, line 477: Should the UnknownNamespace just be a constant? Makes it either to just check `if namespace == UnknownNamespace`

pkg/core/resources/model/resource.go
```diff
@@ -507,32 +534,27 @@ func ComputeLabels(
 		}
 	}
 
-	if isK8s && IsLocallyOriginated(mode, labels) {
-		ns, ok := resourceNameExtensions[mesh_proto.KubeNamespaceTag]
-		if ok && ns != "" {
-			setIfNotExist(mesh_proto.KubeNamespaceTag, ns)
-		}
+	if ns.value != "" && isK8s && IsLocallyOriginated(mode, labels) {
```
lahabana, line 537: There's some magic here that ns == UnknownNamespace would be easier to understand

pkg/core/resources/model/resource.go
lahabana: It's also probably Undefined not Unknown. To me `Unkown` could be something that exists but we don't know what it is. Whereas `Undefined` just doesn't exist.

Btw in any case this and the method after deserve a little comment

lahabana: `Unset` is another option that might be even better

pkg/core/resources/model/resource_test.go
```diff
@@ -259,8 +259,22 @@ var _ = Describe("ComputePolicyRole", func() {
...
+				AddTo(builders.TargetRefMesh(), meshtimeout_api.Conf{
+					IdleTimeout: &kube_meta.Duration{Duration: 123 * time.Second},
+				}).
+				AddFrom(builders.TargetRefMesh(), meshtimeout_api.Conf{
+					IdleTimeout: &kube_meta.Duration{Duration: 123 * time.Second},
+				}).
+				Build().Spec,
+			namespace:    core_model.NewNamespace("kuma-system", true),
+			expectedRole: mesh_proto.SystemPolicyRole,
 		}),
```
lahabana, line 278: So consumer policies can't mix both? Are we sure this isn't an issue? I'd love to see more tests with legacy stuff to make sure we're not breaking backwards compat.

For .e.g I'd expect the following to work:

```yaml
kind: MeshTimeout
namespace: kuma-system
spec:
  to:
   - targetRef:
        kind: MeshService
        name: old_ns_svc_2030
  from:
   - targetRef:
        kind: MeshService
        name: old_ns_svc_2030
```

Sure from reading the code it looks like it works. But we all know we write bugs so a test would be nice!

pkg/core/resources/model/resource.go
lobkovilya: Right, it's better as a constant, also changed the name to `UnsetNamespace`

pkg/core/resources/model/resource.go
lobkovilya: this line is fine, there is another line below `if ns.system || ns.value == "" {` that I changed now to `if ns.system || ns == UnsetNamespace`

pkg/core/resources/model/resource_test.go
lobkovilya: The case you mentioned was already there:
```go
Entry("system policy with both from and to", testCase{
	policy: builders.MeshTimeout().
		WithMesh("mesh-1").WithName("name-1").
		WithTargetRef(builders.TargetRefMesh()).
		AddTo(builders.TargetRefMesh(), meshtimeout_api.Conf{
			IdleTimeout: &kube_meta.Duration{Duration: 123 * time.Second},
		}).
		AddFrom(builders.TargetRefMesh(), meshtimeout_api.Conf{
			IdleTimeout: &kube_meta.Duration{Duration: 123 * time.Second},
		}).
		Build().Spec,
	namespace:    core_model.NewNamespace("kuma-system", true),
	expectedRole: mesh_proto.SystemPolicyRole,
}),
```
added 2 other cases that were not covered

lahabana (approved):

Issue #11759, opened 2024-10-11
Document k8s/uni debugging


slonka:
### Description

We've done a knowledge share session on k8s/uni debugging, let's document the steps to trigger a breakpoint in both of these.

PR #11760, opened 2024-10-11
docs(debug): add debugging on k8s and universal docs

slonka:
## Motivation

We had a knowledge session around this so let's have this in docs so other people can use it.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/11759

> Changelog: skip


lahabana (approved):
Very nice! I almost want make targets that do this but that's for sure a lot more work than just this doc :)

jakubdyszkiewicz (approved):

PR #11764, opened 2024-10-11
fix(e2e): loosen up assertion on traffic route test

Automaat:
## Motivation

We have seen flake on master recently:

Screenshot 2024-10-11 at 09 31 42

I think we could loosen up the assertions to make it more stable


lahabana:
I agree that this test is flaky. However, now things are so loose that it almost feels useless to assert right?

Automaat:
@lahabana that makes sense, updated

jijiechen (approved):

Issue #11765, opened 2024-10-11
whilst upgrading zones labels don't seem to be applied properly


johncowen:
### What happened?

A bug happened!

michaelbeaumont:
Which labels, just Dataplane labels?

lukidzi:
I think it should not be a problem, we did one fix: https://github.com/kumahq/kuma/pull/11766, also https://github.com/kumahq/kuma/pull/11758 and https://github.com/kumahq/kuma/pull/11794

PR #11766, opened 2024-10-11
fix(api): when resource has origin zone assume is local

lukidzi:
## Motivation

When upgrading from 2.7/2.8 to 2.9, not all resources are reconciled, causing them to miss the `kuma.io/zone` label. As a result, they might not correctly match with the `Dataplanes`, which is incorrect behavior. If a resource has an origin zone, it means it was created in that zone, as we don’t sync policies other than the producer ones (starting from 2.9). This caused resources created in 2.7/2.8 to not match with `Dataplanes` after the upgrade.

## Implementation information

If the resource has no `kuma.io/zone` label and `kuma.io/origin: zone` we assume it's created locally and match it.

## Supporting documentation

part of https://github.com/kumahq/kuma/issues/11765


pkg/plugins/policies/core/matchers/dataplane.go
```diff
@@ -229,7 +230,10 @@ func dppSelectedByZone(policyMeta core_model.ResourceMeta, dpp *core_mesh.Datapl
 		policyOrigin, ok := policyMeta.GetLabels()[mesh_proto.ResourceOriginLabel]
 		if ok && policyOrigin == string(mesh_proto.ZoneResourceOrigin) {
 			zone, ok := policyMeta.GetLabels()[mesh_proto.ZoneTag]
-			return ok && meta.GetLabels()[mesh_proto.ZoneTag] == zone
+			if !ok {
+				return true
```
Icarus9913, line 234: Could you add some comments here to explain this case?

pkg/plugins/policies/core/matchers/testdata/matchedpolicies/meshgateways/policies_from_different_zones_no_zone_label.policies.yaml
```diff
@@ -0,0 +1,89 @@
...
+        - type: File
+          file:
+            path: /to-gateway
+---
+type: MeshGateway
+mesh: mesh-1
+name: gateway-in-zone-1
+labels:
+  kuma.io/origin: zone
+  kuma.io/zone: zone-1
```
Icarus9913, line 72: How about removing this `kuma.io/zone` label to validate the updated `filterGatewaysByZone` function?

Icarus9913 (approved):

slonka (approved):
Makes sense to me but I'd ask the original author to review

pkg/plugins/policies/core/matchers/dataplane.go
```diff
@@ -125,14 +125,15 @@ func filterGatewaysByZone(gateways []*core_mesh.MeshGatewayResource, dpp *core_m
...
-	dppZone := dpp.GetMeta().GetLabels()[mesh_proto.ZoneTag]
+	dppZone, dppZoneOk := dpp.GetMeta().GetLabels()[mesh_proto.ZoneTag]
 	for _, gateway := range gateways {
 		gwOrigin, ok := gateway.GetMeta().GetLabels()[mesh_proto.ResourceOriginLabel]
 		if !ok || gwOrigin == string(mesh_proto.GlobalResourceOrigin) {
 			filtered = append(filtered, gateway)
 			continue
 		}
-		if gwZone, ok := gateway.GetMeta().GetLabels()[mesh_proto.ZoneTag]; ok && gwZone == dppZone {
+		gwZone, gwZoneOk := gateway.GetMeta().GetLabels()[mesh_proto.ZoneTag]
```
michaelbeaumont, line 135: Should we have some kind of helper method on Meta or Labels? I find it hard to know what the semantics are of missing zone tag here. Like why is the check on gwZoneOk now inverted?

lukidzi: We inverted it since resources created in 2.7 only have the `kuma.io/origin` label but don’t have the `kuma.io/zone` label. This causes the object to not be recognized as a zone object. I think we could introduce a helper like `IsLocallyOriginated` to address this.

michaelbeaumont (approved):

PR #11791, opened 2024-10-17
test(e2e): check higher number of procs

jakubdyszkiewicz:
## Motivation

Same as https://github.com/kumahq/kuma/pull/11716 but this time not from fork so we can check our runners

## Supporting documentation

Fix #XX


jakubdyszkiewicz:
closing due to no activity and it's very easy to reconstruct. I don't have time to test it more.

PR #11794, opened 2024-10-17
fix(kuma-cp): resources that were created on 2.7.x are missing namespace labels when synced on global

lobkovilya:
## Motivation

In 2.7.x method `GetMeta().GetLabels()` on Kubernetes returned a label map with `k8s.kuma.io/namespace` added dynamically. This behaviour was changed in 2.9.x by https://github.com/kumahq/kuma/pull/11020, the namespace label is now supposed to be set in ComputeLabels function. But this functions is called only on Create/Update of the resources. This means policies that were created on 2.7.x won't have `k8s.kuma.io/namespace` label when synced to Global. Even though the lack of namespace labels affects only how resource looks in GUI on Global it's still worth setting it.

## Implementation information

Set `k8s.kuma.io/namespace` label if not exist to the resources before syncing them to Global.

## Supporting documentation

Fix #XX

### DIFF
--- a/pkg/kds/util/meta.go
+++ b/pkg/kds/util/meta.go
@@ -6,15 +6,17 @@ import (
 	"golang.org/x/exp/maps"
 
 	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
+	config_store "github.com/kumahq/kuma/pkg/config/core/resources/store"
 	"github.com/kumahq/kuma/pkg/core/resources/model"
 )
 
 // KDS ResourceMeta only contains name and mesh.
 // The rest is managed by the receiver of resources anyways. See ResourceSyncer#Sync
 type resourceMeta struct {
-	name   string
-	mesh   string
-	labels map[string]string
+	name           string
+	mesh           string
+	labels         map[string]string
+	nameExtensions model.ResourceNameExtensions
 }
 
 type CloneResourceMetaOpt func(*resourceMeta)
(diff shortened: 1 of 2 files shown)


michaelbeaumont (approved):

pkg/kds/util/meta.go
```diff
@@ -34,21 +36,57 @@ func WithLabel(key, value string) CloneResourceMetaOpt {
...
 }
 
+// PopulateNamespaceLabelFromNameExtension on Kubernetes zones adds 'k8s.kuma.io/namespace' label to the resources
+// before syncing them to Global.
+//
+// In 2.7.x method 'GetMeta().GetLabels()' on Kubernetes returned a label map with 'k8s.kuma.io/namespace' added
+// dynamically. This behavior was changed in 2.9.x by https://github.com/kumahq/kuma/pull/11020, the namespace label is now
+// supposed to be set in ComputeLabels function. But this functions is called only on Create/Update of the resources.
+// This means policies that were created on 2.7.x won't have 'k8s.kuma.io/namespace' label when synced to Global.
+// Even though the lack of namespace labels affects only how resource looks in GUI on Global it's still worth setting it.
```
lahabana, line 46: There should be a cleanup task for this right?

lobkovilya: Added one https://github.com/kumahq/kuma/issues/11797, but it's not a straightforward cleanup, because even on 2.10 user can have policies created in 2.7

Issue #11806, opened 2024-10-21
Restrict MeshGateway resource tags validation


Automaat:
### What happened?

Right now, MeshGateway tags validation is more loose than validation of tags in Mesh*Route which lets you create Gateway you won't be able to reference in Mesh*Route. 

We have lots of examples of this in our docs, which resulted in this issue: https://github.com/kumahq/kuma-website/issues/1810

lukidzi:
> Triage: Let's figure out how to make the tags validation consistent on the MeshRoute and MeshGateway. We should allow using `/` in tags value

PR #11808, opened 2024-10-21
fix(gateway): change MeshGateway tags validation to be consistent with MeshRoute tags validation

Automaat:
Fixes: #11806 

## Motivation

Right now, MeshGateway tags validation is more loose than validation of tags in Mesh\*Route which lets you create Gateway you won't be able to reference in Mesh\*Route.


slonka (approved):

michaelbeaumont:
I thought the MeshGateway was already looser, what's the difference between routes and gateways?

michaelbeaumont:
Why can't we just have
- Leave MeshGateway validation, it already allows `/`
- Change MeshRoute validation, it can now select `/`
What am I missing?

michaelbeaumont (approved):
That's what this PR does :smiley:

Issue #11572, opened 2024-09-26
Relax validation on sectionName with `kind: MeshService`


lahabana:
### Description

When doing:

```yaml
kind: MeshRetry
apiVersion: kuma.io/v1alpha1
metadata:
  name: retry-demo-app-consumer-2
  namespace: kuma-demo
spec:
  targetRef:
    kind: MeshSubset
    tags:
      k8s.kuma.io/service-name: redis
  to:
    - targetRef:
        kind: MeshService
        labels:
          kuma.io/display-name: demo-app
        sectionName: "5000"
      default:
        http:
          numRetries: 127
```

You get:
> Error: YAML contains invalid resource: [spec.to](http://spec.to/)[0].targetRef.sectionName: sectionName should not be combined with labels

There's no real reason for this. If one of the matching meshService doesn't have such a sectionName it should simply not match just like it's already the case for labels.

PR #11819, opened 2024-10-21
feat(kuma-cp): allow `sectionName` and `labels` in `targetRef`

Neyaz:
## Motivation

According to https://github.com/kumahq/kuma/issues/11572 there's no real reason to have such validation.
However, I didn't get if this part from the issue `If one of the matching meshService doesn't have such a sectionName it should simply not match just like it's already the case for labels.`  already works in this way or it needs to be implemented.

## Implementation information

The change is pretty simple.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/11572


jakubdyszkiewicz (dismissed):

jakubdyszkiewicz:
Hey, thanks for the contribution.

> However, I didn't get if this part from the issue If one of the matching meshService doesn't have such a sectionName it should simply not match just like it's already the case for labels. already works in this way or it needs to be implemented.

I believe it already works that way. @lobkovilya to confirm

lobkovilya (changes requested):
Unfortunately, even though it looks like it's supported, method [resolveTargetRef](https://github.com/kumahq/kuma/blob/f153461bd6e33dad7004209969591827604b6732/pkg/plugins/policies/core/rules/resourcerules.go#L250) doesn't take SectionName into account. 

I added the following input file to `pkg/plugins/policies/core/rules/testdata/resourcerules`:
```yaml
# MeshTimeout with targetRef to the real MeshService resource
type: MeshTimeout
name: matched-for-rules-mt-1
mesh: mesh-1
spec:
  targetRef:
    kind: Mesh
  to:
    - targetRef:
        kind: MeshService
        labels:
          env: dev
        sectionName: httpport
      default:
        idleTimeout: 10s
        connectionTimeout: 11s
        http:
          requestTimeout: 12s
---
type: MeshService
name: backend-1
mesh: mesh-1
labels:
  env: dev
spec:
  selector:
    dataplaneTags:
      app: backend
  ports:
    - port: 80
      name: httpport
      targetPort: 80
      appProtocol: http
---
type: MeshService
name: backend-2
mesh: mesh-1
labels:
  env: dev
spec:
  selector:
    dataplaneTags:
      app: backend
  ports:
    - port: 80
      targetPort: 80
      appProtocol: http
---
type: MeshService
name: backend-3
mesh: mesh-1
labels:
  env: prod
spec:
  selector:
    dataplaneTags:
      app: backend
  ports:
    - port: 80
      targetPort: 80
      appProtocol: http
```
and the golden file is not correct.

Neyaz:
@jakubdyszkiewicz @lobkovilya Thank you for a review!

Am I correct in assuming that, in order to implement this, if `sectionName` is set, then in the code, instead of using [implicitSectionName](https://github.com/kumahq/kuma/blob/f153461bd6e33dad7004209969591827604b6732/pkg/plugins/policies/core/rules/resourcerules.go#L277), I should create a resolved policy rule like this
 ```
append(rv, &resolvedPolicyItem{resource: r, item: item, implicitSectionName: item.GetTargetRef().SectionName})
```
?

Neyaz:
@jakubdyszkiewicz @lobkovilya Thank you for a review!

> Unfortunately, even though it looks like it's supported, method [resolveTargetRef](https://github.com/kumahq/kuma/blob/f153461bd6e33dad7004209969591827604b6732/pkg/plugins/policies/core/rules/resourcerules.go#L250) doesn't take SectionName into account.

@lobkovilya I believe it should be handled by [sectionName](https://github.com/kumahq/kuma/blob/f153461bd6e33dad7004209969591827604b6732/pkg/plugins/policies/core/rules/resourcerules.go#L243) method. As I see, it should return refSectionName for every `resolvedPolicyItem` if `sectionName` is set.

lobkovilya:
TBH I think the problem is in `resolveTargetRef`, it never calls `sectionName()` method and operates only with `implicitSectionName` (should also take into account the explicit one)

lahabana:
@lobkovilya what's the path forward on this PR?

lobkovilya:
I added support for Labels with SectionName in https://github.com/kumahq/kuma/pull/12743, once it's merged this PR is good to go

lahabana:
@lobkovilya should we try to push this in 2.10?

lobkovilya:
Yes, I merged master, let me test if it works as expected

lobkovilya (approved):

lukidzi (approved):

PR #11821, opened 2024-10-22
fix(api): return 499 when client cancel context

lukidzi:
## Motivation

A client requesting a resource from the service might close the connection, canceling the context. In this case, the API throws a 5xx server error, but this is more of a client-side issue.

## Implementation information

I used status code 499 as the response code for this situation, which is commonly used in NGINX to indicate that the client closed the connection before the server could send a response.

## Supporting documentation

https://github.com/pocketbase/pocketbase/discussions/834

Fix https://github.com/Kong/kong-mesh/issues/6468


slonka (approved):
I thought that we already had this 🤔

PR #11828, opened 2024-10-22
ci(github): add workflow run notice with version

michaelbeaumont:
## Motivation

Much easier to grab the exact docker image version


slonka (approved):
I remember some issue about some other annotations that we should add to images, but can't find it now 🤔 .

jakubdyszkiewicz (approved):

PR #11832, opened 2024-10-22
test(e2e): use MeshService mode and generated hostname

lukidzi:
## Motivation

I've noticed a flake of the test

## Implementation information

I switched to using MeshService mode instead of manually creating a MeshService. I'm not sure if MeshService might be removed since the Mesh has `meshServices.mode` disabled. Additionally, I changed the hostname to use generated one.

## Supporting documentation

Fix #XX

### DIFF
--- a/test/e2e_env/universal/meshhealthcheck/policy.go
+++ b/test/e2e_env/universal/meshhealthcheck/policy.go
@@ -8,7 +8,9 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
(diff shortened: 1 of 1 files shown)


test/e2e_env/universal/meshhealthcheck/policy.go
```diff
@@ -175,7 +163,7 @@ spec:
 			// check that test-server is healthy
 			Eventually(func(g Gomega) {
 				stdout, _, err := client.CollectResponse(
-					universal.Cluster, "dp-demo-client", "test-server.mesh/content",
+					universal.Cluster, "dp-demo-client", "test-server.universal.ms/content",
```
michaelbeaumont, line 166: Probably we could also use the auto created HostnameGenerators here?

michaelbeaumont (approved):

Issue #11784, opened 2024-10-16
Ignore converting MeshServices


jakubdyszkiewicz:
### Description

https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/041-meshservice-kubernetes-ux.md#decision-outcome

`The Service hasn't opted out via kuma.io/ignore: "true"`

We did not implement it. We should do it

lukidzi:
> Triage: Let's get it

PR #11833, opened 2024-10-23
feat(meshservices): opt-in ignore converting

jakubdyszkiewicz:
## Motivation

We need to have a way to opt out one service

## Implementation information

Support existing kuma.io/ignore annotation.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/11784


lukidzi (approved):
One question: Since we automatically generate MeshService from Dataplane, do we need or have the option to opt-out for universal environments?

jakubdyszkiewicz:
I don't think so. Here the use case is that you label namespace with sidecar injection so all services are converted to mesh service automatically, but if you have one deployment with sidecar-injection: false you end up with a MeshService that points to non-existing dataplane.

PR #11847, opened 2024-10-24
feat(kuma-cp): add option to disable KDS traces

michaelbeaumont:
## Motivation

These traces are potentially not useful and too long-lived.

## Implementation information

Added option to leave the status quo or completely disable KDS traces. Note that Envoy admin operations should not be affected.

### DIFF
--- a/pkg/config/app/kuma-cp/kuma-cp.defaults.yaml
+++ b/pkg/config/app/kuma-cp/kuma-cp.defaults.yaml
@@ -516,6 +516,10 @@ multizone:
       # Response backoff is a time Global CP waits before sending ACK/NACK.
       # This is a way to slow down Zone CP from sending resources too often.
       responseBackoff: 0s # ENV: KUMA_MULTIZONE_GLOBAL_KDS_RESPONSE_BACKOFF
+      tracing:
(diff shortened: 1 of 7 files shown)


pkg/config/app/kuma-cp/kuma-cp.defaults.yaml
```diff
@@ -516,6 +516,10 @@ multizone:
       # Response backoff is a time Global CP waits before sending ACK/NACK.
       # This is a way to slow down Zone CP from sending resources too often.
       responseBackoff: 0s # ENV: KUMA_MULTIZONE_GLOBAL_KDS_RESPONSE_BACKOFF
+      tracing:
```
slonka, line 519: shouldn't this be somewhere under just `tracing`?

slonka (approved):
Not sure about config placement, otherwise LGTM

pkg/config/app/kuma-cp/kuma-cp.defaults.yaml
michaelbeaumont: IMO no, this concerns only KDS. Where as tracing is global

slonka (approved):

PR #11850, opened 2024-10-24
feat(meshservices): api for listing matching DPPs

### LINKED ISSUE #10244: Add endpoint to get the list of dataplanes matching a meshService
### Description

We need an endpoint to easily get all the dataplanes for a meshservice (and maybe for a globalService too).

I believe the most likely path is:

`/meshes/{mesh}/meshservices/{serviceName}/_resources/dataplanes`

As we already to this for policies:

https://github.com/kumahq/kuma/blob/55fa983fc12bdd1a5ea7c1cfe75901320870372b/api/openapi/specs/api.yaml#L109

jakubdyszkiewicz:
## Motivation

We want to list matching DPPs in the UI

## Implementation information

We already had a code for it for policies. In the case of MeshServices, the matching is different.
I extracted "boilerplate" of parsing, marshaling request and validating the access to "generic function", so won't duplicate the logic and whether we have a change it matches both.

When it comes to OAPI, we already have a definition for it
https://github.com/kumahq/kuma/blob/2adb7f53026f3ab38f5967b21f507e6674ce893d/api/openapi/specs/api.yaml#L122
not sure we can do anything more here

## Supporting documentation

Fix #10244

### DIFF
--- a/pkg/api-server/inspect_mesh_service.go
+++ b/pkg/api-server/inspect_mesh_service.go
@@ -1,0 +1,45 @@
+package api_server
+
+import (
+	"github.com/emicklei/go-restful/v3"
+
+	"github.com/kumahq/kuma/pkg/core/resources/access"
+	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
+	"github.com/kumahq/kuma/pkg/core/resources/apis/meshservice"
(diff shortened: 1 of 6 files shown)


michaelbeaumont (approved):

lahabana:
@jakubdyszkiewicz actually this is only meant to work for policies (the doc mentions policyType, MeshService isn't a policy)

Issue #11071, opened 2024-08-08
[MeshExternalService] Possibility to route traffic through specific zone


lukidzi:
### Description

> Few of our cloud applications (mesh enabled) want to communicate with the apps running in Datacenter (non-mesh apps).
Those DC apps are not exposed to the internet. So we installed ZoneCP in the DC. Now those DC applications can be resolved from the Kuma Egress running in the DC.
In this case we can register the DC apps as external services with zone tag, same as the DC zone name.
As per the point number 1, traffic originating from other zones will be routed to DC zone first. Once the traffic is inside DC, the egress can resolve the DC apps.
This is just one possibility/example. We have few more scenarios as well

andrey-dubnik:
We have an upcoming use case where our workloads would go under multiple private networks where within each network services could be both mesh-aware and external to the mesh. We plan registering the external to mesh services as external service with the zone binding as this way mesh knows exactly which zone egress can access the specific "private" service.

> Ideally we would like to be able to bind external service to multiple zones as there can be a scenario where multiple mesh zones are placed within a network and only few can access specific endpoint.

If binding the external service to zone feature is removed from future release of Kuma we won't be able to use Kuma much longer when the described use case becomes a majority of our workload placement use cases.

slonka:
> triage: this would probably be done by defining MeshExternalService in that zone

AyushSenapati:
Why are we willing to define resources at the zone level? We have GlobalCP. So, why can't we simply create the resource in the GlobalCP and let it decide the ZoneCPs it needs the resources to be forwarded?

slonka:
@AyushSenapati - let's chat about the details in the MADR - is that ok?

slonka:
Tagging @justindavies since it's a bigger issue and probably will be converted to aha

PR #11853, opened 2024-10-24
docs(MADR): meshexternalservice routed through the specific zone

lukidzi:
## Motivation

There is a use case where a service is not part of the mesh and is only accessible or resolvable within a single Datacenter (DC). Without the proposed functionality, there is no way to expose this service to other zones.

## Implementation information

MADR https://docs.google.com/document/d/12YrUy-kV3JZu9K4tSv6V6q4Dx8v7mp41_SdxJ0kQ_tc/edit?tab=t.0#heading=h.n6cmlf1eel2z

## Supporting documentation

part of https://github.com/kumahq/kuma/issues/11071


docs/madr/decisions/062-meshexternalservice-through-zone.md
```diff
@@ -0,0 +1,122 @@
+# MeshExternalService Accessible Only Through a Specific Zone
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/11071
+
+## Context and Problem Statement
+
+In certain cloud environments, mesh-enabled applications need to communicate with applications running in a Datacenter (DC) that are not part of the mesh. These DC applications are not exposed to the internet. To address this, we installed a ZoneCP in the DC. This setup allows DC applications to be resolved from the Kuma Egress running in the DC.
```
jijiechen, line 9: These sentences are from users, right? It's better to convert the way to be from our (Kuma authors') perspective?

lukidzi: Changed a bit

jijiechen:
I would prefer creating the resource on global CP with a label.

Because when we ask the user to create a resource in the zone, it implies the resource is "owned/maintained" by the zone. But this is not the case here. 

Zones are phsically located at different clusters/places, so they may have very different permissions when considering the underlying Kubernetes/VM cluster. However, an external service that is only accessible through a zone does not belong to the zone. When the mesh admin/operator decides to open this accessiblity to other zones in the mesh,  the responsibility of managing/maintaining this relationship with the external service should be shared by the whole mesh. So  the `MeshExternalService` should be shown/managed through the global CP.

Let me raise an example scenario:

When the external service fails because of the IP address had been changed by it owner and the consumers within the mesh just found it. So they want to correct the address in `MeshExternalService` defined in the mesh. 

In this case, I think they should go to the global CP admin, instead of the zone admin.

It's slightly different than the `MeshService` case. MeshServices are normally generated automatically and it originates from the Service objects on a Kubernetes cluster. 

So it's reaonable for MeshServices to be "owned" by zones, because:
1. They originate from Services, which are owned by zones
2. People don't need to manage/maintain MeshServices directly, so the operational cost is low

lukidzi:
That makes sense I can agree with most of this. I have another use case that could involve a dedicated database team maintaining a cluster independently. In this scenario, permissions for exposing services from other zones would be managed by the Mesh Operator, while the database team could handle the creation and maintenance of `MeshExternalService` resources. This setup would provide flexibility, allowing the database team to manage their resources without impacting broader mesh configurations.

lukidzi:
I am going to move it to the google docs

slonka (commented):
@lukidzi I feel like there is nothing to add more from my side but there are some discussions that are not closed. Should we create a meeting and talk through them?

jijiechen (approved):

lukidzi:
I am going to reopen it once we plan to work on this

Issue #10618, opened 2024-06-24
Ability to see VIPs/Hostname on global CP


slonka:
### Description

Let's do this on demand, if you go to MeshService/MeshExternalService the global will ask zones for VIPs/Hostnames and then we can show that in the GUI

jakubdyszkiewicz:
I'd consider the generic `GetStatusOfResource` method in KDS Services.

GUI pings Global CP API, Global CP execute KDS method for each zone and collects the data

PR #11865, opened 2024-10-25
feat(inspect): api for retrieving service hostnames

jakubdyszkiewicz:
## Motivation

We want to see hostnames of Mesh*Service in the Global CP.

## Implementation information

While we could add new KDSService to retrieve stats of a resource in specific zones and then aggregate this an easier (less expensive to execute, faster, don't depend on the zone) option is to calculate the hostname. We have all the resources in global CP already (because HostnameGenerator from zone is synced to global for visibility). It's just a matter of relabeling properly.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/10618


api/openapi/specs/api.yaml
```diff
@@ -172,6 +172,41 @@ paths:
           $ref: '#/components/responses/BadRequest'
         '500':
           $ref: '#/components/responses/InternalServerError'
+  /meshes/{mesh}/{serviceType}/{serviceName}/_resources/hostnames:
```
jakubdyszkiewicz, line 175: I'm not sure about this path. What `_resources` stand for?

@lahabana any input here?

lahabana: _resources are meant to map to other resources. For example you get /_resources/dataplanes or /_resources/policies here this doesn't look right.

IMO what you want here is: `/meshes/{mesh}/{serviceType}/{serviceName}/_hostnames`

api/openapi/specs/api.yaml
```diff
@@ -172,6 +172,41 @@ paths:
...
+        - in: path
+          name: mesh
+          example: default
+          schema:
+            type: string
+          required: true
+          description: The mesh the service is part of
+        - in: path
+          name: serviceType
+          example: meshservices
```
lahabana, line 191: You should be able to add validation with just the list of supported values.

api/openapi/specs/api.yaml
```diff
@@ -433,6 +468,45 @@ components:
...
+          type: integer
+          example: 200
+        items:
+          type: array
+          items:
+            $ref: '#/components/schemas/InspectHostname'
+    InspectHostname:
+      type: object
+      title: InspectHostname
+      description: A list of zones in which the hostname is available in
```
lahabana, line 487: This description seems to be the one of the `zones` field not of this entire object.
I think this is more
```suggestion
      description: An supported hostname along with the zones it exists in
```

api/openapi/specs/api.yaml
```diff
@@ -433,6 +468,45 @@ components:
...
+          description: Generated hostname
+          example: redis.redis-system.svc.east.mesh.local
+        zones:
+          type: array
+          items:
+            $ref: '#/components/schemas/InspectHostnameZone'
+    InspectHostnameZone:
+      type: object
+      title: InspectHostnameZone
+      description: A name of the zone with hostname generator that was used to generate the hostname
```
lahabana, line 501: I don't understand this sentence.

pkg/api-server/inspect_mesh_service.go
```diff
@@ -43,3 +67,109 @@ func matchingDataplanesForMeshServices(resManager manager.ResourceManager, resou
...
+		svcMesh := request.PathParameter("mesh")
+		svcType := request.PathParameter("serviceType")
+
+		desc, ok := typeDescForType[svcType]
+		if !ok {
+			rest_errors.HandleError(
+				request.Request.Context(),
+				response,
+				&validators.ValidationError{},
+				fmt.Sprintf("only %s are available for inspection", strings.Join(util_maps.SortedKeys(typeDescForType), ",")),
```
lahabana, line 94:
```suggestion
				fmt.Sprintf("only %q are available for inspection", strings.Join(util_maps.SortedKeys(typeDescForType), ",")),
```

api/openapi/specs/api.yaml
jakubdyszkiewicz: I don't know how to describe a structure with zone and hostname generator name :(
changed with to and, but not sure if it helps. Any suggestions?

jakubdyszkiewicz: we talked offline and decided to just remove hostname generator core name from this object

api/openapi/specs/api.yaml
```diff
@@ -172,6 +172,42 @@ paths:
...
+        - in: path
+          name: serviceType
+          example: meshservices
+          schema:
+            type: string
+            enum: [meshservices, meshmultizoneservices, meshexternalservices]
+          required: true
+          description: The type of the service
+        - in: path
+          name: serviceName
```
lahabana, line 198: We agree that this doesn't support URIs yet?

jakubdyszkiewicz: URIs?

lahabana: unique resource identifier (`<mesh>_<name>_<namespace>_<zone>_<shortType>_<optionalSection>`)

jakubdyszkiewicz: then no. At least not for now

pkg/api-server/inspect_mesh_service.go
```diff
@@ -43,3 +67,114 @@ func matchingDataplanesForMeshServices(resManager manager.ResourceManager, resou
...
+				request.Request.Context(),
+				response,
+				&validators.ValidationError{},
+				fmt.Sprintf("only %q are available for inspection", strings.Join(availableServiceTypes, ",")),
+			)
+		}
+
+		svc := desc.NewObject()
+		if err := resManager.Get(request.Request.Context(), svc, store.GetByKey(svcName, svcMesh)); err != nil {
+			rest_errors.HandleError(request.Request.Context(), response, err, "could not retrieve service")
```
slonka, line 106: I assume this triggers 404 and you don't define 404 in the OAPI spec.

pkg/api-server/testdata/resources/inspect/services/_resources/hostnames/meshservice_hostnames_zone_synced.input.yaml
```diff
@@ -0,0 +1,33 @@
+#/meshes/default/meshservices/test-server-syncedhash/_hostnames 200
```
slonka, line 1: missing an error case

slonka (commented):
In general ok, but missing some bits that would make it better.

api/openapi/specs/api.yaml
```diff
@@ -566,6 +568,12 @@ components:
         application/problem+json:
           schema:
             $ref: './common/error_schema.yaml#/components/schemas/Error'
+    NotFound:
```
slonka, line 571: this probably should be in common like here https://github.com/kumahq/kuma/pull/12497/files#diff-b29f834d002dd609516e6d1f777ba0733baaf8e20c70e3aebde1c412c2c3c1d2R72

slonka (approved):
LGTM - one nit

slonka (approved):

PR #11884, opened 2024-10-28
test(e2e): improve reachable services expectation to allow DNS timeouts

michaelbeaumont:
## Motivation

We suspect that DNS times out initially which leads to exit code 6.


michaelbeaumont:
See https://github.com/kumahq/kuma/pull/11894

PR #11893, opened 2024-10-28
test(e2e): speed up reachable services

jakubdyszkiewicz:
## Motivation

Let's try to speed up some non-shared e2e tests.
They take as much resources as all e2e_env combined.

Before
```
❯❯❯ python test-time.py
Test: [It] > Auto Reachable Services on Kubernetes > should not connect to non auto reachable service, Duration: 74.248 seconds
Test: [It] > Auto Reachable Mesh Services on Kubernetes > should not connect to non auto reachable service, Duration: 61.293 seconds
Test: [It] > Auto Reachable Services on Kubernetes > should connect to ExternalService when MeshTrafficPermission defined, Duration: 60.537 seconds
```
196s

```
Test: [It] > Auto Reachable Mesh Services on Kubernetes > should not connect to non auto reachable service, Duration: 24.698 seconds
Test: [It] > Auto Reachable Services on Kubernetes > should connect to ExternalService when MeshTrafficPermission defined, Duration: 23.711 seconds
Test: [It] > Auto Reachable Services on Kubernetes > should not connect to non auto reachable service, Duration: 19.414 seconds
Test: [BeforeSuite] > , Duration: 17.187 seconds
Test: [AfterSuite] > , Duration: 15.619 seconds
```
100s

## Implementation information

1) Reuse created Kuma deployment for reachable services and reachable backends
2) Adjust Consistently to more reasonable times
3) Setup deployments in parallel (gives 38s boost)

This strategy is useful, because
1) we can merge it with for example CNI to save cluster setup / cleanup.
2) potentially we can run this in parallel just like `e2e_env`
3) potentially we can later move this as a separate zone that we spawn in `e2e_env`

## Supporting documentation

Fix #XX

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -91,6 +91,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require golang.org/x/sync v0.8.0
+
 require (
 	cel.dev/expr v0.16.0 // indirect
 	dario.cat/mergo v1.0.1 // indirect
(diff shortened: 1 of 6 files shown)


slonka (approved):
seems reasonable

PR #11923, opened 2024-10-30
fix(k8s): set annotation kuma.io/display-name for Secrets

michaelbeaumont:
## Motivation

Otherwise we have issues if the Secret name is too long, already handled for other resources in https://github.com/kumahq/kuma/pull/10430
Also this internally fixes `UpdateWithLabels` not being applied for secrets.

## Implementation information

Basically do for Secrets what we do for other resources


lobkovilya:
should we also do the split for pkg/plugins/config/k8s/store.go?

michaelbeaumont:
@lobkovilya added

lobkovilya:
you also need to update `GetLabels()` on KubernetesMetaAdapter like here https://github.com/kumahq/kuma/blob/a0a30050ab6527d14b6450156ca8ad4fc5e93554/pkg/plugins/resources/k8s/store.go#L296

lobkovilya (approved):

PR #11945, opened 2024-10-30
ci(k8s): don't ignore errors on docker network create

michaelbeaumont:
## Motivation

We weren't exiting with an error if `docker network create` failed.

## Implementation information

Avoid repeating the network handling logic by using a Makefile variable.

### DIFF
--- a/mk/k3d.mk
+++ b/mk/k3d.mk
@@ -105,12 +105,23 @@ ifdef K3D_HELM_DEPLOY_ADDITIONAL_OPTS
 	K3D_HELM_DEPLOY_OPTS += $(K3D_HELM_DEPLOY_ADDITIONAL_OPTS)
 endif
 
+define maybe_with_flock
+  if which flock >/dev/null 2>&1; then \
+    SHELL=bash flock -x $(BUILD_DIR)/k3d_network.lock -c $(1); \
+  else \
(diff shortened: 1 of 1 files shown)


lahabana:
Maybe we just don't need the flock and there's a better way to do this...

slonka (approved):
> Maybe we just don't need the flock and there's a better way to do this...

I'd say this is be than before so we can create an issue about flock and merge this

jijiechen (approved):

lahabana:
I think someone agrees that flock is wrong (I'm the one who introduced this :) ): https://stackoverflow.com/a/59724999 :) 

I think [order only prerequisite](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html) would enable us to drop the flock altogether. Let me try a PR :)

michaelbeaumont:
> I think [order only prerequisite](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html) would enable us to drop the flock altogether. Let me try a PR :)

When I opened this PR I knew 50/50 either approved or trigger @lahabana to come up with some make magic :slightly_smiling_face:

michaelbeaumont:
@lahabana im gonna merge this for now cause I think it marginally improves things

PR #11976, opened 2024-11-05
chore(kuma-cp): always clone labels

michaelbeaumont:
## Motivation

Avoid unintentional concurrent map access

## Supporting documentation

Fix #XX


lahabana:
superseeded by #11997

Issue #11968, opened 2024-11-04
Try `MeshLoadBalancingStrategy` with `MeshMultizoneService`


bartsmykla:
### Description

We should try `MeshLoadBalancingStrategy` with `MeshMultizoneService` as it should behave differently that we should prioritize local zone instead of prioritizing splitting traffic between all zones.

This is followup for: https://github.com/kumahq/kuma/issues/11966

lukidzi:
Additionally, we need to ensure that when a user disables locality-aware routing using [MeshLoadBalancingStrategy](https://kuma.io/docs/2.9.x/policies/meshloadbalancingstrategy/#disabling-locality-aware-routing), traffic is properly distributed across zones.

michaelbeaumont:
> we should prioritize local zone instead of prioritizing splitting traffic between all zones.

btw this is already the case

lahabana:
https://github.com/kumahq/kuma/pull/11984

lukidzi:
We have two options to set locality-aware load balancing in Kuma:

1. Mesh: [Locality-Aware Load Balancing](https://kuma.io/docs/2.9.x/policies/locality-aware/#locality-aware-load-balancing)
2. MeshLoadBalancingStrategy (MLBS): [Locality Awareness in MLBS](https://kuma.io/docs/2.9.x/policies/meshloadbalancingstrategy/#localityawareness)

These options allow control over whether traffic should be split across different zones or prioritized within the local zone.

* MeshService: Prioritizes the local zone.
* MeshMultiZoneService: Prioritizes the local zone.
* kuma.io/service: Sets the default priority to 0, distributing traffic across different zones.

When the locality-aware flag is enabled on the Mesh, traffic for kuma.io/service will remain within the same zone. The same effect occurs with MLBS: when set to true, traffic stays within the local zone. If both sets, MLBS can override it

michaelbeaumont:
@lahabana @lukidzi do we have any action to take given the above? or can we close this again

lukidzi:
I think there is one more thing to try, we should check how we behave by the default (no policy/lb mesh configured)

michaelbeaumont:
For this we have an e2e test I think right? It's locality aware

lahabana:
Putting this back in pending so we discuss this

lukidzi:
> lets check if there is a test covering this case and if we can confirm it works as expected, close the issue

slonka:
@lukidzi if you have time could you please check that?

[...]

PR #11980, opened 2024-11-05
fix(MeshLoadBalancingStrategy): set all priorities equal if localityAware is disabled

michaelbeaumont:
## Motivation

We already set `priority: 1` for remote zones in `fillRemoteMeshServices` so we need to explicitly set `priority: 0` if we _don't_ want locality awareness.

Also fix an insufficient condition in an existing case, where we weren't verifying that requests really go to all zones, only that it's OK if they do.

## Supporting documentation

Fixes https://github.com/kumahq/kuma/issues/11968

### DIFF
--- a/test/e2e_env/multizone/localityawarelb/meshmultizoneservice.go
+++ b/test/e2e_env/multizone/localityawarelb/meshmultizoneservice.go
@@ -49,6 +49,7 @@ spec:
 				testserver.WithMesh(meshName),
 				testserver.WithEchoArgs("echo", "--instance", "kube-test-server-1"),
 			)).
+			Install(democlient.Install(democlient.WithNamespace(namespace), democlient.WithMesh(meshName))).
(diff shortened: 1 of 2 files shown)


test/e2e_env/multizone/localityawarelb/meshmultizoneservice.go
```diff
@@ -115,7 +116,9 @@ spec:
 	It("should fallback only to first zone", func() {
 		// given traffic to other zones
 		Eventually(responseFromInstance(multizone.KubeZone2), "30s", "1s").
-			MustPassRepeatedly(5).Should(Or(Equal("kube-test-server-1"), Equal("uni-test-server")))
```
michaelbeaumont, line 118: This doesn't check that both zones are represented, this also passes if it's always `kube-test-server-1`.

lukidzi (approved):

PR #11987, opened 2024-11-06
test(e2e): fix MMZS + MLBS test after merge

michaelbeaumont:
The uni MeshService is autogenerated, without the namespace tag, since #11978 and the first condition in "should fallback only to first zone" was fixed in #11980. This caused the test to start failing on master.


lukidzi (approved):

lahabana:
Should we disallow using k8s tags on uni?

PR #11995, opened 2024-11-08
ci(github): fix release.yaml to use label:dev in version.yaml

lahabana:
## Motivation

This fixes: https://github.com/kumahq/kuma-website/pull/2073 which is needed as we moved to the latest site-generator with: https://github.com/kumahq/kuma-website/pull/1623.


lahabana:
As draft because I just want to use latest ci-tools once it's release

Issue #11994, opened 2024-11-08
Service reconciler injects annotations only when the Namespace is labeled.


lukidzi:
### What happened?

Users can join the mesh by adding the `kuma.io/sidecar-injection=true` label to either the Namespace or the Deployment. When the label is applied at the Namespace level, a service reconciler adds the necessary annotations to the service to ensure it is configured for gateway usage. The annotations applied by the reconciler are:

```
	IngressServiceUpstream      = "ingress.kubernetes.io/service-upstream"
	NginxIngressServiceUpstream = "nginx.ingress.kubernetes.io/service-upstream"
```

When a gateway is deployed within a Namespace that has the `kuma.io/sidecar-injection` label, the reconciler adds these labels correctly. However, if the `kuma.io/sidecar-injection` label is added at the Deployment level rather than the Namespace level, the annotations are not applied as expected.

Expected behavior:
Annotations should be injected when the `kuma.io/sidecar-injection` label is set at the Deployment level, ensuring consistent behavior regardless of where the label is applied.

lukidzi:
>There's no direct connection between a Deployment and a Service, so we can't reliably add annotations to the Service based solely on the presence of the `kuma.io/sidecar-injection` label on the Deployment. Additionally, the Deployment might be created after the Service, making it even harder to automatically sync annotations in such cases.

> One solution could be to require users to manually add the necessary annotations to the Service when applying `kuma.io/sidecar-injection` at the Deployment level. 
> 1. If this approach is adopted, clear documentation is essential to guide users on manually adding the required labels and annotations to ensure consistent behavior. 
> 2. The user should additionally annotate Service with `kuma.io/sidecar-injection` and later we could add these labels automatically (check if that is not supported yet)

lukidzi:
> triage: we should document this better:
Once a user has a kuma.io/sidecar-injection=true on the Deployment/ReplicaSet...., following annotations should be added manually:
> * "ingress.kubernetes.io/service-upstream"
> * "nginx.ingress.kubernetes.io/service-upstream"

jijiechen:
So there is no need to change Kuma code and the only thing we want to do is to update the doc website right?

PR #12005, opened 2024-11-11
fix(controller): include services that have pods labled injection as part of mesh

jijiechen:
## Motivation

Improve reconciler of service controller

## Implementation information

Detect if there is at least one pods of the service is labeled "kuma.io/sidecar-injection=enabled"

## Supporting documentation

fixes https://github.com/kumahq/kuma/issues/11994

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/service_controller.go
+++ b/pkg/plugins/runtime/k8s/controllers/service_controller.go
@@ -3,11 +3,11 @@ package controllers
 import (
 	"context"
 	"fmt"
-
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


michaelbeaumont (changes requested):
It's not clear to me that we want to do this. We are trying to break the `Pod controls Service` logic we've had up until now.

Most of the time there's a Deployment/StatefulSet/DaemonSet controlling these Pods with the Pod template setting these labels. Why wouldn't the user also be able to set a label on the Service?

jijiechen:
@lahabana @lukidzi WDYT? 
The current behaviour is not aligning to what we say in the doc and now it's requested by an actual user.

PR #12017, opened 2024-11-11
fix(gateway): invalid Envoy resources with real MeshService backends and `appProtocol: tcp`

michaelbeaumont:
## Motivation

We were generating invalid Envoy resources if using MeshHTTPRoute and the real backend had `appProtocol: tcp`.

We already weren't setting
`dest.Destination[mesh_proto.ServiceProtocol]` when calling `generateMeshCluster`.


bartsmykla:
@michaelbeaumont what is the state of this PR?

michaelbeaumont:
Obsolete I think

PR #4834, opened 2022-08-16
build: upgrade to google.golang.org/protobuf

michaelbeaumont:
In preparation for #2719 

Blocked by #4830 

Note that we continue to use `github.com/golang/protobuf/jsonpb` [due to the old library unmarshaling go-like `time.Duration`s even though it's against protobuf spec](https://github.com/golang/protobuf/issues/1374).

### Checklist prior to review

- [x] Link to docs PR or issue --
- [x] Link to UI issue or PR --
- [x] Is the [issue worked on linked][1]? --
- [x] The PR does not hardcode values that might break projects that depend on kuma (e.g. "kumahq" as a image registry) --
- [x] The PR will work for both Linux and Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [x] Unit Tests --
- [x] E2E Tests --
- [x] Manual Universal Tests -- none
- [x] Manual Kubernetes Tests --
- [x] Do you need to update [`UPGRADE.md`](/UPGRADE.md)? --
- [x] Does it need to be backported according to the [backporting policy](/CONTRIBUTING.md#backporting)? --

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #12038, opened 2024-11-14
chore(deps): remove deprecated protobuf dependency

### LINKED ISSUE #3499: Update references to deprecated protobuf lib

Icarus9913:
## Motivation

The repo `github.com/golang/protobuf` is deprecated and we should switch to use `google.golang.org/protobuf`.

However we can't totally get rid of it because we still need the `github.com/golang/protobuf/jsonpb` and there's a incompatibility for `google.golang.org/protobuf/encoding/protojson`. So, I add one golangci linter to warn it.

## Supporting documentation

Fix #3499


Icarus9913:
@lahabana How about the new changes?

1. rejected all of the GolangProtobuf new usages
2. loosen the GolangProtobuf jsonpb with `// nolint: depguard`
3. recommend anyone who wants to use GoogleProtobuf protojson to switch back to GolangProtobuf jsonpb. Otherwise if the user consist to use GoogleProtobuf protojson, let them just add  `// nolint: depguard`

---
However, in this way anyone who wants to use `json*` whichever they choosem, they have to add `// nolint: depguard`

lahabana:
LGTM intuitively for the lint rules. I'm not sure I'm grasping everything with the code change so not approving.

.golangci.yml
```diff
@@ -99,6 +99,18 @@ linters-settings:
...
+          - $all
+        deny:
+          - pkg: github.com/golang/protobuf
+            desc: "github.com/golang/protobuf is deprecated; use google.golang.org/protobuf instead"
+      DenyGoogleProtoJson:
+        files:
+          - $all
+        deny:
+          - pkg: google.golang.org/protobuf/encoding/protojson
+            desc: "don't use the protojson package; use github.com/golang/protobuf/jsonpb instead"
```
lukidzi, line 113: Could we add a note explaining why it shouldn't be used? Specifically, that it’s incompatible and might cause issues. We could include this as a comment, but it’s worth noting explicitly.

.golangci.yml
```diff
@@ -99,6 +99,18 @@ linters-settings:
         deny:
           - pkg: "github.com/kumahq/kuma/app"
             desc: "pkg can't import app"
+      DenyGolangProtobuf:
+        files:
+          - $all
+        deny:
```
lukidzi, line 105: Not a specialist here but is it possible to add 
```yaml
allow:
  - "github.com/golang/protobuf/jsonpb"
```

and avoid `// nolint: depguard` ?

Icarus9913: However, I think it's also a note for the user to know what happend for the `json` sdk and be more careful. Does that make sense?

PS: Get a try locally but it doesn't work well (It seems like a bug for golanglint and it gave lots of other unreleated packages warnings)

pkg/core/xds/resource.go
```diff
@@ -5,7 +5,7 @@ import (
 
 	envoy_sd "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	envoy_types "github.com/envoyproxy/go-control-plane/pkg/cache/types"
-	protov1 "github.com/golang/protobuf/proto"
+	"google.golang.org/protobuf/runtime/protoimpl"
```
Icarus9913, line 8: > WARNING: This package should only ever be imported by generated messages. The compatibility agreement covers nothing except for functionality needed to keep existing generated messages operational. Breakages that occur due to unauthorized usages of this package are not the author's responsibility.

https://pkg.go.dev/google.golang.org/protobuf@v1.35.2/runtime/protoimpl
I need to use one public API

Icarus9913: I noticed that in the previous Mike's PR https://github.com/kumahq/kuma/pull/4834 regenerated the objects(proto message v2).

So we don't need to use this internal package voluntarily and we can also remove `protov1.MessageV1` function to convert a `v1/v2` message to a `v2` message.

lukidzi (approved):
Great work 🎉 IMO it looks good. Since files were regenerated and go-control-plane library upgraded I think we are fine with the change.

Issue #12052, opened 2024-11-14
MeshPassthrough doesnt work correctly when CIDR and protocol HTTP


lukidzi:
### What happened?

It seems that when a user creates a policy with CIDRs and the HTTP protocol, we don't generate the correct routes. We should address this and ensure it works properly.

```
    targetRef:
      kind: Mesh
    default:
      appendMatch:
        - type: CIDR
          value: 172.16.0.0/12
          protocol: http
          port: 8126
        - type: CIDR
          value: 192.168.0.0/16
          protocol: http
          port: 8126
        - type: CIDR
          value: 240.0.0.0/4
          protocol: http
          port: 8126
```

PR #12054, opened 2024-11-14
fix(meshpassthrough): generate separate filter chain  when ip/cidr fo…

lukidzi:
## Motivation

While testing the policy, we noticed some unexpected behavior:

* When a CIDR and protocol HTTP were defined, proper outbounds were not generated.
* When the protocol was HTTP, the matcher was missing the port.
* Ordering was incorrect when protocols within the same group (HTTP/HTTP2/gRPC) were used.

## Implementation information

* Updated the logic to create individual FilterChainMatch structures, aligning with Envoy's filter chain matcher parameters.
* Added routes exclusively for HTTP matching with domains, as they need to be placed in virtual hosts.
* Resolved the issue where the port matcher was missing for HTTP.
* Fixed the problem where the combination of protocol HTTP and CIDR did not generate correct routes.
* Corrected the ordering by introducing an order map for protocols, ensuring that more specific routes are not placed at the end for HTTP/HTTP2/gRPC protocols.
* *new limitation* wildcard domains with L7 protocol and all ports doesn't work. It's because envoy doesn't support `*.domain.com:*` in a virtual host

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12052


pkg/plugins/policies/meshpassthrough/plugin/xds/order.go
```diff
@@ -11,108 +11,136 @@ import (
...
-	Protocol core_mesh.Protocol
-	Port     uint32
-	Routes   []Route
+type FilterChainMatch struct {
+	Protocol  core_mesh.Protocol
+	Port      uint32
+	MatchType MatchType
+	Value     string
+	Routes    []Route
 }
```
Icarus9913, line 54: One idea from my mind but not sure if it works well. Since this PR would supplement `MatchType` and `Value` properties for `Matcher` type, which leads to a little bit coupled for type `Route` and `Matcher`.

Not sure if it would be more simple for the implementation with the following structure:

```text
type NewFilterChainMatcher struct {
	Protocol core_mesh.Protocol
	Port     uint32
	Routes   []string
}

var dic = map[MatchType][]NewFilterChainMatcher{}
```

Then just return the `dic` for function `GetOrderedMatchers`

For `listeners_filter_chain.go` structure `FilterChainConfigurer`, we may just use

```text
type FilterChainConfigurer struct {
	APIVersion core_xds.APIVersion
	MatchType MatchType
        FilterChainMatchers []NewFilterChainMatcher{}
	IsIPv6     bool
}
```

pkg/plugins/policies/meshpassthrough/plugin/xds/order.go
```diff
@@ -11,108 +11,136 @@ import (
...
-		filterChainMatchers = append(filterChainMatchers, filterChainMatcher)
-	}
+	// we cannot differentiate between HTTP, HTTP/2, and gRPC on the same port.
 	if err := validatePortAndProtocol(portProtocols); err != nil {
 		return nil, err
 	}
-	orderValues(filterChainMatchers)
+	filterChainMatchers := []FilterChainMatch{}
+	for matcher, routes := range matcherWithRoutes {
+		getOrderedRoutes(routes)
```
jakubdyszkiewicz, line 104: why is this called here?

jakubdyszkiewicz:
as for backporting. It's a significant change, but it's scoped to MeshPasshtrough so I'm ok with backporting it.

lukidzi:
> You mentioned 3 issues being solved by this PR, but I don't see any new tests, only updates to existing golden files and output changes. Are those issues covered by existing tests? If so, how did we not catch it before? (no blaming here) Were golden files too complex to analyze?

I think 2 main things
* complex dump
* not enough edge cases

The problem is that if the configuration is small we cannot hit some route aggregation/protocol ordering, but when is too big we might miss something. Maybe more targeted test cases could help here

pkg/plugins/policies/meshpassthrough/plugin/xds/order.go
lukidzi: I think we should create an issue to refactor it and make it more readable.

pkg/plugins/policies/meshpassthrough/plugin/xds/configurer.go
```diff
@@ -106,30 +110,26 @@ func (c Configurer) configureListenerFilter(listener *envoy_listener.Listener) e
...
-				route.MatchType == IPV6 {
-				return true
-			}
+func hasIPv6Matches(orderedMatchers []FilterChainMatch) bool {
+	return slices.ContainsFunc(orderedMatchers, func(matcher FilterChainMatch) bool {
+		if matcher.MatchType == Domain ||
+			matcher.MatchType == WildcardDomain ||
+			matcher.MatchType == CIDRV6 ||
+			matcher.MatchType == IPV6 {
+			return true
```
Icarus9913, line 131: I mean here maybe we could just use:

```text
if slices.Contains([]string{Domain,WildcardDomain,CIDRV6,IPV6}, matcher.MatchType) {
  return true
}
```

pkg/plugins/policies/meshpassthrough/plugin/xds/order.go
```diff
@@ -11,108 +11,135 @@ import (
...
-		if match.Port != nil {
-			port = *match.Port
-		}
+		port := pointer.DerefOr(match.Port, 0)
 		protocol := core_mesh.ParseProtocol(string(match.Protocol))
+		matchType := getMatchType(match)
+		matcher := Matcher{
+			Protocol:  protocol,
+			Port:      uint32(port),
+			MatchType: matchType,
```
Icarus9913, line 66: In the past, the `Matcher` only has 'Protocol' and 'Port' properties.
However, right now If we have the following policies, we would meet conflict because we would generate 2 elements in array `filterChainMatchers` (in the past we would only have one) :

```text
apiVersion: kuma.io/v1alpha1
kind: MeshPassthrough
metadata:
  name: one
  namespace: kuma-system
  labels:
    kuma.io/mesh: default
spec:
  targetRef:
    kind: Mesh
  default:
    passthroughMode: Matched
    appendMatch:
      - port: 80
        protocol: http
        type: Domain
        value: "www.google.com"
---
apiVersion: kuma.io/v1alpha1
kind: MeshPassthrough
metadata:
  name: two
  namespace: kuma-system
  labels:
    kuma.io/mesh: default
spec:
  targetRef:
    kind: MeshSubset
    proxyTypes:
      - Sidecar
    tags:
      kuma.io/service: nginx-service_kuma-demo_svc_80
  default:
    passthroughMode: Matched
    appendMatch:
      - port: 80
        protocol: http
        type: Domain
        value: "*.google.com"
```

Icarus9913: ### old  orderedMatchers

```text
[{http 80 [{www.google.com 1} {*.google.com 2}]}]
```

### new orderedMatchers: 

```text
[{http 80 2  [{www.google.com 2}]} {http 80 1  [{*.google.com 1}]}]
```

lahabana:
Do we have a test case that covers the case: https://github.com/kumahq/kuma/pull/12054#issuecomment-2503469379

Icarus9913:
Could you please  supplment the new limitation of `WildcardDomain` with `* port` is not support anymore in the current PR description and also add it in our website?

pkg/plugins/policies/meshpassthrough/plugin/xds/order.go
```diff
@@ -11,109 +13,183 @@ import (
...
+					}
+				}
+			} else {
+				matcher.Value = match.Value
+				// there should be no existing matcher if there is ip/cidr
+				matcherWithRoutes[matcher] = map[Route]bool{}
 			}
+		default:
+			matcher.Value = match.Value
+			matcherWithRoutes[matcher] = map[Route]bool{}
```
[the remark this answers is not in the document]
lukidzi: I feel switch much clearer to read than if statement

pkg/plugins/policies/meshpassthrough/plugin/xds/order.go
```diff
@@ -11,109 +13,183 @@ import (
...
+		if _, found := matcherWithRoutesAndAdditionalPorts[matcher]; found {
+			for route := range routes {
+				matcherWithRoutesAndAdditionalPorts[matcher][Route{
+					Value:     route.Value,
+					MatchType: route.MatchType,
+				}] = true
+			}
+		} else {
+			matcherWithRoutesAndAdditionalPorts[matcher] = maps.Clone(routes)
+		}
```
Icarus9913, line 122: Do we have the `found` case? 
Since `matcherWithRoutesAndAdditionalPorts` and `matcherWithRoutes` are the same structure `map[Matcher]map[Route]bool{}`, a map only has one unique key right?

Please correct me if I'm wrong

Icarus9913 (approved):

PR #12059, opened 2024-11-15
fix(kuma-cp):  unnecessary `failed to write error to closed channel` errors

lobkovilya:
## Motivation

Error channel had capacity 0, that's why every time when we write to it like:
```go
select {
case errChan <- err:
default:
    log.Error(err, "failed to write error to closed channel")
}
```
we always go to the `default` section.

## Implementation information

Update channel capacity to be 1.

## Supporting documentation

None


michaelbeaumont (changes requested):
Wait but this isn't even a write to a closed channel, if that happens go panics. Can we fix the error message? Also is a buffer capacity of 1 enough?

lobkovilya:
I checked more carefully, we don't need `errCh` in callbacks at all, should be simpler now

pkg/kds/global/components.go
```diff
@@ -123,19 +103,12 @@ func Setup(rt runtime.Runtime) error {
...
-				select {
-				case errChan <- err:
-				default:
-					log.Error(err, "failed to write error to closed channel")
-				}
-			} else {
-				log.V(1).Info("KDSSyncClient finished gracefully")
-			}
-		}()
+		if err := sink.Receive(); err != nil && (status.Code(err) != codes.Canceled && !errors.Is(err, context.Canceled)) {
```
michaelbeaumont, line 106: Hmm why is this necessarily OK to run synchronously?

lobkovilya: This callback already runs asynchronously like:
```go
processingErrorsCh := make(chan error, 1)
go func() {
	if err := g.zoneToGlobalCb.OnZoneToGlobalSyncConnect(stream); err != nil {
		processingErrorsCh <- err
	}
}()
```
so I don't really understand why `sink.Receive` had to run asynchronously as well

michaelbeaumont: OK so it's fine, since we don't wait for it to return anyway

michaelbeaumont (approved):

Issue #12044, opened 2024-11-14
Add endpoint to retrieve resources by resourceIdentifier


lahabana:
### Description

We now have a systematic [resourceIdentifier](https://github.com/kumahq/kuma/blob/dfb9c5d4fc16b340555029d02f134f4aa43c350e/pkg/core/resources/model/resource.go#L767-L774) for everything.

There's a hash that should be computed from the resource Identifier. Is there a systematic way to compute the hash from a set of labels?

So could we have a new endpoint:

```
/meshes/<mesh>/<resourceType>/_resourceIdentifier/zones/<zoneName>/namespaces/<namespace>/name/<displayName>
```

These would always return the same as:

```
/meshes/<mesh>/<resourceType>/<hashedName>
```

This should be a MADR and this MADR should define:
- resourceIdentifier 
- hashing algorithm from resourceIdentifier to resourceName
- new api endpoint and how it should be implemented

This endpoint would enable the GUI to map from XDS config to resources and to hide altogether resourceNames to only rely on resourceIdentifier everywhere.

xref: https://github.com/kumahq/kuma-gui/issues/2848

PR #12109, opened 2024-11-26
feat(api): add short name support

lahabana:
## Motivation

This adds short names that can be used in kubectl. It also creates a path for the shortname to be also used in resource Identifier like in XDS config for example.

## Implementation information

Add possibility to have shortname which is used in kubectl but also when creating unique resource identifier

This is the first step in a multiple step of changes to make resource identifier easier to use
## Supporting documentation

could help implem of: https://github.com/kumahq/kuma/issues/12044


lobkovilya (approved):
LGTM! 

Probably a good thing to add is to check in RegisterType that short name is unique (the same way we do for `res.Name`) https://github.com/kumahq/kuma/blob/65103c84a0e16c4a1899900f39aa61c72245cd40/pkg/core/resources/registry/registry.go#L89

as there is always a chance in the future to get 2 policies with the same acronym

api/mesh/v1alpha1/dataplane.proto
```diff
@@ -15,6 +15,7 @@ import "validate/validate.proto";
 message Dataplane {
 
   option (kuma.mesh.resource).name = "DataplaneResource";
+  option (kuma.mesh.resource).short_name = "dpp";
```
michaelbeaumont, line 18: Why dpp? Isn't that dataplane proxy?

Issue #12110, opened 2024-11-26
MeshLoadBalancingStrategy support for RingHash and TCP traffic


lukidzi:
### Description

By [design](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/022-lb-policy.md), we chose to support `RING_HASH` routing only for HTTP traffic. However, it is also possible to configure it for TCP traffic.

Use Case:
As a user, I want all requests from a specific source IP to route to the same instance of my database consistently.

Example Envoy configuration:

```
static_resources:
  listeners:
    - name: listener_0
      address:
        socket_address:
          address: 0.0.0.0
          port_value: 10000
      filter_chains:
        - filters:
            - name: envoy.filters.network.tcp_proxy
              typed_config:
                "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
                stat_prefix: tcp_stats
                cluster: service_ring_hash
                hash_policy:
                  - source_ip: {}

  clusters:
    - name: service_ring_hash
      connect_timeout: 0.25s
      type: STRICT_DNS
      lb_policy: RING_HASH
      load_assignment:
        cluster_name: service_ring_hash
        endpoints:
          - lb_endpoints:
              - endpoint:
                  address:
                    socket_address: { address: 127.0.0.1, port_value: 9001 }
              - endpoint:
                  address:
                    socket_address: { address: 127.0.0.1, port_value: 9002 }
              - endpoint:
                  address:
                    socket_address: { address: 127.0.0.1, port_value: 9003 }
```

lobkovilya:
Reopen when someone needs it.

PR #12111, opened 2024-11-26
fix(meshloadbalancingstrategy): deprecate SourceIP and use Connection

lukidzi:
## Motivation

While investigating [kuma#12110](https://github.com/kumahq/kuma/issues/12110), we noticed that the following policy is rejected:

```yaml
kind: MeshLoadBalancingStrategy
apiVersion: kuma.io/v1alpha1
metadata:
  name: ring-hash
  namespace: kuma-system
  labels:
    kuma.io/mesh: default
spec:
  targetRef:
    kind: MeshSubset
    tags: 
      kuma.io/service: demo-app_kuma-demo_svc_5000
  to:
  - targetRef:
      kind: MeshService
      name: redis_kuma-demo_svc_6379
    default:
      loadBalancer:
        type: RingHash
        ringHash:
          hashPolicies:
          - type: Connection
            connection:
              sourceIP: true
```

However, a policy with `type: SourceIP` passes validation, which seems incorrect based on the [documentation](https://kuma.io/docs/2.9.x/policies/meshloadbalancingstrategy/#ringhash). Upon checking the code, I found that the kubebuilder validation uses the incorrect enum `SourceIP`, whereas the correct enum `Connection` should be defined.

## Implementation information

To avoid breaking existing customer policies, I’ve added a new type to the enum while deprecating the old one.

## Supporting documentation

Fix #XX

### DIFF
--- a/pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/deprecated.go
+++ b/pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/deprecated.go
@@ -18,7 +18,7 @@ func validateHashPoliciesType(confs []To) []string {
 		if conf.Default.LoadBalancer == nil {
 			continue
 		}
-	
+
 		switch conf.Default.LoadBalancer.Type {
 		case RingHashType:
 			if conf.Default.LoadBalancer.RingHash == nil || conf.Default.LoadBalancer.RingHash.HashPolicies == nil {
--- a/pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/deprecated.go
+++ b/pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/deprecated.go
@@ -15,25 +15,27 @@ func (t *MeshLoadBalancingStrategyResource) Deprecations() []string {
 func validateHashPoliciesType(confs []To) []string {
 	deprecations := []string{}
 	for ruleIdx, conf := range confs {
-		if conf.Default.LoadBalancer != nil {
-			switch conf.Default.LoadBalancer.Type {
-			case RingHashType:
-				if conf.Default.LoadBalancer.RingHash == nil || conf.Default.LoadBalancer.RingHash.HashPolicies == nil {
-					continue
+		if conf.Default.LoadBalancer == nil {
+			continue
+		}
+	
+		switch conf.Default.LoadBalancer.Type {
+		case RingHashType:
+			if conf.Default.LoadBalancer.RingHash == nil || conf.Default.LoadBalancer.RingHash.HashPolicies == nil {
+				continue
+			}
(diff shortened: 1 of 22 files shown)


jijiechen (approved):

jijiechen:
An issue tracking the removal of `SourceIP` should also be creaated.

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/deprecated.go
```diff
@@ -1,9 +1,42 @@
...
+				if conf.Default.LoadBalancer.Maglev == nil || conf.Default.LoadBalancer.Maglev.HashPolicies == nil {
+					continue
+				}
+				for lbIdx, lbConf := range *conf.Default.LoadBalancer.Maglev.HashPolicies {
+					if lbConf.Type == SourceIPType {
+						deprecations = append(deprecations, fmt.Sprintf("%s type for 'spec.to[%d].default.loadBalancer.maglev.hashPolicies[%d].type' is deprecated, use %s instead", SourceIPType, ruleIdx, lbIdx, ConnectionType))
+					}
+				}
+			}
+		}
```
bartsmykla, line 39: Nit: Maybe instead of nesting it under if statement, we could just continue if it's false:
```suggestion
		if conf.Default.LoadBalancer == nil {
			continue
		}
	
		switch conf.Default.LoadBalancer.Type {
		case RingHashType:
			if conf.Default.LoadBalancer.RingHash == nil || conf.Default.LoadBalancer.RingHash.HashPolicies == nil {
				continue
			}
			for lbIdx, lbConf := range *conf.Default.LoadBalancer.RingHash.HashPolicies {
				if lbConf.Type == SourceIPType {
					deprecations = append(deprecations, fmt.Sprintf("%s type for 'spec.to[%d].default.loadBalancer.ringHash.hashPolicies[%d].type' is deprecated, use %s instead", SourceIPType, ruleIdx, lbIdx, ConnectionType))
				}
			}
		case MaglevType:
			if conf.Default.LoadBalancer.Maglev == nil || conf.Default.LoadBalancer.Maglev.HashPolicies == nil {
				continue
			}
			for lbIdx, lbConf := range *conf.Default.LoadBalancer.Maglev.HashPolicies {
				if lbConf.Type == SourceIPType {
					deprecations = append(deprecations, fmt.Sprintf("%s type for 'spec.to[%d].default.loadBalancer.maglev.hashPolicies[%d].type' is deprecated, use %s instead", SourceIPType, ruleIdx, lbIdx, ConnectionType))
				}
			}
		}
```

bartsmykla (approved):
One nit, but looks good to me overall

lukidzi:
> An issue tracking the removal of `SourceIP` should also be creaated.

Thanks, forgot about it: https://github.com/kumahq/kuma/issues/12145

PR #11363, opened 2024-09-09
chore(deps): bump go to v1.23

michaelbeaumont:
Requires https://github.com/kumahq/kuma/pull/11362

### Checklist prior to review

- [x] [Link to relevant issue][1] as well as docs and UI issues --
- [x] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [x] Tests (Unit test, E2E tests, manual test on universal and k8s) --
  - Don't forget `ci/` labels to run additional/fewer tests
- [x] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [x] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) --

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #12141, opened 2024-11-30
refactor(*): replace experimental `maps` and `slices` with stdlib

Juneezee:
## Motivation

We upgraded to Go 1.23 in https://github.com/kumahq/kuma/pull/11363. The experimental functions are now part of the standard library as of Go 1.21 and Go 1.23.

Reference: https://go.dev/doc/go1.21#slices
Reference: https://go.dev/doc/go1.23#iterators

## Implementation information

The key difference is that `maps.Keys` and `maps.Values` in the `golang.org/x/exp` package return a slice, whereas `maps.Keys` and `maps.Values` in the standard library return an iterator. To work with slices, we need to use `slices.Collect` to convert the iterator into a slice.


jijiechen (approved):
LGTM

Issue #12045, opened 2024-11-14
MeshTrace not working on deployment without kube service


slavogiez:
### What happened?

We configured a MeshTrace policy with a Datadog backend.
This works fine and we can see traces in Datadog, but we have one case where it's not working. 

We run a curl command from the source service container to another mesh service. We can see the traces from the destination sidecar, but not from the source sidecar.

It looks like it doesn't work when the source service doesn't have a kube service (no server listening).

We configured the following MeshTrace policy :

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTrace
metadata:
  labels:
    k8s.kuma.io/namespace: kong-mesh
    kuma.io/mesh: mesh01
    kuma.io/policy-role: system
  name: trace-default
  namespace: kong-mesh
spec:
  default:
    backends:
      - datadog:
          splitService: true
          url: http://trace-svc.datadog-agent.svc.cluster.local:8126
        type: Datadog
    sampling:
      client: 100
      overall: 100
      random: 100
    tags:
      - literal: xxx
        name: org_name
      - literal: yyy
        name: mesh_name
  targetRef:
    kind: Mesh
```

And we have the following deployment : 

```yaml
apiVersion: "apps/v1"
kind: "Deployment"
metadata:
  name: curl
spec:
  replicas: 1
  selector:
    matchLabels:
      app: curl
  template:
    metadata:
      annotations:
        kuma.io/sidecar-env-vars: "DD_ENV=dev;DD_SERVICE=curl-mesh-sidecar"
      labels:
        app: curl
        kuma.io/sidecar-injection: "enabled"
        kuma.io/mesh: mesh01
    spec:
      terminationGracePeriodSeconds: 5
      containers:
      - name: curl
        image: alpine/curl
        command: [ "/bin/sh", "-c", "--" ]
        args: [ "while true; do sleep 30; done;" ]
```

In the XDS config from the source service, we can see the datadog extension, its cluster but not the tracing configuration.
In the manager ui, we can see the MeshTrace policy applied on the DP proxy.

slavogiez:
Not sure what changed but now we can see expected spans from the source service. The only remaining issue is that they don't appear in the whole trace, so we're not able to find the source service in the trace.

lukidzi:
It seems we don't configure listeners for MeshService.

PR #12173, opened 2024-12-04
fix(meshtrace): add support for real resources

lukidzi:
## Motivation

Issue https://github.com/kumahq/kuma/issues/12045 showed that we didn't configure real resources

## Implementation information

e1d1ed73e108b26d24c785b5b93b96cd0fae174a extracted test string to golden files
f6ae3d12ab0f4199af925a248e749114da8afa0f added unit test showing an issue
e7d8c6cef0f08366ee36e8a52df01b256dbbd03c implementation of real resource and generated file

Added a function that loops over real resources to configure listeners for MeshService and MeshMultiZoneService. Since we don't have a `kuma.io/service` as a destination, we set the output of `DestinationName(port)`, which is used for clusters.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12045


pkg/plugins/policies/meshtrace/plugin/v1alpha1/plugin.go
```diff
@@ -137,6 +145,52 @@ func applyToOutbounds(
 	return nil
 }
 
+func applyToRealResources(
+	ctx xds_context.Context,
+	rules core_rules.SingleItemRules,
+	rs *xds.ResourceSet,
+	proxy *xds.Proxy,
+) error {
+	for uri, resType := range rs.IndexByOrigin(xds.NonMeshExternalService) {
```
lahabana, line 154: Why not MES?

lukidzi: Since we don't support policies on Egress, thought maybe no point to add it now

pkg/plugins/policies/meshtrace/plugin/v1alpha1/plugin.go
```diff
@@ -137,6 +145,52 @@ func applyToOutbounds(
...
+						return err
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func getDestinationName(uri model.TypedResourceIdentifier, ctx xds_context.Context) (string, error) {
```
lobkovilya, line 173: is it possible to use https://github.com/kumahq/kuma/blob/882a878cdc0495fccef5300bc107a035bb22cf60/pkg/plugins/policies/core/xds/meshroute/listeners.go#L189 instead?

PR #12175, opened 2024-12-04
feat(xds): enable delta XDS for all tests to verify stability

lukidzi:
## Motivation

Verify stability of the change. Since verify ci statbility can only run from the repo I've created a branch to check state of the change.

## Supporting documentation

Fix #XX


lukidzi:
I feel like it uses a lot of cpu, need investigation

PR #12243, opened 2024-12-11
fix(kds): remove context from map on stream close

lukidzi:
## Motivation

The application's memory usage was slowly increasing but did not correlate with the number of active connections. I decided to investigate the memory usage and noticed some context entries. While reviewing the code, I discovered that we are not cleaning up the map of streams and contexts. I'm not certain if this is the primary issue, but it could be a contributing factor.

## Implementation information

Implemented the `OnDeltaStreamClosed` method to remove entries from the map when a stream is closed.

## Supporting documentation

xref: https://github.com/Kong/kong-mesh/issues/6616


slonka (approved):
Nice catch!

michaelbeaumont (approved):

PR #12244, opened 2024-12-11
fix(helm): add back `KUMA_DATAPLANE_NAME` to egress/ingress deployment

lahabana:
In 2.5.x we introduced a change to not rely on composed env var.

However, sometimes folks update helm before their binaries. Therefore we introduce the removed env var to be able to run 2.4 proxies with helm charts for version 2.6 and 2.7

fix kumahq/kuma-website#2143

## Supporting documentation

Fix #XX


jijiechen:
Hi @lahabana  didn't you tell me that `backport` label can only be added to PRs that are already merged?

jijiechen:
Hi @lahabana I remember you said the `backport` label can only be added to PRs that are already merged, has this changed?

lahabana:
We agreed that this shouldn't be something we support

PR #12245, opened 2024-12-12
chore(app): add common logging of version, config on startup

lahabana:
## Motivation

In a customer investigation it took a while to figure what the version of the dp was.
Also the logs between the CP and DP were quite different.

## Implementation

- Log version and config as early as possible on startup of cp and dp
- Use similar log messages for cp and dp to develop muscle memory
- Simplify api to get json safe to print

### DIFF
--- a/app/kuma-cp/cmd/run.go
+++ b/app/kuma-cp/cmd/run.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"fmt"
 	"time"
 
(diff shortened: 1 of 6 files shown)


bartsmykla (approved):
lgtm, should we backport it maybe?

michaelbeaumont (approved):

lahabana:
I don't think it fits our backporting rules

Issue #11507, opened 2024-09-20
Add mesh to openAPI spec


lahabana:
(this issue has been edited by @slonka)

### Description

Other resources we want to have: Mesh, MeshGateway, Secret/GlobalSecret, Zone.

This is part of "[As a user I want to manage mesh resources using OpenAPI spec](https://docs.google.com/document/d/1Ewe0yRx-vvAuQn5CWRFhJbBPLzdbBQ2wsQosZXk4Go8/edit?tab=t.0#heading=h.6j6867jtxnwu)" user story.

We need to find a way to add Mesh to the openAPI spec.

This is tricky because unlike existing resources in the openAPI spec where we go with resource plugins etc, this has a protobuf first approach.

If this could be automatic in some ways that would be ideal as we could add other resources that are still in protobuf and potentially get a complete openAPI spec.

slonka:
coming from: https://github.com/Kong/kong-mesh/issues/6644 

> have default policies inside a mesh but it should be easy to turn them off.

we have [skip_creating_initial_policies](https://github.com/kumahq/kuma/blob/f8179d7a8d0348a67ac7e7983e6d65c1fee339ed/api/mesh/v1alpha1/mesh.proto#L77) - is it just being able to configure this?

slonka:
In an ideal world we would have a flag "publishToOAS" and it would just work.

PR #12272, opened 2024-12-16
feat(oapi): add mesh and meshgateway to oapi spec

slonka:
## Motivation

We want to be able to mange these resources using Terraform and need OpenAPI spec to be able to generate that. Read the MADR if you're interested in more details: https://docs.google.com/document/d/1Ewe0yRx-vvAuQn5CWRFhJbBPLzdbBQ2wsQosZXk4Go8/ .

## Implementation information

All described in the madr - https://docs.google.com/document/d/1Ewe0yRx-vvAuQn5CWRFhJbBPLzdbBQ2wsQosZXk4Go8/edit?tab=t.0#heading=h.9m1k72xxi21x .

## Supporting documentation

Closes https://github.com/kumahq/kuma/issues/11507


tools/template/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+func main() {
+	// Define command-line flags
+	templateFile := flag.String("template", "", "Path to the Go template file (required)")
+	dataInput := flag.String("data", "", "Path to the YAML/JSON data file or inline JSON/YAML (required)")
+	flag.Parse()
+
+	// Validate required flags
+	if *templateFile == "" || *dataInput == "" {
+		fmt.Println("Usage: go run main.go --template <template_file> --data <data_file_or_inline_JSON/YAML>")
```
slonka, line 21: use: log.Fatal("")

tools/template/resolve_refs.jq
```diff
@@ -0,0 +1,12 @@
+def resolve_refs(obj; defs):
```
slonka, line 1: where to put this?

tools/template/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+	"text/template"
+
+	"sigs.k8s.io/yaml"
+)
+
+func main() {
```
slonka, line 13: write test for this

docs/generated/openapi.yaml
```diff
@@ -1863,6 +1931,74 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/MeshExternalServiceList'
+  /meshgatewayes/{name}:
```
lahabana, line 1934: This path seems wrong

mk/docs.mk
```diff
@@ -37,12 +37,31 @@ docs/generated/openapi.yaml:
...
 	mkdir -p $(OAPI_TMP_DIR)/resources
+
+	for resource in Mesh MeshGateway; do \
+  		lowercase=$$(echo $${resource} | tr '[:upper:]' '[:lower:]'); \
+		mkdir -p $(OAPI_TMP_DIR)/resources/$${lowercase}/; \
+		go run tools/template/main.go --template tools/policy-gen/templates/schema.yaml --data '{"Scope": "Global"}' | \
+		$(YQ) -o json | \
+		$(JQ) --slurpfile defs ./docs/generated/raw/protos/$${resource}.json -f tools/template/resolve_refs.jq | \
+		$(YQ) -P > $(OAPI_TMP_DIR)/resources/$${lowercase}/schema.yaml ; \
+		go run tools/template/main.go --template tools/policy-gen/templates/endpoints.yaml --data "{\"Package\": \"v1.0.0\", \"Name\": \"$${resource}\", \"Scope\": \"Global\", \"Path\": \"$${lowercase}es\"}" > $(OAPI_TMP_DIR)/resources/$${lowercase}/rest.yaml; \
```
lahabana, line 48: One of the reason of the generator was the ability to have access to the [ResourceInfo](https://github.com/kumahq/kuma/blob/2ad2aebba8d59df93a6dd42e55dbae1095efdd3d/tools/resource-gen/genutils/util.go#L45-L69)  then you could use `WsPath` for .e.g

mk/docs.mk
```diff
@@ -37,12 +37,31 @@ docs/generated/openapi.yaml:
 	mkdir -p $(dir $@)
 	mkdir -p $(OAPI_TMP_DIR)/policies
 	mkdir -p $(OAPI_TMP_DIR)/resources
+
+	for resource in Mesh MeshGateway; do \
+  		lowercase=$$(echo $${resource} | tr '[:upper:]' '[:lower:]'); \
+		mkdir -p $(OAPI_TMP_DIR)/resources/$${lowercase}/; \
+		go run tools/template/main.go --template tools/policy-gen/templates/schema.yaml --data '{"Scope": "Global"}' | \
+		$(YQ) -o json | \
+		$(JQ) --slurpfile defs ./docs/generated/raw/protos/$${resource}.json -f tools/template/resolve_refs.jq | \
```
lahabana, line 46: YQ works over json (because json is a subset of yaml), is there a specific reason to use jq here?

lahabana (commented):
I believe you might be making things hard on yourself by not leveraging on of the existing I would add a type `openapi` here: https://github.com/kumahq/kuma/blob/3aef2b26560e7634f0688065f1be630863c0eaa3/tools/resource-gen/main.go#L393

Maybe it's worth moving proto --> json api to https://github.com/pubg/protoc-gen-jsonschema/tree/main and then use this as a lib when doing openAPI generation.

lahabana:
Actually thinking about this some more I think what we can do is simply part with protobufs as soon as generation is done.

Then we can use: https://github.com/invopop/jsonschema/tree/main to generate the jsonschema.

The simple script hereafter gets me an output clearly better than what we already have:

```go
package main

import (
	"github.com/invopop/jsonschema"
	"sigs.k8s.io/yaml"

	"github.com/kumahq/kuma/api/mesh/v1alpha1"
)

func main() {

	reflector := jsonschema.Reflector{
		ExpandedStruct:            true,
		DoNotReference:            true,
		AllowAdditionalProperties: true,
	}
	schema := reflector.Reflect(&v1alpha1.Mesh{})
	//out, _ := json.MarshalIndent(schema, "", "  ")
	out, _ := yaml.Marshal(schema)

	println(string(out))
}
```

I believe once you generate a "type" the path to get to generate the whole endpoint shouldn't be too hard

lahabana:
I've played with this idea a bit and I believe we can quickly get somewhere. I pushed a branch for this: https://github.com/lahabana/kuma/commit/bdab6e587d758af07da6b093efd91128eb2d59fa

slonka:
> The simple script hereafter gets me an output clearly better than what we already have:

If you actually diff the two files [schemas.zip](https://github.com/user-attachments/files/18162373/schemas.zip) there are some things I don't like in the one generated on your branch:
- it's missing descriptions
image
- it doesn't handle enum properly
image
- doesn't contain regexes 

One thing that we definitely would want to get rid of is "additionalProperties" from this branch file - which would be quite easy with `yq`.

mk/docs.mk
slonka: yq doesn't have `walk` function which is needed to resolve `$ref`s.

tools/resource-gen/main.go
```diff
@@ -418,34 +405,99 @@ func main() {
...
+		schemaMap.Set("labels", &jsonschema.Schema{Type: "object", AdditionalProperties: &jsonschema.Schema{Type: "string"}})
+		schemaMap.Set("spec", reflector.ReflectFromType(tpe))
+
+		schema := jsonschema.Schema{
+			Type:       "object",
+			Required:   []string{"type", "name", "spec"},
+			Properties: schemaMap,
+		}
+
+		// TODO merge the schema with a global open api schema for the resource (might be worth to just make these objects
```
slonka, line 463: merging is done by openapi-tool written in node-js so not sure if anything is needed here, I'll do this in the text commit

api/mesh/v1alpha1/mesh/schema.yaml
```diff
@@ -0,0 +1,320 @@
+properties:
+  labels:
+    additionalProperties:
+      type: string
+    type: object
+  mesh:
+    type: string
```
lahabana, line 7: This should be there as it's a Global entity

api/mesh/v1alpha1/meshgateway/rest.yaml
```diff
@@ -0,0 +1,115 @@
+openapi: 3.0.3
+info:
+  version: v1alpha1
+  title: Kuma API
+  description: Kuma API
+  x-ref-schema-name: "MeshGateway"
+
+paths:
+  /meshgateways/{name}:
```
lahabana, line 9: This is missing `/meshes/{meshName}` rigth?

api/mesh/v1alpha1/meshgateway/schema.yaml
```diff
@@ -0,0 +1,140 @@
+properties:
+  labels:
+    additionalProperties:
+      type: string
+    type: object
+  name:
+    type: string
```
lahabana, line 7: Now you are missing mesh so seems like a condition may be reversed :)

api/mesh/v1alpha1/meshgateway/schema.yaml
```diff
@@ -0,0 +1,140 @@
...
+  spec:
```
lahabana, line 8: Ok one of the messy thing of the older policies was that the spec wasn't always nested inside a `spec` field.

For example `MeshGateway` on Universal is:

```
type: MeshGateway
name: the-gateway
mesh: default
labels:
  k8s.kuma.io/namespace: kuma-demo
  kuma.io/display-name: the-gateway
selectors:
  - match:
      kuma.io/service: gw-1
conf:
  listeners:
    - port: 8080
      protocol: HTTP
---
```

The same is the case with `Mesh`: https://github.com/kumahq/kuma/blob/b98468fbd6dcc43f597b99ec2f00cbda09b27599/test/framework/setup.go#L230-L251

mk/docs.mk
```diff
@@ -40,6 +40,7 @@ docs/generated/openapi.yaml:
 	for i in $(API_DIRS); do mkdir -p $(OAPI_TMP_DIR)/$$(echo $${i} | cut -d: -f2); cp -R $$(echo $${i} | cut -d: -f1) $(OAPI_TMP_DIR)/$$(echo $${i} | cut -d: -f2); done
 	for i in $$( find $(POLICIES_DIR) -name '*.yaml' | grep '/api/' | grep -v '/testdata/'); do DIR=$(OAPI_TMP_DIR)/policies/$$(echo $${i} | awk -F/ '{print $$(NF-3)}'); mkdir -p $${DIR}; cp $${i} $${DIR}/$$(echo $${i} | awk -F/ '{print $$(NF)}'); done
 	for i in $$( find $(RESOURCES_DIR) -name '*.yaml' | grep '/api/' | grep -v '/testdata/'); do DIR=$(OAPI_TMP_DIR)/resources/$$(echo $${i} | awk -F/ '{print $$(NF-3)}'); mkdir -p $${DIR}; cp $${i} $${DIR}/$$(echo $${i} | awk -F/ '{print $$(NF)}'); done
+	for i in $$( find $(MESH_API_DIR) -name '*.yaml'); do DIR=$(OAPI_TMP_DIR)/resources/$$(echo $${i} | awk -F/ '{print $$(NF-1)}'); mkdir -p $${DIR}; cp $${i} $${DIR}/$$(echo $${i} | awk -F/ '{print $$(NF)}'); done
```
lahabana, line 43:
```suggestion
	for i in $$( find $(MESH_API_DIR) -name '*.yaml'); do DIR=$(OAPI_TMP_DIR)/protoresources/$$(echo $${i} | awk -F/ '{print $$(NF-1)}'); mkdir -p $${DIR}; cp $${i} $${DIR}/$$(echo $${i} | awk -F/ '{print $$(NF)}'); done
```

I'd use its own folder just for clarity and easier debugging

tools/resource-gen/main.go
```diff
@@ -418,34 +407,120 @@ func main() {
...
+
+		templatePath := path.Join("tools", "openapi", "templates", "endpoints.yaml")
+		tmpl, err := template.ParseFiles(templatePath)
+		if err != nil {
+			return err
+		}
+		opts := map[string]interface{}{
+			"Package": "v1alpha1",
+			"Name":    r.ResourceType,
+			"Scope":   "Global",
```
lahabana, line 491: You should define scope:
```
scope := "Mesh"
if r.Global {
scope = "Global"
}
```
```suggestion
			"Scope":   scope,
```

tools/resource-gen/main.go
```diff
@@ -418,34 +407,120 @@ func main() {
...
+			AllowAdditionalProperties: true,
+		}
+		err := reflector.AddGoComments("github.com/kumahq/kuma/", "api/")
+		if err != nil {
+			return err
+		}
+		schemaMap := orderedmap.New[string, *jsonschema.Schema]()
+		schemaMap.Set("type", &jsonschema.Schema{Type: "string"})
+		schemaMap.Set("name", &jsonschema.Schema{Type: "string"})
+		if r.Global {
```
lahabana, line 453:
```suggestion
		if !r.Global {
```

tools/resource-gen/main.go
```diff
@@ -418,34 +407,120 @@ func main() {
...
+			schemaMap.Set("mesh", &jsonschema.Schema{Type: "string"})
+		}
+		schemaMap.Set("labels", &jsonschema.Schema{Type: "object", AdditionalProperties: &jsonschema.Schema{Type: "string"}})
+		schemaMap.Set("spec", reflector.ReflectFromType(tpe))
```
lahabana, line 457: As mentioned here we do the mistake of adding this to `spec` which is only true for new policies

lahabana (commented):
Very cool!

api/mesh/v1alpha1/meshgateway/schema.yaml
slonka: Universal was [decided to be out of scope for this](https://docs.google.com/document/d/1Ewe0yRx-vvAuQn5CWRFhJbBPLzdbBQ2wsQosZXk4Go8/edit?tab=t.0#heading=h.xx7061aqsfyv). Given that should I just create an issue or should we do something about it?

mk/docs.mk
slonka: changed

tools/resource-gen/main.go
slonka: let's move the conversation to that thread

api/mesh/v1alpha1/meshgateway/rest.yaml
slonka: yes, scope change fixed that

api/mesh/v1alpha1/mesh/schema.yaml
slonka: yes, scope change fixed that

tools/resource-gen/main.go
slonka: done merging the way I described above

api/mesh/v1alpha1/meshgateway/schema.yaml
slonka: it's on lines https://github.com/kumahq/kuma/pull/12272/files#diff-17ba8fbac8021b72549d3c48704eee402f7ca764a8e409622ea5f24715e7c227R6-R7

api/mesh/v1alpha1/meshgateway/schema.yaml
lahabana: But universal IS the api right? There's not rest API with universal entities including against Konnect

slonka: You're 100% right - I don't know why but I thought they were all k8s because they had spec but of course they have spec in new policies 🤷 . Changed the code to not put things in `spec`.

tools/resource-gen/main.go
```diff
@@ -418,34 +407,127 @@ func main() {
...
-		log.Fatalf("%s\n", err)
+func openApiGenerator(pkg string, resources []ResourceInfo) error {
+	protoTypeToType := map[string]reflect.Type{
+		"Mesh":        reflect.TypeOf(v1alpha1.Mesh{}),
+		"MeshGateway": reflect.TypeOf(v1alpha1.MeshGateway{}),
+	}
+
+	for _, r := range resources {
+		tpe, exists := protoTypeToType[r.ResourceType]
+		println("generating", r.ResourceType)
```
lahabana, line 437:
```suggestion
```

tools/resource-gen/main.go
```diff
@@ -418,34 +407,127 @@ func main() {
...
 		log.Fatalf("%s\n", err)
 	}
+}
 
-	if _, err := os.Stdout.Write(out); err != nil {
-		log.Fatalf("%s\n", err)
+func openApiGenerator(pkg string, resources []ResourceInfo) error {
+	protoTypeToType := map[string]reflect.Type{
+		"Mesh":        reflect.TypeOf(v1alpha1.Mesh{}),
+		"MeshGateway": reflect.TypeOf(v1alpha1.MeshGateway{}),
```
lahabana, line 432: Worth adding a comment to indicate that it's here that new types should be added if we want to generate openAPI for it

tools/resource-gen/main.go
```diff
@@ -418,34 +407,127 @@ func main() {
...
+		}
+		schemaMap.Set("labels", &jsonschema.Schema{Type: "object", AdditionalProperties: &jsonschema.Schema{Type: "string"}})
+		properties := reflector.ReflectFromType(tpe).Properties
+		for pair := properties.Oldest(); pair != nil; pair = pair.Next() {
+			schemaMap.Set(pair.Key, pair.Value)
+		}
+
+		schema := jsonschema.Schema{
+			Type:       "object",
+			Required:   []string{"type", "name", "spec"},
```
lahabana, line 464:
```suggestion
			Required:   []string{"type", "name"},
```

lahabana: Also you should add a: `if !r.Global { schema.Required = append(schema.Required, "mesh")}`

lahabana (approved):
LGTM once you've ironed out the mismatch between actual API and spec that you've discovered while testing

PR #12291, opened 2024-12-17
chore(mk): fix k3d start errors when using calico CNI

jijiechen:
## Motivation

To fix k3d start errors when using calica CNI

It's reported in CI run of the latest commit on `release-2.8` and older release versions:

https://github.com/kumahq/kuma/actions/runs/12357114240/job/34492385775

## Implementation information

During backport `\` was forgotten which caused that calico step wasn't a prerequisite for running k3d.

## Supporting documentation

Code on master: https://github.com/kumahq/kuma/pull/12277/files#diff-6a9a6d7b0c33e516a1579d4ea2ccdfdbec5bd689d4f15bcab302e2a80b0d7ee4R109

### DIFF
--- a/mk/k3d.mk
+++ b/mk/k3d.mk
@@ -106,7 +106,7 @@ k3d/cleanup-docker-credentials:
 	@rm -f /tmp/.kuma-dev/k3d-registry.yaml
 
 .PHONY: k3d/start
-k3d/start: ${KIND_KUBECONFIG_DIR} k3d/network/create k3d/setup-docker-credentials
+k3d/start: ${KIND_KUBECONFIG_DIR} k3d/network/create k3d/setup-docker-credentials \
 	$(if $(findstring calico,$(K3D_NETWORK_CNI)),$(TOP)/$(KUMA_DIR)/test/k3d/calico.$(K3D_VERSION).yaml)
 	@echo "PORT_PREFIX=$(PORT_PREFIX)"
 	@KUBECONFIG=$(KIND_KUBECONFIG) \


mk/k3d.mk
```diff
@@ -106,7 +106,7 @@ k3d/cleanup-docker-credentials:
 	@rm -f /tmp/.kuma-dev/k3d-registry.yaml
 
 .PHONY: k3d/start
-k3d/start: ${KIND_KUBECONFIG_DIR} k3d/network/create k3d/setup-docker-credentials
+k3d/start: ${KIND_KUBECONFIG_DIR} k3d/network/create k3d/setup-docker-credentials \
```
slonka, line 109: I don't understand how this helps... this is a list of target prerequisites, with `\` what is it?

lahabana: I agree. Also seeing the weirdness of the fix I think this warrants a lengthier PR description to explain how a magical `\` solves this problem

jijiechen: When it's using `calico`, we need to add another step as dependency which is defined in the next line.
it was good previously and we had a backport PR broke the declaration here yesterday. So here is to bring back the continuation char that was missing from that PR.
https://github.com/kumahq/kuma/pull/12277/files#diff-6a9a6d7b0c33e516a1579d4ea2ccdfdbec5bd689d4f15bcab302e2a80b0d7ee4R109

lukidzi: I didn't notice this. @jijiechen Thanks for spotting it

lukidzi (approved):

mk/k3d.mk
lahabana: Cool added all this info in the PR description

PR #12329, opened 2024-12-19
feat(oapi): properly marshal proto struct and set paths

slonka:
## Motivation

The build was failing in parent repo because of:

https://github.com/kumahq/kuma/blob/39a5f176c7bd1f608fdc0a6e834d33289256d108/docs/generated/openapi.yaml#L11498

Also the paths in the parent repo were not properly set

## Implementation information

We're ignoring the `structpb.Struct` so that it is effectively any JSON map.

## Supporting documentation

Fixes this error:

```
error loading swagger spec in api/openapi/specs/global_insight.yaml
: failed to load OpenAPI specification: error resolving reference "../kuma/openapi.yaml#/components/schemas/GlobalInsight": failed to unmarshal data: json error: invalid character 'o' looking for beginning of value, yaml error: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal bool into field Schema.properties of type openapi3.Schema
make: *** [kuma/mk/generate.mk:108: generate/oas] Error 1
```

> Changelog: feat(oapi): add mesh and meshgateway to oapi spec

### DIFF
--- a/api/mesh/v1alpha1/mesh/schema.yaml
+++ b/api/mesh/v1alpha1/mesh/schema.yaml
@@ -63,15 +63,7 @@ properties:
           properties:
             conf:
               description: Configuration of the backend
-              properties:
-                fields:
-                  additionalProperties:
-                    properties:
-                      Kind: true
-                    required:
-                    - Kind
-                    type: object
-                  type: object
+              properties: {}
               type: object
             format:
               description: |-
@@ -115,15 +107,7 @@ properties:
           properties:
             conf:
               description: Configuration of the backend
-              properties:
-                fields:
-                  additionalProperties:
-                    properties:
-                      Kind: true
-                    required:
-                    - Kind
-                    type: object
-                  type: object
+              properties: {}
               type: object
             name:
               description: Name of the backend, can be then used in Mesh.metrics.enabledBackend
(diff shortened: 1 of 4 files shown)


lukidzi (approved):

api/mesh/v1alpha1/mesh/schema.yaml
```diff
@@ -63,15 +63,7 @@ properties:
...
-              properties:
-                fields:
-                  additionalProperties:
-                    properties:
-                      Kind: true
-                    required:
-                    - Kind
-                    type: object
-                  type: object
+              properties: {}
```
lahabana, line 66: You may need to add additionalProperties on this.

tools/resource-gen/main.go
```diff
@@ -474,20 +479,20 @@ func openApiGenerator(pkg string, resources []ResourceInfo) error {
...
+		outDir := path.Join(rootDir, "api", pkg, "v1alpha1", strings.ToLower(r.ResourceType))
 
 		// Ensure the directory exists
 		err = os.MkdirAll(outDir, 0o755)
 		if err != nil {
 			return fmt.Errorf("failed to create directory: %w", err)
 		}
 
-		err = os.WriteFile(path.Join(outDir, "schema.yaml"), out, 0o600)
+		err = os.WriteFile(path.Join(rootDir, outDir, "schema.yaml"), out, 0o600)
```
lahabana, line 490: outDir already contains rootDir

Issue #12183, opened 2024-12-06
Consider compressing KDS admin responses


lahabana:
### What happened?

We got the error and a large dataplane:

> rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4869150 vs. 4194304)

This happened when calling the stats RPC, I guess the main reason for this is the size of the stats output.

It would be pretty straightforward to then zstd the content which would help with these large payload.

It would make sense to have a compression format flag in: https://github.com/kumahq/kuma/blob/d844aa468f596e6b3972cc2604663e7988081b6e/api/mesh/v1alpha1/kds.proto#L119-L130

https://github.com/kumahq/kuma/blob/d844aa468f596e6b3972cc2604663e7988081b6e/api/mesh/v1alpha1/kds.proto#L157-L169

https://github.com/kumahq/kuma/blob/d844aa468f596e6b3972cc2604663e7988081b6e/api/mesh/v1alpha1/kds.proto#L88-L98

From a backward compat PoV there's no problem as we required Global to be newer than Zone

lukidzi:
> Triage: we should compress them

PR #12339, opened 2024-12-19
feat(kds): use compressor to make requests and responses smaller

slonka:
## Motivation

We're seeing errors like this https://github.com/kumahq/kuma/issues/12183.

## Implementation information

According to the docs https://github.com/grpc/grpc-go/blob/master/Documentation/compression.md#compression if turned on on the client side it will automatically make server also respond with compressed responses.

image

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12183

### DIFF
--- a/pkg/kds/mux/client.go
+++ b/pkg/kds/mux/client.go
@@ -14,6 +14,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/encoding/gzip"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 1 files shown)


jijiechen (approved):

lahabana:
Note that this applies grpc compression to all of KDS. I don't think it's bad for say just that we may keep a look at CPU/mem usage when we push this in prod

slonka:
seems to work:

```
2024/12/20 10:25:34 Unary Request: , Peer: 10.42.0.1:5822, Method: /kuma.mesh.v1alpha1.GlobalKDSService/HealthCheck
2024/12/20 10:25:34 Incoming Metadata: map[:authority:[172.19.0.3:5685] client-id:[zone-renamed] content-type:[application/grpc] features:[zone-ping-health hash-suffix] grpc-accept-encoding:[gzip] kds-version:[v3] user-agent:[kds/0.0.0-preview.v074b0a955 (linux; arm64; Kuma/074b0a9) grpc-go/1.69.0-dev] version:[0.0.0-preview.v074b0a955]]
```

slonka:
> Note that this applies grpc compression to all of KDS. I don't think it's bad for say just that we may keep a look at CPU/mem usage when we push this in prod

I thought this was negligible, do you want me to hide this behind a flag? Should I add `upgrade.md` notes?

pkg/kds/mux/client.go
```diff
@@ -64,6 +65,7 @@ func (c *client) Start(stop <-chan struct{}) (errs error) {
 	}
 	dialOpts := c.metrics.GRPCClientInterceptors()
 	dialOpts = append(dialOpts, grpc.WithUserAgent(version.Build.UserAgent("kds")), grpc.WithDefaultCallOptions(
+		grpc.UseCompressor(gzip.Name),
```
slonka, line 68: PLEASE DON'T MERGE THIS, I'LL DO IT MYSELF AND WILL MONITOR FOR CPU/MEM USAGE - DON'T WANT AN OUTAGE DURING HOLIDAYS

PR #12340, opened 2024-12-19
fix(policy): use new compute for rules and fix rules intersect

### LINKED ISSUE #12273: The virtual_domains not generated correctly with AUTO_REACHABLE_SERVICES and MeshTrafficPermission
### What happened?

version: `2.9.2`

## What happened?

I eanbled the `AUTO_REACHABLE_SERVICES` feature and create 2 MTPs and found the virtual_domains was not generate correctly from the XDS configuration

## Steps to reproduce

1. create a single zone with `2.9.2`
2. enable auto_reachable_services with ENV `KUMA_EXPERIMENTAL_AUTO_REACHABLE_SERVICES`
3. Apply the following yamls:

```yaml
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: default
spec:
  meshServices:
    mode: Exclusive
  mtls:
    enabledBackend: ca-1
    backends:
    - name: ca-1
      type: builtin
---
apiVersion: kuma.io/v1alpha1
kind: HostnameGenerator
metadata:
  name: default
  namespace: kuma-system
spec:
  selector:
    meshService:
      matchLabels:
        kuma.io/origin: zone
  template: "{{ .DisplayName }}.{{ .Namespace }}.default.gg.mesh"
```

4. Create 2 applications with sidecar injected:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    kuma.io/sidecar-injection: enabled
  name: kuma-demo
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    kuma.io/sidecar-injection: enabled
  name: kuma-one
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: 2048-app
  namespace: kuma-demo
spec:
  replicas: 1
  selector:
    matchLabels:
      app: 2048-app
  template:
    metadata:
      labels:
        app: 2048-app
    spec:
      containers:
      - name: 2048-app
        image: ghcr.io/daocloud/dao-2048:v1.4.1
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: service-2048-app
  namespace: kuma-demo
spec:
  type: ClusterIP
  selector:
    app: 2048-app
  ports:
    - name: http
      appProtocol: http
      protocol: TCP
      port: 80
      targetPort: 80
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  namespace: kuma-one
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:stable
        ports:
          - containerPort: 80
            name: http-web-svc
---
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
  namespace: kuma-one
spec:
  selector:
    app: nginx
  ports:
  - name: name-of-service-port
    appProtocol: http
    protocol: TCP
    port: 80
    targetPort: http-web-svc
```

5. Create the following MTPs one by one:

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  name: mtp-allow-kuma-one
  namespace: kuma-system
  labels:
    kuma.io/mesh: default
spec:
  targetRef:
    kind: Mesh
  from:
  - targetRef:
      kind: MeshSubset
      tags:
        k8s.kuma.io/namespace: kuma-one
    default:
      action: Allow
---
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  name: mtp-allow-kuma-other-ns-and-tag
  namespace: kuma-system
  labels:
    kuma.io/mesh: default
spec:
  targetRef:
    kind: Mesh
  from:
  - targetRef:
      kind: MeshSubset
      tags:
        asdfasdfasdf: asdfasdfasdf
        k8s.kuma.io/namespace: kuma-other
    default:
      action: Allow
```

## Result

XDS configurations:

```text
        {
          "name": "kuma:dns",
          "active_state": {
            "version_info": "64c1c7c1-8609-40a6-b7c3-ea3949b850e1",
            "listener": {
              "@type": "type.googleapis.com/envoy.config.listener.v3.Listener",
              "name": "kuma:dns",
              "listener_filters": [
                {
                  "name": "envoy.filters.udp.dns_filter",
                  "typed_config": {
                    "@type": "type.googleapis.com/envoy.extensions.filters.udp.dns_filter.v3.DnsFilterConfig",
                    "stat_prefix": "kuma_dns",
                    "server_config": {
                      "inline_dns_table": {}
                    }
                  }
                }
              ],
```

### LINKED ISSUE #12319: In policy matching subset intersect is incorrect
### What happened?

Went through the [Subset Intersect](https://github.com/kumahq/kuma/blob/86593067f05038a1c7f76446a177538f3efff00f/pkg/plugins/policies/core/rules/rules.go#L182) method source codes, I found there's one case would be failed:

```text
Entry("entry 66", testCase{
  s1: []core_rules.Tag{
    {Key: "key1", Value: "val1"},
    {Key: "key2", Value: "abc"},
  },
  s2: []core_rules.Tag{
    {Key: "key2", Value: "val2"},
    {Key: "key3", Value: "val3"},
  },
  intersect: false,
}),
```

In this case, there should not be one intersection but our codes would give us `intersect: true`

Icarus9913:
## Motivation

1. When I went through the policy matching subset `Intersect` method, I found the behavior is incorrect with one case
2. Rewrite the rule's `Compute` method to use the new `ContainsElement` method instead of using the old `IsSubset` method

## Implementation information

Introduce one new type `type Element map[string]string` for differentiation by `IsSubset`.

## Supporting documentation

Fix #12319
Fix #12273


pkg/plugins/policies/core/rules/rules.go
```diff
@@ -130,6 +130,55 @@ func NewSubset(m map[string]string) Subset {
 	return s
 }
 
+type Element map[string]string
+
+func (ss Subset) ContainsElement(element Element) bool {
```
Icarus9913, line 135: @lobkovilya Is the logic of this method same as your idea?

pkg/plugins/policies/meshtrafficpermission/graph/backends/reachable_test.go
```diff
@@ -0,0 +1,420 @@
+package backends
```
Icarus9913, line 1: Refactor and format this unit test file

pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -300,6 +300,75 @@ var _ = Describe("Rules", func() {
 				s2:        []core_rules.Tag{},
 				intersect: true,
 			}),
+			Entry("entry 7", testCase{
```
Icarus9913, line 303: format the title rather than "entry *"

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -130,6 +130,55 @@ func NewSubset(m map[string]string) Subset {
...
+	// 1. find the overlaps of element and current subset
+	// 2. verify the overlaps
+	// 3. verify the left of current subset
+
+	if len(ss) == 0 {
+		return true
+	}
```
Icarus9913, line 142: Can an empty set Rule have `Action: "Allow"` ?

pkg/plugins/policies/meshtrafficpermission/graph/backends/reachable_test.go
Icarus9913: moved to rules_test.go with golden files

.golangci.yml
```diff
@@ -141,6 +141,12 @@ issues:
     - linters:
         - gocritic
       text: "appendAssign: append result not assigned to the same slice" # None of the instances of this in Kuma were bugs.
+    - linters:
+        - gocritic
+      text: "elseif: can replace 'else {if cond {}}' with 'else if cond {}'" # None of the instances of this in Kuma were bugs.
+    - linters:
+        - gosimple
+      text: "S1008: should use 'return overlapKeyCount != 0' instead of 'if overlapKeyCount == 0 { return false }; return true'" # Ignore this to better read the code
```
lahabana, line 149: Please agree with the linter rather than excluding rules

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -290,6 +360,24 @@ func (r *Rule) GetBackendRefOrigin(hash common_api.MatchesHash) (core_model.Reso
 
 type Rules []*Rule
 
+func (rs Rules) NewCompute(element Element) *Rule {
```
lahabana, line 363: Using New is odd if nothing is created in the method

lahabana (dismissed):
Seems like there's a refactor/cleanup and a bug fix. Is it possible to separate both changes as it's hard to review otherwise

pkg/plugins/policies/core/rules/rules.go
Icarus9913: yeah, I see.
This is just for testing currently, I'll rename it to `Compute` once the PR is ready

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -290,8 +356,29 @@ func (r *Rule) GetBackendRefOrigin(hash common_api.MatchesHash) (core_model.Reso
...
+	if computed != nil {
+		return pointer.To(computed.Conf.(T))
+	}
+
+	return nil
+}
+
+// LegacyCompute returns Rule for the given subset.
+// Deprecated: use Compute instead
+func (rs Rules) LegacyCompute(sub Subset) *Rule {
```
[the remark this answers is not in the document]
lahabana: Ok it's used in one place. Would it make sense to simply inline there a keep the legacy behaviour just 1 one self contained space of the code?

pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go
```diff
@@ -94,19 +95,33 @@ func (c *DeprecatedListenerConfigurer) configureRequestTimeout(routeConfiguratio
 }
 
 func (c *DeprecatedListenerConfigurer) configureRequestHeadersTimeout(hcm *envoy_hcm.HttpConnectionManager) {
-	if conf := c.getConf(c.Subset); conf != nil {
+	// For backward, once a user upgrades from an older version we shouldn't set default timeouts.
+	// Refer to https://github.com/kumahq/kuma/issues/12033
+	deprecatedGetConf := c.legacyGetConf(c.Subset)
```
lahabana, line 100: What I don't understand with this code is when do we actually get rid of this? This code explains why there's so deprecated code but there should be a moment where we can remove this code right? When? Where's the issue tracking this?

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -196,7 +239,7 @@ func (ss Subset) Intersect(other Subset) bool {
 		}
 		oTags, ok := otherByKeysOnlyPositive[tag.Key]
 		if !ok {
-			return true
+			continue
```
lobkovilya, line 242: what's the reason for changing the `Intersect` method?

pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go
```diff
@@ -94,19 +95,33 @@ func (c *DeprecatedListenerConfigurer) configureRequestTimeout(routeConfiguratio
...
+	}
+	return rules.ComputeConf[api.Conf](c.Rules, element)
+}
+
+func (c *DeprecatedListenerConfigurer) legacyGetConf(subset rules.Subset) *api.Conf {
 	if c.Rules == nil {
 		return &api.Conf{}
 	}
-	return rules.ComputeConf[api.Conf](c.Rules, subset)
+	return rules.LegacyComputeConf[api.Conf](c.Rules, subset)
```
lobkovilya, line 124: We can use new Compute here, there is no difference when used with old outbound policies

lobkovilya (changes requested):

pkg/plugins/policies/core/rules/rules.go
lobkovilya: should we remove the change from this PR then, since it's a different problem with a different PR https://github.com/kumahq/kuma/pull/12353?

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -235,6 +278,29 @@ func SubsetFromTags(tags map[string]string) Subset {
...
+	return Element{}
+}
+
+func MeshServiceElement(name string) Element {
+	return Element{mesh_proto.ServiceTag: name}
+}
+
+func MeshExternalServiceElement(name string) Element {
+	return Element{mesh_proto.ServiceTag: name}
+}
```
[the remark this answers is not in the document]
Icarus9913: Because I noticed that we have such old `Subset` codes, so I want to align with it

https://github.com/kumahq/kuma/blob/86593067f05038a1c7f76446a177538f3efff00f/pkg/plugins/policies/core/rules/rules.go#L218-L228

pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go
lobkovilya: I found out why you're getting different results with Compute and LegacyCompute, the problem is in `Element#WithKeyValue` method:
```go
func (e Element) WithKeyValue(key, value string) Element {
	if e == nil {
		e = Element{}
	}
	
	e[key] = value
	return e
}
```
[you're modifying the underlying map `c.Element`](https://github.com/kumahq/kuma/blob/0485cd01ccb6efadda766185180ab0e3beecba04/pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go#L80), but what we probably want is:
```go
func (e Element) WithKeyValue(key, value string) Element {
	c := maps.Clone(e)
	c[key] = value
	return c
}
```
If you fix `WithKeyValue` as I suggest then you won't see the difference between `ComputeConf` and `LegacyComputeConf`. So we can safely remove the legacy version

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -196,7 +248,7 @@ func (ss Subset) Intersect(other Subset) bool {
 		}
 		oTags, ok := otherByKeysOnlyPositive[tag.Key]
 		if !ok {
-			return true
+			continue
```
lobkovilya, line 251: Please add the following comment to the `Intersect` func:

```go
// Intersect returns true if there exists an element that belongs both to 'other' and current set.
// Empty set intersects with all sets.
// 
// We're using this function to check if 2 'from' rules of MeshTrafficPermission can be applied to the same client DPP.
// For example:
//
// from:
//   - targetRef:
//       kind: MeshSubset
//       tags:
//         team: team-a
//   - targetRef:
//       kind: MeshSubset
//       tags:
//         zone: east
// 
// there is a DPP with tags 'team: team-a' and 'zone: east' that's subjected to both these rules. 
// So 'from[0]' and 'from[1]' have an intersection.
// However, in another example:
// 
// from:
//   - targetRef:
//       kind: MeshSubset
//       tags:
//         team: team-a
//   - targetRef:
//       kind: MeshSubset
//       tags:
//         team: team-b
//         zone: east
//
// there is no DPP that'd hit both 'from[0]' and 'from[1]'. So in this case they don't have an intersection.
func (ss Subset) Intersect(other Subset) bool {
```

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -179,6 +233,40 @@ func isSubset(t1, t2 Tag) bool {
...
+  - targetRef:
+      kind: MeshSubset
+      tags:
+        team: team-a
+  - targetRef:
+      kind: MeshSubset
+      tags:
+        team: team-b
+        zone: east
+*/
```
lobkovilya, line 267: can we use single line `//` comment on all lines here? multiline comment looks a bit odd when combined with single line comments

pkg/plugins/policies/core/rules/rules.go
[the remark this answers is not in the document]
lobkovilya: looks like MeshExternalServiceElement is not used in this PR, so just remove it

pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -366,16 +482,118 @@ var _ = Describe("Rules", func() {
...
+				contains: false,
+			}),
+			Entry("no rules matched, rule with negation, element has another key", testCase{
+				ss: []core_rules.Tag{
+					{Key: "key1", Value: "val1", Not: true},
+				},
+				other: core_rules.Element{
+					"key2": "val2",
+				},
+				contains: false,
```
lobkovilya, line 554: I think this one probably should be `true`. If I have a rule:
```
rules:
  subset:
    - key: kuma.io/zone
      value: us-east-2
      not: true
  conf: $conf1
```

and I have a DPP with
```
inbound:
  - tags:
      kuma.io/service: frontend
```

Then we still want `$conf1` to be applied for this DPP right?

Icarus9913: Yes, you're right. 
This case should match it, thanks for reminding.

pkg/plugins/policies/core/rules/rules.go
lobkovilya: That's odd, `make check` works for me as expected with `//`. If you don't mind I can push my change to your branch. Maybe your IDE somehow changes the indentation

lobkovilya (approved):
LGTM

jijiechen (approved):

PR #12353, opened 2024-12-20
fix(policy): fix rules intersect when 2 tags with same key but different value

### LINKED ISSUE #12319: In policy matching subset intersect is incorrect

Icarus9913:
## Motivation

When I went through the policy matching subset Intersect method, I found the behavior is incorrect with one case

## Supporting documentation

Fix #12319

### DIFF
--- a/pkg/plugins/policies/core/rules/rules_test.go
+++ b/pkg/plugins/policies/core/rules/rules_test.go
@@ -241,58 +241,174 @@ var _ = Describe("Rules", func() {
 			func(given testCase) {
 				Expect(given.s1.Intersect(given.s2)).To(Equal(given.intersect))
 			},
-			Entry("entry 1", testCase{
+			Entry("positive, same key and value", testCase{
 				s1: []core_rules.Tag{
-					{Key: "service", Value: "backend"},
-				},
-				s2: []core_rules.Tag{
-					{Key: "service", Not: true, Value: "frontend"},
-					{Key: "version", Value: "v2"},
-				},
-				intersect: true,
-			}),
-			Entry("entry 2", testCase{
-				s1: []core_rules.Tag{
-					{Key: "service", Value: "backend"},
+					{Key: "service", Value: "frontend"},
 				},
 				s2: []core_rules.Tag{
 					{Key: "service", Value: "frontend"},
-					{Key: "version", Value: "v2"},
+				},
+				intersect: true,
+			}),
+			Entry("positive, same key, different value", testCase{
+				s1: []core_rules.Tag{
+					{Key: "service", Value: "frontend"},
+				},
+				s2: []core_rules.Tag{
+					{Key: "service", Value: "backend"},
 				},
 				intersect: false,
 			}),
-			Entry("entry 3", testCase{
+			Entry("positive, multiple key-values with overlap key-value", testCase{
 				s1: []core_rules.Tag{
-					{Key: "service", Not: true, Value: "backend"},
+					{Key: "service", Value: "frontend"},
+					{Key: "version", Value: "v1"},
 				},
 				s2: []core_rules.Tag{
-					{Key: "service", Not: true, Value: "backend"},
-					{Key: "version", Value: "v2"},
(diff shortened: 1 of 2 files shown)


pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -241,37 +241,128 @@ var _ = Describe("Rules", func() {
...
+			Entry("positive, same key and value", testCase{
+				s1: []core_rules.Tag{
+					{Key: "service", Value: "frontend"},
+				},
+				s2: []core_rules.Tag{
+					{Key: "service", Value: "frontend"},
+				},
+				intersect: true,
+			}),
+			Entry("positive, same key, different value", testCase{
```
slonka, line 253: what does positive mean?

pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -241,37 +241,128 @@ var _ = Describe("Rules", func() {
...
+				intersect: false,
+			}),
+			Entry("positive, different key, different value", testCase{
+				s1: []core_rules.Tag{
+					{Key: "service", Value: "frontend"},
 				},
 				s2: []core_rules.Tag{
+					{Key: "version", Value: "v1"},
+				},
+				intersect: true,
```
slonka, line 289: how does this intersect? 🤔

slonka (commented):
I'll be most comfortable with @lobkovilya reviewing this or I need a call to understand some aspects :(

pkg/plugins/policies/core/rules/rules_test.go
Icarus9913: It's the opposite of negation,
We have the structure:

```text
type Tag struct {
	Key   string
	Value string
	Not   bool
}
```

From the codes and comments, I saw some places define it as `negation` with `Tag.Not==true`. 
So, I say it `positive` with `Tag.Not==false`

pkg/plugins/policies/core/rules/rules_test.go
Icarus9913: Great question! This is exactly what confused us so much when we was reading the code initially.

`Intersect` here means([from the code's comments](https://github.com/kumahq/kuma/blob/86593067f05038a1c7f76446a177538f3efff00f/pkg/plugins/policies/core/rules/rules.go#L180-L181)):

```text
// Intersect returns true if there exists an element that matches both tag sets.
// Empty set intersects with all sets.
```

So, the dataplane with following tags will match both the tags in the test case:

```text
service: frontend
version: v1
```

pkg/plugins/policies/core/rules/rules_test.go
Icarus9913: This concept comes from the original code: https://github.com/kumahq/kuma/blob/86593067f05038a1c7f76446a177538f3efff00f/pkg/plugins/policies/core/rules/rules.go#L238-L247

lobkovilya (changes requested):
I think we don't need this change, explained in the issue https://github.com/kumahq/kuma/issues/12319#issuecomment-2559333065

Icarus9913:
New comment from the issue: https://github.com/kumahq/kuma/issues/12319#issuecomment-2560013880

Icarus9913:
Close it, because we have a talk that this fix belongs to PR 12340

PR #12414, opened 2025-01-02
feat(kuma-cp): add isProxy flag to resource descriptor

Automaat:
## Motivation

We need to include proxy type in resource to be get rid of weird one to one mappings and to be able to use it as label for policy matching needed for redesigned Inbound policies


lobkovilya:
I just realised, what we're going to do with gateway? Both `gateway` and `dataplane` use `Dataplane` resource

api/mesh/v1alpha1/zone_ingress.proto
```diff
@@ -23,6 +23,7 @@ message ZoneIngress {
   option (kuma.mesh.resource).scope_namespace = true;
   option (kuma.mesh.resource).has_insights = true;
   option (kuma.mesh.resource).short_name = "zi";
+  option (kuma.mesh.resource).proxy_type = "ingress";
```
lobkovilya, line 26: I'm still not sure if it's a good idea to stick to old values like `ingress` and `egress`. Now they're user facing and it might be difficult for users to associate `ingress` with `ZoneIngress` type.

Should we go with `ZoneIngress` and `ZoneEgress` values?

lobkovilya:
> I just realised, what we're going to do with gateway? Both gateway and dataplane use Dataplane resource

Probably instead of `proxyType` field we should go with `IsProxy bool` and then create a method on DataplaneResource, ZoneIngressResource and ZoneEgressResource that returns the actual type:

```go
func (t *DataplaneResource) GetProxyType() ProxyType {
    // check t.Spec and return either `Sidecar` or `Gateway`
}
```

api/mesh/v1alpha1/zone_ingress.proto
Automaat: If we want to use form `ZoneIngress` instead of `ingress` then maybe we can just don't add this proxy type field and just use `ResourceTypeDescriptor.name`. Also for a dataplane should we use `Sidecar`as we now use in policy `proxyType` field. 

Also I am not sure if can that easily change values of `ProxyType` we use atm, should we then have new type for new ProxyType values?

pkg/core/resources/apis/mesh/dataplane_helpers.go
```diff
@@ -260,3 +260,11 @@ func (d *DataplaneResource) AsOutbounds(resolver core_model.LabelResourceIdentif
 	}
 	return outbounds
 }
+
+func (d *DataplaneResource) GetProxyType() mesh_proto.ProxyType {
+	spec := d.GetSpec().(*mesh_proto.Dataplane)
+	if spec.IsDelegatedGateway() || spec.IsBuiltinGateway() {
```
lobkovilya, line 266: we agreed in the madr that delegated is going to have type `Sidecar`, that's how `targetRef.proxyType` field works today

Automaat: changed

api/common/v1alpha1/ref.go
```diff
@@ -38,8 +38,10 @@ var order = map[TargetRefKind]int{
 type TargetRefProxyType string
 
 var (
-	Sidecar TargetRefProxyType = "Sidecar"
-	Gateway TargetRefProxyType = "Gateway"
+	Sidecar     TargetRefProxyType = "Sidecar"
+	Gateway     TargetRefProxyType = "Gateway"
+	ZoneIngress TargetRefProxyType = "ZoneIngress"
+	ZoneEgress  TargetRefProxyType = "ZoneEgress"
```
lobkovilya, line 44: Once we merge this PR, it shouldn't be possible to use the `ZoneIngress` and `ZoneEgress` types in `targetRef.proxyTypes`. Do we need to add validation somewhere? 

Or should we create a new type with 4 values (maybe in a different package) and deprecate `TargetRefProxyType`?

Automaat: right, I've forgot to check usages of this properly, I've added validation to disallow using new types

lobkovilya: Please create a new type with 4 values, probably in `dataplane_helpers.go` the type is `ProxyTypeLabelValue`

lobkovilya (approved):

PR #12423, opened 2025-01-06
ci(github): remove backport and checklist comment

lahabana:
## Motivation

`pull_request_target` is too hard to use correctly, we should have workflows that rely on it

## Implementation information

Remove the backport workflow and the checklist PR comment.
We will find better ways to implement these that don't require `pull_request_target`


.github/workflows/check.yaml
```diff
@@ -1,6 +1,6 @@
 name: "PR health"
 on:
-  pull_request_target:
+  pull_request:
     # !!!! Be especially careful with checkouts are we are using: pull_request_target
```
lobkovilya, line 4: nit: probably worth updating the comment

lobkovilya (approved):

PR #12444, opened 2025-01-06
ci(github): add PR checklist comment as a cron

lahabana:
## Motivation

Instead of using a potentially dangerous pull_request_target We use a cron that looks at recently updated PRs


lobkovilya (commented):
LGTM

PR #12451, opened 2025-01-07
Revert "chore(sync): sync errors schema"

bartsmykla:
Reverts kumahq/kuma#12416

It doesn't work correctly and it breaks `generate/oas` make target


slonka:
This change does nothing because of https://github.com/kumahq/ci-tools/issues/58

PR #12452, opened 2025-01-07
ci(github): don't fail pr-modification on empty matrix

lahabana:
## Motivation

The action would fail if there were no recently updated PR.
We skip the job if we didn't find any PRs.

## Supporting documentation

A failing action: https://github.com/kumahq/kuma/actions/runs/12648833795

### DIFF
--- a/api/openapi/specs/api.yaml
+++ b/api/openapi/specs/api.yaml
@@ -390,7 +390,7 @@ components:
           type: integer
           description: Number of meshes
           example: 3
-    GlobalInsight:
+    schemas-GlobalInsight:
       type: object
       title: GlobalInsight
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-modification.yaml
```diff
@@ -47,6 +47,7 @@ jobs:
   pr-comment:
     needs: find-prs
     runs-on: ubuntu-24.04
+    if : ${{ needs.find-prs.outputs.recent_prs != '[]' }}
```
bartsmykla, line 50:
```suggestion
    if: needs.find-prs.outputs.recent_prs != '[]'
```

PR #12453, opened 2025-01-07
feat(kuma-cp): add pod labels on dataplane and use proxy type labels

### LINKED ISSUE #12357: Make sure Dataplane `labels` are complete
### Description

* add labels from the pod
* add `kuma.io/proxy-type` 

check that during the migration old DPPs get all the new labels. 
Do we take all labels and put into inbound tags today?

Automaat:
## Motivation

We need to have pod labels on dataplane resource to be able to select dataplane in policies by labels

This needs  https://github.com/kumahq/kuma/pull/12414 to be mergd first

## Supporting documentation

Fix #12357


pkg/core/resources/model/resource.go
```diff
@@ -507,8 +507,7 @@ func GetNamespace(rm ResourceMeta, systemNamespace string) Namespace {
 }
 
 func ComputeLabels(
-	rd ResourceTypeDescriptor,
-	spec ResourceSpec,
+	resource Resource,
```
lobkovilya, line 510: it's better to keep it as `rd` and `spec`, otherwise we'd have to guard `resource.GetMeta()` to not be `nil`, see the [comment](https://github.com/kumahq/kuma/pull/11449#discussion_r1764678386)

pkg/plugins/runtime/k8s/controllers/pod_converter.go
```diff
@@ -60,10 +60,18 @@ func (p *PodConverter) PodToDataplane(
 		return err
 	}
 	// we need to validate if the labels have changed
+	dataplaneRes := core_mesh.NewDataplaneResource()
+	err = dataplaneRes.SetSpec(dataplaneProto)
+	if err != nil {
+		return err
+	}
+	if err := p.ResourceConverter.ToCoreResource(dataplane, dataplaneRes); err != nil {
+		logger.Error(err, "unable to convert ZoneEgress k8s object into core resource")
```
lobkovilya, line 69: why ZoneEgress?

Automaat: no longer valid

pkg/core/resources/model/resource.go
Automaat: changed

pkg/core/resources/model/resource.go
```diff
@@ -566,6 +566,13 @@ func ComputeLabels(
 		labels[mesh_proto.PolicyRoleLabel] = string(role)
 	}
 
+	if rd.IsProxy {
+		proxy, ok := resource.(ProxyResource)
+		if ok {
+			labels[mesh_proto.ProxyTypeLabel] = strings.ToLower(string(proxy.GetProxyType()))
```
lobkovilya, line 572: You can place `GetProxyType` on the ResourceSpec (not on the `Resource`). We already have an interface for policies https://github.com/lobkovilya/kuma/blob/73809a37c7835c8217880f3f8f6a7e0c786834ad/pkg/core/resources/model/resource.go#L730-L733, you can do the same for Proxy

Automaat: Right we can do this, updated

lobkovilya (approved):

PR #12468, opened 2025-01-08
ci(github): use include in pr-modification.yaml

lahabana:
## Motivation

This was causing issues when there were no pr modified.

## Implementation information

 The indirection with include was hard to read so we removed it


lahabana:
Failling action: https://github.com/kumahq/kuma/actions/runs/12668031147
Fixed action: https://github.com/kumahq/kuma/actions/runs/12668034896

bartsmykla (approved):

PR #12477, opened 2025-01-08
ci(gha): reduce permissions in workflows based on scorecard alerts

bartsmykla:
## Motivation

This change reduces GitHub Actions workflow permissions to follow the principle of least privilege, addressing security alerts from Scorecard and improving overall workflow security.

## Implementation information

- `_build_publish.yaml`:
  - Moved `id-token: write` to `build-images` job
  - Removed global `id-token: write` permission

- `_provenance.yaml`:
  - Added `permissions: {}` globally
  - Set specific permissions (`contents`, `actions`, `id-token`, `packages`) per job

- `build-test-distribute.yaml`:
  - Removed global permissions
  - Added job-specific permissions for `checks`, `contents`, `id-token`, and `packages`

- `ci-stability.yaml`:
  - Added explicit `permissions: {}` globally

> Changelog: skip

### DIFF
--- a/.github/workflows/build-test-distribute.yaml
+++ b/.github/workflows/build-test-distribute.yaml
@@ -7,10 +7,7 @@ on:
     branches: ["master", "release-*"]
   workflow_dispatch: # Allows manual trigger from GitHub Actions UI or via REST call
 permissions:
-  contents: write # To upload assets
-  id-token: write # For using token to sign images
-  actions: read # For getting workflow run info to build provenance
-  packages: write # Required for publishing provenance. Issue: https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/container#known-issues
+  contents: read
 env:
   KUMA_DIR: "."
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test-distribute.yaml
```diff
@@ -23,9 +20,8 @@ concurrency:
 jobs:
   check:
     permissions:
-      contents: read
-      # golangci-lint-action
-      checks: write
+      checks: write # needed for golangci/golangci-lint-action to add code annotations in PRs
+      contents: write # needed for sca-project to upload sbom to release assets
```
lukidzi, line 24: We had this `read` and it worked?

lukidzi (approved):

.github/workflows/build-test-distribute.yaml
bartsmykla: You are right, it's unnecessary as we are not using this job to publish SBOM/CVE reports to releases - removed

lukidzi (approved):

PR #12478, opened 2025-01-08
chore(gha): add read pull requests permissions

lukidzi:
## Motivation

Job reads PRs in the first step so permission should be clear

## Supporting documentation

Fix #XX

### DIFF
--- a/.github/workflows/pr-modification.yaml
+++ b/.github/workflows/pr-modification.yaml
@@ -10,6 +10,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: read
 jobs:
   find-prs:
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-modification.yaml
```diff
@@ -10,6 +10,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: read
```
lahabana, line 13: Shouldn't it be in the job that needs the extra perms?

lahabana (approved):

jakubdyszkiewicz (approved):

bartsmykla (approved):

PR #12497, opened 2025-01-09
fix(oapi): add missing bits to fix validation errors in parent project

slonka:
## Motivation

In parent project we were getting errors like: 
- "delete must include 404" 
- "operationId must be defined" 
- "required: true missing"
- property must have required property "createdAt"
- property must have required property "labels"

## Implementation information

So I'm defining it here.

## Supporting documentation

xrel https://github.com/Kong/kong-mesh/issues/7202

> Changelog: skip


api/openapi/specs/common/jsonpatch.yaml
```diff
@@ -15,5 +15,7 @@ components:
           type: string
           description: A JSON Pointer path indicating the part of the document to operate on.
         value:
-          AnyValue:
-            description: The value to be used within the operations.
```
slonka, line 19: this is not a valid schema:

```
Error:   3413:20    error  oas3-schema                    Property "AnyValue" is not expected to be here.                                                                                            components.schemas.JsonPatchItem.properties.value.AnyValue
```

api/openapi/specs/common/resource.yaml
```diff
@@ -74,7 +74,6 @@ components:
           example: MeshRetry
           description: the type of the policy
         proxyRule:
-          description: a rule that affects the entire proxy
```
slonka, line 77: neither is this:

```
Error:   3354:24    error  no-$ref-siblings               $ref must not be placed next to any other properties                                                                                       components.schemas.InspectRule.properties.proxyRule.description
```

api/mesh/v1alpha1/mesh/rest.yaml
```diff
@@ -8,6 +8,7 @@ info:
 paths:
   /meshes/{name}:
     get:
+      operationId: get-Mesh
```
lobkovilya, line 11: nit: mixing dashes and capitalised letters looks a bit unusual, is there a style guide? can we use `getMesh` or `get-mesh` instead?

slonka: that would have to be an additional field here https://github.com/kumahq/kuma/blob/3aef2b26560e7634f0688065f1be630863c0eaa3/pkg/plugins/policies/meshaccesslog/api/v1alpha1/zz_generated.resource.go#L136 or adding a template function like "lowercase" or I can use shortname https://github.com/kumahq/kuma/blob/3aef2b26560e7634f0688065f1be630863c0eaa3/pkg/plugins/policies/meshaccesslog/api/v1alpha1/zz_generated.resource.go#L156

lobkovilya: I think we already [guarantee](https://github.com/kumahq/kuma/blob/3aef2b26560e7634f0688065f1be630863c0eaa3/pkg/core/resources/registry/registry.go#L100) that `ShortName` is unique, so I'd just use it

api/openapi/specs/api.yaml
```diff
@@ -514,20 +531,35 @@ components:
...
+        dataplanes:
+          standard:
+            total: 4
+            online: 2
+            offline: 1
+            partiallyDegraded: 1
+          online: 23
+          offline: 10
+          partiallyDegraded: 17
+          total: 51
```
Icarus9913, line 543: Does `dataplanes` have those properties?
https://github.com/kumahq/kuma/blob/7e4a0efa1567762a63aef57f598eabcecba4a5c0/api/openapi/specs/api.yaml#L349-L361

api/openapi/specs/common/error_schema.yaml
```diff
@@ -54,4 +54,26 @@ components:
...
+            status:
+              example: 404
+            title:
+              example: Not Found
+            type:
+              example: https://httpstatuses.com/404
+            instance:
+              example: kong:trace:1234567890
+            detail:
+              example: Not found
```
Icarus9913, line 72: Do we need to supplement `type: integer/string` for those properties?

api/openapi/specs/api.yaml
slonka: Yes, otherwise the linter would've not complained.

api/mesh/v1alpha1/mesh.proto
```diff
@@ -20,6 +20,7 @@ message Mesh {
   option (kuma.mesh.resource).kds.send_to_zone = true;
   option (kuma.mesh.resource).ws.name = "mesh";
   option (kuma.mesh.resource).ws.plural = "meshes";
+  option (kuma.mesh.resource).short_name = "m";
```
slonka, line 23: Can this break on-wire compatibility? Like if a global is using X and user is using X-1?

api/openapi/specs/api.yaml
slonka: Good point, it was a leftover.

api/openapi/specs/common/error_schema.yaml
slonka: This is taken directly from common (platform) definitions, in order for this to properly dedup I can't change anything.

api/mesh/v1alpha1/mesh/rest.yaml
slonka: used "shortName".

api/mesh/v1alpha1/mesh.proto
slonka: Seems like a similar PR previously did not change anything special or add UPGRADE.md notes so I think it's fine https://github.com/kumahq/kuma/pull/12109/files#diff-ebf17957528e5b0c92852cadc86a5c16e9ebfd594731b109e52f82c7cce59811

lobkovilya: yeah I think we never use these `(kuma.mesh.resource)` in KDS right? It's probably used only when we're generating ResourceTypeDescriptor. So should be safe without UPGRADE.md

lobkovilya (approved):

PR #12506, opened 2025-01-10
ci(gha): update codeql workflow to set checks during scheduled runs

bartsmykla:
## Motivation

Addressing security alert

## Implementation information

Updated CodeQL GitHub Actions workflow to set the `CodeQL` check on the commit being analyzed during scheduled runs, ensuring some commits are marked as having SAST runs by Scorecard


.github/workflows/codeql.yaml
```diff
@@ -1,16 +1,28 @@
 name: "CodeQL"
+
 on:
-  workflow_dispatch: {}
+  push:
+    branches: ["master", "release-*", "!*-merge-master"]
```
lukidzi, line 5: is it needed? `"!*-merge-master"`, also should we check this only on master?

lahabana: Can we just change the job to create a bracnh `release-*/merge-master` or `merge-release-*-to-master`. The bad naming of the branch is the problem here

bartsmykla: Removed this from the current workflow and opened https://github.com/kumahq/kuma/issues/12507 to follow up with the mentioned change

lahabana:
The reason it's not done a PRs is the length the job it takes: https://github.com/kumahq/kuma/pull/8787

I'm afraid adding this back will just cause job queueing.

michaelbeaumont (approved):

.github/workflows/codeql.yaml
```diff
@@ -1,24 +1,27 @@
...
 jobs:
   analyze:
-    timeout-minutes: 30
-    name: Analyze
     runs-on: ubuntu-24.04
+    timeout-minutes: 30
     permissions:
+      contents: read
       actions: read
       security-events: write
```
lahabana, line 17: do you need `checks: write` ?

lahabana (approved):
One nit around perms but LGTM otherwise

PR #12513, opened 2025-01-10
ci(gha): improve CodeQL

bartsmykla:
Currently it's not reviewable as it's only for tests

### DIFF
--- a/.github/workflows/_build_publish.yaml
+++ b/.github/workflows/_build_publish.yaml
@@ -44,6 +44,10 @@ jobs:
   build-binaries:
     timeout-minutes: 40
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      actions: read
+      security-events: write
     outputs:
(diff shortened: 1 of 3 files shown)


.github/workflows/_build_publish.yaml
```diff
@@ -60,8 +64,20 @@ jobs:
           key: ${{ runner.os }}-${{ runner.arch }}-devtools-${{ hashFiles('mk/dependencies/deps.lock') }}
           restore-keys: |
             ${{ runner.os }}-${{ runner.arch }}-devtools
+      - name: Initialize CodeQL
```
lahabana, line 67: Shouldn't it be in check instead?

bartsmykla: The most time takes the build step which here is already done. If we'll put it in `check` we'll have to build binaries there as well which will take 10+ more minutes

lahabana:
@bartsmykla what's the future of this PR?

lahabana:
@bartsmykla what future here?

Issue #12529, opened 2025-01-13
Move from dependabot to renovate


bartsmykla:
### Description

We should start using renovate instead of dependabot

michaelbeaumont:
I like it but for the record, why?

bartsmykla:
@michaelbeaumont there have been a lot of issues with Dependabot lately, for example:
- https://github.com/dependabot/dependabot-core/issues/11215
- https://github.com/dependabot/dependabot-core/issues/11253
- Our security team flagged problems with Dependabot when updating Dockerfiles
- While testing Renovate, I noticed many packages in `go.mod` hadn't been updated by Dependabot in a long time, even though they should have been
- It's not trivial to configure Dependabot to cover release branches, while in Renovate, they can simply be added to the `"baseBranches": []` config

PR #12546, opened 2025-01-14
ci(renovate): add initial renovate.json configuration file

bartsmykla:
## Motivation

We want to move from dependabot to renovate

## Implementation information

This file is not final, but it should be a good base. I can't find any builtin way to dynamically provide `baseBranches` or read them from `active-branches.json` file, so our release scripts should be updated to update them on the same tame as `active-branches.json`.

## Supporting documentation

Part of https://github.com/kumahq/kuma/issues/12529

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,41 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "schedule": "before 5am every weekday",
+  "extends": [
+    "config:best-practices",
+    ":automergeDisabled",
+    ":gitSignOff"
+  ],
+  "baseBranches": [
+    "release-2.6",
+    "release-2.7",
+    "release-2.8",
+    "release-2.9",
+    "master"
+  ],
+  "enabledManagers": [
+    "dockerfile",
+    "github-actions",
+    "gomod",
+    "helm-values"
+  ],
+  "ignorePaths": [
+    "(^|/)([Dd]ocker|[Cc]ontainer)file[^/]*\\.dockerignore$"
+  ],
+  "labels": [
+    "dependencies"
+  ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -0,0 +1,41 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "schedule": "before 5am every weekday",
+  "extends": [
+    "config:best-practices",
+    ":automergeDisabled",
+    ":gitSignOff"
+  ],
+  "baseBranches": [
```
lukidzi, line 9: Should we try first only master and selectively upgrade on branches later?

bartsmykla: we can

lukidzi (approved):

renovate.json
```diff
@@ -0,0 +1,37 @@
...
+    {
+      "description": "go-control-plane v0.12.0 introduced a potential deadlock issue. This issue is being tracked in https://github.com/envoyproxy/go-control-plane/issues/875. Remove this once the issue is resolved",
+      "enabled": false,
+      "matchManagers": ["gomod"],
+      "matchPackageNames": ["github.com/envoyproxy/go-control-plane"]
+    },
+    {
+      "description": "No need to run tests on github actions updates",
+      "matchManagers": ["github-actions"],
+      "addLabels": ["ci/skip-test"]
```
slonka, line 34: Why?

bartsmykla: I haven't spend time thinking about it, but I tried to use the same configuration as for dependabot: https://github.com/kumahq/kuma/blob/84707bb2b53296092dc5423c1a287485415cbd67/.github/dependabot.yml#L42

slonka (approved):

PR #12549, opened 2025-01-14
fix(mesh-insights): skip error on mesh insight creation race condition

Automaat:
## Motivation

We saw this issue multiple times on live environment:

```
unable to resync MeshInsight: failed to execute query: INSERT INTO resources (name, mesh, type, version, spec, creation_time, modification_time, owner_name, owner_mesh, owner_type, labels, status) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12);: ERROR: insert or update on table "resources" violates foreign key constraint "owner_fk" (SQLSTATE 23503)
```

## Implementation information

We probably can just skip this error, as it won't be fixed with retry.

### DIFF
--- a/pkg/insights/resyncer.go
+++ b/pkg/insights/resyncer.go
@@ -29,11 +29,12 @@ import (
 var log = core.Log.WithName("mesh-insight-resyncer")
 
 const (
-	ReasonResync    = "resync"
-	ReasonForce     = "force"
-	ReasonEvent     = "event"
-	ResultChanged   = "changed"
-	ResultNoChanges = "no_changes"
+	ReasonResync             = "resync"
+	ReasonForce              = "force"
+	ReasonEvent              = "event"
+	ResultChanged            = "changed"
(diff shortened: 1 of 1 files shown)


slonka (approved):

slonka (commented):
I'm not 100% sure but I think using a transaction with proper isolation levels would lead to delete operation being blocked until the upsert transaction is done and not resulting in foreign key violation. Not sure if it's worth it in this case.

Automaat:
The other solution we've discussed with @slonka is that we could utilize transaction here, but default transaction level `read commited` would result the same issue, as mesh can be removed in the meantime. We could increase the isolation level to `REPEATABLE READ` or `SERIALIZABLE` but in my opinion we don't need such strong guarantees. If mesh is not present, we don't need mesh insights. We can simply skip the error and don't care about it.

lobkovilya (approved):
Ignoring the error seems like a good approach

PR #12559, opened 2025-01-15
refactor(rules): organize code in the packages, prepare for inbound rules

lobkovilya:
## Motivation

* with the introduction of inbound rules, a single file `rules.go` is insufficient
* new inbound rules are going to introduce a new entry type (i.e. `to[i]` has `targetRef` and `default` fields, while `rules[i]` is going to have `default`, `matches`, `targetRef` fields). It's easier to use generics, like 
    ```go
    type WithPolicyAttributes[T any] struct {
    	Entry T

    	TopLevel  common_api.TargetRef
    	Meta      core_model.ResourceMeta
    	RuleIndex int
    }
    ```
    to share the code between inbound and outbound rules (see packages `merge`, `sort`, `common`)

## Supporting documentation

Check [commit](https://github.com/kumahq/kuma/commit/450bb85eb1c8bfdff090bd4e365a9d4c8d13e78a) to see how new inbound rules are implemented


pkg/plugins/policies/core/rules/subsetutils/subset.go
```diff
@@ -0,0 +1,339 @@
+package subsetutils
```
lobkovilya, line 1: In this file are just copies of functions we had in `rules.go`, you can skip reviewing this

pkg/plugins/policies/core/rules/outbound/resourcerules.go
```diff
@@ -109,34 +81,81 @@ func (rr ResourceRules) Compute(uri core_model.TypedResourceIdentifier, reader R
...
-	SortByTargetRefV2(list)
+	Sort(list)
 
-	var resolvedItems []*resolvedPolicyItem
+	var resolvedItems []*withResolvedResource[T]
 	for _, item := range list {
 		resolvedItems = append(resolvedItems, resolveTargetRef(item, reader)...)
 	}
 
+	indexResources := func() map[core_model.TypedResourceIdentifier]core_model.Resource {
```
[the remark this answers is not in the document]
lobkovilya: I wanted to inline the function to not overload it with generics as it's very small and used only once. I changed the function with a simple for-loop, wdyt?

Automaat (dismissed):

pkg/plugins/policies/core/rules/resourcerules.go
lahabana, line 1: Can we do proper `git mv` ? This would greatly simplify this PR

Automaat (approved):

Issue #12382, opened 2024-12-23
Shared algo for all policies that supports `spec.rules` 


lobkovilya:
### Description

We need an algorithm that can be called from all policy plugins to compute configuration from multiple policies with `spec.rules`.

This doesn't cover:
* `rules[].matches`
* `rules[].targetRef`

PR #12560, opened 2025-01-15
feat(rules): algo for inbound rules

lobkovilya:
## Implementation information

* add a new entry type `RuleEntry` and algo to compute resulting configuration 
* introduce descriptor flag `InterpretFromEntriesAsRules` and set it to `true` for MeshTimeout (and potentially for all policies that support only `Mesh` in `spec.from[].targetRef.kind`)

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12382


pkg/plugins/policies/core/rules/rules.go
```diff
@@ -44,7 +45,8 @@ func (i InboundListener) String() string {
 }
 
 type FromRules struct {
-	Rules map[InboundListener]Rules
+	Rules    map[InboundListener]Rules
+	RulesNew map[InboundListener][]*inbound.Rule
```
Automaat, line 49: can we try finding better name than `RulesNew`? What if we add new new rules in the future?

pkg/plugins/policies/core/matchers/testdata/egressmatchedpolicies/torules/01.golden.yaml
```diff
@@ -39,3 +39,4 @@ Rules:
     - Key: version
       Not: true
       Value: v2
+RulesNew: null
```
Automaat, line 42: should this appear in toRules?

lobkovilya: yes, the test checks for `policies.FromRules` https://github.com/kumahq/kuma/blob/3034ff67fb6a814242083fda083b66509598a21c/pkg/plugins/policies/core/matchers/egress_test.go#L109

it's egress, so toRules here are converted to rules per listener

pkg/plugins/policies/core/rules/rules.go
lobkovilya: Agreed, renamed to `InboundRule` and marked old `Rules` as deprecated

Automaat (approved):

PR #12562, opened 2025-01-15
chore(oapi): add x-speakeasy-entity-operation to all operations

slonka:
## Motivation

So that we can create a terraform provider out of it.

## Implementation information

Just add it in the template.

## Supporting documentation

xrel: https://github.com/Kong/kong-mesh/issues/7201


slonka:
Probably not the best place to add it.

lukidzi (approved):

slonka:
I'm merging this so that I can move forward, I can always revert and add this somewhere else 🤷

Issue #12142, opened 2024-12-02
Fix flaking multizone e2e_env test: TrafficRoute locality aware loadbalancing should loadbalance all requests equally by default


bartsmykla:
### What happened?

https://github.com/kumahq/kuma/blob/b98468fbd6dcc43f597b99ec2f00cbda09b27599/test/e2e_env/multizone/trafficroute/trafficroute.go#L384
Flake noticed in: https://github.com/kumahq/kuma/actions/runs/12082279982/job/33693379224
Test disabled in: https://github.com/kumahq/kuma/pull/12144

PR #12589, opened 2025-01-17
fix(kuma-cp): fix issue with pod labels overriding existing dataplane labels

Automaat:
Fixes: #12142

## Motivation

We should not override already present dataplane labels with pod labels

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/pod_converter.go
+++ b/pkg/plugins/runtime/k8s/controllers/pod_converter.go
@@ -63,7 +63,7 @@ func (p *PodConverter) PodToDataplane(
 	labels, err := model.ComputeLabels(
 		core_mesh.DataplaneResourceTypeDescriptor,
 		currentSpec,
-		pod.Labels,
+		mergeLabels(dataplane.GetLabels(), pod.Labels),
 		model.NewNamespace(pod.Namespace, pod.Namespace == p.SystemNamespace),
 		dataplane.Mesh,
 		p.Mode,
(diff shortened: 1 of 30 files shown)


pkg/plugins/runtime/k8s/controllers/pod_converter.go
```diff
@@ -411,6 +411,16 @@ func MetricsAggregateFor(pod *kube_core.Pod) ([]*mesh_proto.PrometheusAggregateM
 	return aggregateConfig, nil
 }
 
+func mergeLabels(existingLabels map[string]string, podLabels map[string]string) map[string]string {
```
lobkovilya, line 414: From the usage point of view I guess I'd expect input params to be unchanged as long as 
`mergeLabels` returns map. IMO we should either return `maps.Clone` or return nothing. Given the concurrency issues we sometimes have with maps I'd go with `maps.Clone`

pkg/plugins/runtime/k8s/controllers/pod_converter_test.go
```diff
@@ -135,13 +144,13 @@ var _ = Describe("PodToDataplane(..)", func() {
 			}
 
 			// when
-			dataplane := &mesh_k8s.Dataplane{}
-			err = converter.PodToDataplane(context.Background(), dataplane, pod, &namespace, services, otherDataplanes)
+			// dataplane := &mesh_k8s.Dataplane{}
```
lobkovilya, line 147: nit: unnecessary line

pkg/plugins/runtime/k8s/controllers/pod_converter.go
lukidzi: +1, we should avoid editing exisitng map, we might hit concurrency read/write

pkg/plugins/runtime/k8s/controllers/pod_converter_test.go
Automaat: removed

pkg/plugins/runtime/k8s/controllers/pod_converter.go
Automaat: updated

lobkovilya (dismissed):

lukidzi (approved):

Issue #12567, opened 2025-01-16
Support scoped serviceAccount


lahabana:
### Description

- Prior to 1.30.6, When OIDC issued service token with --service-account-extend-token-expiration set to true . With this, the service account tokens injected into pods were given an extended lifetime so they remain valid even after a new refreshed token is provided. Prior to upgrading to 1.30, the metric serviceaccount_stale_tokens_total and the audit annotation authentication.k8s.io/stale-token can be used to monitor for workloads that depend on the extended lifetime and are continuing to use tokens even after a refreshed token is provided to the container. If no action is taken, workloads depending on the extended lifetime will break once the cluster is upgraded to 1.30. See reference [README.md](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md) for details. 

After upgrading to 1.30.6 , It is no longer supported to change “service-account-extend-token-expiration” setting back to “True” for 1.30 OIDC enabled cluster thus this is the reason Kong Mesh CNI and egress pod were not able to renew the token and broke down with Unathorized.

We should support bounded serviceAccount as it seems to be an important security new feature in 1.30.

Docs:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-long-lived-api-token-for-a-serviceaccount
https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md

Is this only for CNI or more generally for other components?

jijiechen:
Correction:  the bound service account feature was introduced in Kubernetes 1.22. They allow the switch `service-account-extend-token-expiration` to play a "deprecation period" support duing v1.22 to v1.30 and this switch was eventually removed in 1.30.

Fortunately, from my study of Kuma code base, **only Kuma-CNI is being impacted by this issue.**

In other components, we are either not interacting with the API server directly or not copying the token to somewhere else before we use it. The `client-go` will read the token file from disk every minute ([link to source](https://github.com/kubernetes/client-go/blob/23900f492969db045e8a36396852381fd189569b/transport/token_source.go#L67-L69)) and the content of the token file is maintained by Kubernetes automatically. **So this is NOT an issue for components using the offical Go SDK without storing somewhere else or caching the token.**

[...]

PR #12590, opened 2025-01-18
refactor(cni): cni installer logic refactoring

bartsmykla:
## Motivation

This PR is a first one which prepares code for the following one which will handle cases where service account tokens are refreshed

## Supporting documentation

Related to: https://github.com/kumahq/kuma/issues/12567

> Changelog: skip


app/cni/pkg/install/validation.go
```diff
@@ -2,93 +2,58 @@ package install
...
 
-	parsed, err := parseFileToHashMap(cniConfPath)
-	if err != nil {
-		return err
+	plugins, ok := parsed["plugins"]
+	if !ok {
+		missingFields = append(missingFields, "plugins")
 	}
 
-	if isPluginChained {
```
jijiechen, line 61: Is this block deleted from the new version?

bartsmykla: No, the whole function was transformed to a method: https://github.com/kumahq/kuma/pull/12590/files/d433ae371ceb2baf56ab9ca2a187e4c09e4220fe#diff-4acb401431435486faf9ee1371e0580938c9dc6cc63720fc594b3cbb7e5dde38R191

app/cni/pkg/install/installer_config.go
```diff
@@ -48,136 +52,179 @@ func (i InstallerConfig) Validate() error {
...
-	matches, err = filepath.Glob(mountedCNINetDir + "/*.conflist")
-	if err != nil {
-		return "", err
-	}
-	file, found = lookForValidConfig(matches, isValidConflistFile)
-	if found {
-		return filepath.Base(file), nil
+		if file, ok := lookForValidConfig(matches, isValidConfFile); ok {
+			log.Info("found CNI config file", "file", file)
+			c.CniConfName = filepath.Base(file)
```
jijiechen, line 69: Changing/overriding a configuration value in the app logic seems not a  good idea. Can we use a return value?

jijiechen: Thanks for noting this. If you agree it's not a good idea to changing the configuration value, could you also improve this logic since you are doing a refactoring

bartsmykla: That's exactly where it should happen. We're using general logic to load env vars into shared configurations: https://github.com/kumahq/kuma/blob/c3afa4a4509998c242a64d2b575a90a175f913c0/pkg/config/config.go#L8  
I'd agree with you if we were explicitly creating an `InstallerConfig` instance, but that's not the case here.

jijiechen: Thanks for explaning and I'll mark as resolved since we are only making default values to this config item

jijiechen (approved):

jijiechen (approved):

PR #12603, opened 2025-01-20
refactor(rules): use `core_model.ResourceList` instead of `[]core_model.Resource`

lobkovilya:
## Motivation

Accepting `[]core_model.Resource` list is a bad function API as it implies resources in the list might have different types. Accepting `core_model.ResourceList` automatically implies all resources in the list have the same type.

## Implementation information

Depends on https://github.com/kumahq/kuma/pull/12559


michaelbeaumont (approved):
Any place where we could use generics to make this nicer?

lobkovilya:
@michaelbeaumont yes, at some point we should remove all autogenerated `Mesh*Resource` and `Mesh*ResourceList` types and replace them with `Resource[T]` and `ResourceList[T]`

michaelbeaumont (approved):

PR #12642, opened 2025-01-22
feat(api): respond with 200/201 with empty json for successful PUT

slonka:
## Motivation

In our OpenAPI spec we say that we're returning `application/json` for put 200/201 https://github.com/kumahq/kuma/blob/a00b2e0f87c69b59e62b580a9f3cad1a36fa5762/tools/openapi/templates/endpoints.yaml#L63 and [we don't always do that](https://github.com/kumahq/kuma/pull/12642/files#diff-a5eb0241e652f23066dcb33b652229602f43c52b565062af3c1cf5ac320f1249L427). I can't find a way to describe this with OpenAPI schema (either you don't set content or you set it, there is no `oneOf` on that level). I tried this:

```
      responses:
        '200':
          description: Updated
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: '#/components/schemas/MeshCreateOrUpdateSuccessResponse'
                  - type: 'null'
```

If we don't do this change terraform does not like the response not having content type:

```
╷
│ Error: failure to invoke API
│
│   with konnect_mesh.default,
│   on main.tf line 24, in resource "konnect_mesh" "default":
│   24: resource "konnect_mesh" "default" {
│
│ unknown content-type received: : Status 200
╵
```

## Implementation information

Respond with 200/201 and empty `{}` when there are no errors.

## Supporting documentation

xrel https://github.com/Kong/kong-mesh/issues/7201


pkg/api-server/resource_endpoints.go
```diff
@@ -424,7 +424,7 @@ func (r *resourceEndpoints) createResource(
 			log.Error(err, "Could not write the response")
```
lahabana, line 424: IMO This is odd, why don't we just return an empty json instead of the different status code? Both are breaking changes to the API but one seems least surprising (didn't even know 204 existed)

slonka: You can see that was my first approach, then I wanted to check how 204 would look like and how many places it would break to have some sort of a proxy "metric" for the clients. I agree, empty json will be better.

pkg/api-server/resource_endpoints.go
```diff
@@ -424,7 +424,9 @@ func (r *resourceEndpoints) createResource(
...
 		}
 	} else {
-		response.WriteHeader(201)
+		if err := response.WriteHeaderAndJson(201, struct{}{}, "application/json"); err != nil {
```
lobkovilya, line 427: there are constants `http.StatusOK` and `http.StatusCreated` we're using them in other places

lahabana:
@slonka title of the PR needs updating

pkg/api-server/resource_endpoints.go
```diff
@@ -476,7 +478,9 @@ func (r *resourceEndpoints) updateResource(
 			log.Error(err, "Could not write the response")
 		}
 	} else {
-		response.WriteHeader(200)
+		if err := response.WriteHeaderAndJson(200, struct{}{}, "application/json"); err != nil {
```
lahabana, line 481: Should we return api_server_types.CreateOrUpdateSuccessResponse instead?

```go
response := api_server_types.CreateOrUpdateSuccessRespons{Warnings: model.Deprecations(currentRes)}
if err := response.WriteHeaderAndJson(200, response, "application/json"); err != nil {
	log.Error(err, "Could not write the response")
}
```

That's how it should have been written originally and we wouldn't have this issue at least :)

pkg/api-server/resource_endpoints.go
slonka: done

Automaat (approved):

Issue #12361, opened 2024-12-23
Inspect API support for top-level targetRef `sectionName`


lobkovilya:
### Description

With the [new design](https://docs.google.com/document/d/1tdIOVVYObHbGKX1AFhbPH3ZQKYvpQG-2s6ShHSECaXM/edit?tab=t.0) it'll be possible select a single inbound port for configuration. Once https://github.com/kumahq/kuma/issues/12375 is implemented we should be able to update the Inspect API to reflect the fact only a single port is configured by the policy.

Today "fromRules" already has [an inbound object]( https://github.com/kumahq/kuma/blob/6d451bc60d0a493024238446ac3ecf971c977403/pkg/api-server/testdata/resources/inspect/dataplanes/_rules/overriding_meshtimeout.golden.json#L13C1-L17C8) we need to check if `sectionName` works out-of-the-box or requires additional support.

PR #12573, opened 2025-01-16
feat(kuma-cp): implement possibility to select proxies in policies by new kind Dataplane

Automaat:
## Motivation

We want to be able to target proxies by new kind Dataplane. We want to select Dataplanes by name/namespace, by labels and select single inbound by its name

## Supporting documentation

This is based on [Inbound policies MADR](https://docs.google.com/document/d/1tdIOVVYObHbGKX1AFhbPH3ZQKYvpQG-2s6ShHSECaXM/edit?pli=1&tab=t.0#heading=h.fxzs6450juto)

Fix #12359

PR #12644, opened 2025-01-22
feat(kuma-cp): add support for inspect api for new kind Dataplane and section name for selecting single inbound

Automaat:
Fix: #12361

## Motivation

We have introduced new kind and we need inspect api support for this, as it turns out it works out of the box so I've just added test for it.

This needs https://github.com/kumahq/kuma/pull/12573 to be merged first


pkg/api-server/testdata/resources/inspect/dataplanes/_rules/dataplane_kind_meshtimeout_section_name.golden.json
```diff
@@ -0,0 +1,83 @@
...
+         "name": "select-whole-dpp",
+         "type": "MeshTimeout"
+        }
+       ]
+      }
+     ]
+    },
+    {
+     "inbound": {
+      "port": 9090,
```
lobkovilya, line 44: we probably should add a `name` field here, it'll be helpful for GUI

Automaat: added

api/openapi/specs/common/resource.yaml
```diff
@@ -108,8 +108,10 @@ components:
            $ref: '#/components/schemas/Rule'
     Inbound:
       type: object
-      required: [tags, port]
+      required: [name, tags, port]
```
lobkovilya, line 111: why is the `name` required?

Automaat: I've made it required, just so it is not pointer to. string but I can change this if you think it should not be required

lobkovilya: Technically it doesn't matter as an empty value `name: ""` means `name` is not set. Could you please check with @johncowen or @schogges to understand what's the preferred way for such kind of situation?

schogges: I think it's better to omit `name` if it's not set. Theoretically it doesn't really matter because we'd catch that in our data-layer, but if `name` is an optional property, I'd treat it as such 🤔 from a TypeScript perspective I think it's better because we are forced to catch that case. An empty string is falsy, but we are not forced to handle that case. If we leave it as an empty string, we'd have to know that it could be an empty string and handle that.

For consistency reasons, are there similar cases and how do we handle that in these cases?

schogges: > This makes the properties of an API response stable and clients are then less prone to error when they expect a property to be there and it suddenly isn't.

If optional properties are excluded, no consumer can expect that it's there. Knowing that this can be an empty string and in that case it means that it's not set, is not obvious from my point of view.

johncowen: How do you know people are using the specs?

schogges: > How do you know people are using the specs?

Can't we just expect that? Is there anything else I don't know about yet other than inspecting API responses?

johncowen: > Can't we just expect that?

I guess someone might be doing a bash thing with curl, so IMO I'd say we can't fully expect that.

---

I think the other thing, even if people are using the specs, they still have to add some sort of guard to check that `name` is there before doing something with it. This could be multiplied across a codebase (unless you have a centralised place like we do), and then you could multiply that by the amount of people using the HTTP API. That's a lot of unnecessary code

Alternatively, we could put whatever logic is needed to set `name: ""` in the backend (i.e. at source) instead then folks using the HTTP API don't have to check for existence before using `name`

schogges: > > Can't we just expect that?
> 
> I guess someone might be doing a bash thing with curl, so IMO I'd say we can't fully expect that.

Regardless of how the api might be consumed, there are plenty ways to check what is in the specs. Even when ignoring the specs and inspecting an API response and the `name` is not always returned one could infer that it's an optional property.

> I think the other thing, even if people are using the specs, they still have to add some sort of guard to check that `name` is there before doing something with it. This could be multiplied across a codebase (unless you have a centralised place like we do), and then you could multiply that by the amount of people using the HTTP API. That's a lot of unnecessary code

I doubt that this is our problem 😅 also when I build something on top of the api and want to exclude the name when it's not set, it doesn't matter if it's returned or an empty string. In any case I have to check for that.

> Alternatively, we could put whatever logic is needed to set `name: ""` in the backend (i.e. at source) instead then folks using the HTTP API don't have to check for existence before using `name`

But they would have to check why it's an empty string and what it means. They could even think it's a bug. This is just not obvious from my point of view and I think this could cause some confusion.

johncowen: Well it looks like we have pretty strong opposite opinions, so kinda feel like we aren't helping here 😅 

@Automaat @lobkovilya any thoughts?

johncowen: ❓ Are Inbound `name`s unique?

Automaat: I'm leaning towards John opinion, to avoid nullable values in api. I think they should be unique, on Kubernetes they will be unique, on universal I think we should add validation for this. What do you think about it @lobkovilya?

johncowen: I was just chatting to @lahabana and he mentioned that kubernetes omits empty values completely, and seeing as we are a very kubernetes piece of software we should probably follow that convention, here's al ink to that https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md. I'm gonna have a quick read of things myself, but I've probably changed my stance on this as I think following this convention is more important. Will be back once I've read up a little more.

lahabana: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

johncowen: Another follow up question here, do we have some sort of doc somewhere that explains the style/feel of our HTTP APIs. So if I'm a beginner contributing to the software I have something to follow to help me make all the APIs feel the same? (Maybe this is just the same link to those kubernetes API conventions)

johncowen: Just to note on this one, we had a discussion around supporting a terraform provider via our HTTP API, and also around what our most common HTTP API client, Envoy, expects. Similar questions and sharp edges arose.

Would be really good to get these sharp edges outlined and a common consistent approach to those documented somewhere so we know (both from a backend and frontend perspective) what to expect. Something I think we are going to follow up on.

johncowen: So I had a good read over the Kubernetes sig conventions and came to the conclusion that it is very much inline with my initial thinking in https://github.com/kumahq/kuma/pull/12644#discussion_r1939161419.

TLDR; explicitness, no ambiguity, good defaults

Also wanted to say to @Automaat and @lobkovilya sorry for taking this thread completely our into space, around the planets and back again 😬 😆 🚀 . Honestly, I'm pretty convinced of my opinion at this point, but that doesn't mean people have to agree with mine, I'll leave it up to you you folks. I'll probably dip out of this conversation after this, unless people have want clarifications or have questions specifically for me.

Lastly, thanks for asking for my opinion, I found out a lot, so thanks 🙏 

Anyway here goes...:

---

The section on defaulting states:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#defaulting

> In general we want default values to be explicitly represented in our APIs, rather than asserting that "unspecified fields get the default behavior". This is important so that:
> 
> - default values can evolve and change in newer API versions
> - the stored configuration depicts the full desired state, making it easier for the system to determine how to achieve the state, and for the user to know what to anticipate

Apart from that my feeling was that the entire document frequently talks about _explicitness_, i.e. avoid ambiguity don't let API consumers guess.

The Optional vs Required section (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) was mentioned. 'Required' fields are obviously filled out, but I'd go as far as to say optional fields should have a default in order to be explicit (see previous `Defaults` link in the same document).

Going back to my original comment. Booleans can only be one of two states, if not its not a boolean by its very definition, a boolean's default should be either `true` or `false`, not non-existent. This explicitness ensures that the end user of an API knows whether a default state is `true` or `false`, we had (I think its been changed now) a problematic lack of explicitness with this via a `health` type property a while back.

Moving on from booleans a little more, if we are using an enum for states or similar, the Kubernetes API conventions mention an explicit `Unknown` state, i.e. the state is `Unknown` not undefined/null.

For numbers, there is almost always a good default, usually `0`, but in some cases as I found out, in our software we have defaults of `2` in some places. For me these are important to expose in the API so we are explicit about what their values are so the consumer doesn't have to guess.

We have a `maxInterval` property somewhere, which multiplies the base interval by its value from what I understand. In my eyes, this should have a default of `1` not `undefined`/`omitempty`.

If an array is "unset" then its an empty array, if an object is unset then its an empty object. Its being explicit about what this property is, leaving no room for ambiguity and errors resulting from that.

The only thing we haven't really mentioned are strings.

I found a couple of mentions relevant to strings and defaults specifically in the kubernete's convention doc:

```
// reason contains a programmatic identifier indicating the reason for the condition's last transition.
// Producers of specific condition types may define expected values and meanings for this field,
// and whether the values are considered a guaranteed API.
// The value should be a CamelCase string.
// This field may not be empty.
// +required
Reason string `json:"reason" protobuf:"bytes,5,opt,name=reason"`
// message is a human readable message indicating details about the transition.
// This may be an empty string.
// +required
Message string `json:"message" protobuf:"bytes,6,opt,name=message"`
```

Both `Reason` and `Message` are strings, both are required but both can be an empty string i.e. `Message: ""`. As they are required I am guessing those don't use `omitempty`.

If we wanted to look specifically at string property that are distinctly `Name:` (quite an important field in kubernetes) I found that `Name` is mentioned in the docs as an example:

```
ports:
  - name: www
    containerPort: 80
```
But the docs doesn't give an guidance or advice on whether to use omitempty or not on a string field specifically called `Name:`

I have comparatively little knowledge of kubernetes, but I did find https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services

> For some Services, you need to expose more than one port. Kubernetes lets you configure multiple port definitions on a Service object. When using multiple ports for a Service, you must give all of your ports names so that these are unambiguous.

I don't know how relevant or important that is to what we are doing here.

At the very least if we are trying to provide consistent APIs to our consumers, wouldn't it be easier at least in this InspectAPI case to just say **"we do not allow omitempty"**?

---

I also found the following related and interesting:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

> The specification is a complete description of the desired state, including configuration settings provided by the user, [default values](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#defaulting) expanded by the system, and properties initialized or otherwise changed after creation by other ecosystem components (e.g., schedulers, auto-scalers), and is persisted in stable storage with the API object

> The status summarizes the current state of the object in the system, and is usually persisted with the object by automated processes but may be generated on the fly. As a general guideline, fields in status should be the most recent observations of actual state, but they may contain information such as the results of allocations or similar operations which are executed in response to the object's spec

---

Just to note there is an ongoing conversation on certain API endpoints that due to internal usage require fields to be omitempty/undefined. I understand why these need to be as they are, and I agree that these are a separate case/group of API endpoints and not necessary covered by what we are discussing here. Also note this PR seems to be seems to be part of the "Inspect API" i.e. whose purpose I believe is to show the expanded system result of the user configuration.

johncowen: Sorry I know I said I would dip out now, but! I noticed this:

https://github.com/kumahq/kuma-website/issues/2163

```
to:
  - targetRef:
      kind: MeshService
      name: backend
      sectionName: port-name # or stringify port's value like "8080"
```

I don't think its such a bad suggestion that a port name could default to `the stringified port's value like "8080"`. One benefit is, as this comes out of the user facing HTTP API, there is no need to document what to do if the port doesn't have a name, because it always has one. If you have explicitly set one yourself, its that, if you haven't its the port number expressed as a string.

As usual I don't understand the details of any under-the-hood technical implications this might have, but from a user-perspective this sounds like another option for a reasonable default to me.

lobkovilya (approved):

Issue #12649, opened 2025-01-23
MeshTimeout inconsistent default values for `spec.to` depending on the presence of the inbound policy


lobkovilya:
## Problem
When there are no MeshTimeout then plugin is not executed https://github.com/kumahq/kuma/blob/820494cf59e22d5a4870ca6d7a01f1f77e38ad46/pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go#L40-L46

But when there are MeshTimeouts that configure `spec.from` then `spec.to` is configured with default values. 

## Steps to reproduce

1. Apply shadow MeshTimeout that configures `spec.from`
```yaml
type: MeshTimeout
name: mt-1
mesh: meshtimeout-envoyconfig
labels:
  kuma.io/effect: shadow
spec:
  from:
    - targetRef:
        kind: Mesh
      default:
        idleTimeout: 50s
        connectionTimeout: 51s
        http:
          requestTimeout: 52s
          streamIdleTimeout: 53s
          maxStreamDuration: 54s
          maxConnectionDuration: 55s
```
2. Run 
```
kumactl inspect dataplane $name --type=config --shadow --include=diff | jq '.diff'
```
```json
[
  {
    "op": "test",
    "path": "/type.googleapis.com~1envoy.config.cluster.v3.Cluster/localhost:3000/connectTimeout",
    "value": "10s"
  },
  {
    "op": "remove",
    "path": "/type.googleapis.com~1envoy.config.cluster.v3.Cluster/localhost:3000/connectTimeout",
    "value": "10s"
  },
  {
    "op": "add",
    "path": "/type.googleapis.com~1envoy.config.cluster.v3.Cluster/localhost:3000/connectTimeout",
    "value": "51s"
  },
  {
    "op": "add",
    "path": "/type.googleapis.com~1envoy.config.cluster.v3.Cluster/meshtimeout-envoyconfig_demo-client__kuma-3_msvc_3000/connectTimeout",
    "value": "5s"
  },
  {
    "op": "add",
    "path": "/type.googleapis.com~1envoy.config.cluster.v3.Cluster/meshtimeout-envoyconfig_test-server__kuma-3_msvc_80/typedExtensionProtocolOptions/envoy.extensions.upstreams.http.v3.HttpProtocolOptions/commonHttpProtocolOptions",
    "value": {
      "idleTimeout": "3600s",
      "maxConnectionDuration": "5s",
      "maxStreamDuration": "0s"
    }
  },
  {
[...]

Issue #12648, opened 2025-01-23
MeshTimeout incorrect default value for `http.streamIdleTimeout`


lobkovilya:
### What happened?

In docs it says 30m https://kuma.io/docs/2.9.x/policies/meshtimeout/#defaults. In practice it's 5s because of this line https://github.com/kumahq/kuma/blob/820494cf59e22d5a4870ca6d7a01f1f77e38ad46/pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go#L231

should be:
```go
StreamIdleTimeout: &kube_meta.Duration{Duration: policies_defaults.DefaultStreamIdleTimeout},
```

lobkovilya:
same for `http.maxConnectionTimeout` 😞 https://github.com/kumahq/kuma/blob/820494cf59e22d5a4870ca6d7a01f1f77e38ad46/pkg/plugins/policies/meshtimeout/plugin/xds/configurer.go#L233

lobkovilya:
closing in favor of https://github.com/kumahq/kuma/issues/12667

PR #12650, opened 2025-01-23
test(meshtimeout): e2e test that checks golden envoy configs

lobkovilya:
## Motivation

When we test `BuildRules` we pass MeshTimeout policies as input and check `outbound.ResourceRules` Go struct.

When we test MeshTimeout plugin we pass `outbound.ResourceRules` Go struct as input and check envoy configs.

We need an easy way to pass MeshTimeout policies and check envoy configs. It's sort of an integration test that involves 2 packages. The easiest way to do that is to add another universal e2e test with golden files. This way we have running Kuma CP and we basically test it as a black box.

I managed to find 2 bugs using these tests:
* https://github.com/kumahq/kuma/issues/12648
* https://github.com/kumahq/kuma/issues/12649

## Implementation information

We check DPP `config` (marshalled Snapshot in Kuma CP) instead of the actual Envoy config dump for a few reasons:
* no need to add logic that waits while config is propagated on DPP
* less setup-specific values we have to reduct to have stable golden file (only IP addresses)
* we can use `kuma.io/effect: shadow` and quickly review the `diff` policies introduces


test/e2e_env/universal/timeout/envoyconfig.go
```diff
@@ -0,0 +1,101 @@
...
+				Name string `json:"name"`
+			} `json:"items"`
+		}
+		Expect(json.Unmarshal([]byte(out), &output)).To(Succeed())
+		for _, item := range output.Items {
+			Expect(universal.Cluster.GetKumactlOptions().RunKumactl("delete", "meshtimeout", item.Name, "--mesh", meshName)).To(Succeed())
+		}
+	})
+
+	redactIPs := func(jsonStr string) string {
```
Automaat, line 73: can we extract this to some common fuctions, as we would like to use this in other similar tests for policies?

test/e2e_env/universal/timeout/envoyconfig.go
```diff
@@ -0,0 +1,101 @@
...
+			input, err := os.ReadFile(inputFile)
+			Expect(err).ToNot(HaveOccurred())
+
+			// when
+			if len(input) > 0 {
+				Expect(universal.Cluster.Install(YamlUniversal(string(input)))).To(Succeed())
+			}
+
+			// then
+			Expect(getConfig("demo-client")).To(matchers.MatchGoldenJSON(strings.Replace(inputFile, "input.yaml", "demo-client.golden.json", 1)))
```
Automaat, line 97: this test is pretty general, can we move it to some other directory than timeout and just add input and golden files for other policies?

test/e2e_env/universal/timeout/envoyconfig.go
Automaat: If we make this test more general, then probably this comment does not make sense

test/e2e_env/universal/timeout/envoyconfig.go
lobkovilya: Moved the test to separate dir `envoyconfig`. The only problem I see with the current approach is that we won't be able to run these tests for different policies concurrently, but I guess we're going to solve this when it becomes a problem

Automaat (approved):

PR #12659, opened 2025-01-24
feat(kuma-cp): allow using Dataplane kind in top level targetRef in all policies

Automaat:
## Motivation

As in pr title, we want to use top level targetRef kind Dataplane in all policies


pkg/plugins/policies/meshtls/api/v1alpha1/validator.go
```diff
@@ -27,6 +27,7 @@ func validateTop(targetRef *common_api.TargetRef) validators.ValidationError {
 		SupportedKinds: []common_api.TargetRefKind{
 			common_api.Mesh,
 			common_api.MeshSubset,
+			common_api.Dataplane,
```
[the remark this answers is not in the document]
Automaat: right, I forgot that this is inbound policy

pkg/plugins/policies/meshaccesslog/api/v1alpha1/validator.go
```diff
@@ -30,8 +30,10 @@ func validateTop(targetRef common_api.TargetRef) validators.ValidationError {
 			common_api.MeshGateway,
 			common_api.MeshService,
 			common_api.MeshServiceSubset,
+			common_api.Dataplane,
 		},
 		GatewayListenerTagsAllowed: true,
+		IsInboundPolicy:            true,
```
lobkovilya, line 36: we have to take this info from the ResourceDescriptor, otherwise we might forget to set it for new policies

Automaat: changed

pkg/plugins/policies/meshhealthcheck/api/v1alpha1/validator.go
```diff
@@ -3,19 +3,20 @@ package v1alpha1
...
 	var verr validators.ValidationError
 	path := validators.RootedAt("spec")
-	verr.AddErrorAt(path.Field("targetRef"), validateTop(r.Spec.TargetRef))
+	verr.AddErrorAt(path.Field("targetRef"), validateTop(r.Spec.TargetRef, r.Descriptor()))
 	verr.AddErrorAt(path, validateTo(pointer.DerefOr(r.Spec.TargetRef, common_api.TargetRef{Kind: common_api.Mesh}), r.Spec.To))
 	return verr.OrNil()
 }
 
-func validateTop(targetRef *common_api.TargetRef) validators.ValidationError {
+func validateTop(targetRef *common_api.TargetRef, descriptor core_model.ResourceTypeDescriptor) validators.ValidationError {
```
lobkovilya, line 19: please make `validateTop` to be a method of `MeshHealthCheckResource`

pkg/plugins/policies/meshfaultinjection/api/v1alpha1/validator.go
```diff
@@ -12,14 +12,14 @@ import (
...
-	verr.AddErrorAt(path.Field("targetRef"), r.validateTop(r.Spec.TargetRef))
+	verr.AddErrorAt(path.Field("targetRef"), r.validateTop(r.Spec.TargetRef, r.Descriptor()))
 	topLevel := pointer.DerefOr(r.Spec.TargetRef, common_api.TargetRef{Kind: common_api.Mesh})
 	verr.AddErrorAt(path, validateFrom(topLevel, r.Spec.From))
 	verr.AddErrorAt(path, validateTo(topLevel, r.Spec.To))
 	return verr.OrNil()
 }
 
-func (r *MeshFaultInjectionResource) validateTop(targetRef *common_api.TargetRef) validators.ValidationError {
+func (r *MeshFaultInjectionResource) validateTop(targetRef *common_api.TargetRef, descriptor core_model.ResourceTypeDescriptor) validators.ValidationError {
```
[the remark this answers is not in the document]
Automaat: changed

lobkovilya:
wait, but the stuff in ResourceDescriptor is static and depends on the policy type, right? we should allow/disallow `sectionName` based on the actual presence of `from` or `rules`. For example, if MeshTimeout has only `to` and `ResourceDescriptor.HasFromTargetRef` is `true` we still shouldn't allow `sectionName`

Automaat:
@lobkovilya I've circled back to isInboundPolicy option

lobkovilya:
probably you need to add some unit tests in validation, as there were no failures with incorrect approach

Automaat:
@lobkovilya I've added this test previosly: https://github.com/kumahq/kuma/blob/1bbc414fa983d0e97f72d1cd429cde41588ab891/pkg/core/resources/apis/mesh/validators_test.go#L1030-L1040

lobkovilya:
No I mean initially you've passed `IsInboundPolicy: true` to the validator without checking `len(r.Spec.From) > 0)`. We agreed that it was incorrect, and I think that some tests should've failed. But since no tests failed I assume we're missing some tests, right?

lobkovilya (dismissed):

pkg/plugins/policies/meshtimeout/api/v1alpha1/validator.go
```diff
@@ -12,7 +12,7 @@ import (
 func (r *MeshTimeoutResource) validate() error {
 	var verr validators.ValidationError
 	path := validators.RootedAt("spec")
-	verr.AddErrorAt(path.Field("targetRef"), r.validateTop(r.Spec.TargetRef))
+	verr.AddErrorAt(path.Field("targetRef"), r.validateTop(r.Spec.TargetRef, len(r.Spec.From) > 0 || len(r.Spec.Rules) > 0))
```
lobkovilya, line 15: maybe you should create a func
```go
func AffectsInbounds(p core_model.Policy) bool
```
that tries to cast to `PolicyWithRules` and `PolicyWithFromList` and checks `len`.

Otherwise we have to update these validations every time we add `spec.rules` support

Automaat: changed

lobkovilya (approved):

Issue #12362, opened 2024-12-23
Deprecate top-level targetRef `MeshSubset`


lobkovilya:
### Description

Once https://github.com/kumahq/kuma/issues/12359 is done, we should be able to deprecate top-level targetRef `MeshSubset`.

PR #12660, opened 2025-01-24
feat(kuma-cp): deprecate MeshSubset kind in top level targetRef

Automaat:
Fix: #12362

## Motivation

Since we've introduced new Dataplane kind we can deprecate usage of MeshSubset for selecting proxies. Dataplane is more powerful and better matches our approach for selecting real resources.


lobkovilya (approved):

lobkovilya:
probably it's a good idea to add a webhook unit test

lobkovilya (commented):
probably it's a good idea to add a webhook unit test

Automaat:
@lobkovilya added test

lobkovilya (approved):

Issue #12508, opened 2025-01-10
Use different registry for release and non release images


lahabana:
### Description

Originally we were going to push preview and release to different repositories.
Now we'd like to push to always the same registry but we'll have an internal repo that contains code to take from the CI repo and push to docker hub.

This will make docker hub cleaner and we can run extra validation before releasing.

lahabana:
There's a discussion about using different repository for either of these and have a staging area for prod artifacts. Waiting for things to be clearer before prioritizing this

slonka:
@lahabana did things clear up?

lahabana:
Nope let's close

PR #12661, opened 2025-01-24
chore(build): publish no release images with -dev suffix

lukidzi:
## Motivation

We want to separate release images from the ones that we build.

## Implementation

When the version has `preview` scripts add a suffix `dev` to the repository 

example:
when tag: kuma-dp, kuma-init, kuma-cp, kumactl, kuma-dni
other: kuma-dp-dev, kuma-init-dev, kuma-cp-dev, kumactl-dev, kuma-cni-dev

fix: https://github.com/kumahq/kuma/issues/12508


mk/docker.mk
```diff
@@ -3,11 +3,18 @@ KUMA_VERSION ?= master
...
 
+ifdef IMAGES_REPOSITORY_SUFFIX
+ifneq ($(strip $(IMAGES_REPOSITORY_SUFFIX)),)
+DOCKER_DEV_IMAGE_SUFFIX = -$(IMAGES_REPOSITORY_SUFFIX)
+endif
+endif
+
 define build_image
-$(addsuffix :$(BUILD_INFO_VERSION)$(if $(2),-$(2)),$(addprefix $(DOCKER_REGISTRY)/,$(1)))
+$(addsuffix :$(BUILD_INFO_VERSION)$(if $(2),-$(2)),$(addprefix $(DOCKER_REGISTRY)/,$(1)$(DOCKER_DEV_IMAGE_SUFFIX)))
```
lahabana, line 17: Can't be actually just do:

```
$(DOCKER_REGISTRY)/$(if $(2),$(2)-)$(1)$(if $(DOCKER_DEV_IMAGE_SUFFIX),-$(DOCKER_DEV_IMAGES_SUFFIX)):$(BUILD_INFO_VERSION)
```

Seems like it's equivalent and more readable to me no?

mk/docker.mk
```diff
@@ -3,11 +3,18 @@ KUMA_VERSION ?= master
...
 DOCKER_REGISTRY ?= $(DOCKER_SERVER)/kumahq
+DOCKER_DEV_IMAGE_SUFFIX ?=
 DOCKER_USERNAME ?=
 DOCKER_API_KEY ?=
 
+ifdef IMAGES_REPOSITORY_SUFFIX
+ifneq ($(strip $(IMAGES_REPOSITORY_SUFFIX)),)
+DOCKER_DEV_IMAGE_SUFFIX = -$(IMAGES_REPOSITORY_SUFFIX)
+endif
+endif
```
lahabana, line 14: Why the indirection? Feels like we could use `IMAGE_REPOSITORY_SUFFIX` directly in the script no?

mk/docker.mk
lahabana: Also any specific rational for prefix vs suffix?

lahabana (changes requested):
Actually there's been further thinking around this and we're looking at using a different registry for non released images.

I think that for Kuma it'll likely be GHCR. This is TBD but I'd close this until we make a decision

mk/docker.mk
lukidzi: I feel suffix is more used in different repos:
https://hub.docker.com/r/kong/kong-gateway-dev
https://hub.docker.com/r/kong/kong-gateway-dev
https://hub.docker.com/r/envoyproxy/envoy-dev

lukidzi:
> Actually there's been further thinking around this and we're looking at using a different registry for non released images.
> 
> I think that for Kuma it'll likely be GHCR. This is TBD but I'd close this until we make a decision

Ok, I can close it for now, and once we agree we can reopen with the provided review suggestions

PR #12664, opened 2025-01-24
fix(xds): ignore watchdog error on context cancelled

### LINKED ISSUE #12663: Error log due to race condition on proxy disconnection
### Kuma Version

2.9.1

### Describe the bug

Sometimes I see `metadata cannot be nil` when proxy disconnects. It's because watchdog is scheduled when proxy disconnects.

As an example, it occurred 4 times over last 3 days with 100s of proxies.

### To Reproduce

Might be a bit hard to reproduce

### Expected behavior

No such log

We need better checks in Dataplane Watchdog that context was already cancelled

jakubdyszkiewicz:
## Motivation

We are producing logs that we should ignore

## Implementation information

I made a change to ignore OnError callback in case of ctx is cancelled.
Other option would be to change log level from ERROR to INFO dynamically in OnError callback implementation, but we would need to repeat this across all watchdog implems. I'm not sure we care about this error. The intention of already existing `!errors.Is(err, context.Canceled)` seems to be to ignore such errors.

While this is fix, I don't think it's worth to backport it. It's not a stability issue, it's just a slight noise in logs.

## Supporting documentation

Fix #12663

### DIFF
--- a/pkg/util/watchdog/watchdog.go
+++ b/pkg/util/watchdog/watchdog.go
@@ -44,7 +44,7 @@ func (w *SimpleWatchdog) Start(ctx context.Context) {
 	defer ticker.Stop()
 	for {
 		if err := w.onTick(ctx); err != nil {
-			if !errors.Is(err, context.Canceled) && w.OnError != nil {
+			if !channels.IsClosed(ctx.Done()) && !errors.Is(err, context.Canceled) && w.OnError != nil {
 				w.OnError(err)
 			}
 		}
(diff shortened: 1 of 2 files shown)


michaelbeaumont:
Why do we need the channel check exactly?

jakubdyszkiewicz:
> Why do we need the channel check exactly?

The idea is to not produce errors caused by the client canceling the connection to the control plane. In such a scenario we have context cancelled which we check already, but that's not enough. There might be something else (like the error described in the linked issue). Instead of relying that ontick will fail with context cancelled error, we can check whether the context is cancelled when ontick failed.

michaelbeaumont:
> we can check whether the context is cancelled

OK I think I got confused because we're checking this already, I guess you mean you're adding a check whether the context is done.

slonka (approved):

michaelbeaumont (approved):

PR #12682, opened 2025-01-28
test(meshtimeout): fix golden file tests to handle IPv4 and IPv6

bartsmykla:

Updated tests to replace both IPv4 and IPv6 addresses with "IP_REDACTED" in generated golden files. This fixes issues with e2e tests checking golden envoy configs (introduced in https://github.com/kumahq/kuma/pull/12650)


lobkovilya (approved):
Awesome, thanks for fixing this!

Automaat (approved):

PR #12688, opened 2025-01-28
test(meshtimeout): stable sort the `.diff` list after redacting IPs

lobkovilya:
## Motivation

* The order of operations in `response.diff` depends on the listeners' and clusters' names that contain IP addresses. If we want this order to be stable, we have to sort it once again after redacting the IPs.
* compile regex once globally


bartsmykla (dismissed):
lgtm

bartsmykla (approved):

Issue #12657, opened 2025-01-23
Flake `SimpleWatchdog [It] should wait for the first tick to happen on WaitForFirstTick`


lobkovilya:
### Description

Unit test flake https://github.com/kumahq/kuma/actions/runs/12931389429/job/36065828247?pr=12650

```
• [FAILED] [0.001 seconds]
SimpleWatchdog [It] should wait for the first tick to happen on WaitForFirstTick
github.com/kumahq/kuma/pkg/util/watchdog/watchdog_test.go:191

  [FAILED] Expected
      <bool>: true
  to be false
  In [It] at: github.com/kumahq/kuma/pkg/util/watchdog/watchdog_test.go:217 @ 01/23/25 14:53:39.112
------------------------------

Summarizing 1 Failure:
  [FAIL] SimpleWatchdog [It] should wait for the first tick to happen on WaitForFirstTick
  github.com/kumahq/kuma/pkg/util/watchdog/watchdog_test.go:217
```

PR #12693, opened 2025-01-28
fix(watchdog): wait for the first tick

lukidzi:
## Motivation

Issue: https://github.com/kumahq/kuma/issues/12657

## Implementation information

Changed to wait for the `HasTicked` to be called before first Ticker event, since there was a race and sometimes on ticked happened before the assertion 

fix: https://github.com/kumahq/kuma/issues/12657

### DIFF
--- a/pkg/util/watchdog/watchdog.go
+++ b/pkg/util/watchdog/watchdog.go
@@ -48,12 +48,12 @@ func (w *SimpleWatchdog) Start(ctx context.Context) {
 				w.OnError(err)
 			}
 		}
-		if w.hasTickedChan != nil && !channels.IsClosed(w.hasTickedChan) { // On the first tick we close the channel
-			close(w.hasTickedChan)
-		}
 		select {
 		case <-ctx.Done():
 		case <-ticker.C:
+		}
+		if w.hasTickedChan != nil && !channels.IsClosed(w.hasTickedChan) { // On the first tick we close the channel
+			close(w.hasTickedChan)
 		}
 		// cases are non prioritized so we first check is the context is done
 		select {
--- a/pkg/util/watchdog/watchdog.go
+++ b/pkg/util/watchdog/watchdog.go
@@ -48,12 +48,12 @@ func (w *SimpleWatchdog) Start(ctx context.Context) {
 				w.OnError(err)
 			}
 		}
+		if w.hasTickedChan != nil && !channels.IsClosed(w.hasTickedChan) { // On the first tick we close the channel
+			close(w.hasTickedChan)
+		}
 		select {
 		case <-ctx.Done():
 		case <-ticker.C:
-		}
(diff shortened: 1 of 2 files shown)


pkg/util/watchdog/watchdog_test.go
```diff
@@ -216,12 +216,16 @@ var _ = Describe("SimpleWatchdog", func() {
 			close(doneCh)
 		}()
 		Expect(watchdog.HasTicked(false)).Should(BeFalse())
-		Consistently(hasTicked).ShouldNot(Receive())
+		Consistently(func(g Gomega) {
+			g.Expect(hasTicked).ShouldNot(Receive())
```
lobkovilya, line 220: isn't it the same as
```go
Consistently(hasTicked, "100ms", "10ms").ShouldNot(Receive())
```

lukidzi: Yes, changed it since it should works well for channel but in case of function( eventually few lines below it won't work)

pkg/util/watchdog/watchdog.go
```diff
@@ -48,13 +48,13 @@ func (w *SimpleWatchdog) Start(ctx context.Context) {
...
 			}
 		}
-		if w.hasTickedChan != nil && !channels.IsClosed(w.hasTickedChan) { // On the first tick we close the channel
-			close(w.hasTickedChan)
-		}
 		select {
 		case <-ctx.Done():
 		case <-ticker.C:
 		}
+		if w.hasTickedChan != nil && !channels.IsClosed(w.hasTickedChan) { // On the first tick we close the channel
```
lobkovilya, line 55: I'm pretty sure the idea behind `HasTicked` is to check that `w.onTick` was called once, not that the timer is actually ticked once, see the comment https://github.com/kumahq/kuma/pull/11590/files#r1818950478

pkg/util/watchdog/watchdog_test.go
```diff
@@ -215,8 +215,11 @@ var _ = Describe("SimpleWatchdog", func() {
 			watchdog.Start(ctx)
 			close(doneCh)
 		}()
-		Expect(watchdog.HasTicked(false)).Should(BeFalse())
-		Consistently(hasTicked).ShouldNot(Receive())
+		// first onTick call
```
lobkovilya, line 218: maybe you can pass function to OnTick that blocks on something, i.e
```go
OnTick: func(ctx context.Context) error {
    <-onTickBlockCh
    return nil
},
```
and then you can check
```go
// before the first OnTick returns false
Expect(watchdog.HasTicked(false)).Should(BeFalse())
// unblock OnTick
close(onTickBlockCh)
// after the first tick returns true
Expect(watchdog.HasTicked(false)).Should(BeTrue())
Expect(hasTicked).Should(BeClosed())
```

PR #12696, opened 2025-01-28
feat(meshtimeout): add support for 'spec.rules'

lobkovilya:
## Implementation information

Depends on https://github.com/kumahq/kuma/pull/12692 so it's obvious this PR doesn't change golden files in `test/e2e_env/universal/envoyconfig`

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12365


Automaat (approved):

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin_test.go
```diff
@@ -245,6 +246,23 @@ var _ = Describe("MeshTimeout", func() {
 						},
 					},
 				},
+				InboundRules: map[core_rules.InboundListener][]*inbound.Rule{
```
Automaat, line 249: shouldn't this replace old from rules in this test? Why we keep them?

PR #12710, opened 2025-01-30
fix(rules): a bug in ResolveTargetRef that caused creating excessive entries in ResourceRules

lobkovilya:
## Motivation

It's a minor bug, I noticed it while covering `ResolveTargetRef` function with unit tests. It resulted in creating extra entities in `ResourceRules` map, i.e if policy has:
```yaml
to:
  - targetRef:
      kind: MeshService
      name: backend
      namespace: kuma-demo
      sectionName: non-existent-section-name
    default: $conf1
```

we're going to produce entry

```json
{
  "..."
  "meshservice:mesh/mesh-1:namespace/kuma-demo:name/backend:section/non-existent-section-name": $conf1,
  "..."
}
```

but it won't be used by the actual policy plugins because `MeshService` doesn't have `non-existent-section-name`.

## Implementation information

* properly handle section names in `ResolveTargetRef` function
* add unit tests
* common interface for ports to avoid depending on specific types like MeshService or MeshMutliZoneService

### DIFF
--- a/pkg/plugins/policies/core/rules/common/resolvetargetref.go
+++ b/pkg/plugins/policies/core/rules/common/resolvetargetref.go
@@ -8,6 +8,7 @@ import (
 
 	common_api "github.com/kumahq/kuma/api/common/v1alpha1"
 	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
+	"github.com/kumahq/kuma/pkg/core/resources/apis/core"
 	meshservice_api "github.com/kumahq/kuma/pkg/core/resources/apis/meshservice/api/v1alpha1"
 	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
 	"github.com/kumahq/kuma/pkg/plugins/policies/core/rules/subsetutils"
@@ -28,6 +29,10 @@ func UniqueKey(r core_model.Resource, sectionName string) core_model.TypedResour
 		ResourceType:       r.Descriptor().Name,
 		SectionName:        sectionName,
 	}
+}
+
+type ResourceWithPorts interface {
+	GetPorts() []core.Port
 }
 
 func ResolveTargetRef(targetRef common_api.TargetRef, tMeta core_model.ResourceMeta, reader ResourceReader) []*ResourceSection {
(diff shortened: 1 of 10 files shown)


pkg/plugins/policies/core/rules/common/resolvetargetref.go
```diff
@@ -80,14 +85,40 @@ func ResolveTargetRef(targetRef common_api.TargetRef, tMeta core_model.ResourceM
...
-	return nil
+	if resourceWithPorts, ok := resolvedResource.(ResourceWithPorts); ok && targetRef.SectionName != "" {
+		for _, port := range resourceWithPorts.GetPorts() {
+			if port.GetName() == targetRef.SectionName {
+				return []*ResourceSection{{
+					Resource:    resolvedResource,
+					SectionName: port.GetName(),
+				}}
+			}
+			if targetRefPort, ok := tryParsePort(targetRef.SectionName); ok && port.GetName() == "" && port.GetValue() == targetRefPort {
```
lukidzi, line 101: I went through the code and we set port value as a port name if there is no name. Won't it affect matching here?

lobkovilya: > we set port value as a port name if there is no name

hmm, could you please share code where it happens

lukidzi: One for universal other for k8s
https://github.com/kumahq/kuma/blob/master/pkg/core/resources/apis/meshservice/generate/generator.go#L90
https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/controllers/meshservice_controller.go#L397

lukidzi: I don't think so, since MeshMultizoneService is created manually and we only update status object. It can be defined by the user: https://kuma.io/docs/2.9.x/networking/meshmultizoneservice/#ports

lobkovilya: right, so the second condition
```go
if targetRefPort, ok := tryParsePort(targetRef.SectionName); ok && port.GetName() == "" && port.GetValue() == targetRefPort {
```
makes sense

lukidzi (approved):

Issue #12381, opened 2024-12-23
Inspect API support for the new `spec.rules` field


lobkovilya:
### Description

Today, the Inspect API endpoint `/_rules` has `fromRules` field. As we're deprecating `spec.from` and replacing it with `spec.rules` we have to make appropriate changes to the `/_rules` endpoint as well.

Automaat:
blocked on: https://github.com/kumahq/kuma/issues/12863

PR #12713, opened 2025-01-30
feat(inspect-api): support for inbound rules

lobkovilya:
## Motivation

https://github.com/kumahq/kuma/pull/12560 added a new field `InboundRules` to `FromRules` structure:
```go
type FromRules struct {
	// Rules is a map of InboundListener to a list of rules built by using 'spec.from' field.
	// Deprecated: use InboundRules instead
	Rules map[InboundListener]Rules
	// InboundRules is a map of InboundListener to a list of inbound rules built by using 'spec.rules' field.
	InboundRules map[InboundListener][]*inbound.Rule
}
```

this structure needs to be visible in the Inspect API `/_rules` endpoint.

## Implementation information

It's a draft PR, at this moment only OpenAPI is changed.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12381


api/openapi/specs/common/resource.yaml
```diff
@@ -27,7 +27,7 @@ components:
           $ref: "#/components/schemas/PolicyDescription"
     PolicyDescription:
       type: object
-      required: [hasToTargetRef, hasFromTargetRef, isTargetRef]
+      required: [hasToTargetRef, hasFromTargetRef, isTargetRef, interpretFromEntriesAsRules]
```
lahabana, line 30: This is not right. `fromAliasForRules` maybe?

Have you tried to use a LLM to find a name? It's usually helpful for these things.

api/openapi/specs/common/resource.yaml
```diff
@@ -87,9 +90,20 @@ components:
             $ref: '#/components/schemas/ResourceRule'
         fromRules:
           type: array
-          description: a set of rules for each inbound of this proxy
+          description: a set of rules for each inbound of this proxy. The field is not set when 'interpretFromEntriesAsRules' on PolicyDescriptor is set to true.
```
johncowen, line 93: Question: Just wondering why we have a specific `interpretFromEntriesAsRules` property for the client to check, when (according to whats said in this description) they could also check for existence of `fromRules` or `inboundRules`?

(might have totally missed reasons for this in other issues etc so lmk!)

lahabana, line 93: Can we just keep it? And just mention that inboundRules are to be preferred.

Also can we rename `toResourceRules` to `outboundRules`?

So we'd have:

```json
proxyRule: {},
outboundRules: {},
inboundRules: {}
```
?

Also getting side tracked here. But I feel like maybe our APIs are just not working out here and we should rethink them.

Should we:

`GET /meshes/<mesh>/dataplanes/<dataplane>/_layout`
```yaml
labels:
   key: value
name: foo
inbounds:
  - port: "..."
    name: "..."
    labels: "..."
    envoy_entity_names: [...]
outbounds:
  - name: "..." # This would be the envoy name potentially
     envoy_entity_names: [...]
     entity: "<meta>" # This would be the meta the find the MeshService/MeshMZService...
  
proxyRules:
   - type: "<policyType>"
     warnings: []
     conf: "interface{}"
     origin: # respects the order in which they are applied
        - meta1
        - meta2
```

Once we have this we can have:

`GET /meshes/<mesh>/dataplanes/<dataplane>/_rules/_outbound/<identifier>`
```yaml
rules:
- type: <the conf type>
```

and something similar for inbounds.

I think this would:

a) make it easier for frontend to extract meaningful data for the DP view
b) make APIs way less verbose for users.


Obviously this would ONLY work with MeshService.

lahabana: Btw this gets rid of your weird "interpretFlag"

docs/generated/openapi.yaml
```diff
@@ -3129,9 +3153,26 @@ components:
...
+            true.
           items:
             $ref: '#/components/schemas/FromRule'
+        inboundRules:
+          type: array
+          description: >-
+            a set of rules for each inbound of the proxy. When policy descriptor
+            has 'interpretFromEntriesAsRules' set to true, this field replaces
+            'fromRules'.
+          items:
```
lahabana, line 3168: Where will the actual sub rules be?

api/openapi/specs/common/resource.yaml
johncowen: > Also getting side tracked here. But I feel like maybe our APIs are just not working out here and we should rethink them.

I agree with this sentiment. It feels like we need something that begins with the inbounds/outbounds (the "connections") and once I have those I can discover the rules for those (the GUI begins with the inbounds/outbounds also). Whereas the `_rules` endpoint feels more like the reverse, I begin with the rules and discover the inbounds/outbounds for those.

lobkovilya: For some policies like MeshTimeout, MeshAccessLog and others we implicitly treat legacy `spec.from` as `spec.rules` because the conversion is straightforward. That's why if this flag `interpretFromEntriesAsRules` is set for the policy you don't need to display both `fromRules` and `inboundRules`, the latter is sufficient.

For MeshTrafficPermission and MeshFaultInjection the conversion of `spec.from` to `spec.rules` is not straightforward. Legacy `spec.from` and new `spec.rules` policies will co-exist and that's why GUI needs to show both `fromRules` and `inboundRules`.

lobkovilya: > Once we have this we can have:
> 
> `GET /meshes/<mesh>/dataplanes/<dataplane>/_rules/_outbound/<identifier>`

If we use ResourceIdentifier as `<identifier>` then we probably don't need `/_outbound` part. Kinds `MeshService`, `MeshMultiZoneService` and `MeshExternalService` are always outbounds. Kind `Dataplane` is always inbound. 

We'd have to come up with `<identifier>` format for public API. Internally we're using `meshservice:mesh/mesh-1:namespace/ns-k8s:name/backend:section/http-port` and it's not ideal for the URL.

lahabana: Don't we have a universal resource identifier no?

lobkovilya: We have ResourceIdentifier as a code structure, it has method [`String()`](https://github.com/kumahq/kuma/blob/8c6aeca48a659bd11e5f7fde7bf62bcd05dfe22e/pkg/core/resources/model/resource.go#L965-L964) that produces a line like `meshservice:mesh/mesh-1:namespace/ns-k8s:name/backend:section/http-port`. But it's not part of public API

api/openapi/specs/common/resource.yaml
lobkovilya: Tried LLM, wdyt about `mapFromToRules`?

docs/generated/openapi.yaml
lobkovilya: Sorry, not sure I follow, what're the sub rules? 

The example of the following schema is:

```yaml
inboundRules:
  - inbound:
      port: 8080
      name: http-api
      tags: {}
    rules:
      - conf: $conf1
        origin: [...]
  - inbound:
      port: 9091
      name: tcp-server
      tags: {}
    rules:
      - conf: $conf2
        origin: [...]
```

api/openapi/specs/common/resource.yaml
schogges: Would it make sense to align it further with the names of the other booleans? Maybe like `isMappedFromToRules`? 🙂

api/openapi/specs/common/resource.yaml
schogges: Do I understand it right that in any case both `fromRules` and `inboundRules` are returned, regardless of the flag being `true` or `false`? In that case I think the description is a little misleading.

Otherwise, if this field is only set based on the conditions you mentioned, we probably don't need the flag as @johncowen questioned 🤔

Icarus9913:
Looks good to me

api/openapi/specs/common/resource.yaml
lobkovilya: The MADRs for [ResourceIdentifier](https://github.com/kumahq/kuma/pull/12756) and Inspect API updates are in progress, but we still want to merge this as an interim solution

api/openapi/specs/common/resource.yaml
```diff
@@ -106,6 +114,32 @@ components:
...
+    InboundRulesEntry:
+      type: object
+      required: [inbound, rules]
+      properties:
+        inbound:
+          $ref: '#/components/schemas/Inbound'
+        rules:
+          type: array
+          items:
+            $ref: '#/components/schemas/InboundRule'
```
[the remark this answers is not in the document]
lobkovilya: `conf` array can have multiple items only in case of using MeshHTTPRoute with `hostnames`. But since [we changed the core `MergeConf` function](https://github.com/kumahq/kuma/pull/8663/files#diff-0ba38a341b14b775bd61353c517d1a06fe30d5869e3496802cd31e051c139d81R22) to return `[]interface{}` and Inspect API is shared between MeshGateway and Sidecars, unfortunately we have `conf: []interface{}` in all rules

api/openapi/specs/common/resource.yaml
lobkovilya: well it's kind of mapped, there is the adapter that allows interpreting old `from` items as new `rules` https://github.com/kumahq/kuma/blob/d7ec0a2b1ac19208fb7dd9726309e3cf8cdc5848/pkg/plugins/policies/core/rules/inbound/inboundrules.go#L27-L33

also, not every policy allows mapping `from` to `rules`, i.e. MeshTrafficPermission and MeshFaultInjection don't support it

api/openapi/specs/common/resource.yaml
```diff
@@ -39,6 +39,9 @@ components:
         hasFromTargetRef:
           description: indicates that this policy can be used as an inbound policy
           type: boolean
+        interpretFromEntriesAsRules:
+          description: indicates that existing 'spec.from' entries are interpreted as 'spec.rules'
```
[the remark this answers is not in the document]
lobkovilya: > I believe this is only there for the transition and we'll soon have all policies have this to true right?

It won't be possible to have this `true` for MeshTrafficPermission and MeshFaultInjection in case we stop encoding DPP tags to the identity certificate. For these policies I expect users to have 2 layers of sequential envoy filters – one is created based on `from` and another one is created based on `rules`.

> Also if the end goal to remove the flag, should false which is the default be the future state? Might make it better to write future proof code?

We'll remove the flag when we remove `spec.rules`. Probably inverting the flag makes sense as there are only 2 policies with `interpretFromEntriesAsRules: false`

api/openapi/specs/common/resource.yaml
schogges: Hm, probably we should not include implementation details in the naming. How about short and concise `isFromAsRules`?

api/openapi/specs/common/resource.yaml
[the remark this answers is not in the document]
lahabana: Sounds good. Can we explain this in the description then?

api/openapi/specs/common/resource.yaml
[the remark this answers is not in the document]
lobkovilya: 1. Users already have MTPs in their system like:
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  name: allow-orders
  namespace: backend-ns
spec:
  from:
    - targetRef:
        kind: MeshSubset
        tags:
          kuma.io/service: orders
      default:
        action: Allow
```
2. To start using `rules` they need to create a new MTP:

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  name: allow-orders-v2
  namespace: backend-ns
spec:
  rules:
    - matches:
        spiffeID: spiffe://mesh-1/orders
      default:
        action: Allow
```
3. Now `allow-orders` and `allow-orders-v2` co-exist. The Inspect API is going to return:
```yaml
fromRules:
  - inbound: $inbound1
    rules:
      - matchers: # tags from cert matching
          - key: kuma.io/service
            value: orders
        conf: 
          action: Allow
inboundRules:
  - inbound: $inbound1
    rules:
      - matches: 
          spiffeID: spiffe://mesh-1/orders
        default:
          action: Allow
```
In Envoy we're going to have 2 RBAC filters – old cert-tag based and new spiffeId based (as long as `allow-orders` and `allow-orders-v2` allow/deny the same requests these 2 filters won't interfere) 
4. When migration is over the user removes `allow-orders`

api/openapi/specs/common/resource.yaml
```diff
@@ -90,6 +93,11 @@ components:
           description: a set of rules for each inbound of this proxy
           items:
             $ref: '#/components/schemas/FromRule'
+        inboundRules:
+          type: array
+          description: a set of rules for each inbound of the proxy. When policy descriptor has 'interpretFromEntriesAsRules' set to true, this field replaces 'fromRules'.
```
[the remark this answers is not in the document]
lobkovilya: this is relevant comment, clients shouldn't look into `fromRules` when the flag is set

api/openapi/specs/common/resource.yaml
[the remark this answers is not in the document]
lahabana: Starting to think that having matches for dp cert tags in inboundRules would make this simpler no?
If that's the case should we go with this so this flag can be a pure rename? (We could even warn in the API to suggest the rewrite)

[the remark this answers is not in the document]
lobkovilya: I think we already had this discussion at some point, the current MTP with cert tags has its complex algorithm that we want to get rid of favoring [the simpler approach with `matches`](https://docs.google.com/document/d/1tdIOVVYObHbGKX1AFhbPH3ZQKYvpQG-2s6ShHSECaXM/edit?tab=t.0):

> Extended GAPI matches order (based on [gateway-api](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteRule))
> * “Exact” spiffeID
> * “Prefix” spiffeID
> * “Exact” path match.
> * “Prefix” path match with largest number of characters.
> * Method match.
> * Largest number of header matches.
> * Largest number of query param matches.

api/openapi/specs/common/resource.yaml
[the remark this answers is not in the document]
lobkovilya: that's hypothetical, we don't have `matches` implemented by any policy yet

api/openapi/specs/common/resource.yaml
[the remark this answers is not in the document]
lobkovilya: they're both set and `inboundRules` supersedes, updated the description

api/openapi/specs/common/resource.yaml
```diff
@@ -106,6 +114,32 @@ components:
...
+          type: array
+          items:
+            $ref: '#/components/schemas/InboundRule'
+    InboundRule:
+      type: object
+      required: [conf, origin]
+      properties:
+        conf:
+          description: The actual conf generated
+          type: array
```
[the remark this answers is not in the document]
lobkovilya: Yes, explained below https://github.com/kumahq/kuma/pull/12713#discussion_r1967340616

api/openapi/specs/common/resource.yaml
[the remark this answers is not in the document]
lobkovilya: updated the doc

pkg/api-server/testdata/resources/inspect/dataplanes/_rules/dataplane_kind_meshtimeout_section_name.golden.json
```diff
@@ -76,6 +76,83 @@
      ]
```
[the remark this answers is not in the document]
lobkovilya: added a test case

api/openapi/specs/common/resource.yaml
lahabana: With a good enough comment should be enough

lahabana (approved):
🥳

Issue #12670, opened 2025-01-27
Remove deprecated code supporting underscore in a resource name


lukidzi:
### Description

We have a deprecated code that supports underscore in resource name. It should be removed in 2.7.x we had no issue to track it https://github.com/kumahq/kuma/blob/master/pkg/core/resources/apis/mesh/meta_validator.go#L32

PR #12714, opened 2025-01-30
chore(cleanup): remove deprecated code

lukidzi:
## Motivation

We deprecated using `_` in a resource name in 2.5.x and we were supposed to remove it in 2.7.x. but we forgot about it. This pr removes this function and adjust to the change

fix: https://github.com/kumahq/kuma/issues/12670


UPGRADE.md
```diff
@@ -12,6 +12,18 @@ does not have any particular instructions.
 
 A successful DELETE and PUT request without any warnings will now set "content-type: application/json" header and return empty json.
 
+### More strict validation rules for resource names
+
+In Kuma 2.7.x we deprecated usage of non [RFC 1123](https://www.rfc-editor.org/rfc/rfc1123.html) characters and since 2.10.x it's not more possible to create or update non RFC complient resource name. In order to be compatible with Kubernetes naming policy we updated the validation rules. Old rule:
```
lahabana, line 17:
```suggestion
In Kuma 2.7.x we deprecated usage of non [RFC 1123](https://www.rfc-editor.org/rfc/rfc1123.html) characters and since 2.10.x it's no longer possible to create or update non RFC compliant resource name. In order to be compatible with Kubernetes naming policy we updated the validation rules. Old rule:
```

lahabana (approved):

bartsmykla (approved):

PR #12741, opened 2025-02-03
chore(e2e): add envoy config test for MeshTLS

Automaat:
## Motivation

We need more tests before we can implement new rules api for MeshTLS


slonka (dismissed):

lobkovilya:
please add another test case for permissive mode

lobkovilya (approved):

PR #12744, opened 2025-02-04
fix(e2e): fix flaky envoyconfig e2e test by waiting for tls ready on MeshService

Automaat:
## Motivation

Envoy config e2e test for MeshTimeout was flaky because it ran before MeshService identities were fully initialized


test/e2e_env/universal/envoyconfig/envoyconfig.go
```diff
@@ -50,6 +51,18 @@ func EnvoyConfigTest() {
 			Setup(universal.Cluster)
 		Expect(err).ToNot(HaveOccurred())
 
+		Eventually(func(g Gomega) {
+			spec, status, err := GetMeshServiceStatus(universal.Cluster, "test-server", meshName)
```
lobkovilya, line 55: you have to check both test-server and demo-client

Automaat: added

lukidzi (approved):

test/e2e_env/universal/envoyconfig/envoyconfig.go
```diff
@@ -50,6 +51,32 @@ func EnvoyConfigTest() {
 			Setup(universal.Cluster)
 		Expect(err).ToNot(HaveOccurred())
 
+		// demo-client identity ready
```
lobkovilya, line 54: I'd just add a local function inside `EnvoyConfigTest`
```go
waitMeshServiceReady := func(name string) {
	Eventually(func(g Gomega) {
		spec, status, err := GetMeshServiceStatus(universal.Cluster, name, meshName)
		g.Expect(spec.Identities).To(Equal([]meshservice_api.MeshServiceIdentity{
			{
				Type:  meshservice_api.MeshServiceIdentityServiceTagType,
				Value: name,
			},
		}))
		g.Expect(status.TLS.Status).To(Equal(meshservice_api.TLSReady))
}
```

Issue #10907, opened 2024-07-16
MeshFaultInjection validation error message should be more specific


slonka:
### Description

```
apiVersion: kuma.io/v1alpha1
kind: MeshFaultInjection
metadata:
  name: default
  namespace: kong-mesh-system
  labels:
    kuma.io/mesh: default # optional, defaults to `default` if it isn't configured
spec:
  targetRef:
    kind: Mesh
  to:
    - targetRef:
        kind: MeshService
        name: httpbin
      default:
        http:
          - abort:
              httpStatus: 500
              percentage: 50
```

```
kubectl apply -f mfi.yaml
The MeshFaultInjection "default" is invalid: spec.from: must not be defined
```

the core question here is: why?

the answer is in the code: https://github.com/kumahq/kuma/blob/cf056262652a58535baedceeb91f2409d826ecbd/pkg/plugins/policies/meshfaultinjection/api/v1alpha1/validator.go#L48 it also targets gateways but just from the error code we don't know :(

bcollard:
Hit this, too.
The issue is that the MeshFaultInjection validator checks for the presence of `proxyType` in the `spec.targetRef` 
The MFI policy works OK when the following is added:
```
proxyTypes: ["Sidecar"]
```

Looks like MeshRateLimit would have the same behavior

PR #12754, opened 2025-02-04
fix(policy): added better message when no proxyTypes

lukidzi:
## Motivation

The validation message wasn't helpful and a user had an issue to understand what was wrong.

## Implementation information

Improved a message to give more details

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/10907


pkg/plugins/policies/meshfaultinjection/api/v1alpha1/validator_test.go
```diff
@@ -207,7 +207,7 @@ from:
 			[]validators.Violation{
 				{
 					Field:   "spec.from",
-					Message: "must not be defined",
+					Message: "must not be defined when the scope includes a Gateway, exclude non-gateway resources or select only gateways and use spec.to",
```
lobkovilya, line 210: "exclude non-gateway" has the same meaning as "select only gateways" :) probably should be
```
must not be defined when the scope includes a Gateway, select only proxyType Sidecar or select only gateways and use spec.to
```

lobkovilya (approved):

Issue #12147, opened 2024-12-02
Replace `-` with `_` before hash in identifiers in xDS config


bartsmykla:
### What happened?

xref: https://github.com/kumahq/kuma-gui/issues/3237

We are currently using `-` as a separator in the identifiers for built-in gateway in xDS configs whenever hash is present, which might cause problems.

Hypothetical situation is when instead of ports we'll decide to use sectionNames and if sectionName will contain `-` in the name

Needs information:
- Kubernetes / DNS RFC which dictates port names - add link in this issue

lukidzi:
> https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#serviceport-v1-core
> The name of this port within the service. This must be a DNS_LABEL. All ports within a ServiceSpec must have unique names. When considering the endpoints for a Service, this must match the 'name' field in the EndpointPort. Optional if only one ServicePort is defined on this service.

> [rfc1035](http://www.ietf.org/rfc/rfc1035.txt)/[rfc1123](http://www.ietf.org/rfc/rfc1123.txt) label (DNS_LABEL): An alphanumeric (a-z, and 0-9) string, with a maximum length of 63 characters, with the '-' character allowed anywhere except the first or last character, suitable for use as a hostname or segment in a domain name.


https://github.com/kubernetes/design-proposals-archive/blob/main/architecture/identifiers.md

lahabana:
@lobkovilya to look whether this should be done in the MADR linked above or if it's an issue by itself.

lobkovilya:
We've put this out of the scope of Resource Identifier MADR. We should think about how we're going to append a hash suffix to `KRI` in https://github.com/kumahq/kuma/issues/13267

lukidzi:
@slonka @bartsmykla was it considered during the MADR?

slonka:
> triage: with unified naming we don't allow `-` in section name so closing this one

PR #12756, opened 2025-02-05
docs(MADR): resource identifier format

lobkovilya:
## Motivation

The goal is to improve Inspect API and introduce an identifier as part of the URL path, i.e `:5681/_rules/<identifier>`. See [the discussion](https://github.com/kumahq/kuma/pull/12713#discussion_r1936824883)

Better to review the [rendered version](https://github.com/kumahq/kuma/blob/2b4f3edebe9091fd3c1328191640de5c2cb9121b/docs/madr/decisions/070-resource-identifier.md) as it contains tables.

Fix https://github.com/kumahq/kuma/issues/12044


docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,99 @@
...
+
+**Cons:**
+- Hard to read when names are poorly chosen, e.g., `kri:default:default:default:meshservice:backend`
+
+### Option 2 - Field names in the identifier
+
+We need to pick two delimiters: one to separate keys from values and another to separate key-value pairs.
+
+```yaml
+kri:meshservice:mesh=mesh-1:zone:zone-1:namespace=kuma-demo:name=backend:section=http-port
```
Automaat, line 91: Would it be possible to use random order and specify only some of the fields, to select multiple resources? 

If not then I think not using field name makes more sense as you still need to remember order and fields

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,99 @@
...
+When mixing identifiers of different resource types, an extended version is used:
+
+```go
+type TypedResourceIdentifier struct {
+    ResourceIdentifier
+
+    ResourceType ResourceType
+    SectionName  string
+}
+```
```
lahabana, line 39: Any reason to not always have the more complete version and make things easier?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,99 @@
...
+- A resource is created on the Global CP and synced down to all Zone CPs.
+- A resource is created on a Zone CP and synced to the Global CP.
+- A resource is created on a Zone CP and synced to other Zone CPs.
+
+For example, the Zone CP in `zone-1` can possess the following MeshTimeouts:
+- MeshTimeouts created in `zone-1`
+- MeshTimeouts created on the Global CP and synced to `zone-1`
+- MeshTimeouts created in `zone-2` and synced to `zone-1`
+
+To uniquely identify a resource regardless of its place of origin, Kuma uses the following Go structure:
```
lahabana, line 19: In the ctx you need to specify where we're planning to use it. And then we need to check that the charset of this fits the places you are planning to use it in

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,99 @@
+# Resource Identifier
+
+* Status: accepted
+
+## Context and Problem Statement
```
lahabana, line 5: There's prior work around this from @Automaat and @lukidzi we should reference it here.

Lukasz made a MADR about cluster names
Automaat made a MADR about internal envoy entities

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,99 @@
...
+For example:
+```
+kri:mesh-1:us-east-2:kuma-demo:meshservice:backend
+kri:mesh-1:us-east-2:kuma-demo:meshservice:backend:http-port
+kri:mesh-1:::meshtimeout:global-timeouts
+```
+
+Having a prefix like `kri` (Kuma Resource Identifier) is useful for two reasons:
+* It visually clarifies the format for users, who can then search for the format description in the documentation.
+* It acts as an implicit version. If we need to update the format, we can use a different prefix (e.g., `uri` or `ri`).
```
lahabana, line 77: IMO we should either always have the type or provide 2 option with and without the type. Otherwise it's hard to figure which one it is (my 2c is that it should be the full with the type always).

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,99 @@
...
+## Pros and Cons of the Options
+
+### Option 1 - Order-based, no field names in the identifier
+
+We need to pick a delimiter that's allowed in the URL path and not allowed in resource names, meshes, namespaces, or zones. Good candidates are `:` and `_`.
+
+There is an identifier format from Amazon called [ARN](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html). We can adopt a similar approach:
+
+```
+kri:<mesh>:<zone>:<namespace>:<resource-type>:<resource-name>:<section-name>
```
lahabana, line 66: What's the charset of each of these components?

I think mesh/zone/namespace/rType/rName is straightforward but how is it for section name?

What are the current possible section name?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,99 @@
...
+type TypedResourceIdentifier struct {
+    ResourceIdentifier
+
+    ResourceType ResourceType
+    SectionName  string
+}
+```
+
+Currently, the resource identifier lacks a string representation and is not exposed to the Kuma public API. The goal of this MADR is to propose a format.
+
```
lahabana, line 42: Should we mention that this identifier will be usable in the API to at least get the value?

docs/madr/decisions/070-resource-identifier.md
lahabana: What's the value for zone when the entity is global?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,99 @@
...
+## Decision Drivers
+
+- Possible to use the identifier in a URL path as `:5681/_rules/<identifier>`
```
lahabana, line 45: Can we consider using this for xds entities too?

docs/madr/decisions/070-resource-identifier.md
lahabana: It's also fine mentioning the PR and the few issues that mention the need for this

docs/madr/decisions/070-resource-identifier.md
lobkovilya: > What's the value for zone when the entity is global?

It's an empty string `""`, i.e. one of the examples below `kri:mesh-1:::meshtimeout:global-timeouts`

docs/madr/decisions/070-resource-identifier.md
lobkovilya: By this line:
> It acts as an implicit version. If we need to update the format, we can use a different prefix (e.g., `uri` or `ri`).

I just wanted to say why it's good to have a special prefix in the resource identifier like "kri" or something else. 

Resource types like `meshservice` or `meshtimeout` should always be part of the identifier.

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,99 @@
...
+- Human-readable: users should be able to type the identifier manually if needed
+
+## Considered Options
+
+* Option 1 - Order-based, no field names in the identifier
+* Option 2 - Field names in the identifier
```
Icarus9913, line 51: Is it possible to use json body to pass the properties(zones, namespaces, meshes)?
Once if we have one property with empty value, it looks weird in the url. My first sense about option 1 and 2 is too long or order-based looks complicated

docs/madr/decisions/070-resource-identifier.md
lobkovilya: No, fields should be sorted, I want resource identifier to be comparable by `==`. When it comes to multiple resources that's a different feature – implement resource selector using the same format. It will be possible in the future.

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Referenced both MADRs in the context.

docs/madr/decisions/070-resource-identifier.md
lobkovilya: yes, updated

docs/madr/decisions/070-resource-identifier.md
lobkovilya: right, no reason to have both https://github.com/kumahq/kuma/issues/12831

docs/madr/decisions/070-resource-identifier.md
lobkovilya: We also want to use resource identifier when naming Envoy resources, that's why we'd like to have a stringified format

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+- MeshTimeouts created in `zone-2` and synced to `zone-1`
+
+To uniquely identify a resource regardless of its place of origin, Kuma uses the following Go structure:
+
+```go
+type ResourceIdentifier struct {
+    Name      string
+    Mesh      string
+    Namespace string
+    Zone      string
```
Automaat, line 26: what is in zone when policy originates from global? I don't remember and it might be good to add this

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+| VirtualHost       | legacy listeners - `<kuma.io/service>`<br>new outbounds - `<mesh>_<name>_<namespace>_<zone>_<short-name>_<port>`                                                                                        | Mesh*Service (with sectionName to select port) |
+| Inbound Cluster   | `localhost:<port>`                                                                                                                                                                                      | Dataplane (with sectionName to select port)    |
+| Outbound Cluster  | legacy clusters - `<kuma.io/service>-hash(dst.tags)`<br>legacy clusters cross-mesh - `<kuma.io/service>-hash(dst.tags)_<mesh>`<br>new clusters - `<mesh>_<name>_<namespace>_<zone>_<short-name>_<port>` | Mesh*Service (with sectionName to select port) |
+| Route             | Routes are set on Listener on VirtualHost.<br>On inbound - `inbound:<kuma.io/service>`<br>On outbound - `<hash_sha256([]Match{...})>`                                                                   | Correlates with a set of MeshHTTPRoutes        |
```
Automaat, line 148: what about `matches` section in new inbound policies api? We might not have a route as a resource but create route from policy

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+| Secret            | `name       = <category>:<scope>:<identifier>`<br>`category   = "mesh_ca" \| "identity_cert"`<br>`scope      = "secret"`<br>`identifier = "all" \| <mesh_name>`                                         | TODO                                           |
+
+##### Builtin Gateway
+
+|             | Name                                                                                                                                                                                                                           | Correlated Resources                    |
+|-------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------|
+| Listener    | `<gateway-name>:<protocol>:<port>` where `gateway-name` is `MeshGatewayResource.Meta.Name`                                                                                                                                     | MeshGateway                             |
```
Automaat, line 155: gateway listeners have name, so maybe this should be treated as `sectionName`? in this case

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+## Pros and Cons of the Options
+
+### Option 1 - Order-based, no field names in the identifier
+
+We need to pick a delimiter that's present in the `delimiter` charset we've specified above.
+
+There is an identifier format from Amazon called [ARN](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html). We can adopt a similar approach, but using `_`:
+
+```
+kri_<mesh>_<zone>_<namespace>_<resource-type>_<resource-name>_<section-name>
```
Automaat, line 299: did we considered having some placeholder for missing values? to avoid multiple _ which can be hard to read?

docs/madr/decisions/070-resource-identifier.md
lobkovilya: it's empty, I added a comment

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Even for outbound routes it says `Correlates with a set of MeshHTTPRoutes`. So it's not really clear how to name the route without hashing

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+    }
+   }
+  }
+ ]
+}
+```
+
+#### [Issue #3249](https://github.com/kumahq/kuma/issues/3249): local envoy clusters should have a better name than localhost_<port> 
+
+Today local clusters have name `localhost:<port>` and in stats they're visible as `localhost_<port>`.
```
lahabana, line 84:
```suggestion
Today inbound clusters have name `localhost:<port>` and in stats they're visible as `localhost_<port>`.
```

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+That's not very informative.
+
+#### [Issue #12733](https://github.com/kumahq/kuma/issues/12733): Envoy statPrefix should not use ip addresses
+
+Using IP address in envoy config statPrefix field can cause issues or generate unnecessary changes in metrics.
+This is also problematic for our e2e tests that test envoy config using golden files.
+
+#### [Issue #12093](https://github.com/kumahq/kuma/issues/12093): xds configs, outbound listeners should use the clustername instead of an IP/port combo
+
+We name outbounds like `outbound:10.43.205.116:6379` where IP address doesn't give any useful information.
```
lahabana, line 94: Even more when there are multiple IPs right?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+                 / "*" / "+" / "," / ";" / "="
+```
+
+#### Envoy resource names
+
+For the context, there is already [MADR](036-internal-listeners.md) that regulated the name of internal listeners.
+Also, there was [work](https://docs.google.com/document/d/1OIZK82Tr-4El2FfdlBn7WNRZ7FatkTuEcZKH0FlSTMA/edit?tab=t.0#heading=h.n6cmlf1eel2z) related to Envoy cluster name unification, but it's not finished.
+Discoveries in this work helped me to fill the tables.
+
+There are no restriction on the name format from the Envoy's side.
```
lahabana, line 134: Even in length?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+
+In Envoy resource names we can use any character except `:` if we want to solve [Issue #2363](https://github.com/kumahq/kuma/issues/2363) without workaround.
+
+In Envoy stats fields we can use any character except `:` to avoid on the fly conversion of `:` to `_` for Cluster names in Prometheus labels.
+
+This leaves us with the following charset:
+
+```
+resource-identifier = *(ALPHA / DIGIT / "-" / "." / "_" / "~" / "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "=" / "@")
+```
```
lahabana, line 241: Shouldn't we use something that is compatible with the statsD charsets for label values?
It doesn't seem that otel has limitations but we should at least mention it here: https://opentelemetry.io/docs/specs/semconv/

docs/madr/decisions/070-resource-identifier.md
lahabana: You need to be explicit about resource-type. Is it the plural/camlCase...?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+
+* Option 1 - Order-based, no field names in the identifier
+
+## Pros and Cons of the Options
+
+### Option 1 - Order-based, no field names in the identifier
+
+We need to pick a delimiter that's present in the `delimiter` charset we've specified above.
+
+There is an identifier format from Amazon called [ARN](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html). We can adopt a similar approach, but using `_`:
```
lahabana, line 296: It's also ensure no overlap between internal entities (prefixed by _kuma) and the rest.

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+```
+delimiter = "_" / "~" / "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "=" / "@"
+```
+
+## Decision Drivers
+
+- New Inspect API endpoints that'd allow getting resolved configuration by the resource identifier
+- Kuma API GET endpoints shouldn't require core name to read the resource 
+- Human-readable: users should be able to type the identifier manually if needed, so it's not hashed
+- Envoy resources that have direct correlation with Kuma resources should be named by using resource identifier, i.e. outbound cluster can be named after MeshService
```
lahabana, line 279: I think another decision driver is being able to use these in prom queries.

.e.g:

```
my_metric{envoy_cluster_name~="kri.*_meshservice_.*"} // Get only actual outbounds from meshservice
```

```
my_metric{envoy_cluster_name~="kri_default_.*.*"} // Get only stuff on the default mesh
```

docs/madr/decisions/070-resource-identifier.md
lahabana: Why not have the resource type before the mesh or at least before the zone?

lahabana (commented):
Great stuff

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -38,12 +38,245 @@ type TypedResourceIdentifier struct {
...
+|-------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------|
+| Cluster     | `meshpassthrough_<protocol>_<match-value>_<port>`<br>when `<port> == 0` Kuma sets port equal to `*`<br>`match-value = CIDR \| IP \| Domain`<br>`CIDR     = i.e. "192.0.2.0/24" or "2001:db8::/32"`<br>`IP       = i.e. "192.0.2.1", or 2001:db8::68", or ::ffff:192.0.2.1"`<br>`Domain   = <dns-name> \| *.<dns-name>`<br>`dns-name = ^([a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62}){1}(\.[a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62})*[\._]?$`<br>  | –                    |
```
Icarus9913, line 183: > passthroughMode: (Optional) Defines behaviour for handling traffic. Allowed values: All, None and Matched. Default: None

What does the identifier should be if `All` and `None` ?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,332 @@
...
+Using IP address in envoy config statPrefix field can cause issues or generate unnecessary changes in metrics.
+This is also problematic for our e2e tests that test envoy config using golden files.
+
+#### [Issue #12093](https://github.com/kumahq/kuma/issues/12093): xds configs, outbound listeners should use the clustername instead of an IP/port combo
+
+We name outbounds like `outbound:10.43.205.116:6379` where IP address doesn't give any useful information.
+
+### Places to use resource identifier
+
+#### URL path
```
schogges, line 98: Could it be that the resource identifier is being used in URL search query, i.e. for filtering?

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Envoy doesn't specify a length limit. I tried to create a cluster with the max expected length of resource identifier 
```
253(name) + 253(zone) + 63(mesh) + 63(namespace) + 15(sectionName) + 30(resourcetype) + 3(kri) + 6(_) = 686
```
and it worked as expected.

docs/madr/decisions/070-resource-identifier.md
lobkovilya: AFAIU statsD as a format doesn't have a way to put a label or tag on a metric https://github.com/statsd/statsd/issues/619, i.e. DD had to extend the format and created [DogStatsD](https://docs.datadoghq.com/developers/dogstatsd/?tab=hostagent#tags). 

For otel attributes there are indeed no limitation, I'm going to add this to the doc https://opentelemetry.io/docs/specs/otel/common/?utm_source=chatgpt.com#attribute

docs/madr/decisions/070-resource-identifier.md
lobkovilya: > did we considered having some placeholder for missing values? to avoid multiple _ which can be hard to read?

what would you use as a placeholder?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -38,12 +38,245 @@ type TypedResourceIdentifier struct {
...
+| Inbound Listener  | `inbound:10.43.205.116:8080`<br>`inbound:[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:8080`                                                                                                                | Dataplane (with sectionName to select port)    |
+| Outbound Listener | `outbound:10.43.205.116:8080`<br>`outbound:[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:8080`                                                                                                              | Mesh*Service (with sectionName to select port) |
+| VirtualHost       | legacy listeners - `<kuma.io/service>`<br>new outbounds - `<mesh>_<name>_<namespace>_<zone>_<short-name>_<port>`                                                                                        | Mesh*Service (with sectionName to select port) |
+| Inbound Cluster   | `localhost:<port>`                                                                                                                                                                                      | Dataplane (with sectionName to select port)    |
```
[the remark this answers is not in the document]
lobkovilya: Yes, the function `GetLocalClusterName` is used in multiple places https://github.com/kumahq/kuma/blob/0fbf884b4bcee672ee93ca5d748b3293be7f3d9a/pkg/xds/envoy/names/resource_names.go#L26

docs/madr/decisions/070-resource-identifier.md
lobkovilya: yeah, we actually removing them when MeshPassthrough is used https://github.com/kumahq/kuma/blob/6c44719740ef47d1de10564c5db094f946f29989/pkg/plugins/policies/meshpassthrough/plugin/v1alpha1/plugin.go#L98

docs/madr/decisions/070-resource-identifier.md
lobkovilya: As it's not a hard requirement, I added as a note what resource identifier and delimiter charset would look like if we wanted to support URL query:
```
resource-identifier = *(ALPHA / DIGIT / "-" / "." / "_" / "~" )
delimiter           = "_" / "~" 
```
they're significantly smaller than those without query support. 

But the good news is if we go with `_` then the resource identifier can be used in a query. So I added this to the `Pros` list

docs/madr/decisions/070-resource-identifier.md
lobkovilya: yeah, I think that's exactly the idea in [Gateway API](https://gateway-api.sigs.k8s.io/concepts/api-overview/#how-it-works):
> SectionName When sectionName is set, the Route selects the listener with the specified name.

I updated the doc

docs/madr/decisions/070-resource-identifier.md
schogges: Great, thank you 🙂

johncowen: (P.S. but one small benefit of not having non-URL safe characters is that it keeps the identifier "pretty" i.e. more recognisable, so a benefit but not super important I would say)

schogges: Still good point @johncowen, but as you said I also think generally it'd be better to not use any non-URL safe chars.

Also depending on what people/tools are using, there might be a difference on which chars are encoded (i.e. `encodeURI` vs `encodeURIComponent`:

```javascript
encodeURI("http://localhost:1234?filter[foo&bar]=baz") // -> 'http://localhost:1234?filter%5Bfoo&bar%5D=baz'

encodeURIComponent("filter[foo&bar]=baz") // -> 'filter%5Bfoo%26bar%5D%3Dbaz'
```

[encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI#description)
> encodeURI() escapes all characters except:
>
>```
>A–Z a–z 0–9 - _ . ! ~ * ' ( )
>
>; / ? : @ & = + $ , #
> ````
>
>The characters on the second line are characters that may be part of the URI syntax, and are only escaped by encodeURIComponent(). Both encodeURI() and encodeURIComponent() do not encode the characters -.!~*'(), known as "unreserved marks", which do not have a reserved purpose but are allowed in a URI "as is". (See [RFC2396](https://datatracker.ietf.org/doc/html/rfc2396))

[encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#description)
> encodeURIComponent() uses the same encoding algorithm as described in [encodeURI()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI). It escapes all characters except:
>
>```
>A–Z a–z 0–9 - _ . ! ~ * ' ( )
>```
>
>Compared to [encodeURI()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI), encodeURIComponent() escapes a larger set of characters. Use encodeURIComponent() on user-entered fields from forms [POST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST)'d to the server — this will encode & symbols that may inadvertently be generated during data entry for [character references](https://developer.mozilla.org/en-US/docs/Glossary/Character_reference) or other characters that require encoding/decoding. For example, if a user writes Jack & Jill, without encodeURIComponent(), the ampersand could be interpreted on the server as the start of a new field and jeopardize the integrity of the data.

johncowen: Yeah gotcha, couple of follow ups:

I'd say people should always use `encodeURIComponent` otherwise they are "holding it wrong" and should change implementation, and I suppose this is my point on expecting people to do always do this how we expect/correctly, there will always be cases where some folks might be "holding it wrong". To be fair, all "values" used in a URL should be using `encodeURIComponent` (or non-JS equivalent) anyway, whether they are expected to be safe or not. So there's also something to be said for not worrying about URL safety, if someone isn't using a correct encoder, they should be. But I'm thinking more about informal "I just want to curl the thing to get a response in my terminal" type of usage. There's benefit in not forcing people to have to encode the thing if for example we chose to use `/` in this case.

So if we want URL safe if only for reasons of "don't make it hard for people to just curl the thing" all in all according to the MADR, that leaves us with:

```
delimiter = "_" / "~"
```

And it sounds like we've landed on a `_`, which is URL safe which is super duper. It's probably a good idea to note that I have seen instances of people using these in hostnames even though a `_` shouldn't be used in hostnames.

@lobkovilya I'm not sure if we validate things like mesh names and zone names to not have `_`, I might be misremembering but do I remember that at least at one point this was possible? Is it definitely not possible to have a mesh/zone name with a `_` in it now?

Just a little side note that's just occurred to me, I'm kinda glad we still have this last "safe character" available `~`, which in a past life has been super useful to have as a usable/meaningful character (i.e. similar to `~/johncowen`), which kinda means "expand `~` to a common string we know about". You never know we might hit a thing at somepoint where we need the same "trick".

docs/madr/decisions/070-resource-identifier.md
johncowen: > Why not have the resource type before the mesh or at least before the zone?

I kinda agree with is, if the resource-type is before anything else, then you can immediately tell whether to expect say an empty mesh in the case of the kri pointing to something that doesn't have a `mesh`

johncowen:
Maybe not part of the MADR and might just be an informal example, but!

> The goal is to improve Inspect API and introduce an identifier as part of the URL path, i.e :5681/_rules/<identifier>. See https://github.com/kumahq/kuma/pull/12713#discussion_r1936824883

When we eventually come to define the endpoint should we include the fact that we are specifically requesting via a `kri` i.e. `:5681/_rules/kri/<identifier>` or `:5681/kri/_rules/<identifier>` or `:5681/_rules/<identifier>/kri`, basically include `kri` in another segment. That way if we ever add another way/identifier to request things we can distinguish via normal URL routing rather than having a single route check for the type of identifier.

Super edge case, but who knows if we are ever gonna change the way we define identifiers. but maybe I'm over thinking it 🤷

docs/madr/decisions/070-resource-identifier.md
lahabana: > I'm not sure if we validate things like mesh names and zone names to not have _, I might be misremembering but do I remember that at least at one point this was possible? Is it definitely not possible to have a mesh/zone name with a _ in it now?

I believe it's no longer supported as this was needed for correct k8s support. But @lobkovilya should check/confirm. I think @jijiechen just removed the back compat in fact

docs/madr/decisions/070-resource-identifier.md
lahabana: And I think there's no lims for prom label length either: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

lahabana: otel has a length limit defined in the SDK api: https://opentelemetry.io/docs/specs/otel/common/#attribute-limits

However, it states the default in infinity.
We should add these in this MADR for reference.

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+
+#### Envoy resource names
+
+For the context, there is already [MADR](036-internal-listeners.md) that regulated the name of internal listeners.
+Also, there was [work](https://docs.google.com/document/d/1OIZK82Tr-4El2FfdlBn7WNRZ7FatkTuEcZKH0FlSTMA/edit?tab=t.0#heading=h.n6cmlf1eel2z) related to Envoy cluster name unification, but it's not finished.
+Discoveries in this work helped me to fill the tables.
+
+There are no restriction on the name format from the Envoy's side.
+Some Envoy resources could be directly correlated with Kuma resources and that's why we should consider renaming them using the resource identifier.
+There are tables below for all Kuma proxies. Resources in `Internal` table exist in all proxies.
```
lahabana, line 143:
```suggestion
The tables below are an inventory of resource usage for all Kuma proxies. Resources in `Internal` table exist in all proxies. 
```

docs/madr/decisions/070-resource-identifier.md
lahabana: Ho that's quite an important issue no? AFAIR there's no "merging" in routes no? Can't we use the child route that matches this branch with the sectionName?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+| Inbound Cluster   | `localhost:<port>`                                                                                                                                                                                      | Dataplane (with sectionName to select port)    |
+| Outbound Cluster  | legacy clusters - `<kuma.io/service>-hash(dst.tags)`<br>legacy clusters cross-mesh - `<kuma.io/service>-hash(dst.tags)_<mesh>`<br>new clusters - `<mesh>_<name>_<namespace>_<zone>_<short-name>_<port>` | Mesh*Service (with sectionName to select port) |
+| Route             | Routes are set on Listener on VirtualHost.<br>On inbound - `inbound:<kuma.io/service>`<br>On outbound - `<hash_sha256([]Match{...})>`                                                                   | Correlates with a set of MeshHTTPRoutes        |
+| Secret            | `name       = <category>:<scope>:<identifier>`<br>`category   = "mesh_ca" \| "identity_cert"`<br>`scope      = "secret"`<br>`identifier = "all" \| <mesh_name>`                                         | TODO                                           |
```
lahabana, line 156: That'd be the kri no?

docs/madr/decisions/070-resource-identifier.md
lahabana: Also routes are specific to a listener IIRC no?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+|-------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------|
+| Listener    | `<gateway-name>:<protocol>:<port>` where `gateway-name` is `MeshGatewayResource.Meta.Name`                                                                                                                                     | MeshGateway (with sectionName to select the listener) |
+| VirtualHost | `<hostname>`                                                                                                                                                                                                                   |                                                       |
+| Cluster     | `<kuma.io/service>-hash(merge(dpp.tags, gateway.tags, listener.tags))`                                                                                                                                                         | Pair of MeshGateway and Mesh*Service                  |
```
lahabana, line 164: So what does this actually look like?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+	return illegalChars.ReplaceAllString(metric, "_")
+}
+```
+
+I can't find explanation why `alt_stat_name` and `stat_prefix` are treated differently by Envoy. 
+When Envoy exports metrics in Prometheus format resource name is always emitted as label value, it's never part of the metric name, i.e:
+```
+envoy_cluster_upstream_cx_total{envoy_cluster_name="localhost_5000"} 0
+```
+so there is no reason to replace `:` with `_`.
```
lahabana, line 225: In the past there were stuff in the metric name, I remember we even had some code to change things around. My is that this is the main reason no?

Also when not using prom format doesn't it actually use `:` as a separator see:

image

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+
+// We need to sanitize metrics in order to  not break statsd and prometheus format.
+// StatsD only allow [a-zA-Z_\-0-9.] characters, everything else is removed
+// Extra dots breaks many regexes that converts statsd metric to prometheus one with tags
+func SanitizeMetric(metric string) string {
+	return illegalChars.ReplaceAllString(metric, "_")
+}
+```
+
+I can't find explanation why `alt_stat_name` and `stat_prefix` are treated differently by Envoy.
```
lahabana, line 220: A use case could be that `stat_prefix` and `alt_stat_name` parameters can be used to consolidate statistics from multiple similar entities under a single name.

For example, if you have separate IPv4 and IPv6 clusters named `my-cluster:v4` and `my-cluster:v6`, you can use `alt_stat_name: my-cluster` for both clusters to aggregate their statistics under the common name `my-cluster` when exporting metrics. This allows you to track related entities as a unified group while maintaining distinct names for configuration purposes.

WDYT?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+I can't find explanation why `alt_stat_name` and `stat_prefix` are treated differently by Envoy. 
+When Envoy exports metrics in Prometheus format resource name is always emitted as label value, it's never part of the metric name, i.e:
+```
+envoy_cluster_upstream_cx_total{envoy_cluster_name="localhost_5000"} 0
+```
+so there is no reason to replace `:` with `_`.
+
+If Envoy tried to support StatsD format then there are still a question, why `stat_prefix` emits metrics with `:` but `alt_stat_name` replaces `:` with `_`.
+Also, StatsD format doesn't seem to be supported out-of-the-box by Envoy and requires configuring `stats_sinks` in the static config.
+
```
lahabana, line 229: Whatever the reason for this, with our new format, do we end up without the need for specifying `stat_prefix` and `alt_stat_name` ? If that's the case that's probably something we should mention and decide whether we still set it (in case) or we don't

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+ 
+In URL path segment we can use:
+
+```
+ALPHA / DIGIT / "-" / "." / "_" / "~" / "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "=" / ":" / "@"
+```
+
+In Envoy resource names we can use any character except `:` if we want to solve [Issue #2363](https://github.com/kumahq/kuma/issues/2363) without workaround.
+
+In Envoy stats fields we can use any character except `:` to avoid on the fly conversion of `:` to `_` for Cluster names in Prometheus labels.
```
lahabana, line 242:
```suggestion
In Envoy stats fields we can use any character except `:` to avoid on the fly conversion of `:` to `_` for Cluster names in Prometheus label values.
```

Labels have keys and values it's worth being thourough

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+
+OpenTelemetry defines "Attribute", it's a key-value pair similar to Prometheus labels.
+There are [no charset limitation on attribute's key or value](https://opentelemetry.io/docs/specs/otel/common/?utm_source=chatgpt.com#attribute).
```
lahabana, line 245:
```suggestion
There are [no charset limitation on attribute's key or value](https://opentelemetry.io/docs/specs/otel/common).
```

Worth removing the utm when adding links

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+  * contain only lowercase alphanumeric characters, '-' or '.'
+  * start with an alphanumeric character
+  * end with an alphanumeric character
+* `namespace`
+  * contain at most 63 characters
+  * contain only lowercase alphanumeric characters or '-'
+  * start with an alphanumeric character
+  * end with an alphanumeric character
+* `resourceType`
+  * only alphanumeric
```
lahabana, line 279: There's no validation for this atm added: https://github.com/kumahq/kuma/issues/12861

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+* `sectionName`
+  * contain at most 15 characters
+  * contain only lowercase alphanumeric characters or '-'
```
lahabana, line 282: Ho yeah? Where is this defined/validated?

docs/madr/decisions/070-resource-identifier.md
lahabana: Makes sense for the ordering

lahabana (commented):
This is good. I think the only gap for getting an approval are:

1. showing actual new names in your table once the whole migration is done
2. telling the future of `stat_prefix`/`alt_stat_name`

The rest are mostly small questions/nits

johncowen:
I just had a thought/question which is maybe just for my understanding, but how come something like SPIFFE IDs aren't an option here? I think that even from my level of understanding, I guess the `/`'s etc in SPIFFE URIs are problematic for all the reasons why we've landed on a `_`? Are there more reasons?

lahabana:
> I just had a thought/question which is maybe just for my understanding, but how come something like SPIFFE IDs aren't an option here? I think that even from my level of understanding, I guess the /'s etc in SPIFFE URIs are problematic for all the reasons why we've landed on a _? Are there more reasons?

Spiffe Id only exist for Dataplane identity.

docs/madr/decisions/070-resource-identifier.md
lobkovilya: There is merging in routes, but I think we're still good and we can name routes in envoy by the name of the most specific MeshHTTPRoute (I guess you were referring to it as "child route"?).

For example, there are producer and consumer routes:
```yaml
kind: MeshHTTPRoute
metadata:
  name: producer-route
  namespace: backend-ns
spec:
  to:
    - targetRef:
        kind: MeshService
        name: backend
      rules:
        - matches:
            - path:
                type: PathPrefix
                value: "/api"
          default: $conf1
        - matches:
            - path:
                type: PathPrefix
                value: "/gui"
          default: $conf2
---
kind: MeshHTTPRoute
metadata:
  name: consumer-route
  namespace: frontend-ns
spec:
  to:
    - targetRef:
        kind: MeshService
        name: backend
        namespace: backend-ns
      rules:
        - matches:
            - path:
                type: PathPrefix
                value: "/api"
          default: $conf3
        - matches:
            - path:
                type: PathPrefix
                value: "/path-of-consumers-interest"
          default: $conf4
```

In consumer's Envoy they're going to result in:

```yaml
routes:
  - name: kri_default_us-east-2_frontend-ns_consumer-route
    match:
      path: "/api"
    route: merge([$conf1, $conf3]).to_envoy_route()
  - name: kri_default_us-east-2_frontend-ns_consumer-route
    match:
      path: "/path-of-consumers-interest"
    route: merge([$conf4]).to_envoy_route()
  - name: kri_default_us-west-2_backend-ns_producer-route
    match:
      path: "/gui"
    route: merge([$conf2]).to_envoy_route()
```

docs/madr/decisions/070-resource-identifier.md
lobkovilya: > I kinda agree with is, if the resource-type is before anything else, then you can immediately tell whether to expect say an empty mesh in the case of the kri pointing to a resource type that doesn't have a mesh

Yeah, probably you're right, it's better to have the first parameter that can't be empty, i.e. `kri_zoneingress__us-east-1_kuma-system_zi-1`. I'm going to change the order then to
```
kri_<resource-type>_<mesh>_<zone>_<namespace>_<resource-name>_<section-name>
```

docs/madr/decisions/070-resource-identifier.md
lobkovilya: > In the past there were stuff in the metric name, I remember we even had some code to change things around. My is that this is the main reason no?

Yeah maybe, but then it's still unclear, why don't they replace `:` in `stat_prefix`? IMO it's just odd that `stat_prefix` and `alt_stat_name` don't have consistent behavior.

> Also when not using prom format doesn't it actually use : as a separator see:

Even though envoy uses `:` as a separator in `/clusters` endpoint, it still takes `cluster.name`, not `cluster.alt_stat_name`:

image

The cluster in the example named `kuma:readiness`

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Ah no, I don't question why there is an actual name and stats name, I just don't understand why `stat_prefix` behaves differently not the same way as `alt_stat_name`.

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Found it in Kubernetes API reference for [ContainerPort](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#containerport-v1-core):
> `name` – If specified, this must be an IANA_SVC_NAME and unique within the pod. Each named port in a pod must have a unique name. Name for the port that can be referred to by services.

And IANA_SVC_NAME is specified in RFC:

> 5.1.  Service Name Syntax
> 
>    Valid service names are hereby normatively defined as follows:
> 
>    o  MUST be at least 1 character and no more than 15 characters long
> 
>    o  MUST contain only US-ASCII [ANSI.X3.4-1986] letters 'A' - 'Z' and
>       'a' - 'z', digits '0' - '9', and hyphens ('-', ASCII 0x2D or
>       decimal 45)
> 
>    o  MUST contain at least one letter ('A' - 'Z' or 'a' - 'z')
> 
>    o  MUST NOT begin or end with a hyphen
> 
>    o  hyphens MUST NOT be adjacent to other hyphens
> 
>    The reason for requiring at least one letter is to avoid service
>    names like "23" (could be confused with a numeric port) or "6000-
>    6063" (could be confused with a numeric port range).  Although
>    service names may contain both upper-case and lower-case letters,
>    case is ignored for comparison purposes, so both "http" and "HTTP"
>    denote the same service.

I checked, and it's indeed validated by the Kubernetes. I'll open an issue so we also validate this on our side

johncowen: > o MUST contain at least one letter ('A' - 'Z' or 'a' - 'z')

> upd: interesting that it doesn't allow names like "8080", at least one letter is required. But we don't have to make sectionName as strict, especially given the fact port's name on Service has a [different requirement](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#serviceport-v1-core):

wow! yes very interesting! Nice spot

> Optional if only one ServicePort is defined on this service.

I think this ^ is interesting also, i.e. required if more than one

docs/madr/decisions/070-resource-identifier.md
jijiechen: I don't remember in which PR that I removed this compatibility. I've just tested on 2.9.3 and confirmed that we still have this validation: 

When I was trying to create a `Mesh` object with name `default_2`, I got this error:

```
Error: Could not process a resource (Resource is not valid)
* name: invalid characters. A lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
```

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+### Option 2 - Field names in the identifier
+
+We need to pick two delimiters: one to separate keys from values and another to separate key-value pairs.
+
+```yaml
+kri;meshservice;mesh=mesh-1;zone=zone-1;namespace=kuma-demo;name=backend;section=http-port
+```
+
+**Pros:**
+- Better handling of gaps; no need for `;;` when a value is not defined
```
jijiechen, line 359: This format is definitely more readable by human.

jijiechen (dismissed):

lahabana:
Should we cover this https://github.com/kumahq/kuma/issues/12147 in the MADR?

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Updated

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Changed the order

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Added in the MADR

docs/madr/decisions/070-resource-identifier.md
lobkovilya: It turned out to be less straightforward with Secrets. Only `mesh_ca` in `builtin` mode has a real secret resource in the store we could use for `kri`. In all other cases, secrets are either generated in memory (i.e. `identity_cert`), or obtained using DataSource (i.e. CA provided), or fetched from external places (i.e. Vault). As we don't have problems with secret naming at that moment and also naming requirements are not yet clear for secrets, I'd move them out of the scope of this MADR. I've added a section:

> #### Secrets naming in Envoy
>
> Renaming secrets in Envoy is out of the scope of this MADR. Only `mesh_ca` secret in `builtin` mode has a corresponding Secret resource in the store. In all other cases, secrets are either stored in memory, or obtained using DataSource or fetched from external systems, so it's not straightforward how to apply `kri` concept. Also, it's not clear what problems and limitations do we have when it comes to naming secrets in Envoy.

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Not much we can do right now as we generate a unique cluster per MeshGateway listener. I looked a bit into this and I think we should be able to stop doing that. I've created an issue https://github.com/kumahq/kuma/issues/13129 and referenced it in the MADR. Once we get rid of hash-suffixes in MeshGateway cluster names we can go with `kri` from Mesh*Service.

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,363 @@
...
+#### Envoy resource names
+
+For the context, there is already [MADR](036-internal-listeners.md) that regulated the name of internal listeners.
+Also, there was [work](https://docs.google.com/document/d/1OIZK82Tr-4El2FfdlBn7WNRZ7FatkTuEcZKH0FlSTMA/edit?tab=t.0#heading=h.n6cmlf1eel2z) related to Envoy cluster name unification, but it's not finished.
+Discoveries in this work helped me to fill the tables.
+
+There are no restriction on the name format from the Envoy's side.
+Some Envoy resources could be directly correlated with Kuma resources and that's why we should consider renaming them using the resource identifier.
+There are tables below for all Kuma proxies. Resources in `Internal` table exist in all proxies. 
+Column "Correlated Resources" provides the Kuma resources that could be used for naming.
```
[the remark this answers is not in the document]
lobkovilya: Added a new column `ResourceIdentifier`

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Updated MADR with newly discovered info on ServicePort's name

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Yes, added a note on that

> As the goal of this MADR is to have unified names in Envoy, Metrics, and API, we want `stat_prefix/alt_stat_name` to always be the same as `name`. This makes the use of `stat_prefix/alt_stat_name` unnecessary and once the migration to new naming is completed Kuma won't set these fields anymore.

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+
+Renaming secrets in Envoy is out of the scope of this MADR. 
+Only `mesh_ca` secret in `builtin` mode has a corresponding Secret resource in the store.
+In all other cases, secrets are either stored in memory, or obtained using DataSource or fetched from external systems,
+so it's not straightforward how to apply `kri` concept.  
+Also, it's not clear what problems and limitations do we have when it comes to naming secrets in Envoy.
+
+**Pros:**
+- Shorter
+- Resembles existing formats from Amazon
```
lahabana, line 359: and Konnect

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+## Pros and Cons of the Options
+
+### Option 1 - Order-based, no field names in the identifier
+
+We need to pick a delimiter that's present in the `delimiter` charset we've specified above.
+
+There is an identifier format from Amazon called [ARN](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html). We can adopt a similar approach, but using `_`:
+
+```
+kri_<resource-type>_<mesh>_<zone>_<namespace>_<resource-name>_<section-name>
```
lahabana, line 331: Could we have an optional `suffix` ? I already envision use cases where actually we can't sumup with a single kri even with a sectionName and we still need to append some `hash` at the end?

I'd argue strongly against the `hashes` if it came up but I'm sure it's gonna come up.

Another probably better option is to explicitly state that if needs for a suffix was required we will need to create a MADR. This enable us to leave this door open but clearly say it should never be required. From looking at the discussion in: https://github.com/kumahq/kuma/issues/13129 clearly it seems that it'd be necessary

lahabana (commented):
Please add at the beginning this what's out of scope.

Because here you talk about a lot of things because we plan to use this in a lot of places but we don't clarify that the only thing we do here is to define the format.
We don't define how it's going to be rolled out to the Envoy config or what the http api looks like

docs/madr/decisions/070-resource-identifier.md
lobkovilya: This reminded me we should probably treat `sectionName` as an optional suffix, because when it's missing, we don't want to have a trailing `_`, like 
```
kri_meshservice_mesh-1_us-east-1_kuma-demo_mt1_
```

So maybe we can put `sectionName` after some delimiter, because strictly speaking `sectionName` is not a resource identifier, it's a reference inside the specific resource:
```
kri_<resource-type>_<mesh>_<zone>_<namespace>_<resource-name>&sectionName=httpport
OR
kri_<resource-type>_<mesh>_<zone>_<namespace>_<resource-name>&sn=httpport
```

and then like you said if we need another suffix, we will create MADR to add another KV pair behind `&`, i.e. `hash` 
```
kri_<resource-type>_<mesh>_<zone>_<namespace>_<resource-name>&sn=httpport,hash=e3a0069bda479bd1
```

WDYT?

lahabana: I'd rather have section be the default with a trailing `_` the rest can wait for a future MADR

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+
+#### [Issue #12093](https://github.com/kumahq/kuma/issues/12093): xds configs, outbound listeners should use the clustername instead of an IP/port combo
+
+We name outbounds like `outbound:10.43.205.116:6379` where IP address doesn't give any useful information.
+
+### Places to use resource identifier
+
+#### URL path
+
+For example `:5681/path-prefix/<identifier>/path-suffix`.
```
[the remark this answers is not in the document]
lobkovilya: Yeah, I'll add that it's out of scope, here it's just an example for clarity. Why isn't `path-prefix` compliant with our specs?

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+* `name` and `zone`
+  * contain no more than 253 characters
+  * contain only lowercase alphanumeric characters, '-' or '.'
+  * start with an alphanumeric character
+  * end with an alphanumeric character
+* `mesh`
+  * contain no more than 63 characters
+  * contain only lowercase alphanumeric characters, '-' or '.'
+  * start with an alphanumeric character
+  * end with an alphanumeric character
```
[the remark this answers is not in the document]
lobkovilya: Good catch, fixed

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+
+
+##### Zone Ingress
+
+|          | Name                                                                                                                        | Correlated Resources                           | ResourceIdentifier                                         |
+|----------|-----------------------------------------------------------------------------------------------------------------------------|------------------------------------------------|------------------------------------------------------------|
+| Listener | `inbound:10.43.205.116:10001`<br>`inbound:[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:10001`                                  | ZoneIngress                                    | kri_zoneingress__us-east-2_kuma-system_zi1                 |
+| Cluster  | legacy services - `<mesh>:<kuma.io/service>`<br>new Mesh*Services - `<mesh>_<name>_<namespace>_<zone>_<short-name>_<port>`  | Mesh*Service (with sectionName to select port) | kri_meshservice_mesh-1_us-east-2_kuma-demo_backend_httppor |
```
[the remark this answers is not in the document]
lobkovilya: thanks, fixed

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+```
+envoy_cluster_upstream_cx_total{envoy_cluster_name="localhost_5000"} 0
+```
+so there is no reason to replace `:` with `_`.
+
+If Envoy tried to support StatsD format then there are still a question, why `stat_prefix` emits metrics with `:` but `alt_stat_name` replaces `:` with `_`.
+Also, StatsD format doesn't seem to be supported out-of-the-box by Envoy and requires configuring `stats_sinks` in the static config.
+
+As the goal of this MADR is to have unified names in Envoy, Metrics, and API, we want `stat_prefix/alt_stat_name` to always be the same as `name`.
+This makes the use of `stat_prefix/alt_stat_name` unnecessary and once the migration to new naming is completed Kuma won't set these fields anymore.
```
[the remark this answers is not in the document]
lobkovilya: I changed so now there are only 2 paragraphs in this section, one that covers the current state and the second explains why we won't use `stat_prefix` anymore. I removed the middle part with the Envoy discrepancy between handling `stat_prefix` and `alt_stat_name` as it's irrelevant to the MADR

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+| Listener    | `inbound:10.43.205.116:10001`<br>`inbound:[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:10001`                                                                      | ZoneEgress                                               | kri_zoneegress__us-east-2_kuma-system_ze1                   |
+| Cluster     | legacy services - `<mesh>:<kuma.io/service>`<br>new Mesh*Services - `<mesh>_<name>_<namespace>_<zone>_<short-name>_<port>`                                      | Mesh*Service (with sectionName to select port)           | kri_meshservice_mesh-1_us-east-2_kuma-demo_backend_httpport |
+| FilterChain | for external services - `<kuma.io/service>_<mesh>`                                                                                                              | MeshExternalService                                      | kri_meshexternalservice_mesh-1___es1                        |
+| Route       | for external services - `outbound:<kuma.io/service>`                                                                                                            | MeshExternalService                                      | kri_meshexternalservice_mesh-1___es1                        |
```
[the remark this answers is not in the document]
lobkovilya: Yes, thanks for noticing

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+We need to pick a delimiter that's present in the `delimiter` charset we've specified above.
+
+There is an identifier format from Amazon called [ARN](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html). We can adopt a similar approach, but using `_`:
+
+```
+kri_<resource-type>_<mesh>_<zone>_<namespace>_<resource-name>_<section-name>
+```
+
+`resource-type` is a lowercased singular resource type, i.e. `meshservice`, `meshtimeout`, etc. 
+We use this name in [kumactl](https://github.com/kumahq/kuma/blob/d7ec0a2b1ac19208fb7dd9726309e3cf8cdc5848/pkg/core/resources/apis/meshservice/api/v1alpha1/zz_generated.resource.go#L153).
```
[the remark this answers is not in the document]
lobkovilya: Changed `<resource-type>` to `shortName`

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+| Cluster     | `meshpassthrough_<protocol>_<match-value>_<port>`<br>when `<port> == 0` Kuma sets port equal to `*`<br>`match-value = CIDR \| IP \| Domain`<br>`CIDR     = i.e. "192.0.2.0/24" or "2001:db8::/32"`<br>`IP       = i.e. "192.0.2.1", or 2001:db8::68", or ::ffff:192.0.2.1"`<br>`Domain   = <dns-name> \| *.<dns-name>`<br>`dns-name = ^([a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62}){1}(\.[a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62})*[\._]?$`<br>  | –                    |
+| FilterChain | when `<match-value> == <protocol>` -> `meshpassthrough_<protocol>_<port>`<br>when `<match-value> != <protocol>` -> `meshpassthrough_<protocol>_<match-value>_<port>`<br>when port is 0 we put `*`, i.e. `meshpassthrough_http_*`<br>`<match-value>` is defined above                                                                                                                                                               | –                    |
```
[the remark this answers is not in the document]
lobkovilya: We're not changing the way we name clusters originating from MeshPassthrough. Even though we could come up with a name like `kri_meshpassthrough...` + `indexOf(spec.appendMatch)`, I discussed with @lukidzi it's important to keep `port` and `protocol` for the metrics

docs/madr/decisions/070-resource-identifier.md
lobkovilya: Added a note:

> #### ResourceIdentifier format is open for extension in the future MADRs
>
> Today when generating clusters for Builtin Gateway, we use the hash suffix to support a MeshService cluster per MeshGateway listener. If we want cluster name to follow `kri` format, we need to extend the format with a new MADR.

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,378 @@
...
+
+##### Internal
+
+|                  | Name                                                                                                                                                                 | Correlated Resources |
+|------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------|
+| Listener         | `kuma:dns`<br>`kuma:envoy:admin`<br>`kuma:metrics:prometheus`<br>`_kuma:metrics:opentelemetry:<backendName>`                                                         | –                    |
+| VirtualHost      | listener's name                                                                                                                                                      | –                    |
+| Internal Cluster | `kuma:readiness`<br>`kuma:envoy:admin`<br>`kuma:metrics:hijacker`<br>`_kuma:metrics:opentelemetry:<backendName>`<br>`tracing:`<br>`access_log_sink`<br>`ads_cluster` | –                    |
```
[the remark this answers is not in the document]
lobkovilya: I marked internal resources that'll be removed. But I'm not sure if we need to rename resources to kri format when they don't correlate with the store's real resources. For example, `kuma:dns` or `kuma:readiness` are just generated for each DPP and we know they won't collide with anything else. 

I can probably mention that we need to add a `_kuma` prefix

docs/madr/decisions/070-resource-identifier.md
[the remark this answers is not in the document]
lahabana: Cause parts of the path that are neither resource name nor types should be prefixed by `_`

lahabana (approved):
Ship it!

docs/madr/decisions/070-resource-identifier.md
```diff
@@ -0,0 +1,380 @@
...
+| Outbound Cluster  | legacy clusters - `<kuma.io/service>-hash(dst.tags)`<br>legacy clusters cross-mesh - `<kuma.io/service>-hash(dst.tags)_<mesh>`<br>new clusters - `<mesh>_<name>_<namespace>_<zone>_<short-name>_<port>` | Mesh*Service (with sectionName to select port)                                                                                           | kri_msvc_mesh-1_us-east-2_kuma-demo_backend_httpport |
+| Route             | Routes are set on Listener on VirtualHost.<br>On inbound - `inbound:<kuma.io/service>`<br>On outbound - `<hash_sha256([]Match{...})>`                                                                   | Envoy Route is a merge product of multiple MeshHTTPRoutes.<br/> We can use only the most specific MeshHTTPRoute policy to name the route | kri_mhttpr_mesh-1_us-east-2_kuma-demo_route-1_       |
```
lukidzi, line 164: What will be the name for default routes, since we don't need a policy to get ones?

lobkovilya: We probably should treat it as an internal entity, so it has to start with `_kuma`, I'll update the doc

lukidzi: How would it look like? Since, we might have a lot of default routes it's hard to say

lobkovilya: What do you mean by "a lot of default routes"? Don't we have just one [catchAll](https://github.com/kumahq/kuma/blob/820494cf59e22d5a4870ca6d7a01f1f77e38ad46/pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go#L196) route? It probably can be named something like `_kuma_catchAll` but I don't want to be too specific in this document about the way we name internal entities, it should be out of the scope, I left a note:
> Following the  [Internal listeners naming convention](036-internal-listeners.md) we need to rename internal listeners and clusters to start with `_`. Additionally, we need to get rid of `:` in internal entities names and use `_` instead.

lukidzi (approved):

johncowen (approved):
Seeing as I commented way back when this started, I had a final read and LGTM 👏

docs/madr/decisions/070-resource-identifier.md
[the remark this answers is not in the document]
lobkovilya: That's interesting idea, I'll add an issue so we could think more about it

lahabana (approved):

Issue #12123, opened 2024-11-27
When I have multiple services I get multiple inbounds with the same port


lahabana:
### What happened?

with:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: demo-app
  namespace: kuma-demo
  labels:
    app: demo-app
    version: v1
spec:
  selector:
    matchLabels:
      app: demo-app
      version: v1
  replicas: 1
  template:
    metadata:
      labels:
        app: demo-app
        version: v1
    spec:
      containers:
        - name: app
          image: ghcr.io/kumahq/kuma-counter-demo:latest
          env:
            - name: OTEL_SERVICE_NAME
              value: demo-app
            - name: OTEL_EXPORTER_OTLP_ENDPOINT
              value: http://opentelemetry-collector.mesh-observability:4317
            - name: KV_URL
              value: "http://kv.kuma-demo.svc.cluster.local:5050"
            - name: APP_VERSION
              valueFrom:
                fieldRef:
                  fieldPath: metadata.labels['version']
          ports:
            - name: http
              containerPort: 5050
---
apiVersion: v1
kind: Service
metadata:
  name: demo-app
  namespace: kuma-demo
spec:
  selector:
    app: demo-app
  ports:
  - protocol: TCP
    appProtocol: http
    port: 5050
    targetPort: 5050
---
apiVersion: v1
kind: Service
metadata:
  name: demo-app-v1
  namespace: kuma-demo
spec:
  selector:
    app: demo-app
    version: v1
  ports:
    - protocol: TCP
      appProtocol: http
      port: 5050
      targetPort: 5050
```

I get:

```yaml
type: Dataplane
mesh: default
name: demo-app-6c79d867f8-wlbww.kuma-demo
labels:
  k8s.kuma.io/namespace: kuma-demo
  kuma.io/display-name: demo-app-6c79d867f8-wlbww
  kuma.io/env: kubernetes
  kuma.io/mesh: default
  kuma.io/origin: zone
  kuma.io/zone: default
networking:
  address: 10.42.0.17
  inbound:
    - port: 5050
      tags:
        app: demo-app
        k8s.kuma.io/namespace: kuma-demo
        k8s.kuma.io/service-name: demo-app
        k8s.kuma.io/service-port: '5050'
        kubernetes.io/hostname: k3d-k3s-default-server-0
        kuma.io/protocol: http
        kuma.io/service: demo-app_kuma-demo_svc_5050
        kuma.io/zone: default
[...]

PR #12760, opened 2025-02-05
feat(kuma-cp): deduplicate dataplane inbounds by address and port combination

Automaat:
## Motivation

We now can have multiple inbounds if multiple services select the same address and port, but we will generate Envoy resources only for one of these. We should deduplicate them on creation so that we don't have inbounds that don't correlate with Envoy resources. This can be also useful for GUI.

Inbound deduplication is enabled only when MeshService mode Exclusive is enabled

xref https://github.com/kumahq/kuma/pull/13108

Fix: https://github.com/kumahq/kuma/issues/12123


pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -158,9 +162,65 @@ func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string
...
+	if len(ifaces) == 0 {
+		name, _, err := i.NameExtractor.Name(ctx, pod)
+		if err != nil {
+			return nil, err
+		}
+
+		ifaces = append(ifaces, inboundForServiceless(zone, pod, name, nodeLabels))
+	}
+
+	// Right now we will build multiple inbounds for each service selecting port, but later on
```
jijiechen, line 197: The code will be left here for future usage, so readers are from the future. Should it be changed to "We are only create one listener for last inbound now, but in the previous version, LegacyInboundInterfacesFor, we built multiple inbounds for each service selecting port"

jijiechen (dismissed):
Looks good to me.

Automaat:
There are still some failing gateway api because of this change that need to be investigated more

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -317,14 +318,27 @@ func (r *PodReconciler) createOrUpdateDataplane(
...
 	others []*mesh_k8s.Dataplane,
 ) error {
+	meshName := util_k8s.MeshOfByLabelOrAnnotation(r.Log, pod, ns)
+	k8sMesh := v1alpha1.Mesh{}
+	if err := r.Client.Get(ctx, kube_types.NamespacedName{Name: meshName}, &k8sMesh); err != nil {
+		return err
+	}
+	mesh := core_mesh.NewMeshResource()
+	if k8sMesh.Spec != nil {
+		r.Log.Info("GOT MESH", "mesh", string(k8sMesh.Spec.Raw))
```
jijiechen, line 328: this wording needs to be improved here: it's not meaningful enough.

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -166,6 +168,7 @@ func (i *InboundConverter) LegacyInboundInterfacesFor(ctx context.Context, zone
 	return ifaces, nil
 }
 
+// Should be used when MeshService mode is Exclusive
 func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string, pod *kube_core.Pod, services []*kube_core.Service) ([]*mesh_proto.Dataplane_Networking_Inbound, error) {
```
jijiechen, line 172: Do we need to sort `services []*kube_core.Service` before iterate over it? To keep consistant results when deduplicated in multiple runs.

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -158,9 +164,66 @@ func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string
...
+	}
+
+	// Right now we will build multiple inbounds for each service selecting port, but later on
+	// we will only create one listener for last inbound. Ignoring other inbounds from dataplane.
+	// Because of this we can safely deduplicate them here. This needs to change when we get rid of kuma.io/service
+	return deduplicateInboundsByAddressAndPort(ifaces), nil
+}
+
+func deduplicateInboundsByAddressAndPort(ifaces []*mesh_proto.Dataplane_Networking_Inbound) []*mesh_proto.Dataplane_Networking_Inbound {
+	inboundKey := func(iface *mesh_proto.Dataplane_Networking_Inbound) string {
```
jijiechen, line 207: In one of my testing, I found when the service `port` is different than the `targetPort`, the two inbounds generated on the DP has the same name and port, but they have different `k8s.kuma.io/service-port` tags: one of them is incorrect. So in this case, we need to remove the one with incorrect tag. Could you also verify this?

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
Automaat: we sort inbounds at the end

pkg/plugins/runtime/k8s/controllers/pod_controller.go
Automaat: ah, this should not be here, I forgot to remove it after debugging

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
Automaat: do you mean the situation when we have service with two ports? 
With this approach we don't care about tags, inbound tags will be gone after we fully switch to MeshService exclusive. I think in this situation we will just create inbound from pod port

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -158,9 +164,66 @@ func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string
...
+	for _, iface := range ifaces {
+		if inboundsPerName[inboundKey(iface)] == nil {
+			inboundsPerName[inboundKey(iface)] = iface
+		}
+	}
+	var deduplicatedInbounds []*mesh_proto.Dataplane_Networking_Inbound
+	for _, v := range inboundsPerName {
+		deduplicatedInbounds = append(deduplicatedInbounds, v)
+	}
+	sort.Slice(deduplicatedInbounds, func(i, j int) bool {
```
lobkovilya, line 221: Probably should be moved out from `deduplicateInbounds` method, "deduplicate" usually implies:
* the complexity is O(N)
* the order of the input slice is preserved in the output slice

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -158,9 +164,66 @@ func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string
...
+	// Because of this we can safely deduplicate them here. This needs to change when we get rid of kuma.io/service
+	return deduplicateInboundsByAddressAndPort(ifaces), nil
+}
+
+func deduplicateInboundsByAddressAndPort(ifaces []*mesh_proto.Dataplane_Networking_Inbound) []*mesh_proto.Dataplane_Networking_Inbound {
+	inboundKey := func(iface *mesh_proto.Dataplane_Networking_Inbound) string {
+		return fmt.Sprintf("%s:%d", iface.Address, iface.Port)
+	}
+
+	inboundsPerName := map[string]*mesh_proto.Dataplane_Networking_Inbound{}
```
lobkovilya, line 211: nit: `inboundsPerAddressPort`?

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -158,9 +164,66 @@ func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string
...
+	for _, iface := range ifaces {
+		if inboundsPerName[inboundKey(iface)] == nil {
+			inboundsPerName[inboundKey(iface)] = iface
```
lobkovilya, line 214: append to the result slice here, that way you preserve the original order and you won't need another loop on `inboundsPerName`

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -131,7 +132,12 @@ func inboundForServiceless(zone string, pod *kube_core.Pod, name string, nodeLab
 	}
 }
 
-func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string, pod *kube_core.Pod, services []*kube_core.Service) ([]*mesh_proto.Dataplane_Networking_Inbound, error) {
+// Deprecated: LegacyInboundInterfacesFor is currently only used for delegated gateway and Mesh without MeshService exclusive
+// to not change order of inbounds.
+// For gateway we pick first inbound to take tags from. Delegated gateway identity relies on this.
+// For Dataplanes when MeshService is disabled we base identity and routing
+// TODO: We should revisit this when we rework identity
```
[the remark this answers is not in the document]
Automaat: added

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -158,9 +164,66 @@ func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string
 
 		ifaces = append(ifaces, inboundForServiceless(zone, pod, name, nodeLabels))
 	}
+
 	return ifaces, nil
 }
 
+// Should be used when MeshService mode is Exclusive
+func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string, pod *kube_core.Pod, services []*kube_core.Service) ([]*mesh_proto.Dataplane_Networking_Inbound, error) {
```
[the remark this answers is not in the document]
Automaat: good idea, chaged this

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
Automaat: changed

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -131,7 +131,29 @@ func inboundForServiceless(zone string, pod *kube_core.Pod, name string, nodeLab
...
+}
+
+// InboundInterfacesFor should be used when MeshService mode is Exclusive
 func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string, pod *kube_core.Pod, services []*kube_core.Service) ([]*mesh_proto.Dataplane_Networking_Inbound, error) {
+	inbounds, err := i.inboundInterfacesFor(ctx, zone, pod, services)
+	if err != nil {
+		return inbounds, err
+	}
+
+	// Right now we will build multiple inbounds for each service selecting port, but later on
```
lobkovilya, line 150: I'm a bit lost with "right now" and "later on", by "later" you mean like "later when Envoy receives the config"? Or "later" when we develop some other feature?

pkg/plugins/runtime/k8s/controllers/pod_converter.go
```diff
@@ -270,9 +272,20 @@ func (p *PodConverter) dataplaneFor(
 			return nil, errors.Errorf("invalid delegated gateway type '%s'", gwType)
 		}
 	} else {
-		ifaces, err := p.InboundConverter.InboundInterfacesFor(ctx, p.Zone, pod, services)
-		if err != nil {
-			return nil, err
+		var ifaces []*mesh_proto.Dataplane_Networking_Inbound
+		var err error
+		// deduplicate inbounds only when MeshService mode is exclusive. Without MeshServices we need duplicated inbounds
+		// for identity and traffic routing
```
lobkovilya, line 278: how identity is affected? it's still based on `kuma.io/service` even with MeshServices no?

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
jijiechen: I agree. The comment will be persist here for long, so we should serve a future reader's perspective.

pkg/plugins/runtime/k8s/controllers/pod_converter.go
Automaat: We base identity on tags, and we are basically removing identities by deduplicating inbounds and removing inbound and its tag kuma.io/service

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
Automaat: Removed the whole comment as it was outdated. I've updated comment for the whole method

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -161,6 +181,23 @@ func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string
 	return ifaces, nil
 }
 
+func deduplicateInboundsByAddressAndPort(ifaces []*mesh_proto.Dataplane_Networking_Inbound) []*mesh_proto.Dataplane_Networking_Inbound {
+	inboundKey := func(iface *mesh_proto.Dataplane_Networking_Inbound) string {
+		return fmt.Sprintf("%s:%d", iface.Address, iface.Port)
```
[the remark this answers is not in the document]
lukidzi: My bad, we use it only as a key

lukidzi (approved):

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -131,7 +131,27 @@ func inboundForServiceless(zone string, pod *kube_core.Pod, name string, nodeLab
...
+func (i *InboundConverter) LegacyInboundInterfacesFor(ctx context.Context, zone string, pod *kube_core.Pod, services []*kube_core.Service) ([]*mesh_proto.Dataplane_Networking_Inbound, error) {
+	return i.inboundInterfacesFor(ctx, zone, pod, services)
+}
+
+// InboundInterfacesFor should be used when MeshService mode is Exclusive. This function deduplicates inbounds by address and port.
+// Since MeshService does not need tags we can safely deduplicate inbounds
 func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string, pod *kube_core.Pod, services []*kube_core.Service) ([]*mesh_proto.Dataplane_Networking_Inbound, error) {
+	inbounds, err := i.inboundInterfacesFor(ctx, zone, pod, services)
+	if err != nil {
+		return inbounds, err
```
Icarus9913, line 148:
```suggestion
		return nil, err
```

Icarus9913 (dismissed):
@Automaat  Please supplement the description to mention that this deduplicate operation only exists in MeshService Exclusive mode

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
Automaat: changed

Icarus9913 (approved):

PR #12766, opened 2025-02-05
test(e2e): add envoyconfig tests for Builtin Gateway with MeshTimeout

lobkovilya:
## Motivation

Same kind of tests we recently introduced for sidecars are now available for Builtin Gateway!


test/e2e_env/universal/envoyconfig/builtingateway.go
```diff
@@ -0,0 +1,157 @@
...
+`, mesh)
+	}
+
+	BeforeAll(func() {
+		setup := NewClusterSetup().
+			Install(
+				Yaml(
+					builders.Mesh().
+						WithName(mesh).
+						WithoutInitialPolicies().
```
Automaat, line 86: don't we need mtls on mesh, like in sidecar tests, for testing MesTLS for example

test/e2e_env/universal/envoyconfig/builtingateway.go
```diff
@@ -0,0 +1,157 @@
...
+
+	E2EAfterEach(func() {
+		Expect(DeleteMeshResources(
+			universal.Cluster,
+			mesh,
+			meshtimeout.MeshTimeoutResourceTypeDescriptor,
+		)).To(Succeed())
+	})
+
+	getConfig := func(dpp string) string {
```
Automaat, line 120: isn't this function the same as in the sidecar test? Maybe we can reuse it

test/e2e_env/universal/envoyconfig/builtingateway.go
lobkovilya: yes, added mtls

Automaat (approved):

PR #12767, opened 2025-02-05
feat(oapi): parse comments to set defaults on pb resources

slonka:
## Motivation

TF distinguishes between null and zero type value so without this we get:

```
      ~ routing                        = {
          + default_forbid_mesh_external_service_access = false
          + zone_egress                                 = false
            # (1 unchanged attribute hidden)
        }
```

## Implementation information

I extract default value from "Default: " in comments and put it in "default" field.

## Supporting documentation

xrel https://github.com/Kong/kong-mesh/issues/7201

### DIFF
--- a/api/mesh/v1alpha1/mesh.proto
+++ b/api/mesh/v1alpha1/mesh.proto
@@ -32,6 +32,7 @@ message Mesh {
     repeated CertificateAuthorityBackend backends = 2;
 
     // If enabled, skips CA validation.
+    // Default: false
     bool skipValidation = 3;
   }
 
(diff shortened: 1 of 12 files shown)


api/mesh/v1alpha1/mesh.proto
```diff
@@ -227,6 +229,7 @@ message DatadogTracingBackendConfig {
   // `backend` service that communicates with a couple of databases, you would
   // get service names like `backend_INBOUND`, `backend_OUTBOUND_db1`, and
   // `backend_OUTBOUND_db2` in Datadog. Default: false
```
lahabana, line 231:
```suggestion
  // `backend_OUTBOUND_db2` in Datadog.
```

slonka:
This no longer be needed.

PR #12768, opened 2025-02-06
fix(kds): don't delete zone created Secret

lukidzi:
## Supporting documentation

Fix #XX


pkg/kds/v2/client/kds_client.go
```diff
@@ -112,7 +113,19 @@ func (s *kdsSyncClient) Receive() error {
...
+				if err := s.kdsStream.NACK(received.Type, err); err != nil {
+					if err == io.EOF {
+						return nil
+					}
+					return errors.Wrap(err, "failed to NACK a discovery response")
+				}
+				s.log.V(1).Info("NACK", "response", received)
+				continue
+			} else {
+				return errors.Wrapf(err, "failed to store %s resources", received.Type)
```
lahabana, line 127: should we include the resource name too?

lahabana:
2 points:

- needs test
- Can we consider backporting this?

api/system/v1alpha1/secret.proto
```diff
@@ -15,6 +15,7 @@ message Secret {
   option (kuma.mesh.resource).package = "system";
   option (kuma.mesh.resource).skip_kubernetes_wrappers = true;
   option (kuma.mesh.resource).kds.send_to_zone = true;
+  option (kuma.mesh.resource).kds.send_to_global = false;
```
Icarus9913, line 18: Does this supplement make impact? I don't see any place call it.
Or is it just a standard supplement to the structure?

pkg/kds/v2/client/kds_client.go
```diff
@@ -112,7 +113,19 @@ func (s *kdsSyncClient) Receive() error {
...
+			}
```
Icarus9913, line 128: Should we filter this `ResourceAlreadyExists` error when creating and then downgrade the error? I feel confused if we send NACK, will we do this create again in the next streaming?

lukidzi: I've changed the logic

api/system/v1alpha1/secret.proto
lukidzi: It makes no impact since the default is `false`. I wanted to make it explicit

lukidzi:
> * needs test
Added both E2E and unit
> * Can we consider backporting this?
I think it's possible to backport this and https://github.com/kumahq/kuma/pull/12776. I am a bit worried since it changes logic a bit.

Icarus9913:
> Do we need kuma.io/origin validation?

In my opinion: no

The cases we may hit:
1. create secret in zone (restart ZCP)
2. create secret in global and it would sync to zone (restart GCP, ZCP)
3. create secret in zone and create the same name secret in global (restart ZCP, GCP)

Since we have used `predicate` to filter the zone secret, it makes no impact for the `kuma.io/origin: global` global synced secret right? Do we have other edge cases?

pkg/kds/v2/store/sync.go
```diff
@@ -41,12 +42,13 @@ type ResourceSyncer interface {
 	//
 	// Sync takes into account only 'Name' and 'Mesh' when it comes to upstream's Meta.
 	// 'Version', 'CreationTime' and 'ModificationTime' are managed by downstream store.
-	Sync(ctx context.Context, upstream client_v2.UpstreamResponse, fs ...SyncOptionFunc) error
+	Sync(ctx context.Context, upstream client_v2.UpstreamResponse, fs ...SyncOptionFunc) (error, error)
```
Icarus9913, line 45: Can we add comments for these 2 errors since this is a public method

lukidzi:
> The cases we may hit:
>
>1. create secret in zone (restart ZCP)
I didn't add this specific e2e tests since creating a new secret triggers reconciliation + restarting CP might be flaky in the e2e
>2. create secret in global and it would sync to zone (restart GCP, ZCP)
We should manually test
>3. create secret in zone and create the same name secret in global (restart ZCP, GCP)
We should manually test

>Since we have used predicate to filter the zone secret, it makes no impact for the kuma.io/origin: global global synced >secret right? Do we have other edge cases?

I think it should all works and don't affect.

test/e2e_env/multizone/sync/sync.go
```diff
@@ -187,6 +213,84 @@ spec:
...
+
+			Expect(multizone.Global.Install(YamlUniversal(globalSecret))).To(Succeed())
+			// should not override existing in zone
+			Eventually(func(g Gomega) {
+				out, err := multizone.Global.GetKumactlOptions().RunKumactlAndGetOutput("get", "secrets", "--mesh", meshName, "-o", "yaml")
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(strings.Count(out, secretName)).To(Equal(1))
+				g.Expect(strings.Count(out, "Z2xvYmFsCg==")).To(Equal(1))
+			}, "30s", "1s").Should(Succeed())
+			Consistently(func(g Gomega) {
```
Icarus9913, line 260: Do we need to use `Consistently` here for Zone? That would lead to a flaky error once network latency. Maybe `Eventually` is enough right?

Icarus9913 (approved):

PR #12769, opened 2025-02-06
fix(meshtimeout): remove if-condition that prevents plugin from running when there are no policies

lobkovilya:
## Motivation

This if-condition at the beginning of the plugin
```go
if len(policies.ToRules.Rules) == 0 && len(policies.FromRules.Rules) == 0 && len(policies.GatewayRules.ToRules.ByListener) == 0 && len(policies.ToRules.ResourceRules) == 0 {
    return nil
}
```
causes inconsistent default values. For example, the default outbound timeouts should be the same regardless of the presence of MeshTimeout policies with `from`. 

## Implementation information

Depends on https://github.com/kumahq/kuma/pull/12766

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12649

### DIFF
--- a/pkg/api-server/testdata/resources/inspect/dataplanes/_rules/proxyconfig_shadow-false.golden.json
+++ b/pkg/api-server/testdata/resources/inspect/dataplanes/_rules/proxyconfig_shadow-false.golden.json
@@ -5,7 +5,7 @@
     "name": "localhost:8080",
     "altStatName": "localhost_8080",
     "type": "STATIC",
-    "connectTimeout": "10s",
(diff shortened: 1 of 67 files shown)


Automaat (approved):

Automaat:
Looking at this pr, it looks like we were missing default timeouts?

lobkovilya:
> Looking at this pr, it looks like we were missing default timeouts?

Most of the time they're the same as Envoy's defaults. But sometimes they're not and bad case is explained in the issue https://github.com/kumahq/kuma/issues/12649

PR #12776, opened 2025-02-07
fix(kds): add all valid resources and skip only not valid

lukidzi:
## Motivation

TBA

## Supporting documentation

Fix #XX

### DIFF
--- a/pkg/kds/v2/client/kds_client.go
+++ b/pkg/kds/v2/client/kds_client.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	std_errors "errors"
 	"io"
 	"time"
 
@@ -15,6 +16,7 @@ type UpstreamResponse struct {
 	ControlPlaneId      string
 	Type                model.ResourceType
 	AddedResources      model.ResourceList
+	InvalidResourcesKey []model.ResourceKey
 	RemovedResourcesKey []model.ResourceKey
 	IsInitialRequest    bool
 }
@@ -23,12 +25,14 @@ func (u *UpstreamResponse) Validate() error {
 	if u.AddedResources == nil {
 		return nil
 	}
+	var err error
 	for _, res := range u.AddedResources.GetItems() {
-		if err := model.Validate(res); err != nil {
-			return err
+		if validationErr := model.Validate(res); validationErr != nil {
+			err = std_errors.Join(err, validationErr)
+			u.InvalidResourcesKey = append(u.InvalidResourcesKey, core_model.MetaToResourceKey(res.GetMeta()))
 		}
 	}
-	return nil
+	return err
 }
 
 type Callbacks struct {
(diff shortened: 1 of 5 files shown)


pkg/kds/v2/client/kds_client.go
```diff
@@ -23,12 +25,14 @@ func (u *UpstreamResponse) Validate() error {
 	if u.AddedResources == nil {
 		return nil
 	}
+	var err error
 	for _, res := range u.AddedResources.GetItems() {
-		if err := model.Validate(res); err != nil {
-			return err
```
Icarus9913, line 28: We can just add 
```text
u.InvalidResourcesKey = append(u.InvalidResourcesKey, core_model.MetaToResourceKey(res.GetMeta()))
```
on top of `return err`

There's no need to introduce `errors.Join` here right?

pkg/kds/v2/client/kds_client.go
```diff
@@ -88,19 +92,19 @@ func (s *kdsSyncClient) Receive() error {
 			return errors.Wrap(err, "failed to receive a discovery response")
 		}
 		s.log.V(1).Info("DeltaDiscoveryResponse received", "response", received)
+		validationErrors := received.Validate()
```
Icarus9913, line 95: I guess we can put the `callbacks` before `received.Validate()` and add some comments to explain the edge case. Then we can avoid the following duplicated twice `NACK` codes

pkg/kds/v2/client/kds_client.go
lukidzi: It's good for NACK message so the user knows that there is an issue with the resource

pkg/kds/v2/client/kds_client.go
lukidzi: We can't if we return NACK after validation we are not going to store the resource that were valid which is an issue

Icarus9913: I mean like this:

```text
s.log.V(1).Info("DeltaDiscoveryResponse received", "response", received)

// comments
if s.callbacks != nil {
	err = s.callbacks.OnResourcesReceived(received)
	if err != nil {
		return errors.Wrapf(err, "failed to store %s resources", received.Type)
	}
}

validationErrors := received.Validate()
if validationErrors != nil {
	s.log.Info("received resource is invalid, sending NACK", "err", validationErrors)
	if err := s.kdsStream.NACK(received.Type, validationErrors); err != nil {
		if err == io.EOF {
			return nil
		}
		return errors.Wrap(err, "failed to NACK a discovery response")
	}
	continue
}
if !received.IsInitialRequest {
	// Execute backoff only on subsequent request.
	// When client first connects, the server sends empty DeltaDiscoveryResponse for every resource type.
	time.Sleep(s.responseBackoff)
}
s.log.V(1).Info("sending ACK", "type", received.Type)
if err := s.kdsStream.ACK(received.Type); err != nil {
	if err == io.EOF {
		return nil
	}
	return errors.Wrap(err, "failed to ACK a discovery response")
}
```

Icarus9913: oh, the `received.Validate()` would store the `InvalidResourcesKey`, I missed it.

Icarus9913 (approved):

pkg/kds/v2/store/sync.go
```diff
@@ -133,18 +134,25 @@ func (s *syncResourceStore) Sync(syncCtx context.Context, upstreamResponse clien
...
-		} else {
-			downstream = filtered
-		}
-		if filtered, err := filter(upstream, opts.Predicate); err != nil {
-			return err
-		} else {
-			upstream = filtered
-		}
+		predicate = opts.Predicate
+	}
```
lahabana, line 147: Another option could be to change filter to simply take a list of predicates. This might make this more readable

pkg/kds/v2/store/sync.go
```diff
@@ -278,6 +285,14 @@ func newIndexed(rs core_model.ResourceList) *indexed {
 	return &indexed{indexByResourceKey: idxByRk}
 }
 
+func indexOfResourceKeys(rks []core_model.ResourceKey) map[core_model.ResourceKey]struct{} {
```
lahabana, line 288: I'm surprised such a function doesn't exist already.

lahabana: https://github.com/kumahq/kuma/blob/d7ec0a2b1ac19208fb7dd9726309e3cf8cdc5848/pkg/core/resources/model/resource.go#L764-L771 is similar no?

lahabana: If there's no possibility to adapt this helper should probably move next to the method above

lahabana:
It's not super clear to me how values that fail validation end up not showing up in the `add`

lukidzi:
> It's not super clear to me how values that fail validation end up not showing up in the `add`

The filter works on the match, so if return true, means we add it. If `indexedInvalidResources` has a specific ResourceKey (which we add in `kds_client.go`) then it means it's invalid and we don;t add it to upstream and downstream lists of resources

Icarus9913 (approved):

PR #12782, opened 2025-02-07
fix(xds): fix token expiration issue caused by inappropriate cache in kuma-cp

jijiechen:
## Motivation

An inappropriate cache of the auth result may cause incorrect auth result between DP and CP.

## Implementation information

Only auth once per gRPC stream.

## Supporting documentation

Fix #XX

### DIFF
--- a/pkg/xds/auth/callbacks.go
+++ b/pkg/xds/auth/callbacks.go
@@ -59,7 +59,8 @@ type stream struct {
 	// Dataplane / ZoneIngress / ZoneEgress associated with this XDS stream.
 	resource model.Resource
 	// nodeID of the stream. Has to be the same for the whole life of a stream.
-	nodeID string
+	nodeID        string
+	authenticated bool
 }
 
 var _ util_xds.Callbacks = &authCallbacks{}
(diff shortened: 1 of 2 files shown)


pkg/xds/auth/callbacks.go
```diff
@@ -87,12 +88,16 @@ func (a *authCallbacks) OnStreamRequest(streamID core_xds.StreamID, req util_xds
 		return err
 	}
 
-	credential, err := ExtractCredential(s.ctx)
-	if err != nil {
-		return errors.Wrap(err, "could not extract credential from DiscoveryRequest")
-	}
-	if err := a.authenticator.Authenticate(user.Ctx(s.ctx, user.ControlPlane), s.resource, credential); err != nil {
-		return errors.Wrap(err, "authentication failed")
+	// If the gRPC stream is already authenticated, we don't need to authenticate it again.
```
jijiechen, line 91: We can't put it in `OnStreamOpen` because we don't have the `resource` object in there.

jijiechen:
fix defaulting behaviour of `EnableReloadableTokens` on CP config

lukidzi:
We should target `master` and backport to other branches

PR #12783, opened 2025-02-07
feat(kuma-cp): inbound name should be taken from pod instead of service

Automaat:
## Motivation

At this moment we are taking inbound name from service port name. In order to get rid of kuma.io/service we need to take name from pod port name. Since we are not using name at the moment, it is safe to switch this.


lobkovilya:
btw it's used [here](https://github.com/kumahq/kuma/blob/820494cf59e22d5a4870ca6d7a01f1f77e38ad46/pkg/xds/topology/outbound.go#L348), but I guess it won't be affected by the change?

lobkovilya (approved):
LGTM, please open an issue to document this behaviour, it looks like today we don't explain in the docs where ports on DPP are coming from

Issue #12785, opened 2025-02-07
Improve logging


lahabana:
### Kuma Version

master

### Describe the bug

We currently only look whether the token review returned that it's Authenticated:

https://github.com/kumahq/kuma/blob/4bda0eaf374136a40c39a014858f4b0b9bfd6bb4/pkg/xds/auth/k8s/authenticator.go#L74

There's an Error field in this status so we should really log this out if it existed and not return a "token doesn't belong to a valid user" but "token verification failed".
Be mindful as this error gets sent to the DP so let's not return the error content.

Issue #12784, opened 2025-02-07
Do not authenticate on each request


lahabana:
### Description

At the moment we authenticate on each request: https://github.com/kumahq/kuma/blob/c264fea580b52ea286e9ed4dc39b8e7909403cc8/pkg/xds/auth/callbacks.go#L84-L101

We can't do this authentication on StreamOpen because we don't have the resource in this case but could we just do it on the first request?

TLS only does the auth on handshake and we don't tear inflight connection when the cert expires. It feels like this should be acceptable for us too. Is that the case in Istio? In Linkerd?

If we judge that we should for reauth on expiry can we simply set a deadline on the token? Can we extract how long a token is valid when we auth first? In any case if we decide that aborting connection is not an option we should design things a bit more.

jijiechen:
xref: https://github.com/kumahq/kuma/pull/4685

jijiechen:
xref: https://github.com/kumahq/kuma/pull/12782

jijiechen:
Istio authenticates once per connection before is a connection is handled by the function `xds.Stream`: https://github.com/istio/istio/blob/1.24.2/pilot/pkg/xds/ads.go#L218-L238
Streams are handled in `pkg/xds/server.go`  and it does nothing than handling stream request/response.

jijiechen:
Linkerd does not authenticate DPs from inside their CP server (the `destination` service)
* Source: https://github.com/linkerd/linkerd2/blob/main/controller/api/destination/server.go#L131

They wrap their `destination` service with the `linkerd-init` container hence using the mTLS capability that is backed by the `identity` service to authenticate their DPs.
https://github.com/linkerd/linkerd2/blob/main/charts/linkerd-control-plane/templates/destination.yaml#L410-L433

The `identity` service validates the service account token at when it's called and then a mTLS certificate is issued:
* token validator: https://github.com/linkerd/linkerd2/blob/main/controller/identity/validator.go#L51
* mTLS CA: https://github.com/linkerd/linkerd2/blob/main/pkg/identity/service.go#L203

lukidzi:
> Triage: Let's do it

jijiechen:
We did not have a MADR on this, instead @jakubdyszkiewicz had clearly described the original considerations here:
https://github.com/kumahq/kuma/pull/4685#issue-1319503247

PR #12788, opened 2025-02-10
fix(xds): only auth once per xds gRPC stream in kuma-cp

jijiechen:
## Motivation

Right now, we are authenticating the DP on every `OnStreamRequest` event in `auth.Callback`, and [the authenticator has a cache](https://github.com/kumahq/kuma/blob/5359d941bf6a7fab22cb177cd94de1112b9b4ad7/pkg/xds/auth/k8s/authenticator.go#L48-L64) which caches the successful verification results. The Kubernetes API Server is only called when it is not hitting the cache, and the cache can last long: when the cache expires, the token is already expired from Kube's point of view.

In a cluster with bound service account token enabled and "extend-token-expiration" disabled, service account tokens expire once an hour by default.

## Implementation information

* Only auth once per gRPC stream.

* Remove auth cache as it's not required when it's only auth once during a gRPC stream/connection.

## Supporting documentation

fixes https://github.com/kumahq/kuma/issues/12784
fixes  https://github.com/kumahq/kuma/issues/12785

replaces https://github.com/kumahq/kuma/pull/12782

### DIFF
--- a/pkg/xds/auth/k8s/authenticator.go
+++ b/pkg/xds/auth/k8s/authenticator.go
@@ -45,7 +45,8 @@ type kubeAuthenticator struct {
 var _ auth.Authenticator = &kubeAuthenticator{}
 
 func (k *kubeAuthenticator) Authenticate(ctx context.Context, resource model.Resource, credential auth.Credential) error {
-	if _, authenticated := k.authenticated.GetIfPresent(credential); authenticated {
+	cacheKey := resource.GetMeta().GetName() + credential
+	if _, authenticated := k.authenticated.GetIfPresent(cacheKey); authenticated {
 		return nil
 	}
 
@@ -61,7 +62,7 @@ func (k *kubeAuthenticator) Authenticate(ctx context.Context, resource model.Res
 		return err
 	}
 
-	k.authenticated.Put(credential, struct{}{})
+	k.authenticated.Put(cacheKey, struct{}{})
 	return nil
 }
 
--- a/pkg/xds/auth/k8s/authenticator.go
+++ b/pkg/xds/auth/k8s/authenticator.go
@@ -73,14 +73,13 @@ func (k *kubeAuthenticator) verifyToken(ctx context.Context, credential auth.Cre
 		},
 	}
 
-	log.V(1).Info("verifying token", "proxy", resourceName, "serviceAccountName", serviceAccountName)
 	if err := k.client.Create(ctx, tokenReview); err != nil {
-		log.Info("[WARNING] fail to call Kubernetes API Server to verify token", "error", err, "proxy", resourceName, "serviceAccountName", serviceAccountName)
+		log.Info("[WARNING] fail to call Kubernetes API Server to verify dataplane token", "error", err, "proxy", resourceName, "serviceAccountName", serviceAccountName)
 		return errors.New("call to TokenReview API failed")
 	}
 	if !tokenReview.Status.Authenticated {
-		log.V(1).Info("fail to verify token", "error", tokenReview.Status.Error, "credential", credential)
-		return errors.Errorf("token doesn't belong to a valid user")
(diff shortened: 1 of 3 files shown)


pkg/xds/auth/k8s/authenticator.go
```diff
@@ -57,22 +44,23 @@ func (k *kubeAuthenticator) Authenticate(ctx context.Context, resource model.Res
...
+func (k *kubeAuthenticator) verifyToken(ctx context.Context, credential auth.Credential, proxyNamespace, serviceAccountName, resourceName string) error {
 	tokenReview := &kube_auth.TokenReview{
 		Spec: kube_auth.TokenReviewSpec{
 			Token: credential,
 		},
 	}
+
 	if err := k.client.Create(ctx, tokenReview); err != nil {
-		return errors.Wrap(err, "call to TokenReview API failed")
+		log.Info("[WARNING] fail to call Kubernetes API Server to verify dataplane token", "error", err, "proxy", resourceName, "serviceAccountName", serviceAccountName)
```
Icarus9913, line 58: Why not `log.Error` ?

jijiechen: Changed to `log.Error`

pkg/xds/auth/k8s/authenticator.go
```diff
@@ -57,22 +44,23 @@ func (k *kubeAuthenticator) Authenticate(ctx context.Context, resource model.Res
...
 	}
+
 	if err := k.client.Create(ctx, tokenReview); err != nil {
-		return errors.Wrap(err, "call to TokenReview API failed")
+		log.Error(err, "fail to call Kubernetes API Server to verify dataplane token", "proxy", resourceName, "serviceAccountName", serviceAccountName)
+		return errors.New("call to TokenReview API failed")
 	}
 	if !tokenReview.Status.Authenticated {
-		return errors.Errorf("token doesn't belong to a valid user")
+		log.Info("[WARNING] fail to verify dataplane token", "error", tokenReview.Status.Error)
```
lahabana, line 62: I'd actually go with:

```
l = logger.With("resourceName", resourceName, "serviceAccountName", serviceAccountName)
l.Debug("dataplane token verification failed")
if tokenReview.Status.Error != "" {
  l.Error("dataplane token verification failed", "error", tokenReview.Status.Error)
  return errors.Errorf("token verification failed with an error")
}
return errors.Errorf("service account token failed authentication")
```

Someone reading the dataplane log should be able to make a difference between a bad token or not succeeding auth no?

pkg/xds/auth/k8s/authenticator.go
```diff
@@ -103,7 +91,7 @@ func (k *kubeAuthenticator) authResource(ctx context.Context, resource model.Res
 		return err
 	}
 
-	if err := k.verifyToken(ctx, credential, namespace, serviceAccountName); err != nil {
+	if err := k.verifyToken(ctx, credential, namespace, serviceAccountName, resource.GetMeta().GetName()); err != nil {
```
lahabana, line 94: This verifyToken is a level of indirection that I find only makes things hard to read don't you think? Should we inline this method instead?

lahabana (commented):
I'd love a bit more content inside the Motivation in your PR description. This would help a lot understand why you are removing the cache for example

pkg/xds/auth/k8s/authenticator.go
jijiechen: Yeah, we can do it. To me, it looks more like this `authResource` method is inlined.

lahabana: Yeah one way or the other :) Just feels that the nesting is just making things harder to read.

Icarus9913 (dismissed):

Icarus9913 (approved):

Issue #12729, opened 2025-02-03
Policies with `spec.from` targeting MeshGateway should be updated to use `spec.rules`


lobkovilya:
### Description

If we want to deprecate `spec.from` then policies like
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTimeout
metadata:
  name: mesh-gateways-timeout-all-default
  namespace: kuma-system
spec:
  targetRef:
    kind: Mesh
    proxyTypes:
    - Gateway
  from:
  - targetRef:
      kind: Mesh
    default:
      http:
        requestHeadersTimeout: 500ms
        streamIdleTimeout: 5s
      idleTimeout: 5m0s
```

also should use `spec.rules`

PR #12791, opened 2025-02-10
feat(kuma-cp): add InboundRules to GatewayRules

Automaat:
## Motivation
We forgot to add this when implementing inbound rules for sidecars

Fix: #12729


lobkovilya (approved):
LGTM

lobkovilya (approved):

PR #12828, opened 2025-02-12
docs(upgrade.md): add doc for dataplane token revocation changes on Universal

jijiechen:
## Motivation

A supplementary change for PR https://github.com/kumahq/kuma/pull/12788 to add `upgrade.md` changes.

The PR introduced a change to the DP authentication that impacts the behaviour of Universal clusters.

## Implementation information

Add change entries in `upgrade.md`


UPGRADE.md
```diff
@@ -45,6 +45,10 @@ New rule:
 The documentation did not mention the `SourceIP` type, but it was possible to create a policy using it instead of `Connection`. Since `SourceIP` 
 is not a correct value, we have decided to deprecate it. If you are using `SourceIP` in your policy, please update it to use `Connection` instead.
 
+#### Changes on revoking dataplane tokens
+
+Revoking a dataplane token on Unversal mesh clusters now requires restarting the mesh control plane, otherwise connected dataplanes will still be able to communicate until they restart/reconnect to the control plane.
```
lahabana, line 50:
```suggestion
Revoking a data-plane token on Universal zones now requires restarting the mesh control plane, otherwise connected data-planes will still be able to communicate until they restart/reconnect to the control plane.
```

I'm not sure this is correct. I think the change is more:

> CP/DP authentication is only checked at connection start. This means that if a token expires or is revoked after the data-plane connects, the connection won't stop. The recommended action on token revocation is to either restart the control-plane or the concerned data-planes

lahabana (commented):
Is there some docs to improve?

UPGRADE.md
jijiechen: Improved. I see more places using `dataplane` than `data-plane`, so I'm using the more common one.

jijiechen:
> Is there some docs to improve?

Created: https://github.com/kumahq/kuma-website/pull/2189

lahabana (approved):

PR #12840, opened 2025-02-12
feat(api-server): add support for label filters

### LINKED ISSUE #12809: Allow listing all resources from the specific zone
### Description

I reviewed the code and found that there's no option to retrieve only ZoneIngress resources—or likely other resources—from a specific zone. Based on a quick look at the data and how it's stored, this isn't straightforward. The zone information is stored in a label, and labels are stored as jsonpb. If filtering by zone is a frequent requirement, we would need to introduce a new column and index it.

# Expected behavior

A user can list all ZoneIngresses in zone-1

lahabana:
## Motivation

We now have labels on all our entities. Some of these labels are very useful (kuma.io/zone for example).
We should be able to use it when filtering api responses.

## Implementation information

Support a basic subset of https://kong-aip.netlify.app/aip/160/ only for labels.

This works for all resource types and while a little costly it should suffice for GUI and kumactl use cases

## Supporting documentation

Fix #12809

### DIFF
--- a/api/mesh/v1alpha1/mesh/rest.yaml
+++ b/api/mesh/v1alpha1/mesh/rest.yaml
@@ -82,6 +82,20 @@ paths:
       operationId: getMeshList
       summary: Returns a list of Mesh in the mesh.
       tags: [ "Mesh" ]
+      parameters:
+        - in: query
+          name: filter
+          description: filter by labels when multiple filters are present, they are ANDed
+          required: false
+          schema:
+            type: object
+            properties:
+              key:
+                type: string
+              value:
+                type: string
+          example:
+            label.k8s.kuma.io/namespace: my-ns
       responses:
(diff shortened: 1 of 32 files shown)


lukidzi (approved):
I feel like it's good for now. 

I have a question
Example:
I have 120 ZoneIngresses in Zone-1 and 1 in Zone-2. If I do a request `filter[label.kuma.io/zone]=zone-2` the resource is not on the first page of the result, will it return the response?

lahabana:
> I have 120 ZoneIngresses in Zone-1 and 1 in Zone-2. If I do a request filter[label.kuma.io/zone]=zone-2 the resource is not on the first page of the result, will it return the response?

Pagination is done after filtering (which makes it very not efficient...) so pages are correct.

PR #12846, opened 2025-02-13
fix(api-server): add defaulter in MeshService and MeshMultizoneService

lahabana:
## Motivation

There are fields like targetPort on port name that default to another value if not set.
This can't be expressed with annotations so defaulting with a `Defaulter` is usefull in these cases.

## Implementation information

There's an open question around what this would do with things like TF as the PUT and GET return something different.


lahabana:
If we go forward with this change we need to make sure the API docs are updated to explain this defaulting

pkg/api-server/testdata/resources/crud/put_meshmultizoneservice_00.golden.json
schogges, line 1: I was wondering why some of the golden files include only empty objects? 🙂

schogges (commented):
LGTM so far 👍 left a Q on the golden files

pkg/api-server/testdata/resources/crud/put_meshmultizoneservice_00.golden.json
lahabana: The default response for a PUT returns an empty object, that's what this first reponse is.

slonka:
Disclaimer: I only checked the description of that it supposed to do (not the code).

There is no other way if this can't be described by simply an annotation. In terraform this can be marked as "computed" and won't show up in the diff. So based on that 👍

PR #12860, opened 2025-02-17
feat(meshcircuitbreaker): supplement HealthyPanicThreshold property

### LINKED ISSUE #11704: Introduce `HealthyPanicThreshold` for MeshCircuitBreaker policy
### Description

I noticed that there's an implementation difference between our MeshCircuitBreaker and Istio's (https://github.com/istio/istio/blob/7b2d8b9dacb7989d75b3b3d860988cadb9f5a848/pilot/pkg/networking/core/cluster_traffic_policy.go#L416-L423), 
maybe we could add one flag to disable panic threshold and it depends on the user's choice.

Icarus9913:
## Motivation

We should supplement HealthyPanicThreshold property in MeshCircuitBreaker policy OutlierDetection section

## Implementation information

Introduce a new `HealthyPanicThreshold` property for MeshCircuitBreaker policy, and its type is `*intstr.IntOrString` just like what we did in MeshHealthCheck policy

- [x] unit test
- [x] e2e test
- [x] verify the MeshCircuitBreaker `HealthyPanicThreshold` overwrite the MeshHealthCheck's

## Supporting documentation

Fix #11704


pkg/plugins/policies/meshcircuitbreaker/plugin/xds/configurer.go
```diff
@@ -210,3 +220,29 @@ func configureTotalFailures(
...
+	}
+	percentage, err := envoyPercent(*value)
+	if err != nil {
+		return err
+	}
+	cluster.CommonLbConfig.HealthyPanicThreshold = percentage
+	return nil
+}
+
+func envoyPercent(intOrStr intstr.IntOrString) (*envoy_type.Percent, error) {
```
lukidzi, line 239: Can we move it to `pkg/util/proto/types.go` ? We have all the Envoy structures there

Icarus9913: I noticed that the `pkg/util/proto` path doesn't import any `github.com/envoyproxy/go-control-plane/envoy/*` dependencies. They import `google.golang.org/protobuf/types/*`

How about putting this function next to https://github.com/kumahq/kuma/blob/0411233912021f6b3392a8155bc0d321ee426ae0/pkg/xds/envoy/listeners/v3/util.go#L84

lukidzi: I'm not sure since this is `listener` code and we configure the cluster. Seems like it should be in some common place

lukidzi (approved):

pkg/plugins/policies/meshcircuitbreaker/plugin/xds/configurer.go
Icarus9913: I'll create a **chore** PR to deal with the `envoyPercent` function in the future once we find a proper place. (This private function exists in multiple places for different policies)

PR #12878, opened 2025-02-18
feat(MeshHealthCheck): deprecate HealthyPanicThreshold property for MeshHealthCheck policy

### LINKED ISSUE #11938: Deprecate `HealthyPanicThreshold` for MeshHealthCheck policy
### Description

The `HealthyPanicThreshold` property is a common configuration for the envoy `config.cluster.v3.Cluster.CommonLbConfig`. It seems more appropriate in MeshCircuitBreaker policy because it's more often used than MeshHealthCheck.

We should deprecate it rather than keep it for both policies, otherwise we hit some conflicts.

Steps from Charly:
 - Deprecate in MeshHealthcheck
 - Document in MeshHealthcheck doc that MeshCircuitBreaker can be used to modify panic threshold
 - During transition make MeshCircuitBreaker override MeshHealthCheck if both are set

Icarus9913:
## Motivation

We need to deprecate `HealthyPanicThreshold` property for MeshHealthCheck policy. Then we gonna supplement this field in MeshCircuitBreaker policy

## Supporting documentation

Fix #11938

### DIFF
--- a/app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
+++ b/app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
@@ -1861,7 +1861,8 @@ spec:
                             Allows to configure panic threshold for Envoy cluster. If not specified,
                             the default is 50%. To disable panic mode, set to 0%.
                             Either int or decimal represented as string.
-                            Deprecated: use MeshCircuitBreaker instead.
+                            Deprecated: the setting has been moved to MeshCircuitBreaker policy,
+                            please use MeshCircuitBreaker policy instead.
                           x-kubernetes-int-or-string: true
                         healthyThreshold:
                           description: |-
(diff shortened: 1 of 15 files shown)


app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
```diff
@@ -1861,6 +1861,7 @@ spec:
                             Allows to configure panic threshold for Envoy cluster. If not specified,
                             the default is 50%. To disable panic mode, set to 0%.
                             Either int or decimal represented as string.
+                            Deprecated: use MeshCircuitBreaker instead.
```
lukidzi, line 1864: Can we add a more clear message? 

example:
> the setting has been moved to MeshCircuitBreaker policy. Please use MeshCircuitBreaker policy instead.

Icarus9913:
Update website: https://github.com/kumahq/kuma-website/pull/2190

lukidzi (approved):

PR #12879, opened 2025-02-18
fix(k8s): fix scope assignment to the resource

lukidzi:
## Motivation

While working on another issue I noticed that we incorrectly set the scope for Kubernetes CRDs.

## Implementation information

When `isNamespaced` set scope `Namespaced`

### DIFF
--- a/pkg/plugins/resources/k8s/native/api/v1alpha1/zz_generated.mesh.go
+++ b/pkg/plugins/resources/k8s/native/api/v1alpha1/zz_generated.mesh.go
@@ -35,7 +35,7 @@ type CircuitBreaker struct {
 }
 
 // +kubebuilder:object:root=true
-// +kubebuilder:resource:scope=Namespaced
+// +kubebuilder:resource:scope=Cluster
 type CircuitBreakerList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata,omitempty"`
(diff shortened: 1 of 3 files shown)


Icarus9913 (changes requested):
2 Questions:
- If we changed the  `+kubebuilder:resource:scope=Namespaced` marker, we need to generate the new k8s CRD yaml right?
- Do we need to touch the legacy policies?

lukidzi:
> If we changed the +kubebuilder:resource:scope=Namespaced marker, we need to generate the new k8s CRD yaml right?

I don't think we need to generate new CRD the scope change touches only `<LegacyPolicy>List` which I belive it's not CRD only `<LegacyPolicy>` is a CRD

> Do we need to touch the legacy policies?

The error was only here

lahabana (commented):
This change has no other side effects? Interesting...

lahabana:
> I don't think we need to generate new CRD the scope change touches only <LegacyPolicy>List which I belive it's not CRD only <LegacyPolicy> is a CRD

Is it maybe unecessary then?

lukidzi:
I feel like they are used in kube clients, for `.List()` but there is no CRD for it. I wanted to keep it consistent

lahabana:
Do we need these annotations yes or no?

lukidzi:
> Do we need these annotations yes or no?

No, we don't

Icarus9913:
OK, it's `*List`. So the comment marker would not impact the CRD, you're right.

Icarus9913 (approved):

lahabana:
> > Do we need these annotations yes or no?
> 
> No, we don't
> 
> https://kubernetes.io/docs/reference/using-api/api-concepts/#collections
> 
> > Keep in mind that the Kubernetes API does not have a kind named List.
> > kind: List is a client-side, internal implementation detail for processing collections that might be of different kinds of object. Avoid depending on kind: List in automation or other code.

So if we don't need them let's remove them no?

lukidzi:
We need objects for k8s client but they are not used in k8s. What I meant `scope` doesn't bother us + we don't need CRDs
> List is a client-side, internal implementation detail for processing collections

PR #12880, opened 2025-02-18
chore(api-server): add offset and size to list resource spec

lahabana:
This was missing from the spec


api/mesh/v1alpha1/mesh/rest.yaml
```diff
@@ -83,6 +83,22 @@ paths:
...
+          description: offset in the list of entities
+          required: false
+          schema:
+            type: integer
+          example: 0
+        - in: query
+          name: size
+          description: the number of items per page
+          required: false
+          default: 100
```
slonka, line 97: default is inside schema @lahabana @lukidzi

Issue #12881, opened 2025-02-18
Collapse Dataplane callbacks


lahabana:
### Description

Currently dataplane callbacks are very complex:

https://github.com/kumahq/kuma/blob/6f98454d0e5948457fc903c62473dc5adcb4a695/pkg/xds/server/v3/components.go#L51-L62

They all wrap the XDS callback: https://github.com/kumahq/kuma/blob/c264fea580b52ea286e9ed4dc39b8e7909403cc8/pkg/xds/server/callbacks/dataplane_callbacks.go#L42

which keeps a map of streamID --> DPP

There are also weirdly tightly coupled (the MetadataTracker is used in the watchdog factory).

There should be a few things we can do to have less callbacks which would lead to less complexity and probably even more efficient code.

I think the first step would be to get rid of the metadataTracker which seems very redundant.

PR #12882, opened 2025-02-18
refactor(dp-server): remove metadata tracker

lahabana:
## Motivation

We were tracking the metadata in its own dataplane callbacks.
In practice this metadata was only used inside the sync_tracker which is itself a dataplane callback.

## Implementation information

We now keep the metadata directly inside the sync_tracker and access it on each watchdog step.
This makes the code more streamlined and removes a hard to understand abstraction that added little value.

## Supporting documentation

Part of #12881


jijiechen:
Looks good to me. 
But can I get this in first to have less conflicts? https://github.com/kumahq/kuma/pull/12886

Issue #12885, opened 2025-02-19
dataplane proxies do not get mTLS identity cert and other xDS resources updated


jijiechen:
### Kuma Version

2.9.3

### Describe the bug

Some of the dataplanes in the mesh were found using expired mTLS certificate, so connection to/from these dataplane proxies could not be established.

In a mTLS enabled mesh cluster, the error seen in other log of other DPPs include:

```
[2025-02-12 00:31:57.150][35][debug][connection] [source/common/tls/ssl_socket.cc:280] [Tags: ""ConnectionId"":""78007""] remote address:172.23.2.43:45846,TLS_error:|268435581:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED:TLS_error_end
```

### To Reproduce

1. Install Kuma service mesh onto a Kubernetes cluster, ideally version 1.30 or above
2. Enable mTLS for the default mesh
3. Install the kuma-demo, and maybe many other apps, with sidecar injection enabled in those app namesapces.
4. Keep triggering traffic among the applications
5. Wait for a few hours/days and check sidecar logs of the DPs

### Expected behavior

Traffic establishes successfully

jijiechen:
From the logs provided by the user, I can see, at `2025-02-09T02:59:31.898Z` the CP was sending stale `envoy.extensions.transport_sockets.tls.v3.Secret resources` (version `565c5411-1595-4b1f-911e-ab75cf750d03`) to DP after it reconnected.

The secret resources should be cleaned up once the DP disconnected, but it wasn't.

* DP disconnected at 2025-02-09T02:59:29.095Z
* It reconnected at 2025-02-09T02:59:31.796Z (edited) 

The expected order is:
* DP disconnect ---> CP clean up (watchers removed) ---> DP re-connect --> watchers created -->  CP re-generate secrets

While the actual order could be:
* DP disconnect ---> DP re-connect  --> watchers created -->  CP clean up (watchers removed) --> CP re-generate secrets

In the second execution process, the updated identity cert (and other xDS updates) will never be received by the DP.

This can happen when the `OnStop` invocation was delayed for some reason: it's run in a go-routine, so it's possible to be scheduled slower than we expected.

https://github.com/kumahq/kuma/blob/master/pkg/xds/server/callbacks/dataplane_sync_tracker.go#L46-L53

jijiechen:
I've seen a 30ms latency on the clean up process of the dataplane connection. 

Here are the callback events:
```
[...]

PR #12886, opened 2025-02-19
fix(kuma-dp): prevent watchers from being cleaned up unexpectedly

jijiechen:
## Motivation

fixing potential xDS update lost by unexpected watcher cleanups

## Implementation information

Only allow a single active connection between a DP and the CP:
* __[this change]__ In the `pkg/xds/server/callbacks/dataplane_sync_tracker.go`, check if there is existing/stale and pending-removing state for a dataplane: if there is, reject this connect attempt
* __[existing behaviour]__ the DP will then retry connecting in a loop-backoff manner
* __[known fact]__ the DP may reach other CP instances which do not have this stale state

This does not change the behaviour of current CP XDS server as we never supported multiple xDS streams from a DP.

## Supporting documentation

fixes https://github.com/kumahq/kuma/issues/12885

### DIFF
--- a/pkg/xds/server/callbacks/dataplane_sync_tracker.go
+++ b/pkg/xds/server/callbacks/dataplane_sync_tracker.go
@@ -2,11 +2,9 @@ package callbacks
 
 import (
 	"context"
-	envoy_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
-	envoy_cache "github.com/envoyproxy/go-control-plane/pkg/cache/v3"
-	"github.com/pkg/errors"
 	stdsync "sync"
-	"time"
+
+	"github.com/pkg/errors"
 
 	"github.com/kumahq/kuma/pkg/core"
 	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
@@ -16,13 +14,11 @@ import (
 
 var dataplaneSyncTrackerLog = core.Log.WithName("xds").WithName("dataplane-sync-tracker")
 
-type NewDataplaneWatchdogFunc func(key core_model.ResourceKey) util_xds_v3.Watchdog
+type NewDataplaneWatchdogFunc func(key core_model.ResourceKey) util_xds_v3.SingletonWatchdog
 
-func NewDataplaneSyncTracker(factoryFunc NewDataplaneWatchdogFunc, hasher envoy_cache.NodeHash, cache envoy_cache.SnapshotCache) DataplaneCallbacks {
+func NewDataplaneSyncTracker(factoryFunc NewDataplaneWatchdogFunc) DataplaneCallbacks {
 	return &dataplaneSyncTracker{
 		newDataplaneWatchdog: factoryFunc,
-		nodeHasher:           hasher,
-		snapshotCache:        cache,
 		watchdogs:            map[core_model.ResourceKey]context.CancelFunc{},
 	}
 }
@@ -42,19 +38,12 @@ type dataplaneSyncTracker struct {
 
 	stdsync.RWMutex // protects access to the fields below
 	watchdogs       map[core_model.ResourceKey]context.CancelFunc
-	nodeHasher      envoy_cache.NodeHash
-	snapshotCache   envoy_cache.SnapshotCache
 }
 
-func (t *dataplaneSyncTracker) OnProxyConnected(streamID core_xds.StreamID, dpKey core_model.ResourceKey, ctx context.Context, _ core_xds.DataplaneMetadata) error {
+func (t *dataplaneSyncTracker) OnProxyConnected(streamID core_xds.StreamID, dpKey core_model.ResourceKey, connCtx context.Context, _ core_xds.DataplaneMetadata) error {
 	// We use OnProxyConnected because there should be only one watchdog for given dataplane.
 	t.Lock()
 	defer t.Unlock()
-
-	// when there is an existing cache to be cleaned up, we wait for it to be removed
-	if err := t.waitForCleaningUpExisting(dpKey, ctx, 3*time.Second); err != nil {
-		return err
-	}
 
 	ctx, cancel := context.WithCancel(context.Background())
 	t.watchdogs[dpKey] = func() {
@@ -62,8 +51,13 @@ func (t *dataplaneSyncTracker) OnProxyConnected(streamID core_xds.StreamID, dpKe
 		cancel()
 	}
 	dataplaneSyncTrackerLog.V(1).Info("starting Watchdog for a Dataplane", "dpKey", dpKey, "streamID", streamID)
+	dataplaneWatchdog := t.newDataplaneWatchdog(dpKey)
+	if dataplaneWatchdog.ExistsStaleState() {
+		dataplaneSyncTrackerLog.Info("[WARNING] rejected a Dataplane connect attempt when it's previous state is still being cleaned up", "dpKey", dpKey, "streamID", streamID)
+		return errors.Errorf("an existing state of the Dataplane is being removed")
+	}
 	//nolint:contextcheck // it's not clear how the parent go-control-plane context lives
-	go t.newDataplaneWatchdog(dpKey).Start(ctx)
+	go dataplaneWatchdog.Start(ctx)
 	return nil
 }
 
@@ -74,43 +68,4 @@ func (t *dataplaneSyncTracker) OnProxyDisconnected(_ context.Context, _ core_xds
 		cancelFn()
 	}
 	delete(t.watchdogs, dpKey)
-}
-
-func (t *dataplaneSyncTracker) waitForCleaningUpExisting(dpKey core_model.ResourceKey, ctx context.Context, maxWait time.Duration) error {
-	timeout := time.After(maxWait)
-	timer := time.NewTicker(50 * time.Millisecond)
-	defer timer.Stop()
-
-	if t.nodeHasher == nil || t.snapshotCache == nil {
-		return nil
-	}
-
-	proxyId := core_xds.FromResourceKey(dpKey)
-	nodeCacheKey := t.nodeHasher.ID(&envoy_core.Node{Id: proxyId.String()})
-	cacheExists := func() error {
-		snapshot, err := t.snapshotCache.GetSnapshot(nodeCacheKey)
-		if err != nil || snapshot == nil {
-			return nil
-		}
-		return errors.New("dataplane is still in the cache")
-	}
-
-	if cacheExists() == nil {
-		return nil
-	}
-
-	for {
-		select {
-		case <-timer.C:
-			if cacheExists() == nil {
-				return nil
-			}
-		case <-timeout:
-			// if the cache is still there, this means there is a performance issue on the CP
-			// in this case, we kick out the connection and the DPP will reconnect later
-			return errors.Errorf("timeout while waiting for dataplane %s to be removed from the cache", proxyId)
-		case <-ctx.Done():
-			return ctx.Err()
-		}
-	}
 }
--- a/pkg/xds/server/callbacks/dataplane_sync_tracker.go
+++ b/pkg/xds/server/callbacks/dataplane_sync_tracker.go
@@ -2,7 +2,11 @@ package callbacks
 
 import (
 	"context"
+	envoy_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoy_cache "github.com/envoyproxy/go-control-plane/pkg/cache/v3"
+	"github.com/pkg/errors"
 	stdsync "sync"
+	"time"
 
 	"github.com/kumahq/kuma/pkg/core"
 	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
@@ -14,9 +18,11 @@ var dataplaneSyncTrackerLog = core.Log.WithName("xds").WithName("dataplane-sync-
 
 type NewDataplaneWatchdogFunc func(key core_model.ResourceKey) util_xds_v3.Watchdog
 
-func NewDataplaneSyncTracker(factoryFunc NewDataplaneWatchdogFunc) DataplaneCallbacks {
+func NewDataplaneSyncTracker(factoryFunc NewDataplaneWatchdogFunc, hasher envoy_cache.NodeHash, cache envoy_cache.SnapshotCache) DataplaneCallbacks {
 	return &dataplaneSyncTracker{
 		newDataplaneWatchdog: factoryFunc,
+		nodeHasher:           hasher,
+		snapshotCache:        cache,
 		watchdogs:            map[core_model.ResourceKey]context.CancelFunc{},
 	}
 }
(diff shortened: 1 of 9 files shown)


lahabana:
There's already code that attempts to do this in xdsCallbacks I believe if anything needs to change it's in there

pkg/xds/sync/dataplane_watchdog.go
```diff
@@ -114,6 +114,48 @@ func (d *DataplaneWatchdog) Cleanup() error {
...
+	if d.dpType == "" {
+		metadata := d.MetadataTracker.Metadata(d.key)
+		if metadata == nil {
+			d.log.Info("[WARNING] could not check state for dataplane without metadata")
+			return false
+		}
+		d.dpType = metadata.GetProxyType()
+	}
+	proxyID := core_xds.FromResourceKey(d.key)
+	switch d.dpType {
```
Icarus9913, line 127: I'd recommend adding some debug-level logs for each case.

jijiechen: It's refactored and this code block is removed.

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -54,22 +52,40 @@ type dpStream struct {
...
+		// execute callback after lock is freed, so heavy callback implementation won't block every callback for every DPP.
+		// proxy disconnect may take some time, we don't want to block the callback for every DPP here
+		disconnectDone := make(chan struct{})
+		d.callbacks.OnProxyDisconnected(dpStream.ctx, streamID, *streamDpKey, func() {
+			go func() { disconnectDone <- struct{}{} }()
+		})
+		go func() {
+			<-disconnectDone
+			d.cleanupDpStream(streamID)
+		}()
```
Icarus9913, line 71: It seems like before calling `d.cleanupDpStream(streamID)` we may call `OnStreamOpen` first, then we gonna hit concurrent problem.

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -54,22 +52,40 @@ type dpStream struct {
...
+	dpStream := d.dpStreams[streamID]
+	streamDpKey = dpStream.dp
+	d.RUnlock()
+
+	if streamDpKey != nil {
+		// execute callback after lock is freed, so heavy callback implementation won't block every callback for every DPP.
+		// proxy disconnect may take some time, we don't want to block the callback for every DPP here
+		disconnectDone := make(chan struct{})
+		d.callbacks.OnProxyDisconnected(dpStream.ctx, streamID, *streamDpKey, func() {
+			go func() { disconnectDone <- struct{}{} }()
```
Icarus9913, line 66: no need goroutine, just call `close(disconnectDone)`

pkg/xds/server/callbacks/dataplane_sync_tracker.go
```diff
@@ -44,21 +46,33 @@ func (t *dataplaneSyncTracker) OnProxyConnected(streamID core_xds.StreamID, dpKe
...
+	t.watchdogs[dpKey] = &trackerCleanup{
+		cancelFunc: func() {
+			dataplaneSyncTrackerLog.V(1).Info("stopping Watchdog for a Dataplane", "dpKey", dpKey, "streamID", streamID)
+			cancel()
+		},
 	}
 	dataplaneSyncTrackerLog.V(1).Info("starting Watchdog for a Dataplane", "dpKey", dpKey, "streamID", streamID)
 	//nolint:contextcheck // it's not clear how the parent go-control-plane context lives
-	go t.newDataplaneWatchdog(dpKey).Start(ctx)
+	go t.newDataplaneWatchdog(dpKey, t.onWatchDogStopped).Start(ctx)
```
lahabana, line 57:
```suggestion
        done := t.watchdogs[dpKey].done
	go func() {
	   defer close(done)
	   t.newDataplaneWatchdog(dpKey, t.onWatchDogStopped).Start(ctx)
	}()
```

pkg/xds/server/callbacks/dataplane_sync_tracker.go
```diff
@@ -44,21 +46,33 @@ func (t *dataplaneSyncTracker) OnProxyConnected(streamID core_xds.StreamID, dpKe
...
 
-func (t *dataplaneSyncTracker) OnProxyDisconnected(_ context.Context, _ core_xds.StreamID, dpKey core_model.ResourceKey) {
+func (t *dataplaneSyncTracker) OnProxyDisconnected(_ context.Context, _ core_xds.StreamID, dpKey core_model.ResourceKey, disconnectDone func()) {
+	t.RLock()
+	defer t.RUnlock()
+	if cleanup := t.watchdogs[dpKey]; cleanup != nil {
+		cleanup.disconnectDone = disconnectDone
+		// kick off stopping of the watchdog
+		cleanup.cancelFunc()
+	}
```
lahabana, line 68:
```go
	t.RLock()
	
        dpData := t.watchdogs[dpKey]
        if dpData == nil {
             t.RUnlock()
             return
       }
       dpData.cancelFunc()
       t.RUnlock()
       <-dpData.done
       t.Lock()
       defer t.Unlock()
       delete(t.watchdogs, dpKey)
```

pkg/xds/server/callbacks/dataplane_sync_tracker.go
```diff
@@ -30,12 +30,14 @@ var _ DataplaneCallbacks = &dataplaneSyncTracker{}
...
-
 	newDataplaneWatchdog NewDataplaneWatchdogFunc
 
 	stdsync.RWMutex // protects access to the fields below
-	watchdogs       map[core_model.ResourceKey]context.CancelFunc
+	watchdogs       map[core_model.ResourceKey]*trackerCleanup
+}
+type trackerCleanup struct {
+	cancelFunc     context.CancelFunc
+	disconnectDone func()
```
lahabana, line 40:
```suggestion
	done chan struct{}
```

pkg/xds/server/callbacks/dataplane_sync_tracker.go
```diff
@@ -30,35 +30,48 @@ var _ DataplaneCallbacks = &dataplaneSyncTracker{}
...
+	dpData := t.watchdogs[dpKey]
+	t.RUnlock()
+
+	if dpData != nil {
+		dpData.cancelFunc()
+		<-dpData.stopped
+		dataplaneSyncTrackerLog.V(1).Info("watchdog for a Dataplane stopped", "dpKey", dpKey, "streamID", streamID)
+		t.Lock()
+		defer t.Unlock()
+		delete(t.watchdogs, dpKey)
```
Icarus9913, line 75: It seems like there would be one concurrent problem:
```text
OnProxyDisconnected RLock --> OnProxyConnected Lock to set 't.watchdogs[dpKey]' --> OnProxyDisconnected Lock to delete 't.watchdogs'
```

Then the `t.watchdogs[dpKey]` would be deleted and it's in `Connected` state.

Please correct me if I'm wrong

jijiechen: RLock will not impact Lock, and when it's Locked, a RLock will not be available. So when it's RLock-ed, we can still lock it in the OnProxyConnected event. The first part is correct.
But I don't think the second part will cause a problem since they are exclusive. 
The two Locking will wait for the lock to be released when it's locked by the other.

jijiechen: OK, so if there is a second connection with the same DPKey established between after the RLock is released and the next Lock statement (for deleting the key), it can cleanup the wrong state.
This seems more like a concern that should be taken care of in the OnProxyConnected event

Icarus9913: OK, the upper dataplane_callback `activeStreams` will prevent this behavior, it's never go into `Connected` state

pkg/xds/server/callbacks/dataplane_callbacks_test.go
```diff
@@ -64,44 +59,90 @@ var _ = Describe("Dataplane Callbacks", func() {
...
 		// when next stream for given data plane proxy is connected
 		err = callbacks.OnStreamOpen(context.Background(), 2, "")
 		Expect(err).ToNot(HaveOccurred())
 		err = callbacks.OnStreamRequest(2, &req)
-		Expect(err).ToNot(HaveOccurred())
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("already an active stream"))
 
-		// then only OnProxyReconnected should be called
+		// then OnProxyReconnected should not be called twice
```
jakubdyszkiewicz, line 84: `OnProxyReconnected` -> `OnProxyConnected`

lahabana (commented):
I quickly read this and it does look right.

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -106,25 +107,26 @@ func (d *xdsCallbacks) OnStreamRequest(streamID core_xds.StreamID, request util_
...
 
-	dpStream := d.dpStreams[streamID]
-	dpStream.dp = &dpKey
-	d.dpStreams[streamID] = dpStream
-
 	activeStreams := d.activeStreams[dpKey]
-	d.activeStreams[dpKey]++
-	d.Unlock()
-
 	if activeStreams == 0 {
```
lahabana, line 115: Can we switch activeStreams to be a bool now?

pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
```diff
@@ -130,51 +130,71 @@ var _ = Describe("Sync", func() {
...
 
+			go callbacks.OnStreamClosed(streamID2, node)
+
 			// when first stream is closed
-			callbacks.OnStreamClosed(1, n)
+			go callbacks.OnStreamClosed(streamID1, node)
+			<-cleanupDone
+			// cleanupDone is inside the watchdog, it's still not the accurate time of cleanup done in the dataplane_callback
+			// to prevent flakiness, we introduce this tiny delay here, it does no harm to the test workflow
+			<-time.After(50 * time.Millisecond)
```
jakubdyszkiewicz, line 178: why do we need it? We have Eventually anyway in the next line

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -106,25 +107,26 @@ func (d *xdsCallbacks) OnStreamRequest(streamID core_xds.StreamID, request util_
...
+		dpStream := d.dpStreams[streamID]
+		dpStream.dp = &dpKey
+		d.dpStreams[streamID] = dpStream
+
+		d.activeStreams[dpKey]++
+		d.Unlock()
```
lahabana, line 121: I'd do:

```
if activeStream == 0 {
   // dpStreamPlay
}
d.Unlock()
if activeStream == 0 {
		if err := d.callbacks.OnProxyConnected(streamID, dpKey, dpStream.ctx, *metadata); err != nil {
			return err
		}
}
```

Rather than having the `Unlock` in two separate code branch

pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
```diff
@@ -130,51 +130,71 @@ var _ = Describe("Sync", func() {
...
 
-			// then watchdog is still active because other stream is opened
+			// then a watchdog is active
 			Eventually(func() int32 {
 				return atomic.LoadInt32(&activeWatchdogs)
 			}, "5s", "10ms").Should(Equal(int32(1)))
 
-			// when other stream is closed
-			callbacks.OnStreamClosed(2, n)
+			// when other stream is closed and the third stream is open
```
jakubdyszkiewicz, line 194: is closed, right?

pkg/xds/server/callbacks/dataplane_callbacks_test.go
```diff
@@ -64,44 +59,90 @@ var _ = Describe("Dataplane Callbacks", func() {
...
+		<-time.After(cleanupCost)
+		// and when the third stream from backend-01 is connected after the first active stream closed and request is sent
+		err = callbacks.OnStreamOpen(context.Background(), streamID3, "")
+		Expect(err).ToNot(HaveOccurred())
+		err = callbacks.OnStreamRequest(streamID3, &envoy_sd.DiscoveryRequest{Node: node})
+		Expect(err).ToNot(HaveOccurred())
+
+		// when other stream is closed and the third stream is open
+		go callbacks.OnStreamClosed(streamID3, node)
+	})
```
jakubdyszkiewicz, line 147: Do we need this test? I feel that this exact behavior is tested in pkg/xds/server/callbacks/dataplane_sync_tracker_test.go already

pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
```diff
@@ -130,51 +130,71 @@ var _ = Describe("Sync", func() {
...
+			err = callbacks.OnStreamRequest(streamID2, &envoy_sd.DiscoveryRequest{Node: node})
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("already an active stream"))
 
 			// then only one watchdog is active
 			Eventually(func() int32 {
 				return atomic.LoadInt32(&activeWatchdogs)
 			}, "5s", "10ms").Should(Equal(int32(1)))
 
+			go callbacks.OnStreamClosed(streamID2, node)
```
jakubdyszkiewicz, line 171: why is it called in a goroutine?

pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
```diff
@@ -130,51 +130,71 @@ var _ = Describe("Sync", func() {
...
+
 			// when first stream is closed
-			callbacks.OnStreamClosed(1, n)
+			go callbacks.OnStreamClosed(streamID1, node)
+			<-cleanupDone
```
jakubdyszkiewicz, line 175: I think channel might be an overkill here? How about just atomic bool you call OnStreamClosed without goroutine and then check if cleanup is Done. This way you check that when OnStreamClosed is finish the watchdog is finished as well, which I think is something you want to verify.

pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
jakubdyszkiewicz: also why is it Eventually? Should this be 0 when cleanup is done? I can't see a race here 🤔

jakubdyszkiewicz (commented):
The fix looks good 👍

pkg/xds/server/callbacks/dataplane_callbacks.go
jijiechen: I'm changing values of this map to streamIDs

jijiechen: I'm changing values of this map to the single active streamID of the paticular data plane.

pkg/xds/server/callbacks/dataplane_callbacks_test.go
jijiechen: Right, it's tested over there. I'm thinking if it's better to test it here and remove the test in that file. The case in that file named `should start only one watchdog per dataplane`, but the `dataplane_sync_tracker.go` itself does not limit it. To me, this feature is provided by this `dataplane_callbacks.go`

pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
jijiechen: Because `OnStreamClosed` is blocking, it includes sending data to channel `cleanupDone` (on line 139) and the receiver of this channel `cleanupDone` is underneath this invocation. Deadlock. 
Well, the deadlock only exists in the case of streamID1, but I also added goroutines for other streams to make it consistency and less confusing.

pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
jijiechen: Let me change it to using an atomic bool. 
I wanted to make an explicit semantic of "waiting for completion", A single atomic field fullfils the requirement of tracking value changes, but does not reflects the feeling of "waiting". This is not so much important though, since we are making the `OnStreamClose` a blocking method.

pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
jijiechen: The `Eventually` was added previously before this refactoring. Let me remove it.

pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
jijiechen: As suggested, I'm changing the `cleanupDone` to an atomic bool, so no need to use the goroutine here any more.

pkg/xds/server/callbacks/dataplane_callbacks_test.go
jijiechen: Removed

jakubdyszkiewicz (approved):

PR #12889, opened 2025-02-19
feat(api): add openApi spec for new proxy and rules API

Automaat:


api/openapi/specs/api.yaml
```diff
@@ -212,6 +212,179 @@ paths:
           $ref: "./common/error_schema.yaml#/components/responses/NotFound"
         '500':
           $ref: '#/components/responses/Internal'
+  /meshes/{mesh}/proxies:
```
lahabana, line 215: Why do we have a list for this?

lahabana:
@Automaat should this be closed as there's a new one?

PR #12890, opened 2025-02-19
Fix/12881 2

lahabana:
## Motivation

Dataplane callbacks was an abstraction causing more complexity then
anything.

## Implementation information

We now have a watchdog_callbacks which start the lifecycle and the
watchdog
We carefully ensure (and test) that there always a single watchdog running at once
even when DP reconnects.

## Supporting documentation

part of https://github.com/kumahq/kuma/issues/12881


pkg/xds/server/callbacks/dataplane_sync_callbacks.go
```diff
@@ -0,0 +1,230 @@
...
+			ctx:              ctx,
+			done:             done,
+			meta:             atomic.Pointer[core_xds.DataplaneMetadata]{},
+			cancelFunc:       cancel,
+		}
+		t.proxyInfos[dpKey] = pInfo
+	}
+	t.activeStreams[streamID].proxyInfo = pInfo
+	ctx := t.activeStreams[streamID].ctx
+	pInfo.Lock()
```
lahabana, line 159: Actually the fact that the long running cleanup runs with this lock means that 2 connections to the same DP may lead to holding the global lock for longer.

pkg/xds/server/callbacks/dataplane_sync_callbacks.go
```diff
@@ -0,0 +1,230 @@
...
+		pInfo.Unlock()
+		return
+	}
+	if streamIdx != len(pInfo.streams)-1 {
+		l.V(1).Info("dpKey is associated with a newer stream", "streams", pInfo.streams)
+		pInfo.streams = append(pInfo.streams[:streamIdx], pInfo.streams[streamIdx+1:]...) // Remove streamID from the list
+		pInfo.Unlock()
+		return // We are not the last stream, we don't care
+	}
+	if sInfo.proxyInfo.cancelFunc != nil {
```
lobkovilya, line 216: since you released the `t.Unlock()` you can't use `sInfo` as it could be accessed for example from the `OnStreamRequest`:
```go
t.activeStreams[streamID].proxyInfo = pInfo
```

pkg/xds/server/callbacks/dataplane_sync_callbacks.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	if metadata.Resource != nil {
+		if core_model.MetaToResourceKey(metadata.Resource.GetMeta()) != dpKey {
+			return fmt.Errorf("proxyId %s does not match proxy resource %s", dpKey, metadata.Resource.GetMeta())
+		}
+	}
+	l := dataplaneSyncLog.WithValues("dpKey", dpKey, "streamID", streamID)
+	t.Lock()
+	if t.activeStreams[streamID].proxyInfo != nil {
+		return nil // We fast return if we already know that this streamID is tracking a specific dataplane
```
lobkovilya, line 141: missing `t.Unlock()` above this line

pkg/xds/server/callbacks/dataplane_sync_callbacks.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			return err
+		}
+
+		l.V(1).Info("starting watchdog")
+		go func() {
+			defer close(pInfo.done)
+			l.V(1).Info("watchdog started")
+			if t.dpWatchdogFactory != nil {
+				t.dpWatchdogFactory.New(dpKey, pInfo.meta.Load).Start(pInfo.ctx)
```
lobkovilya, line 175: I'm trying to understand the purpose of `atomic.Pointer` here. Isn't it the same as 
```go
go func(metaPtr *core_xds.DataplaneMetadata) {
	defer close(pInfo.done)
	l.V(1).Info("watchdog started")
	if t.dpWatchdogFactory != nil {
		t.dpWatchdogFactory.New(dpKey, metaPtr).Start(pInfo.ctx)
	}
}(metadata)
```
assuming of course `t.dpWatchdogFactory.New` accepts `*DataplaneMetadata` instead of ` func() *DataplaneMetadata`. In this case, pointer will be copied as value so you don't need to access `pInfo` field anymore

lahabana:
Switched to draft. Removing this would be good but things have changed a little so this PR needs to be reworked

pkg/xds/server/callbacks/dataplane_sync_callbacks.go
jijiechen: Here `sInfo` is a copy to the value in the map, right? So this statement should still work even the proxyInfo was re-assigned in the map from the method `OnStreamRequest`.

pkg/xds/server/callbacks/dataplane_sync_callbacks.go
```diff
@@ -0,0 +1,230 @@
...
+		}()
+	} else {
+		l.V(1).Info("watchdog was started by a previous stream, not starting a new one", "activeStreams", pInfo.streams)
+	}
+
+	return nil
+}
+
+func (t *dataplaneSyncCallbacks) OnStreamClosed(streamID core_xds.StreamID) {
+	t.Lock()
```
jijiechen, line 186: So many lock/unlock invocations in this method. I'm thinking how can we improve it to make it cleaner.

jijiechen:
I like this change. It simplifies things so much.

PR #12893, opened 2025-02-19
fix(kds): rework cross zone syncing

### LINKED ISSUE #11740: Only allow `MeshMultiZoneService` to be created on global
### What happened?

The remote zones need to know about the existence of MMZS otherwise cross-zone doesn't work, so zone-specific MMZS can't work. We should block creation of them on zones.

lahabana:
# Motivation

Up until now it was really hard to define the behaviour of resources in KDS in a generic way leading to code complexity and bugs:

- Resources that only existed in global could be created on zones
- Problems with secrets colliding between global and zone
- Partial sync (only some secrets synced from Global to zones)

# Implementation details

We rethink KDS using behaviors instead on a per policy basis.
The behaviors are:
- Can a resource be created on global or zone
- Is a resource be "consumed" (used) on global or zone
- Is a resource synced across zones

With this we can decide what gets sent around in KDS the filters and mapper.

What's important is that these flags do not decide *which* resources are synced these are further rules in KDS for .e.g:
- Don't send to global resources that originated from global
- Don't send to zone resources coming from this same zone
- Only send across policies that are of "producer" style (which makes resources like MeshService work out the box)
- Don't sync some configs and secrets that are specific to a CP (tokens private keys for .e.g).

# Supporting documentation

Fix #11740
Created #12916


lukidzi (approved):

lahabana:
Extracting a refactor in #12913

api/mesh/options.proto
```diff
@@ -22,9 +22,12 @@ message KumaResourceOptions {
   // Whether to skip type registration for this resource.
   bool skip_registration = 6;
 
-  KumaKdsOptions kds = 10;
```
lukidzi, line 25: Can we change the type of existing field in proto? Won't it break anything?  https://protobuf.dev/best-practices/dos-donts/#:~:text=Don't%20Change%20the%20Type%20of%20a%20Field,-Almost%20never%20change&text=The%20protobuf%20docs%20outline%20a,superset%20of%20the%20old%20one.

pkg/kds/context/context.go
```diff
@@ -43,11 +44,11 @@ var log = core.Log.WithName("kds")
 
 type Context struct {
 	ZoneClientCtx         context.Context
+	TypesSentByZone       []core_model.ResourceType
+	TypesSentByGlobal     []core_model.ResourceType
 	GlobalProvidedFilter  reconcile_v2.ResourceFilter
 	ZoneProvidedFilter    reconcile_v2.ResourceFilter
 	GlobalServerFiltersV2 []mux.FilterV2
-	// Configs contains the names of system.ConfigResource that will be transferred from Global to Zone
-	Configs map[string]bool
```
lobkovilya, line 50: This field is updated in KM, not sure if you can remove it

pkg/core/resources/model/resource.go
```diff
@@ -54,35 +54,24 @@ type KDSFlagType uint32
...
-	// Can't be used as KDS flag for resource type.
-	GlobalToZoneSelector = GlobalToAllZonesFlag | GlobalToAllButOriginalZoneFlag
-
-	// AllowedOnGlobalSelector is selector for all flags that indicate resource can be created on Global.
-	AllowedOnGlobalSelector = GlobalToAllZonesFlag
-
-	// AllowedOnZoneSelector is selector for all flags that indicate resource can be created on Zone.
-	AllowedOnZoneSelector = ZoneToGlobalFlag | GlobalToAllButOriginalZoneFlag
+	KDSDisabledFlag      = KDSFlagType(0)
+	ConsumedByZoneFlag   = KDSFlagType(1)
```
lobkovilya, line 58: What are the cases, when
* the resource type has `ConsumedByZoneFlag` but **not** `ProvidedByGlobalFlag`
* the resource type has `ConsumedByGlobalFlag` but **not** `ProvidedByZoneFlag`

api/mesh/options.proto
lahabana: yes exactly it's not shared so it's fine

pkg/kds/context/context.go
lahabana: Let me look.

pkg/core/resources/model/resource.go
lobkovilya: Ah, I see, so basically `ConsumedByX` doesn't necessarily mean "consumed through KDS", it means "resource type on X can be created through KDS, API, Kube store". I think it would be nice to add docs for these flags

lahabana: Sounds good!

pkg/kds/context/context.go
lahabana: Ok changed this to be a global var. This feels less akward than the accidental sharing that is happening now.

lukidzi (approved):

pkg/core/resources/model/resource.go
lahabana: done

pkg/kds/context/context.go
lahabana: There will be more KM changes needed anyway with the change to the flags

api/system/v1alpha1/zone.proto
```diff
@@ -16,6 +16,8 @@ message Zone {
   option (kuma.mesh.resource).package = "system";
   option (kuma.mesh.resource).global = true;
   option (kuma.mesh.resource).ws.name = "zone";
+  // Zones are only managed globally
+  option (kuma.mesh.resource).kds = "model.ProvidedByGlobalFlag";
```
lobkovilya, line 20: Found it https://github.com/kumahq/kuma/pull/8584

lahabana: Ok so zone is synced from global to zone?

lahabana: what happens on standalone?

lobkovilya: It's created automatically on standalone with the name `default` https://github.com/kumahq/kuma/pull/8584/files#diff-55d71903df6fb064b2457cc866dcc4a4c379bb99b6ed37eee3475dc7925d74efR20

lahabana: fixing this then

lahabana: Got super confused for a long time here :)

pkg/kds/context/context.go
```diff
@@ -260,31 +256,37 @@ func UpdateResourceMeta(fs ...util.CloneResourceMetaOpt) reconcile_v2.ResourceMa
...
+			_, exists := KDSSyncedConfigs[r.GetMeta().GetName()]
+			return exists
+		case system.GlobalSecretType:
+			if slices.Contains([]string{system.EnvoyAdminCA, system.AdminUserToken, system.InterCpCA, system.UserTokenRevocations}, r.GetMeta().GetName()) {
+				return false
+			}
+			if strings.HasPrefix(r.GetMeta().GetName(), system.UserTokenSigningKeyPrefix) {
+				return false
+			}
+		case system.SecretType:
```
lobkovilya, line 273: nit: the behaviour will be the same even without this line, right?

pkg/kds/context/context.go
```diff
@@ -353,14 +355,8 @@ func SkipUnsupportedHostnameGenerator(ctx context.Context, clusterID string, fea
...
-	if zi, ok := r.(*core_mesh.ZoneIngressResource); ok {
-		// Old zones don't have a 'kuma.io/zone' label on ZoneIngress, when upgrading to the new 2.6 version
-		// we don't want Zone CP to sync ZoneIngresses without 'kuma.io/zone' label to Global pretending
-		// they're originating here. That's why upgrade from 2.5 to 2.6 (and 2.7) requires casting resource
-		// to *core_mesh.ZoneIngressResource and checking its 'spec.zone' field.
-		// todo: remove in 2 releases after 2.6.x
-		return !zi.IsRemoteIngress(localZone)
-	}
-	return core_model.IsLocallyOriginated(config_core.Zone, r.GetMeta().GetLabels()) || r.Descriptor().KDSFlags == core_model.ZoneToGlobalFlag
+	return core_model.IsLocallyOriginated(config_core.Zone, r.GetMeta().GetLabels())
```
lobkovilya, line 359: Don't we still need to check the KDSFlag here?

pkg/core/resources/model/resource.go
```diff
@@ -284,6 +279,16 @@ func (d ResourceTypeDescriptor) IsInsight() bool {
 	return strings.HasSuffix(string(d.Name), "Insight")
 }
 
+func (d ResourceTypeDescriptor) IsReadOnly(isGlobal bool, isFederated bool) bool {
+	if d.ReadOnly {
+		return false
```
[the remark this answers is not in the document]
lahabana: YEs

pkg/kds/context/context.go
lahabana: Nope because we only pass resources that use KDS in the filter already: https://github.com/kumahq/kuma/blob/4ab84acd60a6921909c287444c2276071dcc24ed/pkg/kds/context/context.go#L47-L48

so checking here was just redundant, on top of this we were using `||` which probably means we were sending back up stuff without origin labels which is not longer possible

lobkovilya (approved):

lukidzi (approved):
🎉

Issue #12894, opened 2025-02-20
Invalid `DeepCopyInto` when a property is a pointer to a typealiased map[string]string


slonka:
> [!WARNING]
> Once this is done add `MeshService` to linter in `api-lint` make target!

### Kuma Version

ef080746245a7b6d55994ce961fdb91f039f8e89

### Describe the bug

If I have a struct like this:

```
type DataplaneTags map[string]string

type Selector struct {
	DataplaneTags DataplaneTags `json:"dataplaneTags,omitempty"`
	DataplaneRef  *DataplaneRef `json:"dataplaneRef,omitempty"`
}
```

https://github.com/slonka/kuma/blob/ef080746245a7b6d55994ce961fdb91f039f8e89/pkg/core/resources/apis/meshservice/api/v1alpha1/meshservice.go#L11

then the generated [DeepCopy](https://github.com/slonka/kuma/blob/ef080746245a7b6d55994ce961fdb91f039f8e89/pkg/core/resources/apis/meshservice/api/v1alpha1/zz_generated.deepcopy.go#L154) does not compile:

```
Cannot use 'new(map[string] string)' (type *map[string] string) as the type DataplaneTags
```

### To Reproduce

Clone https://github.com/slonka/kuma/blob/ef080746245a7b6d55994ce961fdb91f039f8e89 and try to compile.

### Expected behavior

It compiles.

### Additional context (optional)

This was needed when I tried to make MeshService compliant with our linter https://github.com/kumahq/kuma/blob/7bafa578aad6e528befcb6c96f025542fd1f6870/tools/ci/api-linter/linter/api-linter.go#L52:

```
go run tools/ci/api-linter/main.go $(find pkg/core/resources/apis/mesh*/api/v1alpha1/ -type d -maxdepth 0 | sed 's|^|github.com/kumahq/kuma/|')

projects/kuma/pkg/core/resources/apis/meshservice/api/v1alpha1/meshservice.go:43:2: field MeshService.State does not match any allowed non-mergeable category
projects/kuma/pkg/core/resources/apis/meshservice/api/v1alpha1/meshservice.go:14:2: field MeshService.Selector.DataplaneTags does not match any allowed non-mergeable category
projects/kuma/pkg/core/resources/apis/meshservice/api/v1alpha1/meshservice.go:19:2: field MeshService.Selector.DataplaneRef.Name does not match any allowed non-mergeable category
[...]

PR #12844, opened 2025-02-13
fix(api): adjust policies properties

slonka:
## Motivation

Best described in https://github.com/kumahq/kuma/issues/12790.

## Implementation information

Hook up the linter and adjust all other policies fields.

## Supporting documentation

Closes https://github.com/kumahq/kuma/issues/12790

PR #12895, opened 2025-02-20
fix(api): adjust resources

slonka:
## Motivation

https://github.com/kumahq/kuma/pull/12844 got split into two so it's smaller change.

It's missing MeshService because of https://github.com/kumahq/kuma/issues/12894 but it's not needed now.

## Implementation information

Just like https://github.com/kumahq/kuma/pull/12844.

## Supporting documentation

Closes https://github.com/kumahq/kuma/issues/12845


pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/meshmultizoneservice.go
```diff
@@ -21,14 +21,15 @@ type MeshMultiZoneService struct {
...
 
 type Port struct {
-	Name string `json:"name,omitempty"`
-	Port uint32 `json:"port"`
+	Name *string `json:"name,omitempty"`
+	Port uint32  `json:"port"`
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=tcp
-	AppProtocol core_mesh.Protocol `json:"appProtocol,omitempty"`
+	AppProtocol core_mesh.Protocol `json:"appProtocol"`
```
lobkovilya, line 32: Looks like we're missing proper validation on MMZS from the beginning, potentially `appProtocol: ""` can be introduced in the code here https://github.com/kumahq/kuma/blob/74af069803862ec569ed5ffe43974ee6b092b274/pkg/plugins/policies/core/xds/meshroute/listeners.go#L213

Technically it was like that before your change, so I don't know maybe it should be a separate issue.

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/validator.go
```diff
@@ -57,10 +57,12 @@ func validateTls(tls *Tls) validators.ValidationError {
 		if tls.Verification.ServerName != nil && !govalidator.IsDNSName(*tls.Verification.ServerName) {
 			verr.AddViolationAt(path.Field("serverName"), "must be a valid DNS name")
 		}
-		if tls.Verification.Mode != nil {
-			if !slices.Contains(allVerificationModes, string(*tls.Verification.Mode)) {
-				verr.AddErrorAt(path.Field("mode"), validators.MakeFieldMustBeOneOfErr("mode", allVerificationModes...))
-			}
+		if tls.Verification.Mode == "" {
+			// code default to be backwards compatible
+			tls.Verification.Mode = TLSVerificationSecured
```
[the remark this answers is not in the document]
slonka: for this we use a switch here: https://github.com/kumahq/kuma/blob/3d0aa873c2bbac6eae062481eab9d33e66e8410e/pkg/xds/topology/outbound.go#L734 that defaults to doing nothing which is equivalent to `TLSVerificationSecured`. I can add the same code default there as well if you want

api/mesh/v1alpha1/mesh.proto
```diff
@@ -227,15 +227,15 @@ message DatadogTracingBackendConfig {
...
-  bool splitService = 3;
+  google.protobuf.BoolValue splitService = 3;
 }
 
 message ZipkinTracingBackendConfig {
   // Address of Zipkin collector.
   string url = 1;
 
   // Generate 128bit traces. Default: false
-  bool traceId128bit = 2;
```
lukidzi, line 238: Doesn't it break helpers?  `bool` is false by default, now it can be nil.

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/validator.go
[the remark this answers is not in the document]
lobkovilya: sh I see, yes I think it's better to handle `""` explicitly in this switch, maybe with a comment that it's backward compatibility

api/mesh/v1alpha1/mesh.proto
slonka: we unpack it with `pointer.Deref` https://github.com/kumahq/kuma/pull/12895/files#diff-adcdddd3271aaed6ae6844e294a9d0674c1dd87bd0117557f5d75a7ff0c1999aR104

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
```diff
@@ -94,8 +98,9 @@ const (
 
 type Verification struct {
 	// Mode defines if proxy should skip verification, one of `SkipSAN`, `SkipCA`, `Secured`, `SkipAll`. Default `Secured`.
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=Secured
-	Mode *VerificationMode `json:"mode,omitempty"`
+	Mode VerificationMode `json:"mode"`
```
lobkovilya, line 103: maybe it is worth adding an enum annotation here

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
```diff
@@ -118,8 +123,9 @@ const (
 
 type SANMatch struct {
 	// Type specifies matching type, one of `Exact`, `Prefix`. Default: `Exact`
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=Exact
-	Type SANMatchType `json:"type,omitempty"`
+	Type SANMatchType `json:"type"`
```
lobkovilya, line 128: and here

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/meshmultizoneservice.go
lobkovilya, line 32: we can also add an enum annotation here, the same way we have on MES:
```go
// +kubebuilder:validation:Enum=tcp;grpc;http;http2
Protocol core_mesh.Protocol `json:"protocol"`
```

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/validator.go
```diff
@@ -13,5 +16,13 @@ func (r *MeshMultiZoneServiceResource) validate() error {
 	if len(r.Spec.Ports) == 0 {
 		verr.AddViolationAt(path.Field("ports"), "cannot be empty")
 	}
+	for i, port := range r.Spec.Ports {
+		if port.Name != nil && *port.Name == "" {
+			verr.AddViolationAt(path.Field("ports").Index(i).Field("name"), validators.MustNotBeEmpty)
+		}
+		if !slices.Contains(mesh.SupportedProtocols, port.AppProtocol) {
```
lobkovilya, line 23: it's a bit concerning MES allows `tcp;grpc;http;http2` but `mesh.SupportedProtocols` also has `kafka`

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
slonka: If we add enum then universal zones that did not have that set can send empty mode and that would mean error on kubebuilder validation right? soooo ideally we would have to write a migration for that

slonka: we agreed on slack to not add enums yet, will create an issue for this

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
slonka: we agreed on slack to not add enums yet, will create an issue for this

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/meshmultizoneservice.go
slonka: we agreed on slack to not add enums yet, will create an issue for this

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/validator.go
slonka: https://github.com/kumahq/kuma/issues/12912 added issue about this

lobkovilya (approved):

PR #12904, opened 2025-02-21
feat(release): update versions/changelog on release branches

lukidzi:
## Motivation

Our `versions.yaml` file on release branches is outdated. While this doesn’t affect the information we retrieve, as we read it from master, it might impact the Helm upgrade E2E tests since we don’t attempt to upgrade to the latest patch version.

## Implementation information

Changed to logic to get versions from `versions.yaml` and later find the latest patch release in helm repository.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -31,71 +31,13 @@ env:
 permissions:
   contents: read
 jobs:
-  get-active-release-branches:
-    runs-on: ubuntu-24.04
-    permissions:
-      contents: read
-    outputs:
-      branches: ${{ steps.get-branches.outputs.branches }}
-    steps:
-      - name: "Get active branches"
-        id: get-branches
-        env:
-          REPOSITORY: ${{ github.repository }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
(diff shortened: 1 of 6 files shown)


lahabana (changes requested):
I don't think that's right.

I agree that latest patch could be annoying but we should find another way

lukidzi:
I noticed a few alternative ways to get the newest patch version:

1. Update patch versions on release branches – This approach updates only the versions.yaml file on the release branch with the latest patch 
2. Use helm search to find the newest patch – This might be unreliable due to potential network issues.
3. Generate in a step/makefile a file based on the releases and set the `VERSIONS_YAML_PATH` or introduce a new env  variable for E2E tests.

lahabana:
I'd go with either:
> Use helm search to find the newest patch – This might be unreliable due to potential network issues.
or use the ednpoint in the docs (I can't remember what it is exactly).

lukidzi:
> I'd go with either:
> 
> > Use helm search to find the newest patch – This might be unreliable due to potential network issues.
> > or use the ednpoint in the docs (I can't remember what it is exactly).

I don’t think we currently expose `versions.yaml` on the Kuma.io (couldn't find, only latest_vesion), but the file is available, and we can expose it if needed. This might be a better option than Helm since we also use compatibility check on Universal.

lukidzi:
I am going to reopen/recreate once we find the best way to handle it.

Issue #12413, opened 2025-01-02
Combine existing defaults configuration into one `mode` switch


slonka:
### Description

Right now we have 4 switches (skipMeshCreation, skipTenantResources, createMeshRoutingResources, skipHostnameGenerators).

Following the [MADR](https://docs.google.com/document/d/1Ewe0yRx-vvAuQn5CWRFhJbBPLzdbBQ2wsQosZXk4Go8/edit?tab=t.0#heading=h.igm0z4k7xidf) we [decided to](https://docs.google.com/document/d/1Ewe0yRx-vvAuQn5CWRFhJbBPLzdbBQ2wsQosZXk4Go8/edit?tab=t.0#heading=h.benn0r4w02v1) combine 4 switches to skip different resources into one `mode` option.

Suggested mode value:

I suggest we have the following values for mode:
- None (prod/IaC use case)
- All - default - (demo use case)

Because for serious use cases (prod, IaC) you want to track things manually in git and for demo you want to skip repeating the same things.

lahabana:
For MinK this would be a parameter of the VCP? Or MinK would be none going forward

slonka:
Most likely it will be none going forward.

slonka:
Chagnes after today's meetings - https://docs.google.com/document/d/1tIfshA7-RFAgpWvptZ2iCYE6KUEMW-VHY-O2NRXo5r0/edit?tab=t.0#bookmark=id.81n9f1qhnhnh

slonka:
Instead of this we decided to extend the `/tenants` endpoints.

slonka:
Not planned for now.

PR #12914, opened 2025-02-24
feat(config): combine default switches into mode

slonka:
combine SkipMeshCreation SkipTenantResources and SkipHostnameGenerators into Mode

## Supporting documentation

Closes https://github.com/kumahq/kuma/issues/12413

### DIFF
--- a/pkg/config/app/kuma-cp/config.go
+++ b/pkg/config/app/kuma-cp/config.go
@@ -32,17 +32,31 @@ var _ config.Config = &Config{}
 
 var _ config.Config = &Defaults{}
 
+const (
+	ModeNone DefaultsMode = "none"
+	ModeDemo DefaultsMode = "demo"
+)
+
+type DefaultsMode string
+
 type Defaults struct {
 	config.BaseConfig
 
+	// Mode for defining default resources. Possible values: "none", "demo".
+	// None - no default resources will be created. This is best suited for IaC and production use cases where you want to precisely control what's created.
+	// Demo - default resources will be created. This is best suited for demoing and testing use cases where you want to quickly get started.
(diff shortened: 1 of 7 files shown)


pkg/config/app/kuma-cp/config.go
```diff
@@ -32,17 +32,31 @@ var _ config.Config = &Config{}
...
+
 type Defaults struct {
 	config.BaseConfig
 
+	// Mode for defining default resources. Possible values: "none", "demo".
+	// None - no default resources will be created. This is best suited for IaC and production use cases where you want to precisely control what's created.
+	// Demo - default resources will be created. This is best suited for demoing and testing use cases where you want to quickly get started.
+	Mode DefaultsMode `json:"mode" envconfig:"kuma_defaults_mode"`
 	// If true, it skips creating the default Mesh
+	// Deprecated: use mode instead.
```
lahabana, line 50: 1. There is a way to add deprecation warnings in the logs which we should probably leverage here.
2. be mindful of mink and CD and more generally backward compat
3. in k8s should we auto default to none and have the default stuff be part of the helm chart?

slonka: 1. In config files it looks the same (word, "Deprecated" followed by ":"):

https://github.com/kumahq/kuma/blob/1f6f11bded0b389da6b911d35405cf3f7f59434f/pkg/config/plugins/runtime/k8s/config.go#L137-L139

There is also a warning in Validate() function:

https://github.com/kumahq/kuma/blob/1f6f11bded0b389da6b911d35405cf3f7f59434f/pkg/config/plugins/runtime/k8s/config.go#L448-L450

So I assume it's about adding it in `Validate()` function.

---

will respond to rest of the points in a later comment

slonka:
Not needed now, closing until we decide to do this work.

PR #12937, opened 2025-02-24
test(e2e): use unique name of the resource

lukidzi:
## Motivation

We were using the same name as other policy. Let's see if it helps


lahabana (dismissed):

test/e2e_env/kubernetes/meshcircuitbreaker/meshcircuitbreaker.go
```diff
@@ -206,7 +206,6 @@ spec:
           maxConnections: 1
           maxPendingRequests: 1
           maxRequests: 1
```
lukidzi, line 208:
```suggestion
          maxRequests: 1
          maxRetries: 1
```

bartsmykla (approved):

PR #12946, opened 2025-02-25
test(e2e): make connections shorter when using MLBS

lukidzi:
## Motivation

Noticed a flake: Once Envoy establishes a connection, it might continue routing traffic to the same instance, which is not expected by our MLBS. This happens because connections are persisted and reused. They can be closed by a timeout or a cluster reload. In the test, I observed that the test eventually passes, but initially, it takes time to start using the policy strategy.

## Implementation Information

I introduced a timeout policy to shorten connection durations. This should ensure that routing aligns with our strategy sooner.

### DIFF
--- a/test/e2e_env/kubernetes/gateway/gateway.go
+++ b/test/e2e_env/kubernetes/gateway/gateway.go
@@ -536,6 +536,28 @@ spec:
 	})
 
 	Context("MeshLoadBalancingStrategy", func() {
+		// Introduce a shorter connection duration because, once a connection is established,
+		// the load balancer might continue routing traffic to existing instances.
+		mt := fmt.Sprintf(`
+apiVersion: kuma.io/v1alpha1
+kind: MeshTimeout
+metadata:
+  name: mt-close-connection
+  namespace: %s
+  labels:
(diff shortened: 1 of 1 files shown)


test/e2e_env/kubernetes/gateway/gateway.go
```diff
@@ -536,6 +536,28 @@ spec:
...
+spec:
+  targetRef:
+    kind: Mesh
+    proxyTypes:
+    - Gateway
+  to:
+  - targetRef:
+      kind: Mesh
+    default:
+      idleTimeout: 10s
```
lahabana, line 558: Is the idleTimeout actually useful?

lahabana (dismissed):

test/e2e_env/kubernetes/gateway/gateway.go
lukidzi: Might not be required in this case since we don't have long periods without any requests.

lukidzi, line 558:
```suggestion
```

lahabana (approved):

PR #12947, opened 2025-02-25
test(e2e): introduce x-set-response-code to simulate upstream error

lukidzi:
## Motivation

Errors returned by the Circuit Breaker do not counts towards outlier detection. It checks only upstream errors that's why we didn't get correctly into panic mode.

## Implementation information

Added `x-set-response-code` to return specific error code and change to 1 failure


lahabana (approved):
Thx!

PR #12948, opened 2025-02-25
test(unit): check correctly active watchdogs

lukidzi:
## Motivation

Noticed a flake of the test.

## Implementation information

I was reviewing the test and noticed that we have an active watchdog, but our assertion was incorrectly checking that we didn't. Since this happens asynchronously, there's a potential race condition where the value increases right before the assertion. I updated it to use `Eventually` to verify that the watchdog is active.


lahabana (approved):

jijiechen:
Thank you so much for fixing this for me.

PR #12960, opened 2025-02-26
test(e2e): use manifest instead of operator to install calico

lukidzi:
## Motivation

We noticed it a bit flaky so maybe using manifest is more stable


jijiechen (approved):
Let's give it a try.

PR #12975, opened 2025-02-27
fix(dpp): save and update labels from Dataplane resource on Universal

Automaat:
## Motivation

We need to handle dpp labels properly in order to be able to select them in policies

### DIFF
--- a/pkg/core/managers/apis/dataplane/dataplane_manager.go
+++ b/pkg/core/managers/apis/dataplane/dataplane_manager.go
@@ -2,8 +2,6 @@ package dataplane
 
 import (
 	"context"
-	"maps"
-
 	"github.com/pkg/errors"
 
 	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
@@ -56,15 +54,10 @@ func (m *dataplaneManager) Create(ctx context.Context, resource core_model.Resou
 	m.setInboundsClusterTag(dp)
 	m.setGatewayClusterTag(dp)
 	m.setHealth(dp)
-	existingLabels := map[string]string{}
-	maps.Copy(existingLabels, opts.Labels)
-	if resource.GetMeta() != nil && resource.GetMeta().GetLabels() != nil {
-		maps.Copy(existingLabels, resource.GetMeta().GetLabels())
-	}
 	labels, err := core_model.ComputeLabels(
 		resource.Descriptor(),
 		resource.GetSpec(),
-		existingLabels,
+		opts.Labels,
 		core_model.UnsetNamespace,
 		opts.Mesh,
 		m.mode,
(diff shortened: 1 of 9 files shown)


pkg/core/managers/apis/dataplane/dataplane_manager.go
```diff
@@ -55,10 +56,15 @@ func (m *dataplaneManager) Create(ctx context.Context, resource core_model.Resou
 	m.setInboundsClusterTag(dp)
 	m.setGatewayClusterTag(dp)
 	m.setHealth(dp)
+	existingLabels := map[string]string{}
+	maps.Copy(existingLabels, opts.Labels)
+	if resource.GetMeta() != nil && resource.GetMeta().GetLabels() != nil {
```
lobkovilya, line 61: Something is odd here, I don't think we expect `resource.GetMeta()` to be set in the `Create` method, no?

Automaat: It is set in case of creating dp. Because we are doing upsert on dp creation: https://github.com/kumahq/kuma/blob/72bc7f18f2f6cbb14816e2477246dd3191fede7e/pkg/xds/server/callbacks/dataplane_lifecycle.go#L123-L129

lobkovilya: It's set only in this PR with `existing.SetMeta(md.Resource.GetMeta())`, I don't think we've done it before. Normally, I don't think you can set meta when calling `Create` because ResourceStores are the only ones who can create store-specific meta.

lobkovilya:
On Universal e2e, we create dataplanes using the same flow. I think it is worth adding labels as another `AppDeploymentOption` [here](https://github.com/kumahq/kuma/blob/4e3c0ae592894da36cc4c602529c51660cd8bdff/test/framework/setup.go#L867) in one of the tests and check they're successfully applied.

test/e2e_env/universal/auth/dp_auth.go
```diff
@@ -51,18 +51,19 @@ func DpAuth() {
...
 		// when
-		err := TestServerUniversal("dp-02", meshName, WithServiceName("test-server"))(universal.Cluster)
+		err := TestServerUniversal("dp-02", meshName, WithServiceName("test-server"), WithAppLabel("test-server"))(universal.Cluster)
 		Expect(err).ToNot(HaveOccurred())
 
 		// then
 		Eventually(func() (string, error) {
 			return universal.Cluster.GetKumactlOptions().RunKumactlAndGetOutput("get", "dataplanes", "-oyaml")
-		}, "30s", "1s").ShouldNot(ContainSubstring("192.168.0.2"))
+		}, "30s", "1s").ShouldNot(And(ContainSubstring("192.168.0.2"), ContainSubstring("not-test-server")))
```
lobkovilya, line 67: Please check if it actually contains `app: test-server`, otherwise this test would'be passed before your fix

Automaat: updated asssertion

lobkovilya (approved):

lobkovilya (approved):

PR #12980, opened 2025-02-28
fix(kuma-cp): move package level variable to struct

lukidzi:
## Motivation

Data race, since in the test we create multiple instances we hit a race condition.

## Implementation information

The logger was on a package level, meaning each call for `NewCollector` overrides the logger. Move it to the structure

### DIFF
--- a/pkg/gc/collector.go
+++ b/pkg/gc/collector.go
@@ -57,14 +57,14 @@ func NewCollector(
 		newTicker:          newTicker,
 		metric:             metric,
 		resourcesToCleanup: resourcesToCleanup,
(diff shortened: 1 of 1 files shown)


pkg/gc/collector.go
```diff
@@ -34,6 +32,7 @@ type collector struct {
 	newTicker          func() *time.Ticker
 	metric             prometheus.Summary
 	resourcesToCleanup map[InsightType]ResourceType
+	gcLog              logr.Logger
```
lobkovilya, line 35: nit: can be just `log` as it's already in `gc` package

lukidzi, line 35:
```suggestion
	log                logr.Logger
```

lobkovilya (approved):

PR #12983, opened 2025-03-03
fix(kds): detect properly hanging stream for the same zone

lukidzi:
## Motivation

In a multi-zone architecture, we run a global control plane and multiple zone control planes. Each zone can have multiple instances running, but only the leader is responsible for communicating with the global control plane.

Currently, if there is an issue with a zone or a proxied stream that does not properly detect the connection closure, we might end up in a situation where a zone control plane restart goes unnoticed. This could result in multiple open streams from a single zone to the global control plane. Since the health check only verifies the zone name and tenant ID, a newly connected instance of the control plane makes all previous health check detectors pass.

This issue occurs when a gRPC client and server, both configured with Keepalive, communicate through a Gateway that is also configured with the gRPC protocol. We can reproduce it with the following setup:
```
Zone → Gateway (gRPC) → Global
```
By introducing packet loss on the Zone, we can observe the following log messages in the control plane:

`keepalive ping failed to receive ACK within timeout`
However, despite this failure, the connection remains open in the control plane, leading to unnecessary resource allocation. This suggests that the Gateway might be misconfigured. In such cases, we should implement safeguards to prevent hanging connections.

## Implementation information

We could track the last connection time by recording it in an event and later checking if the same client attempts to reconnect. If a reconnection occurs, we cancel the old stream.

To handle scenarios where multiple instances of the global control plane are running, we will store this information in zoneInsight. Additionally, we will use a periodic verification mechanism (zone watcher) to ensure the recorded time remains accurate.


pkg/kds/mux/zone_watch.go
```diff
@@ -110,11 +115,69 @@ func (zw *ZoneWatch) Start(stop <-chan struct{}) error {
...
 					})
 					delete(zw.zones, zone)
+					delete(zw.zoneStreams, zone)
+				}
+				// Now we want to check individual stream
+				for stream, connOpenTime := range zw.zoneStreams[zone] {
+					var conf *system_proto.KDSStream
+					switch stream {
+					case service.Clusters:
+						conf = zoneInsight.Spec.GetKdsStreams().GetClusters()
```
lobkovilya, line 125: nit: you can probably create helper func in `zone_insight_helper.go`

```go
func (x *ZoneInsight) GetKDSStream(streamType) *KDSStream 
```

and use it here and in `service.go`

lobkovilya (commented):
The PR makes sense. I'm a bit worried if it's going to be a mess when on each connection at least 6 concurrent processes are going to try to update the same ZoneInsight resource. If at least one of them fails others are going to be canceled (and retried on the client).

As it's potentially can be a regression I'd hide this behind the feature flag and disable by default.

pkg/kds/mux/zone_sync.go
```diff
@@ -144,18 +179,74 @@ func (g *KDSSyncServiceServer) ZoneToGlobalSync(stream mesh_proto.KDSSyncService
...
+	if err != nil {
+		return err
+	}
+
+	// Add delay for Upsert. If Global CP is behind an HTTP load balancer,
+	// it might be the case that each Envoy Admin stream will land on separate instance.
+	// In this case, all instances will try to update Zone Insight which will result in conflicts.
+	// Since it's unusual to immediately execute envoy admin rpcs after zone is connected, 0-10s delay should be fine.
+	// #nosec G404 - math rand is enough
+	time.Sleep(time.Duration(rand.Int31n(10000)) * time.Millisecond)
```
lahabana, line 223: Isn't the retry below just fine?

pkg/kds/service/server.go
```diff
@@ -235,16 +274,32 @@ func (g *GlobalKDSServiceServer) storeStreamConnection(ctx context.Context, zone
...
+		if stream == nil {
+			stream = &v1alpha1.KDSStream{}
+		}
+		if stream.GetConnectTime() == nil || proto.MustTimestampFromProto(stream.ConnectTime).Before(connectTime) {
+			stream.GlobalInstanceId = g.instanceID
+			stream.ConnectTime = proto.MustTimestampProto(connectTime)
 		}
 		switch rpcName {
 		case ConfigDumpRPC:
-			zoneInsight.Spec.EnvoyAdminStreams.ConfigDumpGlobalInstanceId = instance
```
lahabana, line 243: Shouldn't it set both during migration?

pkg/kds/mux/zone_sync.go
lukidzi: It's fine, but it doesn't solve the issue with the first request. It only improves the probability that the first request won't hit a conflict.

lobkovilya (approved):

Issue #12190, opened 2024-12-09
Set `internal_address_config` on HttpConnectionManager explicitly


jijiechen:
### Description

Envoy is warning when HCM does not have  explicit `internal_address_config` set and it will not consider any IP address as internal in a future release, so we need to adapte to this change before our next Envoy version bump.

This is considered a feature because it did not impact our existing versions. 

### Reproduce steps

1. download Kuma
2. Run the zone cp and a DP following the [quick start guide](https://kuma.io/docs/2.9.x/quickstart/universal-demo/)
3. check the sidecar log

### Actual result

A warning shows up:

```
[2024-12-09 06:12:49.716][20][warning][misc] [source/extensions/filters/network/http_connection_manager/config.cc:82] internal_address_config is not configured. The existing default behaviour will trust RFC1918 IP addresses, but this will be changed in next release. Please explictily config internal address config as the migration step.
```

### Expected result

No warning.

### More information

As the warning states, the HCM recognized [IP addresses defined in RFC1918](https://datatracker.ietf.org/doc/html/rfc1918#section-3) as internal by default, and this will be changed in a newer release. This will impact the behaviour of logging and header santination of Envoy.

This behaviour can be controlled by Envoy runtime feature flag
__envoy.reloadable_features.explicit_internal_address_config__
This feature flag defaults to `false` in all existing Envoy versions (1.29.11, 1.32.2), but the code has been changed to defaults to `true` on the `main` branch.

There is a related Envoy CVE:
* [CVE-2024-45806](https://github.com/envoyproxy/envoy/security/advisories/GHSA-ffhv-fvxq-r6mf): (CVSS Score 6.5, Moderate): Potential for x-envoy headers to be manipulated by external sources.

What is an internal address?
> [.InternalAddressConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-internaladdressconfig) Configures what network addresses are considered internal for stats and header sanitation purposes. If unspecified, only RFC1918 IP addresses will be considered internal. See the documentation for [x-envoy-internal](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#config-http-conn-man-headers-x-envoy-internal) for more information about internal/external addresses.
[...]

PR #12986, opened 2025-03-03
feat(xds): add internal address config onto HttpConnectionManager

jijiechen:
## Motivation

add internal address config onto HttpConnectionManager, more details are available in issue https://github.com/kumahq/kuma/issues/12190

## Implementation information

introducing a new configuration item `ipam.knownInternalCIDRs` on zonal CP to allow users specify their known internal address pool and we assign these values when generating Envoy config for HttpConnectionManager

## Supporting documentation

fixes https://github.com/kumahq/kuma/issues/12190


pkg/xds/envoy/listeners/filter_chain_configurers.go
```diff
@@ -86,11 +85,18 @@ func ServerSideStaticTLS(tlsCerts core_xds.ServerSideTLSCertPaths) FilterChainBu
 	})
 }
 
-func HttpConnectionManager(statsName string, forwardClientCertDetails bool) FilterChainBuilderOpt {
-	return AddFilterChainConfigurer(&v3.HttpConnectionManagerConfigurer{
+func HttpConnectionManager(statsName string, forwardClientCertDetails bool, annotators ...v3.HttpConnectionManagerConfigurerAnnotateFunc) FilterChainBuilderOpt {
+	hcmConfigurer := &v3.HttpConnectionManagerConfigurer{
```
jijiechen, line 89: Adding optional "annotator" functions to this Opt so that existing invocation can stay unchanged. Existing code on mutating the HttpConnectionManager by marshal+unmarshal the proto message could potentially benifit from this design too.

pkg/core/xds/types.go
```diff
@@ -188,6 +188,8 @@ type Proxy struct {
 	RuntimeExtensions map[string]interface{}
 	// Zone the zone the proxy is in
 	Zone string
+	// InternalAddresses is a set of address prefixes that are considered internal to the mesh, it will be configured to in Envoy HCM config
+	InternalAddresses []InternalAddress
```
jijiechen, line 192: Adding this field on the `Proxy` object so that the value can be accessed by the generators and plugins. `InternalAddress` carry values of the same category of existing data like `ApiVersion`

lukidzi (commented):
Can we add more test cases? I would like to see Gateway/ZoneEgress/ZoneIngress configuration change either by `outbound_proxy` generator, Mesh*Route. Also, what about `admin_proxy_generator`, `prometheus_endpoint_generator`?

pkg/xds/envoy/listeners/filter_chain_configurers.go
jijiechen: It's refactored, so no more annotators

jijiechen:
We are using Envoy `1.32` in 2.10, and Envoy 1.32  removed the default builtin internal addresses. So this PR is better to be merged in this 2.10 release.
Source: https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/network/http_connection_manager/config.cc#L83-L86

docs/generated/raw/kuma-cp.yaml
```diff
@@ -825,6 +825,15 @@ ipam:
     cidr: 243.0.0.0/8 # ENV: KUMA_IPAM_MESH_MULTI_ZONE_SERVICE_CIDR
   # Interval on which Kuma will allocate new IPs for MeshServices and MeshExternalServices
   allocationInterval: 5s # ENV: KUMA_IPAM_ALLOCATION_INTERVAL
+  # KnownInternalCIDRs contains a list of CIDRs which are considered internal and trusted by Envoy when sending HTTP requests
+  knownInternalCIDRs: # ENV: KUMA_IPAM_ALLOCATION_INTERVAL
```
lukidzi, line 829: Should we include our CIDRs for MeshService, MeshExternalService, and MeshMultiZoneService, as well as possibly 240.0.0.0? I'm not entirely sure when the address IP is validated, but if the outbound listener is bound to one of these IPs and validation happens at this step, we might miss some data.

jijiechen: You mean to include these addresses automatically whenever they were set, right?
I think we can do it, yes

jijiechen: Also, the environment var name is incorrect in this comment

lukidzi: Yes, we could add them automatically.

jijiechen: The `isInternalRequest` is used to check if a request is originated from an internal client, so it's not necessary to include these Mesh*Service CIDRs since they are only used as destinations.
https://github.com/envoyproxy/envoy/blob/main/source/common/http/conn_manager_utility.cc#L222-L251

Envoy attaches a set up request headers when forwarding requests to the upstream. It removes "internal only headers" from requests originated from external clients, so that, in this case, the upstream server only receives trusted/sanitized headers.

lukidzi (approved):
LGTM

PR #13003, opened 2025-03-04
feat(resources): add deprecation for resources whose name breaks RFC-1035

Icarus9913:
## Motivation

Before adding the validation for the Mesh, Zone, MeshService resources to validate whether the name breaks RFC-1035, I supplement the deprecation for them.

## Implementation information

Supplement the `Deprecations` method for those resources

## Supporting documentation

Notice that the difference between [RFC-1123](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) and [RFC-1035](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names) is that RFC-1035 only allow "start with an alphabetic character" which is followed by kubernetes `Service` resource


Icarus9913:
In the future we gonna align with kubernetes to use `RFC-1035`

pkg/core/resources/apis/meshservice/api/v1alpha1/deprecated.go
```diff
@@ -0,0 +1,21 @@
+package v1alpha1
```
lahabana, line 1: Isn't meshmultizoneservice and meshexternalservice supposed to have this aswell?

lahabana (dismissed):
This needs test cases in https://github.com/kumahq/kuma/tree/master/pkg/api-server/testdata/resources/crud to show that the deprecation is actually working.

It also needs an update in UPGRADE.md

UPGRADE.md
```diff
@@ -24,9 +24,11 @@ New rule:
 
 Before the upgrade ensure that your resources don't use unsupported characters.
 
+Further, starting from version `2.10.x` we are deprecating the usage of non [RFC 1035](https://www.rfc-editor.org/rfc/rfc1035.html) characters for names of `Mesh`, `Zone`, `MeshService`, `MeshExternalService`, `MeshMultizoneService` resources. These names will not be accepted in a future release.
```
lahabana, line 27: In the rule above there are no character limit?
```suggestion
Starting from version `2.10.x` we are deprecating the usage of non [RFC 1035](https://www.rfc-editor.org/rfc/rfc1035.html) characters for names of `Mesh`, `Zone`, `MeshService`, `MeshExternalService`, `MeshMultizoneService` resources. These names will be rejected in the future.
```

pkg/core/resources/apis/mesh/validators.go
```diff
@@ -234,14 +239,14 @@ func ValidateHostname(path validators.PathBuilder, hostname string) validators.V
 	}
 
 	if strings.HasPrefix(hostname, "*.") {
-		if !DomainRegexp.MatchString(strings.TrimPrefix(hostname, "*.")) {
+		if !DNS1123SubdomainRegexp.MatchString(strings.TrimPrefix(hostname, "*.")) {
```
lahabana, line 242: Any reason not to use the regexp already in the k8s validaiton package?

lahabana (commented):
I'd love to have someone like @lobkovilya review this. I'm a little concerned if this may cause issues with KDS and the way we create names when syncing things from and to global.

pkg/core/resources/apis/mesh/validators.go
Icarus9913: They are private in k8s

pkg/core/resources/apis/mesh/validators.go
```diff
@@ -234,14 +239,14 @@ func ValidateHostname(path validators.PathBuilder, hostname string) validators.V
...
 			err.AddViolationAt(path, "invalid wildcard domain")
 		}
 
 		return err
 	}
 
-	if !DomainRegexp.MatchString(hostname) {
+	if !DNS1123SubdomainRegexp.MatchString(hostname) {
```
lobkovilya, line 249: The PR description mentions only adding deprecations, but this file contains an actual change in validation. What's the motivation to change the way MeshGateway validates hostnames?

lobkovilya:
> I'd love to have someone like @lobkovilya review this. I'm a little concerned if this may cause issues with KDS and the way we create names when syncing things from and to global.

Looks like there is indeed a problem with KDS hash suffixes, but seems like it exists without this PR. When adding hash-suffix we expect the result length to be less or equal to 253 https://github.com/kumahq/kuma/blob/b2b6262ea5863105816ae1651642f5f5fb95fc3e/pkg/kds/hash/hash.go#L16-L21

So syncing of MS that's longer than 46 chars (63 - 17) is going to result in validation error

pkg/core/resources/apis/mesh/validators.go
Icarus9913: I gonna move it to another PR

lobkovilya (approved):
LGTM

Issue #12990, opened 2025-03-03
api-lint seems to not go into targetRef


slonka:
### Kuma Version

latest

### Describe the bug

The linter seems to not go into `targetRef`. This field does not match any of our predefined fields:

https://github.com/kumahq/kuma/blob/173caaf4e9e50054f169c3bb801a440ba84c1bf6/api/common/v1alpha1/ref.go#L102-L103

This could be because it's imported from another file. The most problematic side-effect is that MeshHTTPRoute.targetRef.proxyTypes is causing a 400 because TF defaults to [] and we have validation that prevents having anything there.

### Expected behavior

It should lint `targetRef`.

PR #8868, opened 2024-01-15
feat(policy): add possibility to target only gateways/sidecars

lukidzi:
### Checklist prior to review

Implementation of: https://github.com/kumahq/kuma/pull/8829

- [ ] [Link to relevant issue][1] as well as docs and UI issues --
- [ ] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [ ] Tests (Unit test, E2E tests, manual test on universal and k8s) --
  - Don't forget `ci/` labels to run additional/fewer tests
- [ ] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [ ] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label) --

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #13004, opened 2025-03-04
fix(api): targetRef linting

slonka:
## Motivation

Best described in the issue https://github.com/kumahq/kuma/issues/12990. The most problematic field is `proxyTypes` because it has conflicting validation - minItems = 1 and `err.Add(disallowedField("proxyTypes", ref.ProxyTypes, ref.Kind))` (for `MeshService`, `MeshHTTPRoute`).

## Implementation information

Modify the structure to pass linter rules.

## Supporting documentation

Closes https://github.com/kumahq/kuma/issues/12990


api/common/v1alpha1/targetref.go
```diff
@@ -88,28 +89,27 @@ type TargetRef struct {
...
 	// Tags used to select a subset of proxies by tags. Can only be used with kinds
 	// `MeshSubset` and `MeshServiceSubset`
-	Tags map[string]string `json:"tags,omitempty"`
+	Tags *map[string]string `json:"tags,omitempty"`
 	// Mesh is reserved for future use to identify cross mesh resources.
-	Mesh string `json:"mesh,omitempty"`
+	Mesh *string `json:"mesh,omitempty"`
 	// ProxyTypes specifies the data plane types that are subject to the policy. When not specified,
 	// all data plane types are targeted by the policy.
-	// +kubebuilder:validation:MinItems=1
```
slonka, line 102: @lukidzi does this need to be validated in code for `Mesh` and `MeshSubset`?

slonka: this is your original PR: https://github.com/kumahq/kuma/pull/8868/files#r1457386523

lukidzi: Yes, it makes sense only for this since when you select something more specific you won't have a mix of sidecar/gateway

lukidzi: If there is a value defined, it has have at least one value, and just omit when no variable

slonka: added validation for these two types

slonka:
We need to figure out if this needs any migration steps.

lobkovilya (approved):

PR #13008, opened 2025-03-04
feat(kds): add support for kuma.io/kds-sync label

### LINKED ISSUE #12916: create label to disable KDS sync
### Description

Atm we don't have a way for 1 resource to not be synced using KDS.

This would be good for some secrets (token signing keys).
It would avoid having to do some special cases directly in the code. We'd simply create the policy with the right label.

lahabana:
## Motivation

There may be cases where either system or user may want to not sync a specific resource with KDS.
Until now this wasn't really possible. So we ended having specific hard coded config types.

This allows disabling KDS for specific resources by adding the label `kuma.io/kds-sync: disabled`

## Implementation information

We add the filter as a check in both filters.
Also added functional testing around the whole KDS.
You can create a set of files of starting state and save end state.
This should make it easy to add a batch of non regression and exploratory tests to KDS.

## Supporting documentation

Fix #12916

### DIFF
--- a/api/mesh/v1alpha1/dataplane_helpers.go
+++ b/api/mesh/v1alpha1/dataplane_helpers.go
@@ -16,6 +16,9 @@ const (
 	KubeNamespaceTag = "k8s.kuma.io/namespace"
 	KubeServiceTag   = "k8s.kuma.io/service-name"
 	KubePortTag      = "k8s.kuma.io/service-port"
+	// KDSSyncLabel a label that controls properties of the KDS sync.
+	// currently only disabled/enabled is supported
+	KDSSyncLabel = "kuma.io/kds-sync"
 )
 
 const (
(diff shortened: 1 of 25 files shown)


api/mesh/v1alpha1/dataplane_helpers.go
```diff
@@ -16,6 +16,9 @@ const (
 	KubeNamespaceTag = "k8s.kuma.io/namespace"
 	KubeServiceTag   = "k8s.kuma.io/service-name"
 	KubePortTag      = "k8s.kuma.io/service-port"
+	// KDSSyncLabel a label that controls properties of the KDS sync.
+	// currently only disabled/enabled is supported
+	KDSSyncLabel = "kuma.io/kds-sync"
```
lukidzi, line 21: Don't we need a validator that forbids setting this label for some resources? Example: `Mesh` object

lahabana: We could though I think this label is niche enough that people using it should know what they are doing.

Also you could imagine having a mesh on global and don't want zones to get it yet.

lukidzi (approved):

lukidzi (approved):

Issue #12653, opened 2025-01-23
Syncing MeshService from Universal with underscores to Kubernetes cluster


jakubdyszkiewicz:
### Kuma Version

2.9.x

### Describe the bug

I cannot sync MeshService from Universal to Kubernetes cluster
```
2025-01-23T09:31:12.592Z        ERROR   kds-mux-client  rpc stream failed prematurely, will restart in background       {"client-id": "dev-eu-west-2-verifone", "rpc": "global-to-zone", "error": "GlobalToZoneSyncClient finished with an error: failed to store MeshService resources: failed to create k8s resource: MeshService.kuma.io \"echo-service_default_8082-x6c24z4bdv2286v8\" is invalid: metadata.name: Invalid value: \"echo-service_default_8082-x6c24z4bdv2286v8\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')", "errorVerbose": "MeshService.kuma.io \"echo-service_default_8082-x6c24z4bdv2286v8\" is invalid: metadata.name: Invalid value: \"echo-service_default_8082-x6c24z4bdv2286v8\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

### To Reproduce

1. Setup multizone env with one Zone on Uni and one Zone on Kube
2. Create DPP with `kuma.io/service` of `echo-service_default_8082`
3. Enable `Exclusive` mode of MeshServices in the mesh
4. MeshService of `echo-service_default_8082` should be generated, synced to global and then fails to be synced on zone

### Expected behavior

I should be able to sync this. Ideally we should not allow `_` for `kuma.io/service`, but I think we need to support this for backwards compatibility

### Additional context (optional)

We could for example replace `_` with `-` because the name is hashed when synced to another CP. We should not get conflicts this way and we can still rely on display name

lukidzi:
This shouldn't be the case. We should remove deprecated code that should be removed in 2.7.x

[...]

PR #13014, opened 2025-03-05
fix(meshservice): enhance the validation of kuma.io/service name for meshservice generation

Icarus9913:
## Motivation

Refer to PR #12751, we need to enhance the kuma.io/service name validation rules used by kubernetes Service resource.
The MeshService resource can be automatically created in universal mode, and the name comes from the value of Dataplane label `kuma.io/service`. 

However, the name of MeshService can be used as part of DNS hostname. If you use underscore, dot sign, it would breaks the DNS Label format.

So, we gonna use RFC-1035-Label-Names rule

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12653


pkg/core/resources/apis/mesh/validators.go
```diff
@@ -18,14 +18,19 @@ import (
...
 
 var (
-	NameCharacterSet     = regexp.MustCompile(`^[0-9a-z.\-_]*$`)
-	DomainRegexp         = regexp.MustCompile("^" + dnsLabel + "(\\." + dnsLabel + ")*" + "$")
-	tagNameCharacterSet  = regexp.MustCompile(`^[a-zA-Z0-9.\-_:/]*$`)
-	tagValueCharacterSet = regexp.MustCompile(`^[a-zA-Z0-9.\-_:]*$`)
-	selectorCharacterSet = regexp.MustCompile(`^([a-zA-Z0-9.\-_:/]*|\*)$`)
+	NameCharacterSet       = regexp.MustCompile(`^[0-9a-z.\-_]*$`)
+	DNS1123SubdomainRegexp = regexp.MustCompile("^" + dns1123SubdomainFmt + "$")
+	DNS1035LabelRegexp     = regexp.MustCompile("^" + dns1035LabelFmt + "$")
```
Icarus9913, line 30: We can't import Kubernetes variable because they're private

Icarus9913:
We still need some time to think about the length problem carefully

Icarus9913:
We decide to use the Kubernetes `NameIsDNS1035Label` validation function directly to verify the value of `label kuma.io/service`.

And for the length restriction, we'll align with the MeshService resource creation validation <=63 length

Issue #13023, opened 2025-03-06
Name validator checks the name of kuma resource on k8s


lukidzi:
### Kuma Version

2.10.0-preview

### Describe the bug

When we enabled mesh service on k8s we noticed a warning logs 
```
│ 2025-03-06T10:38:16.054Z    INFO    kds-delta-global    creating a new resource from upstream    {"type": "MeshService", "name": "redis-84x98dx244v942z9.kong-mesh-system", "mesh": "default"}                                                                                                                                                                                   │
│ 2025-03-06T10:38:16.085Z    INFO    KubeAPIWarningLogger    Name that doesn't conform DNS (RFC 1035) format is deprecated: a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

which shouldn't be displayed since the resource has the correct name `redis-84x98dx244v942z9`. The problem is the we validate kuma resource name with a namespace included and not just a resource (k8s resource) name.

We should also improve the message to include the invalid resource name

### To Reproduce

1. Install kuma on k8s
2. Deploy demo app
3. Enable meshservice

### Expected behavior

* We shouldn't log the message in the case of the resource
* Improve log message to include invalid resource name

lukidzi:
fixed by https://github.com/kumahq/kuma/pull/13028

PR #13028, opened 2025-03-06
fix(resource): use kuma.io/display-name instead of name

lukidzi:
## Motivation

When running on k8s `r.GetMeta().GetName()` returns namespaces name which doesn't pass validation. We shouldn't use this value but instead `GetDisplayName()`

## Implementation information

Check if `kuma.io/display-name` fulfill the requirement.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13023

### DIFF
--- a/pkg/core/resources/apis/mesh/deprecated.go
+++ b/pkg/core/resources/apis/mesh/deprecated.go
@@ -5,15 +5,18 @@ import (
 	"strings"
 
 	apimachineryvalidation "k8s.io/apimachinery/pkg/api/validation"
+
+	"github.com/kumahq/kuma/pkg/core/resources/model"
 )
 
 func (t *MeshResource) Deprecations() []string {
 	var deprecations []string
 
-	allErrs := apimachineryvalidation.NameIsDNS1035Label(t.GetMeta().GetName(), false)
+	name := model.GetDisplayName(t.GetMeta())
+	allErrs := apimachineryvalidation.NameIsDNS1035Label(name, false)
 	if len(allErrs) != 0 {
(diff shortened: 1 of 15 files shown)


pkg/core/resources/apis/mesh/deprecated.go
```diff
@@ -5,15 +5,18 @@ import (
...
 	var deprecations []string
 
-	allErrs := apimachineryvalidation.NameIsDNS1035Label(t.GetMeta().GetName(), false)
+	name := model.GetDisplayName(t.GetMeta())
+	allErrs := apimachineryvalidation.NameIsDNS1035Label(name, false)
 	if len(allErrs) != 0 {
-		nameDeprecationMsg := fmt.Sprintf("Name that doesn't conform DNS (RFC 1035) format is deprecated: %s",
-			strings.Join(allErrs, "; "))
+		nameDeprecationMsg := fmt.Sprintf("Invalid name: '%s'. It does not conform to the DNS format (RFC 1035). This is deprecated. Errors: %s",
+			name, strings.Join(allErrs, "; "))
```
Icarus9913, line 19: What about adding the resource type?

```suggestion
		nameDeprecationMsg := fmt.Sprintf("Invalid %s resource name: '%s'. It does not conform to the DNS format (RFC 1035). This is deprecated. Errors: %s",
			MeshResourceTypeDescriptor.Name, name, strings.Join(allErrs, "; "))
```

Icarus9913 (approved):

Icarus9913 (approved):

PR #13033, opened 2025-03-07
test(e2e): improve universal app run

### LINKED ISSUE #13123: Debug package from failed test doesn't have logs
### Kuma Version

mater

### Describe the bug

I was checking flaky test and the package had no logs of control-plane
In case of gateway api there is no debug package at all https://github.com/kumahq/kuma/actions/runs/13824998229

[e2e-debug-universal-arm64-kind---0 (4).zip](https://github.com/user-attachments/files/19289264/e2e-debug-universal-arm64-kind---0.4.zip)

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

lahabana:
## Motivation

It was very hard to have a complete and readable state of things in the test report because we were executing one of ssh cmds in different places.
We now keep a persistent ssh session for the whole time of the app which makes things easier to track and the output easier to consume.

## Implementation information

the ssh package was completely redesigned to keep an ssh session rather than forking to execute `ssh` commands. This makes things easier to follow, easier to write lengthy scripts without caring about escaping things correctly for bash.
Once this was done we could start a CP or run a DP with a single cmdLine instead of a list of complex calls.

This makes the report a lot easier as it avoids a lot of noisy stuff.

Other minor changes:
- Stop doing retries inside the framework. Either we should retry because it's setup or simply in an eventually. Also stop using the retry from terratest as it's very verbose.
- ignore for dockerfiles were including things that were not necessary, this made the building of images longer.
- Once this change was done we could run into clock skews with tokens. So moved the token issuance date a few seconds in the past

## Supporting documentation

Fix #13123


slonka:
Did you mean "fix https://github.com/kumahq/kuma/issues/13123"?

test/framework/universal_app.go
```diff
@@ -220,11 +226,7 @@ func NewUniversalApp(t testing.TestingT, clusterName, dpName, mesh string, mode
 	app.allocatePublicPortsFor("22")
 
 	if mode == AppModeCP {
-		app.allocatePublicPortsFor("5678", "5680", "5681", "5682", "5685", "9901")
-	}
-
-	if dpName == AppEgress {
```
slonka, line 226: So will Egress not have anything exposed? 🤔

test/framework/universal_app.go
```diff
@@ -220,11 +226,7 @@ func NewUniversalApp(t testing.TestingT, clusterName, dpName, mesh string, mode
...
-		app.allocatePublicPortsFor("9901")
```
slonka, line 227: I think this was useful when debugging right? You can't allocate new ports for an already running container. I know you can always exec and curl but it's a bit easier when it's exposed 🤷

slonka (dismissed):
in generat 👍 - a copule of Qs

test/dockerfiles/universal.dockerignore
```diff
@@ -11,12 +11,6 @@
 !build/artifacts-linux-arm64/kumactl/kumactl
 !build/artifacts-linux-arm64/coredns/coredns
 !build/artifacts-linux-arm64/test-server/test-server
-!build/artifacts-darwin-arm64/kuma-cp/kuma-cp
-!build/artifacts-darwin-arm64/kuma-dp/kuma-dp
-!build/artifacts-darwin-arm64/envoy/*
-!build/artifacts-darwin-arm64/kumactl/kumactl
-!build/artifacts-darwin-arm64/coredns/coredns
-!build/artifacts-darwin-arm64/test-server/test-server
```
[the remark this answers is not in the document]
lahabana: When building images you load everything in your `cwd` in the builder context (that's the infamous: ` => => transferring context: 13.55kB` line when you build).

The more stuff you have the longer it is. If you look here what I remove is the _darwin_ binaries because we don't need _darwin_ binaries when building _Linux_ images. Ideally we'd avoid mounting the binaries for the other arch but dockerignore is too simple to do these conditions so I didn't bother.

test/e2e_env/universal/resilience/resilience.go
```diff
@@ -50,12 +50,13 @@ func ResilienceUniversal() {
...
 
 		// when Kuma CP is killed
-		_, _, err := universal.Exec("", "", AppModeCP, "pkill", "-9", "kuma-cp")
-		Expect(err).ToNot(HaveOccurred())
+		Expect(g.Kill(AppModeCP, "kuma-cp run")).To(Succeed())
 
-		// and DPP is killed while Kuma CP is down
-		_, _, err = universal.Exec("", "", AppModeDemoClient, "pkill", "-9", "envoy")
+		out, _, err := g.Exec("", "", AppModeDemoClient, "ps", "aux")
+		Logf("ps aux: %q", out)
```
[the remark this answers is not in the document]
lahabana: Yep this is done in `g.Kill`

test/framework/docker.go
```diff
@@ -27,22 +23,22 @@ func GetPublishedDockerPorts(
...
 		var err error
 		// Sometimes the port may not be available immediately, and it can take some time.
 		// Since we didn't retry, tests were failing with and an error
 		// `missing port in address` on OSX.
-		retry.DoWithRetry(t, "get port", retries, retryTimeout, func() (string, error) {
+		for i := 0; i < 10; i++ {
 			out, err = shell.RunCommandAndGetStdOutE(t, cmd)
 			if err != nil {
-				return "", err
+				time.Sleep(time.Millisecond * 500)
```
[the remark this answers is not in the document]
lahabana: I actually disagree, if this were repeated a few yes. Here it's used once and very specific to this 1 use case. I think there's more value in understanding everything without any scrolling that useless constants.

test/framework/envs/multizone/env.go
```diff
@@ -178,7 +167,7 @@ func SetupAndGetState() []byte {
 			MADS:        KubeZone2.GetKuma().(*K8sControlPlane).MadsPortFwd(),
 		},
 	}
-	bytes, err := json.Marshal(state)
+	bytes, err := json.Marshal(state) // nolint:govet
```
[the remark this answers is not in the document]
lahabana: state as a mutex. govet complains you try to marshall something with a mutex

test/framework/universal_app.go
lahabana: For dps we have this whole tunnel story: https://github.com/kumahq/kuma/blob/f85020df3d268bde3596cd8e5cfcdcc2cd4eefdb/test/framework/envoy_admin/tunnel/universal.go#L16 that runs the curls from inside the container.

So this was never actually needed (plus not very useful because we bind admin to localhost:9901 so the port-forward wouldn't work...)

test/framework/universal_app.go
lahabana: https://github.com/kumahq/kuma/pull/13033#discussion_r2005635536

slonka (approved):

test/e2e_env/multizone/externalservices/externalservices_multizone_universal.go
```diff
@@ -155,42 +155,42 @@ routing:
...
 		Eventually(func(g Gomega) {
-			stdout, _, err := client.CollectResponse(
+			stdout, stderr, err := client.CollectResponse(
 				zone1, "demo-client", "external-service-1.mesh",
 				client.WithVerbose(),
 			)
 			g.Expect(err).ToNot(HaveOccurred())
-			g.Expect(stdout).To(ContainSubstring("HTTP/1.1 200 OK"))
+			g.Expect(stderr).To(ContainSubstring("HTTP/1.1 200 OK"))
 			g.Expect(stdout).ToNot(ContainSubstring("HTTPS"))
```
slonka, line 164: @lahabana don't we need to change this line as well, is everything now on `stderr`?

PR #13051, opened 2025-03-11
fix(kuma-cp): change usage of deprecated global_downstream_max_connections

lukidzi:
## Motivation

We can see in the logs
```
Usage of the deprecated runtime key overload.global_downstream_max_connections, consider switching to `envoy.resource_monitors.global_downstream_max_connections` instead.This runtime key will be removed in future.
```

the field was deprecated.

## Implementation information

Changed to `envoy.resource_monitors.global_downstream_max_connections`


Icarus9913 (approved):
LGTM

jijiechen (approved):

PR #13060, opened 2025-03-11
added resource details in deprecation message

### LINKED ISSUE #13030: Add the resource type and name as deprecation title
### Description

We saw the following logs in our control plane, and they are not clear to us.

```text
2025-03-07T02:59:56.261Z	INFO	KubeAPIWarningLogger	'from' field is deprecated, use 'rules' instead
---
2025-03-06T10:38:16.085Z    INFO    KubeAPIWarningLogger    Name that doesn't conform DNS (RFC 1035) format is deprecated: a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

I recommend adding the resource type and name as the first line.
Here: https://github.com/kumahq/kuma/blob/b7a4b80a24d1e4fd977f5843b5a212931909abb9/pkg/core/resources/model/resource.go#L127-L132

Maybe we can do something like:

```golang
func Deprecations(resource Resource) []string {
	if v, ok := interface{}(resource).(interface{ Deprecations() []string }); ok {
		deprecations := []string{fmt.Sprintf("%s type resource '%s' has the following fields deprecations: ",
			resource.Descriptor().Name, resource.GetMeta().GetName())}
		return append(deprecations, v.Deprecations()...)
	}
	return nil
}
```

gnana997:
## Motivation

#13030 : Helps user debugging in the usage of deprecated resources

## Implementation information

Added resource name and resource meta information along with the deprecation resource errors

## Supporting documentation

Fix #13030


pkg/core/resources/model/resource.go
```diff
@@ -131,7 +131,9 @@ func Validate(resource Resource) error {
 
 func Deprecations(resource Resource) []string {
 	if v, ok := interface{}(resource).(interface{ Deprecations() []string }); ok {
-		return v.Deprecations()
+		deprecations := []string{fmt.Sprintf("%s type resource '%s' has the following fields deprecations: ",
```
lahabana, line 134:
```suggestion
		deprecations := []string{fmt.Sprintf("%s type resource %q has the following fields deprecations: ",
```

`%q` adds quotes in a safe way.

gnana997: Hi @lahabana , I have updated the message.
Thanks a lot!

bartsmykla (dismissed):
lgtm

Icarus9913:
@gnana997 Could you fix the unit test?
The command `make test UPDATE_GOLDEN_FILES=true` may help you

bartsmykla:
@gnana997 after new golden files were generated via `make test UPDATE_GOLDEN_FILES=true` please commit and push them

gnana997:
> @gnana997 after new golden files were generated via `make test UPDATE_GOLDEN_FILES=true` please commit and push them

Oh ok. Got it. I have pushed the changes.
Thanks!

pkg/api-server/testdata/resources/crud/put_meshaccesslog_00.golden.json
```diff
@@ -1 +1,5 @@
-{}
+{
+ "warnings": [
+  "MeshAccessLog type resource \"my-access-log\" has the following fields deprecations: "
+ ]
+}
```
Icarus9913, line 5: This is the unexpected output that we need to figure out

Icarus9913 (changes requested):

lahabana:
It's a little odd to return the resource name in the deprecation message. What are we trying to achieve here? Readability in kumactl apply?

If that's the case adding the resource type and name in the response should be done in kumactl not in the response of the api

lahabana:
It's a little odd to return the resource name in the deprecation message. What are we trying to achieve here? Readability in kumactl apply?

If that's the case adding the resource type and name in the response should be done in kumactl not in the response of the api

Icarus9913:
> It's a little odd to return the resource name in the deprecation message.

Because I noticed that the `Deprecation` function is also called in the CP, and the corresponding log lost subject object. Here's the logs:

```text
2025-03-07T02:59:56.261Z	INFO	KubeAPIWarningLogger	'from' field is deprecated, use 'rules' instead
---
2025-03-06T10:38:16.085Z    INFO    KubeAPIWarningLogger    Name that doesn't conform DNS (RFC 1035) format is deprecated: a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

pkg/api-server/testdata/resources/crud/put_meshaccesslog_00.golden.json
gnana997: Hi @Icarus9913 , I have refactored the deprecation message to add the messages with the fields and what to use too. Let me know If I need to update.
Thanks!

pkg/api-server/testdata/resources/crud/put_meshaccesslog_00.golden.json
```diff
@@ -1 +1,5 @@
-{}
+{
+ "warnings": [
+  "MeshAccessLog type resource \"my-access-log\" has the following deprecated fields: "
+ ]
+}
```
Icarus9913, line 5: @gnana997 Hi, you can see there are no deprecation details in this sentence, as I said before. It's an obvious unexpected output, and we need to figure out why. I'll have a look at it later.

Icarus9913 (changes requested):

pkg/core/resources/model/resource.go
```diff
@@ -131,7 +131,9 @@ func Validate(resource Resource) error {
 
 func Deprecations(resource Resource) []string {
 	if v, ok := interface{}(resource).(interface{ Deprecations() []string }); ok {
-		return v.Deprecations()
+		deprecations := []string{fmt.Sprintf("%s type resource %q has the following deprecated fields: %s",
+			resource.Descriptor().Name, resource.GetMeta().GetName(), strings.Join(v.Deprecations(), ", "))}
+		return append(deprecations, v.Deprecations()...)
```
Icarus9913, line 136:
```suggestion
		deprecations := v.Deprecations()
		if len(deprecations) == 0 {
			return nil
		}

		deprecationTitle := []string{fmt.Sprintf("'%s' type resource '%q' has the following deprecated fields: ",
			resource.Descriptor().Name, resource.GetMeta().GetName())}
		return append(deprecationTitle, strings.Join(deprecations, ";\n"))
```

Icarus9913 (commented):
@gnana997 Please apply my new changes, then update the golden files again.

lahabana (changes requested):
We need to get background on why KubeAPIWarningLogger doesn't add context. Maybe there's a workaround that's nicer. Worth looking at how other projects to deprecations

Automaat (approved):

PR #13104, opened 2025-03-13
feat(k8s): don't watch configMap/secrets in all namespaces

lukidzi:
## Motivation

For security reasons, some organizations prefer that other components do not watch Secrets and ConfigMaps across all namespaces.

## Implementation information

* Moved RBAC permissions for ConfigMaps from ClusterRole to Role in the system namespace.
* Configured the controller-manager cache to watch only ConfigMaps from the system namespace.
* Configured the controller-manager cache to watch only Secrets from the system namespace, unless `supportGatewaySecretsInAllNamespaces` is defined.
* The check from the controller has been removed, as watching only the system namespace ensures that no events are produced for other namespaces.

### DIFF
--- a/pkg/kds/context/testdata/full_sync/mesh-only/global.golden.yaml
+++ b/pkg/kds/context/testdata/full_sync/mesh-only/global.golden.yaml
@@ -10,14 +10,14 @@ type: Mesh
 creationTime: "0001-01-01T00:00:00Z"
 enabled: true
 modificationTime: "0001-01-01T00:00:00Z"
-name: zone-2
+name: zone-1
 type: Zone
 
 ---
 creationTime: "0001-01-01T00:00:00Z"
 enabled: true
 modificationTime: "0001-01-01T00:00:00Z"
-name: zone-1
+name: zone-2
 type: Zone
 
 # ZoneInsight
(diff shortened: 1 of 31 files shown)


pkg/kds/context/testdata/full_sync/mesh-only/global.golden.yaml
```diff
@@ -10,14 +10,14 @@ type: Mesh
 creationTime: "0001-01-01T00:00:00Z"
 enabled: true
 modificationTime: "0001-01-01T00:00:00Z"
-name: zone-2
```
lukidzi, line 13: Shouldn't be here

lukidzi: https://github.com/kumahq/kuma/pull/13105

bartsmykla (changes requested):
I accidentally gave @lukidzi the wrong information about where the new ConfigMap with transparent proxy settings is stored. I assumed he was only asking about the default behavior and said it’s always in `kuma-system`. However, the user can specify a different ConfigMap via annotation, in which case we first check the workload’s namespace, and if it’s not there, we fall back to `kuma-system`.

slonka (approved):
Updated branch so that unrelated golden file does not show up. LGTM

bartsmykla (approved):

PR #13105, opened 2025-03-13
test(unit): compare file content instead of yaml

lukidzi:
## Motivation

While debugging a flake I noticed that `MatchGoldenYAML` doesn't work correctly. In the match, it compares only the first element of the file. 

https://pkg.go.dev/gopkg.in/yaml.v3#section-readme
> Multi-document unmarshalling is not yet implemented, and base-60 floats from YAML 1.1 are purposefully not supported since they're a poor design and are gone in YAML 1.2.

## Implementation information

Changed to `MatchGoldenEqual` since we are using it in other tests. Also, change sort to `strings.Cmp`

> Changelog: skip

### DIFF
--- a/pkg/kds/context/full_sync_test.go
+++ b/pkg/kds/context/full_sync_test.go
@@ -36,7 +36,7 @@ var _ = Describe("Full sync tests", func() {
 		for _, file := range files {
 			if strings.HasSuffix(file.Name(), ".input.yaml") {
 				zoneName := strings.TrimSuffix(file.Name(), ".input.yaml")
-				resourceStore := memory.NewStore()
+				resourceStore := store.NewPaginationStore(memory.NewStore())
 				fullPath := path.Join(folder, file.Name())
 				Expect(test_store.LoadResourcesFromFile(ctx, resourceStore, fullPath)).To(Succeed())
 				zones[zoneName] = resourceStore
@@ -94,7 +94,7 @@ var _ = Describe("Full sync tests", func() {
 		for zoneName, zoneStore := range zones {
 			out, err := test_store.ExtractResources(ctx, zoneStore)
 			Expect(err).To(Succeed())
-			Expect(out).To(matchers.MatchGoldenYAML(folder, zoneName+".golden.yaml"), "zone %s", zoneName)
+			Expect(out).To(matchers.MatchGoldenEqual(folder, zoneName+".golden.yaml"), "zone %s", zoneName)
 		}
 	}, test.EntriesAsFolder("full_sync"))
 })
--- a/pkg/test/store/load.go
+++ b/pkg/test/store/load.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"regexp"
 	"slices"
+	"strings"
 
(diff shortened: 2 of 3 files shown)


pkg/kds/context/full_sync_test.go
```diff
@@ -36,7 +36,7 @@ var _ = Describe("Full sync tests", func() {
 		for _, file := range files {
 			if strings.HasSuffix(file.Name(), ".input.yaml") {
 				zoneName := strings.TrimSuffix(file.Name(), ".input.yaml")
-				resourceStore := memory.NewStore()
+				resourceStore := store.NewPaginationStore(memory.NewStore())
```
lobkovilya, line 39: Do we need this for resource ordering? Or what's the motivation to wrap this?

pkg/test/store/load.go
```diff
@@ -73,7 +74,7 @@ func ExtractResources(ctx context.Context, rs store.ResourceStore) (string, erro
 		items := slices.SortedFunc(
 			slices.Values(resList.GetItems()),
 			func(a, b model.Resource) int {
-				return cmp.Compare(a.GetMeta().GetName(), b.GetMeta().GetName())
+				return strings.Compare(a.GetMeta().GetName(), b.GetMeta().GetName())
```
lobkovilya, line 77: is there a difference?

lobkovilya (commented):
image


We have a lot of usages of `MatchGoldenYAML`, do we need an issue to revisit all these places?

pkg/kds/context/full_sync_test.go
lukidzi: For resource ordering, we do `.List(..., Ordered)` and without this we won't returned

pkg/test/store/load.go
lukidzi: I wasn't sure and since i did a lot of runs locally with this setup I left it

lukidzi:
> image
> We have a lot of usages of `MatchGoldenYAML`, do we need an issue to revisit all these places?

We should take a look

pkg/test/store/load.go
lahabana: Above uses cmp

lahabana:
Should MatchGoldenYAML simply rely on resource splitting. 

We have a function for this:

https://github.com/kumahq/kuma/blob/7075f0e17466c73ec82aa95649124edad6cafe8d/pkg/util/yaml/split.go#L14-L28

Can't we do something with this?


My suspicion is that most usages are fine as they compare a single doc systematically.

lukidzi:
> Should MatchGoldenYAML simply rely on resource splitting.
> 
> We have a function for this:
> 
> https://github.com/kumahq/kuma/blob/7075f0e17466c73ec82aa95649124edad6cafe8d/pkg/util/yaml/split.go#L14-L28
> 
> Can't we do something with this?
> 
> My suspicion is that most usages are fine as they compare a single doc systematically.

It's not that simple. Since the golden file we read in the gomega library, we have no control on splitting it

lahabana (approved):

Issue #12042, opened 2024-11-14
MeshService status Unavailable but is reachable


schogges:
### What happened?

After I deployed my services as k8s `Service` and configured the `Mesh` with `meshServices.mode: "Exclusive"`, my services became `MeshService` and went into state `Unavailable` (check for `spec.state: Unavailable` in examplary MeshService YAML Dump below). When sending requests I get the expected responses, so the services are actually available.

<details>
<summary>Deployment Configs</summary>

```yaml
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: default
spec:
  mtls:
    enabledBackend: ca-1
    backends:
    - name: ca-1
      type: builtin
  meshServices:
    mode: Exclusive
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: finance-api
  namespace: app
  labels:
    app: finance-api
spec:
  replicas: 1
  selector:
    matchLabels:
      app: finance-api
  template:
    metadata:
      labels:
        app: finance-api
    spec:
      containers:
      - name: finance-api
        image: finance-api:0.0.1
        imagePullPolicy: IfNotPresent
        ports:
        - containerPort: 8002
        readinessProbe:
          httpGet:
            path: /health
            scheme: HTTP
            port: 8002
          initialDelaySeconds: 5
          timeoutSeconds: 5
---
apiVersion: v1
kind: Service
metadata:
  name: finance-api
  namespace: app
  labels:
    app: finance-api
spec:
  selector:
    app: finance-api
  ports:
  - protocol: TCP
    port: 80
    appProtocol: http
    targetPort: 8002
```

</details>

<details>
<summary>MeshService YAML Dump</summary>

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshService
metadata:
  annotations:
    kuma.io/display-name: finance-api
  creationTimestamp: "2024-11-11T15:48:33Z"
  generation: 4
  labels:
    app: finance-api
    k8s.kuma.io/is-headless-service: "false"
    k8s.kuma.io/namespace: app
    k8s.kuma.io/service-name: finance-api
    kuma.io/env: kubernetes
    kuma.io/managed-by: k8s-controller
    kuma.io/mesh: default
    kuma.io/origin: zone
    kuma.io/zone: west
  name: finance-api
  namespace: app
  ownerReferences:
  - apiVersion: v1
    kind: Service
    name: finance-api
[...]

PR #13108, opened 2025-03-14
fix(meshservice): fix inbound port selecting logic on MeshService

jijiechen:
## Motivation

Fixing an issue on the MeshService updater

## Implementation information

> update MeshService updater to take into account target port

## Supporting documentation

fixes https://github.com/kumahq/kuma/issues/12042

### DIFF
--- a/pkg/core/resources/apis/meshservice/status/updater.go
+++ b/pkg/core/resources/apis/meshservice/status/updater.go
@@ -2,6 +2,7 @@ package status
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"reflect"
 	"time"
 
@@ -195,8 +196,16 @@ func dpInboundForMeshServicePort(inbounds []*mesh_proto.Dataplane_Networking_Inb
 		if port.Name != "" && inbound.Name == port.Name {
 			return inbound
 		}
-		if port.Port == inbound.Port {
-			return inbound
+
+		switch port.TargetPort.Type {
+		case intstr.Int:
+			if uint32(port.TargetPort.IntVal) == inbound.Port {
+				return inbound
+			}
+		case intstr.String:
+			if port.TargetPort.StrVal == inbound.Name {
+				return inbound
+			}
 		}
 	}
 	return nil
--- a/pkg/core/resources/apis/meshservice/status/updater.go
+++ b/pkg/core/resources/apis/meshservice/status/updater.go
@@ -6,10 +6,6 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	"github.com/pkg/errors"
-	"github.com/prometheus/client_golang/prometheus"
-	"k8s.io/apimachinery/pkg/util/intstr"
-
 	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
 	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
 	"github.com/kumahq/kuma/pkg/core/resources/apis/meshservice"
@@ -22,6 +18,8 @@ import (
 	core_metrics "github.com/kumahq/kuma/pkg/metrics"
 	"github.com/kumahq/kuma/pkg/util/maps"
 	util_time "github.com/kumahq/kuma/pkg/util/time"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 type StatusUpdater struct {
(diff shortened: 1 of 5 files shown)


pkg/core/resources/apis/meshservice/status/updater_test.go
```diff
@@ -230,10 +230,10 @@ var _ = Describe("Updater", func() {
 				Total:     3,
 			},
 		}),
-		Entry("should count healthy DPPs", dpProxiesTestCase{
+		Entry("should count healthy DPPs and use MeshService.ports[].targetPort to select DP inbound", dpProxiesTestCase{
 			meshService: samples.MeshServiceBackendBuilder().
 				WithDataplaneTagsSelectorKV("app", "backend").
-				AddIntPort(builders.FirstInboundPort+1, builders.FirstInboundServicePort+1, core_mesh.ProtocolHTTP),
```
jijiechen, line 236: The positions of parameters was reversed in previous code: service port should be the first argument, and DP port is  the second.

jijiechen: `MeshServiceBuilder` Sorce: https://github.com/kumahq/kuma/blob/release-2.10/pkg/test/resources/builders/meshservice_builder.go#L67-L77

jijiechen:
I didn't add new test cases because there is already a test case in the existing code base and it covered the scenario. The reason why it didn't fail (as expected without the fix in this PR) in original code is because the test case was not written correctly.

pkg/core/resources/apis/meshservice/status/updater.go
```diff
@@ -195,8 +196,16 @@ func dpInboundForMeshServicePort(inbounds []*mesh_proto.Dataplane_Networking_Inb
 		if port.Name != "" && inbound.Name == port.Name {
```
lobkovilya, line 196: I'm not sure if this is actually correct. I don't think in Kubernetes when a service's port name matches the pod's port name then it's considered a match https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec:

> ports.name (string)
>
> The name of this port within the service. This must be a DNS_LABEL. All ports within a ServiceSpec must have unique names. When considering the endpoints for a Service, this must match the 'name' field in the EndpointPort. Optional if only one ServicePort is defined on this service.

Service's port name is used only for matching with EndpointPort

jijiechen: So let's remove this statement? I think it's sufficient to only use the port number/name matching. My only concern is whether we always have such value of service target port in both Universal and Kubernetes scenarios

lobkovilya: Yes, let's remove this statement.


> My only concern is whether we always have such value of service target port in both Universal and Kubernetes scenarios

On Universal, we always set it to `inbound.Port`: https://github.com/kumahq/kuma/blob/bf28b95f72d634db11cb2378b38d1f022eb4d55f/pkg/core/resources/apis/meshservice/generate/generator.go#L96

On Kubernetes, we set whatever was set in the original Kubernetes Service: https://github.com/kumahq/kuma/blob/3cdbe225910a51ff1520853be20671dd00f10ea4/pkg/plugins/runtime/k8s/controllers/meshservice_controller.go#L403

So I guess it won't be a problem.

But you definitely need to mention this in UPGRADE.md as it's a breaking change.

jijiechen: I tried to compare the behaviour between 2.9 and 2.10. It turns out that the issue (this PR fixes) was newly introduced in this version 2.10: the behaviour from an end user's perspective is not changing, so there is no need to update the `UPGRADE.md`.

The `name` on DP inbound port was automatically assigned from MeshService after a port is matched using `targetPort` in 2.9: https://github.com/kumahq/kuma/commit/bf9c68f475af3eafc17be292204abad6d598619a

And the MeshService updater (the file this PR is trying to change) was also trying to compare this `name` from inbound with MeshService port names, this is also introduced in 2.9: 
https://github.com/kumahq/kuma/commit/11aa52ae0345b483a14d761e3ff4b1f580f398e6#diff-937cc279a26d2767aaca0b4e8cd1d880c38253918c46ddebb01192550f66b7deR146-R187

So, the updater in 2.9 was actually implicitly using `targetPort` because the `name` was assigned to DP inbound only after the port was matching one on MeshService with `targetPort`.  

The issue happens in this release because we changed the logic assigning the `name` of DP inbound port earlier in this PR https://github.com/kumahq/kuma/pull/12783 and we did not change the updater here. So this PR is here to help.

pkg/core/resources/apis/meshservice/status/updater.go
```diff
@@ -176,32 +176,21 @@ func buildDataplaneProxies(
 			}
 		}
 		healthyInbounds := 0
-		for _, port := range ports {
-			if inbound := dpInboundForMeshServicePort(dpp.Spec.GetNetworking().Inbound, port); inbound != nil {
-				if inbound.State == mesh_proto.Dataplane_Networking_Inbound_Ready {
+		for _, meshServicePort := range meshServicePorts {
+			for _, dpInbound := range dpp.Spec.GetNetworking().Inbound {
+				if dpInbound.State == mesh_proto.Dataplane_Networking_Inbound_Ready &&
+					meshservice.MatchInboundWithMeshServicePort(dpInbound, meshServicePort) {
```
lobkovilya, line 182: once at least one inbound is matched you can stop iterating over the inbounds and move on to the next port

jijiechen: Good point

lobkovilya (approved):

Issue #13109, opened 2025-03-14
Kumaclt export prints Mesh resource twice


Automaat:
### Kuma Version

2.9.x

### Describe the bug

When we have Mesh with MeshService exclusive enabled, then if we run `kumactl export` we get `Mesh` resource twice and one of them is empty.

### To Reproduce

Enable MeshService exclusive mode and run kumactl export and run: `kumactl export --format universal`

### Expected behaviour

We should return single Mesh resource

PR #13111, opened 2025-03-14
fix(kumactl): don't export mesh object twice

lukidzi:
## Motivation

Mesh needs to be defined before creating other resources. That's why we need it to be first entry in the exported object:
https://github.com/kumahq/kuma/pull/11511

## Implementation information

I've changed logic to store mesh objects in a different list and later add it first to the exported file.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13109

> run full matrix to check if it doesn't impact federation test


lahabana (approved):
This goes to master should it go to 2.10?

Also should we simply send all `Descriptor().Scope == Global`  first?

lukidzi:
> This goes to master should it go to 2.10?

I wasn't sure since they issue wasn't marked as 2.10


> Also should we simply send all `Descriptor().Scope == Global` first?

Yes it seems better, my bad

slonka (approved):
LGTM

PR #13124, opened 2025-03-17
feat(dns): embedded DNS server instead of coreDNS

lahabana:
## Motivation

Add an embedded DNS server that resolves mesh local hostnames.
This feature is disabled by default and needs to be enabled with `kuma_dns_proxy_port ` on DPs in universal to a non 0 value or simply with `kuma_runtime_kubernetes_injector_builtin_dns_experimental_proxy =true` on CP in Kubernetes (this leverages sidecar injection).

## Implementation information

We leveraged the same system as MeshMetric. The DNS map is exposed in the `_kuma:dynamicconfig` listener, we poll this from the DP at a set frequency and reload the data.

We improved the whole configfetcher to make it reusable across different components. We also setup the endpoint so that it works with [etag caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag) while avoids unnecessary reloading and processing, we can therefore refresh the configs more quickly as well.

For the moment I added a new matrix for e2e test. If this is stable I'll switch the default.


lahabana:
I've made a few separate commits as this is a bit of a mouthful

.github/workflows/_test.yaml
```diff
@@ -69,6 +69,8 @@ jobs:
                   {"target":"universal", "k8sVersion":"${{ env.K8S_MAX_VERSION }}"}
                 ],
                 "include":[
+                  {"embeddedDNS": "true", "k8sVersion": "${{ env.K8S_MAX_VERSION }}", "target": "kubernetes", "arch": "amd64"},
```
lukidzi, line 72: Another option is to enable it on one Universal and Kubernetes zone of mulitzone test

lahabana: Yeah TBH as soon as I make sure this is stable I want to enable it by default and just have 1 fallback option

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -184,22 +185,12 @@ func configureOpenTelemetryBackend(rs *core_xds.ResourceSet, proxy *core_xds.Pro
...
 		return err
 	}
-	rs.Add(&core_xds.Resource{
-		Name:     listener.GetName(),
-		Origin:   OriginDynamicConfig,
-		Resource: listener,
-	})
-
-	return nil
+	return dynconf.AddConfigRoute(proxy, rs, dpapi.PATH, marshal)
```
Automaat, line 193: why do we use dault path `/dns` for meshmetric config endpoint?

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (s *Server) Handler(res dns.ResponseWriter, req *dns.Msg) {
+	start := time.Now()
+	defer func() {
+		s.metrics.RequestDuration.Observe(time.Since(start).Seconds())
+	}()
+	response := &dns.Msg{}
+	// In case it was never loaded
+	s.dnsMap.CompareAndSwap(nil, &dnsMap{
+		ARecords:    make(map[string]*dnsEntry),
```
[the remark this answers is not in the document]
lahabana: What would we do with SRV records? While Envoy handles SRV record the map we configure is empty so Envoy would return NXDOMAIN and fallback to proxying

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -0,0 +1,192 @@
...
+	"time"
+
+	"github.com/miekg/dns"
+
+	"github.com/kumahq/kuma/pkg/core"
+	"github.com/kumahq/kuma/pkg/core/runtime/component"
+	"github.com/kumahq/kuma/pkg/dns/dpapi"
+)
+
+var log = core.Log.WithName("embeededdns")
```
slonka, line 21: nitpick:

```suggestion
var log = core.Log.WithName("embeddeddns")
```

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -0,0 +1,192 @@
...
+type Server struct {
+	address          string
+	port             uint32
+	done             chan struct{}
+	dnsMap           atomic.Pointer[dnsMap]
+	metrics          *metrics
+	upstreamHostPort string
+}
+
+func NewServer(address string, port uint32) *Server {
```
slonka, line 32: do we have a test for this? just starting the server, seeding it and having a client query it and seeing if the results match?

slonka (commented):
In general looks promising, a bunch of questions and nitpicks. Also would love to see a simple test checking integration with a DNS client.

--- 

- [X] Is the PR title satisfactory? Is this part of a larger feature and should be grouped using `> Changelog`?
- [X] PR description is clear and complete. It [Links to relevant issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) as well as docs and UI issues
- [X] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as an image registry)
- [X] IPv6 is taken into account (.e.g: no string concatenation of host port)
- [X] Does this contain a change that needs to be notified to users? In this case, [`UPGRADE.md`](../blob/master/UPGRADE.md) should be updated.
- [X] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? ([this](https://github.com/kumahq/kuma/actions/workflows/auto-backport.yaml) GH action will add "backport" label based on these [file globs](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L6), if you want to prevent it from adding the "backport" label use [no-backport-autolabel](https://github.com/kumahq/kuma/blob/master/.github/workflows/auto-backport.yaml#L8) label)

lahabana:
@lobkovilya pushed a new commit which simplifies the handler API as it doesn't deal with shutdowns anymore (that's what components are for)

pkg/config/plugins/runtime/k8s/config.go
```diff
@@ -389,6 +389,8 @@ type BuiltinDNS struct {
 	Port uint32 `json:"port,omitempty" envconfig:"kuma_runtime_kubernetes_injector_builtin_dns_port"`
 	// Turn on query logging for DNS
 	Logging bool `json:"logging,omitempty" envconfig:"kuma_runtime_kubernetes_injector_builtin_dns_logging"`
+	// Use the embedded DNS instead (This is an experimental feature)
+	UseEmbedded bool `json:"embedded,omitempty" envconfig:"kuma_runtime_kubernetes_injector_builtin_dns_embedded"`
```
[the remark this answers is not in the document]
lahabana: Nope builtin the container embedded in the binary. In any case these 2 are not meant to be present together for too long

app/kuma-dp/pkg/dataplane/configfetcher/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+func newHandlerMetrics(path string) *handlerMetrics {
+	labels := prometheus.Labels{"path": path}
+	handlerTickCount := prometheus.NewCounter(prometheus.CounterOpts{
+		Name:        "kuma_dp_envoyconfigfetcher_handler_call_count",
+		Help:        "Number of times a handler has been called, unlike onchange_duration_seconds_count this is inclusive of not modified cases",
+		ConstLabels: labels,
+	})
+	prometheus.MustRegister(handlerTickCount)
+	handlerTickDuration := prometheus.NewSummary(prometheus.SummaryOpts{
+		Name:        "kuma_dp_envoyconfigfetcher_handler_call_duration_seconds",
```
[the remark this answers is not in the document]
lahabana: Look at the prom API it takes float64 and it's meant to be seconds (that's also why Duration.Seconds() returns a float64)

app/kuma-dp/cmd/run.go
```diff
@@ -216,29 +226,47 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
...
 				return err
 			}
 			components = append(components, envoyComponent)
-
-			observabilityComponents := setupObservability(kumaSidecarConfiguration, bootstrap, cfg)
+			components = append(components, component.NewResilientComponent(
+				runLog.WithName("configfetcher"),
+				confFetcher,
+				time.Second*5,
+				time.Minute,
```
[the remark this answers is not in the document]
lahabana: Is that useful to be configurable though?

app/kuma-dp/pkg/dataplane/configfetcher/metrics.go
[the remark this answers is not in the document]
lobkovilya: I see, I just saw we have lots of `.Observe(float64(time.Since(start).Milliseconds()))`
```bash
❯ rg  Observe\(.*Milliseconds\(\)' . 
./pkg/dns/vips_allocator.go
68:     d.metrics.VipGenerations.Observe(float64(core.Now().Sub(start).Milliseconds()))

./pkg/insights/resyncer.go
365:    r.idleTime.Observe(float64(startProcessingTime.Sub(start).Milliseconds()))
366:    r.timeToProcessItem.Observe(float64(startProcessingTime.Sub(event.time).Milliseconds()))
401:    r.itemProcessingTime.WithLabelValues(reason, result).Observe(float64(time.Since(startProcessingTime).Milliseconds()))

./pkg/metrics/store/counter.go
74:                     s.metric.Observe(float64(core.Now().Sub(start).Milliseconds()))

...
```
but apparently, we haven't named some metrics properly, i.e. it's called `vip_generation` but actually reports milliseconds. I'll open an issue to track this

app/kuma-dp/cmd/run.go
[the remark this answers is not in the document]
slonka: We'd have to dig the history to see why this was first introduced - https://github.com/kumahq/kuma/pull/9892/files - looks like for resiliency e2e test 🤷 - maybe @Automaat can chip in. Basically I wasn't sure if we should follow how the others are set up or not.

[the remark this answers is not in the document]
lahabana: This is for the CP though right? Here we're talking about the DP

app/kuma-dp/pkg/dataplane/configfetcher/component.go
```diff
@@ -0,0 +1,177 @@
...
+		if hasChanged { // Only compute duration when things changed
+			h.metrics.HandlerTickDuration.Observe(time.Since(start).Seconds())
+		}
+	}
+}
+
+func (cf *ConfigFetcher) stepForHandler(h *handlerInfo) (bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), cf.perHandlerTimeout)
+	defer cancel()
+	r, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("http://localhost%s", h.handler.Path()), nil)
```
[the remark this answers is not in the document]
lahabana: Host is just the `host:` header when doing http over unix domain socket because there's no TCP.

app/kuma-dp/cmd/run.go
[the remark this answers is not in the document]
lahabana: Made a config for it

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -0,0 +1,192 @@
...
+			RR:    []dns.RR{},
+		}
+		for _, ipStr := range record.IPs {
+			ip := net.ParseIP(ipStr)
+			if ip == nil {
+				log.Info("[WARNING] invalid IP address", "ip", ipStr, "record", record)
+				continue
+			}
+			switch {
+			case strings.Contains(ipStr, "."):
```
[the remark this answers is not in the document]
Icarus9913: It doesn't matter, I found the `govalidator` repo does the same way.
https://github.com/asaskevich/govalidator/blob/e11347878e2323a0777b40d33eeffd37a185e31b/validator.go#L818-L828

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -0,0 +1,192 @@
...
+
+	done := make(chan error, 1)
+	go func() {
+		defer close(done)
+		err := srv.ListenAndServe()
+		done <- err
+	}()
+	select {
+	case <-stop:
+		err := srv.Shutdown()
```
lukidzi, line 122: Question: Do we handle drainTime? If a user closes the application, we mark the dataplane as offline, but the sidecar may still have outgoing requests. We need to ensure proper handling of DNS resolution.

lahabana: Worth double checking but AFAIK this is all well handled in: https://github.com/kumahq/kuma/blob/c9644985a213dc2e957c3c3337a151a57cc46c51/app/kuma-dp/cmd/run.go#L311-L333

@jijiechen may know more because I think he did things around this recently

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
lahabana: done

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -0,0 +1,214 @@
...
+		var err error
+		if s.upstreamHandler != nil {
+			resp, err = s.upstreamHandler(req)
+		} else {
+			c := new(dns.Client)
+			resp, _, err = c.Exchange(req, s.upstreamHostPort)
+		}
+		if err != nil {
+			s.metrics.UpstreamRequestFailureCount.Inc()
+			log.Error(err, "Failed to write message to upstream")
```
lobkovilya, line 107: logs are lowercased https://github.com/kumahq/kuma/blob/5e12d423e9a6f3f6f9b5eefaed4a8759395c14b7/docs/madr/decisions/003-logging.md#L34

lobkovilya: there are other instances of upper case logs in this file

app/kuma-dp/pkg/dataplane/meshmetrics/component.go
```diff
@@ -285,37 +239,44 @@ func (cf *ConfigFetcher) mapApplicationToApplicationToScrape(applications []xds.
...
 	defer cancel()
-	for backendName := range cf.runningBackends {
-		err := cf.shutdownBackend(ctx, backendName)
-		if err != nil {
-			logger.Error(err, "Failed shutting down metric exporter")
+	hasError := false
+	for backendName := range m.runningBackends {
+		bErr := m.shutdownBackend(ctx, backendName)
+		if bErr != nil {
+			logger.Error(bErr, "Failed to shutdown backend", "backend", backendName)
```
lobkovilya, line 269: and here

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -84,12 +94,18 @@ func (s *Server) Handler(res dns.ResponseWriter, req *dns.Msg) {
...
-		c := new(dns.Client)
-		resp, _, err := c.Exchange(req, s.upstreamHostPort)
+		var resp *dns.Msg
+		var err error
+		if s.upstreamHandler != nil {
+			resp, err = s.upstreamHandler(req)
+		} else {
+			c := new(dns.Client)
+			resp, _, err = c.Exchange(req, s.upstreamHostPort)
+		}
```
slonka, line 104: nitpick: we could just have an interface for `func(m *Msg, address string) (r *Msg, rtt time.Duration, err error)` and inject `dns.Client` or  a mock

pkg/config/app/kuma-dp/config.go
```diff
@@ -150,6 +152,10 @@ type Dataplane struct {
 	DrainTime config_types.Duration `json:"drainTime,omitempty" envconfig:"kuma_dataplane_drain_time"`
 	// Port that exposes kuma-dp readiness status on localhost, set this value to 0 to provide readiness by "/ready" endpoint from Envoy adminAPI
 	ReadinessPort uint32 `json:"readinessPort,omitempty" envconfig:"kuma_readiness_port"`
+	// ResilientComponentBaseBackoff defines the base backoff between restarts of resilient components
+	ResilientComponentBaseBackoff config_types.Duration `json:"resilientComponentBaseBackoff,omitempty" envconfig:"kuma_dataplane_resilient_component_base_backoff"`
+	// ResilientComponentMaxBackoff defines the max backoff between restarts of resilient components
+	ResilientComponentMaxBackoff config_types.Duration `json:"ResilientComponentMaxBackoff,omitempty" envconfig:"kuma_dataplane_resilient_component_max_backoff"`
```
slonka, line 158: shouldn't this be lowercase?

```suggestion
	ResilientComponentMaxBackoff config_types.Duration `json:"resilientComponentMaxBackoff,omitempty" envconfig:"kuma_dataplane_resilient_component_max_backoff"`
```

test/e2e_env/kubernetes/gateway/delegated.go
```diff
@@ -33,6 +34,9 @@ func Delegated() {
 		MeshServiceMode:             mesh_proto.Mesh_MeshServices_Disabled,
 		UseEgress:                   false,
 	}
+	AfterEachFailure(func() {
+		DebugKube(kubernetes.Cluster, config.Mesh, config.Namespace, config.NamespaceOutsideMesh)
+	})
```
slonka, line 39: Seems unrelated right? 🤔

slonka (dismissed):

slonka:
Oh, did not notice that the tests are failing.

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
lahabana: right reworked this way

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func (s *Server) Start(stop <-chan struct{}) error {
+	defer close(s.done)
+	srv := &dns.Server{Addr: s.address, Handler: dns.HandlerFunc(s.Handler), Net: "udp"}
+	config, err := dns.ClientConfigFromFile("/etc/resolv.conf")
+	if err != nil {
+		return fmt.Errorf("failed to read DNS for /etc/resolv.conf %w", err)
+	}
+	s.upstreamHostPort = net.JoinHostPort(config.Servers[0], config.Port)
```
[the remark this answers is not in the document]
lahabana: done

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -0,0 +1,214 @@
...
+
+	done := make(chan error)
```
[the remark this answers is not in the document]
lahabana: done

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
lahabana: I think I went over all the cases

lobkovilya (approved):

slonka (approved):

Issue #13149, opened 2025-03-20
High CPU usage when using multiple `to` policies


lukidzi:
### Kuma Version

master

### Describe the bug

When a user has multiple `to` policies targeting a `MeshService` (old) that affect a single dataplane, along with a default `to` policy for the `Mesh`, our build rules algorithm performs excessive work by iterating over unnecessary permutations. This leads to significantly high CPU usage.

### To Reproduce

1. Deploy kuma
2. Use kuma-tools to generate 40 services
3. Create policies
```
---
apiVersion: kuma.io/v1alpha1
kind: MeshLoadBalancingStrategy
metadata:
  annotations:
    kuma.io/display-name: mesh-load-balancing-strategy-default
  labels:
    kuma.io/mesh: default
  name: mesh-load-balancing-strategy-default
  namespace: kuma-system
spec:
  targetRef:
    kind: Mesh
  to:
  - default:
      loadBalancer:
        leastRequest:
          choiceCount: 3
        type: LeastRequest # check diff with ringhash in envoy docs on absence of header
    targetRef:
      kind: Mesh
---
apiVersion: kuma.io/v1alpha1
kind: MeshLoadBalancingStrategy
metadata:
  name: srv-000-policy
  namespace: kuma-system
spec:
  targetRef:
    kind: MeshService
    name: srv-000_kuma-test_svc_80
  to:
  - default:
      loadBalancer:
        ringHash:
          hashPolicies:
          - header:
              name: x-affinity
            type: Header
        type: RingHash
    targetRef:
      kind: MeshService
      name: srv-005_kuma-test_svc_80
  - default:
      loadBalancer:
        ringHash:
          hashPolicies:
          - header:
              name: x-affinity
            type: Header
        type: RingHash
    targetRef:
      kind: MeshService
      name: srv-004_kuma-test_svc_80
  - default:
      loadBalancer:
        ringHash:
          hashPolicies:
          - header:
              name: x-affinity
            type: Header
        type: RingHash
    targetRef:
      kind: MeshService
      name: srv-003_kuma-test_svc_80
  - default:
      loadBalancer:
        ringHash:
          hashPolicies:
          - header:
              name: x-affinity
            type: Header
        type: RingHash
    targetRef:
      kind: MeshService
      name: srv-002_kuma-test_svc_80
  - default:
      loadBalancer:
        ringHash:
          hashPolicies:
          - header:
              name: x-affinity
            type: Header
        type: RingHash
    targetRef:
[...]

PR #13151, opened 2025-03-20
perf(rules): add `withNegation` flag to simplify `to` policy flow

lukidzi:
## Motivation

> [!NOTE] 
> This is not a problem when using `meshServices.mode: Exclusive` with policies in a new style

When a user has multiple `to` policies for a single top-level target reference and a default to Mesh target, CPU usage spikes due to the expensive matching process.
Screenshot 2025-03-20 at 18 43 36

## Implementation information

After analyzing the issue with @lobkovilya, we discovered that for `to` policies, our logic checks every permutation of tags. However, this is unnecessary because the only possible tags for `to` policies are `kuma.io/service`, or no tags at all when the target is `Mesh`.

The only exception is `MeshHTTPRoute`, where we generate the `__rule-matches-hash__` tag, which points to a hash of the matcher. Since we don’t need to evaluate all permutations, we decided to iterate through the subsets and match them directly with the policies. To optimize further, we first deduplicate entries to avoid redundant computations.

After implementing these changes, I deployed the update and ran a performance profile, which showed significant improvements.

image

fix: https://github.com/kumahq/kuma/issues/13149


pkg/plugins/policies/core/rules/rules.go
```diff
@@ -521,6 +522,46 @@ func BuildRules(list []PolicyItemWithMeta) (Rules, error) {
 	return rules, nil
 }
 
+func createRule(ss subsetutils.Subset, items []PolicyItemWithMeta) ([]*Rule, error) {
+	// 5. For each combination determine a configuration
```
lobkovilya, line 526: nit: please remove the comment from the function

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -417,16 +419,42 @@ func BuildRules(list []PolicyItemWithMeta) (Rules, error) {
...
+		for _, tag := range ss {
+			uniqueKeys[tag.Key] = struct{}{}
+		}
 		subsets = append(subsets, ss)
 	}
 
+	// we don't need to generate all permutations when there is no negations
+	// and we have only 0 or one tag, in other cases we need to generate.
+	// in case of `to` policies it can happen when using MeshHTTPRoute
+	if !withNegations && len(uniqueKeys) <= 1 {
```
lobkovilya, line 439: Why do we need to check `uniqueKeys`? I think for `to` policies it should work even if you add route hash tag

lukidzi: You are right it works, we don't need it

pkg/api-server/testdata/resources/inspect/dataplanes/_rules/meshhttproute.golden.json
```diff
@@ -40,7 +40,15 @@
          "backendRefs": [
           {
            "kind": "MeshServiceSubset",
-           "name": "backend_kuma-demo_svc_3001",
+           "name": "other-svc",
```
lobkovilya, line 43: why did the order change here?

pkg/api-server/testdata/resources/inspect/dataplanes/_rules/resource_rule_meshtimeout_index.golden.json
```diff
@@ -46,17 +46,17 @@
    "toRules": [
     {
      "conf": {
-      "connectionTimeout": "7s",
-      "idleTimeout": "7s",
+      "connectionTimeout": "5s",
```
lobkovilya, line 49: this one is also weird, it's a new `resourceRule` test, so shouldn't be affected no?

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -427,6 +429,29 @@ func BuildRules(list []PolicyItemWithMeta) (Rules, error) {
 		subsets = append(subsets, ss)
 	}
 
+	// we don't need to generate all permutations when there is no negations
+	// and we have only 0 or one tag, in other cases we need to generate.
```
lobkovilya, line 433: need to update the comment without mentioning `0 or one tag`

pkg/api-server/testdata/resources/inspect/dataplanes/_rules/resource_rule_meshtimeout_index.golden.json
lukidzi: It's a `toRules` section changing so doesn't affect `resourceRule`

pkg/api-server/testdata/resources/inspect/dataplanes/_rules/meshhttproute.golden.json
lukidzi: We were missing sorting by name, fixed

lukidzi: Seems like it's still different order.It's because we were sorting by nodes ids and not string name. I think it's just best to keep order of names in tags

pkg/plugins/policies/core/rules/subsetutils/subset.go
```diff
@@ -337,3 +350,48 @@ func (c *SubsetIter) simplified() Subset {
...
 	return result
 }
+
+// Deduplicate returns a new slice of subsetutils.Subset with duplicates removed.
+func Deduplicate(subsets []Subset) []Subset {
+	seen := make(map[string]struct{})
+	result := make([]Subset, 0, len(subsets))
+
+	for _, s := range subsets {
+		key := canonicalSubset(s)
```
lobkovilya, line 360: sorting probably should happen inside the `canonicalSubset` not after it

pkg/plugins/policies/core/rules/subsetutils/subset.go
```diff
@@ -337,3 +350,48 @@ func (c *SubsetIter) simplified() Subset {
...
+		// sort values by key, value, not
+		s.Sorted()
+		if _, exists := seen[key]; !exists {
+			seen[key] = struct{}{}
+			result = append(result, s)
+		}
+	}
+	sort.SliceStable(result, func(i, j int) bool {
```
lobkovilya, line 368: this sort is not really needed, as if we received an already somehow ordered `subsets` as input then by deduplication we won't unstabilize the order

pkg/api-server/testdata/resources/inspect/meshgateways/_rules/meshhttproute.golden.json
```diff
@@ -220,57 +220,13 @@
    "inboundRules": [],
    "toResourceRules": [],
    "toRules": [
-    {
```
lobkovilya, line 223: ugh... this one is actually incorrect, with MeshHTTPRoute for MeshGateway we allow `to[].targetRef: Mesh` and combining with `__rule-matches-hash__` it behaves differently. Probably you should restore your check of number of distinct tags to be 1

lobkovilya (approved):

PR #12982, opened 2025-02-28
test(e2e): make reports much easier to use

lahabana:
We rely on ginkgo reports and make the hierarchy correct with file names that make sense

PR #13191, opened 2025-03-25
chore(e2e): rewrite e2e config init

slonka:
## Motivation

Instead of https://github.com/kumahq/kuma/pull/13165 - because that did not work. Please review the parent PR first.

That way in a parent project we can get the default and overwrite only the parts of the config that we care about. So if a change happens like in https://github.com/kumahq/kuma/pull/12982/files#diff-4097df65e73d6b3a2813df0ea3c87f15a6c43ba3aca9b73b3e60346758ef0344R65 we won't need to modify the parent project.

## Implementation information

Manually init the e2e configuration by moving the init function to `RunE2ESpecs`.


lahabana:
That's a lot better! Thanks for doing this

bartsmykla (approved):

PR #13152, opened 2025-03-20
chore(deps): upgrade min and max k8s version

lukidzi:
## Motivation

We upgraded the minimal and maximal version of K8s since we have more time to validate during a new release period.

## Implementation information

* changed minimal from 1.25.x to 1.27.x
* changed maximal from 1.31.x to 1.32.x

PR #13197, opened 2025-03-25
Revert "chore(deps): upgrade min and max k8s version"

slonka:
Reverts kumahq/kuma#13152

because it started failing CI on main

image


lukidzi (changes requested):
It's not this one

slonka:
I'm closing this because it's not the root cause of the constant failures of conformance tests and we can create issues for the other flaking tests.

PR #13203, opened 2025-03-25
feat(kuma-cp): update default policies to use new rules api instead of deprecated from

Automaat:
## Motivation
We should update our default policies to use new api

### DIFF
--- a/test/e2e_env/kubernetes/defaults/defaults.go
+++ b/test/e2e_env/kubernetes/defaults/defaults.go
@@ -39,7 +39,8 @@ func Defaults() {
 	It("should create default policies for default mesh", func() {
 		Eventually(policyCreated("trafficpermission", "allow-all-default"), "30s", "1s").MustPassRepeatedly(3).Should(BeFalse())
 		Eventually(policyCreated("trafficroute", "route-all-default"), "30s", "1s").MustPassRepeatedly(3).Should(BeFalse())
-		Eventually(policyCreated("meshtimeout", "mesh-timeout-all-default", Config.KumaNamespace), "30s", "1s").MustPassRepeatedly(3).Should(BeTrue())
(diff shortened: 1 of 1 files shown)


pkg/defaults/mesh/mesh.go
```diff
@@ -61,10 +61,12 @@ func EnsureDefaultMeshResources(
 	}
 
 	defaultResourceBuilders := map[string]func() model.Resource{
-		"mesh-gateways-timeout-all": defaulMeshGatewaysTimeoutResource,
-		"mesh-timeout-all":          defaultMeshTimeoutResource,
-		"mesh-circuit-breaker-all":  defaultMeshCircuitBreakerResource,
-		"mesh-retry-all":            defaultMeshRetryResource,
+		"mesh-gateways-timeout-all-outbounds": defaultOutboundMeshGatewaysTimeoutResource,
```
Automaat, line 64: Looking at this code, I am afraid that this is not enough as we will have duplicate timeouts, if old defaults are present. 

Also, simply removing old defaults feels shady, what if someone updated defaults but left the default name?

lukidzi: One more issue. What about zones running older versions? These policies won't work on zones

Automaat: @lukidzi can we protect ourselves from this? some flag on kds or something like this?

lukidzi: It might be possible using a KDS flag, but we would need to translate it to a compatible format at the KDS sync level. That’s a bit tricky, and the global control plane would still display the new version. Since policies are created on the global control plane and it doesn't have awareness of the zone version, it's not possible to determine how to generate the default policy correctly.

Automaat: so should we then wait until we remove support for from entirely?

Automaat:
closing because we need to support LTS version which does not have rules in policies

PR #13213, opened 2025-03-25
Revert "fix(kumactl): don't export mesh object twice"

lukidzi:
Reverts kumahq/kuma#13111

I believe it causes flake of federation test.

1. We need to create an empty Mesh without mTLS
2. Add all secrets
3. Enable mTLS
4. In this case we won't hit an issue with our certificate vs generated certificates


lahabana:
Is the problem more about federation test than the export?

slonka (commented):
> Is the problem more about federation test than the export?

I think that's the case.

> Reverts https://github.com/kumahq/kuma/pull/13111

~~Personally I wouldn't revert this since this actually fixes an issue. The Federation test is now skipped let's fix that instead of reverting.~~ Not sure now - was the original issue ever a problem? 🤔

PR #13219, opened 2025-03-26
ci(gha): pin slsa workflow version and update backport runner to ubuntu-24.04

bartsmykla:
## Motivation

Security + stability

## Implementation information

- Replaced version tags with full commit SHA for slsa-generator workflows
- Updated `runs-on` from `ubuntu-latest` to `ubuntu-24.04` in backport workflow

### DIFF
--- a/.github/workflows/_provenance.yaml
+++ b/.github/workflows/_provenance.yaml
@@ -30,7 +30,7 @@ jobs:
       actions: read # For getting workflow run info to build provenance
       id-token: write # needed for signing the images
     # need to use non hash version because of: https://github.com/slsa-framework/slsa-github-generator/issues/3498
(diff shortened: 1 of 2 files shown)


.github/workflows/_provenance.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
       actions: read # For getting workflow run info to build provenance
       id-token: write # needed for signing the images
     # need to use non hash version because of: https://github.com/slsa-framework/slsa-github-generator/issues/3498
```
lahabana, line 32: This is no longer the case?

slonka: So this is still the case but the actual link should be https://github.com/slsa-framework/slsa-verifier/issues/12 ? right?

lahabana: No idea. It's all old

Issue #13158, opened 2025-03-21
rework sidecarconfig driven by configMap


lahabana:
### Description

There's a few little issues with the existing experimental feature. Let's try to iron them out in this next try.

lahabana:
The first step here is to write a MADR to explain what's currently done, why it's a problem and what we can do to improve

lahabana:
Closing this as just the deprecation is left over and it's tracked with a milestone

PR #13262, opened 2025-03-28
docs(MADR): reworking transparent proxy config with ConfigMaps

bartsmykla:
## Motivation

In version 2.9, we introduced an experimental feature to configure the transparent proxy using a Kubernetes ConfigMap. This version had a major limitation: the control plane needed access to ConfigMaps in *all* namespaces, which some users didn’t allow.

## Implementation information

The proxy config is built from default values, control plane configuration, annotations, and ConfigMaps. Since ConfigMaps are namespace-scoped, the component that builds the final config must access all of these inputs. The simplest way to do that is in the control plane, which then passes the final config to `kuma-init`, `kuma-cni`, and `kuma-sidecar`. But this requires the control plane to access all workload namespaces, which led to expanding its ClusterRole permissions.

This document proposes removing that requirement for setups that don’t use `kuma-cni`. Instead of having the control plane build the full config, each data plane component (`kuma-init` and `kuma-sidecar`) will build its own.

## Supporting documentation

Related to: https://github.com/kumahq/kuma/issues/13158


docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,63 @@
...
+1. Combine default settings from its config and the `kuma-system` ConfigMap
+2. Apply any annotations set on the Pod
+3. Compute the differences from defaults
+4. Add a single annotation `traffic.kuma.io/transparent-proxy-config` with the final config
+
+Components (`kuma-init`, `kuma-dp`) will read this annotation via a mounted volume using the Downward API.
+
+If there's a `traffic.kuma.io/transparent-proxy-configmap-name` annotation, the named ConfigMap will be mounted into `kuma-init` and `kuma-dp`, and its values will override everything else.
+
+Even though environment variables seem easier, we avoid using them for the full config. They take priority over everything else when parsed, and the control plane can't safely set them if it doesn't have access to the workload's ConfigMaps. Annotations give better control over precedence.
```
lahabana, line 59: One option to allow override is to have the possibility to have 2 yaml for the conf. You'd load the main one and then override it with the namespace specific one.

Btw should we even use: `traffic.kuma.io/transparent-proxy-configmap-name` after all a user could just specify their own path for the transparent-proxy-config and that's all. They'd then have to add the mount themselves.
What I mean is that maybe we're just trying too hard with this feature and users simply need to be able to specify the path to they transparent-proxy config...

lahabana (commented):
good direction!

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,63 @@
...
+### Each component gets config in a different way
+
+- `kuma-init`: gets full config as a CLI argument
+- `kuma-dp`: gets config as environment variables (`KUMA_DNS_ENABLED`, `KUMA_DNS_CORE_DNS_PORT`)
+
+This inconsistency makes it hard to track where values are coming from.
+
+## Solution
+
+We remove proxy config from the Dataplane object. Instead, the control plane will pass the config as metadata when `kuma-dp` connects.
```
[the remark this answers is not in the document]
bartsmykla: I probably used the wrong words. I didn’t mean to remove them completely. I meant to stop using them and mark them as deprecated. There won’t be any issues with Universal, since the same information will now be sent as metadata during the xDS connection, using values from the transparent proxy config passed to `kuma-dp` via CLI flag. Improved this paragraph to make it clearer.

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,63 @@
...
+
+`kuma-dp` will accept a CLI flag like `--transparent-proxy-config`, just like `kuma-init`. Some fields can still be overridden using env vars.
+
+During sidecar injection, the control plane will:
+1. Combine default settings from its config and the `kuma-system` ConfigMap
+2. Apply any annotations set on the Pod
+3. Compute the differences from defaults
+4. Add a single annotation `traffic.kuma.io/transparent-proxy-config` with the final config
+
+Components (`kuma-init`, `kuma-dp`) will read this annotation via a mounted volume using the Downward API.
```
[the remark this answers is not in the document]
bartsmykla: done

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
bartsmykla: > One option to allow override is to support two YAMLs for the config. You'd load the main one, then override it with the namespace-specific one.

Yeah exactly, I didn’t spell it out, but that’s what I had in mind too, so I updated the MADR with more explicit explanation.

> Btw should we even use `traffic.kuma.io/transparent-proxy-configmap-name`? A user could just specify their own path for the transparent-proxy config and be done with it. They’d have to add the mount themselves though.  
What I’m getting at is that maybe we’re trying too hard with this, and users just need a way to specify the config path...

> 3. The user mounts the configuration at a specific path, e.g., /etc/kuma.io/transparent-proxy.

Thing is, volume mounts are container-scoped, so the user would have to add them via `ContainerPatches`, which is fragile since the order of the config files matters. I don’t think that’s better than just using the `traffic.kuma.io/transparent-proxy-configmap-name` annotation.

What could work though is: if a user adds a volume with a specific name and a ConfigMap source, we auto-mount it into our containers. @lahabana, @lukidzi Wdyt?

slonka (commented):
I'll do another pass today / tomorrow but I don't see anything obvious that I would change

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,168 @@
...
+### Each component gets config in a different way
+
+- `kuma-init`: gets full config as a CLI argument
+- `kuma-sidecar`: gets config as environment variables (`KUMA_DNS_ENABLED`, `KUMA_DNS_CORE_DNS_PORT`)
+
+This inconsistency makes it hard to track where values are coming from.
+
+## Solution
+
+We stop using the transparent proxy config fields in the Dataplane object and mark them as deprecated. These fields will no longer be relied on by `kuma-sidecar` in Kubernetes deployments. Instead, the same information will be passed as metadata during the xDS connection, using values from the transparent proxy config passed via a CLI flag.
```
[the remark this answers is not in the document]
bartsmykla: Correct

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,168 @@
...
+During sidecar injection, the control plane will:
+1. Merge default values from the control plane config and the `kuma-system` ConfigMap
+2. Apply any Pod-level annotations
+3. Add the resulting values (only those that differ from defaults) as a single annotation: `traffic.kuma.io/transparent-proxy-config`
+
+To avoid unintended overrides, we don’t use environment variables for the full config. Env vars take precedence during parsing and can’t be safely set when the control plane doesn’t have access to workload-level ConfigMaps. Annotations provide better control over precedence and merging.
+
+The annotation will be mounted into each container as a file named `0.yaml` under `/tmp/transparent-proxy`.
+
+If the Pod also has a `traffic.kuma.io/transparent-proxy-configmap-name` annotation, the named ConfigMap will be mounted in the same directory as `1.yaml`. Since files are processed alphabetically, it will override values from the default config.
```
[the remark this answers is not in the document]
bartsmykla: It doesn't because that's incorrect. 0.yaml is the result of merging values from multiple sources (cp config, default tproxy configmap, annotations) by the sidecar injector

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,63 @@
...
+- `kuma-init`: gets full config as a CLI argument
+- `kuma-dp`: gets config as environment variables (`KUMA_DNS_ENABLED`, `KUMA_DNS_CORE_DNS_PORT`)
+
+This inconsistency makes it hard to track where values are coming from.
+
+## Solution
+
+We remove proxy config from the Dataplane object. Instead, the control plane will pass the config as metadata when `kuma-dp` connects.
+
+`kuma-dp` will accept a CLI flag like `--transparent-proxy-config`, just like `kuma-init`. Some fields can still be overridden using env vars.
```
lukidzi: What is the model of this flag? Also, I feel creating a `json` config here a really difficult for the user over simple dataplane configuration (well structured and documented)

bartsmykla: I don't agree that this would be difficult. It's the same YAML config that users already use when installing the transparent proxy, so they'd just need to pass the same file or values as with `kumactl install transparent-proxy` (in the context of universal).

This config is basically the same as what they use in the Dataplane. For example, if before the user had:

```yaml
networking:
  transparentProxying:
    ipFamilyMode: IPv4
    redirectPortInbound: 7777
    redirectPortOutbound: 8888
```

Now they'd pass it via CLI like:

```yaml
ipFamilyMode: ipv4
redirect:
  inbound:
    port: 7777
  outbound:
    port: 8888
```

They already had to provide these values when installing the transparent proxy, and in my experience it's rare for users to modify the defaults anyway.

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,168 @@
...
+- **Pod Reconciler** - creates and updates the `Dataplane` object based on the Pod
+
+The `Dataplane` object includes transparent proxy settings like `ipFamilyMode`, `redirectPortInbound`, and `redirectPortOutbound`. These are used by the control plane to generate the xDS config delivered to `kuma-sidecar`.
+
+Sidecar injection and Dataplane generation are handled by separate components, with no shared state. This means there is no direct way to pass configuration between them.
+
+### Each component gets config in a different way
+
+- `kuma-init`: gets full config as a CLI argument
+- `kuma-sidecar`: gets config as environment variables (`KUMA_DNS_ENABLED`, `KUMA_DNS_CORE_DNS_PORT`)
```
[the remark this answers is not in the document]
bartsmykla: These are the only ones

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
lukidzi: Should we have an option to provide `kuma-cp --transparent-proxy-config or --transparent-proxy-config-file` something like we have for Dataplane?

bartsmykla: I'm not sure why. We already have an option to configure the above values using control plane configuration and adding another source of truth for transparent proxy configuration will introduce another point of complexity which is unnecessary in my opinion.

lukidzi: My bad i meant `kuma-dp --transparent-proxy-config or --transparent-proxy-config-file` we have this for DP and it's easier to provide tp config in a file

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -44,7 +44,14 @@ This inconsistency makes it hard to track where values are coming from.
 
 We stop using the transparent proxy config fields in the Dataplane object and mark them as deprecated. These fields will no longer be relied on by `kuma-sidecar` in Kubernetes deployments. Instead, the same information will be passed as metadata during the xDS connection, using values from the transparent proxy config passed via a CLI flag.
 
-Both `kuma-init` and `kuma-sidecar` will accept a `--config-dir` CLI flag. This directory will contain one or more configuration files, read in alphabetical order. In case of conflicts, values from the later file override earlier ones.
+`kuma-dp` will support a `--transparent-proxy` flag, and `kumactl install transparent-proxy` will support a unified `--config` flag. Both flags can accept one of the following values:
```
slonka, line 47: Nitpick: WDYT about calling it `--transparent-proxy-config`?

```suggestion
`kuma-dp` will support a `--transparent-proxy-config` flag, and `kumactl install transparent-proxy` will support a unified `--config` flag. Both flags can accept one of the following values:
```

slonka (dismissed):

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,173 @@
...
+### Each component gets config in a different way
+
+- `kuma-init`: gets full config as a CLI argument
+- `kuma-sidecar`: gets config as environment variables (`KUMA_DNS_ENABLED`, `KUMA_DNS_CORE_DNS_PORT`)
+
+This inconsistency makes it hard to track where values are coming from.
+
+## Solution
+
+We stop using the transparent proxy config fields in the Dataplane object and mark them as deprecated. These fields will no longer be relied on by `kuma-sidecar` in Kubernetes deployments. Instead, the same information will be passed as metadata during the xDS connection, using values from the transparent proxy config passed via a CLI flag.
```
lahabana, line 45: Will the presence of the metadata define whether we read the fields?
Do we need a feature flag in the bootstrap request for compatibility during upgrades?
Do we need to keep writing these fields for X versions to support CP and DP downgrade/upgrades?

I'd like compatiblity to be discussed here please :)

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
lahabana: :( I wish install transparent-proxy was in the DP instead of kumactl :(

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,173 @@
...
+
+`kuma-dp` will support a `--transparent-proxy` flag, and `kumactl install transparent-proxy` will support a unified `--config` flag. Both flags can accept one of the following values:
+
+- No value: uses the default transparent proxy config (allows running `kuma-dp` without enabling the proxy)
+- A path to a single config file
+- A path to a directory with one or more config files, read in alphabetical order (later files override earlier ones)
```
lahabana, line 51: Wouldn't it be more straightforward to take a list of files and only rely on the order in the arg?

lahabana: I see almost a security risk of reading an entire dir without a lot of value from a user POV

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
```diff
@@ -0,0 +1,173 @@
...
+- A raw YAML string with the config
```
lahabana, line 52: Shouldn't it stick with the more std "-" and read stdin. 
You could then combine: 

```
echo 'baseconf' | kumactl install transparent-proxy --transparent-proxy-config /var/conf.yaml,-
```

lahabana (commented):
I'm almost good with this. I'd honnestly just stay away from alphabetically loading all files in a dir. Let's just add a the files to the command it isn't that complicated.

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
bartsmykla: > Nitpick: WDYT about calling it --transparent-proxy-config?

I went with `--transparent-proxy` so that when no value is provided, it simply enables transparent proxy with the default config. Using `--transparent-proxy-config` makes that less intuitive. But we could support both and make one an alias of the other for flexibility. WDYT?

bartsmykla: @slonka I've updated the MADR and included my suggestion about aliases with an example

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
bartsmykla: addressed

docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md
bartsmykla: I added the section about compatibility

slonka (dismissed):

slonka (approved):

lukidzi (approved):

lahabana (approved):
Should we also add the default transparent-proxy config as an entry in the helm config? This way it would enable us to remove a bunch of configs in the CP no?

PR #13282, opened 2025-04-01
chore(codeql): ignore hostkeycallback in test file

lukidzi:
## Motivation

`paths-ignore` are not supported for golang

> Changelog: skip


lahabana (approved):
Reminder: only use `chore` for user visible changes, this is a `ci` change

PR #13301, opened 2025-04-02
test(framework): support running universal tests on remote host

jijiechen:
## Motivation

Support running universal tests on remote hosts

## Implementation information

* support establishing SSH tunnels to remote hosts before executing commands
* introduce a new `DockerBackend` to encapsulate calls to docker commands: the local backend executes directly, the new remote backend executes commands via an SSH tunnel
* support forwarding local port to remote host within the SSH tunnel so `kumactl` is able to access the CP ports exposed from the container running on the remote host


test/framework/docker.go
```diff
@@ -61,3 +92,247 @@ func GetPublishedDockerPorts(
...
+		if strings.HasPrefix(addr, ":::") {
+			addr = "[::]:" + addr[3:]
+		}
+		_, pubPortStr, err := net.SplitHostPort(addr)
+		if err != nil {
+			return nil, err
+		}
+
+		// setup SSH port forwarding for the remote port
+		pf, err := portForward(u.networking, fmt.Sprintf("%s:%d", u.networking.RemoteHost.Address, pubPortStr))
```
lahabana, line 226: net.JoinHostPort to work with ipv6

lahabana: Question is this actually the host public ip? We need to check this

test/framework/universal/networking.go
```diff
@@ -95,6 +132,31 @@ func (s *Networking) NewSession(exportPath string, cmdName string, verbose bool,
 	return session, nil
 }
 
+func (s *Networking) PortForward(local, remote string, stopChan <-chan struct{}) (net.Addr, error) {
```
lahabana, line 135: ports? address? address+Port? A comment would be helpful here

test/framework/docker.go
jijiechen: I'm going to use `127.0.0.1` here because the local port should be forwarded just to the remote host itself.

test/framework/ssh/portforward.go
```diff
@@ -0,0 +1,58 @@
...
+
+			go func(localConn net.Conn) {
+				remoteConn, err := sshClient.Dial("tcp", remote)
+				if err != nil {
+					log.Fatalf("failed to dial to ssh remote %q on host %s: %q",
+						remote, sshClient.RemoteAddr().String(), err)
+				}
+
+				go pipe(remoteConn, localConn)
+				go pipe(localConn, remoteConn)
```
slonka, line 54: I don't understand this. Doesn't this create some sort of a loop? 🤔

slonka (commented):
There is a lot of utility code that I'm a bit surprise has not been implemented anywhere else. In general I can't find anything obviously wrong anywhere. Just one Q about the pipes.

test/framework/ssh/portforward.go
jijiechen: Sockets are bi-directional. So a socket connection can be used as both `io.Reader` and `io.Writer`.

The `pipe` function takes two parameter as `io.Writer` and `io.Reader`. So a connection is first used as an `io.Writer` and then used as an `io.Reader`, or the other way around.

Each invocation handles a single direction, so there is no loop.

jijiechen:
A lot of thanks to your reviewing! @slonka 

> There is a lot of utility code that I'm a bit surprise has not been implemented anywhere else.

Agreed, and this is the same feeling that @lahabana had.  

I expected there should be some existing SSH tunneling implementations at least. Well, there are such implementations, but those are either out-of-date or very small projects, I could not tell how well they are maintained (or even if they are actually maintained.)

Some example projects are:
* https://github.com/SommerEngineering/SSHTunnel
* https://github.com/glycerine/sshego
* https://github.com/rgzr/sshtun

The other parts of the `RemoteDockendBackend` mostly ported from our existing docker commands (which are powered by the `terratest` library). They are not complicated.

slonka (approved):
I don't see anything obviously wrong here, just the pipe thing that Jay explained.

Issue #13231, opened 2025-03-26
Fix or remove tests: Mtls should enforce traffic permissions (E2E Universal Suite)


bartsmykla:
These tests are currently disabled. We should check if they're still relevant. If they are, we should fix them; if not, we should remove them.

https://github.com/kumahq/kuma/blob/02a8b259fc968f8a5d522e574f5fdaf1652fb628/test/e2e_env/universal/mtls/mtls.go#L160

### Skipped test cases

**Mtls should enforce traffic permissions**

https://github.com/kumahq/kuma/blob/02a8b259fc968f8a5d522e574f5fdaf1652fb628/test/e2e_env/universal/mtls/mtls.go#L195

**Mtls should enforce traffic permissions**

https://github.com/kumahq/kuma/blob/02a8b259fc968f8a5d522e574f5fdaf1652fb628/test/e2e_env/universal/mtls/mtls.go#L205

**Mtls should enforce traffic permissions**

https://github.com/kumahq/kuma/blob/02a8b259fc968f8a5d522e574f5fdaf1652fb628/test/e2e_env/universal/mtls/mtls.go#L219

PR #13304, opened 2025-04-02
test(e2e): enable mtls universal test

lukidzi:
## Motivation

The test was disabled

## Implementation information

Instead of using Consistently, I switched to Eventually with MustPassRepeatedly. Since the change might not propagate immediately and could fail once, it's more reliable to wait for the configuration to take effect before validating that it passes.

Fix [#13231](https://github.com/kumahq/kuma/issues/13231)

> Changelog: skip

### DIFF
--- a/test/e2e_env/universal/mtls/mtls.go
+++ b/test/e2e_env/universal/mtls/mtls.go
@@ -3,7 +3,6 @@ package mtls
 import (
 	"fmt"
 	"net"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


slonka (approved):
How does `Eventually` with `MustPassRepeatedly` work? does it just wait for N consecutive passes? 🤔

lukidzi:
> How does `Eventually` with `MustPassRepeatedly` work? does it just wait for N consecutive passes? 🤔

Yes, wait until it's success and then waits for 10 consecutive success

PR #13306, opened 2025-04-02
test(e2e): enable and fix tp universal test

lukidzi:
## Motivation

We should remove or fix skipped tests

## Implementation information

* fix install command and assertion
* fix uninstall assertion and command
* added one more condition to check if traffic is broken

Fix https://github.com/kumahq/kuma/issues/13232

> Changelog: skip

### DIFF
--- a/test/e2e_env/universal/transparentproxy/transparent_proxy.go
+++ b/test/e2e_env/universal/transparentproxy/transparent_proxy.go
@@ -35,16 +35,7 @@ func TransparentProxy() {
 		Expect(universal.Cluster.DeleteMesh(mesh)).To(Succeed())
 	})
 
-	// This test is currently disabled as it basically never worked as expected
-	// as when installing tproxy again, it was just adding the same set of rules
-	// after the existing ones (with small exception to DNS ones, but it's
-	// irrelevant here). So the tests were passing only because rules form
-	// the second installation have been just ignored.
-	//
-	// When mechanism to uninstall tproxy will be implemented (ref.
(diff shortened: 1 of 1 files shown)


test/e2e_env/universal/transparentproxy/transparent_proxy.go
```diff
@@ -35,42 +35,35 @@ func TransparentProxy() {
...
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(stdout).To(ContainSubstring("transparent proxy cleanup completed successfully"))
+		}).Should(Succeed())
+
+		// then
+		Eventually(func(g Gomega) {
+			failures, err := client.CollectFailure(universal.Cluster, "tp-client", "test-server.mesh")
 			g.Expect(err).ToNot(HaveOccurred())
-			g.Expect(stdout).To(ContainSubstring("Transparent proxy uninstalled successfully"))
+			g.Expect(failures.Exitcode).To((Or(Equal(6), Equal(28))))
```
slonka, line 57: why are there 2 codes that can appear?

slonka: maybe in the future we could have these codes mapped to something more readable / understandable

slonka (approved):

test/e2e_env/universal/transparentproxy/transparent_proxy.go
lukidzi: Sometimes we can have timeout on a dns call so I prefer to keep both

PR #13308, opened 2025-04-02
test(e2e): enable multizone MeshHTTPRoute test

lukidzi:
## Motivation

We should verify these functionalities. I tested locally with `MustPassRepeatedly(100)` and had no fails

## Implementation information

* Increase number of requests
* Change time to be less aggressive

Fix https://github.com/kumahq/kuma/issues/13236

> Changelog: skip


slonka (approved):
We can try this out.

slonka:
> I tested locally with MustPassRepeatedly(100) and had no fails

I've done the same but I it never triggered the same as on CI. Let's see.

Issue #12831, opened 2025-02-12
Converge ResourceIdentifier and TypedResouceIdentifier into a single struct


lobkovilya:
### Description

We have ResourceIdentifier https://github.com/kumahq/kuma/blob/d7ec0a2b1ac19208fb7dd9726309e3cf8cdc5848/pkg/core/resources/model/resource.go#L780-L785

and TypedResourceIdentifier https://github.com/kumahq/kuma/blob/d7ec0a2b1ac19208fb7dd9726309e3cf8cdc5848/pkg/core/resources/model/resource.go#L881-L886

we can have a single struct for simplicity
```go
type ResourceIdentifier struct {
    ResourceType ResourceType
    Name         string
    Mesh         string
    Namespace    string
    Zone         string
    SectionName  string
}
```

PR #13310, opened 2025-04-03
refactor(rules): merge `ResourceIdentifier` and `TypedResourceIdentifier`

lobkovilya:
## Motivation

One struct is simpler than two structs

## Implementation information

* created new type `kri.Identifier` in `pkg/core/kri`
* since `kri` package depends on `core_model`, we can't use `kri` in `core_model`, so I had to move `ResolvedBackendRef` types to `pkg/plugins/policies/core/rules/resolve`
* this PR preserves the original `ResourceIdentifier.String()` method to avoid even bigger diff in golden files 

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12831


slonka (approved):
I don't see anything suspicious looking.

PR #13316, opened 2025-04-04
chore(debug): add dump report to e2e tests

slonka:
## Motivation

I tried to figure out what happened to [Upgrade Multizone with Helm](https://github.com/kumahq/kuma/actions/runs/14263282975/job/39980203825#step:12:1538) but got:

```
No files were found with the provided path: build/reports/e2e-debug*. No artifacts will be uploaded.
```

[here](https://github.com/kumahq/kuma/actions/runs/14263282975/job/39980203825#step:13:20).

Now we get e2e debug:

image

## Implementation information

Add `DebugCPLogs` and 

```
ReportAfterSuite("report suite", report.DumpReport)
```

to a bunch of places.


test/e2e/bootstrap/corefile_template.go
```diff
@@ -68,6 +68,7 @@ data:
 	})
 
 	E2EAfterAll(func() {
+		DebugCPLogs(k8sCluster)
```
lahabana, line 71: That's not the way. It should be in 

It should more something like:
```
	AfterEachFailure(func() {
		DebugUniversal(cluster, "default")
	})
```

because the outcome would be a bit different in that case.

lahabana: The `JustAfterEach` ensure it's really the first teardown stuff that happens.

slonka: Replaced with `AfterEachFailure`

slonka:
Works after changes

image

jijiechen (approved):
Great improvement.

Issue #13217, opened 2025-03-26
Federation kubernetes test is flaky


slonka:
### Description

Federation test is again flaky:

- https://github.com/kumahq/kuma/actions/runs/14060648792/job/39371442422?pr=13191#step:12:2373
```
Federation with Kube Global on federation [BeforeAll] should sync data plane proxies to global cp [job-3]
```
[e2e-debug--amd64-v1.32.2-k3s1-flannel--3 (1).zip](https://github.com/user-attachments/files/19462979/e2e-debug--amd64-v1.32.2-k3s1-flannel--3.1.zip)
[e2e-debug--amd64-v1.32.2-k3s1-flannel--3.zip](https://github.com/user-attachments/files/19462978/e2e-debug--amd64-v1.32.2-k3s1-flannel--3.zip)

---

- https://github.com/kumahq/kuma/actions/runs/14033755777/job/39288003826#step:12:2380
```
Federation with Kube Global on federation [BeforeAll] should sync data plane proxies to global cp [job-3]
```

[e2e-debug--amd64-v1.32.2-k3s1-flannel--3 (1).zip](https://github.com/user-attachments/files/19462990/e2e-debug--amd64-v1.32.2-k3s1-flannel--3.1.zip)
[e2e-debug--amd64-v1.32.2-k3s1-flannel--3 (2).zip](https://github.com/user-attachments/files/19462989/e2e-debug--amd64-v1.32.2-k3s1-flannel--3.2.zip)

lahabana:
Is the cause the same as: https://github.com/kumahq/kuma/pull/13213 ?

slonka:
> Is the cause the same as: https://github.com/kumahq/kuma/pull/13213 ?

Seems like they are related but I'm not 100% sure yet. The problem seems to be that there are duplicated secrets

```
Error from server (AlreadyExists): error when creating \"/tmp/Federation%20with%20Kube%20Global%20on%20federation%20should%20sync%20data%20plane%20proxies%20to%20global%20cp4244208268\": secrets \"default.ca-builtin-key-ca-1\" already exists
```

which might be because we autogenerate the secretes when we an mtls-ed mesh and then we try to import the secrects

PR #13318, opened 2025-04-04
fix(kumactl): change mtls backend builtin to provided and skipvalidation

lukidzi:
## Motivation

When a user exports a mesh with the mTLS type set to builtin, it is exported as is. However, this is incorrect when the user attempts to import it.

During import, this mesh should be treated as provided since the secrets are already included. Additionally, applying a builtin mesh triggers the generation of new secrets, even though existing ones are already available.

## Implementation information

Simplified the logic for better readability:
* Ordered resources: `Mesh` → `Secrets` → ... → `GlobalSecrets`
* Changed mesh from builtin to provided and added `skipValidate` to ensure `Secret` creation can happen after the mesh.
* Enabled federation test.
* Fixed Kubernetes export, as `skipCreatingInitialPolicies` was not being set for the mesh.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13109
Fix: https://github.com/kumahq/kuma/issues/13217


pkg/core/resources/model/resource.go
```diff
@@ -136,6 +136,13 @@ func Deprecations(resource Resource) []string {
 	return nil
 }
 
+func Warnings(resource Resource) []string {
```
slonka, line 139: why is this needed?

slonka (approved):
In general looks good, should we run this with `ci/verify-stability`?

slonka (approved):

lukidzi:
@lahabana you can take a look and let know if it's ok

slonka:
@lukidzi should this be backported?

lukidzi:
No, I don't think we need to backport this since we introduced it in 2.10. Exporting 2 times mesh it's not breaking anything so I don't see a need

Issue #13289, opened 2025-04-01
Set `allowPrivilegeEscalation` to `false` for kuma-dp


lukidzi:
### Description

We don’t explicitly set the `allowPrivilegeEscalation` value to `false`. Since `kuma-dp` shouldn’t require it, we can disable it by default. Users can always override this using `ContainerPatch`.

Docs:
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
https://github.com/kubernetes/design-proposals-archive/blob/main/auth/no-new-privs.md#changes-of-securitycontext-objects
https://github.com/kubernetes/website/issues/30104

PR #13320, opened 2025-04-04
feat(k8s): set allowPrivilegeEscalation to false

lukidzi:
## Motivation

We don't set `allowPrivilegeEscalation`, but it's unnecessary since our process runs with a different UID, and the `kuma-dp` container does not require escalation.

## Implementation information

Set explicitly `allowPrivilegeEscalation` to `false`.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13289


bartsmykla (approved):

lahabana:
@lukidzi can we set a note in UPGRADE.md please?

Issue #13299, opened 2025-04-02
Sometimes init containers (`kuma-init`, `kuma-validation`) take much longer to finish


bartsmykla:
### Kuma Version

2.10.1

### Describe the bug

From time to time, `kuma-init` and `kuma-validation` init containers take significantly longer (up to 2 minutes) to complete.

We observed the following log during container startup:
```
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
Detected at:
    >  goroutine 1 [running, locked to thread]:
    >  runtime/debug.Stack()
    >      runtime/debug/stack.go:26 +0x64
    >  sigs.k8s.io/controller-runtime/pkg/log.eventuallyFulfillRoot()
    >      sigs.k8s.io/controller-runtime@v0.20.4/pkg/log/log.go:60 +0xf4
    >  sigs.k8s.io/controller-runtime/pkg/log.(*delegatingLogSink).WithName(0x4000415640, {0x2f57975, 0x7})
    >      sigs.k8s.io/controller-runtime@v0.20.4/pkg/log/deleg.go:147 +0x34
    >  github.com/go-logr/logr.Logger.WithName(...)
    >      github.com/go-logr/logr@v1.4.2/logr.go:345
    >  sigs.k8s.io/controller-runtime/pkg/metrics/server.init()
    >      sigs.k8s.io/controller-runtime@v0.20.4/pkg/metrics/server/doc.go:26 +0x54
```

This results in init containers being delayed by up to 2 minutes which slows down pod readiness.

This is more frequent when the new transparent proxy configuration via ConfigMaps is enabled.

### To Reproduce

1. Install Kuma with `--set transparentProxy.configMap.enabled=true`  
2. Repeatedly delete pods that are part of the mesh until the issue is observed

### Expected behavior

- Init containers should consistently finish in expected short time
- Logging should be initialized properly to avoid delays

PR #13337, opened 2025-04-07
fix(kumactl): avoid logger issues by importing controller-runtime directly

bartsmykla:

Importing the `controller-runtime/log` subpackage causes global `Log` to be  assigned to the delegated logger, which depends on later `SetLogger` calls. When `controller-runtime` is also imported elsewhere, this re-triggers global init logic, leading to data races and inconsistent logger behavior.

This affects performance and can cause commands like `kumactl install transparent-proxy` or `kumactl install transparent-proxy-validator` to hang for several minutes.

By importing `controller-runtime` directly and using its `Log`/`SetLogger` re-exports, we avoid duplicate initialization and ordering issues.

## Supporting documentation

Closes: https://github.com/kumahq/kuma/issues/13299

### DIFF
--- a/pkg/core/alias.go
+++ b/pkg/core/alias.go
@@ -8,18 +8,18 @@ import (
 	"time"
 
 	"github.com/google/uuid"
-	kube_log "sigs.k8s.io/controller-runtime/pkg/log"
+	kube_ctrl "sigs.k8s.io/controller-runtime"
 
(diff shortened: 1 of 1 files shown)


slonka (approved):

slonka:
It would be nice to include the strace info showing how this was identified.

Automaat (approved):

Issue #3233, opened 2021-11-22
Unify casing in policies


lahabana:
### What happened?

We've got traffic-logs but proxytemplates, traffic-permissions but healthchecks. Casing should be consistent. Also check API urls.

If we change this we will also need to update the GUI.

lahabana:
let's go like k8s .i.e no dashes

lahabana:
xref #3336

lahabana:
We also have `zone-ingress` and `zoneegress` it should be `zoneingress`

lahabana:
Fix this as part of #3330 when we slowly introduce the new polices

PR #13358, opened 2025-04-08
test(e2e): fix Ingress config fetch type

lukidzi:
## Motivation

Our api for kumactl is not consistant and there is no `zoneingresses`, instead there is `zone-ingresses`. We cannot change api since it's a breaking change maybe we could introduce `zoneingresses` as an alias https://github.com/kumahq/kuma/blob/master/api/mesh/v1alpha1/zone_ingress.proto#L23

> Changelog: skip

We should probably cover it https://github.com/kumahq/kuma/issues/3233

### DIFF
--- a/test/framework/debug.go
+++ b/test/framework/debug.go
@@ -208,7 +208,7 @@ type dpType string
 const (
 	dataplaneType   dpType = "dataplane"
 	zoneegressType  dpType = "zoneegress"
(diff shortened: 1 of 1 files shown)


jijiechen (approved):
Approved if it's validated manually. Thanks for the finding and I'm a little surprised that we did not find it earlier. (Probably I was the original author, which is awkward)

Issue #13372, opened 2025-04-10
Reduce WebHooks scope


lukidzi:
### Description

We should limit scope of webhooks. https://github.com/Kong/kong-mesh/blob/master/docs/madr/decisions/015-limit-cp-access-to-k8s-resources.md

We want to watch only for existence of specific label `kuma.io/sidecar-injection`.

PR #13378, opened 2025-04-10
feat(helm): limit webhooks only to namespaces with kuma.io/sidecar-injection label

lukidzi:
## Motivation

Webhooks currently use a very broad namespace selector that only excludes the `kube-system` namespace. To improve both security and performance, we’ve decided to restrict webhooks to only watch namespaces that have the `kuma.io/sidecar-injection` label.

## Implementation information

* Update the namespace selector to include only namespaces labeled with `kuma.io/sidecar-injection`, or `kuma-system` (for mesh defaulter/owner reference cases).
* Add an UPGRADE.md section with a Bash script that checks for pods participating in the mesh that are located in namespaces without the `kuma.io/sidecar-injection` label

Needs: https://github.com/kumahq/kuma/pull/13377

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13372

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -20,6 +20,30 @@ To simplify the namespace selector logic in webhooks, we now require the `kuma.i
 
 Since Kubernetes v1.22, the API server automatically adds the `kubernetes.io/metadata.name` label to all namespaces. As a result, we’ve replaced the use of the custom `kuma.io/system-namespace` label in the secret webhook selector with this standard label.
 
+### Namespaces that are part of the Mesh requires `kuma.io/sidecar-injection` label to exist
+
+Since version 2.11.x, to improve performance and security, we require that each namespace participating in the Mesh has the `kuma.io/sidecar-injection` label set.
+
+Before upgrading, check if you are running any deployments with the `kuma.io/sidecar-injection: true` or `enabled` label in namespaces that do not have the `kuma.io/sidecar-injection` label set. If so, add `kuma.io/sidecar-injection: false` or `disabled` to those namespaces.
+
+You can use this script
+```bash
+for ns in $(kubectl get ns -o jsonpath='{.items[*].metadata.name}'); do
+  ns_label=$(kubectl get ns "$ns" -o jsonpath='{.metadata.labels.kuma\.io/sidecar-injection}' 2>/dev/null)
+  if [ -z "$ns_label" ]; then
+    kubectl get pods -n "$ns" --show-labels --no-headers 2>/dev/null | \
+      grep 'kuma.io/sidecar-injection' | \
+      awk -v ns="$ns" '{print ns "/" $1}'
+  fi
+done
+```
+
+You can later patch namespaces with the following command:
+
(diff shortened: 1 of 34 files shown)


UPGRADE.md
```diff
@@ -20,6 +20,28 @@ To simplify the namespace selector logic in webhooks, we now require the `kuma.i
 
 Since Kubernetes v1.22, the API server automatically adds the `kubernetes.io/metadata.name` label to all namespaces. As a result, we’ve replaced the use of the custom `kuma.io/system-namespace` label in the secret webhook selector with this standard label.
 
+### Namespaces that are part of the Mesh requires `kuma.io/sidecar-injection` label to exist
+
+Since version 2.11.x, to improve performance and security, we require that each namespace participating in the Mesh has the `kuma.io/sidecar-injection` label set.
```
slonka, line 25:
```suggestion
Since version 2.11.x, to improve performance and security, each namespace participating in the Mesh is required to have the `kuma.io/sidecar-injection` label set.
```

UPGRADE.md
```diff
@@ -20,6 +20,28 @@ To simplify the namespace selector logic in webhooks, we now require the `kuma.i
...
+
+Before upgrading, check if you are running any deployments with the `kuma.io/sidecar-injection: true` or `enabled` label in namespaces that do not have the `kuma.io/sidecar-injection` label set. If so, add `kuma.io/sidecar-injection: false` or `disabled` to those namespaces.
```
slonka, line 27: nitpick: I think `disabled` to mean `false` is discouraged - isn't it? If so, I wouldn't recommend setting it.

```suggestion
Before upgrading, check whether any deployments are using the `kuma.io/sidecar-injection: true` or `enabled` label in namespaces that do not have the `kuma.io/sidecar-injection` label set. If so, add `kuma.io/sidecar-injection: false` to those namespaces.
```

UPGRADE.md
```diff
@@ -20,6 +20,28 @@ To simplify the namespace selector logic in webhooks, we now require the `kuma.i
...
+
+You can use this script
```
slonka, line 29:
```suggestion
You can use this script to detect such namespaces:
```

UPGRADE.md
```diff
@@ -20,6 +20,28 @@ To simplify the namespace selector logic in webhooks, we now require the `kuma.i
...
+      grep 'kuma.io/sidecar-injection' | \
+      awk -v ns="$ns" '{print ns "/" $1}'
+  fi
+done
+```
+
+You can later patch namespaces with the following command:
+
+```bash
+kubectl label namespace NAMESPACE_NAME kuma.io/sidecar-injection=disabled
```
slonka, line 44: I think it's a bad merge or something but this point shouldn't be in the middle of another point.

slonka (changes requested):
Some suggestions for upgrade.md and there is definitely something that got merged badly because one point is in the middle of another

slonka (approved):

Issue #13371, opened 2025-04-10
Limit rbac permissions for control-plane


lukidzi:
### Description

We should avoid using wide permissions in ClusterRole for control-plane. We should implement them based on https://github.com/Kong/kong-mesh/blob/master/docs/madr/decisions/015-limit-cp-access-to-k8s-resources.md

Separate ClusterRole:
1. ClusterRole for control-plane bound by ClusterRoleBinding 
2. ClusterRole for namespaces bound by RoleBinding
3. Role for system-namespace bound by RoleBinding

Also, we should introduce flag `enabledNamespaces` which controls where RoleBinding needs to be created, and once not defined we should create ClusterRoleBinding for all namespaces

PR #13387, opened 2025-04-11
feat(helm): reduce scope of rbac permissions and introduce usage of RoleBinding

lukidzi:
## Motivation

To improve security, we are introducing a new option that allows the use of RoleBinding instead of ClusterRoleBinding. The default ClusterRoleBinding will also have reduced access to cluster resources.

## Implementation information

* Permissions in the default ClusterRole have been reduced.
* The default ClusterRole is now split into:
  * A cluster-scoped role for and read only namespaced.
  * A namespaced role for read and write access within specified namespaces.
* Users can now provide a list of namespaces that are part of the Mesh. When this is configured, we create RoleBindings in those namespaces instead of a single ClusterRoleBinding.

> [!WARNING]
> kubelinter recommend usage of explicit resource names instead of `*` in ClusterRole. At the moment I've disabled this validator since we want to watch all kuma resources and avoid changes to ClusterRole. Let's discuss if we want to use `*` with disabled linter or we should explicitly set them.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13371

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -26,6 +26,15 @@ If you are running helm with `noHelmHooks` please set label on the system namesp
 kubectl label namespace SYSTEM_NAMESPACE kuma.io/sidecar-injection=disabled
 ```
 
+### More Restricted `ClusterRole` for Control Plane and CNI
+
+We have split the `ClusterRole` for the control plane into two parts:
+
+* A cluster-scoped `ClusterRole` with read access to namespaced resources .
+* A `ClusterRole` with write permissions, now scoped more narrowly.
+
+By default, we use a `ClusterRoleBinding` to grant write permissions to the control plane. However, starting with version 2.11.x, you can specify the namespaces where the control plane should have write access by using the `enabledNamespaces` configuration.
+
 ## Upgrade to `2.10.x`
 
 ### API Server behaviour changes
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -14,6 +14,12 @@ In previous versions, Kuma did not explicitly set `allowPrivilegeEscalation`. St
 
 Before upgrading, ensure that your configuration does not override this setting.
 
+### System namespace requires the `kuma.io/sidecar-injection: false` label
+
+To simplify the namespace selector logic in webhooks, we now require the `kuma.io/sidecar-injection: false` label to be set on the system namespace.
+
+Since Kubernetes v1.22, the API server automatically adds the `kubernetes.io/metadata.name` label to all namespaces. As a result, we’ve replaced the use of the custom `kuma.io/system-namespace` label in the secret webhook selector with this standard label.
+
 ## Upgrade to `2.10.x`
 
 ### API Server behaviour changes
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -20,6 +20,12 @@ To simplify the namespace selector logic in webhooks, we now require the `kuma.i
 
 Since Kubernetes v1.22, the API server automatically adds the `kubernetes.io/metadata.name` label to all namespaces. As a result, we’ve replaced the use of the custom `kuma.io/system-namespace` label in the secret webhook selector with this standard label.
 
+If you are running helm with `noHelmHooks` please set label on the system namespace:
+
+```bash
+kubectl label namespace SYSTEM_NAMESPACE kuma.io/sidecar-injection=disabled
+```
+
 ## Upgrade to `2.10.x`
 
 ### API Server behaviour changes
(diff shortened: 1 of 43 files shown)


deployments/charts/kuma/templates/cp-rbac.yaml
```diff
@@ -22,42 +22,37 @@ metadata:
...
       - namespaces
       - pods
       - nodes
+      - services
 {{- if .Values.controlPlane.supportGatewaySecretsInAllNamespaces }}
       - secrets
-{{- end }}
-{{- if and .Values.transparentProxy.configMap.enabled .Values.transparentProxy.configMap.config }}
-      - configmaps
 {{- end }}
```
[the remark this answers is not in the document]
slonka: this not fully done, this is still needed

UPGRADE.md
```diff
@@ -26,6 +26,15 @@ If you are running helm with `noHelmHooks` please set label on the system namesp
...
 ```
 
+### More Restricted `ClusterRole` for Control Plane and CNI
+
+We have split the `ClusterRole` for the control plane into two parts:
+
+* A cluster-scoped `ClusterRole` with read access to namespaced resources .
+* A `ClusterRole` with write permissions, now scoped more narrowly.
+
+By default, we use a `ClusterRoleBinding` to grant write permissions to the control plane. However, starting with version 2.11.x, you can specify the namespaces where the control plane should have write access by using the `enabledNamespaces` configuration.
```
bartsmykla, line 36: I feel like this might be a bit too brief. Could you expand on it and explain the implications for the user? It would be helpful to include what the user is expected to do, if anything.

deployments/charts/kuma/templates/cni-rbac.yaml
```diff
@@ -33,6 +28,27 @@ rules:
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "kuma.name" $root }}-cni
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "kuma.name" $root }}-cni
+    namespace: {{ $root.Release.Namespace }}
+{{- end }}
+{{- end }}
+{{- if eq (len .Values.enabledNamespaces) 0 }}
```
bartsmykla, line 51: Nit: wdyt?

```suggestion
{{- else }}
```

deployments/charts/kuma/values.yaml
```diff
@@ -20,6 +20,10 @@ installCrdsOnUpgrade:
 # -- Whether to disable all helm hooks
 noHelmHooks: false
 
+# -- Namespaces that are part of the Mesh. Required to create RoleBindings in these namespaces.
+# If not specified, a ClusterRoleBinding will be created instead.
```
bartsmykla, line 24: Could you refine this comment a bit? It would be more helpful to focus on what the outcome will be when this is used, rather than explaining why it's required or not

UPGRADE.md
```diff
@@ -26,6 +26,15 @@ If you are running helm with `noHelmHooks` please set label on the system namesp
 kubectl label namespace SYSTEM_NAMESPACE kuma.io/sidecar-injection=disabled
 ```
 
+### More Restricted `ClusterRole` for Control Plane and CNI
+
+We have split the `ClusterRole` for the control plane into two parts:
+
+* A cluster-scoped `ClusterRole` with read access to namespaced resources .
```
bartsmykla, line 33:
```suggestion
* A cluster-scoped `ClusterRole` with read access to namespaced resources.
```

deployments/charts/kuma/templates/cp-rbac.yaml
```diff
@@ -104,75 +103,63 @@ rules:
...
-      - delete
       - list
       - watch
-      - create
+  - apiGroups: # required for taint controller
+      - ""
+    resources:
+      - nodes
+    verbs:
       - update
```
bartsmykla, line 128: I understand we can disable the taint controller by setting the right option in the control plane config. I'm wondering if we could expose that setting in `values.yaml`, so disabling it there would also make these permissions unnecessary. What do you think?

bartsmykla (approved):
One nit and a question, but LGTM

deployments/charts/kuma/templates/cp-rbac.yaml
lukidzi: We set it to true when cni is enabled https://github.com/kumahq/kuma/blob/master/deployments/charts/kuma/templates/_helpers.tpl#L310-L316

bartsmykla: What I meant is that it's possible to run the CNI without the taint controller by setting an env var like `KUMA_RUNTIME_KUBERNETES_NODE_TAINT_CONTROLLER_ENABLED` in the control plane runtime. Maybe we could expose this as a `values.yaml` option so when it's disabled, we also skip adding the related permission. It's not a blocker, just a suggestion to consider.

bartsmykla (dismissed):

PR #13392, opened 2025-04-14
ci(ipv6): disable kind ip v6 because of a kernel bug

slonka:
## Motivation

Because of https://github.com/actions/runner-images/issues/11985

## Implementation information

Add if around "runs-on".

### DIFF
--- a/.github/workflows/_e2e.yaml
+++ b/.github/workflows/_e2e.yaml
@@ -23,7 +23,8 @@ jobs:
     timeout-minutes: 60
     # can't use env vars here
     runs-on: ${{ inputs.runner }}
-    if: ${{ inputs.runner != '' }}
+    # workaround for this bug https://github.com/actions/runner-images/issues/11985 - revert when done
+    if: ${{ fromJSON(inputs.matrix).k8sVersion != 'kindipv6' && inputs.runner != '' }}
     strategy:
       fail-fast: false
       matrix:


bartsmykla (dismissed):

.github/workflows/_e2e.yaml
```diff
@@ -22,7 +22,8 @@ jobs:
   e2e:
     timeout-minutes: 60
     # can't use env vars here
-    runs-on: ${{ inputs.runner }}
+    # workaround for this bug https://github.com/actions/runner-images/issues/11985 - revert when done
+    runs-on: ${{ fromJSON(inputs.matrix).k8sVersion == 'kindipv6' && (inputs.runner == 'ubuntu-latest-kong' && 'ubuntu-22.04-kong' || 'ubuntu-22.04') || inputs.runner }}
```
lukidzi, line 26: I think `ubuntu-22.04-kong` doesn't exist

lukidzi: Check here https://github.com/kumahq/kuma/settings/actions/runners we don't have a ubuntu-22.04 runnner

.github/workflows/_e2e.yaml
```diff
@@ -23,7 +23,8 @@ jobs:
     timeout-minutes: 60
     # can't use env vars here
     runs-on: ${{ inputs.runner }}
-    if: ${{ inputs.runner != '' }}
+    # workaround for this bug https://github.com/actions/runner-images/issues/11985 - revert when done
+    if: ${{ fromJSON(inputs.matrix).k8sVersion != 'kindipv6' && inputs.runner != '' }}
```
lukidzi, line 27: Do we want to skip either on kindIpv6 or empty runner?
```suggestion
    if: ${{ fromJSON(inputs.matrix).k8sVersion != 'kindIpv6' || inputs.runner != '' }}
```

slonka: I read this as: We want to "run" when k8s version is not ipv6 and runner is not empty. I don't think your version is correct. We would run when `inputs.runner` is not empty so also when `kindIpv6`.

slonka: Why do we see 4 skips instead of 2 - https://github.com/kumahq/kuma/actions/runs/14444044925/job/40501259796?pr=13392 🤔

lukidzi (approved):

Issue #13386, opened 2025-04-11
Add a message on change of ClusterRole


lukidzi:
### Description

We should add check with a message to update `UPGRADE.md` on ClusterRole change.

PR #13396, opened 2025-04-14
chore(check): verify that UPGRADE.md has changed after Role change

lukidzi:
## Motivation

We want to avoid frequent changes to RBAC and ensure that users who upgrade and have RBAC creation disabled will see a note about it in the UPGRADE.md.

## Implementation information

I've added a `check/rbac` command to check for changed files, specifically looking for `Role`, `RoleBinding`, `ClusterRole`, or `ClusterRoleBinding`. It also verifies that `UPGRADE.md` has been updated. If any RBAC-related changes are detected, the command will fail with a message.

## Supporting documentation

Fix: https://github.com/kumahq/kuma/issues/13386

### DIFF
--- a/mk/check.mk
+++ b/mk/check.mk
@@ -74,7 +74,7 @@ api-lint:
 		github.com/kumahq/kuma/pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1
 
 .PHONY: check
-check: format lint ## Dev: Run code checks (go fmt, go vet, ...)
+check: format lint check/rbac ## Dev: Run code checks (go fmt, go vet, ...)
 	@untracked() { git ls-files --other --directory --exclude-standard --no-empty-directory; }; \
 	check-changes() { git --no-pager diff "$$@"; }; \
 	if [ $$(untracked | wc -l) -gt 0 ]; then \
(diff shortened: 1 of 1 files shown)


mk/check.mk
```diff
@@ -89,6 +89,15 @@ check: format lint ## Dev: Run code checks (go fmt, go vet, ...)
 	fi; \
 	if [ "$$FAILED" = true ]; then exit 1; fi
 
+.PHONY: check/rbac
+check/rbac:
+	@RBAC_CHANGED=$$(git --no-pager diff --name-only -- deployments/ | xargs grep -E 'kind: (Role|RoleBinding|ClusterRole|ClusterRoleBinding)' || true); \
```
slonka, line 94: I'm not sure this will work, git diff only checks unstaged changed so if a person commits it without running the check first this won't fail. I think we should check changes against `base` branch (mostly `master`, but during the release it's `release-X`, but I think `master` would be good enough).

slonka (changes requested):

mk/check.mk
slonka:
```suggestion
	@RBAC_CHANGED=$$(git --no-pager diff --name-only master -- deployments/ | xargs grep -E 'kind: (Role|RoleBinding|ClusterRole|ClusterRoleBinding)' || true); \
```

mk/check.mk
```diff
@@ -89,6 +89,15 @@ check: format lint ## Dev: Run code checks (go fmt, go vet, ...)
...
+	UPGRADE_CHANGED=$$(git --no-pager diff --quiet UPGRADE.md; [ $$? -ne 0 ] && echo true || echo ""); \
```
slonka, line 95: Did the previous one work? don't we need `--` before UPGRADE.md?

```suggestion
	UPGRADE_CHANGED=$$(git --no-pager diff --quiet master -- UPGRADE.md; [ $$? -ne 0 ] && echo true || echo ""); \
```

slonka (approved):

bartsmykla (approved):

Issue #13338, opened 2025-04-07
Accept transparent proxy config from Dataplane Metadata


bartsmykla:
### Summary

Update the control plane to support receiving transparent proxy configuration via Dataplane Metadata instead of the Dataplane resource definition.

### Background

As part of the work described in the MADR [*Transparent Proxy ConfigMap Handling Outside the Control Plane*](https://github.com/kumahq/kuma/blob/a2a5110b718bf5b4d8312c57cbb76ea2af7d1148/docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md), we are moving toward configuring transparent proxy settings via CLI and passing them through Dataplane Metadata. This change is necessary to remove the requirement for the control plane to access ConfigMaps in all namespaces.

### Tasks

- [x] Update the control plane to read transparent proxy settings from Dataplane Metadata
- [x] Ensure the metadata-based config takes precedence over values defined directly in the Dataplane resource

### Acceptance Criteria

- Metadata-based config is preferred when present
- Old config path remains supported for compatibility
- Functionality is verified through tests and documentation is updated accordingly

Issue #13408, opened 2025-04-16
Deprecate inline transparent proxy config in Dataplane resource


bartsmykla:
### Summary

Log a warning when deprecated transparent proxy fields are used in the Dataplane resource.

### Background

To encourage migration to metadata-based configuration for transparent proxy settings, the control plane should log a warning whenever deprecated fields are used in the Dataplane resource. This helps users identify configurations that should be updated.

### Tasks

- [ ] Detect usage of deprecated fields: `redirectPortInbound`, `redirectPortOutbound`, `ipFamilyMode`
- [ ] Log a warning when any of these fields are present in the Dataplane resource
- [ ] Switch e2e tests to use the new way of providing transparent proxy config
- [ ] Update user-facing docs and reference `UPGRADE.md` if needed

### Acceptance Criteria

- A clear warning is logged when deprecated fields are used
- Documentation includes a note about the deprecated fields and the recommendation to migrate to metadata-based config

lahabana:
At the moment the feature is experimental. Once it's on by default we can start warning about the deprecation

lahabana:
Is this on by default? If it's not on by default maybe 2.14 should just suggest that folks should switch as the new option should be the default in 3.0

PR #13409, opened 2025-04-16
feat(kuma-dp): add support for dynamic transparent proxy config

bartsmykla:
## Motivation

As part of the work described in the MADR [*Transparent Proxy ConfigMap Handling Outside the Control Plane*](https://github.com/kumahq/kuma/blob/a2a5110b718bf5b4d8312c57cbb76ea2af7d1148/docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md), we are moving toward configuring transparent proxy settings via CLI and passing them through Dataplane Metadata. This change is necessary to remove the requirement for the control plane to access ConfigMaps in all namespaces.

## Implementation information

This introduces full support for configuring transparent proxy behavior dynamically via the `--transparent-proxy` and `--transparent-proxy-config` flags in `kuma-dp run`.

Key changes:
- Adds `DataplaneConfig` struct with validation, defaults, and feature flags
- Sets proxy redirect, DNS, and IP family settings based on the provided config
- Detects OS and ensures transparent proxy is only allowed on Linux
- Adds support for embedding the config into xDS bootstrap metadata
- Refactors `IsUsingTransparentProxy()` to a config-aware `.Enabled()` check
- Introduces dynamic per-dataplane feature `feature-transparent-proxy-in-dataplane-metadata`
- Ensures config is passed to listeners, metrics, generators, and policies
- Improves CLI flag handling to support inline YAML/JSON or file-based config input

## Supporting documentation

Related to: https://github.com/kumahq/kuma/issues/13338

### DIFF
--- a/pkg/transparentproxy/config/dataplane/config_dataplane.go
+++ b/pkg/transparentproxy/config/dataplane/config_dataplane.go
@@ -1,0 +1,164 @@
+package dataplane
+
+import (
+	"github.com/asaskevich/govalidator"
+	"golang.org/x/exp/constraints"
+
+	core_config "github.com/kumahq/kuma/pkg/config"
+	xds_types "github.com/kumahq/kuma/pkg/core/xds/types"
+	tproxy_config "github.com/kumahq/kuma/pkg/transparentproxy/config"
+	"github.com/kumahq/kuma/pkg/util/pointer"
+)
+
+type PortLike interface {
+	constraints.Integer | constraints.Float | tproxy_config.Port
+}
+
+type DataplaneConfigGetter interface {
+	GetTransparentProxy() *DataplaneConfig
+}
+
+type DataplaneResourcer interface {
+	DataplaneConfigGetter
+	GetAddress() string
+}
+
+type DataplaneMetadater interface {
+	DataplaneConfigGetter
+	HasFeature(string) bool
+	GetDNSPort() uint32
+}
+
+type DatalpaneTrafficFlow struct {
+	Enabled bool               `json:"enabled"`
+	Port    tproxy_config.Port `json:"port"`
+}
+
+func NewDataplaneTrafficFlow[T PortLike](enabled bool, port T) DatalpaneTrafficFlow {
+	return DatalpaneTrafficFlow{
+		Enabled: enabled,
+		Port:    tproxy_config.Port(port),
+	}
+}
+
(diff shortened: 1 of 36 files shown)


pkg/transparentproxy/config/dataplane/config_dataplane.go
```diff
@@ -0,0 +1,164 @@
...
+	"golang.org/x/exp/constraints"
+
+	core_config "github.com/kumahq/kuma/pkg/config"
+	xds_types "github.com/kumahq/kuma/pkg/core/xds/types"
+	tproxy_config "github.com/kumahq/kuma/pkg/transparentproxy/config"
+	"github.com/kumahq/kuma/pkg/util/pointer"
+)
+
+type PortLike interface {
+	constraints.Integer | constraints.Float | tproxy_config.Port
```
slonka, line 14: Why `Float`? 🤔

pkg/transparentproxy/config/dataplane/config_dataplane_test.go
```diff
@@ -0,0 +1,146 @@
...
+		})
+	})
+
+	Describe("GetDataplaneConfig", func() {
+		It("should return fallback if nil", func() {
+			cfg := tproxy_dp.GetDataplaneConfig(nil, nil)
+			Expect(cfg).ToNot(BeNil())
+		})
+
+		It("should use meta and dp values", func() {
```
slonka, line 99: are we asserting here on a dp value? I'm not seeing an assert on `address` 🤔

slonka (approved):
Nothing obviously wrong. Just 2 Qs.

pkg/transparentproxy/config/dataplane/config_dataplane.go
bartsmykla: When converting to `map[string]any` for `structpb.Struct`, numbers are stored as `float64` and stay that way when converting back.

pkg/transparentproxy/config/dataplane/config_dataplane_test.go
bartsmykla: We are not, at least directly, because address is used only in underlying `Enabled()` method

bartsmykla:
> When do we start using this in e2e? 🤔

I'm planning to do it as a last step: https://github.com/kumahq/kuma/issues/13408

Issue #13359, opened 2025-04-08
Naming for resource is not following k8s pattern


lukidzi:
### Description

As part of [kumahq/kuma#3233](https://github.com/kumahq/kuma/issues/3233), we agreed to follow Kubernetes-style resource naming conventions: zoneegress, meshproxypatch, etc. Unfortunately, our legacy resource uses the dash in the name: `zone-ingress` and `global-secret`. We should find a way to rename it by providing backward compatibility.

lahabana:
This really needs further testing

PR #13418, opened 2025-04-16
feat(api): rename zone-ingress/global-secrets to k8s conventions name

lukidzi:
## Motivation

While working on the missing ingress config dump, I noticed a lack of consistency in resource names. The naming differs between inspect and get. We should be consistent, especially since we previously made a change for this purpose, but this resource wasn't covered at the time.

## Implementation information

Introduced a new field in the description that allows users to provide an alias for the path/kumactl command. We then create endpoints for the alias and also include it in the kumactl command.

It shouldn't break the old kumactl since we have 2 endpoints `zone-ingresses` and `zoneingresses`, but the new client does requests to `zoneingresses`.

Other ideas:
Should we return a 301 when someone calls zone-ingresses instead of registering two endpoints? 

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13359


lahabana (commented):
Ho and the resources were correctly named on k8s so this doesn't look too dramatic indeed.

Let's just double check no backward compat it broken (probably worth pushing a test Friday on this).

PR #13422, opened 2025-04-17
chore(kuma-dp): improve transparent proxy config handling and JSON serialization

bartsmykla:
## Motivation

Empty `BaseConfig` field was incorrectly being passed in the metadata

## Implementation information

- Allow transparent proxy on non-linux during tests by checking testing mode
- Ignore `BaseConfig` field when serializing `DataplaneConfig` to JSON
- Skip struct fields tagged with `json:"-"` in `StructToMapOfAny`

## Supporting documentation

Related to: https://github.com/kumahq/kuma/issues/13338

### DIFF
--- a/app/kuma-dp/cmd/run.go
+++ b/app/kuma-dp/cmd/run.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"testing"
 	"time"
 
 	envoy_bootstrap_v3 "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v3"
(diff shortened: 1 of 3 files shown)


app/kuma-dp/cmd/run.go
```diff
@@ -75,7 +76,7 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
 
 			var tpCfg *tproxy_dp.DataplaneConfig
 			if len(tpCfgValues) > 0 || tpEnabled {
-				if runtime.GOOS != "linux" {
+				if !testing.Testing() && runtime.GOOS != "linux" {
```
slonka, line 79: why do we need this?

slonka (approved):

app/kuma-dp/cmd/run.go
bartsmykla: I'll add tests for `kuma-dp run` with the tproxy flags, but that probably belongs in the other PR, not this one.

PR #13426, opened 2025-04-17
fix(secrets): correctly compare other meshes secrets when using crossMesh

lukidzi:
## Motivation

While investigating test failures, I noticed a flake that caused the secret not to be loaded.

## Implementation information

While reviewing the code, I found that we were only comparing the number of meshes, but not verifying whether a specific mesh still exists. I added a change to detect cases where a mesh was removed and another was added in a short period of time. I also added a test and a metric to track mesh generation changes.

### DIFF
--- a/pkg/xds/secrets/secrets.go
+++ b/pkg/xds/secrets/secrets.go
@@ -75,20 +75,11 @@ func NewSecrets(caProvider CaProvider, identityProvider IdentityProvider, metric
 		return nil, err
 	}
 
-	otherMeshesCertGenerationsMetric := prometheus.NewCounterVec(prometheus.CounterOpts{
-		Help: "Number of generated certificates for other meshes",
-		Name: "cert_generation_other_meshes",
-	}, []string{"mesh", "other_mesh"})
-	if err := metrics.Register(otherMeshesCertGenerationsMetric); err != nil {
-		return nil, err
-	}
-
 	return &secrets{
-		caProvider:                       caProvider,
-		identityProvider:                 identityProvider,
-		cachedCerts:                      map[certCacheKey]*certs{},
-		certGenerationsMetric:            certGenerationsMetric,
-		otherMeshesCertGenerationsMetric: otherMeshesCertGenerationsMetric,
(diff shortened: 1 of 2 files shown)


pkg/xds/secrets/secrets.go
```diff
@@ -75,11 +75,20 @@ func NewSecrets(caProvider CaProvider, identityProvider IdentityProvider, metric
 		return nil, err
 	}
 
+	otherMeshesCertGenerationsMetric := prometheus.NewCounterVec(prometheus.CounterOpts{
```
slonka, line 78: feels like this metric is only used for assertions in tests, what we're actually trying to test is `shouldGenerateCerts` - should we just restructure the struct? I know it's already been written that way but it seems kind of clunky to add a metric to cover a test case 🤷

pkg/xds/secrets/secrets_test.go
```diff
@@ -190,13 +202,34 @@ var _ = Describe("Secrets", Ordered, func() {
...
+				// given
+				defaultMesh := newMesh("default")
+
+				// when
+				_, _, err := secrets.GetForDataPlane(context.Background(), newDataplane(), defaultMesh, []*core_mesh.MeshResource{newMesh("mesh-1")})
+
+				// then
+				Expect(err).ToNot(HaveOccurred())
+				Expect(test_metrics.FindMetric(metrics, "cert_generation").GetCounter().GetValue()).To(Equal(1.0))
+				Expect(test_metrics.FindMetric(metrics, "cert_generation_other_meshes", "other_mesh", "mesh-1").GetCounter().GetValue()).To(Equal(1.0))
```
slonka, line 219: as mentioned above - instead of asserting on the metric is it possible to assert on something else? or maybe we should move `shouldGenerateCerts` to a separate module, make it public and test it there? 🤔

slonka (commented):
In general great catch 🎉  - I'm just not 100% convinced about the how we test this module in general

Icarus9913 (approved):
LGTM

PR #13428, opened 2025-04-18
refactor(kuma-dp): remove `BootstrapParams` to simplify bootstrap config fetch

bartsmykla:

- Drop `BootstrapParams` in favor of using `Opts` directly in bootstrap client
- Replace `BootstrapGenerator` function with `BootstrapClient` interface to remove confusion with `BootstrapGenerator` used by control plane
- Refactor remote bootstrap logic to move version check and resource preparation internally
- Simplify test setup with `optsBuilder` and `fakeBootstrapClient`


app/kuma-dp/cmd/run_test.go
```diff
@@ -125,12 +118,11 @@ var _ = Describe("run", func() {
 			// then
 			var actualConfigFile string
 			envoyPid := verifyComponentProcess("Envoy", envoyPidFile, envoyCmdlineFile, func(actualArgs []string) {
-				Expect(actualArgs[0]).To(Equal("--version"))
```
slonka, line 128: Why is the version no longer there? I assume it was not needed 🤔

app/kuma-dp/pkg/dataplane/envoy/remote_bootstrap_test.go
```diff
@@ -81,81 +62,59 @@ var _ = Describe("Remote Bootstrap", func() {
...
+			given.optsBuilder.cpURL(server.URL).build(),
+			given.metadata,
+		)
 
 		// then
 		Expect(err).ToNot(HaveOccurred())
 		Expect(bootstrap).ToNot(BeNil())
 		Expect(kumaSidecar).ToNot(BeNil())
 	},
-		Entry("should support port range with exactly 1 port",
```
slonka, line 98: are these no longer needed?

slonka (approved):

app/kuma-dp/cmd/run_test.go
bartsmykla: That’s because version retrieval was moved to the `BootstrapClient`, and in these tests, we’re using a fake client that doesn’t perform that logic.

app/kuma-dp/pkg/dataplane/envoy/remote_bootstrap_test.go
bartsmykla: These didn’t make sense, there’s no port range involved. I updated the descriptions, removed the duplicate, and improved the golden file names.

Issue #13353, opened 2025-04-08
Investigate and reenable DP DNS e2e tests


lahabana:
### Description

There's a flake on MeshPassthrough with DP DNS on delegated gateway.

For .e.g: https://github.com/kumahq/kuma/actions/runs/14312181078

PR #13457, opened 2025-04-24
fix(dns): enable e2e tests with embedded dns and reduce TTL

lukidzi:
## Motivation

We want to switch to embedded DNS but we had some flakes.

## Implementation information

I've noticed that we fail in the following scenario:

1. We create an ExternalService, which allocates 240.0.0.3 for a real domain (something we don't do for MES).
2. Kong Gateway makes a request to the domain and receives a response with 240.0.0.3.
3. We update the DNS configuration and remove this DNS entry (since we're testing passthrough and removed the ExternalService).
4. Kong Gateway has cached the IP, and we no longer see logs indicating requests to the DNS server.

The default TTL is 3600s, so the cached record isn't refreshed.
This creates a potential issue when a service is removed and later redeployed with a new VIP address. Due to the long TTL, Kong might continue using the stale IP. A shorter TTL would help mitigate this risk.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13353

### DIFF
--- a/pkg/xds/generator/dns_generator.go
+++ b/pkg/xds/generator/dns_generator.go
@@ -53,7 +53,7 @@ func (g DNSGenerator) Generate(ctx context.Context, rs *core_xds.ResourceSet, xd
 		}
 	}
 	if proxy.Metadata.HasFeature(xds_types.FeatureEmbeddedDNS) {
-		dnsInfo := dpapi.DNSProxyConfig{TTL: 3600, Records: []dpapi.DNSRecord{}}
+		dnsInfo := dpapi.DNSProxyConfig{TTL: 30, Records: []dpapi.DNSRecord{}}
 		for name, addresses := range vips {
 			dnsInfo.Records = append(dnsInfo.Records, dpapi.DNSRecord{
 				Name: name,
--- a/.github/workflows/_test.yaml
+++ b/.github/workflows/_test.yaml
@@ -71,6 +71,12 @@ jobs:
                 "include":[
                   {"dpDNS": "true", "k8sVersion": "${{ env.K8S_MAX_VERSION }}", "target": "kubernetes", "arch": "amd64"},
                   {"dpDNS": "true", "k8sVersion": "kind", "target": "universal", "arch": "amd64"},
+                  {"sidecarContainers": "sidecarContainers", "k8sVersion": "${{ env.K8S_MAX_VERSION }}", "target": "kubernetes", "arch": "amd64"},
+                  {"k8sVersion": "${{ env.K8S_MIN_VERSION }}", "target": "multizone", "arch": "amd64"},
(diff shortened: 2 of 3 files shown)


pkg/xds/generator/dns_generator.go
```diff
@@ -53,7 +53,7 @@ func (g DNSGenerator) Generate(ctx context.Context, rs *core_xds.ResourceSet, xd
 		}
 	}
 	if proxy.Metadata.HasFeature(xds_types.FeatureEmbeddedDNS) {
-		dnsInfo := dpapi.DNSProxyConfig{TTL: 3600, Records: []dpapi.DNSRecord{}}
+		dnsInfo := dpapi.DNSProxyConfig{TTL: 30, Records: []dpapi.DNSRecord{}}
```
lahabana, line 56: I think I set 3600 because that's like a DNS common default. So I'd add a comment:

```suggestion
                // This is purposefully set to 30s to avoid DNS cache stale with ExternalService and Kong Gateway see: https://github.com/kumahq/kuma/issues/13353.
		dnsInfo := dpapi.DNSProxyConfig{TTL: 30, Records: []dpapi.DNSRecord{}}
```

lahabana (commented):
So IIUC it's different e2e tests using the same hostname as the external service which causes Kong Gateway to keep the internal DNS cached as a VIP.

Random question related to this and DNS caching. Is there a risk of things like this happening when deleting and recreating MeshService?

.e.g:

1. create MS a --> VIP assigned 240.0.0.1
2. create MS b --> VIP assigned 240.0.0.2
3. delete MS a --> VIP 240.0.0.1 freed
4. create MS c --> VIP assigned 240.0.0.1
5. recreate MS a --> VIP assigned 240.0.0.3

If this happens in a short window the client may cache MS a as 240.0.0.1 when in practice that's MS c and MS a is 240.0.0.3.

This is a edge case because usually you don't recreate MS but it might be worth having an issue to track DNS TTL in relation with VIP allocation.

lukidzi:
> So IIUC it's different e2e tests using the same hostname as the external service which causes Kong Gateway to keep the internal DNS cached as a VIP.
> 
> Random question related to this and DNS caching. Is there a risk of things like this happening when deleting and recreating MeshService?
> 
> .e.g:
> 
> 1. create MS a --> VIP assigned 240.0.0.1
> 2. create MS b --> VIP assigned 240.0.0.2
> 3. delete MS a --> VIP 240.0.0.1 freed
> 4. create MS c --> VIP assigned 240.0.0.1
> 5. recreate MS a --> VIP assigned 240.0.0.3
> 
> If this happens in a short window the client may cache MS a as 240.0.0.1 when in practice that's MS c and MS a is 240.0.0.3.
> 
> This is a edge case because usually you don't recreate MS but it might be worth having an issue to track DNS TTL in relation with VIP allocation.

Yes, it can be the case

PR #13458, opened 2025-04-25
docs(MADR): update resource identifier format limitations

Icarus9913:
## Motivation

Since we have some resource name limitations, we should sync them to the resource identifier MADR

> Changelog: skip

### DIFF
--- a/docs/madr/decisions/070-resource-identifier.md
+++ b/docs/madr/decisions/070-resource-identifier.md
@@ -268,18 +268,23 @@ resource-identifier = *(ALPHA / DIGIT / "-" / "." / "_" / "~" )
 
 As a delimiter we can use only characters that can't be present in resource identifier components – `name`, `mesh`, `namespace`, `zone`, `resourceType` and `sectionName`:
 
-* `name` and `zone`
+* `name`
   * contain no more than 253 characters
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/070-resource-identifier.md
```diff
@@ -268,18 +268,23 @@ resource-identifier = *(ALPHA / DIGIT / "-" / "." / "_" / "~" )
 
 As a delimiter we can use only characters that can't be present in resource identifier components – `name`, `mesh`, `namespace`, `zone`, `resourceType` and `sectionName`:
 
-* `name` and `zone`
+* `name`
   * contain no more than 253 characters
   * contain only lowercase alphanumeric characters, '-' or '.'
```
Icarus9913, line 273: @lobkovilya What does the `name` mean here? Is it a Dataplane resource name or another resource name?

lobkovilya: every resource has `name`, so it's not just Dataplane's name

lobkovilya (approved):

docs/madr/decisions/070-resource-identifier.md
Icarus9913: Let me think about it twice.

PR #13466, opened 2025-04-25
feat(helm): scope webhooks to specific namespaces

### LINKED ISSUE #13460: Configure webhooks with namespaces once provided
### Description

We configure webhooks to watch namespaces labeled with `kuma.io/sidecar-injection`. However, even after a user defines which namespaces are part of the mesh, resources may still be created in namespaces that are not part of the mesh. This behavior can be confusing and unexpected. In this case, we should configure the webhooks to watch only the specified namespaces in.

lukidzi:
## Motivation

Once a user provides `namespaceAllowList` it makes sense to scope webhooks only to these namespaces

## Implementation information

Once `namespaceAllowList` is provided, configure webhook to watch only these namespaces

## Supporting documentation

Fix #13460

### DIFF
--- a/deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
+++ b/deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
@@ -62,6 +62,10 @@ webhooks:
       matchExpressions:
         - key: kuma.io/sidecar-injection
           operator: Exists
+        - key: kubernetes.io/metadata.name
+          operator: NotIn
+          values:
+          - kube-system
       {{- if gt (len .Values.namespaceAllowList) 0 }}
         - key: kubernetes.io/metadata.name
           operator: In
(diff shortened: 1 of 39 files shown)


deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
```diff
@@ -62,10 +62,20 @@ webhooks:
...
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values:
+          - {{ .Release.Namespace }}
+      {{- range $namespace := .Values.namespaceAllowList }}
+          - {{ $namespace }}
+      {{- end }}
+      {{- else }}
         - key: kubernetes.io/metadata.name
           operator: NotIn
```
lobkovilya, line 75: shouldn't we always exclude `kube-system`?

lobkovilya (approved):

PR #13467, opened 2025-04-25
feat(xds): rework DeltaXds to use metadata instead of proto

### LINKED ISSUE #13461: Remove xds protocol from Dataplane object and use metadata
### Description

We shouldn't use Dataplane object to store it. The information should be sent through the metadata.

lukidzi:
## Motivation

We shouldn't change the proto schema but instead use metadata.

## Implementation information

* Remove proto file and field from Dataplane/ZoneIngress/ZoneEgress. It should be safe since clients (mink) were not able to set this property and only global knew about it
* Introduce a new variable for kuma-dp `KUMA_DATAPLANE_RUNTIME_ENVOY_XDS_TRANSPORT_PROTOCOL_VARIANT` which allow setting specific variant of the xds protocol
* Keep `kuma.io/xds-transport-protocol-variant` in the k8s to simplify delta testing for the user

## Supporting documentation

Fix #13461


Icarus9913 (commented):
@lukidzi Please double check `pkg/plugins/runtime/k8s/controllers/pod_converter.go` PodConverter struct `DeltaXds` property, it seems not used anymore

pkg/config/app/kuma-dp/config.go
```diff
@@ -84,6 +84,12 @@ func (c *Config) Features() []string {
 		base = append(base, xds_types.FeatureTransparentProxyInDataplaneMetadata)
 	}
 
+	switch c.DataplaneRuntime.EnvoyXdsTransportProtocolVariant {
+	case "DELTA_GRPC":
+		base = append(base, xds_types.FeatureDeltaGRPC)
+	case "GRPC":
+		base = append(base, xds_types.FeatureGRPC)
+	}
```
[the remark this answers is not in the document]
lukidzi: I don't think we should. Since once not defined we should use control-plane default (which might not be GRPC)

pkg/xds/auth/callbacks.go
```diff
@@ -84,6 +84,7 @@ func (a *authCallbacks) OnStreamOpen(ctx context.Context, streamID core_xds.Stre
 		ctx:      ctx,
 		resource: nil,
 	}
+	core.Log.V(1).Info("OnStreamOpen", "streamID", streamID)
 	return nil
 }
```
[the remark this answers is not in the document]
lukidzi: I think it's at different order and this is nice to debug auth flow

pkg/xds/bootstrap/parameters.go
```diff
@@ -26,32 +27,32 @@ type AggregateMetricsConfig struct {
...
+	DataplaneToken       string
+	DataplaneTokenPath   string
+	DataplaneResource    string
+	CertBytes            []byte
+	Version              *mesh_proto.Version
+	HdsEnabled           bool
+	DynamicMetadata      map[string]string
+	DNSPort              uint32
+	ProxyType            string
+	Features             xds_types.Features
```
lobkovilya, line 52: make sure to update https://github.com/kumahq/kuma/blob/57a37110fc610c2b718ab71f40d3aca1eff66893/pkg/xds/bootstrap/template_v3.go#L63, as it's no longer a list of strings anymore

pkg/xds/bootstrap/parameters.go
```diff
@@ -26,32 +27,32 @@ type AggregateMetricsConfig struct {
...
+	IsGatewayDataplane   bool
+	Resources            types.ProxyResources
+	SystemCaPath         string
+	TransparentProxy     *tproxy_config.DataplaneConfig
+	UseDeltaXds          bool
```
lobkovilya, line 57: why do we need `UseDeltaXds` if you can always call ` params.Features.HasFeature(xds_types.FeatureDeltaGRPC)`

lukidzi: We need since you can always enable it on the control-plane level and later you need to set it

pkg/config/app/kuma-dp/config.go
```diff
@@ -363,6 +371,15 @@ func (d *DataplaneRuntime) Validate() error {
 	if d.BinaryPath == "" {
 		errs = multierr.Append(errs, errors.Errorf(".BinaryPath must be non-empty"))
 	}
+	if d.EnvoyXdsTransportProtocolVariant != "" {
+		switch d.EnvoyXdsTransportProtocolVariant {
+		case "DELTA_GRPC":
+		case "GRPC":
```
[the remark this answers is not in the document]
lukidzi: Envoy configuration reference to SOTW as `GRPC` so I wanted to keep them the same

Icarus9913 (dismissed):
LGTM

lahabana (commented):
This is the right direction! Just don't need a "negative feature flag" the absence of feature flag means you are not using `delta`

pkg/config/app/kuma-dp/config.go
```diff
@@ -363,6 +369,14 @@ func (d *DataplaneRuntime) Validate() error {
 	if d.BinaryPath == "" {
 		errs = multierr.Append(errs, errors.Errorf(".BinaryPath must be non-empty"))
 	}
+	if d.EnvoyXdsTransportProtocolVariant != "" {
+		switch d.EnvoyXdsTransportProtocolVariant {
+		case "DELTA_GRPC":
```
lobkovilya, line 374: I think it's okay to remove `feature-grpc` as @lahabana mentioned, because it's not indeed a "feature". But if the config has the field `ProtocolVariant`, then I believe it should have 2 values `DELTA_GRPC` and `GRPC`. If the value is `DELTA_GRPC` then kuma-dp sends `feature-delta-grpc`, if it's `GRPC` then kuma-dp sends nothing

lobkovilya (approved):

Issue #13340, opened 2025-04-07
Align `kumactl install transparent-proxy` with new `--config` flag behavior


bartsmykla:
## Summary

Update `kumactl install transparent-proxy` to support the new unified `--config` flag with the same behavior and precedence rules as the `--transparent-proxy-config` flag in `kuma-dp`. Deprecate the old `--config-file` flag.

## Background

As part of the work in the MADR [*Transparent Proxy ConfigMap Handling Outside the Control Plane*](https://github.com/kumahq/kuma/blob/a2a5110b718bf5b4d8312c57cbb76ea2af7d1148/docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md), we aim to standardize how transparent proxy configuration is passed to both the data plane and the init container. This change ensures consistent behavior between `kuma-dp` and `kumactl`.

## Tasks

- [ ] Modify `kumactl install transparent-proxy` to support the new `--config` flag
- [ ] Allow the flag to accept:
  - No value (use default transparent proxy config)
  - Comma-separated list of config file paths
  - A `-` to read config from STDIN
- [ ] Allow the flag to be repeated
- [ ] Merge values in order with later values overriding earlier ones
- [ ] Deprecate the `--config-file` flag and print a warning when it is used
- [ ] Update CLI help text
- [ ] Add tests to cover the new config behavior

## Acceptance Criteria

- The `--config` flag in `kumactl install transparent-proxy` matches the behavior of `--transparent-proxy-config` in `kuma-dp`
- Supported input formats:
  - No value (use default config)
  - Comma-separated file paths
  - `-` for STDIN
- Repeated flags are merged in order, with later values taking precedence
- Using `--config-file` prints a deprecation warning
- CLI help text is updated to reflect the change

PR #13478, opened 2025-04-28
chore(tproxy): improve config loading + add new transparent proxy test cases

bartsmykla:
## Motivation

Transparent Proxy config loading was not working properly according to the MADR. Updated the Transparent Proxy test framework and added new test cases to make sure the `--config` flag behaves correctly.

## Implementation information

> [!NOTE]
> **For reviewers**: First commit has the actual changes, second commit adds only the new transparent proxy golden files

Disallow passing raw YAML/JSON directly. Force using stdin or file inputs for clearer error messages. Fix an issue where invalid YAML was wrongly treated as a file path.

Support installing Transparent Proxy with config from multiple sources. Update tests to pass params (flags, files, stdin) instead of only flags.

## Supporting documentation

Closes: https://github.com/kumahq/kuma/issues/13340

MADR: [Transparent Proxy ConfigMap Handling Outside the Control Plane](https://github.com/kumahq/kuma/blob/a2a5110b718bf5b4d8312c57cbb76ea2af7d1148/docs/madr/decisions/071-transparent-proxy-config-map-handling-outside-the-control-plane.md)


test/transparentproxy/testdata/debian-10-default.ip6tables-legacy.golden
slonka, line 1: why is this empty?

test/transparentproxy/testdata/alpine-3-default.ip6tables-legacy.golden
```diff
@@ -0,0 +1 @@
+executable not found
```
slonka, line 1: 🤔 - so should we remove the test altogether or install the binary? or is the binary not available at all?

slonka (commented):
left a couple of comments

test/transparentproxy/testdata/debian-10-default.ip6tables-legacy.golden
bartsmykla: Because that's what `ip6tables-legacy-save` returns, and it is not new, all existing cases for Debian show the same results:
```sh
~ cat /etc/*release* | grep -i pretty
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
~ ip6tables-legacy-save
~
```

test/transparentproxy/transparentproxy_test.go
```diff
@@ -398,62 +307,27 @@ func genEntriesForImage(
 		}
 	}
 
-	// buildArgs is a helper function that combines test case parameters with
-	// any additional decorators.
-	buildArgs := func(
```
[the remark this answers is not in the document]
bartsmykla: I just inlined it: https://github.com/kumahq/kuma/pull/13478/files/a419c2ae70756269a93aa5ec0f5822b796cbbb45#diff-42d160c5757bd37b4345059c63e6885c10e4e424963fa1d5f984c34d7c7e920eR315-R323

test/transparentproxy/transparentproxy_test.go
```diff
@@ -179,41 +181,22 @@ var _ = Describe("Transparent Proxy", func() {
...
-//   - Asserts that no error occurred during execution.
-//   - If the command does not exit successfully (exit code != 0), it reads the
-//     command output and fails the test with an appropriate message.
-//
-// Args:
-//   - ctx (context.Context): The context for controlling the command execution.
-//   - c (testcontainers.Container): The test container where the command will
-//     be executed.
-//   - flags ([]string): Additional flags to pass to the kumactl install
-//     command.
```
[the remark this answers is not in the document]
bartsmykla: Because these functions are simple and the comments do not add real value in my opinion. I made a mistake adding them earlier.

PR #13479, opened 2025-04-28
feat(oapi): add dataplane openapi spec

slonka:
## Motivation

Needed for the front-end.

## Implementation information

Add to `ProtoTypeToType` map.


tools/resource-gen/pkg/generator/main.go
```diff
@@ -451,6 +451,7 @@ var AdditionalProtoTypes = []reflect.Type{
 
 var ProtoTypeToType = map[string]reflect.Type{
 	"Mesh":        reflect.TypeOf(v1alpha1.Mesh{}),
+	"Dataplane":   reflect.TypeOf(v1alpha1.Dataplane{}),
```
slonka, line 454: this is the only change not generated

PR #13480, opened 2025-04-28
test(e2e): upgrade gateway-api test from 1.2.1 to 1.3.0

lukidzi:
## Motivation

https://github.com/kumahq/kuma/pull/13477#pullrequestreview-2798952463

## Implementation information

we should probably upgrade this

## Supporting documentation

> Changelog: skip


bartsmykla:
Why not do it in the same PR as package upgrade: https://github.com/kumahq/kuma/pull/13477?

PR #13486, opened 2025-04-29
feat(api): use golang error wrap for store error detection

### LINKED ISSUE #2604: Use Go error type wrapping for  store error detection

Icarus9913:
## Motivation

Before this PR, we had some error assertion by matching the error message string like:

```text
func IsResourceNotFound(err error) bool {
	return err != nil && strings.HasPrefix(err.Error(), "resource not found")
}

func IsResourceAlreadyExists(err error) bool {
	return err != nil && strings.HasPrefix(err.Error(), "resource already exists")
```

So, I plan to use golang wrap to do the secondary encapsulation for error and supplement the error assertion.

## Implementation information

1. Define the specific brief error with `ErrIsAlreadyExists = errors.New("already exists")`
2. Use `fmt.Errorf("%w")` to implement the secondary encapsulation for the need.
3. Use `errors.Is()` to do the assertion with the basic error in above step1.

## Supporting documentation

Fix #2604


Icarus9913:
I've checked the https://github.com/kumahq/kuma/tree/2.10.1/pkg/core/rest/errors
I think we can integrate them if needed in the future, then we won't define the errors seperately.

jijiechen:
LGTM, it's a common practice to use error types with `fmt`

pkg/core/resources/store/error.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/kumahq/kuma/pkg/core/resources/model"
+)
+
+var (
+	ErrorInvalidOffset = errors.New("invalid offset")
```
Automaat, line 11:
```suggestion
	ErrInvalidOffset = errors.New("invalid offset")
```

please make naming consistent with other errors

pkg/core/resources/store/error.go
```diff
@@ -0,0 +1,61 @@
...
+	ErrInvalidOffset   = errors.New("invalid offset")
+	ErrIsAlreadyExists = errors.New("already exists")
+	ErrConflict        = errors.New("conflict")
+	ErrNotFound        = errors.New("not found")
+	ErrInvalid         = errors.New("invalid")
+	ErrBadRequest      = errors.New("bad request")
+)
+
+func IsInvalidOffset(err error) bool {
+	return err != nil && errors.Is(err, ErrInvalidOffset)
```
lahabana, line 20: I think you don't need the `err != nil` because that's the code for errors.Is:
```
func Is(err, target error) bool {
	if err == nil || target == nil {
		return err == target
	}

	isComparable := reflectlite.TypeOf(target).Comparable()
	return is(err, target, isComparable)
}
```

pkg/core/resources/store/error.go
```diff
@@ -0,0 +1,61 @@
...
+
+func ErrorResourceInvalid(rt model.ResourceType, name, mesh string) error {
+	return fmt.Errorf("resource %w: type=%q name=%q mesh=%q", ErrInvalid, rt, name, mesh)
+}
+
+func IsResourceInvalid(err error) bool {
+	return err != nil && errors.Is(err, ErrInvalid)
+}
+
+func ErrorBadRequest(msg string) error {
```
lahabana, line 55: BadRequest is a poor name as this component doesn't take in requests, it's just called.

pkg/core/resources/store/error.go
```diff
@@ -0,0 +1,61 @@
...
+
+func ErrorResourceNotFound(rt model.ResourceType, name, mesh string) error {
+	return fmt.Errorf("resource %w: type=%q name=%q mesh=%q", ErrNotFound, rt, name, mesh)
+}
+
+func IsResourceNotFound(err error) bool {
+	return err != nil && errors.Is(err, ErrNotFound)
+}
+
+func ErrorResourceInvalid(rt model.ResourceType, name, mesh string) error {
```
lahabana, line 47: This is a missing a reason (why the resource is invalid)

pkg/core/resources/store/error.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/kumahq/kuma/pkg/core/resources/model"
+)
+
+var (
+	ErrInvalidOffset   = errors.New("invalid offset")
+	ErrIsAlreadyExists = errors.New("already exists")
+	ErrConflict        = errors.New("conflict")
+	ErrNotFound        = errors.New("not found")
+	ErrInvalid         = errors.New("invalid")
+	ErrBadRequest      = errors.New("bad request")
```
lahabana, line 16: Some of these errors are too specific to have their own type, yet they are too generic as well as you can't pass in a reason. For example ErrInvalid can never have a reason so the errors are useless to the end consumer of the api.

lahabana (commented):
Let's start by first writing the whole taxonomy and then we'll update the different error handlers of the consumers of the store api

jijiechen (dismissed):

pkg/core/rest/errors/error_handler.go
```diff
@@ -29,13 +29,13 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
...
 	var kumaErr *types.Error
 	switch {
-	case store.IsResourceNotFound(err) || errors.Is(err, &NotFound{}):
+	case store.IsNotFound(err) || errors.Is(err, &NotFound{}):
 		kumaErr = &types.Error{
 			Status: 404,
 			Title:  title,
 			Detail: "Not found",
 		}
-	case errors.Is(err, &rest.InvalidResourceError{}), errors.Is(err, &registry.InvalidResourceTypeError{}), errors.Is(err, &store.PreconditionError{}), errors.Is(err, &BadRequest{}):
```
Icarus9913, line 38: @johncowen @schogges There's one change for the error, can you check if it will cause any breaking changes?


`name can't be empty`, `namespace must be non-empty` and `name %q must include namespace after the dot, ex. "name.namespace"` error title would be changed from **Bad Request** to **Failed to find a resource** or **Failed to create a resource** or **Failed to update a resource**. But the HttpStatus code still is `400`.

pkg/core/rest/errors/error_handler.go
```diff
@@ -65,17 +65,11 @@ func HandleError(ctx context.Context, response *restful.Response, err error, tit
...
 		kumaErr = &types.Error{
 			Status: 400,
 			Title:  title,
-			Detail: "Invalid offset",
-			InvalidParameters: []types.InvalidParameter{
-				{
-					Field:  "offset",
-					Reason: "Invalid format",
-				},
-			},
```
Icarus9913, line 78: @johncowen @schogges Here, the `offset` error will only remove 
```text
{
  Field: "offset"
  Reason: "Invalid format"
}
```

The HttpStatus code and title would not be changed.

schogges: I don't think it would break anything on our side as long as `status` and `title` are set. We have [defaults](https://github.com/kumahq/kuma-gui/blob/761a17a6b0e3ad6b7b8a4e4186c12311e616d0a4/packages/kuma-gui/src/app/kuma/services/kuma-api/ApiError.ts#L29-L34) for the rest.

pkg/core/rest/errors/error_handler.go
schogges: I don't think that it would break anything as I couldn't find a case where we process the `title` any further than just passing it through.

pkg/api-server/testdata/resources/crud/list_mesh_00.golden.json
```diff
@@ -2,18 +2,6 @@
...
- ],
- "details": "Invalid offset",
- "causes": [
-  {
-   "field": "offset",
-   "message": "Invalid format"
-  }
- ]
+ "detail": "invalid: invalid offset: strconv.Atoi: parsing \"invalid\": invalid syntax",
+ "details": "invalid: invalid offset: strconv.Atoi: parsing \"invalid\": invalid syntax"
```
schogges, line 6: @johncowen correct me if I'm wrong, but `details` is not processed on our side.

schogges: I gave it a shot in our `ApiError.spec` unit test:

```typescript
      new ApiError({
        type: '/std-errors',
        status: 400,
        title: 'Could not retrieve resources',
        detail: 'invalid: invalid offset: strconv.Atoi: parsing "invalid": invalid syntax',
        details: 'invalid: invalid offset: strconv.Atoi: parsing "invalid": invalid syntax',
      }),
      {
        type: '/std-errors',
        status: 400,
        title: 'Could not retrieve resources',
        detail: 'invalid: invalid offset: strconv.Atoi: parsing "invalid": invalid syntax',
        instance: null,
        invalidParameters: [],
      },
```

and it passed.

johncowen:
What is it that is changing here, just copy? Or also the shape of the HTTP response when we get an error.?

pkg/api-server/testdata/resources/crud/list_mesh_00.golden.json
Icarus9913: The `details` is a deprecated property in the structure and it lasts for a couple of versions. I think it would be fine.

johncowen:
We had a quick sync on this, and it sounds like its just the copy changing (and in some places we get empty `invalidParameters` whereas previously it would not be empty)

The GUI doesn't parse any copy we receive in these parameters, and if invalidParameters is empty we just don't loop through them to display them its empty.

So all in all sounds like nothing here will break the GUI

PR #13488, opened 2025-04-30
feat(oapi): new api schema for rules inspect

Automaat:
## Motivation

This is api schema to discuss in MADR: https://docs.google.com/document/d/1EzZpk3wwneIxQNPK7WXJqhW3Y3CS1AWMXIcXix9LEoc/edit?tab=t.g7ooo2ntj4jj

### DIFF
--- a/api/openapi/specs/api.yaml
+++ b/api/openapi/specs/api.yaml
@@ -161,6 +161,118 @@ paths:
           $ref: '#/components/responses/BadRequest'
         '500':
           $ref: '#/components/responses/Internal'
+  /meshes/{mesh}/dataplanes/{name}/_layout:
+    get:
+      operationId: get-dataplanes-layout
+      summary: TODO
+      description: TODO
+      parameters:
+        - in: path
+          name: mesh
+          required: true
+          description: The mesh of the DPP to get the diff for.
+          schema:
+            type: string
+        - in: path
+          name: name
+          required: true
+          description: The name of the DPP within the mesh to get the diff for.
+          schema:
+            type: string
+      responses:
+        '200':
+          $ref: '#/components/responses/DataplaneLayoutResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '500':
+          $ref: '#/components/responses/Internal'
+  /meshes/{mesh}/dataplanes/{name}/_proxy/_rules:
+    get:
+      operationId: get-dataplanes-layout
+      summary: TODO
+      description: TODO
+      parameters:
+        - in: path
+          name: mesh
+          required: true
+          description: The mesh of the DPP to get the diff for.
+          schema:
+            type: string
+        - in: path
+          name: name
+          required: true
+          description: The name of the DPP within the mesh to get the diff for.
+          schema:
+            type: string
+      responses:
+        '200':
+          $ref: '#/components/responses/ProxyRulesResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '500':
+          $ref: '#/components/responses/Internal'
+  /meshes/{mesh}/dataplanes/{name}/_inbounds/{kri}/_rules:
+    get:
+      operationId: get-dataplanes-layout
+      summary: TODO
+      description: TODO
+      parameters:
+        - in: path
+          name: mesh
+          required: true
+          description: The mesh of the DPP to get the diff for.
+          schema:
+            type: string
+        - in: path
+          name: name
+          required: true
+          description: The name of the DPP within the mesh to get the diff for.
+          schema:
+            type: string
+        - in: path
+          name: kri
+          required: true
+          description: TODO
+          schema:
+            type: string
+      responses:
+        '200':
+          $ref: '#/components/responses/InboundRulesResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '500':
+          $ref: '#/components/responses/Internal'
+  /meshes/{mesh}/dataplanes/{name}/_outbounds/{kri}/_rules:
+    get:
+      operationId: get-dataplanes-layout
+      summary: TODO
+      description: TODO
+      parameters:
+        - in: path
+          name: mesh
+          required: true
+          description: The mesh of the DPP to get the diff for.
+          schema:
+            type: string
+        - in: path
+          name: name
+          required: true
+          description: The name of the DPP within the mesh to get the diff for.
+          schema:
+            type: string
+        - in: path
+          name: kri
+          required: true
+          description: TODO
+          schema:
+            type: string
+      responses:
+        '200':
+          $ref: '#/components/responses/OutboundRulesResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '500':
+          $ref: '#/components/responses/Internal'
   /meshes/{mesh}/{policyType}/{policyName}/_resources/dataplanes:
     get:
       operationId: inspect-resources
@@ -556,6 +668,27 @@ components:
         name:
           type: string
           example: east
+    DataplaneLayout:
+      type: object
+      title: DataplaneLayout
+      description: TODO
+      required: [ kri, name ]
+      properties:
+        kri:
+          type: string
+          example: TODO
+        name:
+          type: string
+        labels:
+          type: object
+        inbounds:
+          type: array
+          items:
+            $ref: './common/resource.yaml#/components/schemas/DataplaneInbound'
+        outbounds:
+          type: array
+          items:
+            $ref: './common/resource.yaml#/components/schemas/DataplaneOutbound'
   responses:
     IndexResponse:
       description: A response for the index endpoint
@@ -630,6 +763,30 @@ components:
           examples:
             ResponseForHostnames:
               $ref: '#/components/examples/InspectHostnamesExample'
+    DataplaneLayoutResponse:
+      description: TODO
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/DataplaneLayout'
+    ProxyRulesResponse:
+      description: TODO
+      content:
+        application/json:
+          schema:
+            $ref: './common/resource.yaml#/components/schemas/DataplaneProxyRules'
+    InboundRulesResponse:
+      description: TODO
+      content:
+        application/json:
+          schema:
+            $ref: './common/resource.yaml#/components/schemas/DataplaneInboundRules'
+    OutboundRulesResponse:
+      description: TODO
+      content:
+        application/json:
+          schema:
+            $ref: './common/resource.yaml#/components/schemas/DataplaneOutboundRules'
     BadRequest:
       description: Bad Request
       content:
--- a/api/openapi/specs/common/resource.yaml
+++ b/api/openapi/specs/common/resource.yaml
@@ -154,6 +154,61 @@ components:
           x-go-type: 'map[string]string'
         port:
           type: integer
+    DataplaneInbound:
+      type: object
+      required: [ kri, port, protocol ]
+      properties:
+        kri:
+          type: string
+        name:
+          type: string
+        port:
+          type: integer
+        protocol:
+          type: string
+    DataplaneOutbound:
+      type: object
+      required: [ kri, protocol ]
+      properties:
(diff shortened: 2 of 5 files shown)


api/openapi/specs/common/resource.yaml
```diff
@@ -154,6 +156,98 @@ components:
...
+      required: [ policyType, conf, origin ]
+      properties:
+        policyType:
+          type: string
+        conf:
+          description: The actual conf generated
+          type: object
+          additionalProperties: true
+          x-go-type: 'interface{}'
+        origin:
```
Automaat, line 229: can we align this conf with PolicyConf and ResourceRuleOrigin?

api/openapi/specs/api.yaml
```diff
@@ -161,6 +161,118 @@ paths:
...
+          $ref: '#/components/responses/DataplaneNetworkingLayoutResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '500':
+          $ref: '#/components/responses/Internal'
+  /meshes/{mesh}/dataplanes/{name}/_policies:
+    get:
+      operationId: get-proxy-policy-conf
+      summary: Get policies for this dataplane
+      description: Returns proxy level policy configuration derived from combining all policies selecting this dataplane
```
lahabana, line 193:
```suggestion
      description: Returns proxy level policy configuration derived from combining all proxy policies selecting this dataplane 
```

api/openapi/specs/api.yaml
```diff
@@ -161,6 +161,118 @@ paths:
...
+          schema:
+            type: string
+      responses:
+        '200':
+          $ref: '#/components/responses/ProxyPolicyConfResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '500':
+          $ref: '#/components/responses/Internal'
+  /meshes/{mesh}/dataplanes/{name}/_inbounds/{kri}/_policies:
```
lahabana, line 214: needs updating

Automaat: updated

Automaat:
Waiting for MADR to finish

docs/generated/openapi.yaml
```diff
@@ -3667,6 +3859,39 @@ components:
...
+        Dataplane networking layout. It contains information most important
+        information about dataplane and lists of available inbounds and
+        outbounds
+      required:
+        - kri
+        - name
+      properties:
+        kri:
+          type: string
+          example: kri_msvc_mesh-1_us-east-2_kuma-demo_demo-app_http-port
```
lobkovilya: probably worth having DPP's KRI as an example

docs/generated/openapi.yaml
```diff
@@ -3667,6 +3859,39 @@ components:
...
           type: string
           example: east
+    DataplaneNetworkingLayout:
+      type: object
+      title: DataplaneLayout
+      description: >-
+        Dataplane networking layout. It contains information most important
+        information about dataplane and lists of available inbounds and
+        outbounds
+      required:
```
lobkovilya, line 3869: should all fields be required? Otherwise we have these pointers for arrays and maps:
```go
type DataplaneNetworkingLayout struct {
	Inbounds  *[]externalRef0.DataplaneInbound  `json:"inbounds,omitempty"`
	Kri       string                            `json:"kri"`
	Labels    *map[string]string                `json:"labels,omitempty"`
	Name      string                            `json:"name"`
	Outbounds *[]externalRef0.DataplaneOutbound `json:"outbounds,omitempty"`
}
```
but we don't actually need to differentiate between unset labels and empty labels

api/openapi/specs/api.yaml
```diff
@@ -523,6 +703,34 @@ components:
...
+    DataplaneNetworkingLayout:
+      type: object
+      title: DataplaneLayout
+      description: Dataplane networking layout. It contains information most important information about dataplane and lists of available inbounds and outbounds
+      required: [ kri, name ]
+      properties:
+        kri:
+          type: string
+          example: kri_msvc_mesh-1_us-east-2_kuma-demo_demo-app_http-port
+        name:
```
lobkovilya: I know it was in the MADR, but maybe we can remove `name` here as it's exactly the same as the name in path. Also we have `kri` right there.

api/openapi/specs/api.yaml
```diff
@@ -119,6 +119,186 @@ paths:
...
+        '200':
+          $ref: '#/components/responses/DataplaneNetworkingLayoutResponse'
+        '400':
+          $ref: './common/error_schema.yaml#/components/responses/BadRequest'
+        '500':
+          $ref: './common/error_schema.yaml#/components/responses/Internal'
+  /meshes/{mesh}/dataplanes/{name}/_policies:
+    get:
+      operationId: get-proxy-policy-conf
+      summary: Get policies for this dataplane
```
lobkovilya:
```suggestion
      summary: Get proxy policies for this dataplane
```
so it's clear this endpoint doesn't return all policies

api/openapi/specs/common/resource.yaml
```diff
@@ -154,6 +156,125 @@ components:
...
+      type: object
+      required: [ kri, protocol ]
+      properties:
+        kri:
+          type: string
+        port:
+          type: integer
+        protocol:
+          type: string
+    PoliciesList:
```
lobkovilya, line 177: Is it worth having these `*List` types with a single `policies` field? Can we just have `*Response` types to have `policies` list of `*Conf` types?

api/openapi/specs/common/resource.yaml
```diff
@@ -154,6 +156,125 @@ components:
...
+      required: [ kind, conf, origin ]
+      properties:
+        kind:
+          type: string
+        conf:
+          description: The final computed configuration for the data plane proxy, derived by merging all policies whose 'targetRef' field matches the proxy. The merging process follows [RFC 7396 (JSON Merge Patch)](https://datatracker.ietf.org/doc/html/rfc7396), with the order of merging influenced by factors such as where the policy was applied (e.g., custom namespace, system, or global control plane), policy role, and targetRef specificity.
+          type: object
+          additionalProperties: true
+          x-go-type: 'interface{}'
+        origin:
```
lobkovilya: I think we agreed in the MADR we'll nest `kri` to have:
```yaml
origin:
  - kri: $kri1
  - kri: $kri2
```

lobkovilya: Does it make sense to call it in plural `origins`?

api/openapi/specs/common/resource.yaml
```diff
@@ -154,6 +156,125 @@ components:
           x-go-type: 'map[string]string'
         port:
           type: integer
+    DataplaneInbound:
+      type: object
+      required: [ kri, port, protocol ]
+      properties:
+        kri:
+          type: string
+        name:
```
lobkovilya: why `DataplaneInbound` has name, but `DataplaneOutbound` doesn't?

api/openapi/specs/common/resource.yaml
```diff
@@ -154,6 +156,125 @@ components:
           x-go-type: 'map[string]string'
         port:
           type: integer
+    DataplaneInbound:
+      type: object
+      required: [ kri, port, protocol ]
```
lobkovilya, line 159: why port is required for `DataplaneInbound` but not for `DataplaneOutbound`?

api/openapi/specs/api.yaml
```diff
@@ -119,6 +119,186 @@ paths:
...
+          schema:
+            type: string
+      responses:
+        '200':
+          $ref: '#/components/responses/ProxyPolicyConfResponse'
+        '400':
+          $ref: './common/error_schema.yaml#/components/responses/BadRequest'
+        '500':
+          $ref: './common/error_schema.yaml#/components/responses/Internal'
+  /meshes/{mesh}/dataplanes/{name}/_inbounds/{inbound-name}/_policies:
```
lobkovilya: why `inbound-name` and not `inbound-kri` as it says MADR?

Automaat: I've forgot to update it

api/openapi/specs/api.yaml
Automaat: I agree we don't need it here

api/openapi/specs/common/resource.yaml
Automaat: again some old stuff I forgot to update, I now see that in madr it is absent

api/openapi/specs/common/resource.yaml
Automaat: made it required in Outbound

api/openapi/specs/common/resource.yaml
Automaat: right, I forgot about it. It should be ok now

Automaat: it makes sense

api/openapi/specs/common/resource.yaml
Automaat: I am not sure, We already have other lists like: `ResourceTypeDescriptionList`. I would stay with this. As with KRI in origins it gives us more room for expansion in the future if we need something

api/openapi/specs/api.yaml
Automaat: changed

docs/generated/openapi.yaml
Automaat: changed

docs/generated/openapi.yaml
Automaat: righ, this fields should be required

api/openapi/specs/common/resource.yaml
schogges: I'd like to avoid top-level arrays as objects are extendable 🙂 so definitely keep as is even with single entry 👍

schogges:
Generally LGTM, but I'd like to know if those new `*List` endpoints:

1.  support pagination
2. are filterable

if yes please add the spec for the query param(s)

Automaat:
> Generally LGTM, but I'd like to know if those new `*List` endpoints:
> 
> 1. support pagination
> 2. are filterable
> 
> if yes please add the spec for the query param(s)

@schogges Now they are not filterable and they don't support pagination. I don't think it ever will

api/openapi/specs/common/resource.yaml
lobkovilya: @schogges just fyi that's not what I suggested, the response would be exactly the same:
```yaml
policies: [...]
```
but instead of creating 
```yaml
OutboundPolicyConfResponse:
  description: A response containing policies configuration for selected outbound
  content:
    application/json:
      schema:
        $ref: './common/resource.yaml#/components/schemas/PoliciesList'
PoliciesList:
  type: object
  required: [ policies ]
  properties:
    policies:
      type: array
      items:
        $ref: '#/components/schemas/PolicyConf'
```
I suggested:
```yaml
OutboundPolicyConfResponse:
  description: A response containing policies configuration for selected outbound
  content:
    application/json:
      schema:
        type: object
        required: [ policies ]
        properties:
          policies:
            type: array
            items:
              $ref: '#/components/schemas/PolicyConf'
```
because I don't really see value in having `*List` type, IMO just more jumps to read the response schema. 

But I see that we're having this pattern in multiple places, so I'm ok with it.

docs/generated/openapi.yaml
```diff
@@ -3667,6 +3859,39 @@ components:
...
+        - kri
+        - labels
+        - inbounds
+        - outbounds
+      properties:
+        kri:
+          type: string
+          example: kri_dp_default_default_kuma-demo_demo-app-75ff54499c-ttwd7_http-port
```
lobkovilya, line 3877: it's nitpicking since this is an example, but I guess this KRI always will be without the section name, no?

Automaat: That is true it will be without sectionName, I can update it in next prs, to not block this

api/openapi/specs/common/resource.yaml
schogges: Ah sry I misunderstood 🤦 in that case I'd agree with you 👍

PR #13500, opened 2025-05-05
docs(MADR): computed endpoints

lahabana:
This has always been an informal, unpersisted decision. Adding this as a MADR to more easily reference it

### DIFF
--- a/docs/madr/decisions/072-computed-api-endpoints.md
+++ b/docs/madr/decisions/072-computed-api-endpoints.md
@@ -1,0 +1,50 @@
+# Computed API Endpoints
+
+* Status: accepted
+
+## Context and Problem Statement
+
+There are two types of API endpoints in Kuma:
+
+- Resource endpoints - These are endpoints that lead to direct reads to the database. They map to Kuma resources. For example: `GET /mesh/{mesh}/dataplanes/{dataplane}`.
+- Computed endpoints - These are endpoints that provide computed data usually using existing resources. For example: `GET /mesh/{mesh}/dataplanes/{dataplane}/_overview` 
+  or `GET /mesh/{mesh}/dataplanes/{dataplane}/_metrics`.
+
+Computed endpoints represent runtime and composed state, this state is not stored in the database and there's no way to update it from the API.
+
+This MADR provides a convention for computed endpoints. The goal is to provide a consistent way to define and document computed endpoints.
+This doesn't prevent non computed endpoints to include part of the response as computed (examples: warnings or status fields)
+
+## Design
+
+Computed endpoints should be always prefixed with `_`, for example: `GET /mesh/{mesh}/dataplanes/{dataplane}/_overview`.
+
+This has the following benefits:
+
+- It is easy to identify computed endpoints.
+- Names of resources and resource type names can never start with `_`, so there is no risk of collision between a resource and a computed endpoint. For example: `GET /mesh/{mesh}/dataplanes/{dataplane}` and `GET /mesh/{mesh}/dataplanes/_overview` are two different endpoints there's never a risk of having a dataplane names `_overview`.
+- endpoints without `_` prefix are always resource endpoints. This makes it easier to understand the API and reduces the risk of confusion, it also makes generation of clients easier.
+
+## Implications for Kong Mesh
+
+None
+
+## Decision
+
+The decision is to use `_` as a prefix for computed endpoints. This is a convention that should be followed in all future development.
+
+For existing endpoints, we should provide alternative endpoints and progressively deprecate old endpoints.
+All new endpoints should follow this convention and be documented in openAPI spec.
+
+Some examples of endpoints that should be prefixed with `_`:
+
+- xds, clusters, stats: https://github.com/kumahq/kuma/issues/11620
+- `/global-insights`
+- `/config`
+- `/who-am-i`
+
+## Alternatives Considered
+
+- No prefix - This would make it harder to identify computed endpoints and could lead to confusion. It could also lead to collision between resource and computed endpoints.
+- Using request param (`?computed=true`) - the schema of computed endpoints is different than their resource, this would make the API very complex, especially as this kind of response is hard to define in openAPI.
+- Using `Accept` header - accept headers are not meant to change the actual content of the response (see [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2)) it's only meant to define the format of the response.


docs/madr/decisions/072-computed-api-endpoints.md
```diff
@@ -0,0 +1,41 @@
...
+
+* Status: accepted
+
+## Context and Problem Statement
+
+There are two types of API endpoints in Kuma:
+
+- Resource endpoints - These are endpoints that leads to direct reads to the database. They map to Kuma resources. For example: `GET /mesh/{mesh}/dataplanes/{dataplane}`.
+- Computed endpoints - These are endpoints that provide computed data usually using existing resources. For example: `GET /mesh/{mesh}/dataplanes/{dataplane}/_overview` 
+  or `GET /mesh/{mesh}/dataplanes/{dataplane}/_metrics`.
```
schogges, line 11: Maybe it would be good to explain why there are the different types of endpoints? 🙂

docs/madr/decisions/072-computed-api-endpoints.md
```diff
@@ -0,0 +1,41 @@
...
+
+This MADR provides a convention for computed endpoints. The goal is to provide a consistent way to define and document computed endpoints.
+
+## Design
+
+Computed endpoints should be always prefixed with `_`, for example: `GET /mesh/{mesh}/dataplanes/{dataplane}/_overview`.
```
schogges, line 17: Did we consider alternatives such as separate APIs or different scoping?

lahabana: Could you show an example?

docs/madr/decisions/072-computed-api-endpoints.md
```diff
@@ -0,0 +1,41 @@
...
+- Names of resources and resource type names can never start with `_`, so there is no risk of collision between a resource and a computed endpoint. For example: `GET /mesh/{mesh}/dataplanes/{dataplane}` and `GET /mesh/{mesh}/dataplanes/_overview` are two different endpoints there's never a risk of having a dataplane names `_overview`.
+- endpoints without `_` prefix are always resource endpoints. This makes it easier to understand the API and reduces the risk of confusion, it also makes generation of clients easier.
+
+## Implications for Kong Mesh
+
+None
+
+## Decision
+
+The decision is to use `_` as a prefix for computed endpoints. This is a convention that should be followed in all future development.
```
jijiechen, line 31: Will there be a lint tool to enforce this?

docs/madr/decisions/072-computed-api-endpoints.md
schogges: i.e. `/computed/meshes/{mesh}/dataplanes/{dataplane}`

Others might be
- using a query param `?computed=true` or `?embed=(computed-resource)`
- using `Accept` header, i.e. `application/vnd.computed+json` or some custom header

At the moment I think prefixing with `_` is good enough, but I think it's worth adding alternatives to the MADR so we could come to the conclusion that at least in our case `_`-prefixing is the best solution at this point. 🤔

docs/madr/decisions/072-computed-api-endpoints.md
lahabana: not that I know of, also I'm not sure how we could lint this as this is more a logical convention than something we can validate in code no?

schogges (approved):
LGTM 👍

lukidzi (approved):

docs/madr/decisions/072-computed-api-endpoints.md
```diff
@@ -0,0 +1,49 @@
...
+
+This MADR provides a convention for computed endpoints. The goal is to provide a consistent way to define and document computed endpoints.
+
+## Design
+
+Computed endpoints should be always prefixed with `_`, for example: `GET /mesh/{mesh}/dataplanes/{dataplane}/_overview`.
+
+This has the following benefits:
+
+- It is easy to identify computed endpoints.
```
johncowen, line 23: Kinda rhetorical question after confirmation more than anything, but this is an internal concern right?

Some other comments might touch on this a little more (not read the rest yet), and I know this MADR is somewhat for documentation of a historical decision, but curious to get comment on an opinion that from a consumers standpoint the `_` could be seen as exposing internals in the abstraction? i.e. how come a consumer becomes aware/needs to be aware of whether a URL refers to a computed result or not? 

I know this is a pretty well baked approach that we do in our HTTP API but I've always wanted to ask/hear thoughts on that ☝️ , figured now is probably good time to ask.

docs/madr/decisions/072-computed-api-endpoints.md
johncowen: Using headers for something like this would make client side caching harder I think

docs/madr/decisions/072-computed-api-endpoints.md
```diff
@@ -0,0 +1,49 @@
+# Computed API Endpoints
+
+* Status: accepted
+
+## Context and Problem Statement
+
+There are two types of API endpoints in Kuma:
+
+- Resource endpoints - These are endpoints that leads to direct reads to the database. They map to Kuma resources. For example: `GET /mesh/{mesh}/dataplanes/{dataplane}`.
```
slonka, line 9: nitpick:

```suggestion
- Resource endpoints - These are endpoints that lead to direct reads to the database. They map to Kuma resources. For example: `GET /mesh/{mesh}/dataplanes/{dataplane}`.
```

docs/madr/decisions/072-computed-api-endpoints.md
```diff
@@ -0,0 +1,49 @@
...
+- Computed endpoints - These are endpoints that provide computed data usually using existing resources. For example: `GET /mesh/{mesh}/dataplanes/{dataplane}/_overview` 
+  or `GET /mesh/{mesh}/dataplanes/{dataplane}/_metrics`.
+
+Computed endpoints represent runtime and composed state, this state is not stored in the database and there's no way to update it from the API.
+
+This MADR provides a convention for computed endpoints. The goal is to provide a consistent way to define and document computed endpoints.
```
slonka, line 15: might be worth mentioning "computed" parts of the resources, like `status` or `warnings`

docs/madr/decisions/072-computed-api-endpoints.md
slonka: > how come a consumer becomes aware/needs to be aware of whether a URL refers to a computed result or not?

If you're unfamiliar with the API you might be tempted to try and manually modify the computed resource (which is not possible).

slonka (approved):
Just nitpicks

docs/madr/decisions/072-computed-api-endpoints.md
johncowen: (btw don't anybody wait for this convo to be resolved before merging!)

docs/madr/decisions/072-computed-api-endpoints.md
lahabana: Yes done

lukidzi (approved):

slonka (approved):

Issue #13287, opened 2025-04-01
☔ producer policy Mesh*Route support


lahabana:
### Description

Producer policies support Mesh*Service and should now support Mesh*Route

lahabana:
Let's do the docs and then this can be moved to 2.13 to add deprecation

Issue #13267, opened 2025-03-28
Use `KRI` in Envoy resources for Builtin Gateway


lobkovilya:
### Description

A single issue for now to track `KRI` and Builtin Gateway. Probably it's going to unfold into several issues in the future:
* MADR on `KRI` with hash suffixes
* MADR on migration if it's different from https://github.com/kumahq/kuma/issues/13264
* Actual renaming work

slonka:
I'm pausing this until we decide that if we continue developing for `builtin` GW

Currently, the cluster name is computed in
- https://github.com/kumahq/kuma/blob/a8788ad310103113a61fcd7557c8954a2d4a6977/pkg/plugins/runtime/gateway/cluster_generator.go#L299
- https://github.com/kumahq/kuma/blob/26af860614cf0792c0bff004ac95e8f5115808fc/pkg/xds/envoy/tags/match.go#L37
and it contains a hash of the listener tags.

This goes against one of the KRI principles that KRI should be ```"Human-readable: users should be able to type the identifier manually if needed, so it's not hashed"```

Given the fact that we decided keep separate clusters per listener https://github.com/kumahq/kuma/issues/13129#issuecomment-2736669334 we need some human-readable way to identify these clusters.

the only way I thought of doing is to combine these two:

```
kri_mgw_mesh-1_us-east-2__gw-1_{kri_msvc_mesh-1_us-east-2_kuma-demo_backend_httpport}
```

```go
type Identifier struct {
	ResourceType core_model.ResourceType
	Mesh         string
	Zone         string
	Namespace    string
	Name         string
	SectionName  string

    Contains *Identifier
}
```

and estabilish a relationship that a listener has cluster

lahabana:
Removing the parent. Let's do this until we make a better decision around builtin gateway

lobkovilya:
> triage: we won't do this

PR #13502, opened 2025-05-05
feat(meshtimeout): add MeshHTTPRoute support

lobkovilya:
## Motivation

See [MADR-60](https://github.com/kumahq/kuma/blob/abaf5c54ff421a82c447eb8e28a3561dbfaf0e38/docs/madr/decisions/060-rework-mesh-routes-matching.md)

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13287?issue=kumahq%7Ckuma%7C10999


pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
```diff
@@ -263,48 +271,84 @@ func createInboundClusterName(servicePort uint32, listenerPort uint32) string {
...
-		if err != nil {
-			return err
+func configureRoutes(
+	rc *envoy_route.RouteConfiguration,
+	rules outbound.ResourceRules,
+	reader kri.ResourceReader,
+) error {
+	for _, vh := range rc.VirtualHosts {
+		for _, route := range vh.Routes {
+			if !strings.HasPrefix(route.Name, "kri_") {
```
lukidzi, line 331: Should we have some common const so if we need to reference it we won't need to define each time/update each place?

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
```diff
@@ -263,48 +271,79 @@ func createInboundClusterName(servicePort uint32, listenerPort uint32) string {
...
```
lukidzi, line 326: Maybe it's not a question for this but does it work for gateway? I think for gateway we set name of the VH/Route to hash of the match.

lobkovilya: Yeah it doesn't work for gateway until https://github.com/kumahq/kuma/issues/13267 is done. I'm going to create an issue to take care of MeshTimeout and MeshHTTPRoute for Builtin Gateway

lobkovilya: https://github.com/kumahq/kuma/issues/13519

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
lobkovilya: Added `kri.IsValid(s string) bool` to check if string is KRI

lukidzi (approved):

PR #13503, opened 2025-05-05
feat(xds): add option to bind outbounds to real loopback IP:PORT

lukidzi:
## Motivation

It would be beneficial to add an option to bind to real addresses, in case the user doesn't want to manually define `outbounds` for a dataplane when running in Universal mode.

## Implementation information

> [!WARNING]
> Since IPv6 only has a single loopback address `::1/128`, there is no range of loopback addresses, so this approach won't work with IPv6.

* When the `--bind-outbounds` flag is set, Envoy tries to create real addresses using `bind_to_port: true`.
* In this case, the CIDRs used by IP allocators must fall within the loopback range, as you can only bind to those addresses.
* An E2E test has been added to cover this use case.

## Supporting documentation

Fix https://github.com/Kong/kong-mesh/issues/7798


pkg/xds/bootstrap/generator.go
```diff
@@ -127,6 +130,10 @@ func (b *bootstrapGenerator) Generate(ctx context.Context, request types.Bootstr
 		UseDeltaXds:          b.deltaXdsEnabled,
 	}
 
+	if b.dynamicLoopbackOutbounds {
+		params.Features[xds_types.FeatureDynamicLoopbackOutbounds] = true
+	}
+
 	if params.Features.HasFeature(xds_types.FeatureDeltaGRPC) {
```
lahabana, line 137: I think we should go with a DP feature, it's easy to add this to your ECS task and it's a more flexible setup.

The feature could be named: "no-passthrough"

lukidzi: I've implemented here 2 ways:
* Single DP can enable feature
* We can enable feature for all DPs in this case user don't need to define it on each DP

lahabana (dismissed):

PR #13510, opened 2025-05-08
test(framework): add the ability to customize docker volumes and arguments

jijiechen:
## Motivation

Adding the ability to customize docker volumes and arguments when installing Kuma on Universal. With this ability, we can attach certificates and other resources in E2E tests.

## Implementation information

* Introducing helper method `WithCPDockerVolumes` to test framework
* Introducing helper method `WithCPDockerRunOptions` to test framework (to allow specify docker arguments like `--ip`)

## Supporting documentation

An enahancement to https://github.com/kumahq/kuma/pull/13301

> Changelog: skip

### DIFF
--- a/test/framework/interface.go
+++ b/test/framework/interface.go
@@ -35,7 +35,7 @@ type kumaDeploymentOptions struct {
 	helmChartPath               *string
 	helmChartVersion            string
 	helmOpts                    map[string]string
-	noHelmOpts                  []string
+	helmOptsExcluded            []string
 	env                         map[string]string
 	zoneIngress                 bool
 	zoneIngressEnvoyAdminTunnel bool
@@ -56,6 +56,9 @@ type kumaDeploymentOptions struct {
 	// Functions to apply to each mesh after the control plane
 	// is provisioned.
 	meshUpdateFuncs map[string][]func(*mesh_proto.Mesh) *mesh_proto.Mesh
+
+	// extra docker container options to attach when running on universal
+	dockerRunOptions []string
 }
 
 func (k *kumaDeploymentOptions) apply(opts ...KumaDeploymentOption) {
@@ -120,6 +123,7 @@ type appDeploymentOptions struct {
 
 	dockerVolumes       []string
 	dockerContainerName string
+	dockerRunOptions    []string
 }
 
 func (d *appDeploymentOptions) apply(opts ...AppDeploymentOption) {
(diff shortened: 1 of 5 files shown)


test/framework/interface.go
```diff
@@ -35,7 +35,7 @@ type kumaDeploymentOptions struct {
 	helmChartPath               *string
 	helmChartVersion            string
 	helmOpts                    map[string]string
-	noHelmOpts                  []string
+	helmOptsExcluded            []string
```
jijiechen, line 38: Refactoring: changed to a name for better readability

test/framework/interface.go
```diff
@@ -355,6 +360,18 @@ func WithCtlOpts(opts map[string]string) KumaDeploymentOption {
 	})
 }
 
+func WithCPDockerRunOptions(options []string) KumaDeploymentOption {
+	return KumaOptionFunc(func(o *kumaDeploymentOptions) {
+		o.dockerRunOptions = options
+	})
+}
+
+func WithCPDockerVolumes(volumes ...string) KumaDeploymentOption {
```
jijiechen, line 369: Name `WithDockerVolumes` is taken in this package, so adding a `CP` into the name for the scenario of installing Kuma CP.

test/framework/setup.go
```diff
@@ -642,17 +642,15 @@ func universalZoneProxyRelatedResource(
...
 			dpName,
 			"",
 			appType,
-			Config.IPV6,
-			false,
-			[]string{},
-			[]string{},
-			"",
-			concurrency,
+			UniversalAppRunOptions{
```
jijiechen, line 649: Refactoring: introducing a new struct to wrap all these fields (because I wanted to add more)

test/framework/universal_cluster.go
```diff
@@ -133,6 +133,7 @@ func (c *UniversalCluster) DeployKuma(mode core.CpMode, opt ...KumaDeploymentOpt
 	} else {
 		opt = append([]KumaDeploymentOption{WithEnvs(Config.KumaUniversalEnvVars)}, opt...)
 	}
+	c.opts = kumaDeploymentOptions{}
```
jijiechen, line 136: Resetting the `opts` to allow `DeployKuma` to be run multiple times. Otherwise, the options will be populated in the second run.

lukidzi: What if you change the way how w create a deployment of kuma? Now we store it as a `control-plane`. Maybe we need a list of control-planes? Also, why do we need a second run?

jijiechen: Because I wanted to upgrade the CP in a test case, so it shares the existing cluster object while changing the controlPlane state.

There is still one control plane version in my case, so don't need to maintail multiple control plane objects.

lukidzi (approved):

PR #13544, opened 2025-05-13
chore(oapi): ignore structpb.Struct from oas generation

slonka:
## Motivation

https://github.com/kumahq/kuma/pull/13542 did not help because I think I confused it with a fix for another cryptic error. The actual fix should be https://github.com/kumahq/kuma/pull/12329/files#diff-d719909e3a90ac11d2e421de4759332f4795fa80fa0a2d0635a599cdae2dadd6R449 the same as in this PR.

## Implementation information

Add `structpb.Struct` to ignored types.


bartsmykla (approved):
LGTM

PR #13559, opened 2025-05-16
fix(rules): incorrect behaviour of MeshTimeout/MeshRetry policies when referencing routes

lobkovilya:
## Motivation

Method `Compute` is falling back to `Mesh` rules if `MeshHTTPRoute` doesn't exist. See the repro steps in https://github.com/kumahq/kuma/issues/13571.

When computing `MeshHTTPRoute` conf in `Compute` we can't fall back to its `MeshService`, because there might be an ambiguity – `MeshHTTPRoute` could be applied to several `MeshServices`. But when configuring the route, we always know the `MeshService` of the outbound listener. Computing `MeshHTTPRoute` conf is possible only in the context of `MeshService`.

## Implementation information

Instead of passing both `MeshService` and `MeshHTTPRoute` to `Compute` I created a `ResourceContext`:

```go
// ResourceContext represents a hierarchical resource structure and
// is always ready to return the appropriate conf by using Conf method.
// The RootContext is always the mesh. As we're iterating over resource in ResourceSet
// and going deeper to configure Listeners/Routes, we need to add more resources
// to the ResourceContext by using WithID method.
//
// For example:
// 1. At the beginning of Apply method in plugin.go rctx := RootContext()
// 2. As we start iterating over outbound listeners, rctx = rctx.WithID(outboundListenerKRI)
// 3. As we start iterating over the routes of the outbound listener, rctx = rctx.WithID(routeKRI)
//
// At any moment we can call rctx.Conf() and get right configuration.
type ResourceContext[T any] struct {
    ids   []kri.Identifier
    rules ResourceRules
}
```

In `plugin.go` we should start with creating `RootContext`:

```go
func (p plugin) Apply(...) error {
    rctx := outbound.RootContext[meshtimeout_api.Conf](mesh, rules)

    // as we're iterating over listeners/clusters we must add KRI to the context and pass it down
    for _, r := range resources {
        configureListener(rctx.WithID(r.ResourceOrigin), ...)
    }
}

func configureListener(rctx ResourceContext, ...) {
    // as we iterating over routes we again add route's KRI and pass down the context
    for _, route := range routes {
        configureRoute(rctx.WithID(kri.FromString(route.Name)), ...)
    }
}
```

at any moment `rctx.Conf()` is ready to give us the current conf.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13571

### DIFF
--- a/pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
+++ b/pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
@@ -10,9 +10,6 @@ import (
 	"github.com/kumahq/kuma/pkg/core/kri"
 	core_plugins "github.com/kumahq/kuma/pkg/core/plugins"
 	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
-	meshexternalservice_api "github.com/kumahq/kuma/pkg/core/resources/apis/meshexternalservice/api/v1alpha1"
-	meshmultizoneservice_api "github.com/kumahq/kuma/pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1"
-	meshservice_api "github.com/kumahq/kuma/pkg/core/resources/apis/meshservice/api/v1alpha1"
 	core_xds "github.com/kumahq/kuma/pkg/core/xds"
 	xds_types "github.com/kumahq/kuma/pkg/core/xds/types"
 	"github.com/kumahq/kuma/pkg/plugins/policies/core/matchers"
@@ -76,15 +73,7 @@ func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *
 
 	rctx := outbound.RootContext[api.Conf](ctx.Mesh.Resource, policies.ToRules.ResourceRules)
 
-	for _, r := range util_slices.Filter(rs.List(), core_xds.HasResourceOrigin) {
-		switch r.ResourceOrigin.ResourceType {
-		case meshservice_api.MeshServiceType:
-		case meshexternalservice_api.MeshExternalServiceType:
-		case meshmultizoneservice_api.MeshMultiZoneServiceType:
-		default:
-			continue
-		}
-
+	for _, r := range util_slices.Filter(rs.List(), core_xds.HasAssociatedServiceResource) {
 		svcCtx := rctx.
 			WithID(kri.NoSectionName(*r.ResourceOrigin)).
 			WithID(*r.ResourceOrigin)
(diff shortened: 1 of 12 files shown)


pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
```diff
@@ -74,11 +74,24 @@ func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *
...
-	}
+	rctx := outbound.RootContext[api.Conf](ctx.Mesh.Resource, policies.ToRules.ResourceRules)
+
+	for _, r := range util_slices.Filter(rs.List(), core_xds.HasResourceOrigin) {
+		switch r.ResourceOrigin.ResourceType {
+		case meshservice_api.MeshServiceType:
+		case meshexternalservice_api.MeshExternalServiceType:
+		case meshmultizoneservice_api.MeshMultiZoneServiceType:
+		default:
+			continue
```
slonka, line 85: nitpick: to not duplicate this should we extend the `Filter` to accept multiple conditions and have something like:

```suggestion
	for _, r := range util_slices.Filter(rs.List(), core_xds.HasResourceOrigin, core_xds.IsNotMeshStarService) {
```

and have 

```
func IsNotMeshStarService(r *Resource) bool {
	switch r.ResourceOrigin.ResourceType {
	case meshservice_api.MeshServiceType:
	case meshexternalservice_api.MeshExternalServiceType:
	case meshmultizoneservice_api.MeshMultiZoneServiceType:
		return false
	}
	return false
}
```

```
func Filter[A any](input []A, functions... func(A) bool) []A {
	output := make([]A, 0, len(input))
	for _, a := range input {
		include := true
		for _, f := range functions {
			if ok := f(a); !ok {
				include = false
			}
		}
		if include {
			output = append(output, a)
		}
	}
	return output
}
```

(warning, not tested!)

slonka (commented):
I like this approach because it gets rid of these:

```
if rr := rules.Compute(*r.ResourceOrigin, reader); rr != nil {
			conf = rr.Conf[0].(api.Conf)
		}
```

also, should this be called in other plugins now? or is this going to be follow up?

pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go
lobkovilya: I created filter:
```go
func HasAssociatedServiceResource(r *Resource) bool {
	if r.ResourceOrigin == nil {
		return false
	}
	switch r.ResourceOrigin.ResourceType {
	case
		meshservice_api.MeshServiceType,
		meshexternalservice_api.MeshExternalServiceType,
		meshmultizoneservice_api.MeshMultiZoneServiceType:
		return true
	}
	return false
}
```

slonka (approved):
LGTM, let's replace the 

```
if rr := rules.Compute(*r.ResourceOrigin, reader); rr != nil {
			conf = rr.Conf[0].(api.Conf)
		}
```

with the new invocation in the follow up PR

PR #13570, opened 2025-05-19
refactor(meshaccesslog): configurer and tests

lobkovilya:
## Motivation

Unit tests:
* hard to add new test cases
* test code was recreating HCM with FilterChains instead of reusing the actual code

Configurer:
* hard to extend to implement the logic for routes
* a lot of unnecessary parameters passed deep down to functions

## Implementation information

Depends on https://github.com/kumahq/kuma/pull/13559


Icarus9913 (commented):
Good refactor!

pkg/plugins/policies/meshaccesslog/plugin/v1alpha1/plugin.go
```diff
@@ -232,113 +298,56 @@ func applyToGateway(
...
-	dataplane *core_mesh.DataplaneResource,
 	listener *envoy_listener.Listener,
-	backendsAcc *plugin_xds.EndpointAccumulator,
+	backendsAcc *EndpointAccumulator,
+	protocol core_mesh.Protocol,
+	values listeners_v3.KumaValues,
 	accessLogSocketPath string,
 ) error {
-	serviceName := dataplane.Spec.GetIdentifyingService()
+	defaultFmt := DefaultFormat(protocol)
```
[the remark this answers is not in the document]
lobkovilya: The `DefaultFormat` func checks the protocol
```go
```

pkg/envoy/builders/common/builder.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+type Configurer[R any] func(*R) error
+
+func (b *Builder[R]) Configure(configurer Configurer[R]) *Builder[R] {
+	b.configurers = append(b.configurers, configurer)
+	return b
+}
+
+func (b *Builder[R]) ConfigureIf(condition bool, fn func() Configurer[R]) *Builder[R] {
```
Automaat, line 14: is there possibility for this condition to be something more complex than a bool? Like function that checks previous configurers output or something like this?

Automaat (approved):

pkg/envoy/builders/accesslog/accesslog.go
```diff
@@ -0,0 +1,72 @@
+package accesslog
+
+import (
```
lukidzi, line 3: Should this all code be in `xds/listeners/v3` where we have all configurers?

lobkovilya: No :) I'm planning to add README.md for the `pkg/envoy/builders` when I'm done with the MeshAccessLog. But the idea is to have these new Builders and Configurers as a replacement. They're simpler and way less verbose than previous implementation in `xds/listeners` (mostly thanks to generics). 

Also, I'll mention in the README.md: 
* we need to follow the directory structure as in https://github.com/envoyproxy/envoy/tree/main/api/envoy/config
* we shouldn't depend on any dirs from `pkg/plugins/policies`

pkg/xds/envoy/names/resource_names.go
```diff
@@ -117,6 +117,20 @@ func GetGatewayListenerName(gatewayName string, protoName string, port uint32) s
 	return Join(gatewayName, protoName, formatPort(port))
 }
 
+func ParseGatewayListenerName(listenerName string) (string, string, uint32, error) {
```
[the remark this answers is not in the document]
lobkovilya: Done

lukidzi (approved):

Issue #13285, opened 2025-04-01
☔ Restrict set of permissions on Kubernetes


lahabana:
### Description

Some of the requirements:

- As a user, I want to bind a ClusterRole with a ClusterRoleBinding while restricting its permissions to a reduced scope.
- As a user, I want to have the possibility to disable the creation of ClusterRole/ClusterRoleBinding on the upgrade/install
- As a user, I want to have the possibility to run webhooks only on specific namespaces
- As a user, I don't want a mesh control-plane to watch sensitive resources (ConfigMaps, Secrets) across all namespaces
- As a user, I want control-plane to observe only resources in specific namespaces (out-of-scope) - agreed on reduced ClusterRole

More issue will be provided once MADR is ready and approved

PR #13581, opened 2025-05-20
chore(docs/generated): export rendered RBAC manifests to docs for reuse

bartsmykla:
## Motivation

We’re updating the docs to include Kubernetes RBAC resources, so generating these files automatically will help with future maintenance.

## Implementation information

Added make target to generate `docs/generated/raw/rbac.yaml` with control plane RBAC; includes ClusterRoles, RoleBindings, and related resources from Helm chart

## Supporting documentation

Related to: https://github.com/kumahq/kuma/issues/13285

### DIFF
--- a/docs/generated/raw/rbac.yaml
+++ b/docs/generated/raw/rbac.yaml
@@ -1,32 +1,8 @@
-{{- if or .Values.skipRBAC .Values.controlPlane.skipClusterRoleCreation }}
-{{- printf "\n# WARNING: RBAC is skipped. If this was intentional, remember to create the necessary resources for control-plane manually.\n" }}
-{{- end }}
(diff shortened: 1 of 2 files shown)


slonka (approved):

docs/generated/raw/rbac.yaml
```diff
@@ -0,0 +1,333 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kuma-control-plane
```
slonka, line 4: is this going to be correctly overwritten in parent project?

slonka (dismissed):

slonka (approved):

PR #13583, opened 2025-05-20
fix(accesslogstreamer): stop adding newline to the msg

lobkovilya:
## Motivation

* e2e test for MeshAccessLog was incorrectly checking `head -1 nc.out` while new messages are written at the bottom
* AccessLogStreamer adds additional `\n` however we already add `\n` to the message in CP https://github.com/kumahq/kuma/blob/5bcced29dbb0ad857c78e5e6a2d6f75ef03d103d/pkg/plugins/policies/meshaccesslog/plugin/xds/configurer.go#L109

## Implementation information

* fixed the test to use `tail -1`
* stop adding `\n` in accesslogstreamer

### DIFF
--- a/app/kuma-dp/pkg/dataplane/accesslogs/streamer.go
+++ b/app/kuma-dp/pkg/dataplane/accesslogs/streamer.go
@@ -134,8 +134,6 @@ func (s *accessLogStreamer) streamAccessLogs(reader *bufio.Reader) error {
 					continue
 				}
 			}
-
-			accessLogMsg = append(accessLogMsg, '\n')
 		}
 
 		s.RLock()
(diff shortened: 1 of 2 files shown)


app/kuma-dp/pkg/dataplane/accesslogs/streamer.go
```diff
@@ -134,8 +134,6 @@ func (s *accessLogStreamer) streamAccessLogs(reader *bufio.Reader) error {
 					continue
 				}
 			}
-
-			accessLogMsg = append(accessLogMsg, '\n')
```
lukidzi, line 138: Won't it break any integration user might have? Like Dynatrace agent fetching logs and looking for some specific format?

lobkovilya: Added UPGRADE.md section, please take a look

lukidzi (approved):

PR #13591, opened 2025-05-21
feat(inspect-api): use real dataplane metadata when computing config for proxy

Automaat:
## Motivation
There could be policies that needs dynamic metadata for configuration to be computed, and we need to include them in proxy as we recently added these to dataplane insights

### DIFF
--- a/pkg/core/xds/inspect/config.go
+++ b/pkg/core/xds/inspect/config.go
@@ -32,10 +32,10 @@ type ProxyConfigInspector struct {
 	meshContext            xds_context.MeshContext
 	snapshotGenerator      *v3.TemplateSnapshotGenerator
 	knownInternalAddresses []string
-	dataplaneInsight       *core_mesh.DataplaneInsightResource
+	dataplaneMetadata      *core_xds.DataplaneMetadata
 }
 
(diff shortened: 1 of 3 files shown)


pkg/core/xds/inspect/config.go
```diff
@@ -32,9 +32,10 @@ type ProxyConfigInspector struct {
 	meshContext            xds_context.MeshContext
 	snapshotGenerator      *v3.TemplateSnapshotGenerator
 	knownInternalAddresses []string
+	dataplaneInsight       *core_mesh.DataplaneInsightResource
 }
 
-func NewProxyConfigInspector(meshContext xds_context.MeshContext, zone string, knownInternalAddresses []string, hooks ...xds_hooks.ResourceSetHook) (*ProxyConfigInspector, error) {
+func NewProxyConfigInspector(meshContext xds_context.MeshContext, dataplaneInsight *core_mesh.DataplaneInsightResource, zone string, knownInternalAddresses []string, hooks ...xds_hooks.ResourceSetHook) (*ProxyConfigInspector, error) {
```
lahabana, line 38: we should probably just pass `p.dataplaneInsight.Spec.Metadata if that's all we need.

Automaat: changed

lukidzi (approved):

Issue #13592, opened 2025-05-21
On KDS restart we resync MeshServices and reallocate IPs


lukidzi:
### Kuma Version

2.10.1

### Describe the bug

In a multizone setup, we can observe unexpected behavior:
We have a Global control plane and two Zones, both using MeshService. When one of the Zones disconnects from the Global and then reconnects, we detect changes in MeshService and overwrite existing entries, even when no actual change has occurred. This causes the VIP allocator to reassign VIPs unnecessarily.

### To Reproduce

1. Create Global CP
2. Create zone-1
3. create zone-2
4. Deploy demo services in both zones
5. Enable `meshService`
6. Cause one zone to lose the connection to global/or restart (when using postgres/k8s)
7. You can notice that MeshService is created again and VIPs allocated

PR #13618, opened 2025-05-22
fix(kds): don't override status field

lukidzi:
## Motivation

In a multizone setup using MeshService mode, we synchronize objects across zones. Each object contains a status field, which is intentionally removed during synchronization on the global control plane. This allows a remote zone to allocate a VIP and hostname for the MeshService based on its local configuration.

However, when the connection to the global control plane is reestablished, the system detects a change in the resource and overrides the object in the zone, even when there are no actual changes. This happens because the global control plane has a version of the MeshService with an empty status field, while the zone contains an updated status. As a result, the objects appear to differ and trigger an unnecessary update.

The override of MeshService status field was casuing multiple IP allocations and could cause more frequent updates of XDS snapshots.

## Implementation information

First of all, we don't need to compare the status field of synchronized resources, since only the zone runs a generator that sets the status, while the global control plane always has it empty. Additionally, we should ensure that the status field is removed from all resources when syncing them to the global control plane. Since the status shouldn't change, we should set the status of the existing object instead of overriding it.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13592

### DIFF
--- a/pkg/kds/v2/reconcile/snapshot_generator.go
+++ b/pkg/kds/v2/reconcile/snapshot_generator.go
@@ -47,6 +47,12 @@ func IsKubernetes(storeType config_store.StoreType) func(model.Resource) bool {
 	}
 }
 
+func HasStatus() func(model.Resource) bool {
+	return func(r model.Resource) bool {
+		return r.Descriptor().HasStatus
+	}
+}
+
 func NameHasPrefix(prefix string) func(model.Resource) bool {
 	return func(r model.Resource) bool {
 		return strings.HasPrefix(r.GetMeta().GetName(), prefix)
--- a/pkg/kds/v2/store/sync.go
+++ b/pkg/kds/v2/store/sync.go
@@ -221,11 +221,17 @@ func (s *syncResourceStore) Sync(syncCtx context.Context, upstreamResponse clien
 		}
 		newLabels := r.GetMeta().GetLabels()
 		if !core_model.Equal(existing.GetSpec(), r.GetSpec()) ||
-			!maps.Equal(existing.GetMeta().GetLabels(), newLabels) ||
-			!core_model.Equal(existing.GetStatus(), r.GetStatus()) {
+			!maps.Equal(existing.GetMeta().GetLabels(), newLabels) {
 			// we have to use meta of the current Store during update, because some Stores (Kubernetes, Memory)
 			// expect to receive ResourceMeta of own type.
 			r.SetMeta(existing.GetMeta())
+			// we should preserve Status of the resource
+			if r.Descriptor().HasStatus {
+				if err = r.SetStatus(existing.GetStatus()); err != nil {
+					log.Error(err, "failed to set status", "resource", r.GetMeta())
+					continue
+				}
+			}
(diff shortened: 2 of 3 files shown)


Automaat (approved):

pkg/kds/v2/store/sync.go
```diff
@@ -320,6 +334,7 @@ func ZoneSyncCallback(ctx context.Context, syncer ResourceSyncer, k8sStore bool,
 					}
 					return !core_model.IsLocallyOriginated(config_core.Zone, r.GetMeta().GetLabels())
 				}),
+				IgnoreStatusChange(),
```
lahabana, line 337: please add a comment to explain this :)

pkg/kds/context/testdata/full_sync/mesh-service/global.input.yaml
```diff
@@ -0,0 +1,26 @@
+---
+type: Mesh
+name: default
+---
+type: MeshService
+mesh: default
+name: redis-zone-2-ff29vc4685c6f594
```
lobkovilya, line 7: why did you put already "synced" resource to the `*.input.yaml` file? isn't the point of these tests to fill input files only with the resources that originate there and see in golden files they're synced correctly?

pkg/kds/context/testdata/full_sync/mesh-service/zone-1.golden.yaml
```diff
@@ -0,0 +1,86 @@
...
+  - appProtocol: tcp
+    name: some-port
+    port: 16739
+    targetPort: target-port-from-container
+  selector:
+    dataplaneTags:
+      app: redis
+      k8s.kuma.io/namespace: kuma-demo
+  state: Unavailable
+status:
```
lobkovilya, line 68: shouldn't this status be removed?

pkg/kds/v2/reconcile/snapshot_generator.go
```diff
@@ -47,6 +47,12 @@ func IsKubernetes(storeType config_store.StoreType) func(model.Resource) bool {
 	}
 }
 
+func HasStatus() func(model.Resource) bool {
+	return func(r model.Resource) bool {
+		return r.Descriptor().HasStatus
+	}
+}
```
lobkovilya, line 54: You can just have it as
```go
func HasStatus(r model.Resource) bool {
	return r.Descriptor().HasStatus
}
```
and then when using
```go
reconcile_v2.If(reconcile_v2.HasStatus, RemoveStatus()),
```

pkg/kds/context/testdata/full_sync/mesh-service/global.input.yaml
lukidzi: I put it here so I can see that resource is synced to zone-1 and doesn't remove status field

pkg/kds/context/testdata/full_sync/mesh-service/zone-1.golden.yaml
lukidzi: No, this resource was already on the zone and the point is to check if it wasn't removed by sync

lobkovilya (approved):

Issue #10590, opened 2024-06-21
Forbid resources/policies creation with zone label for global


Icarus9913:
### What happened?

Do we need to reinforce our validation for global to forbid those resources/policies creation with `kuma.io/origin: zone` ?

jakubdyszkiewicz:
Yes. Baptiste hit this problem and it caused the confusion. Global was on Kube, we were applying resources with `kuma.io/origin: zone` which broke policy matching.

lahabana:
We should also ensure that federation doesn't expose with zone labels so that this doesn't break the federation flow

PR #13639, opened 2025-05-27
fix(k8s): fix validation of origin label on global

lukidzi:
## Motivation

While testing we noticed that validation was rejecting resource synced from zone to global on k8s

## Implementation information

Rather than validating all resources, we only validate those intended for user creation and not managed administratively.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/10590

### DIFF
--- a/pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
+++ b/pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
@@ -63,15 +63,10 @@ func (c *ResourceAdmissionChecker) isNamespaceAllowed(r core_model.Resource, ns 
 }
 
 func (c *ResourceAdmissionChecker) isResourceAllowed(r core_model.Resource, ns string) *admission.Response {
-	if !c.FederatedZone || !r.Descriptor().IsPluginOriginated {
+	// we don't need to validate fedarated zone and legacy policies
+	if (c.Mode != core.Global && !c.FederatedZone) || !r.Descriptor().IsPluginOriginated {
 		return nil
 	}
-	if !c.DisableOriginLabelValidation && ns == c.SystemNamespace {
-		if origin, ok := core_model.ResourceOrigin(r.GetMeta()); !ok || origin != mesh_proto.ZoneResourceOrigin {
-			return c.resourceIsNotAllowedResponse()
-		}
-	}
-
 	return c.validateLabels(r, ns)
 }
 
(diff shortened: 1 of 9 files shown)


pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -63,15 +63,10 @@ func (c *ResourceAdmissionChecker) isNamespaceAllowed(r core_model.Resource, ns
 }
 
 func (c *ResourceAdmissionChecker) isResourceAllowed(r core_model.Resource, ns string) *admission.Response {
-	if !c.FederatedZone || !r.Descriptor().IsPluginOriginated {
+	// we don't need to validate fedarated zone and legacy policies
```
lobkovilya, line 66: not federated?

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -85,28 +80,33 @@ func (c *ResourceAdmissionChecker) isPrivilegedUser(allowedUsers []string, userI
...
-			return forbiddenResponse(labelsNotAllowedMsg(mesh_proto.ResourceOriginLabel, "global", string(resourceOrigin)))
-		}
-	}
-	if c.Mode != core.Global {
-		resourceOrigin, originPresent := core_model.ResourceOrigin(r.GetMeta())
-		if originPresent && resourceOrigin != mesh_proto.GlobalResourceOrigin {
-			zoneTag, ok := r.GetMeta().GetLabels()[mesh_proto.ZoneTag]
-			if ok && zoneTag != c.ZoneName {
-				return forbiddenResponse(labelsNotAllowedMsg(mesh_proto.ZoneTag, c.ZoneName, zoneTag))
+	if !c.DisableOriginLabelValidation {
```
lobkovilya, line 83: I'd just flip the condition to avoid nesting:
```go
if c.DisableOriginLabelValidation {
    return nil
}

switch c.Mode {
...
```

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -85,28 +80,33 @@ func (c *ResourceAdmissionChecker) isPrivilegedUser(allowedUsers []string, userI
...
+		switch c.Mode {
+		case core.Global:
+			resourceOrigin, originPresent := core_model.ResourceOrigin(r.GetMeta())
+			if !c.DisableOriginLabelValidation && originPresent && resourceOrigin != mesh_proto.GlobalResourceOrigin {
+				return forbiddenResponse(labelsNotAllowedMsg(mesh_proto.ResourceOriginLabel, "global", string(resourceOrigin)))
+			}
+		default:
```
lobkovilya, line 90: nit: since you're using switch you can be specific
```go
case core.Zone, core.Standalone:
```
instead of `default`

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -85,28 +80,33 @@ func (c *ResourceAdmissionChecker) isPrivilegedUser(allowedUsers []string, userI
...
-		resourceOrigin, originPresent := core_model.ResourceOrigin(r.GetMeta())
-		if originPresent && resourceOrigin != mesh_proto.GlobalResourceOrigin {
-			zoneTag, ok := r.GetMeta().GetLabels()[mesh_proto.ZoneTag]
-			if ok && zoneTag != c.ZoneName {
-				return forbiddenResponse(labelsNotAllowedMsg(mesh_proto.ZoneTag, c.ZoneName, zoneTag))
+	if !c.DisableOriginLabelValidation {
+		switch c.Mode {
+		case core.Global:
+			resourceOrigin, originPresent := core_model.ResourceOrigin(r.GetMeta())
+			if !c.DisableOriginLabelValidation && originPresent && resourceOrigin != mesh_proto.GlobalResourceOrigin {
```
lobkovilya, line 87: at this point of the code there is no need to use `c.DisableOriginLabelValidation` (it's `false` since you're in this branch)

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -85,28 +80,33 @@ func (c *ResourceAdmissionChecker) isPrivilegedUser(allowedUsers []string, userI
...
+				return forbiddenResponse(labelsNotAllowedMsg(mesh_proto.ResourceOriginLabel, "global", string(resourceOrigin)))
```
lobkovilya, line 88: we should probably have a constant for `"global"`

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -85,28 +80,33 @@ func (c *ResourceAdmissionChecker) isPrivilegedUser(allowedUsers []string, userI
...
+			}
+		default:
+			resourceOrigin, originPresent := core_model.ResourceOrigin(r.GetMeta())
+			if ns == c.SystemNamespace {
+				if !originPresent || resourceOrigin != mesh_proto.ZoneResourceOrigin {
+					return c.resourceIsNotAllowedResponse()
+				}
+			}
+			if originPresent && resourceOrigin != mesh_proto.GlobalResourceOrigin {
```
lobkovilya, line 97: it's more accurate to check origin is zone (imagine we'll get some new third origin in the future)
```go
if originPresent && resourceOrigin == mesh_proto.ZoneResourceOrigin {
```

pkg/plugins/runtime/k8s/webhooks/resourceadmissionchecker.go
```diff
@@ -85,28 +80,33 @@ func (c *ResourceAdmissionChecker) isPrivilegedUser(allowedUsers []string, userI
...
+				zoneTag, ok := r.GetMeta().GetLabels()[mesh_proto.ZoneTag]
+				if ok && zoneTag != c.ZoneName {
```
lobkovilya, line 99: if we need to check zone even if `DisableOriginLabelValidation` is true?

PR #13654, opened 2025-05-29
Revert "chore(deps): bump golang version from 1.24.2 to 1.24.3"

lukidzi:
Because of https://github.com/golang/go/issues/73617 we need to revert for now

> Changelog: skip


slonka (approved):
could've done "skip/test"

PR #13665, opened 2025-06-03
chore(labels): make the error message about kuma.io/origin more percise

slonka:
## Motivation

I think this can skip the release but there was some confusion around this error. You can have a namespaced policy without this label.

## Implementation information

Add info about system ns


lukidzi:
@slonka ready to merge?

PR #13679, opened 2025-06-06
chore(runtime): remove windows os releated codes and files

Icarus9913:
## Motivation

Currently, we have some Windows os related codes, etc, I plan to remove them for `2.12.x`

### DIFF
--- a/pkg/xds/bootstrap/generator_test.go
+++ b/pkg/xds/bootstrap/generator_test.go
@@ -235,6 +235,58 @@ var _ = Describe("bootstrapGenerator", func() {
 			expectedConfigFile: "generator.custom-config-minimal-request-and-delta.golden.yaml",
 			hdsEnabled:         true,
 		}),
+		Entry("custom config", testCase{
+			dpAuthForProxyType: authEnabled,
+			serverConfig: func() *bootstrap_config.BootstrapServerConfig {
+				return &bootstrap_config.BootstrapServerConfig{
+					Params: &bootstrap_config.BootstrapParamsConfig{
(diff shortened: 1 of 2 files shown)


pkg/xds/bootstrap/generator_test.go
```diff
@@ -235,59 +235,6 @@ var _ = Describe("bootstrapGenerator", func() {
...
-						XdsConnectTimeout:  config_types.Duration{Duration: 2 * time.Second},
-					},
-				}
-			}(),
-			dataplane: defaultDataplane,
-			request: types.BootstrapRequest{
-				Mesh:            "mesh",
-				Name:            "name.namespace",
-				DataplaneToken:  "token",
-				OperatingSystem: "windows",
```
slonka, line 257: it seems like nothing else checks `DataplaneResource:` so I'd just change the OS here and not remove the whole test case

slonka (commented):
One thing, otherwise 👍

pkg/xds/bootstrap/generator_test.go
Icarus9913: The golden file is impacted: https://github.com/kumahq/kuma/pull/13679/files#diff-6f4f8f06bb39fa49da2d1e2097397ca56c255bec0cff314e68f7baf7b07fe057L5-L6

In linux it's `/dev/null`, in windows it's `NUL`.

Icarus9913: Maybe I should remove the `OperatingSystem: "windows"`, let me check it.

bartsmykla (approved):
lgtm

PR #13683, opened 2025-06-06
chore(kuma-dp): add the ability to customize features and dynconfig

lahabana:
This is required for KM to be able to set its own set of feature flags and to move OPA to dynconfig

Part of https://github.com/Kong/kong-mesh/issues/7795


slonka (approved):
LGTM

PR #13684, opened 2025-06-06
feat(KDS): log unique `streamID`

lobkovilya:
## Motivation

Can be useful to see unique connections from the same KDS clients.

## Implementation information

Add interceptors that always add `streamID` into the context.


pkg/kds/mux/server.go
```diff
@@ -104,12 +107,15 @@ func (s *server) Start(stop <-chan struct{}) error {
...
+
 	grpcOptions = append(
 		grpcOptions,
-		grpc.ChainUnaryInterceptor(s.unaryInterceptors...),
+		grpc.ChainStreamInterceptor(
+			append(s.streamInterceptors, kds_middleware.StreamIDStreamInterceptor(&s.streamCount))...,
+		),
+		grpc.ChainUnaryInterceptor(
+			append(s.unaryInterceptors, kds_middleware.StreamIDUnaryInterceptor(&s.streamCount))...,
+		),
```
lahabana, line 118: Instead of doing this here. Shouldn't this be added in the [DefaultContext](https://github.com/kumahq/kuma/blob/7d7b1b6a16715f614804544a06b87d6ac4d42fc1/pkg/kds/context/context.go#L68)?  Seems like this is a more suited place for wiring this no?

lobkovilya: I think both places make sense depending on the interceptor. Stateless interceptors could be placed to DefaultContext, but `StreamIDUnaryInterceptor` keeps streamCounter pointer as a state. 

Nothing wrong with adding the code to DefaultContext like:
```go
func DefaultContext(
	ctx context.Context,
	manager manager.ResourceManager,
	cfg kuma_cp.Config,
) *Context {
    // ...
    var streamCount int64
    return &Context{
        // ...
	ServerStreamInterceptors: []grpc.StreamServerInterceptor{
		kds_middleware.StreamIDStreamInterceptor(&streamCount),
	},
	ServerUnaryInterceptor: []grpc.UnaryServerInterceptor{
		kds_middleware.StreamIDUnaryInterceptor(&streamCount),
	},
    }
```

But I find it less readable and more error prone since now every call to `DefaultContext` is going to produce a different state of `streamCount` variable. 

I prefer `streamCounter` to be server's field https://github.com/kumahq/kuma/blob/93d7675d42dd829dc6b7ef682499c246401c9070/pkg/kds/mux/server.go#L55

lobkovilya: I'm merging the PR, if you strongly feel we need to move it to DefaultContext I'll update in the next PR

PR #13698, opened 2025-06-09
feat(kuma-cp): setup skaffold

Automaat:
## Motivation
Debugging stuff in Kuma is really cumbersome. This PR introduces [Skaffold](https://skaffold.dev) which adds automatic deployments, live reload and super easy debugging setup for Kuma on k8s.


slonka (approved):
The only thing that that might be good to do is to change this default livenessProbe time of 10 minutes / 3 failure attempts to infinity.

Automaat:
skipping probes entirely does not work because of a bug in Skaffold. This should fix it: https://github.com/GoogleContainerTools/skaffold/pull/9831

k8s/demo-kv.yaml
```diff
@@ -0,0 +1,179 @@
...
+            - containerPort: 5050
+              name: http
+---
+apiVersion: kuma.io/v1alpha1
+kind: Mesh
+metadata:
+  name: default
+spec:
+  meshServices:
+    mode: Exclusive
```
lahabana, line 179: Could this use the file https://github.com/kumahq/kuma-counter-demo/blob/main/k8s/001-with-mtls.yaml directly? I'd prefer if we avoid repeating this file.

Automaat: I don't like it either but you cannot reference remote manifests here. You can in `deploy.kubectl` but since this manifest does not install any of the images build with Skaffold it immediately fails with error saying that there is nothing to apply

bartsmykla (approved):

k8s/demo-kv.yaml
lahabana: You could jsut have a make target that downloads it and put it in `build` for .e.g

Issue #13661, opened 2025-06-02
MADR to move to SPIFFE compliant certificates


lahabana:
### Description

This should show the design of Kuma using SPIFFE compliant certs and explain how transparent migration happens.

lahabana:
First step:

- Define scope etc to get a first approval.

PR #13699, opened 2025-06-09
docs(MADR): spiffe compliance - context and user stories

lukidzi:
## Motivation

As we begin work on making Kuma SPIFFE compliant, the first step is to define what we aim to cover and what will be considered out of scope. Additionally, I want to introduce the problem we're facing and outline what we intend to solve.

xref: https://github.com/kumahq/kuma/issues/13661

### DIFF
--- a/docs/madr/decisions/073-spiffe-compliance.md
+++ b/docs/madr/decisions/073-spiffe-compliance.md
@@ -19,7 +19,7 @@ One of the most trusted standards for defining these service identities is SPIFF
 * Tools and rules for issuing, rotating, and checking these identities
 * SPIFFE certificates identify workload, not an API endpoint
 
-Right now, Kuma does something similar by generating SPIFFE-like IDs like `spiffe://<mesh-name>/<service-name>`. But it also adds extra SANs (like `kuma://kuma.io/version/v1`) based on tags. These extra SANs make Kuma's certificates non-compliant with SPIFFE, meaning they can’t be used directly with SPIFFE-based systems like SPIRE, or other tools that expect full compliance. Also, Kuma uses certificates to identify endpoints, whereas SPIFFE-based certificates are used to establish identity via SPIFFE IDs of the workload.
+Right now, Kuma does something similar by generating SPIFFE-like IDs like `spiffe://<mesh-name>/<service-name>`. But it also adds extra SANs (like `kuma://kuma.io/version/v1`) based on tags. These extra SANs make Kuma's certificates non-compliant with SPIFFE, meaning they can’t be used directly with SPIFFE-based systems like SPIRE, or other tools that expect full compliance. Also, Kuma uses certificates to identify endpoints (one Inbound), whereas SPIFFE-based certificates are used to establish identity via SPIFFE IDs of the workload (e.g. Deployment).
 
 In SPIFFE, a Trust Domain is a logical boundary for identity and trust, similar to a DNS domain. Services within the same trust domain can securely identify and trust each other. Communication across different trust domains (such as between organizations or clusters) requires federation, which establishes trust relationships so that identities from one trust domain are recognized and trusted by another.
 


docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,80 @@
...
+
+* A standard way to name services: `spiffe://<trust-domain>/<path>`
+* A strict format for certificates, especially the Subject Alternative Name (SAN) - which must contain exactly one URI (the SPIFFE ID)
+* Tools and rules for issuing, rotating, and checking these identities
+
+Right now, Kuma does something similar by generating SPIFFE-like IDs like `spiffe://<mesh-name>/<service-name>`. But it also adds extra SANs (like `kuma://kuma.io/version/v1`) based on tags. These extra SANs make Kuma's certificates non-compliant with SPIFFE, meaning they can’t be used directly with SPIFFE-based systems like SPIRE, or other tools that expect full compliance.
+
+In SPIFFE, a Trust Domain defines which services are allowed to trust each other. It’s like a boundary — services inside the same trust domain can safely communicate. To talk across trust domains (for example, between two companies or clusters), you need federation, which lets them trust each other's identities.
+
+Kuma currently uses the mesh name as the trust domain. While this works inside Kuma, it doesn’t clearly separate trust boundaries, and it limits how easily Kuma can connect with external services or other meshes.
```
bartsmykla: do you mean other meshes or other service meshes here?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,80 @@
...
+
+This setup causes problems when:
+
+* You want to connect Kuma with other SPIFFE-compliant systems outside the mesh
+* You run multiple clusters or meshes that need to trust each other
```
bartsmykla: You mean service meshes here, right?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,80 @@
...
+* You work in environments that require strict identity standards for security or compliance reasons
+
+To fully support SPIFFE and modern security practices, Kuma should:
+
+* Issue certificates that include only one SPIFFE-compliant URI SAN
```
bartsmykla: Could we mention what is the current state, and what URIs we are currently including?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,80 @@
...
+
+This point is a bit tricky for me, because on one hand, it makes sense to allow users to define a trust domain for an entire mesh (which may span multiple clusters). On the other hand, preventing users from using the same trust domain across different clusters improves security compliance.
+
+By assigning unique trust domains per cluster, we achieve security isolation: if one cluster is compromised, the others remain secure. Additionally, it helps with identity scoping, since we can clearly determine the origin of each identity.
+
+### As a Mesh operator, I want to be able to migrate some workloads to another trust domain without interrupting traffic.
+
+I have a workloads in a trust domain `example.com` and I want to move them to `example.org`. I should be able to do it and don't break the traffic.
+
+
```
bartsmykla, line 110:
```suggestion
```

bartsmykla (commented):
Few comments after the initial read

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,80 @@
...
+To fully support SPIFFE and modern security practices, Kuma should:
+
+* Issue certificates that include only one SPIFFE-compliant URI SAN
+* Let users set the trust domain explicitly, not tie it to the mesh name
+* Allow using other systems (like SPIRE) to manage identities
+
+## User Stories
+
+### As a user, I want a Dataplane (e.g., Service A) inside the mesh to accept mTLS connections from a service (e.g., Service B) outside the mesh.
```
lobkovilya: Does it work today when using `provided` CA backend? If I take that CA and issue a certificate with `spiffe://` and `kuma://` URI SANs and use it for clients outside that mesh, is it going to work today?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,80 @@
...
+
+### As a user, I want to be able to rotate or change the Certificate Authority (CA) used for issuing mTLS identities without interrupting existing connections or causing service downtime.
+
+During the entire process, communication between services must remain secure to ensure uninterrupted, encrypted traffic. This is essential for maintaining zero-trust security guarantees and ensuring operational reliability in production environments.
+
+### As a user, it should not become harder to enable or manage mTLS in either single-zone or multi-zone deployments than it is today.
+
+Currently, users can enable mTLS in both single-zone and multi-zone deployments just by configuring it in the Mesh resource. As we introduce features like trust domains, external identities, or federation, we must preserve this simplicity and avoid adding unnecessary complexity to the mTLS setup process.
+
+### As a Mesh operator, I want to be able to define a trustDomain for an entire mesh or zone
```
lobkovilya: If it's possible to have different trust domains in different zone, do we need another user story to explicitly allow traffic from/to different trust domains?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,80 @@
...
+
+It makes sense to allow users to define a trust domain for a specific part of the infrastructure: whether that's a zone or an entire mesh. However, this becomes tricky when balanced against the need for stricter security guarantees in other areas.
+
+### As a Mesh operator, I should be prevented from configuring the same trust domain in multiple zones when not using SPIRE as the identity provider.
```
lobkovilya: Do we expect users to have multiple trust domains within the same cluster? i.e. trust domain per namespace?

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: I think it should work, but we haven't tested it. The problem is once we base them on trust domain and we want to support communication between them

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: I think we should have a trust domain per zone and don't allow per namespace. Maybe per Mesh make sense but in this case it break the idea of trust domain in individual cluster

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: clarified

docs/madr/decisions/073-spiffe-compliance.md
lobkovilya: I think you need to rephrase the user story then, because in the current state Kuma already kind of implements it.

docs/madr/decisions/073-spiffe-compliance.md
lobkovilya: Maybe it worth adding to "Out of scope" then, user can't go more granular with trust domain than a zone.

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, but we don't want to lose it so I think we should have it as a user story. There is another user story which covers the communication when there is a different trust domain

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,87 @@
...
+Technical Story: https://github.com/kumahq/kuma/issues/13659
+
+## Context and Problem Statement
+
+In today’s service mesh systems, secure communication between services is essential. To make that possible, each service needs a clear and trusted identity: so we know exactly who is talking to whom.
+
+One of the most trusted standards for defining these service identities is SPIFFE (Secure Production Identity Framework for Everyone). It gives us:
+
+* A standard way to name services: `spiffe://<trust-domain>/<path>`
+* A strict format for certificates, especially the Subject Alternative Name (SAN) - which must contain exactly one URI (the SPIFFE ID)
```
slonka, line 18: There was a section in MeshTLS that listed all the requirements if anyone is interested - https://docs.google.com/document/d/1zA1q1hD3PN7c1EsnxkwggtBR08-8i5voZG3yNJyjQfI/edit?tab=t.0#bookmark=id.dbm50hrdz1ec

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,87 @@
...
+Right now, Kuma does something similar by generating SPIFFE-like IDs like `spiffe://<mesh-name>/<service-name>`. But it also adds extra SANs (like `kuma://kuma.io/version/v1`) based on tags. These extra SANs make Kuma's certificates non-compliant with SPIFFE, meaning they can’t be used directly with SPIFFE-based systems like SPIRE, or other tools that expect full compliance.
+
+In SPIFFE, a Trust Domain defines which services are allowed to trust each other. It’s like a boundary — services inside the same trust domain can safely communicate. To talk across trust domains (for example, between two companies or clusters), you need federation, which lets them trust each other's identities.
+
+Kuma currently uses the mesh name as the trust domain. While this works inside Kuma, it doesn’t clearly separate trust boundaries, and it limits how easily Kuma can connect with external services or other applications in the same trust domain.
+
+This setup causes problems when:
+
+* You want to connect Kuma with other SPIFFE-compliant systems outside the mesh
+* You run multiple clusters that need to trust each other
```
slonka, line 31: I'm not sure I 100% understand this - currently if you run multiple clusters (zones) and they share the same `mesh` then everything is fine right?

docs/madr/decisions/073-spiffe-compliance.md
slonka: I feel like it could be merged with `As a user, I want to be able to communicate with applications in the different trust domain.` right?

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, but all have the same trust domain which might not be correct since each cluster can have a different one

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Not exactly, since this application might be in the same trust domain

lahabana: Yes I think 2 stories keeps things more specific and clearer.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,93 @@
...
+
+During the entire process, communication between services must remain secure to ensure uninterrupted, encrypted traffic. This is essential for maintaining zero-trust security guarantees and ensuring operational reliability in production environments.
+
+### As a user, it should not become harder to enable or manage mTLS in either single-zone or multi-zone deployments than it is today.
+
+Currently, users can enable mTLS in both single-zone and multi-zone deployments just by configuring it in the Mesh resource. As we introduce features like trust domains, external identities, or federation, we must preserve this simplicity and avoid adding unnecessary complexity to the mTLS setup process.
+
+### As a user, I want to be able to communicate with applications in the different trust domain.
+
+If a user has a different trust domain in each zone, it should be possible to allow communication between these zones.
```
lahabana:
```suggestion
In the presence of multiple trust domains. Users should be able to trust each domains.
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,93 @@
...
+
+### As a Mesh operator, I want to be able to define a trustDomain for an entire mesh or zone
```
lahabana: I think this is more generic:

> As a Mesh operator, I want to be able to define trustDomain at different levels.

For example if I'm doing a CA migration I might have this existing stuff:

```yaml
spec:
  targetRef:
     kind: Mesh
  conf:
    backend: current
    type: builtin
    trustDomainTemplate: "spiffe://current/ns/..."
```

Next:

```yaml
spec:
  targetRef:
     kind: Dataplane
     labels:
       maturity: beta
  conf:
    backend: new
    type: builtin
    trustDomainTemplate: "spiffe://new/ns/..."
```

And then eventually removing the new `current` one.

In this case at all times communication shouldn't be interrupted and trust should work.
This is IMO how you do easy CA rotation (our migration to SPIRE)

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,93 @@
...
+
+It makes sense to allow users to define a trust domain for a specific part of the infrastructure: whether that's a zone or an entire mesh. However, this becomes tricky when balanced against the need for stricter security guarantees in other areas.
```
lahabana: How is that tricky?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,93 @@
...
+
+### As a Mesh operator, I should be prevented from configuring the same trust domain in multiple zones when not using SPIRE as the identity provider.
```
lahabana: I don't think we can reasonably do this. Zones work idenpendently, people can do dumb things and that's life ;)

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,93 @@
...
+
+This point is a bit tricky for me, because on one hand, it makes sense to allow users to define a trust domain for an entire mesh (which may span multiple clusters). On the other hand, preventing users from using the same trust domain across different clusters improves security compliance.
+
+By assigning unique trust domains per cluster, we achieve security isolation: if one cluster is compromised, the others remain secure. Additionally, it helps with identity scoping, since we can clearly determine the origin of each identity.
+
+Default trust domain: Mesh name
```
lahabana: Actually I think the default trustDomain should be:

```
spiffe://<mesh>.<clusterId>.kuma.io/
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,93 @@
...
+
+### As a user, I want certificate to be valid for a short period of time
+
+Based on the [specification](https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/#spiffe-workload-api) certificates should be short-lived.
+
+## Out of scope
+
+### As a user, I want to specify certificates for a specific outbound
+
+You should use MeshExternalService in this case to configure the outbound connection and provide the required certificates.
```
lahabana: I think this clarifies a common misunderstanding!

Spiffe certs identify WORKLOADS they don't identify APIs or endpoints. That's an entry in the MeshService (identities): https://github.com/kumahq/kuma/blob/a193368f55d9e633876df52f5d87e4f40afee831/pkg/core/resources/apis/meshservice/api/v1alpha1/meshservice.go#L87-L97

I think this may be useful clarify this in the context and problem section maybe?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,93 @@
...
+
+In this scenario, you should also use MeshExternalService and include the necessary certificates in the configuration to enable direct mTLS communication.
+
+### As a user, I want to use two different sources of CA (Kuma and SPIRE) at the same time
+
+This is a complex use case and can make the configuration error prone and difficult to manage.
+Instead, we recommend choosing one of the following approaches:
+
+* Use Kuma as the CA and configure other zones or services to trust its certificates (i.e., share the same root/intermediate CA).
+* Use SPIRE as the CA and federate it across zones and meshes to provide a unified identity source.
```
lahabana: I think this out of scope user story makes Kuma --> Spire migration impossible.

Indeed to do this you need some workloads on Kuma CA and some on Spire.

I think what you may want to say is:

- For a dataplane identity can come from only one place (obviously there's a single ID)
- For the trust domains they must come from a single place (either SPIRE or Kuma never both for a single dataplane).
- 1 dataplane can have its identity from Kuma cp and 1 dataplane can have its identity come from SPIRE, communication between DP should work

We'll need to figure the federation setup for this to work (maybe Kuma could expose a federation endpoint?)

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,93 @@
...
+
+### As a user, I want to have trust domain per namespace
+
+I think we shouldn't allow higher granularity than zone or mesh.
```
lahabana: What's missing is supporting users to move directly from non SPIFFE certs to SPIRE. They will need to first go through SPIFFE compliant Kuma-cp issued certs and then go to SPIRE.

docs/madr/decisions/073-spiffe-compliance.md
lahabana: However, I think it's ok here to say the Kuma-cp --> Spire migration must be made possible but won't be covered in this MADR.

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: It was a bit a conflict with `### As a Mesh operator, I should be prevented from configuring the same trust domain in multiple zones when not using SPIRE as the identity provider.` so I feel like removing this step and keeping `As a Mesh operator, I want to be able to define a trustDomain for an entire mesh or zone` makes more sense and it's clear

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Clarified

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Added note

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Added section about migration and rephrased this one

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Added

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,95 @@
...
+
+In today’s service mesh systems, secure communication between services is essential. To make that possible, each service needs a clear and trusted identity: so we know exactly who is talking to whom.
+
+One of the most trusted standards for defining these service identities is SPIFFE (Secure Production Identity Framework for Everyone). It gives us:
+
+* A standard way to name services: `spiffe://<trust-domain>/<path>`
+* A strict format for certificates, especially the Subject Alternative Name (SAN) - which must contain exactly one URI (the SPIFFE ID)
+* Tools and rules for issuing, rotating, and checking these identities
+* SPIFFE certificates identify workload, not an endpoint or API
+
```
Icarus9913: I don't fully understand this sentence here with the above context. Could you please explain some details?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,95 @@
...
+Right now, Kuma does something similar by generating SPIFFE-like IDs like `spiffe://<mesh-name>/<service-name>`. But it also adds extra SANs (like `kuma://kuma.io/version/v1`) based on tags. These extra SANs make Kuma's certificates non-compliant with SPIFFE, meaning they can’t be used directly with SPIFFE-based systems like SPIRE, or other tools that expect full compliance. Also, Kuma uses certificates to identify endpoints, whereas SPIFFE-based certificates are used to establish identity via SPIFFE IDs of the workload.
+
+In SPIFFE, a Trust Domain defines which services are allowed to trust each other. It’s like a boundary — services inside the same trust domain can safely communicate. To talk across trust domains (for example, between two companies or clusters), you need federation, which lets them trust each other's identities.
```
Icarus9913: How about this?

```suggestion
In SPIFFE, a Trust Domain is a logical boundary for identity and trust, similar to a DNS domain. Services within the same trust domain can securely identify and trust each other. Communication across different trust domains (such as between organizations or clusters) requires federation, which establishes trust relationships so that identities from one trust domain are recognized and trusted by another.
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,95 @@
...
+
+* You want to connect Kuma with other SPIFFE-compliant systems outside the mesh
+* You run multiple clusters that need to trust each other
+* You work in environments that require strict identity standards for security or compliance reasons
+
+To fully support SPIFFE and modern security practices, Kuma should:
+
+* Issue certificates that include only one SPIFFE-compliant URI SAN (currently we generate `spiffe://<mesh>/<service>` for each inbound and `kuma://kuma.io/<tag_name>/<tag_value>` for each tag)
+* Let users set the trust domain explicitly, not tie it to the mesh name
+* Allow using other systems (like SPIRE) to manage identities
```
Icarus9913, line 39: > Let users set the trust domain explicitly, not tie it to the mesh name

So, it's a required property that the user must set, with no default value?

Icarus9913 (commented):
@lukidzi, May I know if this also includes the "CP send requests to DP for Envoy Admin Communication" use case?

lukidzi:
> CP send requests to DP for Envoy Admin Communication

It's not related to this change

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: > SPIFFE certificates identify workload, not an endpoint or API

Each application might expose multiple endpoints. Currently, we create certificate for dataplane but we add identities of all endpoints to this certificate as SAN. That is not a good way, and certificate should just identify the workload and not each certificate for a separate API endpoint.

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: We might introduce same default but it will be in a next PR

docs/madr/decisions/073-spiffe-compliance.md
lahabana: So what you want is to have "zones" that have different trustDomains.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,136 @@
...
+
+In SPIFFE, a Trust Domain is a logical boundary for identity and trust, similar to a DNS domain. Services within the same trust domain can securely identify and trust each other. Communication across different trust domains (such as between organizations or clusters) requires federation, which establishes trust relationships so that identities from one trust domain are recognized and trusted by another.
+
+Kuma currently uses the mesh name as the trust domain. While this works inside Kuma, it doesn’t clearly separate trust boundaries, and it limits how easily Kuma can connect with external services or other applications in the same trust domain.
+
+This setup causes problems when:
+
+* You want to connect Kuma with other SPIFFE-compliant systems outside the mesh
+* You run multiple clusters that need to trust each other
+* You work in environments that require strict identity standards for security or compliance reasons
```
lahabana, line 32: missing:

- You have identity not coming from Kuma (.e.g SPIRE)

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,136 @@
...
+
+To fully support SPIFFE and modern security practices, Kuma should:
+
+* Issue certificates that include only one SPIFFE-compliant URI SAN (currently we generate `spiffe://<mesh>/<service>` for each inbound and `kuma://kuma.io/<tag_name>/<tag_value>` for each tag)
+* Let users set the trust domain explicitly, not tie it to the mesh name
+* Allow using other systems (like SPIRE) to manage identities
```
lahabana, line 39: I think one very important thing:

- Kuma should separate issuing identity from validating trust.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,136 @@
...
+
+### As a user, I want to be able to rotate or change the Certificate Authority (CA) used for issuing mTLS identities without interrupting existing connections or causing service downtime.
+
+During the entire process, communication between services must remain secure to ensure uninterrupted, encrypted traffic. This is essential for maintaining zero-trust security guarantees and ensuring operational reliability in production environments.
+
+### As a user, it should not become harder to enable or manage mTLS in either single-zone or multi-zone deployments than it is today.
+
+Currently, users can enable mTLS in both single-zone and multi-zone deployments just by configuring it in the Mesh resource. As we introduce features like trust domains, external identities, or federation, we must preserve this simplicity and avoid adding unnecessary complexity to the mTLS setup process.
+
+### As a user, I want to be able to communicate with applications in the different trust domain.
```
lahabana:
```suggestion
### As a user, I want to be able to communicate with applications in different trust domains.
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,136 @@
...
+
+I have a workloads in a trust domain `example.com` and I want to move them to `example.org`. I should be able to do it and don't break the traffic.
+
+### As a user, I want certificate to be valid for a short period of time
+
+Based on the [specification](https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/#spiffe-workload-api) certificates should be short-lived.
+
+### As a user, I want to migrate from Kuma non-SPIFFE complient certificates to SPIRE
+
+We should allow users to migrate from Kuma issued certificates, which are not SPIFFE-compliant, to using SPIRE. This may require an intermediate step to first migrate to SPIFFE-compliant certificates, but the details will be specified later.
```
lahabana: So that's a future story right?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Future User Stories (things that are going to be implemented in future releases)
+
+### As a user, I want to have a SPIFFE-compliant SAN in a non-Kubernetes environment
+
+It would be beneficial to support connectivity between different environments. Unfortunately, our services currently lack proper identity on Universal, which makes this challenging.
+
+### As a user, I want to federate different trust domains using the Federation API exposed by the control plane
+
+It should be possible to federate different trust domains without creating additional resources, only by using the federation endpoint exposed by the control plane. This would be a valuable feature, enabling interoperability between different SPIFFE-compliant systems.
```
lahabana, line 89: The federation endpoint is part of SPIRE not spiffe no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,136 @@
...
+### As a user, I want my dataplanes to communicate with Istio services without going through a gateway
+
+In this scenario, you should also use MeshExternalService and include the necessary certificates in the configuration to enable direct mTLS communication.
+
+### As a user, I want to use two different sources of CA (Kuma and SPIRE) at the same time
+
+This is a complex use case and can make the configuration error prone and difficult to manage.
+We want to add some limitation to avoid issues:
+* A dataplane's identity must come from a single source.
+* For trust domains, identities must be issued by one system only — either SPIRE or Kuma, never both for the same dataplane.
```
lahabana: We're mixing things here I think.
I think we're getting the info (identity and trust) either from Spire or Kuma never both.
However, it's possible to add the trust coming from Spire inside Kuma and to implement SPIRE federation api in Kuma to retrieve Trust.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,136 @@
...
+* It's valid for one dataplane to get its identity from Kuma and another from SPIRE — communication between such dataplanes should still work.
+
+To enable this setup, we’ll need to establish a federation mechanism — potentially by having Kuma expose a federation endpoint.
+
+### As a user, I want to have trust domain per namespace
+
+I think we shouldn't allow higher granularity than zone or mesh.
```
lahabana: I don't understand this limitation? AFAIU it shouldn't be complicated to just run 1 dataplane on a different identity.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,136 @@
...
+
+### As a user, I want certificate to be valid for a short period of time
+
+Based on the [specification](https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/#spiffe-workload-api) certificates should be short-lived.
+
+### As a user, I want to migrate from Kuma non-SPIFFE complient certificates to SPIRE
+
+We should allow users to migrate from Kuma issued certificates, which are not SPIFFE-compliant, to using SPIRE. This may require an intermediate step to first migrate to SPIFFE-compliant certificates, but the details will be specified later.
+
+## Future User Stories (things that are going to be implemented in future releases)
```
lahabana, line 81: I think one is automatic CA rotation.

It's fairly straightforward:
```
Preparation Phase:

Generate a new root CA certificate and private key
The new root CA is prepared but not yet active
Both old and new root CAs exist simultaneously during the transition

Bundle Update:

The SPIFFE trust bundle is updated to include both the old and new root CA certificates
This bundle is distributed to all workloads through the Workload API
Workloads now trust certificates issued by either CA

Gradual Transition:

New X.509-SVIDs (SPIFFE Verifiable Identity Documents) are issued by the new CA
Existing SVIDs continue to be valid until they expire
Services can verify certificates from both CAs during this period

Completion:

Once all SVIDs have been renewed with the new CA, the old root CA is removed from the trust bundle
Only the new CA remains active
```

However, we shouldn't do this just right now.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,136 @@
...
+To enable this setup, we’ll need to establish a federation mechanism — potentially by having Kuma expose a federation endpoint.
+
+### As a user, I want to have trust domain per namespace
+
+I think we shouldn't allow higher granularity than zone or mesh.
+
+### As a user, I want to provide MeshTrafficPermission for SPIFFEID
+
+This should be covered by a separate MADR in a separate [issue](https://github.com/kumahq/kuma/issues/12374).
+
```
lahabana, line 134: I think these stories are worth adding:

- Changing trustDomain without changing CA. Or can we do it?

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Federation.md#2-introduction

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: I feel like we can do it. Since we can add another Trust with the same CA

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Rephrased

docs/madr/decisions/073-spiffe-compliance.md
lobkovilya: From SVID [docs](https://github.com/spiffe/spiffe/blob/main/standards/X509-SVID.md?plain=1#L54) on the CA certificate:

> A signing certificate SHOULD itself be an SVID. If present, the SPIFFE ID of a signing certificate MUST NOT have a path component, and SHOULD reside in the trust domain of any leaf SVIDs it issues. A signing certificate MAY be used to issue further signing certificates in the same or different trust domains.

So if trustDomain is part of the CA cert then I believe we can't change trust domain without changing the CA? But I'm a bit confused by their wording `If present`. So it it up to us to include SPIFFE ID into the CA certificate?

lukidzi: Seems so, we should create a CA with SPIFFEID without path part, but `If present` suggest it's optional

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -6,17 +6,22 @@ Technical Story: https://github.com/kumahq/kuma/issues/13659
...
 * A strict format for certificates, especially the Subject Alternative Name (SAN) - which must contain exactly one URI (the SPIFFE ID)
 * Tools and rules for issuing, rotating, and checking these identities
+* SPIFFE certificates identify workload, not an endpoint or API
 
-Right now, Kuma does something similar by generating SPIFFE-like IDs like `spiffe://<mesh-name>/<service-name>`. But it also adds extra SANs (like `kuma://kuma.io/version/v1`) based on tags. These extra SANs make Kuma's certificates non-compliant with SPIFFE, meaning they can’t be used directly with SPIFFE-based systems like SPIRE, or other tools that expect full compliance.
+Right now, Kuma does something similar by generating SPIFFE-like IDs like `spiffe://<mesh-name>/<service-name>`. But it also adds extra SANs (like `kuma://kuma.io/version/v1`) based on tags. These extra SANs make Kuma's certificates non-compliant with SPIFFE, meaning they can’t be used directly with SPIFFE-based systems like SPIRE, or other tools that expect full compliance. Also, Kuma uses certificates to identify endpoints, whereas SPIFFE-based certificates are used to establish identity via SPIFFE IDs of the workload.
```
slonka: What do you mean by "endpoints" here? What's the difference between workload vs endpoint?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -25,63 +30,132 @@ This setup causes problems when:
...
 
+Example:
+I want to move from using certificates issued by Kuma to SPIRE in different trust domain. 
+
 ### As a user, it should not become harder to enable or manage mTLS in either single-zone or multi-zone deployments than it is today.
 
 Currently, users can enable mTLS in both single-zone and multi-zone deployments just by configuring it in the Mesh resource. As we introduce features like trust domains, external identities, or federation, we must preserve this simplicity and avoid adding unnecessary complexity to the mTLS setup process.
 
-### As a user, I want to be able to communicate with applications in the different trust domain.
+Release `2.12.x`: We are going to sync only Trust from global to zones, but not from zone to other zones.
```
slonka: Do we sync "trust" from zones to other zones now? What object is that? Is it somewhere inside MeshService?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -25,63 +30,132 @@ This setup causes problems when:
...
+
+Kuma currently supports various CA (Certificate Authority) backends for identity management. The new mTLS system should at least support the same set of CA providers, to ensure a smooth transition and preserve compatibility with existing deployments.
+
+### As a user, I want to migrate from Kuma non-SPIFFE complient certificates to SPIRE
+
+We should allow users to migrate from Kuma issued certificates, which are not SPIFFE-compliant, to using SPIRE. This may require an intermediate step to first migrate to SPIFFE-compliant certificates, but the details will be specified later.
+
+### As a user, I want to automatically rotate the CA
+
+This could be straightforward to implement, but we shouldn't prioritize it at the moment.
```
slonka, line 109: I think it makes sense to add @lahabana comment here https://github.com/kumahq/kuma/pull/13699/files/7bf39d3fd8dd74da03cfbdf22f1c832d120f9fff..03927ffdadd7e6e30b897d0b7d6a8ef431502078#r2152062929

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -25,63 +30,132 @@ This setup causes problems when:
...
 
 In this scenario, you should also use MeshExternalService and include the necessary certificates in the configuration to enable direct mTLS communication.
 
-### As a user, I want to use two different sources of CA (Kuma and SPIRE)
+### As a user, I want to use two different sources of CA (Kuma non-SPIFFE compliant and SPIRE) at the same time
 
 This is a complex use case and can make the configuration error prone and difficult to manage.
-Instead, we recommend choosing one of the following approaches:
+We want to add some limitation to avoid issues:
+* We should be add a Trust to SPIRE service while using Kuma certificate.
```
slonka: nitpick:

```suggestion
* We should be able to add a Trust to SPIRE service while using Kuma certificate.
```

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: It seems like SPIRE description, I would stay without specific description since it's not implementation but Context and Problem statement

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: * Workload - is a e.g.: Deployment
* Endpoint - is just one Inbound

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Maybe it shouldn't be here since its more design. I will remove

docs/madr/decisions/073-spiffe-compliance.md
slonka:
```suggestion
Right now, Kuma does something similar by generating SPIFFE-like IDs like `spiffe://<mesh-name>/<service-name>`. But it also adds extra SANs (like `kuma://kuma.io/version/v1`) based on tags. These extra SANs make Kuma's certificates non-compliant with SPIFFE, meaning they can’t be used directly with SPIFFE-based systems like SPIRE, or other tools that expect full compliance. Also, Kuma uses certificates to identify endpoints (one Inbound), whereas SPIFFE-based certificates are used to establish identity via SPIFFE IDs of the workload (e.g. Deployment).
```

slonka (approved):

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Added note

lobkovilya (approved):

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -0,0 +1,159 @@
...
+
+This should be covered by a separate MADR in a separate [issue](https://github.com/kumahq/kuma/issues/12374).
+
+### As a user, I want to change a trust domain without changing the CA
+
+Based on the SPIFFE [specification](https://github.com/spiffe/spiffe/blob/main/standards/X509-SVID.md?plain=1#L54), it seems that it's not supported, so we shouldn't support it as well
+
+## Summary
+
+Release 2.12
```
lahabana, line 141: We need Justin to approve this

PR #13732, opened 2025-06-13
fix(oapi): fix generating code from oapi schema

Automaat:
## Motivation
When we generate oapi schema from proto resource, rest schema needs to be in the same directory as components definition. Thanks to this, we avoid issues with schema referencing. 

Previously, when we run make check, generate step was failing silently with errors:
```
error loading swagger spec in api/openapi/specs/api.yaml
: failed to load OpenAPI specification: error resolving reference "/specs/protoresources/dataplaneoverview/schema.yaml#/components/schemas/DataplaneOverview": open /specs/protoresources/dataplaneoverview/schema.yaml: no such file or directory
```
Which caused skipping of go code generation from `api/openapi/specs/api.yaml` file. With this change we don't reference schema from other locations. We only combine it at the end while generating docs.


Icarus9913:
I don't fully understand this PR's goal, it looks like you just moved **2 routes** from `api.yaml` to another new `rest.yaml` and moved some response structure definition like `BadRequest` to `error_schema.yaml` file.

Can you clarify the problems we hit previously?

Automaat:
@Icarus9913 I've updated pr description

slonka (approved):

Issue #13436, opened 2025-04-22
Add support for `Mesh*Route` in `spec.to[].targetRef` in MeshLoadBalancingStrategy policy


lobkovilya:
### Description

Implement as described in [MADR](https://github.com/kumahq/kuma/pull/10858)

PR #13765, opened 2025-06-18
refactor(MeshLoadBalancingStrategy): rewrite using builders

lobkovilya:
## Motivation

Prepare plugin code for routes integration in https://github.com/kumahq/kuma/issues/13436 

## Implementation information

* added new code builders to `pkg/envoy/builders/`
* updated the MLBS plugin's code 
* stopped using UpdateHTTPConnectionManager where it's unnecessary


pkg/plugins/policies/meshloadbalancingstrategy/plugin/v1alpha1/plugin.go
```diff
@@ -299,37 +407,41 @@ func (p plugin) configureGateway(
...
-							proxy,
-							rules.ResourceRules,
-							*realRef,
-							rs,
-							resources,
-							endpoints,
-						); err != nil {
-							return err
+						svcCtx := rctx.
+							WithID(kri.NoSectionName(*realRef)).
```
lukidzi, line 430: Why do we call with ID twice?

lobkovilya: The `ResourceRules` map can contain configurations for both the entire `MeshService` and for individual ports of the `MeshService`.

`ResourceContext` is simply a list of IDs, when calling `Conf()`, first we're trying to lookup conf for the last added ID. If it doesn't exist we're going to the second to last added, etc. 

By adding `WithID(kri.NoSectionName(*realRef)).WithId(*realRef)`, we're guarantee to fall back to MeshService's conf if there is no specific conf for MeshService's port.

PR #13770, opened 2025-06-20
feat(MeshLoadBalancingStrategy): move HashPolicies field to root level

lobkovilya:
## Motivation

Implements [MADR-074](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/074-mlbs-hash-policies.md). 

Moving `HashPolicies` field to the root level is required for implementing https://github.com/kumahq/kuma/issues/13436

## Implementation information

* Added new field `default.hashPolicies`
* Added validation
* Updated envoyconfig tests


pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go
```diff
@@ -33,6 +33,11 @@ type Conf struct {
 	LocalityAwareness *LocalityAwareness `json:"localityAwareness,omitempty"`
 	// LoadBalancer allows to specify load balancing algorithm.
 	LoadBalancer *LoadBalancer `json:"loadBalancer,omitempty"`
+	// HashPolicies specify a list of request/connection properties that are used to calculate a hash.
+	// These hash policies are executed in the specified order. If a hash policy has the “terminal” attribute
+	// set to true, and there is already a hash generated, the hash is returned immediately,
+	// ignoring the rest of the hash policy list.
+	HashPolicies *[]HashPolicy `json:"hashPolicies,omitempty"`
```
lukidzi, line 40: Should we deprecate the previous usage? I think it cant be now since the new version is not yet available but we probably should keep it in mind

lukidzi (approved):

pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go
lobkovilya: I've opened an issue https://github.com/kumahq/kuma/issues/13793

Issue #12229, opened 2024-12-10
Make kumactl apply more verbose


Automaat:
### Description

We should make `kumactl` more verbose on success, like `kubectl`. When you successfully apply resource with `kubectl` you get response:
```
meshtrafficpermission.kuma.io/redis created
```
When you apply resource with `kumactl` it is silent on success

PR #13771, opened 2025-06-20
feat(kumactl): better kumactl apply output

### LINKED ISSUE #10099: Better presentation of method not allowed in kumactl
### What happened?

We just print `Error: Method not allowed`, but we have more info in response (that you can't apply Mesh on zone CP)

lahabana:
## Motivation

The output of kumactl apply when there are multiple resources is missing context. This is very poor UX, I've noticed this debugging e2e tests where I had a bunch of yamls and just an error "something is wrong".

## Implementation information

- We now return errors for each entry in a multi resource yaml. This makes it easier as all errors are returned at once
- We now log to stdout if the resource was updated or created.
- Added golden tests for outputs to make validating this behavior easier
- Avoid having golden files when the matching payload is empty (if no file it means empty).

Fix #10099


app/kumactl/cmd/apply/apply.go
```diff
@@ -133,10 +141,15 @@ $ kumactl apply -f https://example.com/resource.yaml
...
+					warnings, isUpdate, err := upsert(cmd.Context(), pctx.Runtime.Registry, rs, resource)
 					if err != nil {
-						return err
+						return fmt.Errorf("resource type=%q mesh=%q name=%q: failed server side: %s", resource.Descriptor().Name, resource.GetMeta().GetMesh(), resource.GetMeta().GetName(), err.Error())
+					}
+					action := "Created"
+					if isUpdate {
+						action = "Updated"
 					}
+					_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Resource type=%q mesh=%q name=%q %s\n", resource.Descriptor().Name, resource.GetMeta().GetMesh(), resource.GetMeta().GetName(), action)
```
lukidzi:
```suggestion
					_, _ = fmt.Fprintf(cmd.OutOrStdout(), "resource type=%q mesh=%q name=%q %s\n", resource.Descriptor().Name, resource.GetMeta().GetMesh(), resource.GetMeta().GetName(), action)
```

app/kumactl/cmd/apply/apply_test.go
```diff
@@ -117,7 +117,7 @@ var _ = Describe("kumactl apply", func() {
 		// given
 		rootCmd.SetArgs([]string{
 			"--config-file", filepath.Join("..", "testdata", "sample-kumactl.config.yaml"),
-			"apply", "-f", filepath.Join("testdata", "apply-dataplane.yaml"),
+			"apply", "-f", filepath.Join("testdata", "golden", "apply-dataplane.input.yaml"),
```
lukidzi, line 120: can we cleanup files from `testdata/*.yaml`?

lukidzi:
Probably fix: https://github.com/kumahq/kuma/issues/12229

app/kumactl/cmd/apply/apply_test.go
lahabana: we could but the files in there are still useful...

lukidzi:
Needs updated golden files

lukidzi (approved):

PR #13773, opened 2025-06-20
feat(dp-server): exposed grpc reflection server in the dp-server

m4l1c1ou5:
## Motivation
Can be useful in debugging the cp-dp sync issues using cli tools like grpcurl

## Implementation information
Added the reflection server using google.golang.org/grpc/reflection package


pkg/dp-server/server/server.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"google.golang.org/grpc/reflection"
```
slonka, line 7: you need to move the import to the right place - I think `make format` will take care of that

slonka (commented):
Does this pose any security threat if turned on always? Should this be behind a flag? Also can you add to `developer.md` or a separate file in `docs/guides` on how to use `grpcurl` to debug the grpc server?

bartsmykla (changes requested):
I’m not sure it should be enabled by default, I still need to look into it. It makes sense to have an option to turn it on though.

If your gRPC API is accessible to the public, you may not want to expose the reflection service, since this can be a security issue. You need to decide what balance between security and ease of use works best for you and your users.
https://grpc.io/docs/guides/reflection/#tips

m4l1c1ou5:
I evaluated how it can be helpful, initially i thought it might make it easier to debug but i couldn't find any useful way on how will exposing reflection server will help in debugging.
So didn't evaluate it on the security perspective

slonka:
Same as https://github.com/kumahq/kuma/pull/13774#issuecomment-3440705649.

PR #13774, opened 2025-06-20
feat(kuma-cp): exposed grpc reflection server in the intercp communication

m4l1c1ou5:
## Motivation
Can be useful in debugging the intercp sync issues using cli tools like grpcurl

## Implementation information
Added the reflection server using google.golang.org/grpc/reflection package

### DIFF
--- a/pkg/intercp/server/server.go
+++ b/pkg/intercp/server/server.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"google.golang.org/grpc/reflection"
 	"net"
 	"net/http"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


slonka (commented):
Pretty much the same as the other PR, would be great to make the linter pass and add a doc / guide on what can be achieved using `grpcurl`.

bartsmykla (changes requested):
I’m not sure it should be enabled by default, I still need to look into it. It makes sense to have an option to turn it on though.

If your gRPC API is accessible to the public, you may not want to expose the reflection service, since this can be a security issue. You need to decide what balance between security and ease of use works best for you and your users.
https://grpc.io/docs/guides/reflection/#tips

m4l1c1ou5:
I evaluated how it can be helpful, initially i thought it might make it easier to debug but i couldn't find any useful way on how will exposing reflection server will help in debugging.
So didn't evaluate it on the security perspective

slonka:
@m4l1c1ou5 I'm closing this since there is no activity, please reopen with more info if you think this might be useful.

Issue #13266, opened 2025-03-28
MADR on naming "internal" Envoy resources


lobkovilya:
### Description

Some resources in Envoy don't correlate with the real resources in the store. That's why we can't use `KRI` formatted name for these resources. We need to revisit names for the following resources:
* Secrets
* Internal Listeners and Clusters
* Default routes (when there are no MeshHTTPRoute/MeshTCPRoute policies)
* MeshPassthrough resources

PR #13485, opened 2025-04-29
docs(MADR): defining and migrating to consistent naming for non-system Envoy resources and stats

lukidzi:
## Motivation

We want to build new functionalities based on metrics and we should keep them consitant and easy to reference in config

## Implementation information

Fix https://github.com/kumahq/kuma/issues/13264

@slonka edit: poc branch https://github.com/lukidzi/kuma/tree/stats-renaming

PR #13783, opened 2025-06-23
docs(MADR): naming system envoy resources

slonka:
## Motivation

Best described in the issue: https://github.com/kumahq/kuma/issues/13266

## Implementation information

Just the design, no implem.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13266

> Changelog: skip


docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,152 @@
...
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/13266
+
+## Context and Problem Statement
+
+We use resource names to identify listeners, clusters, routes, virtual hosts and secrets in the xDS APIs.
+These names appear in metrics, logs, and debug endpoints (config_dump), and they are often used for monitoring, filtering, and diagnostics.
+
+It's important to know the difference between internal resources (the ones that are defined by Kuma itself, and they usually do not concern the end users) and external resources (the ones that are defined indirectly by the user and the users are interested in).
```
lobkovilya: nit: calling them "external" doesn't feel right, maybe "internal" and "user-defined"?

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,152 @@
...
+    metrics:
+      include:
+        match_type: regexp
+        metric_names: [".*"]  # apply to all metrics
+        attributes:
+          - key: envoy_cluster
+            value: "^_.*"  # matches if envoy_cluster starts with _
+```
+
+**Is this a valid use case if we have profiles in MeshMetric?**
```
lobkovilya: I think with profiles you're just shifting the complexity of maintaining the list of internal resources to a different place. So it's still a valid use case

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,152 @@
...
+For example, clusters like `kuma:envoy:admin` or `_kuma:dynamicconfig` are internal to Kuma, while `kri_msvc_mesh-1_us-east-2_kuma-demo_backend_httpport` cluster is an external resource.
+
+Currently, there is inconsistency in how these resource names are formed, some of them have a `_kuma` prefix, some of them just contain `kuma` and some are free form.
+
+Some internal resources do not map clearly to Kuma resources in the data store, making it difficult to relate metrics to Kuma abstractions.
+
+There are also resources that are in a bit of a gray area, like `meshtrace_datadog` which is a cluster created for `MeshTrace` policy.
+It seems internal, but it could be traced down to MeshTrace resource `kri_meshtrace_mesh-1__kuma-system_my-meshtrace_`.
+
+#### Examples of kuma resource names
```
lobkovilya: It looks like a full list of internal resources, or some internal resources are missing? If it's full list then I'd replace the word "Examples" and call it "Internal Kuma resources". It makes it sound like a list you can work on

docs/madr/decisions/074-naming-internal-envoy-resources.md
lahabana: Maybe is calling these internal is the confusing bit? Maybe we should call these "system" envoy resources.

Also you should really reference: https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/036-internal-listeners.md it might be worth just saying that this MADR superseed 036 as it goes in more details

lahabana:
The correct PR title for a MADR is `docs(MADR)`

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,152 @@
...
+- listeners
+  - ip_port
+  - virtualOutbound
+  - virtualInbound
+
+## Use cases
+
+### Looking at a resource name I want to easily relate it to Kuma concepts
+
+### As a user I want my MeshProxyPatch to continue working after the change
```
lahabana: I think it's fine isn't it part of the whole reason (like for observability) to have this behind a feature flag on the DP?

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,152 @@
...
+sum:envoy.cluster.upstream_rq.count{!envoy_cluster:kuma_readiness , !envoy_cluster:access_log_sink , !envoy_cluster:meshtrace_datadog}.as_count()
+```
+
+we can do:
+
+``` 
+sum:envoy.cluster.upstream_rq.count{!envoy_cluster:_kuma*}.as_count()
+```
+
+**Is this a valid use case if we have profiles in MeshMetric?**
```
lahabana: I think this is the most useful usecase! The idea is that a user should be able to very easily figure what's actually useful to him vs random user.

Even with profiles, it's a lot easier to write a profile if we follow a system

docs/madr/decisions/074-naming-internal-envoy-resources.md
lahabana: Yes having these just makes profiles easier to implement and maintain :)

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,152 @@
...
+
+It's important to know the difference between internal resources (the ones that are defined by Kuma itself, and they usually do not concern the end users) and external resources (the ones that are defined indirectly by the user and the users are interested in).
+For example, clusters like `kuma:envoy:admin` or `_kuma:dynamicconfig` are internal to Kuma, while `kri_msvc_mesh-1_us-east-2_kuma-demo_backend_httpport` cluster is an external resource.
+
+Currently, there is inconsistency in how these resource names are formed, some of them have a `_kuma` prefix, some of them just contain `kuma` and some are free form.
+
+Some internal resources do not map clearly to Kuma resources in the data store, making it difficult to relate metrics to Kuma abstractions.
+
+There are also resources that are in a bit of a gray area, like `meshtrace_datadog` which is a cluster created for `MeshTrace` policy.
+It seems internal, but it could be traced down to MeshTrace resource `kri_meshtrace_mesh-1__kuma-system_my-meshtrace_`.
```
lahabana: I think what is "internal" or "system" is 100% part of the problem statement and should be defined here.

IMO a "system" envoy resource is something that a user is unlikely to want to observe/modify. In other words things that are not really a result of the environment so things not coming from `Mesh*Service` `Mesh*Route`.

So in the example above this is 100% a "system" resource. However, this `_kuma:` I believe doesn't block us from having a kri after. So it could be: `_kuma:kri_meshtrace_mesh-1__kuma-system_my-meshtrace_` no?

PS: I don't if this just an example or comes from code but this is an invalid KRI according to the [MADR](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/070-resource-identifier.md#option-1---order-based-no-field-names-in-the-identifier) the short name for the type should be used so it should be [`mtr`](https://github.com/kumahq/kuma/blob/a7cc1f47184f9b2dee024d283d3cf255fde3ef26/pkg/plugins/policies/meshtrace/api/v1alpha1/zz_generated.resource.go#L175).

docs/madr/decisions/074-naming-internal-envoy-resources.md
slonka: I did not find more but you know how it is - there could be something dynamically defined that is triggered by some specific setting.

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,162 @@
...
+Currently, there is inconsistency in how these resource names are formed, some of them have a `_kuma` prefix, some of them just contain `kuma` and some are free form.
+
+Some internal resources do not map clearly to Kuma resources in the data store, making it difficult to relate metrics to Kuma abstractions.
+
+There are also resources that seem to be in a bit of a gray area, like `meshtrace_datadog` which is a cluster created for `MeshTrace` policy.
+It could be traced down to MeshTrace resource `kri_mtr_mesh-1__kuma-system_my-meshtrace_` and if a cluster operator has problems with tracing collection it might be something to look at.
+
+To make it easier to distinguish between the two types we introduce the following definition:
+
+- **User resource** - any resource that comes from `Mesh*Service` and `Mesh*Route` definition
```
lukidzi: I think in the future we might also have Secrets: since we should probably name it based on MeshIdentity resource. In case of Spire resource name is SPIFFEID, spiffe id with trust domain or `ALL` (federation)

slonka: I think it's the same thing as a cluster for `MeshTrace` - what do others think?

lahabana: Actually with SPIRE will you have the freedom on these resource names?

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,194 @@
...
+- kuma:metrics:hijacker
+- kuma:readiness
+- meshtrace_[zipkin|datadog|otel]
+- envoy_grpc_streams (ads)
+
+Virtual Hosts:
+- _kuma:dynamicconfig
+- kuma:envoy:admin
+
+Routes (not all of them even have a name):
```
lahabana: ideally all our resources should have a name right?

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,194 @@
...
+  - xds-grpc
+  - agent
+- listeners
+  - ip_port
+  - virtualOutbound
+  - virtualInbound
+
+## Use cases
+
+### Looking at a resource name I want to easily relate it to Kuma concepts
```
lahabana: and understand where it's coming from

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,194 @@
...
+
+we can do:
+
+``` 
+sum:envoy.cluster.upstream_rq.count{!envoy_cluster:_kuma*}.as_count()
+```
+
+You might be tempted to think that this use case is already covered by the `MeshMetric` profiles, 
+and it partially is (the end result is the same) but this use case compliments it by making the profiles easier to implement and maintain.
+It also allows users to do their own filtering easier.
```
lahabana:
```suggestion
It also allows users to do their own filtering more easily.
```

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,194 @@
...
+
+Which indicates a system resource that was created for a `MeshTrace` resource.
+
+Resources that are not correlated with any Kuma resource like:
+
+```
+kuma:envoy:admin
+```
+
+Will have a prefix `_kuma_` and adhere to the general format rules (not using `:`):
```
lahabana: Can we be strict about the formatting? Have this as a regex here (and potentially we could have a test to validate these).

Also we should make explicit that using `_kuma_kri` as a prefix implies that it's a valid KRI after.
Should we maybe disallow `_` after the prefix if we are not using a kri?

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,194 @@
...
+          - key: envoy_cluster
+            value: "^_kuma_.*"  # matches if envoy_cluster starts with _
+```
+
+### As a Kuma developer I want to have a consistent naming scheme for all resources in Envoy
+
+### As a Kuma developer I want to distinguish between system and user resources in Envoy
+
+### As a Kuma developer I want to distinguish between system resources exposed and not exposed in metrics
+
```
lahabana: One requirement which happened on the other MADR is that the cardinality of these resource names should be low.

.i.e:

- Avoid IPs / non constant bits
- Avoid dataplane name.

lahabana (approved):
We should mention that this will only happen when the flag defined in https://github.com/kumahq/kuma/pull/13485 is enabled.

docs/madr/decisions/074-naming-internal-envoy-resources.md
slonka: I'll add a use-case for that

docs/madr/decisions/074-naming-internal-envoy-resources.md
slonka: > Should we maybe disallow _ after the prefix if we are not using a kri?

Are there any benefits to this approach? Isn't this just going to be different for the sake of different? Are you afraid of something mixing up with `kri`? If so, it shouldn't - these are defined by us so it's not like a user could "spoof" one.

docs/madr/decisions/074-naming-internal-envoy-resources.md
lukidzi: In case of spire we cannot rename them, usually they have `spiffeID` for identity, and `spiffe://trustDomain`/`ROOTCA`/ `ALL` for CA or bundle

docs/madr/decisions/074-naming-internal-envoy-resources.md
slonka: Added

docs/madr/decisions/074-naming-internal-envoy-resources.md
slonka: I don't think we can control in any way the xds resources from Spire Agent (maybe if we added a custom https://github.com/envoyproxy/xds-relay but that's just too big of an effort - right?). See:
- https://github.com/spiffe/spire/blob/f687bf21e812a4bf027d88136032fc46497e7fe0/pkg/agent/endpoints/sdsv3/handler.go#L395
- https://github.com/spiffe/spire/blob/f6a11a0ae03353b3444007ecaa8c81b34931bb49/pkg/agent/endpoints/sdsv3/handler.go#L282
- https://github.com/spiffe/spire/blob/f6a11a0ae03353b3444007ecaa8c81b34931bb49/pkg/agent/endpoints/sdsv3/handler.go#L291
- https://github.com/spiffe/spire/blob/f6a11a0ae03353b3444007ecaa8c81b34931bb49/pkg/agent/endpoints/sdsv3/handler.go#L312

---

Going back to original question - will secrets in the future be a `User` resource?

docs/madr/decisions/074-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,222 @@
...
+See:
+- https://github.com/spiffe/spire/blob/f687bf21e812a4bf027d88136032fc46497e7fe0/pkg/agent/endpoints/sdsv3/handler.go#L395
+- https://github.com/spiffe/spire/blob/f6a11a0ae03353b3444007ecaa8c81b34931bb49/pkg/agent/endpoints/sdsv3/handler.go#L282
+- https://github.com/spiffe/spire/blob/f6a11a0ae03353b3444007ecaa8c81b34931bb49/pkg/agent/endpoints/sdsv3/handler.go#L291
+- https://github.com/spiffe/spire/blob/f6a11a0ae03353b3444007ecaa8c81b34931bb49/pkg/agent/endpoints/sdsv3/handler.go#L312
+
+## Design
+
+### Use a `^system_[a-z0-9_]+$` regex to name system resources
```
slonka: it's a group of numbers / letters and `_` is outside

docs/madr/decisions/074-naming-internal-envoy-resources.md
slonka: 1. SPIRE = no choice (spiffe ID / trust Domain)
2. Not SPIRE
   1. we can't do KRI because it's a combo of multiple all the MeshTrusts
   2. For MeshIdentity we could use KRI (future)
   3. For current `system_<category>_<scope>_<identifier>
   4. For user defined secrets we can use KRI

docs/madr/decisions/074-naming-internal-envoy-resources.md
slonka: The core thing is having a way of enforcing these rules. Section name can only be a valid DNS name, it applies to all section names in all formats.

docs/madr/decisions/074-naming-internal-envoy-resources.md
slonka: added that to the contents

docs/madr/decisions/074-naming-internal-envoy-resources.md
slonka: Added a simple way of enforcing these rules by checks in the builders

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,239 @@
...
+So instead 
+
+``` 
+sum:envoy.cluster.upstream_rq.count{!envoy_cluster:kuma_readiness , !envoy_cluster:access_log_sink , !envoy_cluster:meshtrace_datadog}.as_count()
+```
+
+we can do:
+
+``` 
+sum:envoy.cluster.upstream_rq.count{!envoy_cluster:_kuma*}.as_count()
```
lahabana: nedds updating

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,262 @@
...
+
+### As a Kuma developer I want to distinguish between system and user resources in Envoy
+
+### As a Kuma developer I want to distinguish between system resources exposed and not exposed in metrics
+
+### As a user I want the cardinality of these resource names to be low
+
+Need to avoid:
+- IPs / non constant bits
+- Avoid dataplane name.
```
lahabana:
```suggestion
- Avoid dataplane name or kri.
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,262 @@
...
+
+See:
+- https://github.com/spiffe/spire/blob/f687bf21e812a4bf027d88136032fc46497e7fe0/pkg/agent/endpoints/sdsv3/handler.go#L395
+- https://github.com/spiffe/spire/blob/f6a11a0ae03353b3444007ecaa8c81b34931bb49/pkg/agent/endpoints/sdsv3/handler.go#L282
+- https://github.com/spiffe/spire/blob/f6a11a0ae03353b3444007ecaa8c81b34931bb49/pkg/agent/endpoints/sdsv3/handler.go#L291
+- https://github.com/spiffe/spire/blob/f6a11a0ae03353b3444007ecaa8c81b34931bb49/pkg/agent/endpoints/sdsv3/handler.go#L312
+
+## Design
+
+### Use a `^system_([a-z0-9]+_{0,1})+$` regex to name system resources
```
lahabana:
```suggestion
### Use a `^system_([a-z0-9-]*_?)+$` regex to name system resources
```

I've created a regex101 to add some options: https://regex101.com/r/2C8IPM/1
Where's the KRI regex for this? Because clearly the regex above wasn't validated with the suggested usage of this.

I'd stay away from `{0,1}` as it isn't supported by all regex engine (also `?` is equivalent).
I think to have trailing kri to work we need to tolerate `__`

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,262 @@
...
+
+All changes will be behind the same feature flag as in [Migrating to consistent and well-defined naming for non-system Envoy resources and stats](./077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md).
+
+All system resources will conform to `^system_([a-z0-9]+_{0,1})+$` regex (we shouldn't use `_system:` because of [this issue](https://github.com/kumahq/kuma/issues/2363)).
```
lahabana:
```suggestion
All system resources will conform to `^system_([a-z0-9]+_{0,1})+$` regex (we shouldn't use `system:` because of [this issue with `:` as separator](https://github.com/kumahq/kuma/issues/2363)).
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,262 @@
...
+
+System resources that can be traced back to a Kuma resource with a valid KRI will have take the form of `system_<KRI>`.
```
lahabana:
```suggestion
System resources that can be traced back to a Kuma resource with a valid KRI have the format `system_<KRI>`.
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,262 @@
...
+
+In the future we will most likely introduce new policies - `MeshTrust`, `MeshIdentity` and `SPIRE` integration.
+
+For `Spire` we can't control the secrets as mentioned in [Non use cases section](#as-a-kuma-developer-i-want-to-rename-resources-that-are-coming-from-outside-kuma).
+
+For a secret related to `MeshTrust`s we can't use `KRI` because it's a combination of all of them and `KRI` currently don't deal with lists of resources.
+
+For `MeshIdentity` and user defined secrets we can use `KRI` but that will be covered by [Migrating to consistent and well-defined naming for non-system Envoy resources and stats](./077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md).
+
+For all other secrets we can use `^system_([a-z0-9]+_{0,1})+$` naming.
```
lahabana: Isn't all this going in so much details that it confuses the reader.

It seems the rule here is:

> Secrets are resources like any others but should most likely always be considered system. Therefore, they will follow if they are an actual Kuma secret `^system_KRI` and `^system_<someusefulName>` otherwise.
There are possible exceptions but these should be explicitly called out in MADRs. (.e.g [MeshTrust and MeshIdentity modify link](linkToSpiffe)).

docs/madr/decisions/076-naming-internal-envoy-resources.md
lahabana: The titles of this section are all over the place.

Isn't it:

```
## Design

### Resource naming

#### 1. Regular resources

Resources that come from users should be named after their KRI.

#### 2. System resources that can be tracked back to user resource

When possible use `system_<KRI>` it should be justified when designing why a user resource is system and not just `<kri>`.

There are other cases where a resource is the result of merging multiple resources together (MTP rbac filters, MeshPassthrough...). In this case we should do: TBD

#### 3. System resources that are not related to user resource

When adding such resource the name should be explicit enough to help anyone understand where this is coming from.

If it's coming from a plugin or a resource for example: `system_mal_dpenvoy_pipe` or `system_mp_someinfo`.

If it's fulfilling a shared or system section it should be clear as well. Some examples: `system_dynconfig` or `system_passthrough_inbound`

#### 4. Resource contextual to the dataplane

There's a few resources that only make sense in the context of the current dataplane.
These are:

1. Resources that would be too high cardinality because the KRI includes the DP name
# This is my own creation but it should be covered in this MADR.
3. Resources that are very DP specific and still worth observing for .e.g:  inbound_passthrough/outbound_passthrough. 

In this case we should use: `self_<shortType>_<usefullName(potentially sectionName)>`

So the examples above could be:

`self_dp_<portName>` or `self_system_inbound_passthrough`

#### 4. Secrets

Nothing special for secrets they either adhere to 1 or 2.

#### 5. Exceptions

We can't control the name of external resources (.e.g SPIRE SDS secrets).
Any new exception and case should result in a MADR which can be linked from here.

### Currently existing system resources and their new naming

... All the changes that will happen as a result of this.

### Enforcement

...

## Rejected alternatives

Using `:` and other things.
```

lahabana: It's ok to have WIPs and jot things progressively. But at some point it's important to rework a MADR in something that is concise and goes straight to the point.

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,262 @@
...
+
+#### Enforcement
+
+##### Builders
+
+To enforce these rules we will add checks in resource builders like [ClusterBuilder](https://github.com/kumahq/kuma/blob/dedaba5b9de1bd134dce813ae49b3475d5d24e6b/pkg/xds/envoy/clusters/cluster_builder.go#L80).
+If anyone tries to add a new resource that doesn't conform to any type it will fail in unit tests.
```
lahabana:
```suggestion
If anyone tries to add a new resource that doesn't conform to any type it will fail tests.
```

doesn't have to be unit tests

lahabana (commented):
This doc needs some rework. It repeats itself, goes left and right.

I get the point now but I'm not sure I'll understand it still in 6months.

This was useful as a draft but it now needs to be reconsidered.

docs/madr/decisions/076-naming-internal-envoy-resources.md
slonka: updated

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,256 @@
...
+kuma:envoy:admin
+```
+
+Will conform with the previously mentioned regex:
+
+```
+system_envoy_admin
+```
+
+and should describe the resource as accurately and plainly as possible (it **MUST** take into account any related configuration option or annotation if exists).
```
lahabana:
```suggestion
and should describe the resource as accurately and plainly as possible (it **MUSTN'T** take into account any related configuration option or annotation if exists).
```

no?

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,256 @@
...
+
+It means that `listener.0.0.0.0_15001` which can be configured by (`kuma.io/transparent-proxying-outbound-port`) will become `system_transparent_proxy_outbound_listener`.
+
```
lahabana: IMO this MADR is still a good place to define `self_` WDYT?

docs/madr/decisions/076-naming-internal-envoy-resources.md
slonka: If it `mustn't` then `listener.0.0.0.0_15001` can't have anything related to `transparent-proxying-outbound` in the name? Why? To someone not familiar with kuma does `15001` mean anything? I doubt that. We have configuration switches that better describe the meaning than the resource names - why not take it into account and have something similar?

docs/madr/decisions/076-naming-internal-envoy-resources.md
slonka: `self_` is defined here - https://github.com/kumahq/kuma/pull/13485/files#diff-15d3fac527452e5002d13b52e9b646b84ed3df20db13fcbd2ed0ed3af3bc6355R29. You mentioned [here](https://github.com/kumahq/kuma/pull/13783#discussion_r2161454561) - "In other words things that are not really a result of the environment so things not coming from Mesh*Service Mesh*Route". Pods that are then converted to DPs are definitely coming from the environment.

docs/madr/decisions/076-naming-internal-envoy-resources.md
slonka: > #### 1. Regular resources

I've renamed this to `User resources` since this is the name we agreed to use.

> There are other cases where a resource is the result of merging multiple resources together (MTP RBAC filters, MeshPassthrough...). 

I’m not sure, but I think RBAC policies which are the result of MTP Policies configured on the listener have a policy name field which can point to the exact KRI of the policy. MeshPassthrough is out of scope because we want to observe it[ https://docs.google.com/document/d/1meJne4SA691P0iucEs7cFaqQjZgvejIx7wCHa-556tg/edit?disco=AAABm3p6H58](https://docs.google.com/document/d/1meJne4SA691P0iucEs7cFaqQjZgvejIx7wCHa-556tg/edit?disco=AAABm3p6H58) so it won't be a system resource.

> In this case we should do: TBD

Can we do anything else than just try and name them something descriptive?

> `system_passthrough_inbound`

So, is passthrough system or not? https://docs.google.com/document/d/1meJne4SA691P0iucEs7cFaqQjZgvejIx7wCHa-556tg/edit?tab=t.0#heading=h.5nlzwysvbtao 

> Resource contextual to the dataplane

It’s defined here - https://github.com/kumahq/kuma/pull/13485/files#diff-15d3fac527452e5002d13b52e9b646b84ed3df20db13fcbd2ed0ed3af3bc6355R29 . “In other words things that are not really a result of the environment” - is dataplane not a result of the environment?

> Nothing special for secrets they either adhere to 1 or 2.

Unless it’s coming from spire then it’s neither 1 or 2.

lahabana: @lobkovilya can weigh in but I'm pretty sure that most policies can't be tracked down to a single policy as they are the merging of potentially many policies (just like mesh passthrough)

For passthrough I think we need to separate meshpassthrough and passthrough.
My guess is that the latter should likely be on self_ and not like system (yes I was wrong above)

docs/madr/decisions/076-naming-internal-envoy-resources.md
lahabana: Ok I get it it's because you talk about annotation. To me this is annotation value. How about:

```suggestion
and should describe the resource as accurately and plainly as possible (it **MUST** take into account any related configuration option or annotation name if exists).
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+### As a Kuma developer I want to distinguish between system and user resources in Envoy
+
+### As a Kuma developer I want to distinguish between system resources exposed and not exposed in metrics
+
+### As a user I want the cardinality of these resource names to be low
+
+Need to avoid:
+- IPs / non constant bits
+- Avoid dataplane name or kri.
+
```
bartsmykla: Nit:
That’s not a problem for system resources since we control their names. Right now, this only affects resources generated from the Dataplane, which is already covered by [MADR-077](https://github.com/kumahq/kuma/blob/52b8f5548739ecd1124181541f105bd6481b7ba5/docs/madr/decisions/077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md), so I wouldn’t mention it here.

```suggestion
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+## Non use cases
```
bartsmykla: Nit:
```suggestion
## Unsupported use cases
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Resource naming
+
+#### 1. User resources
+
+Names for these resources are defined in [Defining and migrating to consistent naming for non-system Envoy resources and stats](./077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md).
+
+#### 2. System resources that can be tracked back to user resource
+
+When possible, use `system_<KRI>`. It should be justified when designing why a user resource is system and not just `<kri>`.
```
bartsmykla: Nit:
```suggestion
When possible, use `system_<kri>`. It should be justified when designing why a user resource is system and not just `<kri>`.
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+sum:envoy.cluster.upstream_rq.count{!envoy_cluster:system_*}.as_count()
+```
+
+You might be tempted to think that this use case is already covered by the `MeshMetric` profiles, 
+and it partially is (the end result is the same) but this use case compliments it by making the profiles easier to implement and maintain.
+It also allows users to do their own filtering more easily.
+
+### As an operator I want to drop all stats related to internal resources
+
+So I can use a processor like:
```
Automaat: what is this processor?

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+### Resource naming
+
+#### 1. User resources
+
+Names for these resources are defined in [Defining and migrating to consistent naming for non-system Envoy resources and stats](./077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md).
+
+#### 2. System resources that can be tracked back to user resource
+
+When possible, use `system_<KRI>`. It should be justified when designing why a user resource is system and not just `<kri>`.
+For example, when a resource is not on a service to service traffic path, is shared between other parts of the system (like Kong Mesh's `MeshGlobalRateLimit` service) or we want to make it easy to exclude from metrics collection.
```
bartsmykla: Nit: Wdyt about writing it like this?
```suggestion
For example, when a resource isn't part of regular service-to-service traffic path, is shared across the system (like Kong Mesh's `MeshGlobalRateLimit` service), or simply needs to be easy to exclude from metrics collection.
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+
+There are other cases where a resource is the result of merging multiple resources together (MTP rbac filters, MeshPassthrough, etc.).
```
bartsmykla: Nit:
```suggestion
There are other cases where a resource is the result of merging multiple resources together (MeshTrafficPermission RBAC filters, MeshPassthrough, etc.).
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+There is nothing special about secrets, they adhere to the same rules (1-4) and can have exceptions (6).
+
+#### 6. Exceptions
+
+We can't control the name of external resources (i.e. SPIRE SDS secrets).
+Any new exception and case should result in a MADR which can be linked from here.
+
+#### 7. Regex to match system names
+
+For `system_<KRI>` resources we will use the same regex as used for KRI but prefixed with `system_`.
```
bartsmykla: Nit:
```suggestion
For `system_<kri>` resources we will use the same regex as used for KRI but prefixed with `system_`.
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
slonka: OTEL processor - https://opentelemetry.io/docs/collector/configuration/#processors a thing that processes metrics

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+| Cluster       | ads_cluster                                                  | system_ads                      |                                                              |
+| Cluster       | plugins:bootstrap:k8s:hooks:apiServerBypass                  | system_kube_api_server_bypass   |                                                              |
+| Listener      | plugins:bootstrap:k8s:hooks:apiServerBypass                  | system_kube_api_server_bypass   |                                                              |
+| Listener      | kuma:dns                                                     | system_dns_builtin              |                                                              |
+
+Deprecated features which resource names won't be changed:
+- virtual probes:
+  - probe:listener
+  - probe:route_configuration
+  - probe
```
bartsmykla: Nit: Wdyt?
```suggestion
Deprecated features whose resource names will remain unchanged:

- [Virtual Probes](https://kuma.io/docs/2.11.x/policies/service-health-probes/#virtual-probes):
  - `probe:listener`
  - `probe:route_configuration`
  - `probe`
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+- Might be really hard to implement
+
+#### Chosen option
+
+We're choosing "Inspecting golden files" as it's the most feasible option that can be implemented quickly and will catch most of the cases.
+
+## Rejected alternatives
+
+- [Using `:`](https://github.com/kumahq/kuma/issues/2363)
+
```
bartsmykla: Nit: I wouldn't mention it as it's not clear enough to understand and doesn't bring value to this MADR
```suggestion
```

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,315 @@
...
+## Implications for Kong Mesh
+
+Kong Mesh doesn't use any special type of resource that is not used by Kuma. 
+In its policies, it only defines `meshglobalratelimit:service` cluster for MeshGlobalRateLimit
+(as well as an endpoint, modifications to routes, virtual hosts and filter chains).
+
+For MeshOPA it only modifies existing resources and doesn't create any new ones.
```
bartsmykla: Nit: Wdyt?
```suggestion
The changes introduced by this document impact the Envoy resources generated by the `MeshGlobalRateLimit` policy. This policy currently adds a cluster with a static name `meshglobalratelimit:service`, which is then referenced in route-level [rate limit configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-rate-limits). As a result of these changes, the cluster name will be updated to follow the `system_<kri>` format, using the valid `kri` of the `MeshGlobalRateLimit` policy that is the source of this cluster (i.e. `system_kri_mgrl___kong-mesh-system_global-rate-limit-policy_`)

The `MeshOPA` policy modifies existing non-system inbound listeners but does not rename any existing Envoy resources or create new ones. Therefore, it is not affected by the changes described in this document.

Other Kong Mesh-specific features or policies do not rely on or modify Envoy resource names or stat prefixes that fall within the scope of this document.
```

bartsmykla (approved):
LGTM with only few nits

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,319 @@
+# Standardized Naming for internal xDS Resources
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/13266
+
+Supersedes: https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/036-internal-listeners.md
```
lahabana: Can we add a `Superseded by` in 36 (and make it very bold)?

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,319 @@
...
+The name should be namespaced so that it groups similar resources together and avoids collisions with other system resources.
+
+For example dynamic config fetcher creates one listener and one route per object type.
+We will namespace it as `system_dynamicconfig` and the DNS route will be `system_dynamicconfig_dns` and MeshMetric route will be `system_dynamicconfig_meshmetric`.
+
+If any related configuration option or annotation name exists it should be taken into account.
+
+#### 4. Resource contextual to the Dataplane
+
+These resources are defined in [Defining and migrating to consistent naming for non-system Envoy resources and stats](./077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md).
```
lobkovilya: Could you please add here these resources have `self_` pattern? It'd be nice to have all resource categories and their prefixes in one place and this MADR is the best candidate as it defines all the categories

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,319 @@
...
+
+For `system_<kri>` resources we will use the same regex as used for KRI but prefixed with `system_`.
+
+For other names we will use `^system_([a-z0-9-]*_?)+$`, here is an example of the usage: https://regex101.com/r/Ic1bk5/2.
+
+### Currently existing system resources and their new naming
+
+| Resource type | Current Name                                                 | New Name                        | Example                                                      |
+|---------------|--------------------------------------------------------------|---------------------------------|--------------------------------------------------------------|
+| Cluster       | kuma:envoy:admin                                             | system_envoy_admin              |                                                              |
```
lobkovilya: do we need `_envoy_`? maybe just `system_admin`?

docs/madr/decisions/076-naming-internal-envoy-resources.md
slonka: @lahabana suggested to have another document that summarizes all the categories in one place.

docs/madr/decisions/076-naming-internal-envoy-resources.md
```diff
@@ -0,0 +1,319 @@
+# Standardized Naming for internal xDS Resources
```
lahabana: we use internal in the MADR name

docs/madr/decisions/076-naming-internal-envoy-resources.md
bartsmykla: Using `envoy` as a namespace gives us room to expand in the future for anything closely tied to Envoy. In contrast, `system_admin` takes up the broader `admin` namespace, which limits our options if we want to group other kinds of admin-related things later.

Also, `admin` on its own is too vague. It's not clear what kind of admin it refers to, so it's harder to understand at a glance. `envoy` is much more specific and immediately tells you what it's about.

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,319 @@
...
+  - PassthroughCluster / InboundPassthroughCluster
+  - BlackHoleCluster
+  - prometheus_stats
+  - sds-grpc
+  - xds-grpc
+  - agent
+- listeners
+  - ip_port
+  - virtualOutbound
+  - virtualInbound
```
slonka: put in notes

docs/madr/decisions/076-naming-internal-envoy-resources.md
Automaat: co we do then: 
```
So I can use a otel processor like:
```

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,319 @@
...
+| Route         | _kuma:dynamicconfig:dns                                      | system_dynamicconfig_dns        |                                                              | 
+| Route         | _kuma:dynamicconfig:meshmetric                               | system_dynamicconfig_meshmetric |                                                              | 
+| Listener      | _kuma:metrics:prometheus:backend-default                     | system_<kri>                    | system_kri_mm_mesh-1_us-east-2_kuma-demo_default_            | 
+| Listener      | _kuma:metrics:prometheus:<backendName>                       | system_<kri>                    | system_kri_mm_mesh-1_us-east-2_kuma-demo_default_            | 
+| Listener      | _kuma:metrics:opentelemetry:<backendName>                    | system_<kri>                    | system_kri_mm_mesh-1_us-east-2_kuma-demo_default_            | 
+| Cluster       | _kuma:metrics:opentelemetry:<backendName>                    | system_<kri>                    | system_kri_mm_mesh-1_us-east-2_kuma-demo_default_            |
```
Automaat: backend name will be part of KRI?

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,314 @@
...
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/13266
+
+Supersedes: https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/036-internal-listeners.md
+
+## Context and Problem Statement
+
+The point of this document is to define a naming strategy of Envoy resources and guidelines for exceptions.
```
lahabana:
```suggestion
The point of this document is to define a naming strategy of Envoy resources, existing exceptions and guidelines for future exceptions.
```

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,314 @@
...
+- System resources - these are internal to Kuma and users are unlikely to be interested in them except when debugging Kuma.
+- User resources - the rest of the resources which are most commonly derived from a Kuma resource and will likely be monitored by users.
+
+For example a cluster named `kuma:envoy:admin` is a system resource, while a cluster `kri_msvc_mesh-1_us-east-2_kuma-demo_backend_httpport` is a user resource.
+
+From now on when we use "system resource" or "user resource" we're talking about Envoy resources, not Kuma resources.
+
+Currently, there is no consistency in how these resources are named, see [additional context](#additional-context) for detailed list.
+
+
```
lahabana:
```suggestion
```

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,314 @@
...
+
```
lahabana:
```suggestion
```

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,314 @@
...
+        attributes:
+          - key: envoy_cluster
+            value: "^system_.*"  # matches if envoy_cluster starts with _
+```
+
+A side benefit of this is that profiles implementation is a lot simpler.
+
+### 5. As a Kuma developer I want to have a consistent naming scheme for all resources in Envoy
+
+### 6. As a Kuma developer I want to distinguish between system and user resources in Envoy
```
lahabana: This could be number 1

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,314 @@
...
+        match_type: regexp
+        metric_names: [".*"]  # apply to all metrics
+        attributes:
+          - key: envoy_cluster
+            value: "^system_.*"  # matches if envoy_cluster starts with _
+```
+
+A side benefit of this is that profiles implementation is a lot simpler.
+
+### 5. As a Kuma developer I want to have a consistent naming scheme for all resources in Envoy
```
lahabana: This could be 2

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,314 @@
...
+
+These should be exceptional cases when we consider such resource as system and not user resource.
+Such exceptions should be defined in a MADR and must use `system_<kri>` naming scheme.
+
+For example the cluster for `MeshGlobalRateLimit` service is a system resource because it is not part of the regular service-to-service traffic path,
+but it is still related to a Kuma resource (the `MeshGlobalRateLimit` policy).
+The cluster is currently called `meshglobalratelimit:service`, but it will be renamed to `system_kri_mgrl___kong-mesh-system_global-rate-limit-policy_`.
+
+There are other cases where a resource is the result of merging multiple resources together (MeshTrafficPermission RBAC filters, MeshPassthrough, etc.).
+This is out of scope and requires a separate MADR (todo: add link)
```
lahabana: todo missing

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,314 @@
...
+
+Refer to [currently existing system resources and their new naming](#currently-existing-system-resources-and-their-new-naming) for the full list.
+
+#### 2. System resources that are not related to Kuma resource
+
+When adding such resource the name should be explicit enough to help anyone understand where this is coming from.
+The name should be namespaced so that it groups similar resources together and avoids collisions with other system resources.
+
+For example dynamic config fetcher creates one listener and one route per object type.
```
lahabana:
```suggestion
For example dynamic config creates one listener and one route per object type.
```

docs/madr/decisions/076-naming-system-envoy-resources.md
```diff
@@ -0,0 +1,314 @@
...
+We will namespace it as `system_dynamicconfig` and the DNS route will be `system_dynamicconfig_dns` and MeshMetric route will be `system_dynamicconfig_meshmetric`.
+
+If the name is coming from a reusable component the namespacing is the responsibility of the reusable component.
```
lahabana:
```suggestion
If the name is coming from a reusable component like it is now the namespacing is the responsibility of the reusable component, not the user of it.
```

lahabana (approved):
Some nits but it's good to go IMO

docs/madr/decisions/076-naming-system-envoy-resources.md
slonka: this "backend name" is the name of the variable, what's actually in there is the address:

```
backends:
  - type: OpenTelemetry
    openTelemetry: 
      endpoint: otel-collector.observability.svc:4317
      refreshInterval: 60s
```

and in theory we can have 2 backends with the same endpoint and different refresh interval so it wouldn't be unique. So that's why I did not put it there. WDYT?

docs/madr/decisions/076-naming-system-envoy-resources.md
slonka: removed

docs/madr/decisions/076-naming-system-envoy-resources.md
slonka: added

docs/madr/decisions/076-naming-system-envoy-resources.md
slonka: removed

docs/madr/decisions/076-naming-system-envoy-resources.md
bartsmykla: As defined in the `sectionName` format:

> The `sectionName` identifies a specific unit within the current context (in this document, the `Dataplane`) that can be referred to by Kuma policies. Currently, this applies to ports (such as outbounds and non-system inbounds), but the format is designed to support future extensions, such as DNS-like labels for `MeshPassthrough`.
> ref. https://github.com/kumahq/kuma/blob/dc2dd5564cac3f97f85085a8aa498f4a520ad4ba/docs/madr/decisions/077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md?plain=1#L125

When referencing a specific part of a resource, the appropriate `sectionName` must be used. In this case, the backend identifier should be placed in the `sectionName` field. For example, the resulting name would be:

```
system_kri_mm_mesh-1_us-east-2_kuma-demo_default_otel_otel-collector.observability.svc_4317
```

We probably need a separate MADR to update the `sectionName` definition with all its use cases and clear rules. One open question is what to do when the `sectionName` needs to include several fields to tell different sections apart (like backends in this case), which could make the `sectionName` very long. We need to decide if that's okay or if there's a better way to handle it so I opened: https://github.com/kumahq/kuma/issues/13887

slonka: I've added section name

docs/madr/decisions/076-naming-internal-envoy-resources.md
slonka: https://github.com/kumahq/kuma/issues/13889 - added

bartsmykla (approved):
LGTM

lobkovilya (approved):

lahabana (approved):

PR #13786, opened 2025-06-23
feat(meshcontext): move MeshServices, MeshExternalService and MeshMultizoneServices to new DestinationIndex inside BaseMeshContext

Automaat:
## Motivation
For implementing new Inspect API and Dataplane Layout endpoint, we need to compute list of outbounds. We have all the needed code in place, but it was in `MeshContext`. We want to limit the scope of things that need to be built for this endpoint. To achieve this, we can move this information from `MeshContext` to `BaseMeshContext` as building it is lightweight.


pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -105,7 +105,7 @@ func GenerateClusters(
 								// services that are synced can be accessed only with TLS through ZoneIngress
 								tlsReady = !ms.IsLocalMeshService() || ms.Status.TLS.Status == meshservice_api.TLSReady
 								if port, found := ms.FindPortByName(realResourceRef.Resource.SectionName); found {
-									protocol = port.AppProtocol
+									protocol = core_mesh.ParseProtocol(port.GetProtocol())
```
lahabana: why this here and not in the helpers?

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/helpers.go
```diff
@@ -59,3 +64,43 @@ func (t *MeshExternalServiceResource) Domains() *xds_types.VIPDomains {
...
+func (m MesPort) GetName() string {
+	return fmt.Sprintf("%d", m.Value)
+}
+
+func (m MesPort) GetValue() uint32 {
+	return uint32(m.Value)
+}
+
+func (m MesPort) GetNameOrStringifyPort() string {
+	return fmt.Sprintf("%d", m.Value)
```
lahabana: The implem here doesn't seem right no?

Also I think the API itself is broken because GetName should do this because that's the first thing folks will think about calling no?

pkg/core/resources/apis/core/interfaces.go
```diff
@@ -4,4 +4,5 @@ type Port interface {
 	GetName() string
 	GetValue() uint32
 	GetNameOrStringifyPort() string
+	GetProtocol() string
```
lahabana: I know we're very bad at documenting interfaces but can we try to improve?

for example: this GetProtocol does it returned a sanitized version of the protocol?

pkg/xds/context/destination_index.go
```diff
@@ -0,0 +1,150 @@
...
+	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
+	"github.com/kumahq/kuma/pkg/plugins/policies/core/rules/resolve"
+)
+
+type Destination interface {
+	GetPorts() []core.Port
+	FindPortByName(name string) (core.Port, bool)
+}
+
+type DestinationIndex struct {
```
lahabana: Same what's a DestinationIndex? When should I use it? Does it have constraints?

pkg/xds/context/destination_index.go
```diff
@@ -0,0 +1,150 @@
...
+	MeshServiceByIdentifier             map[kri.Identifier]*meshservice_api.MeshServiceResource
+	MeshServicesByLabelByValue          LabelsToValuesToResourceIdentifier
+	MeshExternalServiceByIdentifier     map[kri.Identifier]*meshexternalservice_api.MeshExternalServiceResource
+	MeshExternalServicesByLabelByValue  LabelsToValuesToResourceIdentifier
+	MeshMultiZoneServiceByIdentifier    map[kri.Identifier]*meshmzservice_api.MeshMultiZoneServiceResource
+	MeshMultiZoneServicesByLabelByValue LabelsToValuesToResourceIdentifier
```
lahabana: Looks from the way you use your apis that all these fields shouldn't be public no?

pkg/xds/context/destination_index.go
```diff
@@ -0,0 +1,150 @@
+package context
```
lahabana, line 1: BTW I think if we want to start making things a bit smoother to use we may want to add [internal packages](https://www.bytesizego.com/blog/golang-internal-package) I don't know if it's the case here but I think this would greatly help us defining more clear API boundaries.

pkg/xds/context/mesh_context_builder.go
```diff
@@ -298,12 +272,60 @@ func (m *meshContextBuilder) BuildBaseMeshContextIfChanged(ctx context.Context,
...
-		Mesh:        mesh,
-		ResourceMap: rmap,
+		hash:             newHash,
+		Mesh:             mesh,
+		ResourceMap:      rmap,
+		DestinationIndex: buildDestinationsIndex(rmap),
 	}, nil
 }
 
+func buildDestinationsIndex(resourceMap ResourceMap) *DestinationIndex {
```
lahabana: Feels like this should be NewDestinationIndex inside the destination_index.go

pkg/plugins/policies/core/xds/meshroute/clusters.go
Automaat: I couldn't return Protocol type from this method since I was getting circular dependencies errors on this. I can try maybe it could be entangled somehow

pkg/xds/context/mesh_context_builder.go
Automaat: moved

pkg/xds/context/destination_index.go
Automaat: changed to private fields

pkg/plugins/policies/core/xds/meshroute/clusters.go
Automaat: I managed to untangle some of this mess. But there are still issues that we sometimes use plugin code or other stuff in core packages. I think targetRef is mostly to blame. Maybe this should be promoted to core now?

pkg/xds/context/destination_index.go
Automaat: I am not sure if DestinationIndex should be moved into internal package. It still exposes public api that is used in other places

pkg/plugins/policies/core/xds/meshroute/clusters.go
lobkovilya: > we sometimes use plugin code or other stuff in core packages

where in core packages we use plugin code?

Automaat: https://github.com/kumahq/kuma/blob/a9be0381debb3febc0447fd78861b7873f4023e0/pkg/core/xds/matched_policies.go#L9

https://github.com/kumahq/kuma/blob/820494cf59e22d5a4870ca6d7a01f1f77e38ad46/pkg/core/xds/inspect/rules.go#L11

pkg/xds/context/destination_index.go
```diff
@@ -0,0 +1,192 @@
...
+	meshMultiZoneServiceByIdentifier    map[kri.Identifier]*meshmzservice_api.MeshMultiZoneServiceResource
+	meshMultiZoneServicesByLabelByValue LabelsToValuesToResourceIdentifier
+}
+
+func NewDestinationIndex(resourceMap ResourceMap) *DestinationIndex {
+	var meshServices []*meshservice_api.MeshServiceResource
+	if resourceMap[meshservice_api.MeshServiceType] != nil {
+		meshServices = resourceMap[meshservice_api.MeshServiceType].(*meshservice_api.MeshServiceResourceList).Items
+	}
+	meshServicesByKri := make(map[kri.Identifier]*meshservice_api.MeshServiceResource, len(meshServices))
```
lobkovilya: you can easily prevent repeating the logic 3 times if you implement 2 func:
```go
func indexByKri[T core_model.Resource](list []T) map[kri.Identifier]T {}

func indexByLabelKeyValue[T core_model.Resource](list []T) LabelsToValuesToResourceIdentifier {}
```

Automaat: Right, I will update it, it was copy paste from what we had previously

pkg/plugins/policies/core/xds/meshroute/clusters.go
lobkovilya: Ah, that sounded a bit misleading, `core_rules` is not "plugin code". The fact that `xds` and `inspect` packages are using `core_rules` sounds okay and I wouldn't expect `core_rules` to use either `xds` or `inspect` packages. 

> Maybe this should be promoted to core now?

I'd prefer dragging out the content of "core" into appropriate packages rather than bringing more stuff into it.

pkg/xds/context/destination_index.go
Automaat: added doc

pkg/core/resources/apis/core/interfaces.go
Automaat: added docs

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/helpers.go
Automaat: removed GetNameOrStringifyPort and left only GetName and documented it properly

pkg/core/resources/apis/core/interfaces.go
```diff
@@ -1,7 +1,22 @@
 package core
 
+import core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
```
lahabana, line 3: This is an interface package. That suddenly pulls an implem. This doesn't sound right to me.

It seems like what we want is:

https://github.com/kumahq/kuma/blob/1ead87e0c94bec07899322496d8c2f908ddc8341/pkg/core/resources/apis/mesh/dataplane_helpers.go#L24-L55

In this case I'd move these here and add a type alias in the original place.

This saves you from creating a weird dep, you only pull up a type

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/helpers.go
```diff
@@ -39,12 +40,12 @@ func (m *MeshMultiZoneServiceResource) findPort(port uint32) (Port, bool) {
...
 func (m *MeshMultiZoneServiceResource) FindSectionNameByPort(port uint32) (string, bool) {
 	if port, found := m.findPort(port); found {
-		return port.GetNameOrStringifyPort(), true
+		return port.GetName(), true
 	}
 	return "", false
 }
 
-func (m *MeshMultiZoneServiceResource) FindPortByName(name string) (Port, bool) {
+func (m *MeshMultiZoneServiceResource) FindPortByName(name string) (core.Port, bool) {
```
lahabana, line 32: The lines inside the for loop below are just:

That's:
```suggestion
if p.GetName() == name {
   return p, true
}
```

no?

pkg/core/resources/apis/core/interfaces.go
```diff
@@ -1,7 +1,22 @@
...
+
+// Port is a common abstraction for a destination port. It provides us basic information about port
 type Port interface {
+	// GetName returns port name or stringified port value. This can be used when building KRI of a destination
```
lahabana: Worth stating it's never empty

pkg/xds/context/destination_index.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/kumahq/kuma/pkg/core/resources/apis/core"
+	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
+	meshexternalservice_api "github.com/kumahq/kuma/pkg/core/resources/apis/meshexternalservice/api/v1alpha1"
+	meshmzservice_api "github.com/kumahq/kuma/pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1"
+	meshservice_api "github.com/kumahq/kuma/pkg/core/resources/apis/meshservice/api/v1alpha1"
+	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
+	"github.com/kumahq/kuma/pkg/plugins/policies/core/rules/resolve"
+)
+
+type DestinationIndex struct {
```
lahabana, line 17: I really appretiate the introduction of this type. I'd love a little of docs though :)

pkg/xds/context/destination_index.go
```diff
@@ -0,0 +1,205 @@
...
+	meshServiceByIdentifier             map[kri.Identifier]*meshservice_api.MeshServiceResource
+	meshServicesByLabelByValue          LabelsToValuesToResourceIdentifier
```
lahabana: I think you should pull this type here no? This will likely enable you to not make this type public

pkg/xds/context/destination_index.go
```diff
@@ -0,0 +1,205 @@
...
+	reachable := dc.getResourceNamesForLabels(kind, labels)
+	for ri, count := range reachable {
+		if count == len(labels) {
+			result = append(result, ri)
+		}
+	}
+	return result
+}
+
+func (dc *DestinationIndex) getSectionName(kind core_model.ResourceType, key kri.Identifier, port uint32) string {
```
lahabana: I don't get this method. Didn't you just do this in Destination.GetName() ?

pkg/xds/context/destination_index.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/kumahq/kuma/pkg/plugins/policies/core/rules/resolve"
+)
+
+type DestinationIndex struct {
+	meshServiceByIdentifier             map[kri.Identifier]*meshservice_api.MeshServiceResource
+	meshServicesByLabelByValue          LabelsToValuesToResourceIdentifier
+	meshExternalServiceByIdentifier     map[kri.Identifier]*meshexternalservice_api.MeshExternalServiceResource
+	meshExternalServicesByLabelByValue  LabelsToValuesToResourceIdentifier
+	meshMultiZoneServiceByIdentifier    map[kri.Identifier]*meshmzservice_api.MeshMultiZoneServiceResource
+	meshMultiZoneServicesByLabelByValue LabelsToValuesToResourceIdentifier
```
lahabana: Feels like the follow methods could be LOT simpler if this struct only had:

```
resourceByIdentifier map[kri.Indentifier]interface{}
resourceByLabelByValue LabelsToValuesToResourceIdentifier
```

we're using KRIs everywhere so there's no risk of conflicts, I think it would greatly simplify a bunch of methods after too

lahabana: one you've actually done this, it's even worth asking if we need a dep on resourceMap, and MeshServiceResource and friends in this file. Probalby not

`NewDestinationIndex(...[]Destination)` of something like this

pkg/xds/sync/dataplane_proxy_builder.go
```diff
@@ -122,7 +124,7 @@ func (p *DataplaneProxyBuilder) resolveVIPOutbounds(
 	bindOutbounds bool,
 ) []*xds_types.Outbound {
 	if !tpEnabled && !bindOutbounds {
-		return dataplane.AsOutbounds(meshContext.ResolveResourceIdentifier)
+		return asOutbounds(dataplane, meshContext.ResolveResourceIdentifier)
```
lahabana, line 127: nice occurence of the Scout rule here :)

lahabana (commented):
Good progress here! I think there's a bit more we can do but it's really helping clarifying things as well!

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/helpers.go
Automaat: not exactly, here we are checking if this name matches name or port. If we do `p.GetName()` we will prioritize matching by actual name and this method is used mostly for BackendRef where you cannot specify actual name, you can only specify port. 

I guess we should add possibility to use sectionName in backedRef

pkg/core/resources/apis/core/interfaces.go
Automaat: added

pkg/xds/context/destination_index.go
Automaat: changed

pkg/xds/context/destination_index.go
Automaat: this resolves actual section name from resource by port value. I've simplified it

pkg/xds/context/destination_index.go
Automaat: rewritten destination index to simplify it

pkg/xds/sync/dataplane_proxy_builder.go
Automaat: thanks

pkg/core/resources/apis/core/interfaces.go
Automaat: Moving this and adding type alias in original place will bring a lot of changes. I think that the best option would be to just move the whole type here, it will be similar effort. But I don't want to do it in this pr as this will become unreadable and big

pkg/xds/context/destination_index.go
Automaat: added doc

pkg/core/resources/apis/core/interfaces.go
lahabana: It's very unfortunate because that would like simplify a lot of the dependency graph but yes let's do this later

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/helpers.go
lahabana: yes can we create an issue? Btw if we select just the port why do we actually look at the name?

lahabana: Looking at the code I clearly don't understand the same as you. Also I seen it exists on all 3 service types why isn't that part of your interface?

pkg/xds/context/context.go
```diff
@@ -139,109 +124,40 @@ func (mc *MeshContext) ResolveResourceIdentifier(resType core_model.ResourceType
 }
 
 func (mc *MeshContext) GetMeshServiceByKRI(id kri.Identifier) *meshservice_api.MeshServiceResource {
-	return mc.MeshServiceByIdentifier[kri.NoSectionName(id)]
+	if id.ResourceType != meshservice_api.MeshServiceType {
```
lahabana: Should this be an assertion?

lahabana: Actually should these individual functions disappear in favour of `GetServiceByKRI()` in most cases it's easier contextual (inside a switch statement) so it will already be selected correctly.

there's even a couple of cases where the API is painful: https://github.com/kumahq/kuma/blob/ce3342fe9e4489f9f42259f1e19d8b294d4c25d3/pkg/plugins/policies/core/xds/meshroute/listeners.go#L280-L288

https://github.com/kumahq/kuma/blob/ce3342fe9e4489f9f42259f1e19d8b294d4c25d3/pkg/xds/context/context.go#L107-112

I think this is the only place where we'd need to ensure that if something is returned it must be a MeshServiceType: https://github.com/kumahq/kuma/blob/ce3342fe9e4489f9f42259f1e19d8b294d4c25d3/pkg/plugins/policies/core/xds/meshroute/listeners.go#L115

This can for sure be done just there.

pkg/xds/context/destination_index.go
```diff
@@ -0,0 +1,148 @@
...
+	core.Destination
+}
+
+func NewDestinationIndex[T DestinationResource](destinations []T) *DestinationIndex {
+	destinationByIdentifier := make(map[kri.Identifier]core.Destination)
+	destinationsByLabelByValue := labelsToValuesToResourceIdentifier{}
+	for _, item := range destinations {
+		ri := kri.From(item, "")
+		destinationByIdentifier[ri] = item
+		buildLabelValueToServiceNames(ri, destinationsByLabelByValue, item.GetMeta().GetLabels())
```
lahabana: That's a bit of personal preference but this function is called once and something larger functions are more readable than extracting subfunctions (especially when they mutate in place).

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/helpers.go
```diff
@@ -59,3 +64,39 @@ func (t *MeshExternalServiceResource) Domains() *xds_types.VIPDomains {
...
+	if name == strconv.Itoa(int(t.Spec.Match.Port)) {
+		return MesPort{
+			Value:    t.Spec.Match.Port,
+			Protocol: t.Spec.Match.Protocol,
+		}, true
+	}
+	return MesPort{}, false
+}
+
+type MesPort struct {
```
lahabana: Why redefine a type instead of adding the methods on the existing `t.Spec.Match` ?

pkg/xds/context/mesh_context_builder.go
```diff
@@ -297,10 +291,23 @@ func (m *meshContextBuilder) BuildBaseMeshContextIfChanged(ctx context.Context,
...
+	var destinations []DestinationResource
+	for _, resources := range rmap {
+		if !isDestinationResource(resources) {
+			continue
+		}
+
+		for _, res := range resources.GetItems() {
+			destinations = append(destinations, res.(DestinationResource))
+		}
+	}
```
lahabana: This is the hotpath (we create these contexts all the time).
The destinations list you are creating here is thrown away straight after the init building an iterator may be a better option.

Also using type descriptor after this code could be part of:
https://github.com/kumahq/kuma/blob/ce3342fe9e4489f9f42259f1e19d8b294d4c25d3/pkg/xds/context/mesh_context_builder.go#L276-L300

```go
case desc.IsDestination:
   destResources = append(destResources, resources.GetItems()
```

And later passed to `NewDestinationIndex(destResources...)`

This leverages existing patterns to make code easier to read and the day we add an extra Destination it works very easily.

pkg/xds/context/mesh_context_builder.go
```diff
@@ -583,3 +574,11 @@ func inferServiceProtocol(endpoints []xds.Endpoint) core_mesh.Protocol {
 	}
 	return serviceProtocol
 }
+
+func isDestinationResource(resources core_model.ResourceList) bool {
```
lahabana: This is exactly stuff that should be in the type descriptor. I think there should be a `IsDestination` akin to `IsProxy`: https://github.com/kumahq/kuma/blob/e279bba108307837bbe966f032c5e95039759372/pkg/core/resources/model/resource.go#L204

If you are worried about type explosion turning these 2 in flags like we did for KDS may make things simpler.

Advantage of this is that you could also ensure at generation that the types implement the `Destination` interface by generating a `var _ Destination = MyType{}`. Thus helping anyone that adds a destination in the future

lahabana (commented):
Great stuff! Things are getting there. I still have a couple of minor changes to simplify the code further possible the resourceType stuff could be done as a follow up but I think it should be done as part of this work stream as it simplifies greatly the code.

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/helpers.go
Automaat: I was thinking of just implementing interface on existing `t.Spec.Match` but @lobkovilya said that we should add dedicated type for it

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/helpers.go
Automaat: This method is part of the Destination interface.

So `GetName` method returns name if present or stringified port. Since `backendRef` can only select section by port value, if we use `GetName` here, we won't be able to select MS/MMZS/MES with named port. Because in this case we will always be comparing value with actual name. 

When you use `sectionName` in policies, you need to specify port name if it exists. If you specify port by value and name is present, policy won't be applied. That is how we handle sectionName. Even if we allow specifying `backendRef` with `sectionName` we will still need to support this for backward compatibility

pkg/xds/context/mesh_context_builder.go
Automaat: This makes sense, I somehow didn't think earlier about type descriptor. I updated it to your suggestions

pkg/xds/context/mesh_context_builder.go
Automaat: changed

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/helpers.go
Automaat: When you change this method to use `GetName` all route tests fail as it cannot resolve backendRefs to `realResources` since they all have named ports

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/helpers.go
lahabana: Any reasons other than just using more memory? We do this for other data types.

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/helpers.go
lahabana: I'm not sure I follow but seeing how complex that is it may make sense to explain this in the comments :)

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/helpers.go
Automaat: removed

pkg/xds/context/context.go
Automaat: moved these to single method

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/helpers.go
Automaat: created issue for backendRef with sectionName: https://github.com/kumahq/kuma/issues/13840

lahabana (commented):
sorry just noticed I didn't submit this :( 

I'll go over with you recent updates and potential resolve some threads

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/helpers.go
lahabana: Can you reference this issue in the interface definition. Noone reads issues, people read the code though :)

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -147,30 +148,18 @@ func SniForBackendRef(
 	meshCtx xds_context.MeshContext,
 	systemNamespace string,
 ) string {
-	var resource core_model.Resource
 	var name string
 	var port int32
```
lahabana, line 153: nit: I think you don't have to define these first anymore.
Also there's a case where port is 0, is this handled somewhere else or it doesn't matter? Should we creatre an issue to look into this?

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -147,30 +148,18 @@ func SniForBackendRef(
...
-		name = core_model.GetDisplayName(resource.GetMeta())
-		if p, ok := mzms.FindPortByName(backendRef.Resource.SectionName); ok {
-			port = p.Port
-		}
+	dest := meshCtx.GetServiceByKRI(pointer.Deref(backendRef.Resource))
+	if p, ok := dest.FindPortByName(backendRef.Resource.SectionName); ok {
+		port = p.GetValue()
+	}
+	resource := dest.(core_model.Resource)
+	name = core_model.GetDisplayName(resource.GetMeta())
```
lahabana:
```suggestion
	name := core_model.GetDisplayName(resource.GetMeta())
```

pkg/plugins/policies/core/xds/meshroute/listeners.go
```diff
@@ -108,70 +101,26 @@ func CollectServices(
...
 
-func collectMeshService(
+func collectService(
 	outbound *xds_types.Outbound,
 	meshCtx xds_context.MeshContext,
 ) *DestinationService {
-	ms := meshCtx.GetMeshServiceByKRI(pointer.Deref(outbound.Resource))
-	if ms == nil {
-		// we want to ignore service which is not found. Logging might be excessive here.
-		// We don't have other mechanism to bubble up warnings yet.
```
lahabana, line 118: Should we keep the comment?

pkg/plugins/policies/core/xds/meshroute/listeners.go
```diff
@@ -108,70 +101,26 @@ func CollectServices(
 	return dests
 }
 
-func collectMeshService(
+func collectService(
 	outbound *xds_types.Outbound,
 	meshCtx xds_context.MeshContext,
 ) *DestinationService {
```
lahabana, line 112: Feels like this function is simple enough to be inlined now no?

pkg/plugins/policies/core/xds/meshroute/listeners.go
```diff
@@ -108,70 +101,26 @@ func CollectServices(
...
-	if port.AppProtocol != "" {
-		protocol = port.AppProtocol
-	}
-	return &DestinationService{
-		Outbound:    outbound,
-		Protocol:    protocol,
-		ServiceName: svc.DestinationName(port.Port),
+		ServiceName: service.DestinationName(port.GetValue()),
 	}
 }
```
lahabana, line 133: same for this next function, It's so simple yet named so poorly that I think it's existence is more hurtful than helpful

pkg/plugins/policies/core/xds/meshroute/listeners.go
```diff
@@ -191,43 +140,17 @@ func GetServiceProtocolPortFromRef(
...
+	dest := meshCtx.GetServiceByKRI(pointer.Deref(ref.Resource))
+	if dest == nil {
 		return "", "", 0, false
 	}
+	port, ok := dest.FindPortByName(ref.Resource.SectionName)
+	if !ok {
+		return "", "", 0, false
+	}
+	service := dest.DestinationName(port.GetValue())
+	protocol := port.GetProtocol()
```
lahabana, line 147: There was a TODO in the MeshService case, is this something always solved with your interface?

tools/policy-gen/generator/cmd/core_resource.go
```diff
@@ -231,6 +238,7 @@ var {{.Name}}ResourceTypeDescriptor = model.ResourceTypeDescriptor{
 		KumactlListArg: "{{.Path}}",
 		AllowToInspect: {{.IsPolicy}},
 		IsPolicy: {{.IsPolicy}},
+        IsDestination: {{.IsDestination}},
```
lahabana, line 241:
```suggestion
{{- if .IsDestination }}
        IsDestination: true,
{{- end }}
```
This avoids the unecessary updates of all the type descriptors which makes things easier to review

pkg/plugins/policies/core/xds/meshroute/listeners.go
Automaat: added comment back

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/helpers.go
Automaat: extended comment on interface and linked issue

pkg/plugins/policies/core/xds/meshroute/clusters.go
Automaat: done

pkg/plugins/policies/core/xds/meshroute/listeners.go
Automaat: inlined

pkg/plugins/policies/core/xds/meshroute/listeners.go
Automaat: we have defaults in code now in all destinations and it defaults to tcp

tools/policy-gen/generator/cmd/core_resource.go
Automaat: I would leave this as we always add new fields here, I think that `ResourceTypeDescriptor` should contain all fields always so that when you read it you don't wonder if you are not forgetting some information about it

pkg/plugins/policies/core/xds/meshroute/listeners.go
Automaat: I don't like the idea of inlining it. I like small, easily readable functions even if they are used once. When I inline this there will be nested ifs which will make this harder to read and understand, from my perspective.

I've renamed this function to make it clear what it does

pkg/plugins/policies/core/xds/meshroute/clusters.go
Automaat: I've created issue https://github.com/kumahq/kuma/issues/13842

pkg/xds/context/destination_index.go
Automaat: I personally like having these separately, it make it easier to read for me. As I said in some other comments, I like small self-contained functions that builds up to bigger functionality

lahabana (approved):

lahabana (approved):

Issue #13054, opened 2025-03-11
☂️ resources name validation


Icarus9913:
### Description

xref:

* issues including:
  * https://github.com/kumahq/kuma/issues/9246
  * https://github.com/kumahq/kuma/issues/12954
  * https://github.com/kumahq/kuma/issues/12825

* First to add deprecation (2.10.x):
  * https://github.com/kumahq/kuma/pull/13003

* Cleanup deprecation and enforce the validation (2.12.x):
  * https://github.com/kumahq/kuma/issues/13047
  * https://github.com/kumahq/kuma/issues/12985

lukidzi:
@Icarus9913 please add more issues since most of them are closed and let's add them as a subissues so we have a correct relationships

lahabana:
@Icarus9913 is https://github.com/kumahq/kuma/issues/13015 fixed by what has already be done here?

Icarus9913:
> [@Icarus9913](https://github.com/Icarus9913) is [#13015](https://github.com/kumahq/kuma/issues/13015) fixed by what has already be done here?

I think we should wait for `2.13.x` for the compatibility(we added deprecations for some resource names' length in `2.10`) and enhance the KDS resource syncing logic for resource name suffix.

PR #13804, opened 2025-06-25
docs(MADR): design SPIFFE compliance

lukidzi:
* Added limitation for egress and admin-api cert
* design of MeshIdentity and MeshTrust resource

xref: #13661 

Needs merge of https://github.com/kumahq/kuma/pull/13699 so the diff is easier to read

Rendered: https://github.com/lukidzi/kuma/blob/madr-spiffe-design/docs/madr/decisions/073-spiffe-compliance.md

### DIFF
--- a/docs/madr/decisions/073-spiffe-compliance.md
+++ b/docs/madr/decisions/073-spiffe-compliance.md
@@ -290,7 +290,7 @@ The control plane uses the default CA bundle to validate the provided CA. By def
 
 **Potential collision**
 
-In scenarios where a user creates a `MeshIdentities` resource that specifies a trustDomain already used by another `MeshIdentity` or legacy trust domain (Mesh name), a conflict may arise. Because it's not always possible to detect this conflict during resource creation, we propose adding a `status` field to `MeshIdentities`. This field will be used to report post-creation validation issues, such as trust domain conflicts, making them visible to the user.
+In scenarios where a user creates a `MeshIdentities` resource that specifies a trustDomain already used by another `MeshIdentity` or legacy trust domain (Mesh name), a conflict may arise. Because it's not possible to detect this conflict during resource creation, we propose adding a `status` field to `MeshIdentities`. This field will be used to report post-creation validation issues, such as trust domain conflicts, making them visible to the user. The status will be synced to the global, as we do for all resources with `status` field.
 
 **ServiceAccount** 
 We are going to set `k8s.kuma.io/service-account` with the name of service account. The length of the possible value is [253 characters](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-multiple-service-accounts) which makes us to put it into the annotation and not a label.


docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+
+This resource allows the creation of Certificate Authorities (CAs) that are supported within a specific Mesh. It does not actively issue certificates — it simply defines the authority. The user must later configure which CA is used as the default, and which ones should continue to be supported during a migration period.
+
+* The resource can be created on the Global control plane and synced to Zones.
+* It can also be created on a Zone, in which case it will be synced to Global, but not propagated to other Zones.
+
+KDS flags:
+```golang
+model.GlobalToZonesFlag | model.ZoneToGlobalFlag
+```
```
lahabana, line 235: What's the shortName?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+spec:
+  selector:
+    dataplane:
+      matchLabels:
+        app: test
+  spiffeID: # optional
+    trustDomain: "prod.example.fr" # { .Mesh }.{ .Zone }.{ .ClusterID}.kuma.io
+    path: "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}" # let's use the same as HostnameGenerator
+  provider:
+    type: builtin | provided | spire
```
lahabana: I think the case here is not right.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+    builtin:
+      dataplaneCertificate:
+        duration: 24h
+    spire:
+      agent:
+        addresss: "/run/spire/sockets/agent.sock"
```
lahabana:
```suggestion
        address: "/run/spire/sockets/agent.sock"
```

Should the address be there or should it be a parameter of the CP?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+    provided: # to extend in KM
+      certificate:
+        # one of
+        secret:
+        inline:
+        inlineString:
+      privateKey:
+        # one of
+        secret:
+        inline:
```
lahabana: We shouldn't have inline and inlineString anymore. However, a file path or env var could be useful WDYT?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+        inlineString:
+      dataplaneCertificate:
+        duration: 24h
+```
+
+Currently, the `ServiceAccount` is not exposed in the `Dataplane` resource, which is problematic since we need to access this information. A potential solution is to add a dedicated field to the `Dataplane` object to store the associated `ServiceAccount`.
```
lahabana: Shouldn't this just be part of the bootstrap request and ctx?

I guess we can't because of needing the SA for building MeshService identities?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+Users can provide certificates in the following ways:
+
+* Inline,
+* File path,
+* Secret reference.
+
+This gives flexibility for organizations that already have an established PKI or need to integrate with existing certificate authorities.
+
+```yaml
+    provided: # to extend in KM
```
lahabana: Should builtin just be an option of this one in the end?

This would be:

```
insecure_autogenerate: true # Will generate a CA and store them in secrets, the name will be: "TBD"
```

lahabana: And potentially have an option `trust_extraction_disabled: true`.

Also it would be good to state in this doc what is required from the CAs that is used there.

lahabana: Also one thing for the future is with all this stuff I think we can do CA auto rotation :)

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+```yaml
+    spire:
+      agent:
+        addresss: "/run/spire/sockets/agent.sock"
+        timeout: 1s
+``` 
+
+Another thing is that we need to mount a volume with Spire Agent unix socket.
+
+1. Add a new section to the injector configuration:
```
lahabana: So do you need this to be aligned with the address in MeshIdentity no? Feels like it's unecessary there no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+
+```yaml
+runtime:
+  kubernetes:
+    injector:
+      spire:
+        path: /run/spire/sockets
+        mountPath: /run/spire/sockets
```
lahabana, line 375: injector-mode: annotation | always | never (never is default).

This probably makes it easier when it's systematic...

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+```
+* and inject label
+
+```yaml
+kuma.io/spire-integration: enabled
+```
+
+Purpose of the Label:
+* This label signals that the pod has SPIRE integration enabled.
+* It prevents Kuma from delivering Spire-related configuration (e.g., SDS config, identity provider references) to pods that don't have the socket mounted.
```
lahabana: So there'll be a DP feature-flag to indicate this right? Just like we do for dynconfig no?
Also we don't need a label we can directly set the correct env var (That's similar to the thing from @bartsmykla ).

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+
+2. We could expose the configuration in control-plane to be more generic
```
lahabana, line 399: Ideally our Spire integration should be very straightforward. This is too hard IMO

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+  sidecarPatch:
+    - op: add
+      path: /volumeMounts/-
+      value:
+        name: spire-agent-socket
+        mountPath: /run/spire/sockets/
+        readOnly: true
+```
+
+3. We could introduce templating/overlay (out-of-scope for this MADR)
```
lahabana, line 432: I don't think we should even consider this. It's more stuff for something we already have (contrainer-patches)

docs/madr/decisions/073-spiffe-compliance.md
lahabana: Though do you need this? In the end you'll send this SDS config etc IIF the MeshIdentity is spire no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+      matchLabels:
+        app: demo-app
+```
+
+The `MeshIdentity` resource’s selector field allows selection based on dataplane labels. Users can define it to apply to either all dataplanes or a specific subset.
+If there are multiple (X) identities selecting the same dataplane, the control plane will resolve the conflict by choosing one based on the following order of precedence:
+
+1. Selector specificity – the identity with most specific selector is preferred.
+2. Lexicographical order – if selector sizes are equal, the identity with the lexicographically smallest name is chosen.
+3. Age – if names are the same, the oldest identity (based on creation timestamp) is selected.
```
lahabana: Name can't be the same right?

docs/madr/decisions/073-spiffe-compliance.md
lahabana: Also want to say: https://github.com/kumahq/kuma/blob/e279bba108307837bbe966f032c5e95039759372/pkg/core/resources/model/resource.go#L216

It almost makes sense to show the entire type descriptor here (at least the non default fields).

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+2. User creates MeshTrust with CA of a new MeshIdentity `ca-2`
+3. Trust is propagated. We provide identity and keep the information(supported CA (Trust), and Identity) in DataplaneInsight
+```yaml
+...
+name: demo-client-65f94cb577-jxrrq-44wfv42db9z44xxf
+type: DataplaneInsight
+mTLS:
+  certificateExpirationTime: "2025-06-03T17:35:33Z"
+  certificateRegenerations: 1
+  issuedBackend: ca-1 # name of MeshIdentity
```
lahabana: KRI no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+  lastCertificateRegeneration: "2025-06-02T17:35:33.624189876Z"
+  supportedBackends: # list of MeshTrust name
```
lahabana, line 590: KRI no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+
+Based on this information we will update MeshService
+
+```yaml
+...
+name: second-test-server-b7dwz26bb9b4z676
+spec:
+  identities:
+  - type: SpiffeID
+    value: spiffe://example.com/zone/my-new-zone/ns/real-resource-ns/sa/default
```
lahabana: needs updating at least to have valid example

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+![Rotation flow](assets/073/rotation-flow.png)
+
+#### Secret naming
+
+We want to use a [KRI](070-resource-identifier.md) for resource naming. Since `MeshIdentity` creates a separate resource (Secret), it should be named based on the KRI to simplify correlation between the `Secret` and the `MeshIdentity` configuration.
+
+Proposed name:
+
+```yaml
+kri_mi_mesh-1_us-east-2_kuma-system_identity-1
```
lahabana: not going to work on k8s as this is a secret so it needs to be the right DNS format

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+
+Also, we need to name the Spire agent cluster, in this case is it internal or maybe based on MeshIdentity:
+Proposed name:
+```yaml
+system_spire-agent
+```
+
+or 
+```yaml
+kri_mi_mesh-1_us-east-2_kuma-system_spire-identity
```
lahabana: I think @slonka left a path just for this: `system_kri_mi_mesh-1_us-east-2_kuma-system_spire-identity`

lahabana: Also I think mid might be a better shortname no?

lahabana: I think this whole part goes too fast. It covers 2 options 1 where we're talking about envoy secret resource name and the other is an envoy cluster for sds

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+Proposed name:
+
+```yaml
+kri_mi_mesh-1_us-east-2_kuma-system_identity-1
+```
+
+Also, we need to name the Spire agent cluster, in this case is it internal or maybe based on MeshIdentity:
+Proposed name:
+```yaml
+system_spire-agent
```
lahabana:
```suggestion
system_sds-spire-agent
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshTrust
+metadata:
+  name: trust-hash
+  namespace: kuma-system # only in system namespace
+  labels:
+    kuma.io/mesh: default
+spec:
+  ca: inline | secret | file
```
lahabana: That's a caBundle right?

lahabana: Btw the CA rotation without trustDomain change is it just a matter of having the old and new bundle and then removing the old once it's expired or if a user wants?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+The resource should also be manually creatable by the user. This enables users to define additional trust domains and allows the local sidecar to accept incoming mTLS traffic from services outside the mesh.
+This functionality is essential for:
+
+* Enabling ExternalClients,
+* Accepting mTLS connections from external services, and
+* Federating multiple separate clusters through shared trust.
+
+**MeshTrust based on MeshIdentity**
+
+When a user creates a `MeshIdentity`, we will automatically create a corresponding `MeshTrust` resource using the CA and trust domain from that identity. We can implement a dedicated generator that creates the MeshTrust based on the `MeshIdentity`.
```
lahabana: I suggested we'd make this optional in the MeshIdentity resource.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,711 @@ Future releases
...
+To avoid issues where a user removes a `MeshIdentity` but an old trust remains in use, we will not remove a `MeshTrust` and we let user to clean it up. In case it's problematic for the user we can implement it later.
```
lahabana: What happens with updates to MeshIdentity or deletion/recreation? Is the MeshTrust replaced? Updated with both?

docs/madr/decisions/073-spiffe-compliance.md
lahabana: would it be useful to have an optional origin?

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes it makes sense. I based this on https://github.com/kumahq/kuma/blob/master/api/common/v1alpha1/datasource.go
```golang
// DataSource defines the source of bytes to use.
type DataSource struct {
	// Data source is a secret with given Secret key.
	Secret *string `json:"secret,omitempty"`
	// Data source is inline bytes.
	Inline *[]byte `json:"inline,omitempty"`
	// Data source is inline string`
	InlineString *string `json:"inlineString,omitempty"`
}
```

We might need a new type

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, we need SA for building MeshService identity

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, it makes sense to remove here and don't make an error when configuring CP and policy

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, it maybe makes more sense to have it in configuration of the control-plane to avoid potential issues

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, it's problematic without the flag, we shouldn't probably update existing one. It makes sense to add a `MeshIdentity.trustExtractionDisabled` and let user control it.

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, it makes sense to add origin

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: It's not k8s resource it's secret name in Envoy config

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, it shouldn't be the same

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: The only issue is that once you enable spire and you don't have a mounted volume it will send config which will be rejected

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+spiffe://<trustDomain>/ns/<namespace>/sa/<serviceAccount>
+```
+
+Trust domain: `{ .Mesh }.{ .Zone }.{ .ClusterID }.kuma.io`
+
+Once a TrustDomain is defined on the Global control plane, we require it to include the zone name in the template. We are going to implement a validator for this case.
+
+Question:
+* Is there an issue once we rename zone ? Isn't it an edge case that we might don't need to support?
+  It's not supported and can cause many issues, we shouldn't think about it.
```
lahabana, line 199: +1 do we already ensure this at runtime by crashing the CP if the zone has changed name? If not can we open an issue to track this?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+
+### MeshIdentity
+
+This resource allows the creation of Certificate Authorities (CAs) that are supported within a specific Mesh. It does not actively issue certificates — it simply defines the authority. The user must later configure which CA is used as the default, and which ones should continue to be supported during a migration period.
```
lahabana: I don't fully understand the last sentency. The user doesn't really configure the default anymore right? It's done through conflict resolution right? Also the supported for the migration period part. What do you mean here?

It might be worth mentioning that this resource is only about Issuing identity, it's not about trust

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+  labels:
+    kuma.io/mesh: default
+spec:
+  selector:
+    dataplane:
+      matchLabels:
+        app: test
+  spiffeID: # optional
+    trustDomain: "prod.example.fr" # { .Mesh }.{ .Zone }.{ .ClusterID}.kuma.io
+    path: "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}" # let's use the same as HostnameGenerator
```
lahabana:
```suggestion
    path: "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}" # let's use the same templating format as HostnameGenerator
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+* spire
+
+**Provided**
+
+The provided type allows users to supply their own CA certificates. These can be configured per zone or as a global definition.
+Users can provide certificates in the following ways:
+
+* Environment variable,
+* File path,
+* Secret reference.
```
lahabana:
```suggestion
* Secret reference.
* Auto generated and stored in a Secret
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+
+This gives flexibility for organizations that already have an established PKI or need to integrate with existing certificate authorities.
+
+```yaml
+    provided: # to extend in KM
+      insecureAutogenerate: false
```
lahabana: I've opened: https://github.com/Kong/aip/issues/126 to cover this

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+
+**Proposal**
+
+For the current release, we propose defining the default SPIFFE ID for Kubernetes workloads with the following structure:
+
+```
+spiffe://<trustDomain>/ns/<namespace>/sa/<serviceAccount>
+```
+
+Trust domain: `{ .Mesh }.{ .Zone }.{ .ClusterID }.kuma.io`
```
lobkovilya: We started limiting `mesh` and `zone` to be 63 char max not that long ago https://github.com/kumahq/kuma/pull/9191, so technically users with long mesh and zone names can break domain limit of 253 characters. We need to make sure we're handling this case properly

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+  namespace: kuma-system # only in system namespace
+  labels:
+    kuma.io/mesh: default
+spec:
+  selector:
+    dataplane:
+      matchLabels:
+        app: test
+  spiffeID: # optional
+    trustDomain: "prod.example.fr" # { .Mesh }.{ .Zone }.{ .ClusterID}.kuma.io
```
lobkovilya: nit: could you please add an example that follows the template `{ .Mesh }.{ .Zone }.{ .ClusterID}.kuma.io`

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+      privateKey:
+        # one of
+        secret:
+        path:
+        envVar:
+      dataplaneCertificate:
+        duration: 24h
+```
+
+Additionally, we could introduce a configuration option called `insecureAutogenerate`, which automatically generates a default CA and persists it in the store. This would functionally be equivalent to the current builtin CA.
```
lobkovilya: today for builtin you can specify expiration and some other info, are we going to default it with `insecureAutogenerate`?

docs/madr/decisions/073-spiffe-compliance.md
lahabana: I'm actually unsure how insecure that is... Maybe it's worth @bartsmykla to dig for some info around this...

docs/madr/decisions/073-spiffe-compliance.md
lahabana: I think it's fine it's just part of doing the thigns in the right way no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+```
+
+* mount volume
+```yaml
+volumeMounts:
+  - name: spire-agent-socket
+    mountPath: /run/spire/sockets
+    readOnly: true
+```
+* and inject label
```
lahabana: annotaiton no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+
+1. User has a Mesh with MeshIdentity `identity-1`
+```yaml
+...
+name: demo-client-65f94cb577-jxrrq-44wfv42db9z44xxf
+type: DataplaneInsight
+mTLS:
+  certificateExpirationTime: "2025-06-03T17:35:33Z"
+  certificateRegenerations: 1
+  issuedBackend: system_kri_mi_mesh-1_us-east-2_kuma-system_identity-1
```
lahabana: Why system here? I don't think that's necessary, this should just be a KRI no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+
+Based on this information we will update MeshService
+
+```yaml
+...
+name: second-test-server-b7dwz26bb9b4z676
+spec:
+  identities:
+  - type: SpiffeID
+    value: spiffe://old.zone-1.0c5afd2e-de93-437b-8ba0-3decdff805b6.kuma.io/ns/real-resource-ns/sa/default
```
lahabana: seriously the clusterID is a UUID?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+
+> [!WARNING]
+> The naming is different when using SPIRE, since SPIRE uses SPIFFEID or `default` as a certificate name
+> https://spiffe.io/docs/latest/deploying/spire_agent/#sds-configuration
+> https://spiffe.io/docs/latest/microservices/envoy/#tls-certificates
+#### MeshTrust
+
+Based on the existing `MeshIdentity`, we should create a `MeshTrust` resource, which will be used to properly generate trust domains for other zones.
+`MeshTrust` should be created in each zone and then synchronized with the global control plane.
+1. Can be created on the Zone and synced to Global
```
lahabana, line 674: Actually why?

lahabana: I think this can all be figured as part of 4 no?

docs/madr/decisions/073-spiffe-compliance.md
lahabana: could we have a "first expiry" date for the bundle?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+The resource should also be manually creatable by the user. This enables users to define additional trust domains and allows the local sidecar to accept incoming mTLS traffic from services outside the mesh.
+This functionality is essential for:
+
+* Enabling ExternalClients,
+* Accepting mTLS connections from external services, and
+* Federating multiple separate clusters through shared trust.
+
+**MeshTrust based on MeshIdentity**
+
+When a user creates a `MeshIdentity`, we will automatically create a corresponding `MeshTrust` resource using the CA and trust domain from that identity. We can implement a dedicated generator that creates the MeshTrust based on the `MeshIdentity`. This option can be disabled and `MeshTrust` might not be created.
```
lahabana: hwo can it be disabled?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+To avoid issues where a user removes a `MeshIdentity` but an old trust remains in use, we will not remove a `MeshTrust` and we let user to clean it up. In case it's problematic for the user we can implement it later. Once user updates a MeshIdentity we shouldn't update existing MeshTrust and keep it as it is. We added a new type into `MeshIdentity.provided.trustExtractionDisabled` which allows controlling this.
```
lahabana: Should MeshTrust be a list and we allow appending? This would support the use case of rotating CA and recreation. The user can then remove entries they no longer judge valid.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+  namespace: kuma-system # only in system namespace
+  labels:
+    kuma.io/mesh: default
+spec:
+  ca: inline | secret | file
+  trustDomain: "prod.zone-1.id.kuma.io"
+  originRef:
+    kind: MeshIdentity
+    name: default
+    namespace: kuma-system
```
lahabana: This works well across zone?

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Currently we don't crash, we have some errors while running. I will open an issue https://github.com/kumahq/kuma/issues/13838

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+kind: MeshIdentity
+metadata:
+  name: old-identity
+  namespace: kuma-system # only in system namespace
+  labels:
+    kuma.io/mesh: default
+spec:
+  selector:
+    dataplane:
+      matchLabels: {} # select all
```
lobkovilya, line 473: It would be nice to also cover if any of the following is allowed and what means if allowed:
```yaml
# no selector
spec:
  provider: ...
```
```yaml
spec:
  selector: {}
  provider: ...
```
```yaml
spec:
  selector:
     dataplane: {}
  provider: ...
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+...
+name: demo-client-65f94cb577-jxrrq-44wfv42db9z44xxf
+type: DataplaneInsight
+mTLS:
+  certificateExpirationTime: "2025-06-03T17:35:33Z"
+  certificateRegenerations: 1
+  issuedBackend: system_kri_mi_mesh-1_us-east-2_kuma-system_identity-1
+  lastCertificateRegeneration: "2025-06-02T17:35:33.624189876Z"
+  supportedBackends: # list of MeshTrust name
+  - system_kri_mtrust_mesh-1_us-east-2_kuma-system_identity-1
```
lobkovilya: why `system_`?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+Update SpiffeID based on identities
+```yaml
+...
+name: second-test-server-b7dwz26bb9b4z676
+spec:
+  identities:
+  - type: SpiffeID
+    value: spiffe://old.zone-1.0c5afd2e-de93-437b-8ba0-3decdff805b6.kuma.io/ns/real-resource-ns/sa/default
+  - type: SpiffeID
+    value: spiffe://new.zone-1.0c5afd2e-de93-437b-8ba0-3decdff805b6.kuma.io/ns/real-resource-ns/sa/default
```
lobkovilya: why `old` and `new`? it's supposed to be a mesh name, no?

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: In a way, it makes sense - you want SPIRE, and the configuration is being delivered. However, if the required volume is missing, the pod isn't actually ready. This mechanism provides a safer way to enable SPIRE without breaking existing running workloads. It might not be required during first iteration but I think it's nice to have

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+    targetPort: main
+  selector:
+    dataplaneTags:
+      app: second-test-server
+      k8s.kuma.io/namespace: real-resource-ns
+  state: Available
+status:
+  ...
+  tls:
+    status: Ready <- only if all DPPs have identity provided
```
lobkovilya: all DPPs have "identity provided" or "trust provided"?

lobkovilya: Also I thought we don't want real traffic to be affected by the info we gathered through DataplaneInsights? But currently we use `status.tls.status` when configuring outbounds for MeshService

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,753 @@ Future releases
...
+  - system_kri_mtrust_mesh-1_us-east-2_kuma-system_identity-1
+  - system_kri_mtrust_mesh-1_us-east-2_kuma-system_identity-2
+...
+```
+
+Based on this information we will update MeshService
+
+```yaml
+...
+name: second-test-server-b7dwz26bb9b4z676
```
lobkovilya: In this example, does MeshService `second-test-server-b7dwz26bb9b4z676` select Dataplane  `demo-client-65f94cb577-jxrrq-44wfv42db9z44xxf`? or it's some other service that demo-client communicates with?

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: I don't think we should support incorrect names and try to be strict with 63 long validation of each part

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, we should configure it to secure way but I feel we don't need to allow customization since if a user wants a more secure it can provide it's own CA. I will add more details what kind of CA is a default

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Updated example with correct names

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: We are using it now so I don't want to change this logic

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Fixed

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: I this case we check only Identity Provided

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: We don't sync across zones MeshTrust, so we might not have an issue. If we want to sync it can be an issue since MeshIdentity might not be in another zone, the other option is to put it in status and don't sync status cross zone

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: based on this yes https://github.com/kumahq/kuma/blob/master/pkg/clusterid/creator.go#L29-L45

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: `MeshIdentity.provided.trustExtractionDisabled` can disable creation, also we could introduce a flag for controller which will be responsible for this

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: The question is: Do we want to make Multizone easy or not? I think we have options:

1. Create on Global -> Sync to all zones (makes sense)
2. Create on Zone and Sync to Global -> it would improve debugging
3. Create on Zone and Sync to other Zone -> I am not sure if that should be the case, it makes multizone easy but I think user might not want to sync some resource to other zone. 
But I am open for other opinion

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,840 @@ Future releases
...
+        path:
+        envVar:
+      dataplaneCertificate:
+        duration: 24h
+```
+
+Additionally, we could introduce a configuration option called `insecureAutogenerate`, which automatically generates a default CA and persists it in the store. This would functionally be equivalent to the current builtin CA.
+Currently, Kuma uses a CA with the RSA algorithm and a key size of 2048 bits, which is widely supported and compatible.
+
+* Algorithm: RSA
```
lukidzi: @bartsmykla does it make sense? or should we switch to ECDSA?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,840 @@ Future releases
...
+    trustDomain: "default.zone-1.0c5afd2e-de93-437b-8ba0-3decdff805b6.kuma.io" # { .Mesh }.{ .Zone }.{ .ClusterID }.kuma.io
+    path: "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}" # let's use the same templating format as HostnameGenerator
+  provider:
+    type: Provided | Spire
+    spire:
+      agent:
+        timeout: 1s
+    provided: # to extend in KM
+      trustExtractionDisabled: false # allows to disable extraction of CA into MeshTrust
+      insecureAutogenerate: false
```
slonka: nitpick: should we have enums instead of booleans (align with k8s api guide)? like `trustExtraction: disabled|enabled` and `autogenerate: off|insecure|...`

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,840 @@ Future releases
...
+apiVersion: kuma.io/v1alpha1
+kind: MeshIdentity
+...
+spec:
+  selector:
+    dataplane:
+      matchLabels: {}
+```
+
+3. Select nathing
```
slonka: nitpick:

```suggestion
3. Select nothing
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,840 @@ Future releases
...
+func LabelSelectorAsSelector(ps *LabelSelector) (labels.Selector, error) {
+	if ps == nil {
+		return labels.Nothing(), nil
+	}
+	if len(ps.MatchLabels)+len(ps.MatchExpressions) == 0 {
+		return labels.Everything(), nil
+	}
+```
+
+#### How would look the process of switching backends?
```
slonka: nitpick:

```suggestion
#### What would the process of switching backends look like?
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,916 @@ Future releases
...
+spiffe://<trustDomain>/ns/<namespace>/sa/<serviceAccount>
+```
+
+Trust domain: `{ .Mesh }.{ .Zone }.kuma.io`
+
+Once a TrustDomain is defined on the Global control plane, we require it to include the zone name in the template. We are going to implement a validator for this case. Also, we need to validate length of each segment where Mesh and Zone has limit of 63 characters.
+
+Limitation:
+* Mesh and Zone needs to have max length of 63 characters each
+
```
Icarus9913: In 2.13.*, the `Mesh`, `Zone` etc resources would follow [RFC-1035](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names). In the current implementation, do you mean to use truncate?

Here's the reference: https://github.com/kumahq/kuma/blob/master/UPGRADE.md#add-rfc-1035-label-constraints-for-specific-resources-names

lukidzi: No, just that we are going to validate a length of Mesh and Zone before creating identity

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,916 @@ Future releases
...
+  spiffeID: # optional
+    trustDomain: "default.zone-1.kuma.io" # { .Mesh }.{ .Zone }.kuma.io
+    path: "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}" # let's use the same templating format as HostnameGenerator
+  provider:
+    type: Bundled | Spire
+    spire:
+      agent:
+        timeout: 1s
+    bundled: # to extend in KM
+      trustExtraction: Disabled | Enabled # allows to disable extraction of CA into MeshTrust
```
lahabana:
```suggestion
      MeshTrustCreation: Disabled | Enabled # allows to disable extraction of CA into MeshTrust
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,916 @@ Future releases
...
+      insecureAllowSelfSigned: true
```
lahabana, line 261: it would like be useful to add that this is ensured that CA cert or autogenerate are not self signed

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,916 @@ Future releases
...
+          envVar:
+        privateKey:
+          type: File | Secret | EnvVar
+          # OneOf
+          secret:
+          path:
+          envVar:
+```
+
+CP uses a default CA Bundle to validate provided CA.
```
lahabana:
```suggestion
CP uses the default CA Bundle to validate provided CA.
```

Linked to the config would likely be useful here.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,916 @@ Future releases
...
+          # OneOf
+          secret:
+          path:
+          envVar:
+```
+
+Additionally, we could introduce a configuration option called `autogenerate`, which automatically generates a default CA and persists it in the store. This would functionally be equivalent to the current builtin CA.
+Currently, Kuma uses a CA with the RSA algorithm and a key size of 2048 bits, which is widely supported and compatible.
+
+* Algorithm: Ed25519 - it works for newer versions of TLS.
```
lahabana, line 344: Have we checked with security whether we can do this?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,916 @@ Future releases
...
+#### MeshTrust
+
+Based on the existing `MeshIdentity`, we should create a `MeshTrust` resource, which will be used to properly generate trust domains for other zones.
+`MeshTrust` should be created in each zone and then synchronized with the global control plane.
+1. Can be created on the Zone and synced to Global
+2. Can be created on the Global and synced to all Zones
+3. Can be created on the Zone and not synced to global by providing label `kuma.io/kds-sync: disabled`
+4. Can be created on the Zone and synced to all zones (Future release) (`SyncedAcrossZonesFlag`)
+
+We can describe it as KDS flag:
```
lahabana: that's a type descriptor

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,916 @@ Future releases
...
+  name: trust-hash
+  namespace: kuma-system # only in system namespace
+  labels:
+    kuma.io/mesh: default
+spec:
+  origin:
+    kri: kri_mid_mesh-1_us-east-2_kuma-system_identity-1
+  caBundles:
+    - caBundle:
+    - caBundle:
```
lahabana: what's the format? PEM?

lahabana (approved):

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes, @bartsmykla recommended this after conversation with security team

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,916 @@ Future releases
...
+metadata:
+  name: identity
+  namespace: kuma-system # only in system namespace
+  labels:
+    kuma.io/mesh: default
+spec:
+  selector:
+    dataplane:
+      matchLabels:
+        app: demo-app
```
lobkovilya: since it's for all services the matchLabels should be updated

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,916 @@ Future releases
...
+
+Yes, if we set `trusted_ca` envoy doesn't do validation of any SAN, hash, etc. https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/ssl#example-configuration
+
+> If only trusted_ca is specified, Envoy will verify the certificate chain of the presented certificate, but not its subject name, hash, etc. Other validation context configuration is typically required depending on the deployment.
+
+SPIFFE validator is experimental but SPIRE uses it. It sets a `custom_validator_config` which allows separation of each bundle to separate trust domain:
+
+
+```yaml
+                  custom_validator_config:
```
lobkovilya, line 920: So how is it going to look when we're switching from old builtin mTLS to new MeshIdentity? does it mean in trust domains we have both `spiffe://<mesh>/` for old and `spiffe://{ .Mesh }.{ .Zone }.kuma.io/` for new?

lukidzi: Yes we need both at this time

lobkovilya: Is new `SPIFFECertValidatorConfig` able to validate our certs that were generated the old way? Is this validator okay with the fact our old certs are not SPIFFE compatible?

lukidzi: Yes, seems it has no issue. You can take a look here https://github.com/lukidzi/envoy-spiffe-certs/blob/main/server.yaml

lukidzi: Based on the envoy code, we check only first san, and since we put in kuma `spiffe` san as a first record it works correctly

```cc
  for (const GENERAL_NAME* general_name : san_names.get()) {
    if (general_name->type != GEN_URI) {
      continue;
    }

    const std::string san = Utility::generalNameAsString(general_name);
    trust_domain = SPIFFEValidator::extractTrustDomain(san);
    // We can assume that valid SVID has only one URI san.
    break;
  }
...
...
...
std::string SPIFFEValidator::extractTrustDomain(const std::string& san) {
  static const std::string prefix = "spiffe://";
  if (!absl::StartsWith(san, prefix)) {
    return "";
  }

  auto pos = san.find('/', prefix.size());
  if (pos != std::string::npos) {
    return san.substr(prefix.size(), pos - prefix.size());
  }
  return "";
}

```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,917 @@ Future releases
...
+
+**Proposal**
+
+For the current release, we propose defining the default SPIFFE ID for Kubernetes workloads with the following structure:
+
+```
+spiffe://<trustDomain>/ns/<namespace>/sa/<serviceAccount>
+```
+
+Trust domain: `{ .Mesh }.{ .Zone }.kuma.io`
```
lobkovilya: Maybe we shouldn't use `kuma.io` domain as a default:
* it'll be visible in Kong Mesh
* it might give wrong impression the default cert is going to be rooted from real `kuma.io` cert which is not the case

Maybe `.mesh.local` as we do for MeshServices by default https://github.com/kumahq/kuma/blob/8af10f0f889ec4c46c7aa69fdac331565e04312b/pkg/defaults/hostname_generator.go#L54?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,917 @@ Future releases
...
+      certificateParameters: # lets make rotation percentage static for the moment
+        expiry: 24h
+      autogenerate:
+        enabled: true
+      ca:
+        certificate:
+          type: File | Secret | EnvVar
+          # OneOf
+          secret:
+          path:
```
lobkovilya: the type is `File`, need to change either field or type

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: @lahabana What do you think? I think it makes sense since we use `mesh.local` as a default in HostnameGenerator

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Changed to `file`

docs/madr/decisions/073-spiffe-compliance.md
lahabana: Is that compliant with spiffe? If it is and you feel it doesn't create confusion go for it.

For the 1st Argument Kuma is all over the place in KM.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+```
+spiffe://<trustDomain>/ns/<namespace>/sa/<serviceAccount>
+```
+
+Trust domain: `{ .Mesh }.{ .Zone }.kuma.io`
+
+Once a TrustDomain is defined on the Global control plane, we require it to include the zone name in the template. We are going to implement a validator for this case. Also, we need to validate length of each segment where Mesh and Zone has limit of 63 characters.
+
+Limitation:
+* Mesh and Zone needs to have max length of 63 characters each
```
lahabana: Should we consider this to be a future restriction? If yes, can we open an issue?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+          envVar:
+        privateKey:
+          type: File | Secret | EnvVar
+          # OneOf
+          secret:
+          file:
+          envVar:
+```
+
+CP uses the default CA Bundle to validate provided CA. Currently we don't set it, but we can set it additionally by setting `KUMA_GENERAL_TLS_CA_BUNDLE_FILE`.
```
lahabana:
```suggestion
CP uses the default CA Bundle to validate provided CA. By default, we'll use the system bundle but a user can override this with `KUMA_MESHIDENTITY_TLS_CA_BUNDLE_FILE` which defaults to `KUMA_GENERAL_CA_BUNDLE_FILE`
```

We usually have a specific config that default to the general one no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+
+`insecureAllowSelfSigned` allows preventing users from using self signed CA either provided by the user or autogenrated.
```
lahabana:
```suggestion
`insecureAllowSelfSigned` makes using self signed CA an explicit opt-in. This is used with both provided by the user or autogenrated CAs.
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+
+**ServiceAccount** 
+We are going to set `k8s.kuma.io/service-account` with the name of service account. The length of the possible value is [253 characters](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-multiple-service-accounts) which makes us to put it into the annotation and not a label. That let us resolve SPIFFEID.
```
lahabana:
```suggestion
We are going to set `k8s.kuma.io/service-account` with the name of service account. The length of the possible value is [253 characters](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-multiple-service-accounts) which makes us to put it into the annotation and not a label.
This will be used to generate spiffeId 
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+
+#### Provider types
+
+There can be multiple difference identity providers but in the first iteration we would like to focus on:
```
lahabana:
```suggestion
There can be multiple identity providers but in the first iteration we would like to focus on:
```
(multiple usually implies different)

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+### API
+
+#### Default SPIFFE ID
+
+Spire, which is an implementation of SPIFFE, defines a standard way for Subject Alternative Name (SAN) definition. By default, on Kubernetes, it uses the Service Account (SA) and namespace in a path format such as:
+
+```
+spiffe://<trustDomain>/ns/<namespace>/sa/<serviceAccount>
+```
+This default model works for us on Kubernetes but will not be supported in Universal deployments. We therefore need to find a default SPIFFE ID format that works consistently for both Kubernetes and Universal environments.
```
bartsmykla, line 178: Nit
```suggestion

This default model works for us on Kubernetes but will not be supported in Universal deployments. We therefore need to find a default SPIFFE ID format that works consistently for both Kubernetes and Universal environments.
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+
+For the current release, we propose defining the default SPIFFE ID for Kubernetes workloads with the following structure:
+
+```
+spiffe://<trustDomain>/ns/<namespace>/sa/<serviceAccount>
+```
+
+Trust domain: `{ .Mesh }.{ .Zone }.kuma.io`
+
+Once a TrustDomain is defined on the Global control plane, we require it to include the zone name in the template. We are going to implement a validator for this case. Also, we need to validate length of each segment where Mesh and Zone has limit of 63 characters.
```
bartsmykla:
```suggestion
Once a trust domain is defined on the Global control plane, we require it to include the zone name in the template. We are going to implement a validator for this case. Also, we need to validate length of each segment where Mesh and Zone has limit of 63 characters.
```

bartsmykla: Could you also elaborate on why these two elememts has limit of 63 characters>

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+runtime:
+  kubernetes:
+    injector:
+      spire:
+        enabled: true | false 
+        path: /run/spire/sockets
+        mountPath: /run/spire/sockets
+```
+
+Annotations always works, `enabled: true | false` take action when there is no annotation.
```
lahabana: You don't say anywhere anymore what the annotation is.

lahabana: You do later but feels like it should be done before.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+```
+
+* mount volume
+```yaml
+volumeMounts:
+  - name: kuma-spire-agent-socket
+    mountPath: /run/spire/sockets
+    readOnly: true
+```
+* and inject annotation
```
lahabana: Why does it do this? I feel like it doesn't have to no?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+
+```yaml
+kuma.io/spire-integration: enabled
+```
+
+Purpose of the Label:
```
lahabana: there's no label.

docs/madr/decisions/073-spiffe-compliance.md
bartsmykla: Nit: Wdyt?
```suggestion
The control plane uses the default CA bundle to validate the provided CA. Right now, we don't set it explicitly, but we can do so by setting `KUMA_GENERAL_TLS_CA_BUNDLE_FILE`.
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+              op: add
+              value:
+                name: spire-agent-socket
+                mountPath: /run/spire/sockets/
+                readOnly: true          
+```
+
+#### Define active identity provider
+
+Previously, defining the active identity for a group of services was done through a single, perhaps less flexible, mechanism. Now, the `MeshIdentity` resource, leveraging its `selector`, offers a more granular and controlled approach to activate specific configurations.
```
lahabana:
```suggestion
Previously, selecting the active identity for a group of services was done through a single, perhaps less flexible, mechanism. Now, the `MeshIdentity` resource, leveraging its `selector`, offers a more granular and controlled approach to activate specific configurations.
```

docs/madr/decisions/073-spiffe-compliance.md
bartsmykla: it should be other way around as this should be set to `false` by default so I would explain what will be the result of setting it to `true`

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+                  type: Directory
+            - path: spec.template.spec.containers.kuma-sidecar.volumeMounts.spire-agent-socket
+              op: add
+              value:
+                name: spire-agent-socket
+                mountPath: /run/spire/sockets/
+                readOnly: true          
+```
+
+#### Define active identity provider
```
lahabana:
```suggestion
#### Selecting active identity provider
```

docs/madr/decisions/073-spiffe-compliance.md
bartsmykla: Nit: Wdyt?
```suggestion
We will set the `k8s.kuma.io/service-account` annotation with the service account name. Since the value can be up to [253 characters](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-multiple-service-accounts), it must go into an annotation rather than a label. This allows us to resolve the SPIFFE ID.
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+  selector:
+    dataplane:
+      matchLabels:
+        app: demo-app
+```
+
+The `MeshIdentity` resource’s selector field allows selection based on dataplane labels. Users can define it to apply to either all dataplanes or a specific subset.
+If there are multiple (X) identities selecting the same dataplane, the control plane will resolve the conflict by choosing one based on the following order of precedence:
+
+1. Selector specificity – the identity with most specific selector is preferred.
```
lahabana: This should be more rigourous on what specific mean. I think it's by number of keys in the selector.

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+
+**ServiceAccount** 
+We are going to set `k8s.kuma.io/service-account` with the name of service account. The length of the possible value is [253 characters](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-multiple-service-accounts) which makes us to put it into the annotation and not a label. That let us resolve SPIFFEID.
+
+#### Provider types
+
+There can be multiple difference identity providers but in the first iteration we would like to focus on:
+
+* bundled
+* spire
```
bartsmykla: Nit:
```suggestion
* Bundled
* Spire
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+apiVersion: kuma.io/v1alpha1
+kind: MeshIdentity
+...
+spec:
+  selector:
+    dataplane:
+      matchLabels: {}
+```
+
+3. Select nothing
```
lahabana, line 521: I select not more of less specific than select all dp?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+* Environment variable,
+* File path,
+* Secret reference,
+* Auto generated and stored in a Secret
+
+This gives flexibility for organizations that already have an established PKI or need to integrate with existing certificate authorities.
+
+```yaml
+    bundled: # to extend in KM
+      trustExtraction: Disabled | Enabled # allows to disable extraction of CA into MeshTrust
```
bartsmykla:
```suggestion
      meshTrustCreation: Disabled | Enabled # allows to disable extraction of CA into MeshTrust
```

docs/madr/decisions/073-spiffe-compliance.md
bartsmykla: That's true

docs/madr/decisions/073-spiffe-compliance.md
bartsmykla: I know what it means but without context it seems a little bit vague, could you extend this paragraph?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+    dataplane:
+      matchLabels:
+        app: demo-app
+```
+
+The `MeshIdentity` resource’s selector field allows selection based on dataplane labels. Users can define it to apply to either all dataplanes or a specific subset.
+If there are multiple (X) identities selecting the same dataplane, the control plane will resolve the conflict by choosing one based on the following order of precedence:
+
+1. Selector specificity – the identity with most specific selector is preferred.
+2. Lexicographical order – if selector sizes are equal, the identity with the lexicographically smallest name is chosen.
```
bartsmykla:
```suggestion
1. Selector specificity - the identity with most specific selector is preferred.
2. Lexicographical order - if selector sizes are equal, the identity with the lexicographically smallest name is chosen.
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+spec:
+  selector:
+    dataplane:
+      matchLabels:
+        app: demo-app
+  spiffeID: # optional
+    trustDomain: "prod.zone-1.kuma.io"
+    path: "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}"
+  bundled: # to extend in KM
+    trustExtraction: Disabled | Enabled # allows to disable extraction of CA into MeshTrust
```
bartsmykla:
```suggestion
    meshTrustCreation: Disabled | Enabled # allows to disable extraction of CA into MeshTrust
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+func LabelSelectorAsSelector(ps *LabelSelector) (labels.Selector, error) {
+	if ps == nil {
+		return labels.Nothing(), nil
+	}
+	if len(ps.MatchLabels)+len(ps.MatchExpressions) == 0 {
+		return labels.Everything(), nil
+	}
+```
+
+#### What would the process of switching backends look like?
```
lahabana: what's the value of keeping the same fields? They mention `backends` which are not things that exist anymore in this new world

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+
+We want to use a [KRI](070-resource-identifier.md) for resource naming. Since `MeshIdentity` generates secrets for each dataplane, it should be named based on the KRI to simplify correlation between the Envoy `Secret` and the `MeshIdentity` configuration.
+
+Proposed name:
+
+```yaml
+kri_mid_mesh-1_us-east-2_kuma-system_identity-1
+```
+
+Also, we need to name the Spire agent cluster, in this case is it internal or maybe based on MeshIdentity:
```
lahabana:
```suggestion
Also, we need to name the Spire agent cluster, in this case it is internal or maybe based on MeshIdentity:
```

lahabana: I'm confused this say A or B I only see 1 option after...

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+Proposed name:
+```yaml
+system_mid_sds-spire-agent
+```
+
+> [!WARNING]
+> The naming is different when using SPIRE, since SPIRE uses SPIFFEID or `default` as a certificate name
```
lahabana:
```suggestion
> The naming of secrets is different when using SPIRE, since SPIRE uses SPIFFEID or `default` as a certificate name
```

lahabana (approved):

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,921 @@ Future releases
...
+metadata:
+  name: trust-hash
+  namespace: kuma-system # only in system namespace
+  labels:
+    kuma.io/mesh: default
+spec:
+  origin:
+    kri: kri_mid_mesh-1_us-east-2_kuma-system_identity-1
+  caBundles:
+    - caBundle: # PEM
```
lahabana: I think this should potentially be pluggable. inline is ok but ideally I should be able to have vault here.

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: This selector selects no dataplane. Based on K8s selectors it seems we should select nothing

docs/madr/decisions/073-spiffe-compliance.md
slonka: I'm leaning towards `.mesh.local`

docs/madr/decisions/073-spiffe-compliance.md
slonka: I think it's coming from https://github.com/kumahq/kuma/blob/b42f6da0df63b1cc7c8bc3d7f4f0c051119e3883/UPGRADE.md#L162 / https://github.com/kumahq/kuma/issues/12653

docs/madr/decisions/073-spiffe-compliance.md
slonka: it's already a restriction - right? https://github.com/kumahq/kuma/issues/12653 or do you mean combined together?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -242,45 +245,55 @@ metadata:
...
-      privateKey:
-        # one of
-        secret:
-        path:
-        envVar:
-      dataplaneCertificate:
-        duration: 24h
+    bundled: # to extend in KM
+      meshTrustCreation: Disabled | Enabled # allows to disable extraction of CA into MeshTrust
+      insecureAllowSelfSigned: true
```
slonka, line 261: Does it make sense to do an enum? 🤔 

`insecureSelfSigned: allowed | disallowed`

docs/madr/decisions/073-spiffe-compliance.md
slonka: I think I like this option better.

docs/madr/decisions/073-spiffe-compliance.md
slonka: nitpick: you've added two dots there Charly - just edited

slonka: do you mean by default it should be self signed?

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: With boolean it's easier to just set default `false` once not defined, also there is no more benefit since we won't add additional values in the future

docs/madr/decisions/073-spiffe-compliance.md
slonka: https://github.com/kumahq/kuma/pull/13804/files/97ad7e8f0a941f7474794cabfdd0ec0e38ae400a..3105b5c47d29a340d39781c3115ccf85b0e502af#diff-2624c543bb2b588a32c9e2efc57fac0c007db25634c6160be4247472f13c484cR259 this is how it's called up there

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: I think covered by this: https://github.com/kumahq/kuma/issues/13054

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Default `insecureAllowSelfSigned : false`

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: https://github.com/kumahq/kuma/issues/13054

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: There should be no issue with changing the name since it's SPIFFE is elastic about it

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,919 @@ Future releases
...
+mTLS:
+  certificateExpirationTime: "2025-06-03T17:35:33Z"
+  certificateRegenerations: 1
+  issuedBackend: kri_mid_mesh-1_us-east-2_kuma-system_identity-1
+  lastCertificateRegeneration: "2025-06-02T17:35:33.624189876Z"
+  supportedBackends: # list of MeshTrust name
+  - kri_mtrust_mesh-1_us-east-2_kuma-system_identity-1
+...
+```
+1. User creates MeshTrust with CA of a new MeshIdentity `identity-2`
```
slonka: nitpick: 

```suggestion
2. User creates MeshTrust with CA of a new MeshIdentity `identity-2`
```

slonka: image

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,919 @@ Future releases
...
+        # OneOf
+        secret:
+        file:
+        envVar:
+      privateKey:
+        type: File | Secret | EnvVar
+        # OneOf
+        secret:
+        file:
+        envVar:
```
slonka: do we need this if we have "autogenerate: enabled: true"?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,919 @@ Future releases
...
+  name: allow-all
+  namespace: kuma-demo
+  labels:
+    kuma.io/mesh: default
+spec:
+  match:
+    rules:
+    ....
+```
+4. Enable MeshIdentity for all service
```
slonka: nitpick

```suggestion
4. Enable MeshIdentity for all services
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,919 @@ Future releases
...
+  labels:
+    kuma.io/mesh: default
+spec:
+  from:
+  - targetRef:
+      kind: Mesh
+    default:
+      action: Allow
+```
+or more specific but it's not a part of this MADR
```
slonka: Is it going to be defined in https://github.com/kumahq/kuma/pull/13848 ?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,919 @@ Future releases
...
+  labels:
+    kuma.io/mesh: default
+spec:
+  selector:
+    dataplane:
+      matchLabels: {}
+  spiffeID: # optional
+    trustDomain: "prod.zone-1.mesh.local"
+    path: "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}"
+  provider:
```
slonka, line 883: isn't it bundled now?

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,919 @@ Future releases
...
+      privateKey:
+        # one of
+        secret:
+        inline:
+        inlineString:
+      dataplaneCertificate:
+        duration: 24h
+```
+5. Check if works by verifing DataplaneInsights
+6. Wait for propagation of values, we caa verify this in `MeshInsights`
```
slonka: nitpick:

```suggestion
6. Wait for propagation of values, we can verify this in `MeshInsights`
```

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,919 @@ Future releases
...
+7. User remove mTLS section from Mesh object once all dataplanes uses new identity
+
+**Does Envoy supports our certs?**
+
+Yes, if we set `trusted_ca` envoy doesn't do validation of any SAN, hash, etc. https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/ssl#example-configuration
+
+> If only trusted_ca is specified, Envoy will verify the certificate chain of the presented certificate, but not its subject name, hash, etc. Other validation context configuration is typically required depending on the deployment.
+
+SPIFFE validator is experimental but SPIRE uses it. It sets a `custom_validator_config` which allows separation of each bundle to separate trust domain:
```
slonka, line 916: What is istio using?

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: In this case it's optional

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Yes

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: It's provider with `type: Bundled`

docs/madr/decisions/073-spiffe-compliance.md
lukidzi: Istio uses `trusted_bundle`

docs/madr/decisions/073-spiffe-compliance.md
bartsmykla: that makes sense, can we point to this issue in the MADR so the reader can understand where this limit comes from?

slonka (approved):

slonka (dismissed):

docs/madr/decisions/073-spiffe-compliance.md
```diff
@@ -153,7 +157,937 @@ Future releases
...
+      reason: # Collision | Generated | TemplateError
+```
+
+The control plane uses the default CA bundle to validate the provided CA. By default, it relies on the system CA bundle, as no bundle is set explicitly. However, the user can override this by setting the `KUMA_MESHIDENTITY_TLS_CA_BUNDLE_FILE` environment variable, which defaults to the value of `KUMA_GENERAL_CA_BUNDLE_FILE`.
+
+`insecureAllowSelfSigned` makes using self signed CA an explicit opt-in. This is used with both provided by the user or autogenrated CAs.
+
+**Potential collision**
+
+In scenarios where a user creates a `MeshIdentities` resource that specifies a trustDomain already used by another `MeshIdentity` or legacy trust domain (Mesh name), a conflict may arise. Because it's not always possible to detect this conflict during resource creation, we propose adding a `status` field to `MeshIdentities`. This field will be used to report post-creation validation issues, such as trust domain conflicts, making them visible to the user.
```
lobkovilya: nit: I don't think it's possible at all
```suggestion
In scenarios where a user creates a `MeshIdentities` resource that specifies a trustDomain already used by another `MeshIdentity` or legacy trust domain (Mesh name), a conflict may arise. Because it's not possible to detect this conflict during resource creation, we propose adding a `status` field to `MeshIdentities`. This field will be used to report post-creation validation issues, such as trust domain conflicts, making them visible to the user.
```

lobkovilya: Please also mention if we're going to sync `status` over KDS, I believe we'll sync?

lobkovilya (approved):
LGTM!

PR #13809, opened 2025-06-26
ci(lint): enable usestdlibvars linter

mmorel-35:
## Motivation

A linter that detect the possibility to use variables/constants from the Go standard library.

## Supporting documentation

https://golangci-lint.run/usage/linters/#usestdlibvars

> Changelog: skip


.golangci.yml
```diff
@@ -138,6 +139,8 @@ linters:
         - cancelled
     nonamedreturns:
       report-error-in-defer: false
+    usestdlibvars:
+      http-status-code: false
```
lobkovilya, line 143: why opting out for status code checks?

mmorel-35: It's activated by default, and I prefer to activate the rules progressively so if you don't like the linter at all, si did not spent to much time. If you take a look are several other rules that can be activated for this linter

PR #13810, opened 2025-06-27
ci(lint): enable exptostd linter

mmorel-35:
## Motivation

Detects functions from golang.org/x/exp/ that can be replaced by std functions.

## Supporting documentation

https://golangci-lint.run/usage/linters/#exptostd

> Changelog: skip

### DIFF
--- a/pkg/transparentproxy/config/dataplane/config_dataplane.go
+++ b/pkg/transparentproxy/config/dataplane/config_dataplane.go
@@ -2,7 +2,7 @@ package dataplane
 
 import (
 	"github.com/asaskevich/govalidator"
-	"golang.org/x/exp/constraints"
+	"golang.org/x/exp/constraints" //nolint:exptostd // https://github.com/kumahq/kuma/pull/13810/files#r2179434552
 
 	core_config "github.com/kumahq/kuma/pkg/config"
 	xds_types "github.com/kumahq/kuma/pkg/core/xds/types"
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,7 @@ linters:
     - contextcheck
     - depguard
     - errcheck
+    - exptostd
     - ginkgolinter
     - gocritic
     - gomodguard
--- a/pkg/api-server/inspect_mesh_service.go
+++ b/pkg/api-server/inspect_mesh_service.go
@@ -2,11 +2,11 @@ package api_server
 
 import (
 	"fmt"
+	"maps"
 	"sort"
 	"strings"
 
 	"github.com/emicklei/go-restful/v3"
-	"golang.org/x/exp/maps"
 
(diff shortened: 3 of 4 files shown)


pkg/transparentproxy/config/dataplane/config_dataplane.go
```diff
@@ -11,7 +10,7 @@ import (
 )
 
 type PortLike interface {
-	constraints.Integer | constraints.Float | tproxy_config.Port
+	~int | ~int8 | ~int16 | ~int32 | ~int64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr | ~float32 | ~float64
```
lobkovilya: Strictly saying that's not a replacement from std :)

lobkovilya (changes requested):
Even though I like the idea of having such linter in the project, it seems like https://github.com/ldez/exptostd is not widely adopted yet. Basically it can detect only `golang.org/x/exp/maps` and `golang.org/x/exp/slices` packages.

But I think we need an issue to go over `golang.org/x/exp/...` imports and see what can be replaced with std packages as we haven't revisited for a while.

@slonka as you're the second reviewer, what do you think?

pkg/transparentproxy/config/dataplane/config_dataplane.go
mmorel-35: tproxy_config.Port was conflicted with it's native value so I only kept natives . Was it inappropriate?

lobkovilya: No it's not about `tproxy_config.Port`. What I meant is `constraints.Integer` and `constraints.Float` allows to avoid explicit type enumeration. The idea of the linter is to replace `golang.org/x/exp/constraints` with std alternative, but simple type enumeration technically is not an alternative.

mmorel-35: I understand, maybe I didn't look to the right place but I haven't found a correct replacement in the recommended new package. Maybe you can find it ?

slonka:
@lobkovilya for excluding `exp` we can use `gomodguard` and configure it with:
- golang.org/x/exp/maps
- golang.org/x/exp/slices
- golang.org/x/exp/constraints

or implement / vote for implementing https://github.com/ryancurrah/gomodguard/issues/47

> imports and see what can be replaced with std packages as we haven't revisited for a while

I see it only used in 2 places:
- https://github.com/kumahq/kuma/blob/ad81fbf4367ea87c4e4953b70cfce349530bdccf/pkg/api-server/inspect_mesh_service.go#L175
- https://github.com/kumahq/kuma/blob/4479a75c816f16645d48e8d21336756783e57803/pkg/transparentproxy/config/dataplane/config_dataplane.go#L14

PR #13811, opened 2025-06-27
ci(lint): enable var-declaration rule from revive

mmorel-35:
## Motivation

This rule proposes simplifications of variable declarations.

## Implementation information

https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md#var-declaration

> Changelog: skip


lobkovilya (dismissed):
LGTM

PR #13812, opened 2025-06-27
ci(lint): clean up golangci-lint configuration

mmorel-35:
## Motivation

Clean up golangci-lint configuration

## Implementation information

Remove unused exclusions. Also remove legacy preset to list them explicitly in their dedicated linters config so they can be identified clearly.

## Supporting documentation

> Changelog: skip

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -97,7 +97,9 @@ linters:
                 - os
     gosec:
       excludes:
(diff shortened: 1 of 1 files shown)


slonka (approved):
As far as I can see we're not loosening any checks so 👍

PR #13815, opened 2025-06-27
ci(linter): fix linter for meshservice

slonka:
## Motivation

Because of https://github.com/kumahq/kuma/issues/12894 we excluded MeshService (there was a problem with typealiased pointer to a struct).

## Implementation information

Adjust MeshService Resouce and HostnameGenerator

## Supporting documentation

Closes https://github.com/kumahq/kuma/issues/12894


pkg/test/store/store_test_template.go
```diff
@@ -119,17 +120,18 @@ func ExecuteStoreTests(
...
 						},
 						Ports: []meshservice_api.Port{
 							{
 								Port:        80,
-								TargetPort:  intstr.FromInt(80),
+								TargetPort:  pointer.To(intstr.FromInt(80)),
 								AppProtocol: "http",
 							},
 						},
+						State: meshservice_api.StateUnavailable,
```
slonka, line 134: @lobkovilya - will this be a problem? 🤔 the fact that I needed to manually add that, it shouln't be, right?

bartsmykla (approved):

pkg/core/resources/apis/hostnamegenerator/api/v1alpha1/hostnamegenerator.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 type LabelSelector struct {
-	MatchLabels map[string]string `json:"matchLabels,omitempty"`
+	MatchLabels map[string]string `json:"matchLabels"`
```
lobkovilya: It was possible to create HG without `MatchLabels`. You need to check if previously created HG can be returned with `matchedLabels: null`. If that's the case then they won't pass OAPI validation

bartsmykla (dismissed):

slonka:
There seems to be more problems with removing `omitempty`. When it's a struct - removing it seems to be fine but when it's an array we get:

```
spec.ports in body must be of type array
```

on the global. I'm closing this and opening https://github.com/kumahq/kuma/pull/13914 to exclude everything non matching in `meshservice` and we can revisit this PR when we need make `MeshService` pass the validation (probably when people request using TF provider on the zone CPs).

PR #13816, opened 2025-06-27
ci(linter): add linter rule for int32

slonka:
## Motivation

During knowledge share we talked about k8s guidelines for APIs. We decided as part of test friday write a linter to check for it.

## Implementation information

Add check for int types.


bartsmykla (approved):

slonka:
@Automaat please pick this up when you have time

PR #13819, opened 2025-06-27
ci(skaffold): simplify skaffold handling

lahabana:
When added in #13698 we relied on modifying our tags and making things work for skaffold Here we make skaffold work for us by making skaffold use our tag format.

We also stop embedding the demo and have a `make dev/fetch-demo` target instead

Finally use `docker/tag/kuma-cp` to remove the arch prefix from the local tags. This is a good way to go about this because this is what's already done in e2e tests: https://github.com/kumahq/kuma/blob/8dc904d4a84bc8304784866501721e5153757b09/mk/k3d.mk#L219-L225


mk/docker.mk
```diff
@@ -7,7 +7,7 @@ DOCKER_USERNAME ?=
 DOCKER_API_KEY ?=
 
 define build_image
-$(addsuffix :$(BUILD_INFO_VERSION)$(if $(and $(2),$(if $(findstring true,$(SKIP_PLATFORM)),,1)),-$(2)),$(addprefix $(DOCKER_REGISTRY)/,$(1)))
+$(addsuffix :$(BUILD_INFO_VERSION)$(if $(2),-$(2)),$(addprefix $(DOCKER_REGISTRY)/,$(1)))
```
lahabana, line 10: These should be a revert to what it was before https://github.com/kumahq/kuma/pull/13698

mk/dev.mk
```diff
@@ -2,7 +2,7 @@ KUMA_DIR ?= .
 TOOLS_DIR = $(KUMA_DIR)/tools
 # Important to use `:=` to only run the script once per make invocation!
 BUILD_INFO := $(shell $(TOOLS_DIR)/releases/version.sh)
-BUILD_INFO_VERSION ?= $(word 1, $(BUILD_INFO))
+BUILD_INFO_VERSION = $(word 1, $(BUILD_INFO))
```
lahabana, line 5: These should be a revert to what it was before https://github.com/kumahq/kuma/pull/13698

skaffold.yaml
```diff
@@ -21,26 +24,23 @@ build:
       custom:
         buildCommand: |
           platform=$(echo $PLATFORMS | cut -d'/' -f2)
-          image_tag=$(echo $IMAGE | cut -d':' -f2)
-          SKIP_PLATFORM=true BUILD_INFO_VERSION=$image_tag make image/kuma-dp/$platform
+          make image/kuma-dp/$platform docker/tag/kuma-cp
```
Automaat:
```suggestion
          make image/kuma-dp/$platform docker/tag/kuma-dp
```

Automaat (approved):

Automaat (approved):

Issue #11770, opened 2024-10-14
Consider mise for internal tooling dependency management


slonka:
### Description

Right now our hand written script with checksums work but we might want to consider https://github.com/jdx/mise for these things.

lukidzi:
> Triage: We can do it, seems like it's easy to switch between versions and might improve new user experience

Automaat:
So I tested it here: https://github.com/kumahq/kuma/pull/13835

I found two issues with it now:
- Install clang-format can take up to 20 minutes. We could outsource this to separate action, like golangci-lint https://github.com/jidicula/clang-format-action  Also mise cache is pretty good so it will be only reinstalled after updating `.mise.toml` file
- We need to move getting proto dependencies from installing go deps script: https://github.com/kumahq/kuma/blob/f6f367f8970eb351ef42593d894159c90ff579c8/mk/dependencies/go-deps.sh#L54-L78 to separate makefile targer

PR #13835, opened 2025-07-01
feat(tools): move to Mise from our custom scripts

Automaat:
## Motivation

Introducing [Mise](https://mise.jdx.dev) to Kuma. We already use Mise in projects like `kuma-website` or `kuma-counter-demo`. 
- With mise we can streamline tools installation for different Kuma versions, which replaces our custom scripts and `make dev/tools` target. 
- While switching branches for example from release to master mise will automatically enable specified versions of tools which removes need of reinstalling everything always and reduces mistakes while working with older branches
- We can now easily update tools version with [Renovate](https://docs.renovatebot.com/modules/manager/mise/) which was now not possible, and we almost never updated them
- Mise action cache works much better than our custom cache with `make dev/tools` which will speed up CI a little bit

### Kong Mesh or other projects using Kuma compatibility
Right now, we don't have any extra dependencies than we have in Kuma but adding them would be easy as you can have multiple config files that mise will merge: https://mise.jdx.dev/configuration.html#configuration

### Concerns
- There is one slight issue with this now, installation of clang-format can take up to 17 minutes eating all the time available for check action. We can address this by running https://github.com/jidicula/clang-format-action as we do with `golang-ci lint`

Fix: https://github.com/kumahq/kuma/issues/11770


mk/dev.mk
```diff
@@ -38,32 +38,46 @@ endef
...
-SHELLCHECK=$(CI_TOOLS_BIN_DIR)/shellcheck
-CONTAINER_STRUCTURE_TEST=$(CI_TOOLS_BIN_DIR)/container-structure-test
+ifeq (,$(shell which mise))
+$(error "mise - https://github.com/jdx/mise - not found. Please install it.")
+endif
+MISE := $(shell which mise)
+
+PROTOS_DEPS_PATH=$(shell $(MISE) where protoc)/include
+
+XDS_VERSION=$(shell go list -f '{{ .Version }}' -m github.com/cncf/xds/go)
```
Automaat, line 52: this probably can be done cleaner

mk/dev.mk
```diff
@@ -38,32 +38,46 @@ endef
...
+PROTO_ENVOY=$(shell go mod download github.com/envoyproxy/data-plane-api@main && go list -f '{{ .Dir }}' -m github.com/envoyproxy/data-plane-api@main)
+
+CLANG_FORMAT=$(MISE) x clang-format -- clang-format
+YQ=$(shell mise which yq)
+HELM=$(MISE) x helm -- helm
+K3D_BIN=$(MISE) x k3d -- k3d
+KIND=$(MISE) x kind -- kind
+KUBEBUILDER=$(MISE) x kubebuilder -- kubebuilder
+# TODO this should be automated with max k8s version
+KUBEBUILDER_ASSETS_VERSION=1.32.0
```
Automaat: how to properly store this version? This should depend on our max k8s version but not all versions minor versions of k8s are supported

mk/dev.mk
```diff
@@ -38,32 +38,46 @@ endef
...
-PROTOC_GEN_VALIDATE=$(CI_TOOLS_BIN_DIR)/protoc-gen-validate
-PROTOC_GEN_KUMADOC=$(CI_TOOLS_BIN_DIR)/protoc-gen-kumadoc
-PROTOC_GEN_JSONSCHEMA=$(CI_TOOLS_BIN_DIR)/protoc-gen-jsonschema
-GINKGO=$(CI_TOOLS_BIN_DIR)/ginkgo
-GOLANGCI_LINT=$(CI_TOOLS_BIN_DIR)/golangci-lint
-HELM_DOCS=$(CI_TOOLS_BIN_DIR)/helm-docs
-KUBE_LINTER=$(CI_TOOLS_BIN_DIR)/kube-linter
-HADOLINT=$(CI_TOOLS_BIN_DIR)/hadolint
-OAPI_CODEGEN=$(CI_TOOLS_BIN_DIR)/oapi-codegen
+PROTOC_GEN_GO=$(shell mise which protoc-gen-go)
```
Automaat: In other projects we do 
```
$(MISE) x shellcheck -- shellcheck
```
but in Kuma some targets need absolute path to binary that is why i needed to use
```
mise which
```

Should we move all to use `mise which`?

api/mesh/v1alpha1/gateway_route.proto
```diff
@@ -42,11 +42,10 @@ message MeshGatewayRoute {
     message Rule {
       // repeated Match matches = 1;
 
-      repeated Backend backends = 2
-          [ (validate.rules).repeated .min_items = 1 ];
+      repeated Backend backends = 2 [ (validate.rules).repeated.min_items = 1 ];
```
Automaat, line 45: these changes to proto files are result of updating our prehistoric version of `clang-format`

mk/dev.mk
lahabana: mise which is probs best

mk/dev.mk
lahabana: My guess is always go to `.0` patch. I doubt these change much inside a k8s version.

mk/dev.mk
```diff
@@ -38,32 +38,46 @@ endef
...
+
+CLANG_FORMAT=$(MISE) x clang-format -- clang-format
+YQ=$(shell mise which yq)
+HELM=$(MISE) x helm -- helm
+K3D_BIN=$(MISE) x k3d -- k3d
+KIND=$(MISE) x kind -- kind
+KUBEBUILDER=$(MISE) x kubebuilder -- kubebuilder
+# TODO this should be automated with max k8s version
+KUBEBUILDER_ASSETS_VERSION=1.32.0
+KUBEBUILDER_ASSETS=$(shell setup-envtest use $(KUBEBUILDER_ASSETS_VERSION) --bin-dir $(CI_TOOLS_BIN_DIR) -p path)
```
lahabana: where does this setup-envtest come from? Should it be a mise which?

mk/dev.mk
```diff
@@ -38,32 +38,46 @@ endef
...
-CONTAINER_STRUCTURE_TEST=$(CI_TOOLS_BIN_DIR)/container-structure-test
+ifeq (,$(shell which mise))
+$(error "mise - https://github.com/jdx/mise - not found. Please install it.")
+endif
+MISE := $(shell which mise)
+
+PROTOS_DEPS_PATH=$(shell $(MISE) where protoc)/include
+
+XDS_VERSION=$(shell go list -f '{{ .Version }}' -m github.com/cncf/xds/go)
+PROTO_XDS=$(shell go mod download github.com/cncf/xds@$(XDS_VERSION) && go list -f '{{ .Dir }}' -m github.com/cncf/xds@$(XDS_VERSION))
```
lahabana, line 53: I'm not a fan of these because if we don't have internet this fails.

I guess these can't work with a mise tool? Why?
Would it make sense to write a mise plugin to do this?

lahabana (commented):
Go progress on this!

mk/dev.mk
Automaat: changed to use `mise which`

mk/dev.mk
Automaat: changed to `mise which`

.github/workflows/pr-comments.yaml
```diff
@@ -54,8 +54,7 @@ jobs:
           key: ${{ runner.os }}-${{ runner.arch }}-devtools-${{ hashFiles('mk/dependencies/deps.lock') }}
           restore-keys: |
             ${{ runner.os }}-${{ runner.arch }}-devtools
-      - run: |
```
lahabana, line 57: I think you forgot to remove the cache above

.github/workflows/transparentproxy-tests.yaml
```diff
@@ -26,8 +26,7 @@ jobs:
           key: ${{ runner.os }}-${{ runner.arch }}-devtools-${{ hashFiles('mk/dependencies/deps.lock') }}
           restore-keys: |
             ${{ runner.os }}-${{ runner.arch }}-devtools
-      - run: |
-          make dev/tools
+      - uses: jdx/mise-action@13abe502c30c1559a5c37dff303831bab82c9402 # v2.2.3
```
lahabana, line 22: same re cache

mk/dev.mk
```diff
@@ -38,32 +38,46 @@ endef
...
+PGV_VERSION=$(shell go list -f '{{.Version}}' -m github.com/envoyproxy/protoc-gen-validate)
+PROT_PGV=$(shell go mod download github.com/envoyproxy/protoc-gen-validate@$(PGV_VERSION) && go list -f '{{ .Dir }}' -m github.com/envoyproxy/protoc-gen-validate@$(PGV_VERSION))
```
lahabana:
```suggestion
PROTO_PGV=$(shell go mod download github.com/envoyproxy/protoc-gen-validate@$(PGV_VERSION) && go list -f '{{ .Dir }}' -m github.com/envoyproxy/protoc-gen-validate@$(PGV_VERSION))
```

mk/api.mk
```diff
@@ -1,4 +1,8 @@
 PROTOC := $(PROTOC_BIN) \
+	--proto_path=$(PROTO_GOOGLE_APIS) \
+	--proto_path=$(PROTO_XDS) \
+	--proto_path=$(PROT_PGV) \
```
lahabana: PROTO_PGV

lahabana (dismissed):
Let's go with this but also let's find a solution for the deps...

mk/dev.mk
Automaat: Right now, mise does not have support for this. For go mise can only do `go install`, and we need to do `go mod download`

Looking at the mise plugins. I think we would need to create a custom plugin for every proto dependency. If I understand correctly, one mise plugin should install one dependency. 

I am not sure if there is an easy way of depending on protos. I see that people most of the time are just downloading this schema manually: https://github.com/search?q=github.com%2Fenvoyproxy%2Fdata-plane-api&type=code&p=1

.github/workflows/pr-comments.yaml
Automaat: removed cache

.github/workflows/transparentproxy-tests.yaml
Automaat: removed cache

mk/dev.mk
Automaat: changed

mk/api.mk
Automaat: changed

mk/dev.mk
```diff
@@ -38,32 +38,46 @@ endef
...
+CLANG_FORMAT=$(shell $(MISE) which clang-format)
+YQ=$(shell $(MISE) which yq)
+HELM=$(shell $(MISE) which helm)
+K3D_BIN=$(shell $(MISE) which k3d)
+KIND=$(shell $(MISE) which kind)
+KUBEBUILDER=$(shell $(MISE) which kubebuilder)
+# TODO this should be automated with max k8s version
+KUBEBUILDER_ASSETS_VERSION=1.32.0
+SETUP_ENVTEST=$(shell $(MISE) which setup-envtest)
+KUBEBUILDER_ASSETS=$(shell $(SETUP_ENVTEST) use $(KUBEBUILDER_ASSETS_VERSION) --bin-dir $(CI_TOOLS_BIN_DIR) -p path)
```
Automaat, line 65: `CI_TOOLS_BIN` still relies on kuma version:
```
CI_TOOLS_DIR ?= ${HOME}/.kuma-dev/${PROJECT_NAME}-${CI_TOOLS_VERSION}
```

Should we change it to something like:
```
CI_TOOLS_DIR ?= ${HOME}/.kuma-dev/${PROJECT_NAME}
```
?

mk/dev.mk
Automaat: I moved this just under envs with K8S versions for now. It will be easier to remember about this when updating k8s version in Kuma

mk/dev.mk
Automaat: I think that beside this place `CI_TOOLS_DIR` and `CI_TOOLS_BIN_DIR` should not be used

mk/dev.mk
lahabana: Can we open an issue to track this and refer to this comment?

mk/dev.mk
lahabana: can't we use something from `mise which` ?

mk/dev.mk
Automaat: https://github.com/kumahq/kuma/issues/13880

mk/dev.mk
Automaat: moved to mise dirs

PR #13853, opened 2025-07-04
fix(dynconfig): change the direct response body size dynamically

### LINKED ISSUE #13852: kuma listener dynamic config exceeded the direct response maxinum size
### Kuma Version

master and release 2.11.*

### Describe the bug

Since we introduced the embedded DNS server in 2.11.* to replace coreDNS, we stored the DNS data in the xds listener dynamic config to return it directly. However, once if your data is very big then it would exceed the default maximum size  4096 bytes and get the envoy errors.

### To Reproduce

In 2.11.*, build a large mesh cluster and use the embedded DNS server.

### Expected behavior

No errors occurred

Here's the error logs from envoy:

```text
2025-07-03T08:46:52.548Z  INFO  xds.nack-backoff  config was previously rejected by Envoy. Applying backoff before resending it {"backoff": "5s", "nodeID": "default.2048-app-85c76944c7-jprqh.kuma-demo", "reason": "Error adding/updating listener(s) _kuma:dynamicconfig: response body size is 4261 bytes; maximum is 4096\n"}
```

[max_direct_response_body_size_bytes](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route.proto) property in HTTP route configuration.

Icarus9913:
## Motivation

Since we introduced embedded DNS in `2.11.*`, the DNS data would be stored in the direct response. However, the default max direct response body size is 4 KiB, and if you exceed it, the envoy would return an error. So, We need to change the data size.

## Implementation information

In the CP side, we calculate the size and set it in the XDS configuration dynamically.

## Supporting documentation

Fix #13852


pkg/xds/dynconf/listener.go
```diff
@@ -105,6 +107,12 @@ func AddConfigRoute(proxy *core_xds.Proxy, rs *core_xds.ResourceSet, name string
 				},
 			},
 		)
+		// set the MaxDirectResponseBodySizeBytes value dynamically
+		if routeConfig.MaxDirectResponseBodySizeBytes == nil {
+			routeConfig.MaxDirectResponseBodySizeBytes = wrapperspb.UInt32(uint32(len(bytes)) + maxDirectResponseBodyPadding)
```
lahabana: What's the padding for? It would be useful to add a comment about this because it's quite cryptic to me :)

pkg/xds/dynconf/listener.go
```diff
@@ -105,6 +107,12 @@ func AddConfigRoute(proxy *core_xds.Proxy, rs *core_xds.ResourceSet, name string
...
+		} else {
+			routeConfig.MaxDirectResponseBodySizeBytes = wrapperspb.UInt32(uint32(len(bytes)) + routeConfig.MaxDirectResponseBodySizeBytes.GetValue())
```
lahabana: This needs to be check but AFAIU this is per response so each endpoint accounts individually. So here you don't want to maxDirectReponse to be the sum of all. You instead want the size to be the max of all no?

```suggestion
			routeConfig.MaxDirectResponseBodySizeBytes = wrapperspb.UInt32(math.Max(uint32(len(bytes) + padding), routeConfig.MaxDirectResponseBodySizeBytes.GetValue())
```

pkg/xds/dynconf/listener.go
```diff
@@ -18,8 +19,9 @@ import (
 )
 
 const (
-	Origin       = "dynamic-config"
-	ListenerName = "_kuma:dynamicconfig"
+	Origin                       = "dynamic-config"
+	ListenerName                 = "_kuma:dynamicconfig"
+	maxDirectResponseBodyPadding = 256
```
lahabana: nit: but I think this sort of global is just making reading harder. You may as well just define this as a variable in the code, the compiler will simplify the variable out and use a constant anyway.

lahabana (approved):

lahabana (changes requested):
Didn't mean to approve

pkg/xds/dynconf/listener.go
Icarus9913: Here's an example of the `route_config`, you can see the current implementation would combine the `DNS` and `MeshMetrics` data together for the route config. So I think we should use the sum right?

```text
"route_config": {
  "virtual_hosts": [
    {
      "name": "_kuma:dynamicconfig",
      "domains": [
        "*"
      ],
      "routes": [
        {
          "match": {
            "path": "/dns",
            "headers": [
              {
                "name": "If-None-Match",
                "string_match": {
                  "exact": "c05b1311c2331be58862578ce1c7c47cdc95d4266cef778bd50b7da5fd8e4fb1"
                }
              }
            ]
          },
          "direct_response": {
            "status": 304
          }
        },
        {
          "match": {
            "path": "/dns"
          },
          "direct_response": {
            "status": 200,
            "body": {
              "inline_string": "{\"records\":[{\"name\":\"httpbin.kuma-demo.svc.8000.mesh\",\"ips\":[\"240.0.0.2\",\"::ffff:f000:2\"]},{\"name\":\"httpbin_kuma-demo_svc_8000.mesh\",\"ips\":[\"240.0.0.2\",\"::ffff:f000:2\"]},{\"name\":\"nginx-service.kuma-demo.svc.80.mesh\",\"ips\":[\"240.0.0.1\",\"::ffff:f000:1\"]},{\"name\":\"nginx-service_kuma-demo_svc_80.mesh\",\"ips\":[\"240.0.0.1\",\"::ffff:f000:1\"]}],\"ttl\":30}"
            }
          },
          "response_headers_to_add": [
            {
              "header": {
                "key": "Etag",
                "value": "c05b1311c2331be58862578ce1c7c47cdc95d4266cef778bd50b7da5fd8e4fb1"
              }
            }
          ],
          "name": "_kuma:dynamicconfig:/dns"
        },
        {
          "match": {
            "path": "/meshmetric",
            "headers": [
              {
                "name": "If-None-Match",
                "string_match": {
                  "exact": "39fa51495718bc31d1be9553a00cb3fb266e5a56d7ee60fa5eb3b4ad9839192a"
                }
              }
            ]
          },
          "direct_response": {
            "status": 304
          }
        },
        {
          "match": {
            "path": "/meshmetric"
          },
          "direct_response": {
            "status": 200,
            "body": {
              "inline_string": "{\"observability\":{\"metrics\":{\"applications\":null,\"backends\":[{\"type\":\"Prometheus\",\"name\":null}],\"sidecar\":{\"profiles\":{\"appendProfiles\":[{\"name\":\"None\"}],\"include\":[{\"type\":\"Regex\",\"match\":\"envoy_cluster_external_upstream_rq_.*\"},{\"type\":\"Regex\",\"match\":\"envoy_cluster_default_total_match_count\"},{\"type\":\"Regex\",\"match\":\"envoy_cluster_external_upstream_rq_active\"},{\"type\":\"Regex\",\"match\":\"envoy_cluster_external_upstream_rq_pending_total\"},{\"type\":\"Regex\",\"match\":\"envoy_cluster_external_upstream_rq_pending_overflow\"},{\"type\":\"Regex\",\"match\":\"envoy_cluster_external_upstream_rq_pending_failure_eject\"}]}},\"extraLabels\":{\"app\":\"nginx\",\"apps\":\",nginx,\",\"k8s_kuma_io_name\":\"nginx-5f54549bb4-8wfw7\",\"k8s_kuma_io_namespace\":\"kuma-demo\",\"k8s_kuma_io_namespaces\":\",kuma-demo,\",\"k8s_kuma_io_service_name\":\"nginx-service\",\"k8s_kuma_io_service_names\":\",nginx-service,\",\"k8s_kuma_io_service_port\":\"80\",\"k8s_kuma_io_service_ports\":\",80,\",\"kubernetes_io_hostname\":\"k3d-km-server-0\",\"kubernetes_io_hostnames\":\",k3d-km-server-0,\",\"kuma_io_protocol\":\"http\",\"kuma_io_protocols\":\",http,\",\"kuma_io_service\":\"nginx-service_kuma-demo_svc_80\",\"kuma_io_services\":\",nginx-service_kuma-demo_svc_80,\",\"kuma_io_zone\":\"default\",\"kuma_io_zones\":\",default,\",\"pod_template_hash\":\"5f54549bb4\",\"pod_template_hashs\":\",5f54549bb4,\"}}}}"
            }
          },
          "response_headers_to_add": [
            {
              "header": {
                "key": "Etag",
                "value": "39fa51495718bc31d1be9553a00cb3fb266e5a56d7ee60fa5eb3b4ad9839192a"
              }
            }
          ],
          "name": "_kuma:dynamicconfig:/meshmetric"
        }
      ]
    }
  ],
  "max_direct_response_body_size_bytes": 8192
}
```

lahabana: But we only call 1 endpoint at a time right? So the body is just 1 endpoint at once. Can't you just check this with starting an envoy and loading each config?

Icarus9913: You're right, and I did a test for it, I should choose a single endpoint. Thanks for correcting me, let me fix it

pkg/xds/dynconf/listener.go
```diff
@@ -22,6 +23,8 @@ const (
 	ListenerName = "_kuma:dynamicconfig"
 )
 
+var maxDirectResponseBodyPadding uint32 = 256 // 256 bytes
```
lahabana: my argument about padding still holds.

pkg/xds/generator/dns_generator_test.go
```diff
@@ -9,66 +9,83 @@ import (
 	. "github.com/onsi/gomega"
 
 	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
+	core_plugins "github.com/kumahq/kuma/pkg/core/plugins"
 	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
 	model "github.com/kumahq/kuma/pkg/core/xds"
 	xds_types "github.com/kumahq/kuma/pkg/core/xds/types"
+	core_rules "github.com/kumahq/kuma/pkg/plugins/policies/core/rules"
```
lahabana: NO let's not do these crazy imports. You are importing something very specific in a very generic package. This is why our linters and linkers take forever. 

You should be able to just unit test the generic behaviour of dynconf in the dynconf package. This will create a nice small test that doesn't need to evolve when implementations change. That's "write once, works forever" which is the core of engineering.

pkg/xds/generator/testdata/dns/4-envoy-config.golden.yaml
lahabana, line 1: you should name these 0X so we can have more than 10 tests.

It's nicer and easier to use: https://github.com/kumahq/kuma/blob/fbdc26e98e1af65b35377510810d5341f759c7ee/pkg/test/ginkgo.go#L67 

This avoids requiring to add many things whenever you add a test.

pkg/xds/dynconf/listener.go
Icarus9913: All right, I've also verified it works without padding. 
Let's remove it.

pkg/xds/dynconf/listener.go
```diff
@@ -105,6 +106,11 @@ func AddConfigRoute(proxy *core_xds.Proxy, rs *core_xds.ResourceSet, name string
 				},
 			},
 		)
+		if routeConfig.MaxDirectResponseBodySizeBytes == nil {
+			routeConfig.MaxDirectResponseBodySizeBytes = wrapperspb.UInt32(uint32(len(bytes)))
```
lahabana, line 110: why did the padding go away?

lahabana (approved):
LGTM

pkg/xds/dynconf/listener.go
lahabana: @Icarus9913 wasn't convinced of the need for it

Issue #13753, opened 2025-06-17
Implement new `_layout` endpoint


Automaat:
### Description

For this new endpoint, we need to implement a collection of outbound from MeshContext. We need to combine MeshService + MeshMultiZoneService + MeshExternalService, and filter it by reachable backends from Dataplane resource.

PR #13855, opened 2025-07-04
feat(kuma-cp): implement dataplane layout endpoint

Automaat:
## Motivation

Implementing new `_layout` endpoint as described in MADR: https://docs.google.com/document/d/1EzZpk3wwneIxQNPK7WXJqhW3Y3CS1AWMXIcXix9LEoc/edit?tab=t.g7ooo2ntj4jj

Fix https://github.com/kumahq/kuma/issues/13753

### DIFF
--- a/pkg/api-server/dataplane_layout_endpoint.go
+++ b/pkg/api-server/dataplane_layout_endpoint.go
@@ -1,0 +1,186 @@
+package api_server
+
+import (
+	"context"
+	"net/http"
+	"slices"
+	"strconv"
+	"strings"
+
+	"github.com/emicklei/go-restful/v3"
+
+	"github.com/kumahq/kuma/api/mesh/v1alpha1"
+	"github.com/kumahq/kuma/api/openapi/types"
+	api_common "github.com/kumahq/kuma/api/openapi/types/common"
+	"github.com/kumahq/kuma/pkg/core/kri"
+	"github.com/kumahq/kuma/pkg/core/resources/access"
+	"github.com/kumahq/kuma/pkg/core/resources/apis/core"
+	core_mesh "github.com/kumahq/kuma/pkg/core/resources/apis/mesh"
+	"github.com/kumahq/kuma/pkg/core/resources/manager"
+	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
+	"github.com/kumahq/kuma/pkg/core/resources/store"
+	rest_errors "github.com/kumahq/kuma/pkg/core/rest/errors"
+	"github.com/kumahq/kuma/pkg/core/user"
+	util_slices "github.com/kumahq/kuma/pkg/util/slices"
+	xds_context "github.com/kumahq/kuma/pkg/xds/context"
+)
+
+type dataplaneLayoutEndpoint struct {
+	resManager         manager.ResourceManager
+	meshContextBuilder xds_context.MeshContextBuilder
+	resourceAccess     access.ResourceAccess
+}
+
+func newDataplaneLayoutEndpoint(resManager manager.ResourceManager, meshContextBuilder xds_context.MeshContextBuilder, resourceAccess access.ResourceAccess) *dataplaneLayoutEndpoint {
+	return &dataplaneLayoutEndpoint{
+		resManager:         resManager,
+		meshContextBuilder: meshContextBuilder,
+		resourceAccess:     resourceAccess,
+	}
+}
+
+func (dle *dataplaneLayoutEndpoint) addEndpoint(ws *restful.WebService) {
+	ws.Route(
+		ws.GET("/meshes/{mesh}/dataplanes/{name}/_layout").
+			To(dle.getLayout).
+			Doc("Get Dataplane Layout").
+			Returns(http.StatusOK, "OK", nil),
+	)
+}
+
+func (dle *dataplaneLayoutEndpoint) getLayout(request *restful.Request, response *restful.Response) {
+	meshName := request.PathParameter("mesh")
+	dataplaneName := request.PathParameter("name")
+
+	if err := dle.resourceAccess.ValidateGet(
+		request.Request.Context(),
+		core_model.ResourceKey{Mesh: meshName, Name: dataplaneName},
+		core_mesh.DataplaneResourceTypeDescriptor,
+		user.FromCtx(request.Request.Context()),
+	); err != nil {
+		rest_errors.HandleError(request.Request.Context(), response, err, "Access Denied")
+		return
+	}
+
+	mesh, err := dle.getMeshResource(request.Request.Context(), meshName)
+	if err != nil {
+		rest_errors.HandleError(request.Request.Context(), response, err, "Failed to retrieve Mesh")
+		return
+	}
+
+	if mesh.Spec.GetMeshServices().GetMode() == v1alpha1.Mesh_MeshServices_Disabled {
+		response.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	dataplane, err := dle.getDataplaneResource(request.Request.Context(), meshName, dataplaneName)
+	if err != nil {
+		rest_errors.HandleError(request.Request.Context(), response, err, "Failed to retrieve Dataplane")
+		return
+	}
+
+	baseMeshContext, err := dle.meshContextBuilder.BuildBaseMeshContextIfChanged(request.Request.Context(), meshName, nil)
+	if err != nil {
+		rest_errors.HandleError(request.Request.Context(), response, err, "Failed to build MeshContext")
+	}
+
+	inbounds := util_slices.Map(dataplane.Spec.GetNetworking().GetInbound(), func(inbound *v1alpha1.Dataplane_Networking_Inbound) api_common.DataplaneInbound {
+		sectionName := inbound.GetName()
+		if sectionName == "" {
+			sectionName = strconv.Itoa(int(inbound.GetPort()))
+		}
+
+		return api_common.DataplaneInbound{
+			Kri:      kri.From(dataplane, sectionName).String(),
+			Port:     int32(inbound.GetPort()),
+			Protocol: inbound.GetProtocol(),
+		}
+	})
+
+	var outbounds []api_common.DataplaneOutbound
+	if dataplane.Spec.GetNetworking().GetOutbound() != nil {
+		// TODO handle user defined outbounds on universal without transparent proxy
+	} else {
+		// handle reachable backend and outbounds when using transparent proxy
+		reachableBackends := baseMeshContext.DestinationIndex.GetReachableBackends(mesh, dataplane)
+		outbounds = dle.getOutbounds(baseMeshContext.DestinationIndex, reachableBackends)
+	}
+
+	slices.SortStableFunc(outbounds, func(a, b api_common.DataplaneOutbound) int {
+		return strings.Compare(a.Kri, b.Kri)
+	})
+
+	networkingLayout := types.DataplaneNetworkingLayout{
+		Inbounds:  inbounds,
+		Kri:       kri.From(dataplane, "").String(),
+		Labels:    dataplane.GetMeta().GetLabels(),
+		Outbounds: outbounds,
+	}
+
+	err = response.WriteHeaderAndJson(http.StatusOK, networkingLayout, "application/json")
+	if err != nil {
+		log.Error(err, "Could not write response")
+	}
+}
+
+func (dle *dataplaneLayoutEndpoint) getMeshResource(ctx context.Context, meshName string) (*core_mesh.MeshResource, error) {
+	meshResource := core_mesh.NewMeshResource()
+	err := dle.resManager.Get(ctx, meshResource, store.GetByKey(meshName, core_model.NoMesh))
+	if err != nil {
+		return nil, err
+	}
+	return meshResource, nil
+}
+
+func (dle *dataplaneLayoutEndpoint) getDataplaneResource(ctx context.Context, meshName string, dataplaneName string) (*core_mesh.DataplaneResource, error) {
+	dataplaneResource := core_mesh.NewDataplaneResource()
+	err := dle.resManager.Get(ctx, dataplaneResource, store.GetByKey(dataplaneName, meshName))
+	if err != nil {
+		return nil, err
+	}
+	return dataplaneResource, nil
+}
+
+func (dle *dataplaneLayoutEndpoint) getOutbounds(destinationContext *xds_context.DestinationIndex, reachableBackends *xds_context.ReachableBackends) []api_common.DataplaneOutbound {
+	var outbounds []api_common.DataplaneOutbound
+	if reachableBackends == nil {
+		for destinationKri, destination := range destinationContext.GetAllDestinations() {
+			outbounds = append(outbounds, expandDestination(destinationKri, destination)...)
+		}
+		return outbounds
+	}
+
+	for destinationKri := range *reachableBackends {
+		dest := destinationContext.GetDestinationByKri(destinationKri)
+		if dest == nil {
+			continue
+		}
+		if destinationKri.HasSectionName() {
+			port, ok := dest.FindPortByName(destinationKri.SectionName)
+			if !ok {
+				continue
+			}
+			outbounds = append(outbounds, api_common.DataplaneOutbound{
+				Kri:      destinationKri.String(),
+				Port:     port.GetValue(),
+				Protocol: string(port.GetProtocol()),
+			})
+		} else {
+			outbounds = append(outbounds, expandDestination(destinationKri, dest)...)
+		}
+	}
+
+	return outbounds
+}
+
+func expandDestination(destinationKri kri.Identifier, dest core.Destination) []api_common.DataplaneOutbound {
+	var outbounds []api_common.DataplaneOutbound
+	for _, port := range dest.GetPorts() {
+		outbounds = append(outbounds, api_common.DataplaneOutbound{
+			Kri:      kri.WithSectionNameFromPort(destinationKri, port.GetName()).String(),
+			Port:     port.GetValue(),
+			Protocol: string(port.GetProtocol()),
+		})
+	}
+	return outbounds
+}
--- a/pkg/api-server/testdata/resources/inspect/dataplanes/_layout/empty_reachable_backend.golden.json
+++ b/pkg/api-server/testdata/resources/inspect/dataplanes/_layout/empty_reachable_backend.golden.json
@@ -1,0 +1,19 @@
+{
+ "inbounds": [
+  {
+   "kri": "kri_dp_default_test-zone__dp-1_main-port",
+   "port": 8080,
+   "protocol": "http"
+  },
+  {
+   "kri": "kri_dp_default_test-zone__dp-1_secondary-port",
+   "port": 8081,
+   "protocol": "tcp"
+  }
+ ],
+ "kri": "kri_dp_default_test-zone__dp-1_",
+ "labels": {
+  "kuma.io/zone": "test-zone"
+ },
+ "outbounds": null
+}
--- a/pkg/api-server/testdata/resources/inspect/dataplanes/_layout/multiple_mixed_destinations.golden.json
+++ b/pkg/api-server/testdata/resources/inspect/dataplanes/_layout/multiple_mixed_destinations.golden.json
@@ -1,0 +1,40 @@
+{
+ "inbounds": [
+  {
+   "kri": "kri_dp_default_test-zone__dp-1_main-port",
+   "port": 8080,
+   "protocol": "http"
+  },
+  {
+   "kri": "kri_dp_default_test-zone__dp-1_secondary-port",
+   "port": 8081,
+   "protocol": "tcp"
+  }
+ ],
+ "kri": "kri_dp_default_test-zone__dp-1_",
+ "labels": {
+  "kuma.io/zone": "test-zone"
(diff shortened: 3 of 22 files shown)


pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func (dle *dataplaneLayoutEndpoint) getDataplaneResource(ctx context.Context, meshName string, dataplaneName string) (*core_mesh.DataplaneResource, error) {
+	dataplaneResource := core_mesh.NewDataplaneResource()
+	err := dle.resManager.Get(ctx, dataplaneResource, store.GetByKey(dataplaneName, meshName))
+	if err != nil {
+		return nil, err
+	}
+	return dataplaneResource, nil
+}
```
lahabana: Shouldn't these 2 just use the entity in the BaseMeshContext?

lahabana: nevermind I get it now

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+	baseMeshContext, err := dle.meshContextBuilder.BuildBaseMeshContextIfChanged(request.Request.Context(), meshName, nil)
+	if err != nil {
+		rest_errors.HandleError(request.Request.Context(), response, err, "Failed to build MeshContext")
+	}
+
+	inbounds := util_slices.Map(dataplane.Spec.GetNetworking().GetInbound(), func(inbound *v1alpha1.Dataplane_Networking_Inbound) api_common.DataplaneInbound {
+		sectionName := inbound.GetName()
+		if sectionName == "" {
+			sectionName = strconv.Itoa(int(inbound.GetPort()))
+		}
```
lahabana: Isn't there a function in inbound to return a section name already?

lahabana: If not maybe there should be one

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+	err := dle.resManager.Get(ctx, dataplaneResource, store.GetByKey(dataplaneName, meshName))
+	if err != nil {
+		return nil, err
+	}
+	return dataplaneResource, nil
+}
+
+func (dle *dataplaneLayoutEndpoint) getOutbounds(destinationContext *xds_context.DestinationIndex, reachableBackends *xds_context.ReachableBackends) []api_common.DataplaneOutbound {
+	var outbounds []api_common.DataplaneOutbound
+	if reachableBackends == nil {
```
lahabana: Need to look more into it but I'm suprised we need to do this here. To me the destinationIndex would have already resolved this.

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	mesh, err := dle.getMeshResource(request.Request.Context(), meshName)
+	if err != nil {
+		rest_errors.HandleError(request.Request.Context(), response, err, "Failed to retrieve Mesh")
+		return
+	}
+
+	if mesh.Spec.GetMeshServices().GetMode() == v1alpha1.Mesh_MeshServices_Disabled {
+		response.WriteHeader(http.StatusBadRequest)
```
lahabana: I think this should be:

```
rest_errors.HandleError(..., response, errors.NewBadRequest("can't use _layout endpoint without meshService"))
```

Otherwise it's likely not very straightforward for users to understand why this isn't valid and it's also not compliant with the AIP: https://kong-aip.netlify.app/aip/193/

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+	inbounds := util_slices.Map(dataplane.Spec.GetNetworking().GetInbound(), func(inbound *v1alpha1.Dataplane_Networking_Inbound) api_common.DataplaneInbound {
+		sectionName := inbound.GetName()
+		if sectionName == "" {
+			sectionName = strconv.Itoa(int(inbound.GetPort()))
+		}
+
+		return api_common.DataplaneInbound{
+			Kri:      kri.From(dataplane, sectionName).String(),
+			Port:     int32(inbound.GetPort()),
+			Protocol: inbound.GetProtocol(),
```
lahabana, line 83: are you going to backfill with the recent addition later?

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	slices.SortStableFunc(outbounds, func(a, b api_common.DataplaneOutbound) int {
+		return strings.Compare(a.Kri, b.Kri)
+	})
+
+	networkingLayout := types.DataplaneNetworkingLayout{
+		Inbounds:  inbounds,
+		Kri:       kri.From(dataplane, "").String(),
+		Labels:    dataplane.GetMeta().GetLabels(),
```
lahabana, line 104: Sorry only realising this but shouldn't this be a meta?

https://github.com/kumahq/kuma/blob/a19f7a26cd2636dd15c816e408e0ce6816e02323/api/openapi/specs/common/resource.yaml#L45 with potentially adding an option KRI field to this meta.

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+		Outbounds: outbounds,
+	}
+
+	err = response.WriteHeaderAndJson(http.StatusOK, networkingLayout, "application/json")
+	if err != nil {
+		log.Error(err, "Could not write response")
+	}
+}
+
+func (dle *dataplaneLayoutEndpoint) getMeshResource(ctx context.Context, meshName string) (*core_mesh.MeshResource, error) {
```
lahabana: I don't find these functions super useful. They add double error handling and hide things that might actually not need to be hidden...

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,186 @@
...
+func (dle *dataplaneLayoutEndpoint) getDataplaneResource(ctx context.Context, meshName string, dataplaneName string) (*core_mesh.DataplaneResource, error) {
+	dataplaneResource := core_mesh.NewDataplaneResource()
+	err := dle.resManager.Get(ctx, dataplaneResource, store.GetByKey(dataplaneName, meshName))
+	if err != nil {
+		return nil, err
+	}
+	return dataplaneResource, nil
+}
+
+func (dle *dataplaneLayoutEndpoint) getOutbounds(destinationContext *xds_context.DestinationIndex, reachableBackends *xds_context.ReachableBackends) []api_common.DataplaneOutbound {
```
lahabana: looks like this could be a function and not need `dle`

pkg/api-server/dataplane_layout_endpoint.go
Automaat: added

pkg/api-server/dataplane_layout_endpoint.go
lahabana: Yes the more I look at this code, the more it feels like the destinationContext should be built accurately and that reachableBackends and universal etc are stuff that shouldn't leak here. WDYT?

pkg/core/kri/kri.go
```diff
@@ -100,3 +100,13 @@ func IsValid(s string) bool {
 func Compare(a, b Identifier) int {
 	return strings.Compare(a.String(), b.String())
 }
+
+func (i Identifier) HasSectionName() bool {
+	return i.SectionName != ""
+}
+
+func WithSectionNameFromPort(id Identifier, sectionName string) Identifier {
```
lahabana: I don't get the relationship between the name of the function and what it does

pkg/api-server/dataplane_layout_endpoint.go
Automaat: yes, we need to have common logic that will build this new names so we can reuse it in layout endpoint

pkg/api-server/dataplane_layout_endpoint.go
Automaat: I don't think we need full meta here. We only use labels everywhere

pkg/core/kri/kri.go
Automaat: It was some old artifact. Updated name

pkg/api-server/dataplane_layout_endpoint.go
Automaat: removed whole function

pkg/api-server/dataplane_layout_endpoint.go
Automaat: right, forgot to update it

pkg/api-server/dataplane_layout_endpoint.go
Automaat: removed

pkg/api-server/dataplane_layout_endpoint.go
Automaat: I moved this whole logic to DestinationIndex

api/mesh/v1alpha1/dataplane_helpers.go
```diff
@@ -376,6 +376,17 @@ func (d *Dataplane_Networking_Inbound) GetProtocol() string {
 	return d.Tags[ProtocolTag]
 }
 
+// GetSectionName returns either inbound name or stringified port value
+func (d *Dataplane_Networking_Inbound) GetSectionName() string {
+	if d == nil {
+		return ""
+	}
```
lahabana, line 383: I really feel this should just panic WDYT?

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		rest_errors.HandleError(request.Request.Context(), response, err, "Failed to retrieve Dataplane")
+		return
+	}
+
+	inbounds := util_slices.Map(dataplane.Spec.GetNetworking().GetInbound(), func(inbound *v1alpha1.Dataplane_Networking_Inbound) api_common.DataplaneInbound {
+		sectionName := inbound.GetSectionName()
+
+		return api_common.DataplaneInbound{
+			Kri:      kri.From(dataplane, sectionName).String(),
```
lahabana:
```suggestion
			Kri:      kri.From(dataplane, inbound.GetSectionName()).String(),
```

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/kumahq/kuma/pkg/core/user"
+	util_slices "github.com/kumahq/kuma/pkg/util/slices"
+	xds_context "github.com/kumahq/kuma/pkg/xds/context"
+)
+
+type dataplaneLayoutEndpoint struct {
+	resManager         manager.ResourceManager
+	meshContextBuilder xds_context.MeshContextBuilder
+	resourceAccess     access.ResourceAccess
+}
```
lahabana, line 29: It's 100% a nit. But I feel weird we make an object for this.

We could just pass:

```
createDPEndpoint(resManager manager.ResourceManager, meshContextBuilder xds_context.MeshContextBuilder, resourceAccess access.ResourceAccess)
```

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,114 @@
...
+	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
+	"github.com/kumahq/kuma/pkg/core/resources/store"
+	rest_errors "github.com/kumahq/kuma/pkg/core/rest/errors"
+	"github.com/kumahq/kuma/pkg/core/user"
+	util_slices "github.com/kumahq/kuma/pkg/util/slices"
+	xds_context "github.com/kumahq/kuma/pkg/xds/context"
+)
+
+type dataplaneLayoutEndpoint struct {
+	resManager         manager.ResourceManager
```
lahabana: I think this can be a ReadOnlyResourceManager (if there's such a thing though...)

pkg/api-server/testdata/resources/inspect/dataplanes/_layout/empty_reachable_backend.golden.json
```diff
@@ -0,0 +1,19 @@
...
+   "kri": "kri_dp_default_test-zone__dp-1_secondary-port",
+   "port": 8081,
+   "protocol": "tcp"
+  }
+ ],
+ "kri": "kri_dp_default_test-zone__dp-1_",
+ "labels": {
+  "kuma.io/zone": "test-zone"
+ },
+ "outbounds": null
```
lahabana: should we have a `[]` instead?

pkg/api-server/testdata/resources/inspect/dataplanes/_layout/empty_reachable_backend.golden.json
```diff
@@ -0,0 +1,19 @@
+{
```
lahabana, line 1: @johncowen might appreciate reviewing these examples output early on

pkg/api-server/testdata/resources/inspect/dataplanes/_layout/multiple_mixed_destinations.golden.json
```diff
@@ -0,0 +1,40 @@
...
+ ],
+ "kri": "kri_dp_default_test-zone__dp-1_",
+ "labels": {
+  "kuma.io/zone": "test-zone"
+ },
+ "outbounds": [
+  {
+   "kri": "kri_extsvc_default___mes-1_9090",
+   "port": 9090,
+   "protocol": "http"
```
lahabana, line 22: random thoughts for the future: should we consider adding the hostnames and IPs here?

pkg/core/kri/kri.go
```diff
@@ -100,3 +100,13 @@ func IsValid(s string) bool {
 func Compare(a, b Identifier) int {
 	return strings.Compare(a.String(), b.String())
 }
+
+func (i Identifier) HasSectionName() bool {
+	return i.SectionName != ""
+}
+
+func WithSectionName(id Identifier, sectionName string) Identifier {
+	idCopy := id
```
lahabana: it's already a copy no?

pkg/xds/context/context.go
```diff
@@ -62,6 +62,10 @@ func (g BaseMeshContext) Hash() string {
 	return base64.StdEncoding.EncodeToString(g.hash)
 }
 
+func (g BaseMeshContext) GetReachableOutbounds(dataplane *core_mesh.DataplaneResource) map[kri.Identifier]core_resources.Port {
```
lahabana: I think a comment here would help.

pkg/xds/context/destination_index.go
```diff
@@ -82,6 +82,49 @@ func (dc *DestinationIndex) GetDestinationByKri(id kri.Identifier) core.Destinat
 	return dc.destinationByIdentifier[kri.NoSectionName(id)]
 }
 
+func (dc *DestinationIndex) getReachableOutbounds(mesh *core_mesh.MeshResource, dataplane *core_mesh.DataplaneResource) map[kri.Identifier]core.Port {
+	if dataplane.Spec.GetNetworking().GetOutbound() != nil {
+		// TODO handle user defined outbounds on universal without transparent proxy
```
lahabana: I think you already opened an issue for this. Can you already add the reference here please?

pkg/xds/context/destination_index.go
```diff
@@ -82,6 +82,49 @@ func (dc *DestinationIndex) GetDestinationByKri(id kri.Identifier) core.Destinat
 	return dc.destinationByIdentifier[kri.NoSectionName(id)]
 }
 
+func (dc *DestinationIndex) getReachableOutbounds(mesh *core_mesh.MeshResource, dataplane *core_mesh.DataplaneResource) map[kri.Identifier]core.Port {
```
lukidzi: Does it works with `--bind-outbounds`?

Automaat: Does it change anything? Also We don't look into proxy here, this is only based on Dataplane config

lukidzi: It's like a transparent proxy but no transparent proxy section. In this case you have all outbounds

lahabana (approved):
Some small nits

pkg/api-server/testdata/resources/inspect/dataplanes/_layout/empty_reachable_backend.golden.json
Automaat: I guess so, but in schema we have this field required and field in struct does not have `omitempty` and it still generates null. I am not sure what to do with this

pkg/core/kri/kri.go
Automaat: it is you are right

pkg/xds/context/context.go
Automaat: this no longer apply

pkg/xds/context/destination_index.go
Automaat: linked

pkg/api-server/dataplane_layout_endpoint.go
Automaat: I am not sure, we do it like this in other places so I wanted to be consistent with current approach

pkg/api-server/dataplane_layout_endpoint.go
Automaat: changed

pkg/api-server/testdata/resources/inspect/dataplanes/_layout/multiple_mixed_destinations.golden.json
Automaat: I am not sure, you can easily get this information when you extract resource by its KRI which is already there

pkg/api-server/testdata/resources/inspect/dataplanes/_layout/empty_reachable_backend.golden.json
lahabana: I think you need to explicitly set an empty array

pkg/api-server/dataplane_layout_endpoint.go
```diff
@@ -0,0 +1,112 @@
...
+
+	inbounds := util_slices.Map(dataplane.Spec.GetNetworking().GetInbound(), func(inbound *v1alpha1.Dataplane_Networking_Inbound) api_common.DataplaneInbound {
+		return api_common.DataplaneInbound{
+			Kri:      kri.From(dataplane, inbound.GetSectionName()).String(),
+			Port:     int32(inbound.GetPort()),
+			Protocol: inbound.GetProtocol(),
+		}
+	})
+
+	var outbounds []api_common.DataplaneOutbound
```
lahabana:
```suggestion
	outbounds := []api_common.DataplaneOutbound{}
```

should do the trick about the below comment

Automaat: done

pkg/api-server/testdata/resources/inspect/dataplanes/_layout/empty_reachable_backend.golden.json
Automaat: fixed

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/helpers.go
```diff
@@ -73,7 +71,7 @@ func (t *MeshExternalServiceResource) FindPortByName(name string) (core.Port, bo
 }
 
 func (m Match) GetName() string {
-	return fmt.Sprintf("%d", m.Port)
+	return ""
```
lahabana, line 75: This seems super odd to me. @lukidzi WDYT?

lahabana: My guess is that there can only ever be 1 port in a MES and that's why

lukidzi: Yes, MES only has 1 port probably that's why

lahabana:
```suggestion
        // MES can only have one port so there's no point in naming it.
	return ""
```

lahabana (approved):

pkg/api-server/testdata/resources/inspect/dataplanes/_layout/empty_reachable_backend.golden.json
johncowen: I seem to remember we added a further lookup key as well as the kri here? I think it was to be called `proxyResourceName` a bit like this:

```
inbounds:
  - kri: <resource_identifier> # Inbound KRI
    port: <port_number> # port number
    protocol: http # port protocol
    proxyResourceName: <common_prefix>_sectionName
```

I believe the plan was to add this to both inbounds and outbounds (even though the outbounds one is probably identical to the kri)

johncowen: p.s. @bartsmykla would have more info on this I think

lukidzi (dismissed):

slonka (approved):
I'm approving for @lukidzi since he made changes to this PR so his approve doesn't count

Issue #13755, opened 2025-06-17
Implement new `_policy` endpoint


Automaat:
### Description

As described in [MADR](https://docs.google.com/document/d/1EzZpk3wwneIxQNPK7WXJqhW3Y3CS1AWMXIcXix9LEoc/edit?tab=t.g7ooo2ntj4jj). We need to implement filtering of already merged policies for a given dpp/inbound/outbound

PR #13871, opened 2025-07-07
feat(kuma-cp): implement new _policies endpoints for dataplane

Automaat:
## Motivation
Implementing new inspect api according to the design: https://docs.google.com/document/d/1EzZpk3wwneIxQNPK7WXJqhW3Y3CS1AWMXIcXix9LEoc/edit?tab=t.g7ooo2ntj4jj#heading=h.hfhi2p9zepu1

Fix https://github.com/kumahq/kuma/issues/13755

> Changelog: feat(api): implement new _policies endpoints for dataplane

### DIFF
--- a/pkg/api-server/resource_endpoints.go
+++ b/pkg/api-server/resource_endpoints.go
@@ -905,7 +905,7 @@ func (r *resourceEndpoints) getPoliciesConf(mapToResponse matchedPoliciesToRespo
 		}
 
 		if baseMeshContext.Mesh.Spec.GetMeshServices().GetMode() == mesh_proto.Mesh_MeshServices_Disabled {
-			response.WriteHeader(http.StatusBadRequest)
+			rest_errors.HandleError(request.Request.Context(), response, rest_errors.NewBadRequestError("can't use _policies endpoint without meshService enabled"), "Bad Request")
 			return
 		}
 
--- a/pkg/api-server/testdata/resources/inspect/dataplanes/_policies/meshservice_disabled.golden.json
+++ b/pkg/api-server/testdata/resources/inspect/dataplanes/_policies/meshservice_disabled.golden.json
@@ -1,0 +1,7 @@
+{
+ "type": "/std-errors",
+ "status": 400,
+ "title": "Bad Request",
+ "detail": "bad request: can't use _policies endpoint without meshService enabled",
+ "details": "bad request: can't use _policies endpoint without meshService enabled"
+}


api/mesh/v1alpha1/dataplane_helpers.go
```diff
@@ -288,6 +288,15 @@ func (n *Dataplane_Networking) GetInboundForPort(port uint32) *Dataplane_Network
 	return nil
 }
 
+func (n *Dataplane_Networking) GetInboundForSectionName(sectionName string) *Dataplane_Networking_Inbound {
+	for _, inbound := range n.Inbound {
+		if inbound.Name == sectionName || strconv.Itoa(int(inbound.Port)) == sectionName {
```
lahabana: again this complex code. Any chance we can use something generic in `inbound` ?

pkg/api-server/testdata/resources/inspect/dataplanes/_policies/merged_outbound_polices.golden.json
lahabana, line 1: typo in policies

pkg/api-server/testdata/resources/inspect/dataplanes/_policies/merged_outbound_polices.golden.json
```diff
@@ -0,0 +1,24 @@
...
+   ],
+   "kind": "MeshTimeout",
+   "origins": [
+    {
+     "kri": "kri_mt_default___mt-2_"
+    },
+    {
+     "kri": "kri_mt_default___mt-1_"
+    }
+   ]
```
lahabana: Are we sure about this ordering? I would have thought it was the other way round (mt-1 first then mt-2)

pkg/api-server/testdata/resources/inspect/dataplanes/_policies/single_outbound_policy_selecting_port.input.yaml
lahabana, line 1: good thorough test suite. I think it would help if you create 1 folder for inbound and one for outbound. After all they different APIs

api/mesh/v1alpha1/dataplane_helpers.go
Automaat: I am not sure what, do you have any ideas?

pkg/api-server/testdata/resources/inspect/dataplanes/_policies/merged_outbound_polices.golden.json
Automaat: this is what we have in other places: https://github.com/kumahq/kuma/blob/c3781d4fe676766bdbae6add5f3eab3f6c35a46d/pkg/api-server/testdata/resources/inspect/dataplanes/_rules/overriding_meshtimeout.golden.json#L29-L45

it goes from lest to most specific

pkg/api-server/testdata/resources/inspect/dataplanes/_policies/single_outbound_policy_selecting_port.input.yaml
Automaat: moved to different directories

pkg/api-server/testdata/resources/inspect/dataplanes/_policies/merged_outbound_polices.golden.json
Automaat: fixed

api/mesh/v1alpha1/dataplane_helpers.go
slonka: I guess Charly means something like:

```
func (d *Dataplane_Networking_Inbound) IsSection(sectionName string) bool {
	if d.Name == sectionName || strconv.Itoa(int(d.Port)) == sectionName {
		return true
	}
	return false
}
```

? or maybe something similar?

pkg/api-server/resource_endpoints.go
```diff
@@ -139,6 +143,25 @@ func (r *resourceEndpoints) addFindEndpoint(ws *restful.WebService, pathPrefix s
...
+	if r.descriptor.Name == core_mesh.DataplaneType {
+		ws.Route(ws.GET(pathPrefix+"/{name}/_policies").To(r.getPoliciesConf(matchedPoliciesToProxyPolicy)).
+			Doc(fmt.Sprintf("Get policy config %s", r.descriptor.Name)).
+			Param(ws.PathParameter("name", fmt.Sprintf("Name of a %s", r.descriptor.Name)).DataType("string")).
+			Returns(200, "OK", nil).
+			Returns(404, "Not found", nil))
+		ws.Route(ws.GET(pathPrefix+"/{name}/inbounds/{inbound_kri}/_policies").To(r.getPoliciesConf(matchedPoliciesToInboundConfig)).
+			Doc("Get policy config for inbound").
+			Param(ws.PathParameter("name", fmt.Sprintf("Name of a %s", r.descriptor.Name)).DataType("string")).
+			Param(ws.PathParameter("inbound_kri", "KRI of a inbound").DataType("string")).
```
slonka, line 155: inbounds are not going to be KRI https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md#naming-for-non-system-inbound-related-resources - should we call it `contextual_descriptor`? 🤔 FYI @bartsmykla

pkg/api-server/resource_endpoints.go
```diff
@@ -847,6 +870,148 @@ func (r *resourceEndpoints) configForProxyParams(request *restful.Request) (*api
 	return params, nil
 }
 
+func (r *resourceEndpoints) getPoliciesConf(mapToResponse matchedPoliciesToResponse) restful.RouteFunction {
```
slonka, line 875: This is a comment to this file in general not to the specific function. We shouldn't mix business logic with controller code. The question is if we want to pull it out in this PR or as a separate work.

pkg/api-server/resource_endpoints.go
```diff
@@ -847,6 +870,148 @@ func (r *resourceEndpoints) configForProxyParams(request *restful.Request) (*api
...
+			if res.Type == "" {
+				rest_errors.HandleError(request.Request.Context(), response, err, fmt.Sprintf("matched policy didn't set type for policy plugin %s", policyPlugin.Name))
+			}
+
+			matchedPolicies = append(matchedPolicies, res)
+		}
+
+		out, err := mapToResponse(matchedPolicies, request, dataplane, baseMeshContext.Resources())
+		if err != nil {
+			rest_errors.HandleError(request.Request.Context(), response, err, "Failed to TODO")
```
slonka: TODO

pkg/api-server/resource_endpoints.go
```diff
@@ -847,6 +870,148 @@ func (r *resourceEndpoints) configForProxyParams(request *restful.Request) (*api
...
+			res, err := policyPlugin.Plugin.MatchedPolicies(dataplane, baseMeshContext.Resources())
+			if err != nil {
+				rest_errors.HandleError(request.Request.Context(), response, err, fmt.Sprintf("could not apply policy plugin %s", policyPlugin.Name))
+			}
+			if res.Type == "" {
+				rest_errors.HandleError(request.Request.Context(), response, err, fmt.Sprintf("matched policy didn't set type for policy plugin %s", policyPlugin.Name))
+			}
+
+			matchedPolicies = append(matchedPolicies, res)
+		}
```
slonka, line 924: this seems generic and a bit similar (but simplified) to https://github.com/kumahq/kuma/blob/e4aaf62370225d07be0bff673e3835d0c156f539/pkg/plugins/policies/core/matchers/dataplane.go#L40 - can this be reused somehow?

api/mesh/v1alpha1/dataplane_helpers.go
lobkovilya: yes, it's the same logic we already have in https://github.com/kumahq/kuma/blob/e4aaf62370225d07be0bff673e3835d0c156f539/pkg/plugins/policies/core/matchers/dataplane.go#L239

can be reused

pkg/api-server/resource_endpoints.go
Automaat: It will use KRI in inspect api, nothing changed here. It will just have an extra name to correlate it with envoy resources/stats

pkg/api-server/resource_endpoints.go
Automaat: but we use `MatchedPolicies` func here. We need to run it for every policy plugin

api/mesh/v1alpha1/dataplane_helpers.go
Automaat: made some refactoring, this should be better now

pkg/api-server/resource_endpoints.go
slonka: We chatted offline and decided that until there is another codepath that would need the code to be in a service or some other module we're not going to move it.

slonka (approved):

Issue #12374, opened 2024-12-23
MeshTrafficPermission MADR for `spec.rules[].matches`


lobkovilya:
### Description

All other inbound policies make sense even without `rules[].matches`. MeshTrafficPermission doesn't provide any values without `matches` field.

We agreed in the [MADR](https://docs.google.com/document/d/1tdIOVVYObHbGKX1AFhbPH3ZQKYvpQG-2s6ShHSECaXM/edit?tab=t.0#heading=h.fxzs6450juto) to introduce SPIFEE matcher support as the client cert already has it in the lists of URIs. We need to decide how is this `matches` going to look like? Do we need only `Prefix` and `Exact`?

PR #13879, opened 2025-07-07
docs(MADR): MeshTrafficPermission Is a Special Case of Inbound Policies and Requires a Custom Matching Algorithm

lobkovilya:
## Motivation

While working on https://github.com/kumahq/kuma/issues/12374 I realized MeshTrafficPermission is just too different to try to invent a unified matching algo for all inbound policies.

This MADR aims to make MeshTrafficPermission a special case and justify custom matching algorithm.

### DIFF
--- a/docs/madr/decisions/078-special-mtp-algo.md
+++ b/docs/madr/decisions/078-special-mtp-algo.md
@@ -1,0 +1,119 @@
+# MeshTrafficPermission Is a Special Case of Inbound Policies and Requires a Custom Matching Algorithm
+
+* Status: accepted
+
+## Context and Problem Statement
+
+In the [Inbound Policies MADR](https://docs.google.com/document/d/1tdIOVVYObHbGKX1AFhbPH3ZQKYvpQG-2s6ShHSECaXM) we said 
+all inbound policies are going to have similar structure:
+
+```yaml
+spec:
+  targetRef: {}
+  rules:
+     - matches: []
+       default: {}
+```
+
+and we'll create a unified matching algorithm that'll work for all inbound policies and potentially use Envoy Matching API.
+
+But turns out, it's a real challenge to come up with unified algorithm that'd work both for `MeshTrafficPermission` and other policies.
+
+Technically it's possible to build a matching tree for Envoy that'd be similar for all inbound policies.
+But, the size of the tree is getting big really fast.
+
+Let's look at the example:
+
+```yaml
+rules:
+   - matches:
+        - spiffeId:
+            type: Prefix
+            value: spiffe://trust-domain.mesh/
+     default: $conf1
+   - matches:
+        - path:
+            type: Prefix
+            value: /foo
+     default: $conf2
+   - matches:
+        - method: GET
+     default: $conf3
+```
+
+We’re matching on three independent dimensions:
+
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/078-special-mtp-algo.md
```diff
@@ -0,0 +1,119 @@
+# MeshTrafficPermission Is a Special Case of Inbound Policies and Requires a Custom Matching Algorithm
+
+* Status: accepted
+
+## Context and Problem Statement
+
+In the [Inbound Policies MADR](https://docs.google.com/document/d/1tdIOVVYObHbGKX1AFhbPH3ZQKYvpQG-2s6ShHSECaXM) we said
```
lahabana: How can this is not a MADR in markdown (or at least an number with a link to the google docs)?

There's no point in having MADRs if there's no centralised source of truth.

lobkovilya: There is MADR with link https://github.com/kumahq/kuma/blob/abaf5c54ff421a82c447eb8e28a3561dbfaf0e38/docs/madr/decisions/066-inbound-policies.md#L7-L6, I just wanted to save few clicks for readers, but I can replace the link

lahabana (approved):
Makes sense and gets to the points quickly!

docs/madr/decisions/078-special-mtp-algo.md
lobkovilya: updated the link

docs/madr/decisions/078-special-mtp-algo.md
```diff
@@ -0,0 +1,118 @@
...
+     default: $conf3
+```
+
+We’re matching on three independent dimensions:
+
+1. Whether the SPIFFE ID starts with `spiffe://trust-domain.mesh/`
+2. Whether the request path starts with `/foo`
+3. Whether the request method is `GET`
+
+Each condition has two possible outcomes: it either matches or doesn’t.
```
jijiechen, line 49: The MADR overall looks good to me. 
Can we add some explanation that the above example is not from a single user-defined policy and it's actually a composite of all user-defined **policies** that elements under `rules` do not maintain a sequencial order?
It costed me some time to figure out why there can be 8 combinations because they look comimg from a single policy and we could just match the rules one by one.

lobkovilya: Single or composite is not actually relevant in this scenario. It'll be 8 combinations coming from a single policy. 

> we could just match the rules one by one

That's the issue with MeshTrafficPermission, we can't match only the first matching rule, we need to check all matching policies. And if we want to check all policies we need to build a tree for envoy that has 8 leaves

jijiechen: Understood, thanks!
So this is decided by the different semantics between MTP and other policies.

lobkovilya: Yes, MTP semantic is quite different, and if we try to come up with the unified algo that works for all policies then it'll be highly suboptimal. That's why this MADR suggests to treat MTP differently

lahabana (approved):

Issue #13754, opened 2025-06-17
Implement new `/meshes/<mesh_name>/dataplanes/<dataplane_name>/_outbounds/<outbound_kri>/_routes` endpoint


Automaat:
### Description

We need to implement functionality for getting routes for a given outbound. We need to follow schema from [MADR](https://docs.google.com/document/d/1EzZpk3wwneIxQNPK7WXJqhW3Y3CS1AWMXIcXix9LEoc/edit?tab=t.g7ooo2ntj4jj) 

List of routes needs to be list of merged Mesh*Route policies.

PR #13892, opened 2025-07-08
feat(kuma-cp): implement endpoint to get routes for dataplane outbound

Automaat:
## Motivation

We need to add _routes endpoint to be able to extract computed view of routes for dataplane outbound

Fix https://github.com/kumahq/kuma/issues/13754


bartsmykla (approved):
lgtm

PR #13893, opened 2025-07-08
feat(kuma-dp): Introduced WorkDir in DataplaneRuntime

### LINKED ISSUE #9522: Introduce a new config item "WorkingDir" to replace socketDir/configDir config items
### What happened?

We currently have these directores to be set on the `DataplaneRuntime` object:

```go
// DataplaneRuntime defines the context in which dataplane (Envoy) runs.
type DataplaneRuntime struct {
	config.BaseConfig

        ...
	// Dir to store auto-generated Envoy bootstrap config in.
	ConfigDir string `json:"configDir,omitempty" envconfig:"kuma_dataplane_runtime_config_dir"`
	// SocketDir dir to store socket used between Envoy and the dp process
	SocketDir string `json:"socketDir,omitempty" envconfig:"kuma_dataplane_runtime_socket_dir"`
}
```

Sometimes, we also need to generate content into a temporary directory, and it's better for the `tempDir` to be customized as well. However, as we introduce more writable directores, we'll need to introduce more config items. This adds more complexity to the configuration.

Can we make the user to set the `workingDir` to be a parent and writable directory, and kuma creates sub-directories when needed, and then put generated content into each sub-directory according?

m4l1c1ou5:
## Motivation

This change introduces WorkDir in DataplaneRuntime to replace socketDir/configDir config

## Supporting documentation

Fixes #9522


m4l1c1ou5:
Hi @bartsmykla, thanks for the review
I've added backward support in my PR by adding a check if `workDir` is given, override both `configDir` and `socketDir` with `workDir`, but if `workDir` is not present, then, pick `workDir` value from `configDir`, and use `workDir` in the code flow.
I've also added the unit test case for testing this flow

app/kuma-dp/cmd/run.go
```diff
@@ -416,6 +429,11 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
 			"Use this flag to pass detailed transparent proxy settings to kuma-dp.",
 	)
 
+	err := cmd.PersistentFlags().MarkDeprecated("config-dir", "use --work-dir instead")
+	if err != nil {
```
bartsmykla: Nit:
```suggestion
	if err := cmd.PersistentFlags().MarkDeprecated("config-dir", "use --work-dir instead"); err != nil {
```

pkg/config/app/kuma-dp/config.go
```diff
@@ -217,7 +219,7 @@ type DataplaneRuntime struct {
 	// Resources defines the resources for this proxy.
 	Resources DataplaneResources `json:"resources,omitempty"`
 	// SocketDir dir to store socket used between Envoy and the dp process
-	SocketDir string `json:"socketDir,omitempty" envconfig:"kuma_dataplane_runtime_socket_dir"`
+	SocketDir string `json:"socketDir,omitempty" envconfig:"kuma_dataplane_runtime_socket_dir" deprecated:"use WorkDir instead"`
```
bartsmykla: Nit: This change isn’t needed because we don’t use this field’s tags. Instead, I suggest adding a comment above the field to explain it
```suggestion
	// Deprecated: use WorkDir instead
	SocketDir string `json:"socketDir,omitempty" envconfig:"kuma_dataplane_runtime_socket_dir"`
```

pkg/config/app/kuma-dp/config.go
```diff
@@ -194,7 +194,9 @@ type DataplaneRuntime struct {
 	// Path to Envoy binary.
 	BinaryPath string `json:"binaryPath,omitempty" envconfig:"kuma_dataplane_runtime_binary_path"`
 	// Dir to store auto-generated Envoy bootstrap config in.
-	ConfigDir string `json:"configDir,omitempty" envconfig:"kuma_dataplane_runtime_config_dir"`
+	ConfigDir string `json:"configDir,omitempty" envconfig:"kuma_dataplane_runtime_config_dir" deprecated:"use WorkDir instead"`
+	// WorkDir to store auto-generated Envoy bootstrap config in.
+	WorkDir string `json:"workDir,omitempty" envconfig:"kuma_dataplane_runtime_work_dir" overrides:"ConfigDir,SocketDir"`
```
bartsmykla: The deprecated tag on `ConfigDir` isn't needed since we don't currently use struct tags for deprecation warnings. Instead, I'd suggest putting the deprecation note in a comment above the field for clarity and consistency.

```suggestion
	// ConfigDir was used to store Envoy bootstrap config.
	// Deprecated: use WorkDir instead.
	ConfigDir string `json:"configDir,omitempty" envconfig:"kuma_dataplane_runtime_config_dir"`

	// WorkDir is the directory to store auto-generated Envoy bootstrap config.
	// It overrides values from deprecated ConfigDir and SocketDir.
	WorkDir string `json:"workDir,omitempty" envconfig:"kuma_dataplane_runtime_work_dir"`
```

app/kuma-dp/cmd/run.go
```diff
@@ -136,15 +136,27 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
...
+				if cfg.DataplaneRuntime.ConfigDir != "" {
+					runLog.Info("configDir(--config-dir) in config file is deprecated, use workDir(--work-dir) instead")
+					cfg.DataplaneRuntime.WorkDir = cfg.DataplaneRuntime.ConfigDir
+				}
+				if cfg.DataplaneRuntime.SocketDir != "" {
+					runLog.Info("socketDir in config file is deprecated, use workDir(--work-dir) instead")
+				}
+			} else {
+				cfg.DataplaneRuntime.SocketDir = cfg.DataplaneRuntime.WorkDir
+			}
```
bartsmykla: The logic is incorrect. If `SocketDir` is explicitly set, it should take priority to avoid breaking existing setups. To handle this correctly, we need to differentiate between `SocketDir` and `WorkDir` instead of always overriding `SocketDir` based on `WorkDir`.

Wdyt:
```suggestion
			if cfg.DataplaneRuntime.WorkDir == "" && cfg.DataplaneRuntime.ConfigDir != "" {
				runLog.Info("ConfigDir is deprecated, please use WorkDir instead")
				cfg.DataplaneRuntime.WorkDir = cfg.DataplaneRuntime.ConfigDir
			}

			if cfg.DataplaneRuntime.SocketDir != "" {
				runLog.Info("SocketDir is deprecated, please use WorkDir instead")
			} else {
				cfg.DataplaneRuntime.SocketDir = cfg.DataplaneRuntime.WorkDir
			}
```

app/kuma-dp/cmd/run.go
```diff
@@ -136,15 +136,27 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
...
+
+			if cfg.DataplaneRuntime.WorkDir == "" || cfg.DNS.ConfigDir == "" {
```
bartsmykla: Following the earlier comment
```suggestion
			if cfg.DataplaneRuntime.WorkDir == "" || cfg.DataplaneRuntime.SocketDir == "" || cfg.DNS.ConfigDir == "" {
```

app/kuma-dp/cmd/run.go
```diff
@@ -233,7 +245,7 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
 			opts.AdminPort = bootstrap.GetAdmin().GetAddress().GetSocketAddress().GetPortValue()
 
 			confFetcher := configfetcher.NewConfigFetcher(
-				core_xds.MeshMetricsDynamicConfigurationSocketName(cfg.DataplaneRuntime.SocketDir),
+				core_xds.MeshMetricsDynamicConfigurationSocketName(cfg.DataplaneRuntime.WorkDir),
```
bartsmykla: Following the earlier comment:
```suggestion
				core_xds.MeshMetricsDynamicConfigurationSocketName(cfg.DataplaneRuntime.SocketDir),
```

app/kuma-dp/cmd/run.go
```diff
@@ -461,7 +479,7 @@ func setupObservability(ctx context.Context, kumaSidecarConfiguration *types.Kum
 	accessLogStreamer := component.NewResilientComponent(
 		runLog.WithName("access-log-streamer"),
 		accesslogs.NewAccessLogStreamer(
-			core_xds.AccessLogSocketName(cfg.DataplaneRuntime.SocketDir, cfg.Dataplane.Name, cfg.Dataplane.Mesh),
+			core_xds.AccessLogSocketName(cfg.DataplaneRuntime.WorkDir, cfg.Dataplane.Name, cfg.Dataplane.Mesh),
```
bartsmykla: Following the earlier comment:

```suggestion
			core_xds.AccessLogSocketName(cfg.DataplaneRuntime.SocketDir, cfg.Dataplane.Name, cfg.Dataplane.Mesh),
```

app/kuma-dp/cmd/run.go
```diff
@@ -477,7 +495,7 @@ func setupObservability(ctx context.Context, kumaSidecarConfiguration *types.Kum
 		tpEnabled,
 	)
 	metricsServer := metrics.New(
-		core_xds.MetricsHijackerSocketName(cfg.DataplaneRuntime.SocketDir, cfg.Dataplane.Name, cfg.Dataplane.Mesh),
+		core_xds.MetricsHijackerSocketName(cfg.DataplaneRuntime.WorkDir, cfg.Dataplane.Name, cfg.Dataplane.Mesh),
```
bartsmykla: Following the earlier comment:

```suggestion
		core_xds.MetricsHijackerSocketName(cfg.DataplaneRuntime.SocketDir, cfg.Dataplane.Name, cfg.Dataplane.Mesh),
```

app/kuma-dp/cmd/run_test.go
```diff
@@ -220,6 +220,35 @@ var _ = Describe("run", func() {
 				expectedFile: filepath.Join(tmpDir, "bootstrap.yaml"),
 			}
 		}),
+		Entry("can be launched with args and given config dir", func() testCase {
```
bartsmykla: Nit:
```suggestion
		Entry("can be launched with args and given work dir", func() testCase {
```

pkg/config/app/kuma-dp/config_test.go
```diff
@@ -59,6 +59,7 @@ var _ = Describe("Config", func() {
 				"KUMA_DATAPLANE_PROXY_TYPE":                                     "ingress",
 				"KUMA_DATAPLANE_RUNTIME_BINARY_PATH":                            "envoy.sh",
 				"KUMA_DATAPLANE_RUNTIME_CONFIG_DIR":                             "/var/run/envoy",
+				"KUMA_DATAPLANE_RUNTIME_WORK_DIR":                               "/var/run/envoy",
```
bartsmykla, line 63:
```suggestion
				"KUMA_DATAPLANE_RUNTIME_SOCKET_DIR":                             "/var/run/envoy",
				"KUMA_DATAPLANE_RUNTIME_WORK_DIR":                               "/var/run/envoy",
```

pkg/config/app/kuma-dp/config_test.go
```diff
@@ -93,6 +94,7 @@ var _ = Describe("Config", func() {
 			Expect(cfg.Dataplane.DrainTime.Duration).To(Equal(60 * time.Second))
 			Expect(cfg.DataplaneRuntime.BinaryPath).To(Equal("envoy.sh"))
 			Expect(cfg.DataplaneRuntime.ConfigDir).To(Equal("/var/run/envoy"))
+			Expect(cfg.DataplaneRuntime.WorkDir).To(Equal("/var/run/envoy"))
```
bartsmykla, line 99:
```suggestion
			Expect(cfg.DataplaneRuntime.SocketDir).To(Equal("/var/run/envoy"))
			Expect(cfg.DataplaneRuntime.WorkDir).To(Equal("/var/run/envoy"))
```

bartsmykla (approved):
Great job @m4l1c1ou5, congrats on getting another contribution merged! 🎉 Thanks for your continued work on the project!

Issue #13593, opened 2025-05-22
Mark envoy as ready only if `identity_cert` is initialized when mTLS is enabled


jijiechen:
### Description

We need take `identity_cert` readiness into account when reporting dataplane readiness.

when mTLS is enabled, `identity_cert` is referenced by Envoy clusters and is required lastly by Envoy. We are currently responding the readiness as `Ready` by a cluster pointing to a simple HTTP handler in the `kuma-dp` process. We are not taking `identity_cert` into account so a DP can reporte itself as ready before its `identity_cert` is initialized. 

Current implementation:

https://github.com/kumahq/kuma/blob/60de16bb8c29890e2b6f3eca791495f61634b79f/pkg/xds/generator/testdata/admin/06.envoy-config.golden.yaml#L57-L68

When `identity_cert` takes long, like a few seconds or even longer, there can be traffic drop on this dataplane: 
1. It can not verify mTLS traffic
2. the traffic sent from this dataplane will fail mTLS validation

jijiechen:
The "uninitialized" state lasted 614ms in my env, so I think  can't be ignored:

I have 51 services and 1 dataplane in each of them in my mesh.

Here are the envoy configs dumped:

```json
{
   "dynamic_warming_secrets": [
    {
     "name": "identity_cert:secret:default",
     "version_info": "uninitialized",
     "last_updated": "2025-05-28T03:16:40.579Z",
     "secret": {
      "@type": "type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret",
      "name": "identity_cert:secret:default"
     }
    },
    {
     "name": "mesh_ca:secret:default",
     "version_info": "uninitialized",
     "last_updated": "2025-05-28T03:16:40.579Z",
     "secret": {
      "@type": "type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret",
      "name": "mesh_ca:secret:default"
     }
    }
   ]
  }
```

```json
{
   "dynamic_active_secrets": [
    {
     "name": "identity_cert:secret:default",
     "version_info": "aab17341-71cd-4f7f-aced-f14b29630000",
     "last_updated": "2025-05-28T03:16:41.193Z",
     "secret": {
      "@type": "type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret",
      "name": "identity_cert:secret:default",
      "tls_certificate": {
       "certificate_chain": {
[...]

PR #13902, opened 2025-07-09
feat(kuma-dp): add identity cert readiness as part of the readiness check

jijiechen:
### What

Block data planes without identity cert ready in a mTLS enabled mesh

Details at https://github.com/kumahq/kuma/issues/13593

### How

**Option 1: Use the `kuma:envoy:admin` listener to push updates to the DP showing whether the `identity_cert` is well received.**

**Option 2: Request Envoy admin api and check if the `identity_cert:secret:*` is expected and ready/valid.**

Steps:

In the existing [ReadinessReporter](https://github.com/kumahq/kuma/blob/28b1b7562b6ccb010ab7540b69a9efbe2766e8aa/app/kuma-dp/pkg/dataplane/readiness/component.go#L93), introduce a new step to check readiness of `identity_cert`:

1. if the DP is terminating, don't check and return ready
2. if no admin port is defined on the dataplane, don't check and return ready
3. check if mesh has mTLS enabled and an `identity_cert` is expected by requesting admin API at `/config_dump?resource=dynamic_active_listeners&name_regex=kuma:envoy:admin`
4. check if `identity_cert` is ready and valid by requesting admin API at `/config_dump?resource=dynamic_active_secrets&name_regex=identity_cert:secret:.*`

Do check `identity_cert` every time when the readiness reporter is invoked by the `/ready` endpoint because we want to respond an accurate result of whether the DP is ready.

### Why

There are four kinds of state of `identity_cert`:
1. `N/A`: not expected, because the mesh does not have mTLS enabled
2. `uninitialized`: expected but not received yet
3. `Normal`: received and is valid
4. `Expired`: received but is expired 

So the first step is to confirm if the mesh has mTLS enabled and the `identity_cert` is expected, and then the next step is to check if the cert is ready and valid.

#### Checking if `identity_cert` is expected

Unfortunately, we don't have an existing way of getting the mTLS swtich from the kuma-dp process:
* it is dynamically changable, so it's not part of the DP file or metadata: we have to get it from the xDS. But, 
* it is not an existing field or annotation/tag/label in the xDS (We could potentially infer it from clusters by checking if `identity_cert` is used in `transport_socket`, but parsing/filtering clusters is not perormant, see "performance")
* it is not part of the metrics of the DP/Envoy

The idea here is to attach this information to the `kuma:envoy:admin` listener by using metadata: listeners are re-generated when user switches on or off mTLS on mesh and this triggers an SDS request from the DP to fetch the `identity_cert`. So checking the listener information is a good point to tell if the DP is waiting for the `identity_cert`.

Once this information is attached, we can check from the ReadinessReporter and tell whether the `identity_cert` is expected.

#### Checking if identity_cert is received and ready

By using the config dump endpoint provided by Envoy admin API, we can check if `identity_cert` is received. We can also get the certificate bytes and validate if it is expired.

My test shows that when the `identity_cert` is expected but not received, it will be shown as `uninitialized` when I do a full config dump. But when I request the single resource type by invoking `/config_dump?resource=dynamic_active_secrets&name_regex=identity_cert:secret:.*` at this time, uninitialized secrets are not returned.

When validate the validity of the `identity_cert`, we only validate the expiry and skip validting other items like CA and common name, etc. Those are ensured by the control plane.

#### Resilience and Performance

Introducing new HTTP requests into the readiness reporter can impact the stability and performance of the reporter, and when these requests are sending back to Envoy, it will further impact the performance of Envoy.

In Envoy, endpoint `/config_dump` is handled by the main working thread, and not the worker thread that serves external requests, so there is no loop-dependency in this scenario. However in a busy mesh, when we send many xDS updates and Envoy is busy handling these updates it may be stuck at responding the admin APIs.  This was reported and impacted overall usability of business applications before https://github.com/kumahq/kuma/issues/11034

We can set a 2-second timeout (which is shorter than the default timeout of the readiness check itself) to requests invoking the `/config_dump` endpoint, meaning that when Envoy is not able to respond in time, the whole DP can be marked as not ready. 

By using the `resource` and `name_regex` queries on the `/config_dump` endpoint, it significantly reduces the footprint of the response and it goes to a much faster branch inside Envoy.

https://github.com/envoyproxy/envoy/blob/main/source/server/admin/config_dump_handler.cc#L170-L175

## Supporting documentation

fixes https://github.com/kumahq/kuma/issues/13593

### DIFF
--- a/app/kuma-dp/pkg/dataplane/readiness/identity_cert_client.go
+++ b/app/kuma-dp/pkg/dataplane/readiness/identity_cert_client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	envoy_admin "github.com/envoyproxy/go-control-plane/envoy/admin/v3"
 	envoy_tls "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
@@ -93,9 +94,13 @@ func certSecretReady(resp []byte) (bool, error) {
 			return false, fmt.Errorf("failed to unmarshal tls secret: %w", err)
 		}
 
-		certBytes := tlsSecret.GetTlsCertificate().GetCertificateChain().GetInlineBytes()
+		pemBytes := tlsSecret.GetTlsCertificate().GetCertificateChain().GetInlineBytes()
+		pemBlock, _ := pem.Decode(pemBytes)
+		if pemBlock == nil {
+			return false, fmt.Errorf("invalid PEM block: %w", err)
+		}
 		var x509Cert *x509.Certificate
-		x509Cert, err = x509.ParseCertificate(certBytes)
+		x509Cert, err = x509.ParseCertificate(pemBlock.Bytes)
 		return x509Cert != nil && x509Cert.NotAfter.After(time.Now().UTC()), err
 	}
 }
--- a/app/kuma-dp/pkg/dataplane/readiness/identity_cert_client_test.go
+++ b/app/kuma-dp/pkg/dataplane/readiness/identity_cert_client_test.go
@@ -68,7 +68,7 @@ var _ = Describe("IdentityCertClient", func() {
 		ready, err := checkClient.CheckIfReady()
 
 		Expect(err).ToNot(HaveOccurred())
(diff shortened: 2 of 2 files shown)


jijiechen:
Closing as the solution needs to be redesigned.

Issue #13795, opened 2025-06-25
Define and register `feature-unified-resource-naming` feature flag


bartsmykla:
### Description

Introduce the `feature-unified-resource-naming` feature flag to enable consistent naming for non-system Envoy resources and stats. This includes both KRI-based formats and the `self_` format for inbounds. This task sets up core wiring to propagate the flag from config to xDS metadata for all proxy types.

### Acceptance criteria

* [x] Define `FeatureUnifiedResourceNaming` constant in `pkg/xds/types/features.go` with a clear comment
* [x] Add `UnifiedResourceNamingEnabled` to `DataplaneRuntime` config and wire it into `kuma-dp`
* [x] Ensure `kuma-dp` includes `feature-unified-resource-naming` in xDS metadata when enabled
* [ ] Ensure `ZoneIngress` and `ZoneEgress` proxies propagate the feature flag in both Universal and Kubernetes modes
* [ ] Ensure the control plane reads `feature-unified-resource-naming` from xDS metadata for all proxy types
* [ ] Add unit tests to cover propagation of the flag from config to xDS metadata and detection in the control plane

PR #13908, opened 2025-07-09
feat(xds): add feature flag for unified resource naming

bartsmykla:
## Motivation

Introduce the `feature-unified-resource-naming` feature flag to enable consistent and predictable naming of Envoy resources and stats across all proxy types. This PR sets up the necessary wiring between the control plane and data plane to support the flag, ensuring it can be configured and propagated correctly when enabled.

## Implementation information

- New `FeatureUnifiedResourceNaming` constant in `xds/types/features.go`
- Added `UnifiedResourceNamingEnabled` to `DataplaneRuntime` config, exposed via env var
- Propagate feature flag from `kuma-dp` config to xDS metadata
- Enable automatic injection of the feature in Kubernetes via the injector and container factory
- Updated `DataplaneProxyFactory` to handle the new flag
- Wired through Kubernetes plugin and injector setup

## Supporting documentation

Closes: https://github.com/kumahq/kuma/issues/13795

Related:
- [MADR-076 Standardized Naming for system xDS Resources](https://github.com/kumahq/kuma/blob/210d6f7a354bc279b0f2184bb4a4ae9229394b8c/docs/madr/decisions/076-naming-system-envoy-resources.md)
- [MADR-077 Defining and migrating to consistent naming for non-system Envoy resources and stats](https://github.com/kumahq/kuma/blob/dc2dd5564cac3f97f85085a8aa498f4a520ad4ba/docs/madr/decisions/077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md)

### DIFF
--- a/pkg/xds/bootstrap/generator_test.go
+++ b/pkg/xds/bootstrap/generator_test.go
@@ -221,10 +221,10 @@ var _ = Describe("bootstrapGenerator", func() {
 				return dp
 			},
 			request: types.BootstrapRequest{
-				Mesh:    "mesh",
-				Name:    "name.namespace",
-				Version: defaultVersion,
-				Workdir: "/tmp",
+				Mesh:     "mesh",
+				Name:     "name.namespace",
+				Version:  defaultVersion,
+				Workdir:  "/tmp",
 				Features: []string{xds_types.FeatureUnifiedResourceNaming},
 			},
 			expectedConfigFile: "generator.custom-config-minimal-request-with-unified-resource-naming-feature-flag.golden.yaml",


slonka (commented):
> I'm currently checking if we already have a way to verify that the feature flag is passed from the data plane to the control plane, or if it's feasible to add this kind of check as an integration test. That's why this PR is in draft

Should we check it the same way as "delta" - [https://github.com/kumahq/kuma/blob/76d9da887dc94394d1e24025da908772408619c5/pkg/x[…]a/generator.custom-config-minimal-request-and-delta.golden.yaml](https://github.com/kumahq/kuma/blob/76d9da887dc94394d1e24025da908772408619c5/pkg/xds/bootstrap/testdata/generator.custom-config-minimal-request-and-delta.golden.yaml#L52) ?

Issue #13885, opened 2025-07-08
Mismatch between struct field and JSON tag in `BootstrapRequest.AppProbeProxyEnabled`


bartsmykla:
### Describe the bug

In `BootstrapRequest`, the field is called `AppProbeProxyEnabled` but the JSON tag says `appProbeProxyDisabled`:
https://github.com/kumahq/kuma/blob/76d9da887dc94394d1e24025da908772408619c5/pkg/xds/bootstrap/types/bootstrap_request.go#L21

We should check which one is actually correct (enabled or disabled), make the name match, and update any related logic if needed.

lahabana:
If we can't do it in a clean manner we should:

1. Close the PR
2. Add a comment to this issue indicating why it's not possible
3. Add warning in the code if it's necessary to make it clear of this "edgy/complex" API to avoid accidentally breaking this (worth linking to this issue in the comments)
4. Close this

bartsmykla:
* `BootstrapRequest` uses the wire key `appProbeProxyDisabled` since 2.9 (public API). The old struct field was named `AppProbeProxyEnabled` but carried the **disabled** wire value due to the JSON tag mismatch. Both DP and CP shared that buggy shape, so behavior was consistent despite the confusing name

* The PR flips the struct to `AppProbeProxyDisabled` and updates logic/tests accordingly. This **changes semantics under version skew**:

  * **New DP → Old CP:** old CP deserializes `appProbeProxyDisabled: true` into its `AppProbeProxyEnabled = true` and treats the probe proxy as **enabled**, which is the opposite of what the DP intended
  * **Old DP → New CP:** new CP now reads `AppProbeProxyDisabled`, but old DPs still send the value the CP used to interpret as “enabled,” again inverting the behavior

* Kuma supports mixed-version upgrades (CP with older DPs). Because the wire key is already `appProbeProxyDisabled`, there’s **no additive, non-breaking way** to correct the field name/meaning without:

  * introducing a second field and complicated precedence rules, or
  * versioning the bootstrap endpoint and running a long deprecation cycle
    Either path still risks flips during rolling upgrades and violates the project’s backward-compat expectations. Reviewers explicitly called out the upgrade hazard, and the PR was closed with "no path forward."

* Net: fixing the name/meaning now would **flip behavior during CP↔DP skew** and can break probes/traffic during upgrades, so it’s not possible without breaking backward compatibility

PR #13909, opened 2025-07-09
fix(xds/bootstrap): rename field to match `appProbeProxyDisabled` JSON tag

bartsmykla:
The `BootstrapRequest` struct had a mismatch between the field name (`AppProbeProxyEnabled`) and its JSON tag (`appProbeProxyDisabled`). Since `appProbeProxyDisabled` is already part of the public API, this commit renames the field to match the tag.

Updated related logic and tests to use the new `AppProbeProxyDisabled` field name while preserving the same behavior.

## Supporting documentation

Closes: https://github.com/kumahq/kuma/issues/13885


app/kuma-dp/pkg/dataplane/envoy/remote_bootstrap.go
```diff
@@ -185,14 +185,14 @@ func (b *remoteBootstrapClient) requestForBootstrap(ctx context.Context, client
...
-		ReadinessPort:        opts.Config.Dataplane.ReadinessPort,
-		AppProbeProxyEnabled: opts.Config.ApplicationProbeProxyServer.Port > 0,
-		OperatingSystem:      b.operatingSystem,
-		Features:             features,
-		Resources:            resources,
-		Workdir:              opts.Config.DataplaneRuntime.SocketDir,
+		DynamicMetadata:       metadata,
+		DNSPort:               opts.Config.DNS.EnvoyDNSPort,
+		ReadinessPort:         opts.Config.Dataplane.ReadinessPort,
+		AppProbeProxyDisabled: opts.Config.ApplicationProbeProxyServer.Port == 0,
```
slonka, line 191: so this inverts

pkg/xds/bootstrap/generator.go
```diff
@@ -115,7 +115,7 @@ func (b *bootstrapGenerator) Generate(ctx context.Context, request types.Bootstr
 		DynamicMetadata:      request.DynamicMetadata,
 		DNSPort:              request.DNSPort,
 		ReadinessPort:        request.ReadinessPort,
-		AppProbeProxyEnabled: request.AppProbeProxyEnabled,
+		AppProbeProxyEnabled: !request.AppProbeProxyDisabled,
```
slonka: and this inverts, so it's double negation so it's the same

slonka (approved):

app/kuma-dp/pkg/dataplane/envoy/remote_bootstrap.go
slonka: it can't be negative because it's `uint32`

pkg/xds/bootstrap/generator.go
lahabana: Does it make sense to fix the parameter here as well? (TBH I have this configParameters struct it just seems useless...)

pkg/xds/bootstrap/types/bootstrap_request.go
```diff
@@ -14,18 +14,18 @@ type BootstrapRequest struct {
...
-	Resources            ProxyResources             `json:"resources"`
-	Workdir              string                     `json:"workdir"`
-	MetricsResources     MetricsResources           `json:"metricsResources"`
-	SystemCaPath         string                     `json:"systemCaPath"`
-	TransparentProxy     *tproxy_dp.DataplaneConfig `json:"dataplaneConfig,omitempty"`
+	CaCert                string                     `json:"caCert"`
+	DynamicMetadata       map[string]string          `json:"dynamicMetadata"`
+	DNSPort               uint32                     `json:"dnsPort,omitempty"`
+	ReadinessPort         uint32                     `json:"readinessPort,omitempty"`
+	AppProbeProxyDisabled bool                       `json:"appProbeProxyDisabled,omitempty"`
```
jijiechen, line 21: I think the only change you need to make is to change this json tag to align with the field name right?
The meaning of the field is not confused/error right now, the issue is only the json tag.

jijiechen (changes requested):
Can be implemented with much smaller footprint.

pkg/xds/bootstrap/types/bootstrap_request.go
lahabana: Changing the jsonTag is a breaking change. DP and CP run different versions so you'd break our version compat.

jijiechen: Yes, makes sense. Thanks.

jijiechen (dismissed):

pkg/xds/bootstrap/generator.go
bartsmykla: Agree with the comment about this struct to be useless and also it seems that changing it as well doesn't require lot of changes so I changed it as suggested.

pkg/xds/bootstrap/template_v3.go
```diff
@@ -352,7 +352,7 @@ func genConfig(parameters configParameters, proxyConfig xds.Proxy, enableReloada
 		res.Node.Metadata.Fields[core_xds.FieldDataplaneReadinessPort] = util_proto.MustNewValueForStruct(strconv.Itoa(int(parameters.ReadinessPort)))
 	}
 	// a request from an old DP will not include this field, so defaults to false
-	if parameters.AppProbeProxyEnabled {
+	if !parameters.AppProbeProxyDisabled {
```
slonka, line 355: Look at the comment above - doesn't it mean that during upgrade old proxies will always send `false` here so it will by default be `true`?

slonka (changes requested):
look at my previous comment, we might have a problem during upgrade

pkg/xds/bootstrap/template_v3.go
jijiechen: Yes and no, thanks for pointing this out. 

This code [was introduced in 2.9](https://github.com/kumahq/kuma/commit/002e301acf53484f3508a2a7e4ebce8e7df97f66), at that time the CP needs to keep the compatibility with DPs of versions 2.7 and 2.8. Now we are on a much higher version and [we are always setting this annotation](https://github.com/kumahq/kuma/blob/4479a75c816f16645d48e8d21336756783e57803/pkg/plugins/runtime/k8s/probes/pod_probe_overrider.go) since 2.9. 

So technically we can do this change now.

**Background**

The field was initially called `appProbeProxyDisabled`, and it was then renamed to `appProbeProxyEnabled`, but unfortunately I forgot updating the json tag.

lahabana:
Closing this as we agreed there's no path forward. @bartsmykla needs to update the issue with some explanation.

Issue #13875, opened 2025-07-07
Create a new model for defining different datasources


lukidzi:
### Description

We have a Datasource model in kuma which supports `Inline`, `InlineString`, and `Secret`. In the [MADR](https://github.com/kumahq/kuma/pull/13804) we agreed that we need a new model for support
- Secret
- File
- EnvVar

PR #13910, opened 2025-07-09
docs(MADR): introduction of common DataSource struct

lukidzi:
## Motivation

As discussed in #13804 we want to introduce common DataSource structure

Here I discuss how should it be designed and potential issues

xref: https://github.com/kumahq/kuma/issues/13875

Rendered: https://github.com/lukidzi/kuma/blob/madr-datasource/docs/madr/decisions/081-datasource-api.md

### DIFF
--- a/docs/madr/decisions/081-datasource-api.md
+++ b/docs/madr/decisions/081-datasource-api.md
@@ -86,13 +86,17 @@ This ambiguity in resolution context leads to confusion and potential misconfigu
 
 ## User stories
 
-### As a Mesh Operator, I want to provide a CA and private key for `MeshIdentity` using a file or environment variable
+### As a Mesh Operator, I want to provide a CA and private key for `MeshIdentity` using a file, environment variable or secret
 
-Mesh Operator should be able to configure `MeshIdentity` by referencing a file path or environment variable instead of using inline data.
+Mesh Operators should be able to configure a `MeshIdentity` by referencing a file path, environment variable, or `Secret`. This avoids the need to embed sensitive data inline, improving security and aligning with best practices for secret management.
 
 ### As a Mesh Operator, I want to configure mTLS for MeshExternalService securely
 
 It should be possible to configure a specific `MeshExternalService` to use a certificate and private key mounted from a file or injected via an environment variable on the `ZoneEgress` - without transmitting them over the network.
+
+### As a Mesh Operator, I want to configure TLS for MeshExternalService
+
+It should be possible to configure only the CA bundle (inline) for verifying the upstream server, as it is safe to expose this information to users.
 
 ### As a Mesh Operator, I want to define a MeshExternalService with mTLS settings once and use it across all zones
 
@@ -312,6 +316,156 @@ The proposed model introduces clear semantics for resource resolution, allowing 
 #### Cons
 * Two separate models, which may introduce some code duplication and increase maintenance overhead.
 
+### Option 4: Separate Structures for Secrets and Other Data Sources with an Resolution Context directly in the resource
+
+The idea is to avoid embedding the resolution logic (i.e., where the resource should be resolved) into the `DataSource` itself. Instead, this responsibility should remain with the resource that uses it. This way, we keep the `DataSource` object clean and focused.
+
+```golang
+// +kubebuilder:validation:Enum=File;Secret;EnvVar;InsecureInline
+type SecretDataSourceType string
+
+const (
+	SecretDataSourceFile      SecretDataSourceType = "File"
+	SecretDataSourceSecretRef SecretDataSourceType = "Secret"
+	SecretDataSourceEnvVar    SecretDataSourceType = "EnvVar"
+	SecretDataSourceInline    SecretDataSourceType = "InsecureInline"
+)
+
+type SecretDataSource struct {
+	Type           SecretDataSourceType `json:"type,omitempty"`
+	File           *File                `json:"file,omitempty"`
+	SecretRef      *SecretRef           `json:"secretRef,omitempty"`
+	EnvVar         *EnvVar              `json:"envVar,omitempty"`
+	InsecureInline *Inline              `json:"insecureInline,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=File;EnvVar;Inline
+type DataSourceType string
+
+const (
+	DataSourceFile   DataSourceType = "File"
+	DataSourceEnvVar DataSourceType = "EnvVar"
+	DataSourceInline DataSourceType = "InsecureInline"
+)
+
+type DataSource struct {
+	Type   DataSourceType `json:"type,omitempty"`
+	File   *File          `json:"file,omitempty"`
+	EnvVar *EnvVar        `json:"envVar,omitempty"`
+	Inline *Inline        `json:"inline,omitempty"`
+}
+
+type File struct {
+	Path string `json:"path,omitempty"`
+}
+
+type EnvVar struct {
+	Name string `json:"name,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=Secret
+type RefType string
+
+const (
+	SecretRefType RefType = "Secret"
+)
+
+type SecretRef struct {
+	Kind RefType `json:"kind,omitempty"`
+	Name string  `json:"name,omitempty"`
+}
+
+type Inline struct {
+	Value string `json:"value,omitempty"`
+}
+```
+
+#### How would it looks for `MeshExternalService`?
+
+> [!NOTE]
+> This is not a finalized design for `MeshExternalService` and should be covered by a separate design document.
+
+For `MeshExternalService`, we could introduce an additional field in the certificate definition, such as `type`, with values like `ZoneEgress` or `ControlPlane`. This field would indicate where the resource should be resolved.
+
+By doing so, we can:
+
+* Keep the DataSource structure small and focused.
+* Avoid embedding resolution logic directly inside `DataSource`.
+* Allow individual resources (like `MeshExternalService`) to define resolution semantics explicitly and contextually.
+
+This approach promotes cleaner separation of concerns and aligns better with the needs of different resource types.
+
+```yaml
+      clientCert:
+        type: ControlPlane
+        controlPlane:
+          type: File | EnvVar | Secret | InsecureInline
+          file:
+            path: "/cert"
+      clientKey:
+        zoneEgress:
+          type: EnvVar
+          envVar:
+            name: "PRIVATE_KEY"
+```
+or
+```yaml
+    auth:  
+      type: ZoneEgress | ControlPlane
+      zoneEgress:
+        clientCert:
+          type: File
+          file:
+            path: "/cert"
+        clientKey:
+          type: EnvVar
+          envVar:
+            name: "PRIVATE_KEY"
+      controlPlane:
+        clientCert:
+          type: File
+          file:
+            path: "/cert"
+        clientKey:
+          type: EnvVar
+          envVar:
+            name: "PRIVATE_KEY"
+```
+
+In the case of `MeshIdentity`, we don’t need a `type` field because the data is always accessed by the Control Plane, and there’s no need to control or vary the resolution behavior.
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshIdentity
+metadata:
+  name: identity
+  namespace: kuma-system # only in system namespace
+  labels:
+    kuma.io/mesh: default
+spec:
+  ...
+  provider:
+    type: Bundled
+    bundled: # to extend in KM
+      ca:
+        certificate:
+          type: EnvVar
+          envVar:
+            name: "PRIVATE_KEY"
+        privateKey:
+          type: EnvVar
+          envVar:
+            name: "PRIVATE_KEY"
+```
+
+#### Pros
+* Clear API boundaries - users can explicitly define where the resource should be resolved (control plane or data plane).
+* Inline secrets are clearly marked as insecure, improving security visibility and awareness.
+* Resolution logic is embedded in the object itself, making it clear what is supported and expected for each resource.
+
+#### Cons
+* Two separate models, which may introduce some code duplication and increase maintenance overhead.
+
 ## Security implications and review
 
 Following changes increases security:
@@ -326,4 +480,4 @@ Following changes increases security:
 
 ## Decision
 
-Option 3 offers the most flexibility, enabling users to explicitly choose whether a value is resolved by the control plane or accessed directly by the data plane. Additionally, by providing two separate models, it becomes easier to understand which options are available and valid in each API context.
+Option 4 offers the most flexibility, enabling users to explicitly (on specific resource) choose whether a value is resolved by the control plane or accessed directly by the data plane. Additionally, by providing two separate models, it becomes easier to understand which options are available and valid in each API context.
--- a/docs/madr/decisions/081-datasource-api.md
+++ b/docs/madr/decisions/081-datasource-api.md
@@ -325,19 +325,19 @@ The idea is to avoid embedding the resolution logic (i.e., where the resource sh
 type SecureDataSourceType string
 
 const (
-	CPDataSourceFile      SecureDataSourceType = "File"
-	CPDataSourceSecretRef SecureDataSourceType = "Secret"
-	CPDataSourceEnvVar    SecureDataSourceType = "EnvVar"
-	CPDataSourceInline    SecureDataSourceType = "InsecureInline"
+	SecureDataSourceFile      SecureDataSourceType = "File"
+	SecureDataSourceSecretRef SecureDataSourceType = "Secret"
+	SecureDataSourceEnvVar    SecureDataSourceType = "EnvVar"
+	SecureDataSourceInline    SecureDataSourceType = "InsecureInline"
 )
 
 // SecureDataSource is a way to securely provide data to the component
 type SecureDataSource struct {
-	Type              SecureDataSourceType `json:"type,omitempty"`
-	File              *File                `json:"file,omitempty"`
-	EnvVar            *EnvVar              `json:"envVar,omitempty"`
-	InsecureInline    *Inline              `json:"insecureInline,omitempty"`
-	SecretRef         *SecretRef           `json:"secretRef,omitempty"`
+	Type           SecureDataSourceType `json:"type,omitempty"`
+	File           *File                `json:"file,omitempty"`
+	EnvVar         *EnvVar              `json:"envVar,omitempty"`
+	InsecureInline *Inline              `json:"insecureInline,omitempty"`
+	SecretRef      *SecretRef           `json:"secretRef,omitempty"`
 }
 // +kubebuilder:validation:Enum=File;EnvVar
 type DataSourceType string
@@ -345,8 +345,11 @@ type DataSourceType string
 const (
 	DataSourceFile   DataSourceType = "File"
 	DataSourceEnvVar DataSourceType = "EnvVar"
+  DataSourceEnvVar DataSourceType = "Inline"
 )
 
+// DataSource is just a way to provide data. Not necessarily secrets, 
+// can be any data, i.e. certs, configs, OPA policies written in rego, lua plugins etc.
 type DataSource struct {
 	Type   DataSourceType `json:"type,omitempty"`
 	File   *File          `json:"file,omitempty"`
--- a/docs/madr/decisions/081-datasource-api.md
+++ b/docs/madr/decisions/081-datasource-api.md
@@ -8,7 +8,7 @@ Technical Story: https://github.com/kumahq/kuma/issues/13875
 
 While working on SPIFFE compliance, we identified limitations in our existing `DataSource` structure. It currently supports only inline values or secret references:
 
-[Source](https://github.com/kumahq/kuma/blob/master/api/common/v1alpha1/datasource.go)
+[Source](https://github.com/kumahq/kuma/blob/4ac717ba99aae596da3114455d4ba8967e084191/api/common/v1alpha1/datasource.go#L10-L18)
 ```golang
 // DataSource defines the source of bytes to use.
 type DataSource struct {
@@ -25,10 +25,7 @@ The current solution has following limitations:
 * Users cannot provide values via file path or environment variable.
 * Defining private keys inline is insecure, especially since these values are synced to the global control plane.
 
-
-There is another issue when user wants to explicitly use the `DataSource` in the resource. It's not clear if value provided
-by the data source should be accessed by the data plane or control plane. In some resources it makes sense to have it on the control plane
-while for others on the data plane.
+There is another issue when user wants to explicitly use the `DataSource` in the resource. It's not clear if value provided by the data source should be accessed by the proxy or control plane. In some resources it makes sense to have it on the control plane while for others on the proxy.
 
 **Example**
 
@@ -82,7 +79,7 @@ This ambiguity in resolution context leads to confusion and potential misconfigu
 * Create a reusable structure applicable across different resources.
 * Clearly define where the `DataSource` should be resolved:
   * On the control plane (e.g., secrets)
-  * On the data plane (e.g., files or environment variables)
+  * On the proxy (data plane, zone ingress, or zone egress) via files or environment variables
 
 ## User stories
 
@@ -108,7 +105,7 @@ It should be feasible to define a `MeshExternalService` with all required mTLS c
 
 ### As a Mesh Operator, I want to configure mTLS for MeshExternalService using a Secret
 
-To maintain a strong security boundary, the control plane should only access `Secrets` from the system namespace (e.g., kuma-system). When a `MeshExternalService` is configured with a `Secret`, it makes sense for the control plane to read that `Secret` and securely deliver it to the data plane via SDS.
+To maintain a strong security boundary, the control plane should only access `Secrets` from the system namespace (e.g., kuma-system). When a `MeshExternalService` is configured with a `Secret`, it makes sense for the control plane to read that `Secret` and securely deliver it to the proxy via SDS.
 
 ## Design
 
@@ -155,7 +152,7 @@ type EnvVar struct {
 This model is extensible and aligns with Kubernetes conventions (e.g., `SecretRef`). However, it introduces ambiguity regarding where the value should be resolved.
 To address this, we could introduce a resource-specific `ResourceDescriptor` parameter that defines where the `DataSource` should be resolved:
 * Control plane - for example, when signing certificates in `MeshIdentity`.
-* Data plane - for example, when performing mutual TLS in `MeshExternalService` via `ZoneEgress`.
+* Proxy (data plane, zone ingress, or zone egress) - for example, when doing mTLS for `MeshExternalService` via `ZoneEgress`
 
 #### Pros
 * Provides a unified structure.
@@ -163,8 +160,8 @@ To address this, we could introduce a resource-specific `ResourceDescriptor` par
 * Inline secrets are no longer synced to the global control plane, improving security.
 
 #### Cons
-* Responsibility for resolving the configuration (control plane vs. data plane) is ambiguous.
-* It's unclear where values like file paths or environment variables should be configured - in the control plane or data plane context.
+* Responsibility for resolving the configuration (control plane vs. proxy) is ambiguous.
+* It's unclear where values like file paths or environment variables should be configured - in the control plane or proxy context.
 * There's a potential security risk if private keys are delivered by the control plane.
 * The API is not self-explanatory: the resolution behavior is only discoverable by reading the code.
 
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,256 @@
...
+* Clear API boundaries — it’s explicitly defined what can be configured for each resource type.
+* Improved security — avoids delivering sensitive material like private keys to the data plane via SDS.
+* Inline secrets are not synced to the global anymore
+
+#### Cons
+* Two similar but separate structures — which adds some duplication and potential maintenance overhead.
+* No explicit support for Secret in data plane context — users would need to manually mount secrets as files or environment variables.
+* Lack of inline
+* Might be problem with rotation - Envoy might not read if it's rotated - might be support by: https://www.envoyproxy.io/docs/envoy/latest/configuration/security/secret#key-rotation
+
```
lahabana, line 467: Any alternative options?

Thinking about it:
Secret makes little sense on the DP (except if we use the serviceaccount token to retrieve the secret on the dp but that looks weird).
So we could for example have a resolveMode: dp|cp on the file and env var datasource?
We could then depending on the use allow only one mode?

I'd love us to come up with some alternatives and probably also clarify what is it we're trying to achieve here.

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,256 @@
...
+  * EnvVar and File result in a cluster referencing the path or environment variable on the `ZoneEgress` (pod/VM),
+  * Secret is read by the control plane and passed via SDS.
+
+This results in a divergence in behavior across resources. For example:
+* In MeshIdentity, the control plane always reads the source and delivers it.
+* In MeshExternalService, it’s unclear whether the control plane or the data plane (`ZoneEgress`) is responsible.
+
+**How can we solve it?**
+
+To resolve the ambiguity around how and where a `DataSource` is accessed (i.e., control plane vs data plane), we could introduce a resource specific ResourceDescriptor parameter. This parameter would define whether the `DataSource` should be:
```
slonka: would that parameter be somewhere on the data source? something similar to:

```
type: File
resolve: onCp | onDp 
file:
  path: "/etc/ssl/certs/my-cert.crt"
```

? we could then validate it on specific resource (MES / MID)

docs/madr/decisions/081-datasource-api.md
slonka: Is this the same thing I'm suggesting here https://github.com/kumahq/kuma/pull/13910#discussion_r2196931897 ? 🤔

docs/madr/decisions/081-datasource-api.md
lukidzi: Added a 3rd option

docs/madr/decisions/081-datasource-api.md
lukidzi: Added a goal.

I feel that once I've added datasource on MeshExternalService it wasn't clear for me if I should load file on control-plane or maybe I should mount it to the ZoneEgress. Also, I can see potential of having similar issue in different policies

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+    Type      Type       `json:"type,omitempty"`
+    File      *File      `json:"file,omitempty"`
+    SecretRef *SecretRef `json:"secretRef,omitempty"`
+    EnvVar    *EnvVar    `json:"envVar,omitempty"`
+}
+
+type ResolveMode string
+
+const (
+    CPResolveMode ResolveMode = "Cp"
```
lahabana: ControlPlane or DataPlane no?

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+    DPResolveMode ResolveMode = "Dp"
+)
+
+type File struct {
+    ResolveMode ResolveMode `json:"resolveMode,omitempty"`
```
lahabana: ResolveOn maybe? I can't find a good name here.

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+#### Pros
+* Clear API boundaries — user can configure where the resource should be loaded
+* Inline secrets are not synced to the global anymore
+
+#### Cons
+* Additional parameter which needs to be set by the user
+
+## Security implications and review
+
+Following changes increses security:
```
lahabana:
```suggestion
Following changes increases security:
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+* secrets are not more visiable in global control-plane
```
lahabana:
```suggestion
* secrets are not more visible in global control-plane
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+
+## Reliability implications
+
+## Implications for Kong Mesh
+
+`MeshOPA` uses current API, if we decide to change current we should probably update it (breaking change), or just maintain current.
+
+## Decision
+
+Option 3 offers the most flexibility, enabling users to explicitly choose whether a value is resolved by the control plane or accessed directly by the data plane.
```
lahabana: You say Option 3 and there's no number next to the alternatives. This is not very nice to the reader

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+* Improved security — avoids delivering sensitive material like private keys to the data plane via SDS.
+* Inline secrets are not synced to the global anymore
+
+#### Cons
+* Two similar but separate structures — which adds some duplication and potential maintenance overhead.
+* No explicit support for Secret in data plane context — users would need to manually mount secrets as files or environment variables.
+* Lack of inline
+* Might be problem with rotation - Envoy might not read if it's rotated - might be support by: https://www.envoyproxy.io/docs/envoy/latest/configuration/security/secret#key-rotation
+
+### Use one DataSource with resolveMode
```
lahabana:
```suggestion
### 3. Use one DataSource with resolveMode
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+* Can be reused consistently across different resources
+* Inline secrets are not synced to the global anymore
+
+#### Cons
+* Unclear who is responsible for delivering the configuration (control plane or ZoneEgress)
+* Not obvious where values like file paths or environment variables need to be set (control plane or data plane context)
+* Potential security risk: private keys might be delivered by the control plane
+
+
+### Create different DataSources based on where should be accessed
```
lahabana:
```suggestion
### 2. Create different DataSources based on where should be accessed
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+
+The main goals is:
+* Provide a reusable structure that can be consistently applied across different resources.
+* Clearly define whether the data source should be:
+  * Loaded by the data plane (e.g., via `File` or `EnvVar`), or
+  * Delivered by the control plane (e.g., via `Secret` or other resolved sources).
+
+## Design
+
+### Create a one DataSource structure
```
lahabana:
```suggestion
### 1. Expand existing DataSource structure and rely on where the DS is used
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+	Name string `json:"name,omitempty"`
+}
+```
+
+This model is extensible and allows defining one of several specific sources. It also follows the Kubernetes convention by referencing other resources using `SecretRef`.
+
+> Do we need a namespace?
+Since we only list secrets from the system namespace (default: `kuma-system`), we likely don't need the namespace field for now.
+
+**Problem**
```
lahabana: This seems to need to go in the context and problem statement no?

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+    FileType   Type = "File"
+    SecretType Type = "Secret"
+    EnvVarType Type = "EnvVar"
+)
+
+type ControlPlaneDataSource struct {
+    Type      Type       `json:"type,omitempty"`
+    File      *File      `json:"file,omitempty"`
+    SecretRef *SecretRef `json:"secretRef,omitempty"`
+    EnvVar    *EnvVar    `json:"envVar,omitempty"`
```
lahabana: I think we need to keep inline for back compat

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+// +kubebuilder:validation:Enum=File;Secret;EnvVar
+type Type string
+
+const (
+    FileType   Type = "File"
+    SecretType Type = "Secret"
+    EnvVarType Type = "EnvVar"
+)
+
+type ControlPlaneDataSource struct {
```
lahabana:
```suggestion
type DataSource struct {
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+
+Following changes increses security:
+* secrets are not more visiable in global control-plane
+* In case of 2nd option, private key is not delivered by the network to the dataplane
+
+## Reliability implications
+
+## Implications for Kong Mesh
+
+`MeshOPA` uses current API, if we decide to change current we should probably update it (breaking change), or just maintain current.
```
lahabana: How is it a breaking change? Can a Default when resolveMode is set fix this?

This is only a problem for MeshOPA? Feels like it's also a problem for MES no?

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+    File      *File      `json:"file,omitempty"`
+    SecretRef *SecretRef `json:"secretRef,omitempty"`
+    EnvVar    *EnvVar    `json:"envVar,omitempty"`
+}
+
+type ResolveMode string
+
+const (
+    CPResolveMode ResolveMode = "Cp"
+    DPResolveMode ResolveMode = "Dp"
```
lahabana: What's the default when it's unset?

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,318 @@
...
+)
+
+type File struct {
+    ResolveMode ResolveMode `json:"resolveMode,omitempty"`
+    Path        string      `json:"path,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=Secret
```
lahabana, line 290: Should this same annotation be present on resolveMode?

docs/madr/decisions/081-datasource-api.md
lukidzi: I think it should be ControlPlane

docs/madr/decisions/081-datasource-api.md
lukidzi: MES is experimental, while MeshOPA not. Also, we want to introduce `inline` while MeshOPA has also access to `inlineString`

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,307 @@
...
+type DataSource struct {
+	Type   DataSourceType `json:"type,omitempty"`
+	File   *File          `json:"file,omitempty"`
+	EnvVar *EnvVar        `json:"envVar,omitempty"`
+	Inline *Inline        `json:"inline,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=ControlPlane;DataPlane
+// default: ControlPlane
+type ResolveOn string
```
lobkovilya, line 273: TBH I don't like the `ResolveOn` field. It doesn't feel like something user should decide, the resolution mode is likely implied by the place where the DataSource is used:
* CA on MeshIdentity is always read by the Control Plane, it doesn't make any sense to read CA private key on the DPP
* MeshExternalService's private key should be read by the ZoneEgress, otherwise we're transferring private key over the xDS which is not ideal

lobkovilya (commented):
Maybe this MADR is missing user stories i.e. in which cases user wants resolution on CP and in which on DP. And is there a case when user actually wants to switch between CP and DP modes

docs/madr/decisions/081-datasource-api.md
```diff
@@ -312,6 +316,154 @@ The proposed model introduces clear semantics for resource resolution, allowing
...
+
+For `MeshExternalService`, we could introduce an additional field in the certificate definition, such as `type`, with values like `ZoneEgress` or `ControlPlane`. This field would indicate where the resource should be resolved.
+
+By doing so, we can:
+
+* Keep the DataSource structure small and focused.
+* Avoid embedding resolution logic directly inside `DataSource`.
+* Allow individual resources (like `MeshExternalService`) to define resolution semantics explicitly and contextually.
+
+This approach promotes cleaner separation of concerns and aligns better with the needs of different resource types.
```
slonka, line 395: I like this approach but I'm not sure about the names, they sort of apply that it's some sort of CP cert (which is not). Since it's the responsibility of the resource to name these I think we don't have to focus on this now.

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,481 @@
...
+type ControlPlaneDataSourceType string
+
+const (
+	CPDataSourceFile      ControlPlaneDataSourceType = "File"
+	CPDataSourceSecretRef ControlPlaneDataSourceType = "Secret"
+	CPDataSourceEnvVar    ControlPlaneDataSourceType = "EnvVar"
+	CPDataSourceInline    ControlPlaneDataSourceType = "InsecureInline"
+)
+
+type ControlPlaneDataSource struct {
```
lobkovilya: I'm not sure if this one should be called `ControlPlaneDataSource`. I like your previous suggestion `SecretDataSource` more. Because it's not really about CP or not CP, if user inlines private key it's considered insecure (even if DP reads it), but if they inline cert it's fine. 

Alternatively, we can call it either `SecureDataSource`, `ProtectedDataSource` or `SensitiveDataSource`

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,481 @@
...
+
+// +kubebuilder:validation:Enum=File;EnvVar
+type DataSourceType string
+
+const (
+	DataSourceFile   DataSourceType = "File"
+	DataSourceEnvVar DataSourceType = "EnvVar"
+)
+
+type DataSource struct {
```
lobkovilya, line 350: Is this one missing `Secret` and `Inline`? I thought option 4 is the same as option 3 but without `ResolveOn`

lukidzi: I was thinking about OpenAPI spec, and I think we need something like: 
* ControlPlaneDataSource
* SecureDataSource
* DataplaneDataSource
Why? 

on ZoneEgress it doesn't make sense to define `Secret` or `Inline` so we don't want to pollute OpenAPI schema. In case of ControlPlane it should be possible to define them that's why it makes sense to include them there. Also, there might be case for Secured and not secured. WDYT?

lobkovilya: > on ZoneEgress it doesn't make sense to define Secret or Inline so we don't want to pollute OpenAPI schema

I wouldn't say it doesn't make sense as if it's something insane. Technically we could've add code to kuma-dp that'd access Kubernetes API to get the `Secret`. It's just that we don't want to do that, so we can validate and return `type "Secret" is not supported with ZoneEgress mode`.

I think we need just 2 types:
```go
// DataSource is just a way to provide data. Not necessarily secrets, 
// can be any data, i.e. certs, configs, OPA policies written in rego, lua plugins etc.
type DataSource struct {
	Type   DataSourceType `json:"type,omitempty"`
	File   *File          `json:"file,omitempty"`
	EnvVar *EnvVar        `json:"envVar,omitempty"`
	Inline *Inline        `json:"inline,omitempty"`
}

// SecureDataSource is a way to securely provide data to the component
type SecureDataSource struct {
	Type              SecureDataSourceType `json:"type,omitempty"`
	File              *File                `json:"file,omitempty"`
	EnvVar            *EnvVar              `json:"envVar,omitempty"`
	InsecureInline    *Inline              `json:"insecureInline,omitempty"`
	SecretRef         *SecretRef           `json:"secretRef,omitempty"`
}
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/13875
+
+## Context and Problem Statement
+
+While working on SPIFFE compliance, we identified limitations in our existing `DataSource` structure. It currently supports only inline values or secret references:
+
+[Source](https://github.com/kumahq/kuma/blob/master/api/common/v1alpha1/datasource.go)
```
bartsmykla: Nit: It's better to use permalink as this link will be invalid when the structure will change in the future
 ```suggestion
[Source](https://github.com/kumahq/kuma/blob/4ac717ba99aae596da3114455d4ba8967e084191/api/common/v1alpha1/datasource.go#L10-L18)
```

File: docs/madr/decisions/081-datasource-api.md [hunk 22]
```diff
@@ -0,0 +1,485 @@
...
+	Inline *[]byte `json:"inline,omitempty"`
+	// Data source is inline string`
+	InlineString *string `json:"inlineString,omitempty"`
+}
+```
+
+The current solution has following limitations:
+* Users cannot provide values via file path or environment variable.
+* Defining private keys inline is insecure, especially since these values are synced to the global control plane.
+
```
bartsmykla, line 27: Nit:
```suggestion
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+        secret: my-secret
+```
+
+This ambiguity in resolution context leads to confusion and potential misconfiguration, particularly when reusing the same structure across diverse resource types.
+
+**Goal**
+* Create a reusable structure applicable across different resources.
+* Clearly define where the `DataSource` should be resolved:
+  * On the control plane (e.g., secrets)
+  * On the data plane (e.g., files or environment variables)
```
bartsmykla: Nit: Wdyt?
```suggestion
  * On the proxy (data plane, zone ingress, or zone egress) via files or environment variables
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+
+type EnvVar struct {
+    Name string `json:"name,omitempty"`
+}
+```
+
+This model is extensible and aligns with Kubernetes conventions (e.g., `SecretRef`). However, it introduces ambiguity regarding where the value should be resolved.
+To address this, we could introduce a resource-specific `ResourceDescriptor` parameter that defines where the `DataSource` should be resolved:
+* Control plane - for example, when signing certificates in `MeshIdentity`.
+* Data plane - for example, when performing mutual TLS in `MeshExternalService` via `ZoneEgress`.
```
bartsmykla: Nit: Wdyt?
```suggestion
* Proxy (data plane, zone ingress, or zone egress) - for example, when doing mTLS for `MeshExternalService` via `ZoneEgress`
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+
+#### Pros
+* Provides a unified structure.
+* Can be reused consistently across different resources.
+* Inline secrets are no longer synced to the global control plane, improving security.
+
+#### Cons
+* Responsibility for resolving the configuration (control plane vs. data plane) is ambiguous.
```
bartsmykla: Nit:
```suggestion
* Responsibility for resolving the configuration (control plane vs. proxy) is ambiguous.
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+* It's unclear where values like file paths or environment variables should be configured - in the control plane or data plane context.
```
bartsmykla:
```suggestion
* It's unclear where values like file paths or environment variables should be configured - in the control plane or proxy context.
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+)
+
+type SecretRef struct {
+    Kind RefType `json:"kind,omitempty"`
+    Name string  `json:"name,omitempty"`
+}
+
+type EnvVar struct {
+    Name string `json:"name,omitempty"`
+}
```
bartsmykla: Nit:
```suggestion
// +kubebuilder:validation:Enum=File;Secret;EnvVar
type ControlPlaneType string

const (
	FileType   ControlPlaneType = "File"
	SecretType ControlPlaneType = "Secret"
	EnvVarType ControlPlaneType = "EnvVar"
)

type ControlPlaneDataSource struct {
	Type      ControlPlaneType `json:"type,omitempty"`
	File      *File            `json:"file,omitempty"`
	SecretRef *SecretRef       `json:"secretRef,omitempty"`
	EnvVar    *EnvVar          `json:"envVar,omitempty"`
}

// +kubebuilder:validation:Enum=File;EnvVar
type ProxyType string

const (
	FileType   ProxyType = "File"
	EnvVarType ProxyType = "EnvVar"
)

type ProxyDataSource struct {
	Type   ProxyType `json:"type,omitempty"`
	File   *File     `json:"file,omitempty"`
	EnvVar *EnvVar   `json:"envVar,omitempty"`
}

type File struct {
	Path string `json:"path,omitempty"`
}

// +kubebuilder:validation:Enum=Secret
type RefType string

const (
	SecretRefType RefType = "Secret"
)

type SecretRef struct {
	Kind RefType `json:"kind,omitempty"`
	Name string  `json:"name,omitempty"`
}

type EnvVar struct {
	Name string `json:"name,omitempty"`
}
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+```
+
+The proposed model addresses the issues of the previous approach by clearly specifying where the `DataSource` should be resolved - either on the control plane or the data plane. It also avoids exposing unsupported options, such as referencing a `Secret` in contexts where it cannot be accessed by the data plane.
```
bartsmykla: Nit:
```suggestion
The proposed model addresses the issues of the previous approach by clearly specifying where the `DataSource` should be resolved - either on the control plane or the proxy. It also avoids exposing unsupported options, such as referencing a `Secret` in contexts where it cannot be accessed by the proxy.
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+
+#### Pros
+* Clear API boundaries — it’s explicitly defined what can be configured for each resource type.
+* Improved security — avoids delivering sensitive materials like private keys to the data plane via SDS.
```
bartsmykla:
```suggestion
* Improved security — avoids delivering sensitive materials like private keys to the proxy via SDS.
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+* Inline secrets are no longer synced to the global control plane.
+
+#### Cons
+* Requires maintaining two similar but separate structures, which introduces duplication and potential maintenance overhead.
+* No explicit support for Secret in the data plane context — users must manually mount secrets as files or environment variables.
```
bartsmykla:
```suggestion
* No explicit support for Secret in the proxy context — users must manually mount secrets as files or environment variables.
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// +kubebuilder:validation:Enum=ControlPlane;DataPlane
+// default: ControlPlane
+type ResolveOn string
+
+const (
+	ResolveOnCP ResolveOn = "ControlPlane"
+	ResolveOnDP ResolveOn = "DataPlane"
+)
```
bartsmykla: Nit:
```suggestion
// +kubebuilder:validation:Enum=ControlPlane;Proxy
// default: ControlPlane
type ResolveOn string

const (
	ResolveOnControlPlane ResolveOn = "ControlPlane"
	ResolveOnProxy        ResolveOn = "Proxy"
)
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+	Kind RefType `json:"kind,omitempty"`
+	Name string  `json:"name,omitempty"`
+}
+
+type Inline struct {
+	Value string `json:"value,omitempty"`
+}
+```
+
+The proposed model introduces clear semantics for resource resolution, allowing users to specify whether a file (or other data source) should be loaded by the control plane or the data plane. Additionally, it separates secret-related data sources from general ones, making it clearer which options are allowed in which contexts.
```
bartsmykla: Nit:
```suggestion
The proposed model introduces clear semantics for resource resolution, allowing users to specify whether a file (or other data source) should be loaded by the control plane or the proxy. Additionally, it separates secret-related data sources from general ones, making it clearer which options are allowed in which contexts.
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+	InsecureInline *Inline              `json:"insecureInline,omitempty"`
+	SecretRef      *SecretRef           `json:"secretRef,omitempty"`
+}
+// +kubebuilder:validation:Enum=File;EnvVar
+type DataSourceType string
+
+const (
+	DataSourceFile   DataSourceType = "File"
+	DataSourceEnvVar DataSourceType = "EnvVar"
+  DataSourceEnvVar DataSourceType = "Inline"
```
bartsmykla: Nit:
```suggestion
	DataSourceEnvVar DataSourceType = "Inline"
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+          envVar:
+            name: "PRIVATE_KEY"
+        privateKey:
+          type: EnvVar
+          envVar:
+            name: "PRIVATE_KEY"
+```
+
+#### Pros
+* Clear API boundaries - users can explicitly define where the resource should be resolved (control plane or data plane).
```
bartsmykla: Nit:
```suggestion
* Clear API boundaries - users can explicitly define where the resource should be resolved (control plane or proxy).
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+#### Cons
+* Two separate models, which may introduce some code duplication and increase maintenance overhead.
+
+## Security implications and review
+
+Following changes increases security:
+* Inline secrets are explicitly marked as `insecure`, helping to discourage their use in production environments.
+* In Option 2, the private key is not transmitted over the network to the data plane.
+
+## Reliability implications
```
bartsmykla: Nit:
```suggestion
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+
+## Implications for Kong Mesh
+
+`MeshOPA` currently uses the existing API. If we decide to change it, we should either update `MeshOPA` accordingly (introducing a breaking change), or maintain the existing API for backward compatibility.
+
+## Decision
+
+Option 4 offers the most flexibility, enabling users to explicitly (on specific resource) choose whether a value is resolved by the control plane or accessed directly by the data plane. Additionally, by providing two separate models, it becomes easier to understand which options are available and valid in each API context.
```
bartsmykla: Nit:
```suggestion
Option 4 offers the most flexibility, enabling users to explicitly (on specific resource) choose whether a value is resolved by the control plane or accessed directly by the proxy. Additionally, by providing two separate models, it becomes easier to understand which options are available and valid in each API context.
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+```
+
+The current solution has following limitations:
+* Users cannot provide values via file path or environment variable.
+* Defining private keys inline is insecure, especially since these values are synced to the global control plane.
+
+
+There is another issue when user wants to explicitly use the `DataSource` in the resource. It's not clear if value provided
+by the data source should be accessed by the data plane or control plane. In some resources it makes sense to have it on the control plane
+while for others on the data plane.
```
bartsmykla: Nit:
```suggestion
There is another issue when user wants to explicitly use the `DataSource` in the resource. It's not clear if value provided by the data source should be accessed by the proxy or control plane. In some resources it makes sense to have it on the control plane while for others on the proxy.
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+
+It should be feasible to define a `MeshExternalService` with all required mTLS configuration in a single location, allowing it to be synced and consumed across multiple zones without additional configuration.
+
+### As a Mesh Operator, I want to prevent users from seeing private keys in the GUI
+
+`Inline` DataSource values can expose sensitive information like private keys in the user interface. For security, `inline` definitions should be discouraged or clearly marked as insecure for use with private keys.
+
+### As a Mesh Operator, I want to configure mTLS for MeshExternalService using a Secret
+
+To maintain a strong security boundary, the control plane should only access `Secrets` from the system namespace (e.g., kuma-system). When a `MeshExternalService` is configured with a `Secret`, it makes sense for the control plane to read that `Secret` and securely deliver it to the data plane via SDS.
```
bartsmykla: Nit:
```suggestion
To maintain a strong security boundary, the control plane should only access `Secrets` from the system namespace (e.g., kuma-system). When a `MeshExternalService` is configured with a `Secret`, it makes sense for the control plane to read that `Secret` and securely deliver it to the proxy via SDS.
```

docs/madr/decisions/081-datasource-api.md
```diff
@@ -0,0 +1,485 @@
...
+* Resolution logic is embedded in the object itself, making it clear what is supported and expected for each resource.
+
+#### Cons
+* Two separate models, which may introduce some code duplication and increase maintenance overhead.
+
+## Security implications and review
+
+Following changes increases security:
+* Inline secrets are explicitly marked as `insecure`, helping to discourage their use in production environments.
+* In Option 2, the private key is not transmitted over the network to the data plane.
```
bartsmykla: Nit:
```suggestion
* In Option 2, the private key is not transmitted over the network to the proxy.
```

bartsmykla (approved):
Few nit picks, but LGTM

slonka (approved):

PR #13924, opened 2025-07-11
feat(kuma-dp): use unix socket for readiness reporter

Icarus9913:
## Supporting documentation

Fix #XX

### DIFF
--- a/app/kuma-dp/cmd/run.go
+++ b/app/kuma-dp/cmd/run.go
@@ -306,12 +306,8 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
 			components = append(components, observabilityComponents...)
 
 			var readinessReporter *readiness.Reporter
-			if cfg.Dataplane.ReadinessPort > 0 {
-				readinessReporter = readiness.NewReporter(
-					bootstrap.GetAdmin().GetAddress().GetSocketAddress().GetAddress(),
-					cfg.Dataplane.ReadinessPort)
-				components = append(components, readinessReporter)
-			}
+			readinessReporter = readiness.NewReporter(cfg.DataplaneRuntime.SocketDir)
+			components = append(components, readinessReporter)
 
 			if err := rootCtx.ComponentManager.Add(components...); err != nil {
 				return err
--- a/app/kuma-dp/pkg/dataplane/readiness/component.go
+++ b/app/kuma-dp/pkg/dataplane/readiness/component.go
@@ -8,11 +8,11 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/asaskevich/govalidator"
 	"github.com/bakito/go-log-logr-adapter/adapter"
 
 	"github.com/kumahq/kuma/pkg/core"
 	"github.com/kumahq/kuma/pkg/core/runtime/component"
+	core_xds "github.com/kumahq/kuma/pkg/core/xds"
 )
 
 const (
@@ -23,28 +23,21 @@ const (
 
 // Reporter reports the health status of this Kuma Dataplane Proxy
 type Reporter struct {
-	localListenAddr string
-	localListenPort uint32
-	isTerminating   atomic.Bool
+	socketDir     string
+	isTerminating atomic.Bool
 }
 
 var logger = core.Log.WithName("readiness")
 
-func NewReporter(localIPAddr string, localListenPort uint32) *Reporter {
+func NewReporter(socketDir string) *Reporter {
 	return &Reporter{
-		localListenPort: localListenPort,
-		localListenAddr: localIPAddr,
+		socketDir: socketDir,
 	}
 }
 
 func (r *Reporter) Start(stop <-chan struct{}) error {
-	protocol := "tcp"
-	addr := r.localListenAddr
-	if govalidator.IsIPv6(addr) {
-		protocol = "tcp6"
-		addr = fmt.Sprintf("[%s]", addr)
-	}
-	lis, err := net.Listen(protocol, fmt.Sprintf("%s:%d", addr, r.localListenPort))
+	socketPath := core_xds.ReadinessReporterSocketName(r.socketDir)
+	lis, err := net.Listen("unix", socketPath)
 	if err != nil {
 		return err
 	}
--- a/pkg/config/app/kuma-dp/config.go
+++ b/pkg/config/app/kuma-dp/config.go
@@ -30,7 +30,6 @@ var DefaultConfig = func() Config {
 			Name:                          "", // Dataplane name must be set explicitly
 			DrainTime:                     config_types.Duration{Duration: 30 * time.Second},
 			ProxyType:                     "dataplane",
-			ReadinessPort:                 9902,
 			ResilientComponentMaxBackoff:  config_types.Duration{Duration: 1 * time.Minute},
 			ResilientComponentBaseBackoff: config_types.Duration{Duration: 5 * time.Second},
 		},
@@ -140,8 +139,6 @@ type Dataplane struct {
 	ProxyType string `json:"proxyType,omitempty" envconfig:"kuma_dataplane_proxy_type"`
 	// Drain time for listeners.
 	DrainTime config_types.Duration `json:"drainTime,omitempty" envconfig:"kuma_dataplane_drain_time"`
-	// Port that exposes kuma-dp readiness status on localhost, set this value to 0 to provide readiness by "/ready" endpoint from Envoy adminAPI
-	ReadinessPort uint32 `json:"readinessPort,omitempty" envconfig:"kuma_readiness_port"`
 	// ResilientComponentBaseBackoff defines the base backoff between restarts of resilient components
 	ResilientComponentBaseBackoff config_types.Duration `json:"resilientComponentBaseBackoff,omitempty" envconfig:"kuma_dataplane_resilient_component_base_backoff"`
 	// ResilientComponentMaxBackoff defines the max backoff between restarts of resilient components
@@ -331,10 +328,6 @@ func (d *Dataplane) Validate() error {
 	// Notice that d.AdminPort is always valid by design of PortRange
 	if d.DrainTime.Duration <= 0 {
 		errs = multierr.Append(errs, errors.Errorf(".DrainTime must be positive"))
-	}
-
-	if d.ReadinessPort > 65353 {
-		return errors.New(".ReadinessPort has to be in [0, 65353] range")
 	}
 
 	return errs
--- a/pkg/core/xds/sockets.go
+++ b/pkg/core/xds/sockets.go
@@ -24,6 +24,11 @@ func OpenTelemetrySocketName(workdir string, backendName string) string {
 	return socketName(filepath.Join(workdir, "kuma-otel-"+backendName))
 }
 
+// ReadinessReporterSocketName generates a socket path that will fit the Unix socket path limitation of 104 chars
+func ReadinessReporterSocketName(wordkir string) string {
+	return socketName(filepath.Join(wordkir, "kuma-readiness-reporter"))
+}
+
 func socketName(s string) string {
 	trimLen := len(s)
 	if trimLen > 98 {
--- a/pkg/xds/generator/admin_proxy_generator.go
+++ b/pkg/xds/generator/admin_proxy_generator.go
@@ -10,6 +10,7 @@ import (
 
 	core_xds "github.com/kumahq/kuma/pkg/core/xds"
 	util_maps "github.com/kumahq/kuma/pkg/util/maps"
+	"github.com/kumahq/kuma/pkg/util/pointer"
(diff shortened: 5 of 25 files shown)


Icarus9913:
> 1. In the new version, the CP and DPPS configure the kuma-sidecar to listen on a Unix socket by default, and the Envoy XDS redirects the `/ready` route to this Unix socket. Note that this behavior cannot be disabled.
> 2. For compatibility with the older versions of DPPs:
> * If the DPP uses a port range(0,65353], meaning the kuma-sidecar listens on a TCP port, the Envoy XDS will retain the old behavior and redirect the `/ready` route to the TCP listener.
> * If the DPP uses port 0, which means the kuma-sidecar doesn't listen on a TCP Port, the envoy XDS will redirect the `/ready` route to the Envoy admin API.

I did a test and found that once if I put the following ENV for CP:
```text
       env:
        - name: KUMA_RUNTIME_KUBERNETES_INJECTOR_SIDECAR_CONTAINER_ENV_VARS
          value: KUMA_READINESS_PORT:0
```

then the DP bootstrap metadata didn't have 
```text
"dataplane.readinessReporter.port": "0"
```

In this situation, we can't distinguish if the old DPPs disabled the readiness reporter. We need to add extra flag for the Unix socket switching :(

pkg/config/app/kuma-dp/config.go
```diff
@@ -30,6 +30,7 @@ var DefaultConfig = func() Config {
 			Name:                          "", // Dataplane name must be set explicitly
 			DrainTime:                     config_types.Duration{Duration: 30 * time.Second},
 			ProxyType:                     "dataplane",
+			ReadinessUnixSocketEnabled:    false,
 			ReadinessPort:                 9902,
```
Icarus9913: In this implementation, the `ReadinessUnixSocketEnabled` is `false` by default, and the user needs to enable it explicitly.

jijiechen: I think we can just make it enabled by default as this will be a break change anyway for those users who wants to stick with TCP ports. Making this enabled by default can let most users experience the new feature sooner.

jijiechen:
LGTM, let's remove the port as Windows is already not supported in 2.11.

jijiechen (approved):

pkg/config/app/kuma-dp/config.go
Icarus9913: Updated `ReadinessUnixSocketEnabled ` to `true` by default

pkg/xds/bootstrap/types/bootstrap_request.go
```diff
@@ -14,18 +14,19 @@ type BootstrapRequest struct {
...
-	Features             []string                   `json:"features"`
-	Resources            ProxyResources             `json:"resources"`
-	Workdir              string                     `json:"workdir"`
-	MetricsResources     MetricsResources           `json:"metricsResources"`
-	SystemCaPath         string                     `json:"systemCaPath"`
-	TransparentProxy     *tproxy_dp.DataplaneConfig `json:"dataplaneConfig,omitempty"`
+	CaCert                     string                     `json:"caCert"`
+	DynamicMetadata            map[string]string          `json:"dynamicMetadata"`
+	DNSPort                    uint32                     `json:"dnsPort,omitempty"`
+	ReadinessUnixSocketEnabled bool                       `json:"readinessUnixSocketEnabled,omitempty"`
```
lahabana: We use features for this these days.

app/kuma-dp/cmd/run.go
```diff
@@ -307,13 +307,12 @@ func newRunCmd(opts kuma_cmd.RunCmdOpts, rootCtx *RootContext) *cobra.Command {
...
 
-			var readinessReporter *readiness.Reporter
-			if cfg.Dataplane.ReadinessPort > 0 {
-				readinessReporter = readiness.NewReporter(
-					bootstrap.GetAdmin().GetAddress().GetSocketAddress().GetAddress(),
-					cfg.Dataplane.ReadinessPort)
-				components = append(components, readinessReporter)
-			}
+			readinessReporter := readiness.NewReporter(
+				cfg.Dataplane.ReadinessUnixSocketEnabled,
```
lahabana: This should be `ReadinessUnixSocketDisabled`.

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -23,28 +24,45 @@ const (
...
+		addr = socketPath
+
+	case r.localListenPort == 0:
+		return nil
+
+	case r.localListenPort != 0:
+		if govalidator.IsIPv6(addr) {
+			protocol = "tcp6"
+			addr = fmt.Sprintf("[%s]:%d", addr, r.localListenPort)
+		}
```
lahabana, line 54: This is tricky to read is this simpler:

```go
var lis net.Listener
if r.unixSocketDisabled {
   if r.localListenPort == 0 {
      return nil
   }
  addr := fmt.Sprintf("%s:%d", r.localListenAddr, r.localListenPort)
  proto := tcp
  if govalidator.IsIPv6(addr) {
          proto = "tcp6"
          addr = fmt.Sprintf("[%s]:%d, addr, r.localListenPort)
   }
   if l, err := net.Listen(proto, addr); err != nil {
     return err
   }
   lis = l
} else {
   if l, err := net.Listen("unix", core_xds.ReadinessReporterSocketName(r.socketDir)); err != nil {
     return err
   }
   lis = l
}
```

The advantage here is that once we go full unix socket it's much easier to remove the code (you remove 1 if block)

pkg/core/xds/types/features.go
```diff
@@ -33,3 +33,5 @@ const FeatureBindOutbounds string = "feature-bind-outbounds"
 // proxy passthrough, and system format for internal Kuma resources that users typically
 // don't need to care about unless debugging Kuma.
 const FeatureUnifiedResourceNaming string = "feature-unified-resource-naming"
+
+const FeatureReadinessUnixSocket = "feature-readiness-unix-socket"
```
lahabana, line 39: Comment please. What is this feature for?

pkg/xds/generator/admin_proxy_generator.go
```diff
@@ -98,13 +98,16 @@ func (g AdminProxyGenerator) Generate(ctx context.Context, _ *core_xds.ResourceS
...
+		case unixSocketEnabled, tcpReadinessReporterEnabled:
 			se.ClusterName = dppReadinessClusterName
-		} else {
-			// we keep the previous behavior if readinessPort is not set
-			// this can happen when an existing DPP is connecting to this CP, it does not have this metadata
+		// if readinessPort is set to 0, we use envoy admin cluster for readiness check
+		case tcpReadinessReporterDisabled:
 			se.ClusterName = envoyAdminClusterName
 		}
 	}
```
lahabana: Why is this a function?

Should it be:

```go
readinessClusterName := dppReadinessClusterName
if readinessPort == 0 && !proxy.Metadata.HasFeature(types.FeatureReadinessUnixSocket) {
   readinessClusterName = envoyAdminClusterName // Fallback to envoyAdmin for health.
}
```

Then you actually just do `se.clusterName = readinessClusterName` later in the function no?

pkg/xds/generator/admin_proxy_generator.go
```diff
@@ -161,25 +164,35 @@ func (g AdminProxyGenerator) Generate(ctx context.Context, _ *core_xds.ResourceS
...
+			UnixDomainPath: core_xds.ReadinessReporterSocketName(proxy.Metadata.WorkDir),
+		}
+	case tcpReadinessReporterEnabled:
+		xdsEndpoint = core_xds.Endpoint{
+			Target: adminAddress,
+			Port:   readinessPort,
 		}
+	case tcpReadinessReporterDisabled:
+		return resources, nil
+	}
```
lahabana, line 166: This is usually easier to read with a if with an early return and an if/else later.

pkg/xds/generator/admin_proxy_generator.go
lahabana: It's also strange that the 2 mutation in for loops here after are outside the if block where the collections are used.

pkg/xds/generator/admin_proxy_generator_test.go
```diff
@@ -130,6 +130,15 @@ var _ = Describe("AdminProxyGenerator", func() {
 				xds_types.FeatureUnifiedResourceNaming: true,
 			},
 		}),
+		Entry("should generate admin resources, Unix socket enabled", testCase{
+			dataplaneFile: "09.dataplane.input.yaml",
+			expected:      "09.envoy-config.golden.yaml",
+			adminAddress:  "127.0.0.1",
+			readinessPort: 9902,
+			features: map[string]bool{
+				xds_types.FeatureReadinessUnixSocket: true,
```
lahabana: worth adding `xds_types.FeatureUnifiedResourceNaming` too

pkg/config/app/kuma-dp/config.go
```diff
@@ -140,6 +140,8 @@ type Dataplane struct {
 	ProxyType string `json:"proxyType,omitempty" envconfig:"kuma_dataplane_proxy_type"`
 	// Drain time for listeners.
 	DrainTime config_types.Duration `json:"drainTime,omitempty" envconfig:"kuma_dataplane_drain_time"`
+	// ReadinessUnixSocketDisabled disables readiness check via Unix socket.
+	ReadinessUnixSocketDisabled bool `json:"readinessUnixSocketDisabled,omitempty" envconfig:"kuma_readiness_unix_socket_disabled"`
 	// Port that exposes kuma-dp readiness status on localhost, set this value to 0 to provide readiness by "/ready" endpoint from Envoy adminAPI
 	ReadinessPort uint32 `json:"readinessPort,omitempty" envconfig:"kuma_readiness_port"`
```
lahabana, line 147: I'd think we should add a link to the issue to indicate this should be removed.

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -23,28 +24,45 @@ const (
...
-		addr = fmt.Sprintf("[%s]", addr)
+	addr := fmt.Sprintf("%s:%d", r.localListenAddr, r.localListenPort)
+
+	switch {
+	case !r.unixSocketDisabled:
+		protocol = "unix"
+		socketPath := core_xds.ReadinessReporterSocketName(r.socketDir)
+		addr = socketPath
+
+	case r.localListenPort == 0:
```
lahabana: Think about this some more. The ability to disable this was added in https://github.com/kumahq/kuma/pull/11107 as a possibility to get out of a potential bug on our side.

So it's been around for quite a while and I don't remember ever hearing about anyone complaining about this.
Would the whole code become simpler if `r.localListenPort` could never be 0 so readiness always go through the go process.
In which case there'd be only 2 choices each time: tcp socket and unixDomain.

UPGRADE.md
```diff
@@ -12,6 +12,12 @@ does not have any particular instructions.
 
 These endpoints were deprecated, and are now removed. You can achieve the same functionality with `/zones/_overview`.
 
+### Deprecation of readiness reporter TCP port in favor of Unix socket
+
+The readiness reporter TCP port is deprecated and will be removed in a future release, also TCP port 0 is now not allowed to be used.
```
lahabana: It is also not longer possible to disable readiness reporter

pkg/core/xds/sockets.go
```diff
@@ -24,6 +24,11 @@ func OpenTelemetrySocketName(workdir string, backendName string) string {
 	return socketName(filepath.Join(workdir, "kuma-otel-"+backendName))
 }
 
+// ReadinessReporterSocketName generates a socket path that will fit the Unix socket path limitation of 104 chars
+func ReadinessReporterSocketName(wordkir string) string {
```
lahabana: Typo workdir

pkg/xds/generator/admin_proxy_generator_test.go
```diff
@@ -84,92 +84,114 @@ var _ = Describe("AdminProxyGenerator", func() {
...
+			expected:      "06.envoy-config.golden.yaml",
 			adminAddress:  "::1",
 			readinessPort: 9400,
 		}),
-		Entry("should generate admin resources, unified naming", testCase{
+		Entry("should generate admin resources, unified naming, readiness with TCP port 9902", testCase{
+			dataplaneFile: "07.dataplane.input.yaml",
+			expected:      "07.envoy-config.golden.yaml",
+			adminAddress:  "",
+			readinessPort: 9902,
```
lahabana, line 127: Should these be removed?

Icarus9913: I think we can keep it currently as the recent 2 minor versions, the 9902 TCP Port is still supported for compatibility. With this, we can have the golden files that show the readiness clusters use IP addr + Port.

I created the cleanup issue to track this: https://github.com/kumahq/kuma/issues/14039

PR #13928, opened 2025-07-11
fix(xds): check resource type when resolving identifier

lukidzi:
## Motivation

https://github.com/kumahq/kuma/actions/runs/16160736111/job/45614269770

I was investigating a flake in the test and noticed that the generated route was incorrect. Based on the config dump, I observed that instead of a `MeshService`, we had a `MeshMultiZoneService`.

After a discussion with @lobkovilya and further debugging, I confirmed that this was caused by the lack of a type check. We were selecting the oldest resource without verifying its type.

## Implementation information

Ensure the resource type is checked before choosing the oldest object.


bartsmykla (approved):
lgtm

pkg/xds/context/context.go
```diff
@@ -102,7 +102,7 @@ func (mc *MeshContext) ResolveResourceIdentifier(resType core_model.ResourceType
 	var oldestTri *kri.Identifier
 	for _, tri := range mc.BaseMeshContext.DestinationIndex.resolveResourceIdentifiersForLabels(labels) {
```
lobkovilya: I think `resolveResourceIdentifiersForLabels` should accept `resType` as the first arg. I don't think it's ever correct to use this method for multiple types at once

lobkovilya (approved):

Issue #13903, opened 2025-07-09
[system resources] Implement unified name generation for system Envoy resources


slonka:
### Description

Mirror of: https://github.com/kumahq/kuma/issues/13798

Requires: https://github.com/kumahq/kuma/issues/13795

### Description

Add support for generating Envoy resource and stat names using a consistent naming format when the `feature-unified-resource-naming` flag is enabled. This includes:

* The `system_<kri>` structure defined in [MADR-076](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/076-naming-system-envoy-resources.md?plain=1#L110)
* The `system_<descriptor>`  defined in [MADR-076](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/076-naming-system-envoy-resources.md?plain=1#L124)

These formats ensure consistent naming across all system resources and stats.

### Acceptance criteria

* [x] Use `system_<kri>` structure for resources that can be traced back to kuma resource
* [x] Use `system_<descriptor>` structure for resources that can not be traced back to kuma resource
* [x] Ensure legacy name generation remains unchanged when the feature flag is not enabled
* [x] Remember about Kong Mesh resources (MeshOPA, MeshGlobalRateLimit)

PR #13942, opened 2025-07-14
feat(kuma-cp): rename envoy admin system resource to follow unified naming format

slonka:
## Motivation

Best described in https://github.com/kumahq/kuma/issues/13903 and related MADRs

## Implementation information

Add a switch on top of a FF to rename envoy admin resources.

## Supporting documentation

xrel https://github.com/kumahq/kuma/issues/13903
closes https://github.com/kumahq/kuma/issues/13966

### DIFF
--- a/pkg/hds/tracker/callbacks.go
+++ b/pkg/hds/tracker/callbacks.go
@@ -175,7 +175,7 @@ func (t *tracker) OnEndpointHealthResponse(streamID xds.StreamID, resp *envoy_se
 		status := clusterHealth.LocalityEndpointsHealth[0].EndpointsHealth[0].HealthStatus
 		health := status == envoy_core.HealthStatus_HEALTHY || status == envoy_core.HealthStatus_UNKNOWN
 
-		if clusterHealth.ClusterName == names.GetEnvoyAdminClusterName() {
+		if clusterHealth.ClusterName == names.GetEnvoyAdminClusterName() || clusterHealth.ClusterName == names.SystemGetAdminResourceName() {
 			envoyHealth = health
 		} else {
 			port, err := names.GetPortForLocalClusterName(clusterHealth.ClusterName)
--- a/pkg/hds/tracker/callbacks.go
+++ b/pkg/hds/tracker/callbacks.go
@@ -25,7 +25,7 @@ import (
 	"github.com/kumahq/kuma/pkg/util/watchdog"
 	util_xds_v3 "github.com/kumahq/kuma/pkg/util/xds/v3"
 	"github.com/kumahq/kuma/pkg/xds/envoy/names"
-	"github.com/kumahq/kuma/pkg/xds/generator"
+	"github.com/kumahq/kuma/pkg/xds/generator/system_names"
 )
 
 type streams struct {
@@ -176,7 +176,8 @@ func (t *tracker) OnEndpointHealthResponse(streamID xds.StreamID, resp *envoy_se
 		status := clusterHealth.LocalityEndpointsHealth[0].EndpointsHealth[0].HealthStatus
 		health := status == envoy_core.HealthStatus_HEALTHY || status == envoy_core.HealthStatus_UNKNOWN
 
-		if clusterHealth.ClusterName == names.GetEnvoyAdminClusterName() || clusterHealth.ClusterName == generator.AdminResourceName {
+		// TODO(unified-resource-naming): adjust when legacy naming is removed
+		if clusterHealth.ClusterName == names.GetEnvoyAdminClusterName() || clusterHealth.ClusterName == system_names.EnvoyAdminResourceName {
 			envoyHealth = health
 		} else {
 			port, err := names.GetPortForLocalClusterName(clusterHealth.ClusterName)
--- a/pkg/hds/tracker/callbacks.go
+++ b/pkg/hds/tracker/callbacks.go
@@ -25,6 +25,7 @@ import (
 	"github.com/kumahq/kuma/pkg/util/watchdog"
 	util_xds_v3 "github.com/kumahq/kuma/pkg/util/xds/v3"
 	"github.com/kumahq/kuma/pkg/xds/envoy/names"
+	"github.com/kumahq/kuma/pkg/xds/generator"
 )
 
 type streams struct {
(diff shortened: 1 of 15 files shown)


pkg/xds/generator/admin_proxy_generator.go
```diff
@@ -117,19 +122,23 @@ func (g AdminProxyGenerator) Generate(ctx context.Context, _ *core_xds.ResourceS
 	resources := core_xds.NewResourceSet()
 	// We bind admin to 127.0.0.1 by default, creating another listener with same address and port will result in error.
 	if g.getAddress(proxy) != adminAddress {
+		adminListenerName := envoy_names.GetAdminListenerName()
```
bartsmykla: Nit: for variable containing cluster name you are using `envoyAdmin` prefix and for listener `admin`. Maybe it would be worth to use the same format (probably `envoyAdmin`). Wdyt?

pkg/hds/tracker/callbacks.go
```diff
@@ -175,7 +175,7 @@ func (t *tracker) OnEndpointHealthResponse(streamID xds.StreamID, resp *envoy_se
 		status := clusterHealth.LocalityEndpointsHealth[0].EndpointsHealth[0].HealthStatus
 		health := status == envoy_core.HealthStatus_HEALTHY || status == envoy_core.HealthStatus_UNKNOWN
 
-		if clusterHealth.ClusterName == names.GetEnvoyAdminClusterName() {
+		if clusterHealth.ClusterName == names.GetEnvoyAdminClusterName() || clusterHealth.ClusterName == names.SystemGetAdminResourceName() {
```
bartsmykla: Nit: maybe worth adding a `TODO(unified-resource-naming): adjust when legacy naming is removed` comment above, or something similar, to make it clear this will need cleanup later. That way we can grep for these TODOs and easily track what to update. Wdyt?

pkg/hds/tracker/healthcheck_generator.go
```diff
@@ -56,8 +57,15 @@ func (g *SnapshotGenerator) GenerateSnapshot(ctx context.Context, node *envoy_co
 		return nil, err
 	}
 
+	md := xds.DataplaneMetadataFromXdsMetadata(node.Metadata)
+	unifiedNamingEnabled := md.HasFeature(xds_types.FeatureUnifiedResourceNaming)
+	clusterName := names.GetEnvoyAdminClusterName()
```
bartsmykla, line 64: Nit: maybe worth adding a `TODO(unified-resource-naming): adjust when legacy naming is removed` comment above, or something similar, to make it clear this will need cleanup later. That way we can grep for these TODOs and easily track what to update. Wdyt?

pkg/xds/generator/admin_proxy_generator.go
```diff
@@ -58,13 +59,17 @@ func (g AdminProxyGenerator) Generate(ctx context.Context, _ *core_xds.ResourceS
...
 	adminPort := proxy.Metadata.GetAdminPort()
 	readinessPort := proxy.Metadata.GetReadinessPort()
+	unifiedNamingEnabled := proxy.Metadata.HasFeature(types.FeatureUnifiedResourceNaming)
 	// We assume that Admin API must be available on a loopback interface (while users
 	// can override the default value `127.0.0.1` in the Bootstrap Server section of `kuma-cp` config,
 	// the only reasonable alternatives are `::1`, `0.0.0.0` or `::`).
 	// In contrast to `AdminPort`, we shouldn't trust `AdminAddress` from the Envoy node metadata
 	// since it would allow a malicious user to manipulate that value and use Prometheus endpoint
 	// as a gateway to another host.
 	envoyAdminClusterName := envoy_names.GetEnvoyAdminClusterName()
```
bartsmykla, line 71: Nit: maybe worth adding a `TODO(unified-resource-naming): adjust when legacy naming is removed` comment above, or something similar, to make it clear this will need cleanup later. That way we can grep for these TODOs and easily track what to update. Wdyt?

pkg/xds/generator/admin_proxy_generator.go
bartsmykla: Nit: maybe worth adding a `TODO(unified-resource-naming): adjust when legacy naming is removed` comment above, or something similar, to make it clear this will need cleanup later. That way we can grep for these TODOs and easily track what to update. Wdyt?

pkg/xds/envoy/names/system_names.go
```diff
@@ -0,0 +1,9 @@
+package names
+
+import "fmt"
+
+const SystemPrefix = "system_"
+
+func SystemGetAdminResourceName() string {
+	return fmt.Sprintf("%senvoy_admin", SystemPrefix)
+}
```
bartsmykla: Nit: Wdyt about changing the name to `SystemGetEnvoyAdminResourceName`

bartsmykla (dismissed):
Few nits, but LGTM

pkg/xds/envoy/names/system_names.go
lahabana: I'm curious about this package because it has very general things (it's a bit of a `util`) that likely may make more sense to be defined close to where they are used in a non public manner.

For example: https://github.com/kumahq/kuma/blob/2f0a82a05de4e87e505f886bdffbdee509938c19/pkg/xds/dynconf/listener.go#L23

The reason for this is that it raises the bar on doing cross dependencies of things.
By this I mean: "If you want to have something depend on something else, you should think about it because dependencies have a cost".

So what I'd do here is still have a `pkg/xds/envoy/system` (likely more `pkg/core/system_names` so it's next to `kri`) and then in package define:

- IsSystem(name string) bool
- AsSystemName(name string) string // Adds the prefix

Needing to know exactly that this is the admin listener (and that folks can reference the admin listener) should be part of where the listener pacakge is (with a dpapi package for .e.g like we did for plugins: https://github.com/kumahq/kuma/blob/master/pkg/plugins/policies/meshmetric/dpapi/api.go)

This is just a suggestion btw :)

lahabana: Actually you can already see this in your code. Seems like healthcheck generator etc relying on this listener is an important dependency of the admin listener not of the name package

slonka: I'm on the fence about this one. On one hand this:

> The reason for this is that it raises the bar on doing cross dependencies of things.

makes sense but on the other hand it goes against:
- smaller dependencies are generally better than bigger ones (better to depend on a package with 0 dependencies than 10)
- depending in direction of stability
- this can lead to circular dependencies (but I guess if we already have a circular dependency that is not visible now because we only reference a name then I guess it's better to remove this loop)
- seem like other meshes opted for this as well ([consul](https://github.com/hashicorp/consul/blob/863d6aeedde94d785e473cefbcd91443557bd09d/envoyextensions/xdscommon/xdscommon.go#L18))

I think I'll ask the team to chip in.

lahabana: For smaller dep the way to go is to have a package with just the api (see example of dpapi) this solves most of your concerns I believe.

pkg/hds/tracker/healthcheck_generator.go
slonka: added a couple of lines above

pkg/xds/generator/system_names/api.go
```diff
@@ -0,0 +1,5 @@
+package system_names
```
slonka, line 1: maybe the whole `admin_generator` should be moved to a separate package, so it's not in general "generator" package? and then move this to that package? WDYT?

bartsmykla: Nit: I like this idea. Let's use the prefix `SystemResourceName` (in this case `SystemResourceNameEnvoyAdmin`) for these type of variables so it's easily grepable when searching for all related constants. Wdyt?

PR #13949, opened 2025-07-14
test(transparentproxy): remove deprecated debian 10 and add 13

lukidzi:
## Motivation

Debian 10 is deprecated since the last year and we cannot find iptables packages now https://packages.debian.org/buster/allpackages

## Implementation information

Remove debian 10 and add 13

## Supporting documentation

https://www.debian.org/releases/trixie/release-notes/
https://www.debian.org/releases/buster/


bartsmykla (changes requested):
It seems you forgot to commit the updated [`test/transparentproxy/config.go`](https://github.com/kumahq/kuma/blob/ee83aaa6fd5f45eaf0b2950391ad32f5dcb548e1/test/transparentproxy/config.go) as there are only golden files

bartsmykla (approved):
It seems fine now

PR #13952, opened 2025-07-15
test(meshtls): improve meshtls test by running requets between asserts

slonka:
## Motivation

@lukidzi noticed a flake:  https://github.com/kumahq/kuma/actions/runs/16269540267/job/45939071797

```
• [FAILED] [40.527 seconds]
MeshTLS [It] should set cypher and version
github.com/kumahq/kuma/test/e2e_env/universal/meshtls/policy.go:332

Captured StdOut/StdErr Output

  [FAILED] Timed out after 30.001s.
  The function passed to Eventually failed at github.com/kumahq/kuma/test/e2e_env/universal/meshtls/policy.go:411 with:
  no stat found: &{[]}
Error: It 07/14/25 15:42:24.33
```

## Implementation information

Which we think was caused by the policy not being propagated quickly enough so that the requests done in the previous lines did not trigger the stats change. We moved everything after policy application to be in eventually which should re-trigger stats if there is a failure in eventually.

### DIFF
--- a/test/e2e_env/universal/meshtls/policy.go
+++ b/test/e2e_env/universal/meshtls/policy.go
@@ -125,36 +125,30 @@ spec:
 			)
 			g.Expect(err).ToNot(HaveOccurred())
 			g.Expect(responses.Instance).To(Equal("test-server"))
-		}, "30s", "500ms").Should(Succeed())
 
-		// and
-		// can access test-server-2 from service in the mesh
-		Eventually(func(g Gomega) {
-			responses, err := client.CollectEchoResponse(
+			// and
+			// can access test-server-2 from service in the mesh
+			responses, err = client.CollectEchoResponse(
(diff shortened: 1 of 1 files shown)


bartsmykla (approved):

test/e2e_env/universal/meshtls/policy.go
```diff
@@ -125,36 +125,30 @@ spec:
...
 			g.Expect(responses.Instance).To(Equal("test-server"))
-		}, "30s", "500ms").Should(Succeed())
 
-		// and
-		// can access test-server-2 from service in the mesh
-		Eventually(func(g Gomega) {
-			responses, err := client.CollectEchoResponse(
+			// and
+			// can access test-server-2 from service in the mesh
+			responses, err = client.CollectEchoResponse(
```
slonka, line 131: hold on with merging until patch release done

bartsmykla: it's done

Issue #13319, opened 2025-04-04
MeshRateLimit validate `local.http.onRateLimit.status` to be >= 400


lobkovilya:
### Description

Today, it's possible to apply:

```yaml
type: MeshRateLimit
mesh: default
name: mrl
spec:
  from:
    - default:
        local:
          http:
            requestRate:
              num: 10
              interval: 1m0s
            onRateLimit:
              status: 204
```

But Envoy is going to [default the status to 429](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto#envoy-v3-api-field-extensions-filters-http-local-ratelimit-v3-localratelimit-status):

> If this is set to < 400, 429 will be used instead. 

We need to improve validation and disallow `status` to be less than 400.

slonka:
> triage: done https://github.com/kumahq/kuma/blob/fa7b39e3e912364f5323df9b5dab40aeb4c86a7a/pkg/plugins/policies/meshratelimit/api/v1alpha1/deprecated.go#L42

PR #13958, opened 2025-07-15
fix(meshratelimit): add warning log about status code

lukidzi:
## Motivation

If the status is lower than 400 envoy fallback to 429. We shouldn't allow setting incorrect status

## Implementation information

Added log information that setting status lower than 400 is not supported

## Supporting documentation

part of: #https://github.com/kumahq/kuma/issues/13319

### DIFF
--- a/pkg/plugins/policies/meshratelimit/api/v1alpha1/deprecated.go
+++ b/pkg/plugins/policies/meshratelimit/api/v1alpha1/deprecated.go
@@ -1,6 +1,8 @@
 package v1alpha1
 
 import (
+	"fmt"
+
 	"github.com/kumahq/kuma/pkg/plugins/policies/core/jsonpatch/validators"
 	"github.com/kumahq/kuma/pkg/util/pointer"
 )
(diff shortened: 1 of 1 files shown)


pkg/plugins/policies/meshratelimit/api/v1alpha1/deprecated.go
```diff
@@ -9,6 +11,35 @@ func (t *MeshRateLimitResource) Deprecations() []string {
 	var deprecations []string
 	if len(pointer.Deref(t.Spec.From)) > 0 {
 		deprecations = append(deprecations, "'from' field is deprecated, use 'rules' instead")
+		for i, rule := range pointer.Deref(t.Spec.From) {
+			if rule.Default.Local != nil && isStatusInvalid(pointer.Deref(rule.Default.Local)) {
```
slonka: you don't need pointer.Deref if you just checked that `rule.Default.Local` is not nil

pkg/plugins/policies/meshratelimit/api/v1alpha1/deprecated.go
```diff
@@ -9,6 +11,35 @@ func (t *MeshRateLimitResource) Deprecations() []string {
...
+				deprecations = append(deprecations, fmt.Sprintf("'spec.from[%d].default.local.http.requestRate.status' must be 400 or higher. Please update your configuration.", i))
```
slonka, line 16: should we mention why that's the case? that envoy uses 429 when that happens?

slonka: should this be in deprecations or in [warnings](https://github.com/kumahq/kuma/blob/62a33aec0311c7484df58f2374e5683f9a6ef93d/pkg/plugins/policies/meshpassthrough/plugin/v1alpha1/plugin.go#L42)?

lukidzi: `warnings` are not presented anywhere so it might not be a solution here

slonka (approved):

Issue #13873, opened 2025-07-07
Create new resource MeshIdentity API


lukidzi:
### Description

Based on MADR we should create a new resource https://github.com/kumahq/kuma/pull/13804

PR #13985, opened 2025-07-17
refactor(api): move selector and status to the common api struct

lukidzi:
## Motivation

We create the same structs in different places. It makes sense to have them in a common struct and use them in differenr resources.

## Implementation information

Extract resource to the common struct

## Supporting documentation

Part of #13873


app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
```diff
@@ -3901,11 +3901,7 @@ spec:
                       matchLabels:
                         additionalProperties:
                           type: string
-                        description: MatchLabels matches multiple MeshServices by
```
lukidzi, line 3904: @slonka Is it ok since we have a validation in the code?

slonka: I think it's fine to loosen validation in oapi schema as long as we have validation in the code. Tagging @lobkovilya to confirm.

app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
```diff
@@ -521,6 +521,8 @@ ingress:
     type: LoadBalancer
     # -- (string) Optionally specify IP to be used by cloud provider when configuring load balancer
     loadBalancerIP:
+    # -- A list of CIDRs or IPs for the cloud-provider load balancer to allow access from.
+    loadBalancerSourceRanges: []
```
slonka: I don't think this is related, was this pulled from another PR? 🤔

lukidzi: I think it should be in master, I don't fully understand why it doesn't fail if we are missing it 😕

lukidzi: Ok merged master and seems ok

slonka (approved):

PR #13989, opened 2025-07-17
ci(lint): enable extra-rules from gofumpt and paramTypeCombine from go-critic

mmorel-35:
## Motivation

Check if code and import statements are formatted, with additional rules.

This is the application of `golangci-lint fmt ./...`

## Supporting documentation

* https://golangci-lint.run/usage/formatters/#gofumpt
* https://go-critic.com/overview.html#paramtypecombine

## Note

I appended `DO NOT EDIT.` to `Code generated ...` so they are not linted by golangci-lint

> Changelog: skip


mmorel-35:
I don't like to try to defend each linter rule so if you don't want to take it after seeing the two docs I provided, I won't try to convince anyone. It's just that it makes thing more readable I guess. If you don't want this, I can revert golang changes and add a WON'T FIX comment in the golangci-lint configuration

PR #13991, opened 2025-07-20
feat(meshtrust): create api definition 

### LINKED ISSUE #13874: Create a new resource MeshTrust API
### Description

Based on MADR we should create a new resource API https://github.com/kumahq/kuma/pull/13804

lukidzi:
## Supporting documentation

Fix #13874


pkg/core/resources/apis/meshtrust/api/v1alpha1/validator.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+
+	"github.com/kumahq/kuma/pkg/core/validators"
+)
+
+func (r *MeshTrustResource) validate() error {
+	var verr validators.ValidationError
+	path := validators.RootedAt("spec")
+	if r.Spec.Origin != nil {
+		verr.AddViolationAt(path.Field("origin"), "cannot be provided by the user")
```
slonka: We have `MustNotBeDefined`

pkg/core/resources/apis/meshtrust/api/v1alpha1/validator.go
```diff
@@ -0,0 +1,57 @@
...
+	return verr.OrNil()
+}
+
+func validateTrustDomain(path validators.PathBuilder, td string) validators.ValidationError {
+	var verr validators.ValidationError
+	if td == "" {
+		verr.AddViolationAt(path.Field("trustDomain"), "trustDomain needs to be defined")
+	}
+	if len(td) > 253 {
+		verr.AddViolationAt(path.Field("trustDomain"), fmt.Sprintf("trustDomain needs to shorter than 253 sings, current has %d", len(td)))
```
slonka: we have `ValidateLength` in common validators

slonka: does it even work if we use templates like 

```
{ .Mesh }.{ .Zone }.mesh.local
```

? aren't we just checking the length of the template?

slonka: if we can do this post templating should we check if this is an actual domain?

pkg/core/resources/apis/meshtrust/api/v1alpha1/validator.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return verr
+}
+
+func validateCABundles(path validators.PathBuilder, bundles []CABundle) validators.ValidationError {
+	var verr validators.ValidationError
+	if len(bundles) == 0 {
+		verr.AddViolationAt(path, "atleast 1 bundle needs to be provided")
```
slonka: I think we have `MustNotBeEmpty` or you can adjust a common validator to be more precise

pkg/core/resources/apis/meshtrust/api/v1alpha1/validator.go
```diff
@@ -0,0 +1,57 @@
...
+		return verr
+	}
+	for i, bundle := range bundles {
+		switch bundle.Type {
+		case PemCABundleType:
```
slonka, line 34: missing default to check for invalid type

pkg/core/resources/apis/meshtrust/api/v1alpha1/validator.go
```diff
@@ -0,0 +1,57 @@
...
+			path := path.Index(i).Field("pem")
+			if bundle.Pem == nil {
+				verr.AddViolationAt(path, "pem needs to be defined")
```
slonka: we have `MustBeDefined`

pkg/core/resources/apis/meshtrust/api/v1alpha1/validator.go
```diff
@@ -0,0 +1,57 @@
...
+			if !isPEMFormat(bundle.Pem.Value) {
+				verr.AddViolationAt(path.Field("value"), "provided certificate has incorrect format")
+			}
+		}
+	}
+	return verr
+}
+
+func isPEMFormat(s string) bool {
+	block, _ := pem.Decode([]byte(s))
```
slonka, line 51: should we check if `block.Type` is not a private key?

pkg/core/resources/apis/meshtrust/api/v1alpha1/meshtrust.go
```diff
@@ -0,0 +1,45 @@
...
+const (
+	PemCABundleType CABundleType = "Pem"
+)
+
+type CABundle struct {
+	// Type specifies the format or source type of the CA bundle.
+	// +required
+	Type CABundleType `json:"type"`
+	// Pem contains the PEM-encoded CA bundle if the Type is set to a PEM-based format.
+	Pem *Pem `json:"pem,omitempty"`
```
slonka: Shouldn't this be named `PEM` ?

pkg/core/resources/apis/meshtrust/api/v1alpha1/validator.go
lukidzi: We don't use it after (unless it's autogenerated from MeshIdentity). There is a case when a user can define a trust domain manually so it's possible to define longer than 253 signs so it's nice to verify

pkg/core/resources/apis/meshtrust/api/v1alpha1/validator.go
lukidzi: I discovered that we can't do it here since we reject autogenerated MeshTrust

PR #14011, opened 2025-07-21
feat(kuma-cp): rename dynamic config system resource to follow unified naming format

slonka:
## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13967


bartsmykla (dismissed):

pkg/core/system_names/system_names.go
```diff
@@ -1,15 +1,37 @@
...
 import (
+	"regexp"
 	"strings"
 )
 
 const SystemPrefix = "system_"
 
+const separator = "_"
+
+var cleanNameRegex = regexp.MustCompile(`[^a-z0-9-_]`)
```
lukidzi: Should we clean the name or just name it correctly and validate that we set it based on the MADR?

bartsmykla (approved):
lgtm

PR #14027, opened 2025-07-22
feat(tools): move to Mise from our custom scripts

Automaat:
## Motivation

Introducing [Mise](https://mise.jdx.dev) to Kuma. We already use Mise in projects like `kuma-website` or `kuma-counter-demo`. 
- With mise we can streamline tools installation for different Kuma versions, which replaces our custom scripts and `make dev/tools` target. 
- While switching branches for example from release to master mise will automatically enable specified versions of tools which removes need of reinstalling everything always and reduces mistakes while working with older branches
- We can now easily update tools version with [Renovate](https://docs.renovatebot.com/modules/manager/mise/) which was now not possible, and we almost never updated them
- Mise action cache works much better than our custom cache with `make dev/tools` which will speed up CI a little bit

### Kong Mesh or other projects using Kuma compatibility
Right now, we don't have any extra dependencies than we have in Kuma but adding them would be easy as you can have multiple config files that mise will merge: https://mise.jdx.dev/configuration.html#configuration

## Addressed concerns
- I've previously moved clang-format to separate action so that we don't need to install it on CI. We can safely disable this with `MISE_DISABLE_TOOLS: "clang-format,golangci-lint,skaffold"`.  Thanks to this installation of tools take 1 minute at most without cache
- we now use GITHUB_TOKEN to avoid rate limit issues we hit on master last time we merged this.

Fix: https://github.com/kumahq/kuma/issues/11770


bartsmykla (dismissed):
I love it 🫶

bartsmykla (dismissed):

PR #14028, opened 2025-07-22
feat(api): create common datasource

lukidzi:
## Motivation

Our current data source doesn't cover all our needs, so we’d like to introduce a new one

## Implementation information

API definition

## Supporting documentation

Fix #https://github.com/kumahq/kuma/issues/13875


api/common/v1alpha1/datasource.go
```diff
@@ -7,6 +7,7 @@ import (
 	system_proto "github.com/kumahq/kuma/api/system/v1alpha1"
 )
 
+// DEPRECATED
```
slonka: I think we usually do "Deprecated: use xyz":

```suggestion
// Deprecated: use api/common/v1alpha1/datasource/datasource.go
```

api/common/v1alpha1/datasource/datasource.go
```diff
@@ -0,0 +1,63 @@
...
+// can be any data, i.e. certs, configs, OPA policies written in rego, lua plugins etc.
+type DataSource struct {
+	Type   DataSourceType `json:"type"`
+	File   *File          `json:"file,omitempty"`
+	EnvVar *EnvVar        `json:"envVar,omitempty"`
+	Inline *Inline        `json:"inline,omitempty"`
+}
+
+type File struct {
+	Path string `json:"path,omitempty"`
```
slonka: When this field is used in a policy it will start getting linted by the api-linter and it won't pass. I think this field is required so it shouldn't have "omitempty" according to our [types doc](https://docs.google.com/document/d/1ZV8KaXr1kV4P5O-eq4gy5IxSMJQ4VzvTT-BmeBABuDo/edit?tab=t.0).

```suggestion
	Path string `json:"path"`
```

api/common/v1alpha1/datasource/datasource.go
```diff
@@ -0,0 +1,63 @@
...
+const (
+	DataSourceFile   DataSourceType = "File"
+	DataSourceEnvVar DataSourceType = "EnvVar"
+	DataSourceInline DataSourceType = "Inline"
+)
+
+// DataSource is just a way to provide data. Not necessarily secrets,
+// can be any data, i.e. certs, configs, OPA policies written in rego, lua plugins etc.
+type DataSource struct {
+	Type   DataSourceType `json:"type"`
```
slonka, line 37: This will probably need:

```suggestion
	// +kuma:discriminator
	Type   DataSourceType `json:"type"`
```

api/common/v1alpha1/datasource/datasource.go
```diff
@@ -0,0 +1,63 @@
...
+const (
+	SecureDataSourceFile      SecureDataSourceType = "File"
+	SecureDataSourceSecretRef SecureDataSourceType = "Secret"
+	SecureDataSourceEnvVar    SecureDataSourceType = "EnvVar"
+	SecureDataSourceInline    SecureDataSourceType = "InsecureInline"
+)
+
+// SecureDataSource is a way to securely provide data to the component
+type SecureDataSource struct {
+	Type           SecureDataSourceType `json:"type,omitempty"`
```
slonka, line 17: Same as below

```suggestion
	// +kuma:discriminator
	Type           SecureDataSourceType `json:"type,omitempty"`
```

api/common/v1alpha1/datasource/datasource.go
```diff
@@ -0,0 +1,63 @@
...
+	EnvVar *EnvVar        `json:"envVar,omitempty"`
+	Inline *Inline        `json:"inline,omitempty"`
+}
+
+type File struct {
+	Path string `json:"path,omitempty"`
+}
+
+type EnvVar struct {
+	Name string `json:"name,omitempty"`
```
slonka:
```suggestion
	Name string `json:"name"`
```

api/common/v1alpha1/datasource/datasource.go
```diff
@@ -0,0 +1,63 @@
...
+// +kubebuilder:validation:Enum=Secret
+type RefType string
+
+const (
+	SecretRefType RefType = "Secret"
+)
+
+type SecretRef struct {
+	Kind RefType `json:"kind,omitempty"`
+	Name string  `json:"name,omitempty"`
```
slonka: These are not optional right?

```suggestion
	Kind RefType `json:"kind"`
	Name string  `json:"name"`
```

api/common/v1alpha1/datasource/datasource.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+type Inline struct {
+	Value string `json:"value,omitempty"`
```
slonka: This one is also not optional, right? 🤔 

```suggestion
	Value string `json:"value"`
```

slonka (commented):
The linter will only run on this when it's used in a policy so just pointing things out that would probably be an error in the linter once it's used

api/common/v1alpha1/datasource.go
lukidzi: I left is as `TODO` since linter fails. We should change it once we move to use a new model in MES

slonka (approved):

Issue #13960, opened 2025-07-16
Add support for unified naming in MeshHTTPRoute and MeshTCPRoute


bartsmykla:
Update the XDS generation logic for `MeshHTTPRoute` and `MeshTCPRoute` to support the new consistent naming format defined in [MADR-070](https://github.com/kumahq/kuma/blob/df678629aa3bdce62a84c0e3752c7e5d45bf1e98/docs/madr/decisions/070-resource-identifier.md), activated by the `feature-unified-resource-naming` flag.

### Acceptance criteria

* Confirm whether default routes use `MeshHTTPRoute`/`MeshTCPRoute` logic
  * If yes, skip applying the new naming to default routes
* Apply the `kri_` format for non-default resources only
* Include `sectionName` as required by MADR-070
* Reuse and update shared logic in [`pkg/plugins/policies/core/xds`](https://github.com/kumahq/kuma/tree/6fddf59333f41c03ec8931856dbcd89e198a0d1b/pkg/plugins/policies/core/xds)
* Exclude built-in gateway resources from the new naming format
* Ensure legacy naming remains unchanged when the feature flag is not set

PR #14038, opened 2025-07-23
feat(kuma-cp/xds): enable unified resource naming for MeshHTTPRoute listeners

bartsmykla:
## Motivation

We want to support a new consistent way of naming Envoy resources, as described in MADR-077. This makes it easier to understand, track, and work with generated resources in tools and dashboards. The new naming is optional and controlled by the `feature-unified-resource-naming` flag.

## Implementation information

We updated how outbound listeners are built for `MeshHTTPRoute` to support the new naming format. The changes include:

- New helper functions to generate names in one place (`destinationname.Resolve`, `MustResolve`)
- A simpler way to get names with fallback values (`MaybeResolveKRIWithFallback`, `ResolveKRIWithFallback`, `MaybeResolveKRI`)
- Updated `GenerateOutboundListener` to use the new naming when the feature flag is on
- Renamed `DestinationService.ServiceName` to `KumaServiceTagValue` to clarify that it comes from the `kuma.io/service` tag
- Renamed `MeshServiceElement` to `KumaServiceTagElement` for clarity
- Replaced `GetAddressOrDefault` with `GetAddressWithFallback` for consistency
- Extracted names like `routeConfigName`, `virtualHostName`, `listenerName`, and `statPrefix` using the new helpers
- Switched to a more flexible listener builder
- Added support for setting the `statPrefix` in listeners
- Cleaned up old logic and updated related tests

These changes let us switch between legacy and unified names without breaking existing behavior.

## Supporting documentation

* [MADR-077](https://github.com/kumahq/kuma/blob/52b8f5548739ecd1124181541f105bd6481b7ba5/docs/madr/decisions/077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md)
* Related issue: https://github.com/kumahq/kuma/issues/13960


pkg/core/resources/apis/core/destinationname/destinationname.go
```diff
@@ -1,13 +1,48 @@
...
+		panic(err)
+	}
+	return name
+}
+
+func Resolve(unifiedNaming bool, dest core.Destination, port core.Port) (string, error) {
+	switch {
+	case dest == nil:
+		return "", errors.New("dest is nil: expected a non-nil dest implementing core.Destination")
+	case unifiedNaming && port != nil && port.GetValue() > 0:
```
Automaat: you should use Port.GetName() instead of checking port name/value: https://github.com/kumahq/kuma/blob/da963defb1b0a387d54b8f98b914885b0cf2a785/pkg/core/resources/apis/core/interfaces.go#L7-L8

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/listeners.go
```diff
@@ -149,27 +170,29 @@ func generateListeners(
 	return resources, nil
```
bartsmykla, line 169: generation logic is the same

pkg/core/resources/apis/core/destinationname/destinationname.go
bartsmykla: addressed

pkg/xds/topology/outbound.go
```diff
@@ -246,27 +247,6 @@ type MeshServiceIdentity struct {
 	Identities map[string]struct{}
 }
 
-func BuildMeshServiceIdentityMap(
```
slonka, line 249: this was never used

PR #14051, opened 2025-07-25
chore(dev/mise): make dev tools updatable by Renovate

bartsmykla:
### Motivation

Some dev tools in `mise.toml` were defined using short names that Renovate does not support. As a result, they were not being updated automatically. To fix that, the full tool names from the [mise registry](https://mise.jdx.dev/registry.html#tools) were used, which allows Renovate to properly detect and manage them. This helps keep our dev dependencies up to date.

### Implementation information

* Replaced tool short names in `mise.toml` with their full names for tools not supported by the Renovate mise manager's default registry
* Added `.renovate/mise.json` preset with custom rules for:
  * Disabling updates for `clang-format` (legacy version in use)
  * Allowing `protoc` updates only in the 3.20.x patch range
  * Mapping full tool names back to short aliases used in our tooling
* Updated `renovate.json` to include the new shared `mise` preset

### Supporting documentation

* Renovate mise manager short name support: https://docs.renovatebot.com/modules/manager/mise/#supported-default-registry-tool-short-names
* Fool tool names sourced from: https://mise.jdx.dev/registry.html#tools

Tested on personal fork with live Renovate runs.


renovate.json
```diff
@@ -27,7 +28,7 @@
         "github-actions"
       ],
       "extends": [
-        "local>kumahq/kuma//.renovate/github-actions"
```
slonka, line 30: why is this 'local' removed?

bartsmykla: it was redundant, `local>` is equivalent of `github>` which can be skipped because Renovate is configured on GitHub

Issue #13868, opened 2025-07-07
Add user defined universal outbounds to `_layout` endpoint


Automaat:
### Description

Right now, we only have a list of dynamically computed outbound. We need to provide information about user defined outbounds in universal setup.

PR #14081, opened 2025-07-29
feat(api): add universal outbounds to _layout endpoint

Automaat:
## Motivation

We need to add user defined outbounds to `_layout` api if user is not using transparent proxy

Fix: https://github.com/kumahq/kuma/issues/13868

### DIFF
--- a/pkg/xds/context/destination_index.go
+++ b/pkg/xds/context/destination_index.go
@@ -44,7 +44,7 @@ func NewDestinationIndex(resources ...[]core_model.Resource) *DestinationIndex {
 }
 
 // GetReachableBackends return map of reachable port by its KRI, and bool to indicate if any backend were match or all destinations were returned
-func (dc *DestinationIndex) GetReachableBackends(dataplane *core_mesh.DataplaneResource) (map[kri.Identifier]core.Port, bool) {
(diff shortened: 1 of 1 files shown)


pkg/xds/context/destination_index.go
```diff
@@ -111,6 +140,28 @@ func (dc *DestinationIndex) GetDestinationByKri(id kri.Identifier) core.Destinat
...
 
+// ResolveResourceIdentifier resolves one resource identifier based on the labels.
+// If multiple resources match the labels, the oldest one is returned.
+// The reason is that picking the oldest one is the less likely to break existing traffic after introducing new resources.
+func (dc *DestinationIndex) ResolveResourceIdentifier(resType core_model.ResourceType, labels map[string]string) *kri.Identifier {
+	if len(labels) == 0 {
+		return nil
+	}
+	var oldestCreationTime *time.Time
+	var oldestTri *kri.Identifier
```
slonka: nitpick: is this a typo?

```suggestion
	var oldestKri *kri.Identifier
```

Automaat: it was copied, renamed to better match new context

PR #14082, opened 2025-07-29
feat(api): add _policies endpoint to get routes policies config

Automaat:
## Motivation

Next part of implementing new Inspect API

Fix https://github.com/kumahq/kuma/issues/13867


pkg/api-server/resource_endpoints.go
```diff
@@ -174,6 +174,13 @@ func (r *resourceEndpoints) addFindEndpoint(ws *restful.WebService, pathPrefix s
 			Param(ws.PathParameter("outbound_kri", "KRI of a outbound").DataType("string")).
 			Returns(200, "OK", nil).
 			Returns(404, "Not found", nil))
+		ws.Route(ws.GET(pathPrefix+"/{name}/outbounds/{outbound_kri}/_routes/{route_kri}/_policies").To(r.getPoliciesConf(ordered.Policies, matchedPoliciesToRouteConfig)).
```
Automaat, line 177: we could simplify this endpoint as we don't actually need outbound kri for this. We could do:
```
"/{name}/routes/{route_kri}/_policies"
```

pkg/api-server/testdata/resources/inspect/dataplanes/_policies/routes/multiple_merged_policies_different_to.golden.json
```diff
@@ -0,0 +1,17 @@
...
+   {
+    "http": {
+     "requestTimeout": "20s"
+    }
+   }
+  ],
+  "kind": "MeshTimeout",
+  "origins": [
+   {
+    "kri": "kri_mt_default___mt-1_"
```
slonka, line 13: aren't we missing mt-2?

slonka: would the resulting Envoy config have 

```
streamIdleTimeout: 10s
```

?

Automaat: We are not missing it, this is what actually will be applied in plugin

Automaat: I just checked it again and that is how this is applied, we are not merging real resources, so this conig is correct

Issue #10212, opened 2024-05-10
Strategy on sync name conflict


jakubdyszkiewicz:
### Description

When we apply policies on global and zone, we hash them so it's very unlikely that we get into name conflicts. However, it's still possible with the following steps

1) Apply object on global named xyz, see that synced name in zone is xyz-hashname
2) I delete object on global
3) I apply object on zone with name xyz-hashname. It’s synced to global but that does not matter here
4) I apply the same object on global named xyz

Let see what happens. We need to have a defined strategy for such case.

Automaat:
Applying steps from issue description ends with error on zone:
```
2024-05-17T08:28:55.172Z    INFO    kds-delta-zone    creating a new resource from upstream    {"type": "MeshTimeout", "name": "mesh-timeout-all-default-bv8z44b7bzb99xc7.kong-mesh-system", "mesh": "default"}
2024-05-17T08:28:55.292Z    INFO    kds-delta-zone    error during callback received, sending NACK    {"kds-version": "v2", "err": "resource already exists: type=\"MeshTimeout\" name=\"mesh-timeout-all-default-bv8z44b7bzb99xc7.kong-mesh-system\" mesh=\"default\""}
```

slonka:
> > Triage: rejection is fine. It would be great to bubble up this error better than logs.

Is it the idea of having a "status" field (that we had somewhere) or something else?

lukidzi:
We now reject object if it already exists. We show it in logs but still we could show it better

lobkovilya:
> triage: Kuma handles conflicts the correct way, resources are rejected. The issue is that NACK is not propagated properly. We are not planning to fix it now

PR #14083, opened 2025-07-29
chore(kds): define user and internal errors and log user errors as info

slonka:
## Motivation

There are 2 reasons for NACK:
1. We've messed up with validation/feature-flags on Global. User applies resource that's considered valid from Global's point of view and invalid from Zone CP point of view. That shouldn't happen normally. **We'd like to see these errors in logs (on error level)**.
2. On Zone CP user created resource with the name `mesh-timeout-all-default-bv8z44b7bzb99xc7` and that's exactly the same name as `mesh-timeout-all-default` will have when synced from Global to Zone (original issue is https://github.com/kumahq/kuma/issues/10212). In that case, the collision is created on user's side. It looks like we can't do anything about it. We probably **shouldn't log this as an error** in this case.

## Implementation information

Add "ErrUserNack" and "ErrInternalNack" and join the correct errors.

## Supporting documentation

xrel: https://github.com/Kong/kong-mesh/issues/7890 https://github.com/kumahq/kuma/issues/10212


lukidzi (approved):

slonka:
Tested locally:

image

bartsmykla (approved):

Issue #8748, opened 2024-01-04
Potential to trigger ConfigMapReconciler less on MeshGateway changes


lobkovilya:
### What happened?

1. Kuma CP enqueues events for all MeshGateways even though we can do this only for MeshGateways with crossMesh listeners https://github.com/kumahq/kuma/blob/beb3c363def6e741803285e0ea9d445595b638b8/pkg/plugins/runtime/k8s/controllers/configmap_controller.go#L157

2. By looking at the algorithm that generates VIPs it seems like we don't need to watch MeshGatewayRoutes https://github.com/kumahq/kuma/blob/beb3c363def6e741803285e0ea9d445595b638b8/pkg/plugins/runtime/k8s/controllers/configmap_controller.go#L95

jakubdyszkiewicz:
> Triage: @michaelbeaumont to check if need this

slonka:
Is this all https://github.com/kumahq/kuma/pull/14095/files#diff-773a586d5ff0b30db8ea18424c875571a12bf162ba38f4bb15ddb73d19f3bb74L96 ?

slonka:
> Triage: Might not need the reconciler at all when we use `MeshService: exclusive`

PR #14095, opened 2025-07-30
perf(kuma-cp): do not trigger on MeshGatewayRoute

slonka:
## Motivation

Closes https://github.com/kumahq/kuma/issues/8748

## Implementation information

Remove trigger


slonka:
@lobkovilya @bartsmykla @Automaat - can we move forward with this?

Automaat (approved):

PR #14104, opened 2025-07-31
feat(inbound-rules): stop merging inbound rules for internal representation

Automaat:
## Motivation

We need to stop merging inbound rules to be able to implement SpiffeId support in MeshTrafficPermission. Rest of the policies that utilize rules will work as previosly 

## Supporting documentation

Fix #XX


slonka (approved):
> old Inspect API endpoint _rules is going to return unmerged rules, but schema is the same, so it's not a breaking change

Does this affect GUI in any way? Does it just mean that we're going to display a list of rules instead of just one?

Automaat:
@slonka Since we don't merge rules it will be displayed differently in gui but we don't change api so GUI will work out of the box. This was described in MADR https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/080-inbound-policies-matches.md

PR #14125, opened 2025-08-04
feat(kuma-cp): rename mesh metric system resource to follow unified naming format

slonka:
## Motivation

https://github.com/kumahq/kuma/issues/13968

## Implementation information

Rename Mesh Metric resources.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13968

### DIFF
--- a/pkg/core/kri/kri.go
+++ b/pkg/core/kri/kri.go
@@ -106,6 +106,11 @@ func (i Identifier) HasSectionName() bool {
 }
 
 func WithSectionName(id Identifier, sectionName string) Identifier {
+	// cannot add section name to empty identifier
+	if id == (Identifier{}) {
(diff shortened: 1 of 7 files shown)


pkg/core/kri/kri.go
```diff
@@ -106,6 +106,11 @@ func (i Identifier) HasSectionName() bool {
 }
 
 func WithSectionName(id Identifier, sectionName string) Identifier {
+	// cannot add section name to empty identifier
+	if id == (Identifier{}) {
+		return id
+	}
```
slonka, line 112: same change is in https://github.com/kumahq/kuma/pull/14128/files but there is a bug there so to speed  this PR up I'm adding it here

Issue #14025, opened 2025-07-22
MeshTrafficPermission: implement API support for SpiffeID matches


lobkovilya:
### Description

YAML example is in [MADR-80](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/080-inbound-policies-matches.md)

```yaml
type: MeshTrafficPermission
mesh: default
name: by-service-owner
spec:
  targetRef:
     kind: Dataplane
     labels:
        app: backend
     sectionName: http-port
  rules:
    - default:
        deny:
           - spiffeId:
               type: Exact
               value: "spiffe://trust-domain.mesh/ns/default/sa/api-gateway"
        allowWithShadowDeny:
           - spiffeId:
               type: Prefix
               value: "spiffe://trust-domain.mesh/ns/legacy"
        allow:
           - spiffeId:
               type: Prefix
               value: "spiffe://trust-domain.mesh/"
```

we'll need a new `Default` conf type for MeshTrafficPermission

```go
type RuleConf struct {
    Deny                *[]common_api.Match `json:"deny,omitempty"`
    AllowWithShadowDeny *[]common_api.Match `json:"allowWithShadowDeny,omitempty"`
    Allow               *[]common_api.Match `json:"allow,omitempty"`
}
```

`Match` has to be introduced in the [common](https://github.com/kumahq/kuma/tree/4004a7231090fb2786d5ede41b3d95b73188d745/api/common) package as it'll be used later in other inbound policies

```go
// +kubebuilder:validation:Enum=Exact;PathPrefix
type SpiffeIdMatchType string

type SpiffeIdMatch struct {
    Type  SpiffeIdMatchType `json:"type"`
    Value string            `json:"value"`
}

type Match struct {
    SpiffeId *SpiffeIdMatch `json:"spiffeId,omitempty"`
}
```

PR #14127, opened 2025-08-04
feat(MeshTrafficPermission): implement new rules api

Automaat:
## Motivation

Adding new rules api for MeshTrafficPermission as described in [MADR](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/080-inbound-policies-matches.md)

Fix https://github.com/kumahq/kuma/issues/14025


api/common/v1alpha1/match.go
```diff
@@ -0,0 +1,16 @@
+package v1alpha1
+
+type Match struct {
+	// SpiffeId defines a matcher configuration for SpiffeId matching
+	SpiffeId *SpiffeIdMatch `json:"spiffeId,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=Exact;PathPrefix
```
lukidzi: Should it be `PathPrefix` or maybe just `Prefix`?

Automaat: that was in the issue, but I checked madr again and it should be `Path` updated

pkg/core/resources/apis/mesh/validators.go
```diff
@@ -449,6 +450,17 @@ func ValidateTargetRef(
 	return err
 }
 
+func ValidateMatch(match common_api.Match) validators.ValidationError {
```
lukidzi, line 453: I think we also need to validate type since kubevalidator works only on k8s and not universal

Automaat: I am not if we need to do this We don't validate enums in any other policy

Automaat: It is validated with open api schema we generate from it

lukidzi (approved):

PR #14128, opened 2025-08-04
refactor(KRI): replace `*kri.Identifier` with value-based identifiers in policy logic

bartsmykla:
## Motivation

The codebase used `*kri.Identifier` in many places across policy handling, which made the code more complex and less consistent. This change replaces all usages with `kri.Identifier` to simplify the API, remove unnecessary pointer handling, and improve clarity when resolving references in matchers, resolvers, and route generators.

## Implementation information

Replaced all uses of `*kri.Identifier` with `kri.Identifier` throughout the policy system. Updated the resolution of `TargetRef` and `BackendRef` to use `kri.Identifier` directly instead of `core_model.ResourceMeta`, making the origin of references clearer.

Simplified the `kri.From` and `kri.FromResourceMeta` functions by removing the `sectionName` parameter. Introduced a new helper `kri.WithSectionName(...)` for cases where a section name needs to be set. This helper works with both string and numeric names and avoids modifying empty identifiers.

All related code and tests were updated accordingly.


pkg/plugins/policies/core/rules/resolve/backendref.go
```diff
@@ -28,24 +28,27 @@ func BackendRef(meta core_model.ResourceMeta, br common_api.BackendRef, resolver
...
+	if rr.Resource.IsEmpty() {
+		return ResolvedBackendRef{}, false
 	}
 
-	if br.Port != nil {
-		rr.Resource.SectionName = fmt.Sprintf("%d", *br.Port)
+	if sn := pointer.Deref(br.SectionName); sn != "" {
+		rr.Resource.SectionName = sn
+	} else if p := pointer.Deref(br.Port); p > 0 {
+		rr.Resource.SectionName = fmt.Sprintf("%d", p)
```
slonka: this is different, let's check if this causes e2e to fail

pkg/plugins/policies/core/rules/resolve/backendref.go
```diff
@@ -42,10 +42,8 @@ func BackendRef(origin kri.Identifier, br common_api.BackendRef, resolver LabelR
 		return ResolvedBackendRef{}, false
 	}
 
-	if sn := pointer.Deref(br.SectionName); sn != "" {
```
slonka: do we have an issue about this potential bug?

bartsmykla: not yet, will open soon

PR #14140, opened 2025-08-05
chore(mise): adjust make targets and gha after migration to mise

Icarus9913:
## Motivation

This is a cleanup operation to adjust make targets and gha since we moved to use mise


Icarus9913:
I've discussed with @bartsmykla to add the `skip-test` label because we still need another PR to make the github action work well

PR #14112, opened 2025-08-01
fix(helm): properly set addresses when a list is provided

lukidzi:
## Motivation

Seems like values from the list are not set correctly: https://github.com/kumahq/kuma/pull/13965/files

## Implementation information

Use range to iterate over elements

PR #14141, opened 2025-08-05
fix: fix loadbalancersourcerange templating

facchettos:
This fixes an issue that arises when the loadBalancerSourceRanges field is set with more than one value. Old behavior created a string with all the ranges in it which resulted in an invalid CIDR

## Motivation
helm chart is broken and I need to deploy stuff with loadbalancersourceranges

## Supporting documentation

Fix #XX


Icarus9913:
@facchettos Hi, there's one PR already landed on the master branch to solve the `loadBalancerSourceRanges` field for ingress service:  https://github.com/kumahq/kuma/pull/14112

PR #14143, opened 2025-08-05
fix(tools): revert mise

Automaat:
## Motivation

revert mise temporarily

> Changelog: skip


jijiechen:
Adding some context: we are trying to release a patch sooner as some features are eagerly asked by users, and the `mise` changes were not complete, so we need to revert them temporarily and add them back after the coming patch release.

Issue #14026, opened 2025-07-22
MeshTrafficPermission: plugin support for SpiffeId matches


lobkovilya:
### Description

The algorithm is explained in [MADR-80#Algorithm](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/080-inbound-policies-matches.md#algorithm).

Please use new [builders](https://github.com/kumahq/kuma/tree/79b2415831544448a880f3dcb75996230a96fc8f/pkg/envoy/builders) packages when building Matching API configuration. This code will be highly reusable when implementing `matches` for other inbound policies.

PR #14147, opened 2025-08-05
feat(MeshTrafficPermission): spiffeId matcher support

Automaat:
## Motivation

Implementing plugin support in MeshTrafficPermission as described in MADR: https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/080-inbound-policies-matches.md#algorithm

Fix: https://github.com/kumahq/kuma/issues/14026


pkg/plugins/policies/meshtrafficpermission/xds/rbac_configurer_test.go
```diff
@@ -0,0 +1,413 @@
...
+                    name: default
+    shadowMatcher:
+        onNoMatch:
+            action:
+                name: envoy.filters.rbac.action
+                typedConfig:
+                    '@type': type.googleapis.com/envoy.config.rbac.v3.Action
+                    action: DENY
+                    name: default
+    statPrefix: allow_all_prefix.`,
```
slonka: is this how it used to work? that we shadow deny things that are allowed? 🤔

pkg/plugins/policies/meshtrafficpermission/xds/rbac_configurer_test.go
```diff
@@ -0,0 +1,413 @@
...
+                            typedConfig:
+                                '@type': type.googleapis.com/envoy.extensions.matching.common_inputs.ssl.v3.UriSanInput
+                        valueMatch:
+                            exact: spiffeId://trust-domain.mesh/ns/backend/v1
+                - onMatch:
+                    action:
+                        name: envoy.filters.rbac.action
+                        typedConfig:
+                            '@type': type.googleapis.com/envoy.config.rbac.v3.Action
+                            name: kri_mtp_default_zone-1_ns-1_mtp-2_
```
slonka, line 287: why is there no `action` field here? is `ALLOW` the default so it's not needed? 🤔

Automaat: if you set allow action it is removed from output yaml as this is default

pkg/plugins/policies/meshtrafficpermission/xds/rbac_configurer_test.go
Automaat: changed

slonka (approved):

Issue #13876, opened 2025-07-07
Implement `Bundled` provider for MeshIdentity


lukidzi:
### Description

Based on [MADR](https://github.com/kumahq/kuma/pull/13804) we should implement Bundled provider which allows providing certificates and autogenerate them

PR #14149, opened 2025-08-05
feat(plugins): add ability to register core resources plugins

lukidzi:
## Motivation

Once MeshIdentity is implemented we need to create some resources (Secrets, Cluster - Spire agent) it makes sense to have specific generator for them.

## Implementation information

* Add tag which allows registering generators in core resources
* Add option to distinguish between specific plugins 

## Supporting documentation

Part of #13876

### DIFF
--- a/pkg/core/plugins/initializer.go
+++ b/pkg/core/plugins/initializer.go
@@ -25,3 +25,16 @@ func Init(enabledPlugins []string, plugins map[string]*PluginInitializer) {
 		}
 	}
 }
+
+func InitAllIf(enabledPlugins []string, pluginName string, plugins map[string]*PluginInitializer) {
+	for _, plugin := range enabledPlugins {
+		if plugin == pluginName {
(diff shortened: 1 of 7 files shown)


pkg/core/plugins/initializer.go
```diff
@@ -25,3 +25,16 @@ func Init(enabledPlugins []string, plugins map[string]*PluginInitializer) {
...
+
+func InitAllIf(enabledPlugins []string, pluginName string, plugins map[string]*PluginInitializer) {
+	for _, plugin := range enabledPlugins {
+		if plugin == pluginName {
+			for _, initializer := range plugins {
+				if !initializer.Initialized {
+					initializer.InitFn()
+					initializer.Initialized = true
+				}
+			}
```
slonka, line 37: I guess you could break after the plugin is found

```suggestion
			}
			break;
```

PR #14150, opened 2025-08-05
feat(xds): added builders for tls context

lukidzi:
## Motivation

We introduced builders for Envoy config which are more flexibable and reusable.

## Implementation information

* added builders for Downstream and Upstream TLS context
* added new configures option to override downstream and upstream context
## Supporting documentation

Part of #13876

### DIFF
--- a/pkg/envoy/builders/auth/secret.go
+++ b/pkg/envoy/builders/auth/secret.go
@@ -78,7 +78,7 @@ func CertificateChain(datasource *Builder[envoy_core.DataSource]) Configurer[env
 	return func(s *envoy_auth.Secret_TlsCertificate) error {
 		ds, err := datasource.Build()
 		if err != nil {
-			return nil
+			return err
 		}
 		if s.TlsCertificate == nil {
 			s.TlsCertificate = &envoy_auth.TlsCertificate{}
@@ -92,7 +92,7 @@ func PrivateKey(datasource *Builder[envoy_core.DataSource]) Configurer[envoy_aut
 	return func(s *envoy_auth.Secret_TlsCertificate) error {
 		ds, err := datasource.Build()
 		if err != nil {
-			return nil
+			return err
 		}
 		if s.TlsCertificate == nil {
 			s.TlsCertificate = &envoy_auth.TlsCertificate{}
@@ -110,7 +110,7 @@ func TrustDomainBundle(trustDomain string, datasource *Builder[envoy_core.DataSo
 	return func(s *envoy_auth.SPIFFECertValidatorConfig_TrustDomain) error {
 		ds, err := datasource.Build()
 		if err != nil {
-			return nil
+			return err
 		}
 		s.Name = trustDomain
 		s.TrustBundle = ds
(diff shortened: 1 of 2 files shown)


pkg/envoy/builders/auth/secret.go
```diff
@@ -0,0 +1,119 @@
...
+
+func NewTlsCertificate() *Builder[envoy_auth.Secret_TlsCertificate] {
+	return &Builder[envoy_auth.Secret_TlsCertificate]{}
+}
+
+func CertificateChain(datasource *Builder[envoy_core.DataSource]) Configurer[envoy_auth.Secret_TlsCertificate] {
+	return func(s *envoy_auth.Secret_TlsCertificate) error {
+		ds, err := datasource.Build()
+		if err != nil {
+			return nil
```
slonka: why are you returning nil when the `err` is not nil?

slonka:
```suggestion
			return err
```

pkg/envoy/builders/auth/secret.go
```diff
@@ -0,0 +1,119 @@
...
+		s.TlsCertificate.CertificateChain = ds
+		return nil
+	}
+}
+
+func PrivateKey(datasource *Builder[envoy_core.DataSource]) Configurer[envoy_auth.Secret_TlsCertificate] {
+	return func(s *envoy_auth.Secret_TlsCertificate) error {
+		ds, err := datasource.Build()
+		if err != nil {
+			return nil
```
slonka: same as above

```suggestion
			return err
```

pkg/envoy/builders/auth/secret.go
```diff
@@ -0,0 +1,119 @@
...
+
+func NewSPIFFECertValidator() *Builder[envoy_auth.SPIFFECertValidatorConfig_TrustDomain] {
+	return &Builder[envoy_auth.SPIFFECertValidatorConfig_TrustDomain]{}
+}
+
+func TrustDomainBundle(trustDomain string, datasource *Builder[envoy_core.DataSource]) Configurer[envoy_auth.SPIFFECertValidatorConfig_TrustDomain] {
+	return func(s *envoy_auth.SPIFFECertValidatorConfig_TrustDomain) error {
+		ds, err := datasource.Build()
+		if err != nil {
+			return nil
```
slonka:
```suggestion
			return err
```

pkg/envoy/builders/tls/tls.go
```diff
@@ -0,0 +1,202 @@
...
+func NewDefaultValidationContext() *Builder[envoy_tls.CertificateValidationContext] {
+	return &Builder[envoy_tls.CertificateValidationContext]{}
+}
+
+func SANs(builders []*Builder[envoy_tls.SubjectAltNameMatcher]) Configurer[envoy_tls.CertificateValidationContext] {
+	return func(c *envoy_tls.CertificateValidationContext) error {
+		for _, builder := range builders {
+			config, err := builder.Build()
+			if err != nil {
+				return nil
```
slonka:
```suggestion
				return err
```

slonka (commented):
I think you're missing some error returns

PR #14171, opened 2025-08-07
feat(meshidentity): bundled provider implementation

### LINKED ISSUE #13876: Implement `Bundled` provider for MeshIdentity

lukidzi:
## Motivation

Implement MeshIdentity provider - bundled, it's a first part which creates a provider but it doesn't use it yet.

## Implementation information

* Register provider
* Create autogenerated certificates and allow creating workload identities
* Added a WorkloadIdentity field into proxy object so we can store information next to the proxy not in one global storage (SecretTracker)

## Supporting documentation

Fix #13876


pkg/core/resources/apis/meshidentity/api/v1alpha1/helpers.go
```diff
@@ -0,0 +1,137 @@
...
+func (i *MeshIdentity) getSpiffeIDTemplate() string {
+	builder := strings.Builder{}
+	builder.WriteString("spiffe://")
+	if i.SpiffeID != nil {
+		builder.WriteString("{{ .TrustDomain }}")
+		if i.SpiffeID.Path == nil {
+			builder.WriteString(defaultPathTemplate)
+		} else {
+			builder.WriteString(pointer.Deref(i.SpiffeID.Path))
+		}
```
slonka: nitpick: you could use `pointer.DerefOr`:

```suggestion
		builder.WriteString(pointer.DerefOr(i.SpiffeID.Path, defaultPathTemplate))
```

pkg/core/resources/apis/meshidentity/generator/v1alpha1/generator_test.go
```diff
@@ -0,0 +1,86 @@
...
+	resourceSet *core_xds.ResourceSet,
+	typ envoy_resource.Type,
+) []byte {
+	resources, err := resourceSet.ListOf(typ).ToDeltaDiscoveryResponse()
+	Expect(err).ToNot(HaveOccurred())
+	actual, err := util_proto.ToYAML(resources)
+	Expect(err).ToNot(HaveOccurred())
+
+	return actual
+}
```
slonka: Should we move this to some sort of `util`? I think we have this in a bunch of places already 🤔

pkg/core/resources/apis/meshidentity/providers.go
```diff
@@ -0,0 +1,10 @@
+package meshidentity
+
+import (
+	"github.com/kumahq/kuma/pkg/core/plugins"
+	"github.com/kumahq/kuma/pkg/core/resources/apis/meshidentity/providers/bundled"
+)
+
+var NameToModule = map[string]*plugins.PluginInitializer{
+	"bundled": {InitFn: bundled.InitProvider, Initialized: false},
```
slonka, line 13: This is not generated right? Should we put some comment somewhere that when you add a new provider you need to manually update this list?

pkg/core/resources/apis/meshidentity/providers/bundled/autogenerated.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+
+	"github.com/kumahq/kuma/pkg/core"
+	core_ca "github.com/kumahq/kuma/pkg/core/ca"
+	util_tls "github.com/kumahq/kuma/pkg/tls"
+)
+
+const (
+	DefaultCACertValidityPeriod = 10 * 365 * 24 * time.Hour
+	DefaultKeySize              = 2048
```
slonka: Nitpick: Should this be private? I don't think we use them outside of here.

pkg/core/resources/apis/meshidentity/providers/bundled/autogenerated.go
```diff
@@ -0,0 +1,78 @@
...
+	domain, err := spiffeid.TrustDomainFromString(trustDomain)
+	if err != nil {
+		return nil, err
+	}
+	uri, err := spiffeid.FromSegments(domain)
+	if err != nil {
+		return nil, err
+	}
+	tmpl := &x509.Certificate{
+		SerialNumber: big.NewInt(0),
```
slonka: this is supposed to be unique, is this correct?

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
```diff
@@ -0,0 +1,394 @@
...
+		logger:          logger,
+		roSecretManager: roSecretManager,
+		secretManager:   secretManager,
+		cache:           c,
+		zone:            zone,
+	}, nil
+}
+
+func (b *bundledIdentityProvider) Validate(ctx context.Context, identity *meshidentity_api.MeshIdentityResource) error {
+	if !pointer.DerefOr(identity.Spec.Provider.Bundled.InsecureAllowSelfSigned, false) {
```
slonka, line 84: are we sure everything from `identity.Spec.Provider.Bundled` is not nil? 🤔

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
```diff
@@ -0,0 +1,394 @@
...
+	workloadCert, err := x509.CreateCertificate(rand.Reader, template, caCert, publicKey, caPrivateKey)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to generate X509 certificate")
+	}
+	identityPair, err := util_tls.ToKeyPair(privateKey, workloadCert)
+	if err != nil {
+		return nil, err
+	}
+	identifier := kri.From(identity)
+	resources, err := additionalResources(identifier.String(), identityPair)
```
slonka, line 259: shouldn't the name be `system_<kri>`?

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+var maxUint128, one *big.Int
+
+func init() {
+	one = big.NewInt(1)
+	m := new(big.Int)
+	m.Lsh(one, 128)
+	maxUint128 = m.Sub(m, one)
+}
```
slonka, line 351: This is repeated, should we dry it up?

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
```diff
@@ -0,0 +1,394 @@
...
+
+func isSelfSigned(certPEM []byte) (bool, error) {
+	block, _ := pem.Decode(certPEM)
+	if block == nil || block.Type != "CERTIFICATE" {
+		return false, errors.New("failed to decode PEM certificate")
```
slonka: Error message is a bit misleading, could be:

```suggestion
		return false, errors.New("failed to decode PEM block or block is not of type CERTIFICATE")
```

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
lukidzi: I would prefer to keep it local since it's easier to locate it and don't need to jump over project

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
lukidzi: We already validate these values during creation. This check is only performed during reconciliation for values that cannot be validated at creation time.

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
lukidzi: In the MADR I had KRI without system prefix

pkg/core/resources/apis/meshidentity/providers/bundled/bundled_test.go
```diff
@@ -0,0 +1,231 @@
...
+			meshIdentity := builders.MeshIdentity().
+				WithName("matching-1").
+				WithSelector(&v1alpha1.LabelSelector{
+					MatchLabels: &map[string]string{
+						"app":     "test-app",
+						"version": "v1",
+					},
+				}).WithBundledAutoGenerated().Build()
+			// create MeshIdentity
+			Expect(resourceManager.Create(context.TODO(), meshIdentity, core_store.CreateBy(model.MetaToResourceKey(meshIdentity.Meta)))).To(Succeed())
```
slonka, line 82: Do you remember the reason we're not using [SpecContext](https://onsi.github.io/ginkgo/#spec-timeouts-and-interruptible-nodes)?

pkg/core/resources/apis/meshidentity/providers/manager.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (i *IdentityProviderManager) SelectedIdentity(dataplane *core_mesh.DataplaneResource, identities []*meshidentity_api.MeshIdentityResource) *meshidentity_api.MeshIdentityResource {
+	identity, _ := meshidentity_api.Matched(dataplane.Meta.GetLabels(), identities)
+	return identity
+}
+
+func (i *IdentityProviderManager) GetWorkloadIdentity(ctx context.Context, proxy *xds.Proxy, identity *meshidentity_api.MeshIdentityResource) (*xds.WorkloadIdentity, error) {
+	if identity == nil {
+		return nil, nil
```
slonka, line 35: this looks a bit weird, shouldn't we return an error? 🤔 also I don't think this function is used anywhere

pkg/core/xds/types.go
```diff
@@ -201,6 +208,41 @@ func (p *Proxy) GetTransparentProxy() *tproxy_dp.DataplaneConfig {
 	return tproxy_dp.GetDataplaneConfig(p.Dataplane, p.Metadata)
 }
 
+type ManageType string
```
slonka: nitpick: this sounds a bit awkward, maybe

```suggestion
type ManagementMode string
```

or 

```suggestion
type ManagementType string
```

?

slonka (approved):
A couple of Qs / nitpicks

pkg/core/resources/apis/meshidentity/providers/bundled/bundled_test.go
lukidzi: Not sure :/

pkg/core/resources/apis/meshidentity/providers/manager.go
lukidzi: It's not used now, but its going to be used in another PR

lukidzi: Maybe I will refactor it in a PR with usage to not enter once identity is empty (since it's not an error to be not selected by the MeshIdentity policy)

slonka (approved):

PR #14180, opened 2025-08-08
feat(kuma-cp): contextual (unified) inbound names and route config

bartsmykla:
## Motivation

Inbound resources in Kuma currently use service or port-based names that do not provide enough context when unified naming is enabled. This makes it harder to trace and correlate inbound clusters, listeners, and route configurations across the control plane and Envoy configuration. We also need a consistent way to reference inbound resources in the API server and tests, ensuring they match the names used in the data plane.

## Implementation information

Introduced a new helper `naming.ContextualInbound` to generate contextual inbound names in the format `self_inbound_<section>`. This helper is used in the API server layout endpoint to populate the new `proxyResourceName` field for dataplane inbounds.

When unified naming is enabled:
- Inbound cluster names, listener names, statPrefix, route config names, and virtual host names now use the contextual inbound name
- HTTP inbound route configuration takes explicit `routeConfigName` and `virtualHostName` parameters instead of deriving them internally

Refactored the inbound proxy generator to use the generic listener builder and set statPrefix explicitly. Updated `HttpInboundRoutes` and its v3 implementation to support the new signature.

Updated all relevant policy plugin tests to pass explicit route and virtual host names, and refreshed API server golden files to include `proxyResourceName` values.

## Supporting documentation

Closes https://github.com/kumahq/kuma/issues/14007


pkg/core/naming/contextual.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func ContextualInboundName[T sectionName](r core_model.Resource, sectionName T) (string, error) {
+	if r == nil {
+		return "", errors.New("cannot build contextual inbound name: resource is nil")
+	}
+
+	desc, err := core_registry.Global().DescriptorFor(r.Descriptor().Name)
+	if err != nil {
+		return "", errors.Wrapf(
```
slonka, line 34: Is it any different than what was mentioned in error wrapping madr? https://github.com/kumahq/kuma/blob/3220f15dad5ecf795c5ff27321e4885830c66bf6/docs/madr/decisions/073-error-wrapping.md#L118

pkg/core/naming/contextual_test.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+var _ = Describe("ContextualInboundName", func() {
+	It("should return error when resource is nil", func() {
+		// when
+		name, err := naming.ContextualInboundName(nil, "http")
+
+		// then
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("resource is nil"))
```
slonka, line 31: nitpick: I know there's `MatchError` as well, does it give us anything better than this?

pkg/xds/generator/inbound_proxy_generator.go
```diff
@@ -29,6 +30,7 @@ type InboundProxyGenerator struct{}
 
 func (g InboundProxyGenerator) Generate(ctx context.Context, _ *core_xds.ResourceSet, xdsCtx xds_context.Context, proxy *core_xds.Proxy) (*core_xds.ResourceSet, error) {
 	resources := core_xds.NewResourceSet()
+	unifiedNaming := proxy.Metadata.HasFeature(xds_types.FeatureUnifiedResourceNaming)
```
slonka, line 33: nitpick: we usually wrap this in `getNameOrDefault`

slonka (approved):

pkg/xds/generator/inbound_proxy_generator.go
bartsmykla: ack

pkg/core/naming/contextual_test.go
bartsmykla: ack

pkg/core/naming/contextual.go
bartsmykla: ack

PR #14199, opened 2025-08-11
feat(kuma-cp): fix missing unified naming for secret resources

bartsmykla:
## Motivation

Some code paths missed updating secret resource names to use the unified naming feature introduced in #14166. This caused inconsistencies in how secrets were referenced during mTLS and filter chain configuration.

## Implementation information

Added the `unifiedResourceNaming` flag to all relevant function calls and configurers that manage secrets, including server-side and client-side mTLS configurers. Updated inbound, outbound, gateway, and external service proxy generators to pass the flag consistently. Adjusted tests to account for the new parameter.

## Supporting documentation

Related to:
- #14166
- #13973

### DIFF
--- a/pkg/xds/generator/prometheus_endpoint_generator.go
+++ b/pkg/xds/generator/prometheus_endpoint_generator.go
@@ -113,7 +113,7 @@ func (g PrometheusEndpointGenerator) Generate(ctx context.Context, _ *core_xds.R
 		case mesh_proto.PrometheusTlsConfig_activeMTLSBackend:
 			listenerBuilder = listenerBuilder.Configure(envoy_listeners.FilterChain(
 				envoy_listeners.NewFilterChainBuilder(proxy.APIVersion, envoy_common.AnonymousResource).Configure(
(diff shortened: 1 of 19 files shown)


pkg/xds/generator/prometheus_endpoint_generator.go
```diff
@@ -113,7 +113,7 @@ func (g PrometheusEndpointGenerator) Generate(ctx context.Context, _ *core_xds.R
 		case mesh_proto.PrometheusTlsConfig_activeMTLSBackend:
 			listenerBuilder = listenerBuilder.Configure(envoy_listeners.FilterChain(
 				envoy_listeners.NewFilterChainBuilder(proxy.APIVersion, envoy_common.AnonymousResource).Configure(
-					envoy_listeners.ServerSideMTLS(xdsCtx.Mesh.Resource, proxy.SecretsTracker, nil, nil),
+					envoy_listeners.ServerSideMTLS(xdsCtx.Mesh.Resource, proxy.SecretsTracker, nil, nil, false),
```
slonka: shouldn't this be also plugged in?

bartsmykla: addressed

Issue #13883, opened 2025-07-07
Create MeshTrust based on MeshIdentity


lukidzi:
### Description

We decided to create a MeshTrust based on MeshIdentity, we should implement it

PR #14216, opened 2025-08-12
feat(meshidentity): set MeshService TLSReadiness and Identity

lukidzi:
## Motivation

We were setting Identity and TLSReadiness base on DataplaneInsight and service tag. DataplaneInsights should be used only to display information in the UI for traffic shaping.

## Implementation information

* Set a SpiffeID in MeshService if there is MeshIdentity targeting this dataplane
* Dont set service tag if SpiffeID is set and mesh has no tls
* Set TLSReady if there is identity targeting all dataplanes for specific MeshService and there is a MeshTrust for this trust domain

## Supporting documentation

part of #13883 

> Changelog: feat(MeshIdentity): set MeshService TLSReadiness and Identity

### DIFF
--- a/pkg/core/resources/apis/meshservice/status/updater.go
+++ b/pkg/core/resources/apis/meshservice/status/updater.go
@@ -216,7 +216,7 @@ func (s *StatusUpdater) buildTLS(
 			Status: meshservice_api.TLSNotReady,
 		}
 	}
-	if (existing.Status == meshservice_api.TLSReady && (mesh.MTLSEnabled() || len(meshIdentities) > 0)) {
+	if existing.Status == meshservice_api.TLSReady && (mesh.MTLSEnabled() || len(meshIdentities) > 0) {
 		// If mTLS is enabled, the status should go only one way.
 		// Every new instance always starts with mTLS, so we don't want to count issued backends.
 		// Otherwise, we could get into race when new Dataplane did not receive cert yet,
--- a/pkg/core/resources/apis/meshservice/status/updater_test.go
+++ b/pkg/core/resources/apis/meshservice/status/updater_test.go
@@ -129,9 +129,9 @@ var _ = Describe("Updater", func() {
 			err := resManager.Get(context.Background(), ms, store.GetByKey("backend", "tls-mesh"))
 			g.Expect(err).ToNot(HaveOccurred())
 			g.Expect(pointer.Deref(ms.Spec.Identities)).To(ContainElements(meshservice_api.MeshServiceIdentity{
-					Type:  meshservice_api.MeshServiceIdentityServiceTagType,
-					Value: "backend",
-				},
+				Type:  meshservice_api.MeshServiceIdentityServiceTagType,
+				Value: "backend",
+			},
 				meshservice_api.MeshServiceIdentity{
 					Type:  meshservice_api.MeshServiceIdentitySpiffeIDType,
 					Value: "spiffe://tls-mesh.east.mesh.local/ns/my-ns/sa/default",


pkg/core/resources/apis/meshservice/status/updater.go
```diff
@@ -107,6 +109,16 @@ func (s *StatusUpdater) updateStatus(ctx context.Context) error {
 		return errors.Wrap(err, "could not list of Meshes")
 	}
 
+	meshIdentities := meshidentity_api.MeshIdentityResourceList{}
+	if err := s.roResManager.List(ctx, &meshIdentities); err != nil {
+		return errors.Wrap(err, "could not list of MeshIdentities")
```
slonka:
```suggestion
		return errors.Wrap(err, "could not list MeshIdentities")
```

pkg/core/resources/apis/meshservice/status/updater.go
```diff
@@ -107,6 +109,16 @@ func (s *StatusUpdater) updateStatus(ctx context.Context) error {
...
+	}
+	meshTrusts := meshtrust_api.MeshTrustResourceList{}
+	if err := s.roResManager.List(ctx, &meshTrusts); err != nil {
+		return errors.Wrap(err, "could not list of MeshIdentities")
```
slonka:
```suggestion
		return errors.Wrap(err, "could not list MeshTrusts")
```

pkg/core/resources/apis/meshservice/status/updater.go
```diff
@@ -193,18 +203,20 @@ func buildDataplaneProxies(
...
+	trustDomains map[string]struct{},
 ) meshservice_api.TLS {
-	if !mesh.MTLSEnabled() {
+	if !mesh.MTLSEnabled() && len(meshIdentities) == 0 {
 		return meshservice_api.TLS{
 			Status: meshservice_api.TLSNotReady,
 		}
 	}
-	if mesh.MTLSEnabled() && existing.Status == meshservice_api.TLSReady {
+	if (mesh.MTLSEnabled() && existing.Status == meshservice_api.TLSReady) || (len(meshIdentities) > 0 && existing.Status == meshservice_api.TLSReady) {
```
slonka: This could be simplified to:

```suggestion
	if (existing.Status == meshservice_api.TLSReady && (mesh.MTLSEnabled() || len(meshIdentities) > 0)) {
```

right?

pkg/core/resources/apis/meshservice/status/updater.go
```diff
@@ -221,29 +235,81 @@ func buildTLS(
...
+					continue
+				}
+				spiffeIDs[spiffeID] = struct{}{}
+			}
+		}
 	}
 	var identites []meshservice_api.MeshServiceIdentity
-	for _, identity := range maps.SortedKeys(serviceTagIdentities) {
+
+	if (len(spiffeIDs) != 0 && mesh.MTLSEnabled()) || len(spiffeIDs) == 0 {
```
slonka: I think this can be simplified to:

```suggestion
	if len(spiffeIDs) == 0 || mesh.MTLSEnabled() {
```

pkg/core/resources/apis/meshservice/status/updater_test.go
```diff
@@ -68,12 +72,82 @@ var _ = Describe("Updater", func() {
...
+		trustDomain := "test.east.mesh.local"
+		Expect(samples.MeshServiceBackendBuilder().Create(resManager)).To(Succeed())
+		Expect(builders.MeshTrust().WithTrustDomain(trustDomain).Create(resManager)).To(Succeed())
+		Expect(resManager.Create(context.TODO(), mid, store.CreateByKey(mid.Meta.GetName(), "default"))).To(Succeed())
+		Expect(resManager.Create(context.TODO(), samples.DataplaneBackendBuilder().WithMesh("default").Build(), store.CreateByKey("dp-1", "default"), store.CreateWithLabels(map[string]string{
+			metadata.KumaServiceAccount: "default",
+			mesh_proto.KubeNamespaceTag: "my-ns",
+			"app":                       "test",
+		}))).To(Succeed())
+		Expect(samples.DataplaneWebBuilder().Create(resManager)).To(Succeed()) // identity of web should not be added
```
slonka: I don't understand the comment, does the line in any way check that `web` is not added? 🤔

pkg/core/resources/apis/meshservice/status/updater_test.go
```diff
@@ -68,12 +72,82 @@ var _ = Describe("Updater", func() {
...
+			g.Expect(ms.Spec.Identities).To(Equal(&[]meshservice_api.MeshServiceIdentity{
+				{
+					Type:  meshservice_api.MeshServiceIdentityServiceTagType,
+					Value: "backend",
+				},
+				{
+					Type:  meshservice_api.MeshServiceIdentitySpiffeIDType,
+					Value: "spiffe://tls-mesh.east.mesh.local/ns/my-ns/sa/default",
+				},
+			}))
```
slonka: If the array is not sorted then this can fail. We can use 

```suggestion
			g.Expect(pointer.Deref(ms.Spec.Identities)).To(ContainElements(meshservice_api.MeshServiceIdentity{
					Type:  meshservice_api.MeshServiceIdentityServiceTagType,
					Value: "backend",
				},
				meshservice_api.MeshServiceIdentity{
					Type:  meshservice_api.MeshServiceIdentitySpiffeIDType,
					Value: "spiffe://tls-mesh.east.mesh.local/ns/my-ns/sa/default",
				},
			))
```

PR #14217, opened 2025-08-12
feat(meshidentity): create mesh trust based on identity

### LINKED ISSUE #13883: Create MeshTrust based on MeshIdentity

lukidzi:
## Motivation

We want to simplify some logic and introduce a component which creates MeshTrust based on MeshIdentity

## Implementation information

* component check every 5seconds for meshidentity policy and reconcile them
* creates MeshTrust or appends if one already exists

## Supporting documentation

Fix #13883


pkg/core/resources/apis/meshidentity/status/updater.go
```diff
@@ -148,10 +143,101 @@ func (i *IdentityProviderReconciler) initialize(ctx context.Context, mid *meshid
...
+		return err
+	}
+
+	origin := kri.From(identity).String()
+	caPEM := string(ca)
+
+	if update {
+		// Check if the CA PEM is already present in the MeshTrust resource
+		for _, bundle := range meshTrust.Spec.CABundles {
+			if bundle.PEM.Value == caPEM {
```
slonka: we only have one type right now but in the future there might be more, should we check the type here?

slonka: is string comparison completely safe here? is there any chance that there is some additional whitespace char here and this equality could fail? should we `pem.Decode` it?

slonka (commented):
Just 2 Qs

pkg/core/resources/apis/meshidentity/status/updater.go
lukidzi: Yes, should be better, changed

PR #14233, opened 2025-08-13
feat(xds): calculate mTLS insights for MeshIdentity based on events

lukidzi:
## Motivation

With introduction of MeshIdentity we moving away from depending on DataplaneInsights for traffic shaping - this apply only when using MeshIdentity.

## Implementation information

* once we create an identity we sent an event that identity was provided
* when identity is empty we produce an empty event
* dataplane insights store the last event and check if there is a change
* in case of missing event, simple restart of dataplane recalculate the dataplane insight

Part of: #13883


pkg/core/resources/apis/meshidentity/providers/manager_test.go
```diff
@@ -0,0 +1,249 @@
...
+				}).Build(),
+				builders.MeshIdentity().WithName("matching-specific").WithSelector(&v1alpha1.LabelSelector{
+					MatchLabels: &map[string]string{
+						"app": "test-app",
+					},
+				}).Build(),
+			},
+			expectedIdentityName: "matching-specific",
+		}),
+		Entry("select the most specific identity with 2 tags", testCase{
```
slonka: is this test different than:

https://github.com/kumahq/kuma/blob/73c93eb34f1f821738b899758e691f8601b135dd/pkg/core/resources/apis/meshidentity/api/v1alpha1/helpers_test.go#L52-L74

also should there be a test case where the dpp has 3 labels like:

```
dpp: builders.Dataplane().WithLabels(map[string]string{
				"app":     "test-app",
				"version": "v1",
				"env": "prod",
			}).AddInboundHttpOfService("test-app").Build(),
```

and there is MID with:

```
					MatchLabels: &map[string]string{
						"app":     "test-app",
						"version": "v1",
					},
```

and another with:

```
					MatchLabels: &map[string]string{
						"app":     "test-app",
						"env":     "prod",
					},
```

which one would match then? 🤔

pkg/xds/server/callbacks/dataplane_status_sink.go
```diff
@@ -98,31 +109,60 @@ func (s *dataplaneInsightSink) Start(stop <-chan struct{}) {
...
+		if e, ok := event.(events.WorkloadIdentityChangedEvent); ok && e.ResourceKey == dataplaneID {
+			return true
+		}
+		return false
+	})
+	defer listener.Close()
 
-	flush := func(closing bool) {
+	flush := func(closing bool, event *events.WorkloadIdentityChangedEvent) {
+		ctx := context.TODO()
```
slonka, line 130: should you just wire in appContext from `runtime` object from here:

https://github.com/kumahq/kuma/blob/211f511dd2bd81e29f9970dd757ba4e3070c01a0/pkg/xds/server/v3/components.go#L170

?

pkg/xds/server/callbacks/dataplane_status_sink.go
```diff
@@ -98,31 +109,60 @@ func (s *dataplaneInsightSink) Start(stop <-chan struct{}) {
...
-
+		var secretsInfo *secrets.Info
+		switch {
+		case event != nil && event.Operation == events.Delete:
+			usesIdentity = false
+			secretsInfo = nil
+		case event != nil:
+			usesIdentity = true
+			secretsInfo = &secrets.Info{
+				IssuedBackend:     event.Origin.String(),
```
slonka: don't we need to fill out the `ownMesh`, `otherMeshes` etc?

pkg/xds/server/callbacks/dataplane_status_sink.go
```diff
@@ -98,31 +109,60 @@ func (s *dataplaneInsightSink) Start(stop <-chan struct{}) {
...
+				SupportedBackends: s.listMeshTrustBackends(ctx, dataplaneID.Mesh),
+				Expiration:        pointer.Deref(event.ExpirationTime),
+				Generation:        pointer.Deref(event.GenerationTime),
+			}
+		case usesIdentity:
```
slonka: so we get into this case if we also get into 138, this looks a bit weird 🤔

pkg/xds/server/callbacks/dataplane_status_sink_test.go
```diff
@@ -60,6 +62,11 @@ var _ = Describe("DataplaneInsightSink", func() {
 			ticker := &time.Ticker{
 				C: ticks,
 			}
+			metrics, err := core_metrics.NewMetrics("")
```
slonka, line 70: do we trigger any of the new paths in this tests?

https://github.com/kumahq/kuma/blob/8e9d4c16461f57e8da6c008f4aa6c1ce2bb66d56/pkg/xds/server/callbacks/dataplane_status_sink.go#L135-L150

pkg/core/resources/apis/meshidentity/providers/manager_test.go
lukidzi: It's the same test, I will remove this and add to helpers the case you mention

pkg/xds/server/callbacks/dataplane_status_sink.go
lukidzi: No, it's a separate goroutine, and if we provide direct context we might kill it before flushing finishing

slonka: appContext is the context that's for the lifetime of the app so if that's killed then the entire app is killed

pkg/xds/server/callbacks/dataplane_status_sink.go
lukidzi: It's for a cross mesh and cert providing, not used in insights - we don't need it in new cert generation flow

pkg/xds/server/callbacks/dataplane_status_sink_test.go
lukidzi: added test

pkg/xds/server/callbacks/dataplane_status_sink.go
lukidzi: The problem is with the flusing, if we base it on a appContext we might cancel write in the middle, while in a separate context, we can finish it first and later `stop` close the goroutine

slonka (approved):

PR #14240, opened 2025-08-13
feat(meshidentity): enable usage of identity and provide e2e tests

lukidzi:
## Motivation

This PR needs previous ones to be merged, so the diff will be smaller

## Implementation information

* use a workload identity
* introduce a new method in meshcontext `mtlsEnabeld` which checks if mesh has enabled mtls section or we have mesh identities since we might have it
* configure downstream and upstream tls context
* add bundled provider e2e test multizone
* add spire provider e2e test kubernetes


pkg/core/resources/apis/meshidentity/api/v1alpha1/helpers.go
```diff
@@ -62,10 +62,10 @@ func Matched(
 func (i *MeshIdentity) getSpiffeIDTemplate() string {
 	builder := strings.Builder{}
 	builder.WriteString("spiffe://")
+	builder.WriteString("{{ .TrustDomain }}")
```
lukidzi, line 65: This needs to be always for spiffeID template

pkg/xds/topology/outbound.go
```diff
@@ -180,17 +182,15 @@ func fillRemoteMeshServices(
...
+	mtlsEnabled bool,
 ) {
-	ziInstances := map[string]struct{}{}
-
-	if !mesh.MTLSEnabled() {
-		// Ingress routes the request by TLS SNI, therefore for cross
-		// cluster communication MTLS is required.
-		// We ignore Ingress from endpoints if MTLS is disabled, otherwise
-		// we would fail anyway.
+	if !mtlsEnabled {
```
lukidzi, line 187: Since we might have meshidentity instead of mTLS on mesh we need to check both (in mesh context is a helper function)

slonka:
Re-approving because of Bart's updates

PR #14256, opened 2025-08-14
refactor(core): move Protocol type to metadata subpackage

bartsmykla:
## Motivation

The `Protocol` type was previously defined in a package that brought in unnecessary dependencies when imported, even in cases where only the protocol definition was needed. This increased compile times and coupling between unrelated parts of the codebase.

## Implementation information

Refactored the codebase by moving the `Protocol` type into a new `metadata` subpackage. This makes the protocol definition lightweight and eliminates unnecessary dependencies when the type is imported. Updated all references across the codebase to use the new import path.

While making this change, performed minor related refactorings to ensure consistency and maintain clean imports.


pkg/core/metadata/protocol.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// GetCommonProtocol returns a common protocol between given two.
+//
+// E.g.,
+// a common protocol between HTTP and HTTP2 is HTTP2,
+// a common protocol between HTTP and HTTP  is HTTP,
+// a common protocol between HTTP and TCP   is TCP,
+// a common protocol between GRPC and HTTP2 is HTTP2,
+// a common protocol between HTTP and HTTP2 is HTTP.
```
slonka, line 98: this contradicts first sentence, we have:

```
a common protocol between HTTP and HTTP2 is HTTP2
a common protocol between HTTP and HTTP2 is HTTP
```

so which one is it?

slonka: and the unit tests say TCP: https://github.com/kumahq/kuma/pull/14256/files#diff-123b97ec69ad93b1afb8f055f92f18ecace699bcee2ce09cfbc2092760181221R74-R78

pkg/xds/envoy/listeners/filter_chain_configurers.go
```diff
@@ -139,13 +132,18 @@ func (s *splitAdapter) HasExternalService() bool { return s.hasExternalService }
...
 	for _, cluster := range clusters {
-		cluster := cluster.(*envoy_common.ClusterImpl)
-		splits = append(splits, &splitAdapter{
+		sa := &splitAdapter{
 			clusterName:        cluster.Name(),
-			weight:             cluster.Weight(),
 			lbMetadata:         cluster.Tags(),
 			hasExternalService: cluster.IsExternalService(),
-		})
+			weight:             1,
```
slonka, line 139: why do we set it to 1 by default? 🤔

pkg/xds/envoy/listeners/filter_chain_configurers.go
```diff
@@ -116,13 +116,6 @@ func NetworkRBAC(statsName string, rbacEnabled bool, permission *core_mesh.Traff
 	})
 }
 
-func DownstreamTlsContext(downstreamTlsContext *envoy_tls.DownstreamTlsContext) FilterChainBuilderOpt {
```
slonka, line 119: is this one replaced somewhere? if it was not used in Kuma did you check if it was used in the parent project?

bartsmykla: It was unused (in partent project as well)

pkg/core/metadata/protocol.go
bartsmykla: I copy pasted it, and didn't look into comment, but that's a good point and I'll followup with the fixed comment

pkg/xds/envoy/listeners/filter_chain_configurers.go
bartsmykla: because the new envoy_common.Cluster implementation does not set weights (pkg/plugins/policies/core/xds/cluster.go), we now default them to 1

PR #14262, opened 2025-08-15
refactor(xds): rework outbound route API, add SNI, helpers

bartsmykla:
## Motivation

This change improves the outbound route configuration API and cluster handling in XDS to make it safer, clearer, and more flexible. The previous `HttpOutboundRoute` signature was prone to accidental name mismatches and lacked explicit parameters for route and virtual host names. Adding SNI support to the `Cluster` interface improves TLS interoperability, and new helpers like `naming.GetNameOrFallback` reduce boilerplate for safe name resolution.

## Implementation information

- Introduced `naming.GetNameOrFallback` and `GetNameOrFallbackFunc` with tests to simplify name resolution when primary values are missing
- Relaxed generic constraint on contextual `sectionName` to `constraints.Unsigned` for broader applicability
- Added `SNI` to `Cluster` interface and cluster builder (`WithSNI`) for explicit TLS server name configuration
- Updated `HttpOutboundRoute` to accept `routeConfigName` and `virtualHostName` for better clarity
- Switched to `NewHttpStaticRouteConfigurer` for static HTTP routes to unify route-building logic
- Optimized builders (`FilterChainBuilder`, `ClusterBuilder`) to append directly and skip nil checks internally
- Made `UpdateHTTPConnectionManager` generic and able to accept functions with or without error returns
- Hardened `routes_configurer` with safe type assertions for timeout and weights
- Cleaned up unused imports and parameters

### DIFF
--- a/pkg/xds/envoy/virtualhosts/routes_configurer.go
+++ b/pkg/xds/envoy/virtualhosts/routes_configurer.go
@@ -170,8 +170,11 @@ func (c RoutesConfigurer) routeAction(clusters []envoy_common.Cluster, modify *m
 	if len(clusters) != 0 {
 		// Timeout can be configured only per outbound listener. So all clusters in the split
 		// must have the same timeout. That's why we can take the timeout from the first cluster.
-		cluster := clusters[0].(*envoy_common.ClusterImpl)
-		routeAction.Timeout = util_proto.Duration(cluster.Timeout().GetHttp().GetRequestTimeout().AsDuration())
+		if cluster, ok := clusters[0].(*envoy_common.ClusterImpl); ok {
+			routeAction.Timeout = util_proto.Duration(cluster.Timeout().GetHttp().GetRequestTimeout().AsDuration())
+		} else {
+			routeAction.Timeout = util_proto.Duration(0)
+		}
 	}
 	if len(clusters) == 1 {
 		routeAction.ClusterSpecifier = &envoy_config_route_v3.RouteAction_Cluster{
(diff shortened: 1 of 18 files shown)


pkg/xds/envoy/virtualhosts/routes_configurer.go
```diff
@@ -170,21 +170,29 @@ func (c RoutesConfigurer) routeAction(clusters []envoy_common.Cluster, modify *m
 	if len(clusters) != 0 {
 		// Timeout can be configured only per outbound listener. So all clusters in the split
 		// must have the same timeout. That's why we can take the timeout from the first cluster.
-		cluster := clusters[0].(*envoy_common.ClusterImpl)
-		routeAction.Timeout = util_proto.Duration(cluster.Timeout().GetHttp().GetRequestTimeout().AsDuration())
+		if cluster, ok := clusters[0].(*envoy_common.ClusterImpl); ok {
+			routeAction.Timeout = util_proto.Duration(cluster.Timeout().GetHttp().GetRequestTimeout().AsDuration())
+		} else {
+			routeAction.Timeout = util_proto.Duration(0)
+		}
```
Icarus9913, line 177: One question here, according to the comments of the `Timeout` field, "If not specified, the default is 15s".  I'd like to know if this change would break anything?

Icarus9913 (commented):
LGTM

pkg/xds/envoy/virtualhosts/routes_configurer.go
bartsmykla: It actually doesn't change anything, if it would we would see change in the golden files. This is just fallback :-)

Issue #14006, opened 2025-07-21
Add support for unified naming for traffic routed through ZoneEgress by data planes


bartsmykla:
Update the XDS generation logic in data plane proxies to use the unified naming format when routing outbound traffic through `ZoneEgress`, as defined in [MADR-070](https://github.com/kumahq/kuma/blob/df678629aa3bdce62a84c0e3752c7e5d45bf1e98/docs/madr/decisions/070-resource-identifier.md) and [MADR-077](https://github.com/kumahq/kuma/blob/52b8f5548739ecd1124181541f105bd6481b7ba5/docs/madr/decisions/077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md), enabled by the `feature-unified-resource-naming` flag.

#### Acceptance criteria

* Apply the unified naming format to outbound XDS resources and stats in data planes that route traffic through `ZoneEgress`
* Ensure correct naming for traffic targeting `MeshExternalService` and `MeshMultiZoneService`
* Preserve legacy naming when the feature flag is not set

PR #14280, opened 2025-08-18
feat(kuma-cp): support unified naming in zone proxies

bartsmykla:
## Motivation

We want to enable unified resource naming in zone proxies so that service references are consistent and unambiguous across meshes. This avoids collisions, simplifies debugging, and ensures that egress, ingress, and external services all follow the same naming conventions.

## Implementation information

This change introduces support for unified naming in zone proxies:
- `BackendRef` now uses `SectionName` when available, falling back to `Port`
- `MeshExternalService` derives its section name from port via `GetName`
- Outbounds for external services use `kri.WithSectionName`
- Cluster names are resolved from full `kri.Identifier` when unified naming is enabled
- `InternalServices` and `ExternalServices` generators are merged into a single egress `PolicyGenerator`
- Ingress and secrets generators are refactored to build `ResourceSet`s and produce clearer error messages
- Tests are updated to reflect the new generator wiring

We considered keeping separate generators for internal and external services, but merging them into a single `PolicyGenerator` makes the flow simpler and easier to maintain.

## Supporting documentation

Closes: https://github.com/kumahq/kuma/issues/14005
Closes: https://github.com/kumahq/kuma/issues/14006


pkg/xds/generator/egress/testdata/06.mixed-services-with-external-in-other-zone.golden.yaml
```diff
@@ -138,34 +138,34 @@ resources:
...
+        - service-2-zone-2{mesh=mesh-1}
         transportProtocol: tls
       filters:
       - name: envoy.filters.network.tcp_proxy
         typedConfig:
           '@type': type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
-          cluster: mesh-1:externalservice-2
-          statPrefix: mesh-1_externalservice-2
+          cluster: mesh-1:service-2-zone-2
+          statPrefix: mesh-1_service-2-zone-2
```
bartsmykla, line 168: `filterChains` are now ordered by their cluster names, that's why it changed

pkg/xds/generator/egress/testdata/04.mixed-services-custom-trafficroute.golden.yaml
```diff
@@ -1,35 +1,4 @@
...
-              envoy.transport_socket_match:
-                kuma.io/protocol: http
-                mesh: mesh-1
-    name: mesh-1:externalservice-1
-    type: STRICT_DNS
-    typedExtensionProtocolOptions:
-      envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
-        '@type': type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
-        explicitHttpConfig:
-          httpProtocolOptions: {}
```
bartsmykla, line 32: Previously we did not check if any endpoint existed before creating clusters for legacy external services. This removal is done because we now perform that check

api/common/v1alpha1/targetref.go
```diff
@@ -164,7 +164,7 @@ type BackendRef struct {
 func (b BackendRef) ReferencesRealObject() bool {
 	switch b.Kind {
 	case MeshService:
-		return b.Port != nil
+		return pointer.Deref(b.SectionName) != "" || b.Port != nil
```
slonka, line 167: is this a fix for the bug that you discovered a while back?

pkg/xds/generator/egress/testdata/04.mixed-services-custom-trafficroute.golden.yaml
slonka: Can this be a breaking change?

api/common/v1alpha1/targetref.go
bartsmykla: that's correct

pkg/xds/generator/egress/testdata/04.mixed-services-custom-trafficroute.golden.yaml
slonka: Talked offline: nothing routed to them so it's safe

api/common/v1alpha1/targetref.go
slonka: Talked offline: yes, but it's a small fix so we did not want to create a separate issue for that

slonka (approved):

pkg/xds/generator/egress/testdata/04.mixed-services-custom-trafficroute.golden.yaml
bartsmykla: It can't, because we were creating clusters, but nothing was routing to them

PR #14287, opened 2025-08-19
fix(helm): remove `sh` usage in CRD install helm hook job

bartsmykla:
## Motivation

Bitnami announced they will remove their `bitnami` registry soon. We currently rely on images from this registry that include a shell. When switching to other `kubectl` images (such as rancher) which do not ship with `sh`, the Helm hook Job responsible for installing CRDs breaks. We need to remove the shell dependency before the registry is removed.

## Implementation information

The Helm `pre-install,pre-upgrade` hook Job is updated so that the main container no longer executes a script via `sh`. Instead, it directly invokes `kubectl apply` on the CRD files generated by the init container. No new containers are introduced and the init container is unchanged. Since CRDs are generated with `kumactl install crds` without the `--only-missing` flag, the generated files always contain CRDs, so `kubectl apply` has resources to apply.

## Supporting documentation

- Related issue: https://github.com/bitnami/charts/issues/35164

### DIFF
--- a/deployments/charts/kuma/templates/pre-upgrade-install-crds-job.yaml
+++ b/deployments/charts/kuma/templates/pre-upgrade-install-crds-job.yaml
@@ -72,16 +72,6 @@ metadata:
   labels:
   {{- include "kuma.labels" . | nindent 4 }}
 data:
-  install_crds.sh: |
-    #!/usr/bin/env sh
-    set -e
-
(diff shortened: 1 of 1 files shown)


jijiechen (approved):

slonka (approved):
Seems like this functionality of `--only-missing` is not used so this PR looks safe 👍

PR #14303, opened 2025-08-20
feat(kuma-cp/xds): support unified naming in MeshTLS

bartsmykla:
## Motivation

Make `MeshTLS` work with unified resource naming so clusters, routes, listeners, SNI, and SAN matchers use new naming formats when the feature is enabled.

## Implementation information

- Resolve names from full `kri.Identifier` including `SectionName`
- Read `SectionName` from `BackendRef` when present, otherwise derive it from the port
- Apply section-aware names to SNI, SAN matchers, cluster names, and related MeshTLS wiring
- Keep legacy behavior when unified naming is off to maintain backward compatibility
- Refactor helpers and split functions to improve readability without changing behavior outside of naming
- Update and align tests with the new naming flow

## Supporting documentation

Closes: https://github.com/kumahq/kuma/issues/13961

### DIFF
--- a/pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin.go
+++ b/pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin.go
@@ -12,7 +12,9 @@ import (
 	common_tls "github.com/kumahq/kuma/api/common/v1alpha1/tls"
 	mesh_proto "github.com/kumahq/kuma/api/mesh/v1alpha1"
 	"github.com/kumahq/kuma/pkg/core"
+	"github.com/kumahq/kuma/pkg/core/kri"
 	core_meta "github.com/kumahq/kuma/pkg/core/metadata"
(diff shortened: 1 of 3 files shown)


pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin.go
```diff
@@ -36,33 +39,43 @@ import (
...
+	log := logger.WithValues(
+		"proxyName", proxy.Dataplane.GetMeta().GetName(),
+		"mesh", ctx.Mesh.Resource.GetMeta().GetName(),
+	)
+
+	switch {
+	case ctx.Mesh.Resource.MTLSEnabled():
+	case proxy.WorkloadIdentity != nil:
+	default:
+		log.V(1).Info("skip applying MeshTLS, mTLS is disabled")
```
slonka, line 74: It took me a while but I think these two are equivalent so 👍

slonka (approved):

Issue #13799, opened 2025-06-25
[user resources] Add e2e tests for consistent naming feature with full and partial feature flag rollout


bartsmykla:
### Description

Implement e2e tests to validate correct generation of unified naming formats for Envoy resources and stats across all deployment scenarios described in the [MADR-077 Defining and migrating to consistent naming for non-system Envoy resources and stats](https://github.com/kumahq/kuma/blob/dc2dd5564cac3f97f85085a8aa498f4a520ad4ba/docs/madr/decisions/077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md). This includes testing KRI-based names for resources tied to distinct Kuma resources, contextual format for proxy-scoped resources like non-system inbounds and transparent proxy passthrough, and system format for internal components.

PR #14352, opened 2025-08-26
fix(kuma-cp): fix match MeshHTTPRoute in policy targets with unified naming

bartsmykla:
## Motivation

With unified naming enabled, generated Envoy route names for a `MeshHTTPRoute` include the full KRI with a `SectionName` (the rule index that produced the route). Policies like `MeshTimeout` and `MeshRetry` target the whole `MeshHTTPRoute`, so their `ToRules` carry the base KRI without a `SectionName`. This mismatch caused the policy wiring to skip those routes because the names did not match. We found this while enabling unified naming for one zone in our multizone e2e tests.

## Implementation information

- Make route lookups section-aware: when a policy targets a `MeshHTTPRoute` by its base KRI, match routes derived from that resource even if their names include a `SectionName`
- Keep behavior unchanged when unified naming is off
- Add e2e coverage by enabling unified naming in one zone of the multizone suite
  - Rationale: we do not have a good unit or integration test seam that exercises the full path from policy targets to Envoy route names with sectioned KRIs
  - Enabling unified naming in one zone validates the fix end-to-end with minimal churn, while leaving other zones on legacy naming for stability

## Supporting documentation

Related to: https://github.com/kumahq/kuma/issues/13799

### DIFF
--- a/pkg/core/kri/kri.go
+++ b/pkg/core/kri/kri.go
@@ -62,7 +62,7 @@ func FromResourceMeta(rm core_model.ResourceMeta, resourceType core_model.Resour
 
 func FromString(s string) (Identifier, error) {
 	parts := strings.Split(s, "_")
-	if len(parts) != 7 {
+	if len(parts) < 7 {
 		return Identifier{}, errors.Errorf("invalid identifier string: %q", s)
 	}
 	if parts[0] != "kri" {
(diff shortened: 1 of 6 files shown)


pkg/core/kri/kri.go
```diff
@@ -62,7 +62,7 @@ func FromResourceMeta(rm core_model.ResourceMeta, resourceType core_model.Resour
 
 func FromString(s string) (Identifier, error) {
 	parts := strings.Split(s, "_")
-	if len(parts) != 7 {
+	if len(parts) < 7 {
```
slonka, line 65: this is because section name can have `_`

Issue #14322, opened 2025-08-22
Dataplane inbounds are duplicated when multiple services select the same port


lobkovilya:
### Description

It's exactly as https://github.com/kumahq/kuma/issues/12123 but it should've been fixed in https://github.com/kumahq/kuma/pull/12760.

It also causes `/_layout` and `/_policies` endpoints to return duplicated data:

```json
{
 "inbounds": [
  {
   "kri": "kri_dp_default_zone-1_kuma-demo_demo-app-5d497d665f-88k49_http",
   "port": 5050,
   "protocol": "http",
   "proxyResourceName": "self_inbound_dp_http"
  },
  {
   "kri": "kri_dp_default_zone-1_kuma-demo_demo-app-5d497d665f-88k49_http",
   "port": 5050,
   "protocol": "http",
   "proxyResourceName": "self_inbound_dp_http"
  }
 ]
 ...
}
```
```json
{
 "policies": [
  {
   "kind": "MeshTimeout",
   "origins": [
    {
     "kri": "kri_mt_default__kong-mesh-system_mesh-timeout-all-default_"
    },
    {
     "kri": "kri_mt_default__kong-mesh-system_mesh-timeout-all-default_"
    }
   ],
   "rules": [
    {
     "conf": {
      "connectionTimeout": "10s",
      "idleTimeout": "2h0m0s",
      "http": {
       "requestTimeout": "0s",
       "streamIdleTimeout": "1h0m0s",
       "maxStreamDuration": "0s"
      }
     },
     "kri": "kri_mt_default__kong-mesh-system_mesh-timeout-all-default_"
    },
    {
     "conf": {
      "connectionTimeout": "10s",
      "idleTimeout": "2h0m0s",
      "http": {
       "requestTimeout": "0s",
       "streamIdleTimeout": "1h0m0s",
       "maxStreamDuration": "0s"
      }
     },
     "kri": "kri_mt_default__kong-mesh-system_mesh-timeout-all-default_"
    }
   ]
  }
 ]
}

```

Steps to reproduce: 

1. Apply https://github.com/kumahq/kuma-counter-demo
2. See the Dataplane has 2 inbounds
```yaml
apiVersion: kuma.io/v1alpha1
kind: Dataplane
mesh: default
spec:
  networking:
    address: 10.64.7.5
    admin:
      port: 9901
    inbound:
    - health:
        ready: true
      name: http
      port: 5050
      tags:
        app: demo-app
[...]

PR #14356, opened 2025-08-27
fix(kuma-cp): reconcile pods if MeshService mode changed to remove duplicated inbounds

Automaat:
## Motivation

When we use MS mode exclusive, we need to deduplicate inbounds on Dataplane. This works well if you start with this mode. This PR covers the situation when you start with MS mode other than exclusive and migrate to exclusive, or if you go back to old services from MS exclusive

Fix https://github.com/kumahq/kuma/issues/14322

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/pod_controller.go
+++ b/pkg/plugins/runtime/k8s/controllers/pod_controller.go
@@ -338,7 +338,6 @@ func (r *PodReconciler) createOrUpdateDataplane(
 		},
 	}
 	operationResult, err := kube_controllerutil.CreateOrUpdate(ctx, r.Client, dataplane, func() error {
-		r.Log.Info("UPDATING DATAPLANE", "dataplane", dataplane)
 		if err := r.PodConverter.PodToDataplane(ctx, dataplane, pod, services, others, mesh); err != nil {
 			return errors.Wrap(err, "unable to translate a Pod into a Dataplane")
 		}
--- a/pkg/plugins/runtime/k8s/controllers/pod_controller.go
+++ b/pkg/plugins/runtime/k8s/controllers/pod_controller.go
@@ -338,6 +338,7 @@ func (r *PodReconciler) createOrUpdateDataplane(
 		},
 	}
 	operationResult, err := kube_controllerutil.CreateOrUpdate(ctx, r.Client, dataplane, func() error {
+		r.Log.Info("UPDATING DATAPLANE", "dataplane", dataplane)
 		if err := r.PodConverter.PodToDataplane(ctx, dataplane, pod, services, others, mesh); err != nil {
 			return errors.Wrap(err, "unable to translate a Pod into a Dataplane")
 		}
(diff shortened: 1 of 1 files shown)


pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -541,3 +546,63 @@ func EndpointSliceToPodsMapper(l logr.Logger, client kube_client.Client) kube_ha
...
+		for _, pod := range pods.Items {
+			req = append(req, kube_reconcile.Request{
+				NamespacedName: kube_types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name},
+			})
+		}
+		return req
+	}
+}
+
+type MeshServiceExclusivePredicate struct {
```
lobkovilya, line 573: have you checked if changing something else on the mesh (like mtls) doesn't reconcile pods?

Automaat: Yes, I checked it locally

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -541,3 +546,63 @@ func EndpointSliceToPodsMapper(l logr.Logger, client kube_client.Client) kube_ha
...
 }
+
+func MeshToPodsMapper(l logr.Logger, client kube_client.Client) kube_handler.MapFunc {
+	l = l.WithName("mesh-to-pods-mapper")
+	return func(ctx context.Context, obj kube_client.Object) []kube_reconcile.Request {
+		mesh := obj.(*mesh_k8s.Mesh)
+
+		l := l.WithValues(
+			"Mesh",
+			kube_types.NamespacedName{Namespace: mesh.Namespace, Name: mesh.Name},
```
lobkovilya: what's `mesh.Namespace`? mesh resource is globally scoped

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -541,3 +546,63 @@ func EndpointSliceToPodsMapper(l logr.Logger, client kube_client.Client) kube_ha
...
+		)
+
+		pods := &kube_core.PodList{}
+		if err := client.List(ctx, pods, kube_client.InNamespace(obj.GetNamespace()), kube_client.MatchingLabels(map[string]string{
```
lobkovilya: same, not sure if we need to add `kube_client.InNamespace`

pkg/plugins/runtime/k8s/controllers/pod_controller.go
Automaat: right, removed

pkg/plugins/runtime/k8s/controllers/pod_controller.go
Automaat: good catch, it was from copy pasting. Updated

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -334,6 +338,7 @@ func (r *PodReconciler) createOrUpdateDataplane(
 		},
 	}
 	operationResult, err := kube_controllerutil.CreateOrUpdate(ctx, r.Client, dataplane, func() error {
+		r.Log.Info("UPDATING DATAPLANE", "dataplane", dataplane)
```
lobkovilya: judging by the upper case it's probably a leftover?

Automaat: it was debug log, removed

lobkovilya (approved):

PR #14361, opened 2025-08-28
test(framework): migrate to `portforward` API and add `ConfigDump` support

bartsmykla:
## Motivation

The old `tunnel.NewK8sEnvoyAdminTunnel` API was limited and required manual port-forward handling. We are moving to the new `portforward` API which provides a simpler and unified way of creating and reusing admin tunnels. Additionally, we add `GetConfigDump` support to improve envoy config validation in tests.

## Implementation information

- replaced all usages of `tunnel.NewK8sEnvoyAdminTunnel` with `Cluster.GetOrCreateAdminTunnel` using `portforward.Spec`
- removed old `tunnel` import and added new `portforward` import
- updated e2e tests in `spire` and multizone `meshservice` to use the new API
- extended the admin tunnel with a `GetConfigDump` function for easier envoy configuration checks

### DIFF
--- a/test/framework/envoy_admin/config_dump/config_dump.go
+++ b/test/framework/envoy_admin/config_dump/config_dump.go
@@ -1,0 +1,71 @@
+package config_dump
+
+import (
+	envoy_admin_v3 "github.com/envoyproxy/go-control-plane/envoy/admin/v3"
+
+	xds_types "github.com/kumahq/kuma/pkg/core/xds/types"
+	"github.com/kumahq/kuma/pkg/util/proto"
+)
+
+type EnvoyConfig struct {
+	Boostrap     envoy_admin_v3.BootstrapConfigDump
+	Cluster      envoy_admin_v3.ClustersConfigDump
+	Endpoints    envoy_admin_v3.EndpointsConfigDump
+	Listeners    envoy_admin_v3.ListenersConfigDump
+	ScopedRoutes envoy_admin_v3.ScopedRoutesConfigDump
+	Routes       envoy_admin_v3.RoutesConfigDump
+}
+
+func ParseEnvoyConfig(bs []byte) (*EnvoyConfig, error) {
+	var cd envoy_admin_v3.ConfigDump
+	if err := proto.FromJSON(bs, &cd); err != nil {
+		return nil, err
+	}
+
+	var config EnvoyConfig
+
+	for _, raw := range cd.GetConfigs() {
+		var err error
+
+		switch {
+		case raw.MessageIs(&config.Boostrap):
+			err = raw.UnmarshalTo(&config.Boostrap)
+		case raw.MessageIs(&config.Cluster):
+			err = raw.UnmarshalTo(&config.Cluster)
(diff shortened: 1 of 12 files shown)


test/framework/envoy_admin/config_dump/config_dump.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+type EnvoyConfig struct {
+	Boostrap     envoy_admin_v3.BootstrapConfigDump
+	Cluster      envoy_admin_v3.ClustersConfigDump
+	Endpoints    envoy_admin_v3.EndpointsConfigDump
+	Listeners    envoy_admin_v3.ListenersConfigDump
+	ScopedRoutes envoy_admin_v3.ScopedRoutesConfigDump
+	Routes       envoy_admin_v3.RoutesConfigDump
+}
```
slonka, line 17: aren't we missing `Secrets` and `ECDS`?

test/framework/envoy_admin/config_dump/config_dump.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (m *Metadata) HasFeature(feature string) bool {
+	return m.features[feature]
+}
+
+func (c *EnvoyConfig) Metadata() *Metadata {
+	meta := Metadata{features: xds_types.Features{}}
+
+	if raw, ok := c.Boostrap.GetBootstrap().GetNode().GetMetadata().Fields["features"]; ok {
```
slonka, line 64: is it possible that any of these is nil?

slonka (approved):

test/framework/envoy_admin/config_dump/config_dump.go
bartsmykla: it's not

test/framework/envoy_admin/config_dump/config_dump.go
bartsmykla: I will do it as a followup

test/framework/k8s_cluster.go
```diff
@@ -169,34 +165,46 @@ func (c *K8sCluster) PortForward(
...
-	return c.portForwards[name]
+func (c *K8sCluster) GetPortForward(spec portforward.Spec) portforward.Tunnel {
+	return c.portForwards[spec]
 }
 
-func (c *K8sCluster) ClosePortForward(name string) {
-	if fwd, ok := c.portForwards[name]; ok {
-		fwd.Close()
-		delete(c.portForwards, name)
+func (c *K8sCluster) ClosePortForwards(specs ...portforward.Spec) {
```
lobkovilya, line 183: TBH I think easier way to implement this would be to simply pass `filters` funcs:
```go
func (c *K8sCluster) ClosePortForwards(filters ...func(portforward.Spec) bool) {
```
I think it's more familiar looking API and you don't need to create `Matches` relation between specs

bartsmykla: I'll improve it as a followup

PR #14386, opened 2025-09-04
feat(api): merge inbound policies in inspect api

Automaat:
## Motivation

We should merge inbound policies other than MeshTrafficPermission in inspect api, to make it simpler to understand for users and more concise to display in gui.


lahabana:
@Automaat why is this targeting just 2.12? shouldn't it go to master and be backported?

lahabana:
Let's figure if this needs to be closed and finished.

Issue #14340, opened 2025-08-25
Provide migration path from `mesh.mTLS` to `MeshIdentity`


slonka:
### Description

When both `mesh.mTLS` section and `MeshIdentity` is present, TLS context on the cluster contains certs from both of them but on the inbound listener it only contains data from `MeshIdentity`. If we wanted to migrate from `mesh.mTLS` to MeshIdentity without interruption we need to make the inbound listener contain both.

PR #14387, opened 2025-09-04
fix(meshtrust): add trust independently of WorkloadIdentity

lukidzi:
## Motivation

We introduced MeshTrust, which allows users to provide a CA that the mesh should use to accept traffic. Unfortunately, it was originally tightly coupled to the existence of a WorkloadIdentity generated by the MeshIdentity. This shouldn’t be the case, since MeshTrust should also work without a MeshIdentity - for example, when operating with older Mesh resources.

## Implementation information

*If Mesh.mTLS is enabled and MeshTrust is defined, we add the mesh mTLS CA to the set of Trusts.
* A CAProvider is used in MeshContext to retrieve the mesh CA.
* By default, use the `system_trust_bundle` secret for the validation context (when WorkloadIdentity is provided). Otherwise, use a configurer.
* Sorting of Trusts was added in `dataplane_status_sink` to avoid unnecessary updates.

## Supporting documentation

Part of https://github.com/kumahq/kuma/issues/14340


test/e2e_env/multizone/meshidentity/migration.go
```diff
@@ -0,0 +1,249 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return r.GetSpec().(*meshtrust_api.MeshTrust), nil
+	}
+
+	indent := func(pem string) string {
+		pad := strings.Repeat(" ", 10)
+		return pad + strings.ReplaceAll(pem, "\n", "\n"+pad)
+	}
```
slonka: nitpick: could be moved to test utils, we have one like that already https://github.com/kumahq/kuma/blob/98ceeba150132bbc9f795118758aa0e14b2330f4/test/e2e_env/multizone/meshidentity/meshidentity.go#L90

test/e2e_env/multizone/meshidentity/migration.go
```diff
@@ -0,0 +1,249 @@
...
+			var err error
+			trust2, err = getMeshTrust(multizone.KubeZone2.Name())
+			g.Expect(err).ToNot(HaveOccurred())
+		}, "30s", "1s").Should(Succeed())
+		Expect(NewClusterSetup().
+			Install(YamlK8s(fmt.Sprintf(trustTmpl, multizone.KubeZone2.Name(), meshName, multizone.KubeZone1.Name(), indent(trust2.CABundles[0].PEM.Value), trust2.TrustDomain))).
+			Setup(multizone.KubeZone1)).To(Succeed())
+
+		// and
+		// select dataplane
```
slonka:
```suggestion
		// select all dataplanes
```

right?

test/e2e_env/multizone/meshidentity/migration.go
```diff
@@ -0,0 +1,249 @@
...
+				builders.Mesh().
+					WithName(meshName).
+					WithMeshServicesEnabled(mesh_proto.Mesh_MeshServices_Exclusive),
+			)).
+			Setup(multizone.Global)).To(Succeed())
+
+		// then
+		// Traffic may sometimes fail when switching to the new identity.
+		// This happens because changes to MeshService are propagated
+		// with the delay while some applications might already use a new identity.
```
slonka, line 224: I don't understand this. So the client change is propagated faster than the server change? Isn't this the whole point of smooth migration?

lobkovilya: I think the problem is explained here https://github.com/kumahq/kuma/pull/14503/files#diff-6b19182e323fc0e4cf026f63532816ba839fec0f303fbc30f834d20ecbdc6f35R7

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
```diff
@@ -269,13 +268,12 @@ func (b *bundledIdentityProvider) CreateIdentity(ctx context.Context, identity *
 	b.logger.V(1).Info("identity created", "dpp", model.MetaToResourceKey(proxy.Dataplane.GetMeta()), "spiffeId", spiffeID, "identity", model.MetaToResourceKey(identity.GetMeta()))
 
 	return &xds.WorkloadIdentity{
-		KRI:                        identifier,
-		ManagementMode:             xds.KumaManagementMode,
-		ExpirationTime:             pointer.To(template.NotAfter),
-		GenerationTime:             pointer.To(now),
-		IdentitySourceConfigurer:   sourceConfigurer(identifier.String()),
-		ValidationSourceConfigurer: sourceConfigurer(system_names.SystemResourceNameCABundle),
```
lobkovilya, line 277: Should this field `ValidationSourceConfigurer` even exist in `WorkloadIdentity` struct? It's not identity's concern how it'll be verified no?

pkg/core/resources/apis/meshtrust/generator/v1alpha1/secrets.go
```diff
@@ -26,18 +26,17 @@ func NewPlugin() core_plugins.CoreResourcePlugin {
 }
 
 func (p *plugin) Generate(rs *core_xds.ResourceSet, xdsCtx xds_context.Context, proxy *core_xds.Proxy) error {
-	kumaManaged := pointer.Deref(proxy.WorkloadIdentity).ManagementMode == core_xds.KumaManagementMode
+	externallyManaged := pointer.Deref(proxy.WorkloadIdentity).ManagementMode == core_xds.ExternalManagementMode
```
lobkovilya, line 33: Why do we even check `proxy.WorkloadIdentity` in the `MeshTrust` plugin? IMO this plugin should function purely based on MeshTrust resources without knowing about MeshIdentities existence

pkg/xds/context/context.go
```diff
@@ -91,7 +90,7 @@ type MeshContext struct {
 	ServicesInformation         map[string]*ServiceInformation
 	DataSourceLoader            datasource.Loader
 	ReachableServicesGraph      ReachableServicesGraph
-	TrustsByTrustDomain         map[string][]*meshtrust_api.MeshTrust
+	TrustsByTrustDomain         map[string][]string
```
lobkovilya: Now I need to read the usage of `TrustsByTrustDomain` to understand what `[]string` is storing. Maybe it'd make sense to create type alias like 
```go
type PEMBytes []byte
```
and then rename the field like
```go
CAsByTrustDomain map[string][]PEMBytes
```

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
lukidzi: There is a problem when using SPIRE. In case of spire we don't want to use a MeshTrust (as we want spire to deliver the CA), so in this case is connected to it. Also when using SPIRE we don't want to create a secrets from MeshTrust as they won't be used.

pkg/core/resources/apis/meshtrust/generator/v1alpha1/secrets.go
lukidzi: The problem is with the SPIRE, since spire doesn't need these certificates since it delivers their own by SDS. In this case it makes it a bit complicated since when using SPIRE, MeshTrust doesn't have an effect on listener.

slonka (approved):

PR #14494, opened 2025-09-16
chore(k8s): upgrade min and max k8s version

lukidzi:
## Motivation

We were using old versions of the K8s

## Implementation information

* Update minimal and maximum version of k8s
* We should update soon to 1.34 since it's released by k3d image is not yet available (It's a RC)

### DIFF
--- a/mk/dev.mk
+++ b/mk/dev.mk
@@ -27,7 +27,7 @@ CI_TOOLS_BIN_DIR=$(CI_TOOLS_DIR)/bin
 # Change here and `make check` ensures these are used for CI
 # Note: These are _docker image tags_
 # If changing min version, update mk/kind.mk as well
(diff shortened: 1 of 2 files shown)


mk/dev.mk
```diff
@@ -27,10 +27,10 @@ CI_TOOLS_BIN_DIR=$(CI_TOOLS_DIR)/bin
 # Change here and `make check` ensures these are used for CI
 # Note: These are _docker image tags_
 # If changing min version, update mk/kind.mk as well
-K8S_MIN_VERSION = v1.27.16-k3s1
-K8S_MAX_VERSION=v1.32.2-k3s1
```
lukidzi, line 31: there is 1.34.1 but the image is a RC so waiting for the release

bartsmykla (approved):

Issue #14388, opened 2025-09-04
Provide a way to migrate from one spiffeId to another


lukidzi:
### Description

While working on the migration from `mesh.mTLS` to `MeshIdentity`, I discovered an issue with transitioning from one SPIFFE ID to another.

Let’s assume we have two zones: `zone-client` and `zone-server`. A client runs in zone-client, and a server runs in zone-server.

1. Both zone-client and zone-server use `mesh.mTLS`, and the MeshService identity is based on `kuma.io/service`.
2. Both zones switch to MeshIdentity.
3. The client in zone-client can no longer communicate with the server in zone-server.
4. The MeshService of the server in zone-server is updated with a SPIFFE ID.
5. That MeshService is synced to zone-client.
6. The client’s configuration is updated with a SAN based on the new SPIFFE ID.
7. Traffic resumes and works correctly.

This illustrates a case where both zones begin using the new identity, but the server presents itself as `spiffe://domain.zone-server.mesh.local/ns/server/sa/server` while the client cannot yet verify this SAN, since the corresponding MeshService update has not propagated.

lahabana:
Such a design will need a MADR

PR #14503, opened 2025-09-17
docs(MADR): migration from one spiffeID to another using MeshService

lukidzi:
## Motivation

We want to migrate from one SpiffeID to another without a downtime. Currently the problem was that MeshService was missing a new identifier.

## Implementation information

Provider a design proposal how we can solve it.

## Supporting documentation

xref: https://github.com/kumahq/kuma/issues/14388

### DIFF
--- a/docs/madr/decisions/083-meshservice-meshidentity-identity.md
+++ b/docs/madr/decisions/083-meshservice-meshidentity-identity.md
@@ -1,0 +1,298 @@
+# Migration from one SpiffeID to another without a downtime using MeshService
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/14388
+
+## Context and Problem Statement
+
+MeshService stores identities in the object:
+
+```yaml
+spec:
+  identities:
+  - type: ServiceTag
+    value: test-server_default_kuma-demo_svc_80
+  - type: SpiffeID
+    value: spiffe://default.zone.mesh.local/ns/kuma-demo/sa/default
+```
+`ServiceTag` identity is a value that can be precomputed and set on a `MeshService` without checking whether mTLS is enabled.
+In the case of `MeshIdentity`, we set the `SpiffeID` based on the matching resource. We determine which `MeshIdentity` targets a specific `MeshService`, and based on that, we assign the corresponding identity.
+
+Later, this information is used to create the Envoy cluster configuration. Once a `MeshIdentity` is enabled and targets specific dataplanes, Kuma begins delivering identities to each sidecar. At this point, the `MeshService` is recalculated, new identities are added, and these are then translated into SAN matching.
+
+### Where is the issue?
+
+Let’s walk through the flow:
+
+Prerequisites:
+* `MeshService` is in `Exclusive` mode
+* There is constant traffic from the client to the service
+
+1. Mesh with legacy mTLS (`Mesh.mTLS`) is enabled
+2. Client and Server details
+* Client presents identity: `client_default_kuma-demo_svc_80`
+* Server presents identity: `server_default_kuma-demo_svc_80`
+
+MeshService of Client
+```yaml
+spec:
+  identities:
+  - type: ServiceTag
+    value: client_default_kuma-demo_svc_80
+```
+
+MeshService of Server
+```yaml
+spec:
+  identities:
+  - type: ServiceTag
+    value: server_default_kuma-demo_svc_80
+```
+3. Envoy configuration based on MeshService identities
+From the `MeshService` definitions, Kuma generates Envoy configuration that enforces identity checks.
+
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/083-meshservice-meshidentity-identity.md
```diff
@@ -0,0 +1,298 @@
...
+Prerequisites:
+* `MeshService` is in `Exclusive` mode
+* There is constant traffic from the client to the service
+
+1. Mesh with legacy mTLS (`Mesh.mTLS`) is enabled
+2. Client and Server details
+* Client presents identity: `client_default_kuma-demo_svc_80`
+* Server presents identity: `server_default_kuma-demo_svc_80`
+
+MeshService of Client
```
lobkovilya, line 37: Does MeshService of Client play any role in this communication? I don't think it's relevant if `Client` also acts as server for someone else

docs/madr/decisions/083-meshservice-meshidentity-identity.md
```diff
@@ -0,0 +1,298 @@
...
+- Introduce a new field `mode`
+- Use a `kuma.io/effect` label
+
+## Decision Outcome
+
+- Introduce a new field `mode`
+
+## Design
+
+### Introduce a new field `mode`
```
lobkovilya, line 183: What if instead of new field `mode` we just allow creating MeshIdentity without the `provider`:
```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshIdentity
metadata:
  name: identity
  namespace: kuma-system
  labels:
    kuma.io/mesh: default
spec:
  selector:
    dataplane:
      matchLabels: {}
  spiffeID:
    trustDomain: "{{ .Mesh }}.{{ .Zone }}.mesh.local"
    path: "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}"
```
In this case spiffeID is propagated to MeshService, but no actual crypto identity is propagated to the DPP (because there is no provider).

docs/madr/decisions/083-meshservice-meshidentity-identity.md
lukidzi: You are right, just wanted to show the example with update of both. If you feel like it isn't required I can remove it

docs/madr/decisions/083-meshservice-meshidentity-identity.md
lukidzi: I like this idea. I've updated the MADR

docs/madr/decisions/083-meshservice-meshidentity-identity.md
```diff
@@ -0,0 +1,361 @@
...
+    dataplane:
+      matchLabels: {}
+  spiffeID:
+    trustDomain: "{{ .Mesh }}.{{ .Zone }}.mesh.local"
+    path: "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}"
+```
+
+In this case, we would not create any certificates or issue new identities. The only change would be that the `MeshService` is updated with a new `SpiffeID`.
+
+### Status field
```
lobkovilya: It's a subsection of `### Generate identity only when there is no provider section` right?
```suggestion
#### Status field
```

docs/madr/decisions/083-meshservice-meshidentity-identity.md
```diff
@@ -0,0 +1,361 @@
...
+### Introduce a new field `mode`
+
+We could introduce a new field into MeshIdentity resource:
+
+```yaml
+type: MeshIdentity
+name: identity
+mesh: default
+spec:
+  mode: Active (Default) | SAN
```
slonka, line 192: note to myself: SAN mode is aliasing in MeshService

docs/madr/decisions/083-meshservice-meshidentity-identity.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Status field
+
+When a user creates a `MeshIdentity`, a controller sets its status to indicate whether the resource is ready and initializes all related resources.
+In this case, since there is no provider, we don’t have an initialization phase. However, we still want to set a status on the resource for better debuggability.
+
+Proposed status:
+
+```golang
+        common_api.Condition{
```
lobkovilya, line 249: please fix indentation

docs/madr/decisions/083-meshservice-meshidentity-identity.md
```diff
@@ -0,0 +1,361 @@
...
+					Type:    meshidentity_api.SANProviderConditionType,
```
lukidzi: Type: `SpiffeIDProvider`
Reason: `SpiffeIDProvided`

PR #14518, opened 2025-09-21
fix(kds): fix kds deadlock with grpc streaming

### LINKED ISSUE #14486: KDS deadlock
### Kuma Version

2.9.4

### Describe the bug

In the multizone deployment, the policy that is created in Zone could be synced to Global. However, the policy created in Global can not be synced to Zone.

### To Reproduce

If you create a massive amount of policies in Global, the problem may happen.

### Expected behavior

The policies could be synced from each side.

## version
kuma version: `2.9.4`
go-control-plane version: `v0.13.1-kong-1`
grpc-go version: `v1.67.1` 

### zone CP pprof

Image

### global CP pprof

Image

### LINKED ISSUE #14517: potential kds goroutine leak
### Kuma Version

master

### Describe the bug

When I went through the kds codes, I found that there would be a case to lead goroutine leak.

For example, 
on ZoneCP we start startGlobalToZoneSync in a goroutine
https://github.com/kumahq/kuma/blob/2.12.0/pkg/kds/mux/client.go#L115,

later, we'll call `OnGlobalToZoneSyncStarted` callback function and it's running in a goroutine.
https://github.com/kumahq/kuma/blob/2.12.0/pkg/kds/mux/client.go#L138

Also,  the `handleProcessingErrors` would be stuck waiting for the `processingErrorsCh` channel
https://github.com/kumahq/kuma/blob/2.12.0/pkg/kds/mux/client.go#L257-L263

And, if you go back to check the `OnGlobalToZoneSyncStarted` callback function
https://github.com/kumahq/kuma/blob/2.12.0/pkg/kds/zone/components.go#L74-L85

There might be a case that once if we hit the error `context.Canceled`, we won't send error to the `errChan` channel. In such a situation, the above `handleProcessingErrors` function would be stuck indefinitely

### To Reproduce

_No response_

### Expected behavior

no goroutine leak

_No response_

Icarus9913:
## Motivation

Decoupling the kds grpc client `Send` and `Recv` methods in different goroutines to avoid deadlock.

## Supporting documentation

Fix #14486
Fix #14517


pkg/kds/mux/client.go
```diff
@@ -105,10 +105,9 @@ func (c *client) Start(stop <-chan struct{}) (errs error) {
 	defer cancel()
 
 	log := muxClientLog.WithValues("client-id", c.clientID)
-	errorCh := make(chan error)
+	errorCh := make(chan error, 6)
```
Icarus9913: Setting the channel capability to 6 to avoid the goroutine leak due to the following 6 goroutines methods `errorCh` sending stuck

pkg/kds/mux/client.go
```diff
@@ -261,22 +260,22 @@ func (c *client) handleProcessingErrors(
...
-	}
-	if err := stream.CloseSend(); err != nil {
-		log.Error(err, "CloseSend returned an error")
-	}
 	if err != nil {
+		if status.Code(err) == codes.Unimplemented {
+			log.Error(err, "rpc stream failed, because global CP does not implement this rpc. Upgrade remote CP.")
+			// backwards compatibility. Do not rethrow error, so KDS multiplex can still operate.
+			return
+		}
```
Icarus9913: We should clean up these codes in master branch later.

pkg/kds/mux/zone_sync.go
```diff
@@ -143,12 +140,8 @@ func (g *KDSSyncServiceServer) ZoneToGlobalSync(stream mesh_proto.KDSSyncService
 	shouldDisconnectStream := g.watchZoneHealthCheck(stream.Context(), zone, service.ZoneToGlobal, connectTime)
 	defer shouldDisconnectStream.Close()
 
-	processingErrorsCh := make(chan error, 1)
-	go func() {
-		if err := g.zoneToGlobalCb.OnZoneToGlobalSyncConnect(stream); err != nil {
-			processingErrorsCh <- err
-		}
-	}()
+	processingErrorsCh := make(chan error, 2)
```
lobkovilya: why 2?

pkg/kds/v2/client/kds_client.go
```diff
@@ -74,14 +75,18 @@ func NewKDSSyncClient(
 	}
 }
 
-func (s *kdsSyncClient) Receive() error {
+func (s *kdsSyncClient) SendReq() error {
```
lobkovilya: TBH I'm not sure if this fix is correct. All you did is moved `SendReq` to a separate goroutine, but this function runs once at the start and sends at most 20 DeltaDiscoveryRequests and exits (see it just loops over `s.resourceTypes`). 

However, the second function `ReceiveResp` despite the name actually not only receives DeltaDiscoveryResponses but also sends DeltaDiscoveryRequests (in NACK and ACK functions).

So it doesn't look like this PR actually decouples KDS grpc client `Send` and `Recv` methods in different goroutines.

lobkovilya (changes requested):

pkg/kds/mux/zone_sync.go
Icarus9913: Because the `SendReq` and `ReceiveResp` functions are in separate goroutines and share this error channel

pkg/kds/mux/client.go
Icarus9913: put the comments in codes in another PR to solve the goroutine leak problem

pkg/kds/v2/client/kds_client.go
lobkovilya: @Icarus9913 I checked today grpc [docs](https://github.com/grpc/grpc-go/blob/7235bb7dcd13cc2c219f8e680ba65e31199d4791/stream.go#L129-L132) today :
> It is safe to have a goroutine calling SendMsg and another goroutine calling RecvMsg on the same stream at the same time, **but it is not safe to call SendMsg on the same stream in different goroutines**. It is also not safe to call CloseSend concurrently with SendMsg.

But I think if `Subscribe` and `Watch` run concurrently from different goroutines then `SendMsg` is called concurrently on the same stream.

Icarus9913: oops, this might be a problem :(

Icarus9913: @lobkovilya Ideally, maybe we should have 3 goroutines like this:

```text

// DeltaDiscoveryRequest signal
go func() {
  for _, typ := range resourceTypes {
    deltaDiscoveryReqCh <- typ
  }
}

// grpc SendReq
go func() {
  for{
    select {
    case <-ackCh:
      grpc.SendMsg()

    case <-nackCh:
      grpc.SendMsg()

    case <-deltaDiscoveryReqCh:
      grpc.SendMsg()
    }
  }
}

// grpc ReceiveResp
go func() {
  for{
    grpc.RecvMsg()
    if err != nil {
      nackCh <- struct{}
    }else {
      ackCh <- struct{}
    }
  }
}
```

Or do you have any better ideas?

jijiechen: We need to make sure the ACK/NACK is sent after the subscribing requests.

Icarus9913: > We need to make sure the ACK/NACK is sent after the subscribing requests.

It's fine since the ACK/NACK needs to carry the `resourceType` parameter, see: [ACK](https://github.com/kumahq/kuma/blob/6ffe623803707cb56e866630e6eef276b7d295e5/pkg/kds/v2/client/stream.go#L119), [NACK](https://github.com/kumahq/kuma/blob/2.12.0/pkg/kds/v2/client/stream.go#L137)

In my new pseudocode, actually the `ACK/NACK` won't impact the `deltaDiscoveryReq` since they're belong to different resources types.

jijiechen: Just had a review of this PR and had a discussion on this thread, I believe this fix demonstrated by this pseudo code will do the trick.

lobkovilya:
1. I'd revert changes with `Subscribe` and `Watch` keeping only `Receive` method.
2. We can have dedicated goroutine that reads from `sendCh` and calls `SendMsg` and another dedicated goroutine that calls `RecvMsg` and puts them into `recvCh`. Method `Receive` will look kind of similar to what it is now, but instead of calling `SendMsg` and `RecvMsg` directly it'll just put msgs in `sendCh` and `recvCh`. That's basically what you've suggested as "3 goroutine" solution. 
3. Please keep all changes within `Receive` method. So "clients" of `Receive` method could still use it as before. Don't change callbacks like `OnGlobalToZoneSyncConnectFunc` and `OnZoneToGlobalSyncConnectFunc`. These callbacks needs to be removed altogether in the following PRs (I might do that later).

Icarus9913:
After discussing with @lobkovilya, we'd like to create the grpc `stream` within the `Receive()` function, then we can have full control to cancel the stream.

One thing needs to be mentioned based on the above sentence: the grpc `Recv` function is I/O blocking, and `Send` could be I/O blocking.

Icarus9913:
close it in favor of https://github.com/kumahq/kuma/pull/15042

Issue #13941, opened 2025-07-14
Review all policy plugins to ensure they skip xDS config when no matching policy applies


bartsmykla:
### Description

During work on #13798, we found a bug where the `MeshTLS` policy plugin incorrectly generated Envoy resources even when no `MeshTLS` policy was configured (covered in #13939). The root cause was an incomplete check that only verified if the policy type existed, not whether actual rules were defined.

We should review **all policy plugins** in Kuma and Kong Mesh to ensure that none of them suffer from the same issue. Specifically:

* Verify that each plugin properly checks for the existence of actual policy rules before generating any Envoy resources.
* Determine if there's a reliable way (such as code checks, linters, or unit tests) to automatically detect or prevent similar issues in the future.

### Goals:

* Identify and fix plugins with incomplete checks.
* Suggest or implement improvements to prevent similar mistakes.

---

### Review

There seems to be a misunderstanding about what the following check is doing:

```go
policies, ok := proxy.Policies.Dynamic[api.MeshTimeoutType]
if !ok {
	return nil
}
```

It seems to be assumption this verifies whether any `MeshTimeout` policies apply to the current data plane proxy. But that's incorrect. The `ok` check only confirms the presence of the policy type in the map, not whether any actual rules exist. Even when no `MeshTimeout` policy is defined, `ok` will still be `true` and the returned object will look like this:

```json
{
  "Type": "MeshTimeout",
  "InboundPolicies": null,
  "OutboundPolicies": null,
  "ServicePolicies": null,
  "DataplanePolicies": [],
  "FromRules": {
    "Rules": {},
    "InboundRules": {}
  },
  "ToRules": {
    "Rules": [],
    "ResourceRules": {}
  },
  "GatewayRules": {
    "ToRules": {
      "ByListener": {},
      "ByListenerAndHostname": {}
    },
    "FromRules": {},
    "InboundRules": {}
  },
  "SingleItemRules": {
    "Rules": []
  },
  "Warnings": null
}
```

Instead of relying on `ok`, each plugin should explicitly check whether any rules are present for that policy type.

It's unclear whether this behavior changed recently or if these checks were ineffective from the start. I haven’t looked into the history in detail yet.

#### Incorrect checks

* **MeshAccessLog**

  https://github.com/kumahq/kuma/blob/1ead87e0c94bec07899322496d8c2f908ddc8341/pkg/plugins/policies/meshaccesslog/plugin/v1alpha1/plugin.go#L59-L62

[...]

PR #14529, opened 2025-09-22
feat(kuma-cp): do not run policies code if policy is not configured

Automaat:
## Motivation

We don't want to run policies code if no policy was configured. The only exception is MeshTimeout where we want to configure defaults even if no policies were configured. This should have positive impact on performance as we won't run code that we know won't produce any result

Fix https://github.com/kumahq/kuma/issues/13941

### DIFF
--- a/pkg/plugins/policies/core/rules/rules.go
+++ b/pkg/plugins/policies/core/rules/rules.go
@@ -55,9 +55,17 @@ type FromRules struct {
 	InboundRules map[InboundListener][]*inbound.Rule
 }
 
+func (f FromRules) Configured() bool {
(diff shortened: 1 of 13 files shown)


bartsmykla (approved):

lobkovilya (changes requested):
I think correct way to solve this is to implement [Policy Zero](https://github.com/Kong/kong-mesh/issues/7442) concept. If we just skip plugins then it's difficult to show in the Inspect API what values are applied when no policy matches.

Automaat:
Closing in favour of: https://github.com/Kong/kong-mesh/issues/7442#issue-2869535525

Issue #9161, opened 2024-02-07
allow using envoy's zone aware routing


aniket-z:
### Description

Our current setup has the following components:
- kuma-cp running in universal mode
- hundreds of ECS services across multiple AWS ECS clusters
- all kuma-dps configured with same zone (i.e. "kuma.io/zone" is same for all dataplanes)

Now, we want to use [envoy's zone aware routing feature](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/zone_aware) to reduce our inter-AZ network cost

We are aware that kuma supports [locality-aware-routing](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/031-locality-aware-load-balancing.md) but there it seems we are configuring routing using [envoy's priority based load balancing](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/priority) rather than using [envoy's zone aware routing feature](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/zone_aware).

Consider a scenario where source service calls destination service and source service has 4 tasks in zone A and 1 task in zone B but destination service has 1 task in zone A and 4 tasks in zone B. Here, if we use kuma's locality aware routing, from what I have understood, it seems that traffic on all tasks of destination service would not be same (and thus CPU % would not be same for all tasks) whereas if we were to use envoy's zone aware routing throughput per task of destination service (and thus the CPU%) would be same as it does routing taking zone level task count of both source & destination service in account. Please correct me if I have misunderstood envoys' zone aware routing or kuma's locality aware routing and if the problem I have described is not valid?

Is there any way we could use envoy's zone aware routing feature with kuma? We don't want to change kuma.io/zone as that would require us to setup kuma's zone ingress & egress and we don't want to introduce an extra hop (and thus extra cost & latency) and an extra component into our system. Please suggest how should we proceed here

lukidzi:
Hi, I think the thing you want to achieve can be done with Kuma by using https://kuma.io/docs/2.6.x/policies/meshloadbalancingstrategy/#disable-cross-zone-traffic-and-prioritize-traffic-the-dataplanes-on-the-same-node-and-availability-zone

First you need to add a tag to your dataplane in the same availability zone. 

e.g: 
```yaml
type: Dataplane
mesh: default
name: { { name } }
networking:
  address: { { address } }
  inbound:
    - port: 8000
      servicePort: 80
[...]

PR #14532, opened 2025-09-23
docs(MADR): add MADR-083 for Envoy Zone-Aware Routing

m4l1c1ou5:
Introduce a new MADR that defines how Envoy’s zone-aware routing feature will be integrated into Kuma’s control plane.

### What changed
- Added MADR-083: “Enable Envoy Zone-Aware Routing in Kuma”
- Defined decision context, drivers, and outcome for injecting zone_aware_lb_config via EDS/CDS 
- Specified new MeshLoadBalancingStrategy API with type: ZoneAware  
- Provided example YAML showing ZoneAware configuration using zoneIdentifier and minClusterSize  

### Why
- Ensure per-host throughput balance and minimize cross-AZ data transfer costs
- Preserve Kuma’s declarative abstraction and existing kuma.io/zone labels  

Implements MADR for #9161

### DIFF
--- a/docs/madr/decisions/083-zone-aware-routing.md
+++ b/docs/madr/decisions/083-zone-aware-routing.md
@@ -1,0 +1,87 @@
+# Envoy Zone Aware Routing Integration
+
+**Status:** accepted
+
+## Context and Problem Statement
+
+Kuma’s existing **locality-aware load balancing** uses Envoy’s priority-based routing via `MeshLoadBalancingStrategy` to favor dataplanes sharing the same `kuma.io/zone` tag. However, it does **not** account for the relative number of instances per availability zone when distributing traffic. In environments where tasks are unevenly distributed across zones, this can overload individual instances and degrade performance.
+
+**Example scenario:**
+- **Source service:** 4 tasks in AZ-A, 1 in AZ-B
+- **Destination service:** 1 task in AZ-A, 4 in AZ-B  
+  Priority-based routing would still favor AZ-A—even though it has a single instance—resulting in uneven per-instance load. Envoy’s **zone aware routing** would balance evenly across all endpoints while favoring the local zone.
+
+## Decision Drivers
+
+- **True zone-awareness:** factor zone-level instance counts into routing decisions.
+- **Minimal infrastructure changes:** no additional proxies or sidecars.
+- **Leverage Envoy’s native implementation:** avoid reimplementing complex load-balancing logic.
+- **Compatibility:** support both Kubernetes and universal modes.
+- **Maintain existing `kuma.io/zone` tags:** avoid forcing users to reconfigure zones.
+
+## Considered Options
+
+1. **Enhance priority-based strategy**  
+   Extend `MeshLoadBalancingStrategy` to calculate custom weights mimicking zone-aware behavior.  
+   *Pros:* No external flags. *Cons:* Reinventing Envoy’s algorithm, maintenance burden.
+
+2. **Expose Envoy’s `zone_aware_lb_config` directly**  
+   Add a new `type: ZoneAware` mode in `MeshLoadBalancingStrategy` that sets `common_lb_config.zone_aware_lb_config`.  
+   *Pros:* Minimal CP logic; leverages Envoy. *Cons:* Requires documentation and policy extension.
+
+## Decision Outcome
+
+Chosen option: **Expose Envoy’s `zone_aware_lb_config` directly** as a new `ZoneAware` mode in `MeshLoadBalancingStrategy`.
+
+### Implementation Details
+
+- Extend `MeshLoadBalancingStrategy.spec.default.localityAwareness.localZone` to accept:
+    - `type: ZoneAware`
+    - `minClusterSize`: mapped to Envoy’s `zone_aware_lb_config.min_cluster_size`
+    - `zoneIdentifier`: tag key for zone (e.g., `topology/availability-zone`)
+  - Kuma CP will inject into each dataplane’s bootstrap:
+    ```yaml
+    cluster_manager:
+      local_cluster_name: {{ service_name }}
+    node:
+      locality:
+        zone: {{ zone_tag }}
+    ```
+  - Kuma CP will inject into each relevant cluster:
+    ```yaml
+    common_lb_config:
+      zone_aware_lb_config:
+        min_cluster_size: {{ minClusterSize }}
+    ```
+
+### Example Configuration
+
+```yaml
+type: MeshLoadBalancingStrategy
+name: zone-aware
+mesh: default
+spec:
+  targetRef:
+    kind: Dataplane
+    labels:
+      app: frontend
+  to:
+    - targetRef:
+        kind: MeshService
+        name: backend
+        sectionName: http
+      default:
+        loadBalancer:
+          type: ZoneAware
+          zoneAware:
+            # Minimum total endpoints before zone-aware logic activates
+            minClusterSize: 1
+            # Tag used on dataplanes to identify their zone
+            zoneIdentifier: topology/availability-zone
+```
+
+## Consequences
+
+- **Backward compatibility:** Existing weighted strategies remain unchanged.
+- **Policy changes:** Users opt in via `type: ZoneAware`.
+- **Documentation:** Need examples, guides, and upgrade notes.


docs/madr/decisions/083-zone-aware-routing.md
```diff
@@ -0,0 +1,87 @@
...
+## Decision Outcome
+
+Chosen option: **Expose Envoy’s `zone_aware_lb_config` directly** as a new `ZoneAware` mode in `MeshLoadBalancingStrategy`.
+
+### Implementation Details
+
+- Extend `MeshLoadBalancingStrategy.spec.default.localityAwareness.localZone` to accept:
+    - `type: ZoneAware`
+    - `minClusterSize`: mapped to Envoy’s `zone_aware_lb_config.min_cluster_size`
+    - `zoneIdentifier`: tag key for zone (e.g., `topology/availability-zone`)
```
slonka, line 41: what about 

```
upstream.zone_routing.force_local_zone.min_size
```

?

docs/madr/decisions/083-zone-aware-routing.md
```diff
@@ -0,0 +1,87 @@
...
+  - Kuma CP will inject into each dataplane’s bootstrap:
+    ```yaml
+    cluster_manager:
+      local_cluster_name: {{ service_name }}
+    node:
+      locality:
+        zone: {{ zone_tag }}
```
slonka, line 48: this is the value of `zoneIdentifier` evaluated at the dp, right?

docs/madr/decisions/083-zone-aware-routing.md
slonka: also do you know what @lukidzi wanted to achieve with this `subZoneIdentifier` mentioned in the issue?

docs/madr/decisions/083-zone-aware-routing.md
```diff
@@ -0,0 +1,87 @@
...
+
+Chosen option: **Expose Envoy’s `zone_aware_lb_config` directly** as a new `ZoneAware` mode in `MeshLoadBalancingStrategy`.
+
+### Implementation Details
+
+- Extend `MeshLoadBalancingStrategy.spec.default.localityAwareness.localZone` to accept:
+    - `type: ZoneAware`
+    - `minClusterSize`: mapped to Envoy’s `zone_aware_lb_config.min_cluster_size`
+    - `zoneIdentifier`: tag key for zone (e.g., `topology/availability-zone`)
+  - Kuma CP will inject into each dataplane’s bootstrap:
```
slonka, line 42: this would be in https://github.com/kumahq/kuma/blob/002e301acf53484f3508a2a7e4ebce8e7df97f66/app/kuma-dp/pkg/dataplane/envoy/remote_bootstrap.go#L134 right?

slonka (commented):
I think it makes sense, would love some input from @justindavies as well

docs/madr/decisions/083-zone-aware-routing.md
```diff
@@ -0,0 +1,87 @@
...
+    kind: Dataplane
+    labels:
+      app: frontend
+  to:
+    - targetRef:
+        kind: MeshService
+        name: backend
+        sectionName: http
+      default:
+        loadBalancer:
```
lobkovilya, line 74: IIUC `zone_aware_lb_config` and `locality_weighted_lb_config` are 2 alternative approaches and they're mutually exclusive. At this moment, we are already using `locality_weighted_lb_config` when configuring policy like
```yaml
type: MeshLoadBalancingStrategy
name: local-zone-affinity-backend
mesh: default
spec:
  to:
  - targetRef:
      kind: MeshService
      name: backend
    default:
      localityAwareness:
        localZone:
          affinityTags:
            - key: kubernetes.io/hostname
              weight: 9000
            - key: topology.kubernetes.io/zone
              weight: 9
```

Your suggested API change doesn't reflect the fact 2 approaches are mutually exclusive, i.e. we'd have to add extra validation to reject policies like this (because it's not possible to use zone_aware_lb and locality_weighted_lb at the same time):
```yaml
type: MeshLoadBalancingStrategy
name: local-zone-affinity-backend
mesh: default
spec:
  to:
  - targetRef:
      kind: MeshService
      name: backend
    default:
      loadBalancer:
        type: ZoneAware
        zoneAware:
          # Minimum total endpoints before zone-aware logic activates
          minClusterSize: 1
          # Tag used on dataplanes to identify their zone
          zoneIdentifier: topology/availability-zone
      localityAwareness:
        localZone:
          affinityTags:
            - key: kubernetes.io/hostname
              weight: 9000
            - key: topology.kubernetes.io/zone
              weight: 9
```

I think MeshLoadBalancingStrategy policy API should make it clear to users that `zoneAware` and `localZone` are 2 mutually exclusive approaches.

lobkovilya (commented):
Thanks for the MADR, I think I understand the motivation, I left a comment related to the API

docs/madr/decisions/083-zone-aware-routing.md
m4l1c1ou5: yes

docs/madr/decisions/083-zone-aware-routing.md
m4l1c1ou5: @lobkovilya i was also concerned about this issue, but since the existing policy structure does not have generic keys like 'type' that takes kind of enums, I'm not able to solve this issue
All i can think of is adding validation layer during policy apply
If you have any better solution, please help

docs/madr/decisions/083-zone-aware-routing.md
m4l1c1ou5: @slonka 'upstream.zone_routing.force_local_zone' is a runtime config and as far as i understand, runtime config does not support cluster level configurations.
Is it correct or I'm missing something?

m4l1c1ou5: I'm also not sure about why @lukidzi suggested subzoneid

docs/madr/decisions/083-zone-aware-routing.md
m4l1c1ou5: Yes

docs/madr/decisions/083-zone-aware-routing.md
slonka: @lukidzi - can you weight in here?

docs/madr/decisions/083-zone-aware-routing.md
slonka: @lobkovilya any suggestions here?

docs/madr/decisions/083-zone-aware-routing.md
m4l1c1ou5: @lukidzi any update here?

docs/madr/decisions/083-zone-aware-routing.md
m4l1c1ou5: @lobkovilya any update here?

docs/madr/decisions/083-zone-aware-routing.md
```diff
@@ -0,0 +1,87 @@
...
+   Add a new `type: ZoneAware` mode in `MeshLoadBalancingStrategy` that sets `common_lb_config.zone_aware_lb_config`.  
+   *Pros:* Minimal CP logic; leverages Envoy. *Cons:* Requires documentation and policy extension.
+
+## Decision Outcome
+
+Chosen option: **Expose Envoy’s `zone_aware_lb_config` directly** as a new `ZoneAware` mode in `MeshLoadBalancingStrategy`.
+
+### Implementation Details
+
+- Extend `MeshLoadBalancingStrategy.spec.default.localityAwareness.localZone` to accept:
```
Automaat, line 38: **issue:** This says to extend `MeshLoadBalancingStrategy.spec.default.localityAwareness.localZone` with the new `type`/`minClusterSize`/`zoneIdentifier` fields, but the "Example Configuration" below (lines 74-80) nests those exact same fields under a different path instead: `spec.to[].default.loadBalancer.type: ZoneAware` + a new `loadBalancer.zoneAware` block.

Checked against the current schema (`pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go`): `LocalZone` has only one field, `AffinityTags` — no `type` discriminator exists there today. `LoadBalancer`, however, already has a `// +kuma:discriminator` `Type` field with a same-named nested struct per variant (`RoundRobin`, `LeastRequest`, `RingHash`, `Random`, `Maglev`). The Example Configuration's shape matches `LoadBalancer` (adding a `ZoneAware` variant), not `LocalZone`. An implementer following this bullet literally would extend the wrong struct.

Worth aligning this section with the Example Configuration before implementation starts, since resolving exactly this kind of ambiguity is the point of the Implementation Details section.

_Generated by Sybra harness_

Automaat (commented):
Review Verdict: NEEDS_FIXES

Adversarial review of MADR-083 (Envoy Zone-Aware Routing). One issue-level finding: the "Implementation Details" section names a different schema location than the "Example Configuration" actually uses for the same new fields — see inline comment. No blocking issues; nothing else survived adversarial refutation.

_Generated by Sybra harness_

Issue #12372, opened 2024-12-23
Add `spec.rules` for MeshFaultInjection


lobkovilya:
### Description

We should add a `rules` list without L7 `matches` (only `rules[].default`)

```yaml
spec:
  targetRef:
    kind: Dataplane
    labels:
      app: frontend
  rules:
    - default:
        http:
          - abort:
              httpStatus: 500
              percentage: "2.5"
```

This issue covers only API change.

PR #14533, opened 2025-09-23
feat(MeshFaultInjection): implement rules API

Automaat:
## Motivation

We now have Matching api so we can implement rules API for MeshFaultInjection

Fix https://github.com/kumahq/kuma/issues/12372


pkg/plugins/policies/meshfaultinjection/api/v1alpha1/validator_test.go
```diff
@@ -119,6 +148,77 @@ from:
...
+				{
+					Field:   "spec.rules[0].default.http.delay[1].value",
+					Message: "must not be negative when defined",
+				},
+				{
+					Field:   `spec.rules[0].default.http.delay[1].percentage`,
+					Message: `must be in inclusive range [0.0, 100.0]`,
+				},
+				{
+					Field:   `spec.rules[0].default.http.responseBandwidth[2].responseBandwidth`,
```
slonka: I think path is missing `limit` here, should be:

```suggestion
					Field:   `spec.rules[0].default.http.responseBandwidth[2].responseBandwidth.limit`,
```

slonka (commented):
One thing, apart from that 👍

pkg/plugins/policies/meshfaultinjection/api/v1alpha1/validator_test.go
Automaat: fixed

PR #14543, opened 2025-09-25
chore(.github): add workflow sync kuma oas to kuma-gui

schogges:
## Motivation

To keep the kuma-gui TypeScript types that are based on and generated from the kuma OAS in sync.

## Implementation information

The workflow should only run on pushes to the trunk, on release branches and only when there are changes in `docs/generated/openapi.yaml`.

Please also refer to the [workflow in kuma-gui](https://github.com/kumahq/kuma-gui/blob/master/.github/workflows/sync-kuma-oas.yaml) that is being triggered by the dispatched event of this workflow.

Part of https://github.com/kumahq/kuma-gui/issues/2876

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -58,7 +58,7 @@ Before upgrading, ensure that your configuration does not override this setting.
 
 ### System namespace requires the `kuma.io/sidecar-injection: false` label
 
-To simplify the namespace selector logic in webhooks, we now require the `kuma.io/sidecar-injection: false` label to be set on the system namespace.
+To simplify the namespace selector logic in webhooks, we now require the `kuma.io/sidecar-injection: false` label to be set on Kuma's system namespace (`kuma-system` by default).
 
 Since Kubernetes v1.22, the API server automatically adds the `kubernetes.io/metadata.name` label to all namespaces. As a result, we’ve replaced the use of the custom `kuma.io/system-namespace` label in the secret webhook selector with this standard label.
 


.github/workflows/sync_kuma_oas_to_kuma-gui.yaml
```diff
@@ -0,0 +1,38 @@
...
+      TARGET_REPO: kuma-gui
+      EVENT_TYPE: SYNC_KUMA_OAS
+    steps:
+      - uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          github-token: ${{ secrets.NOTIFY_BOT_PAT_TOKEN }}
+          script: |
+            const branch = process.env.GITHUB_REF_NAME;
+            return github.rest.repos.createDispatchEvent({
+              owner: '${{ secrets.NOTIFY_OWNER }}',
```
schogges, line 32: I took `secrets.NOTIFY_OWNER` from another workflow in this repo. `kuma-gui` and `kuma` have the same owner, so I think this should be the same. Please lmk 🙂

.github/workflows/sync_kuma_oas_to_kuma-gui.yaml
```diff
@@ -0,0 +1,38 @@
...
+  dispatch-event:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    env:
+      TARGET_REPO: kuma-gui
+      EVENT_TYPE: SYNC_KUMA_OAS
+    steps:
+      - uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          github-token: ${{ secrets.NOTIFY_BOT_PAT_TOKEN }}
```
schogges, line 28: I'm not super sure about using `secrets.NOTIFY_BOT_PAT_TOKEN`, so please let me know. Otherwise I'd use `actions/create-github-app-token` to generate a token.

lahabana (approved):
That's right it has to be this token. Because the bot token is limited to the kuma org

schogges:
@lahabana if it's good to go, could you merge it please? I don't have permissions 😄

Issue #11001, opened 2024-07-29
Deprecate possibility to specify `Mesh*Route` in `spec.targetRef` policy section


Automaat:
### Description

As described in [MADR](https://github.com/kumahq/kuma/pull/10858)

lahabana:
Should that even be 2.11.x ? Also have we already added a note in UPGRADE.md?

lahabana:
Can only happen after: #11000, #10999 is done

PR #14544, opened 2025-09-25
chore(targetRef): deprecate possibility to target MeshHTTPRoute in top level targetRef

Automaat:
## Motivation

We can now target MeshHTTPRoute in`to` section of policy, which should be used for targeting. Since this is now released, we can deprecate old way of targeting routes

Fix https://github.com/kumahq/kuma/issues/11001

### DIFF
--- a/pkg/plugins/policies/core/jsonpatch/validators/validator.go
+++ b/pkg/plugins/policies/core/jsonpatch/validators/validator.go
@@ -129,7 +129,12 @@ func TopLevelTargetRefDeprecations(targetRef *common_api.TargetRef) []string {
 	}
 	if _, ok := replacedByDataplane[targetRef.Kind]; ok {
 		return []string{
-			fmt.Sprintf("%s value for 'targetRef.kind' is deprecated, use %s with labels instead", common_api.MeshSubset, common_api.Dataplane),
+			fmt.Sprintf("%s value for 'targetRef.kind' is deprecated, use %s with labels instead", targetRef.Kind, common_api.Dataplane),
+		}
+	}
+	if targetRef.Kind == common_api.MeshHTTPRoute {
+		return []string{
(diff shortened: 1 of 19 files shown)


pkg/plugins/runtime/k8s/webhooks/testdata/validation/cli-user_create_mt-route-deprecated.global.golden.yaml
```diff
@@ -0,0 +1,10 @@
+Patches: null
+allowed: false
+status:
+  code: 403
```
slonka, line 4: should this be a 403? seems unrelated to the issue - it's missing label

pkg/plugins/policies/core/jsonpatch/validators/validator.go
```diff
@@ -129,7 +129,12 @@ func TopLevelTargetRefDeprecations(targetRef *common_api.TargetRef) []string {
 	}
 	if _, ok := replacedByDataplane[targetRef.Kind]; ok {
 		return []string{
-			fmt.Sprintf("%s value for 'targetRef.kind' is deprecated, use %s with labels instead", common_api.MeshSubset, common_api.Dataplane),
```
slonka, line 132: so was meshsubset here by mistake?

Automaat: yes it was by mistake

pkg/plugins/runtime/k8s/webhooks/testdata/validation/cli-user_create_mt-route-deprecated.global.golden.yaml
Automaat: this is expected, this policy is applied on zone, when you look at other test cases it looks the same

slonka (approved):

PR #14568, opened 2025-09-30
ci(k3d): cluster creation timeout is insufficient

lobkovilya:
## Motivation

Pulling k3s images from docker.io registry sometimes takes more than 2 minutes. Long-term fix will resolve the actual problem of long pulling, but for now we just want to have CI running on `master`.

## Implementation information

Increased timeout up to 300s


lobkovilya:
We can merge this if we want to get `master` to working state as quickly as possible. I opened and issue to solve the actual root cause https://github.com/kumahq/kuma/issues/14569

bartsmykla (approved):

Issue #14462, opened 2025-09-10
Add KRI to resource/policy response


lukidzi:
### Description

We use KRI internally for linking between resource. This is beneficial and we want to use it also in the frontend. Unfortunately currently it's not available and it's required to be build by the fronted. That is problematic and might lead to diffrent issues. We should investigate the possibility to add KRI to the response

PR #14571, opened 2025-10-01
docs(MADR): record decision to add KRI to REST API resources

bartsmykla:
## Motivation

The GUI and other clients need a stable machine generated identifier to reliably correlate REST API resources with the origins that produced their Envoy configuration. Reconstructing KRI on the client is fragile and inconsistent across endpoints.

## Implementation information

This PR adds a MADR documenting the decision to compute and inject `meta.kri` during JSON marshalling of `ResourceMeta`. The decision explains why this is preferred over adding fields across many structs and code paths. It also notes that OpenAPI templates will be updated to document `meta.kri` as a readOnly string and clarifies that legacy protobuf resources are out of scope beyond existing Inspect API coverage.

## Supporting documentation

- Technical Story: https://github.com/kumahq/kuma/issues/14462
- GUI Issue: https://github.com/kumahq/kuma-gui/issues/4220

### DIFF
--- a/docs/madr/decisions/084-add-kri-to-rest-api-resources.md
+++ b/docs/madr/decisions/084-add-kri-to-rest-api-resources.md
@@ -1,0 +1,138 @@
+# Add KRI to All REST API Resources
+
+* Status: Accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/14462
+GUI Issue: https://github.com/kumahq/kuma-gui/issues/4220
+
+## Context and Problem Statement
+
+The GUI needs a stable, machine-generated identifier to link any resource or policy from the REST API with the origin that produced its Envoy configuration. We already use KRI (Kuma Resource Identifier) in new components and in the Inspect API, where per-rule origins are exposed. But most REST list/get endpoints do not return the KRI field. Clients must reconstruct it manually, which is fragile and error-prone.
+
+We need a lightweight way to expose KRI across REST API responses without major refactors or storage changes.
+
+## Decision Drivers
+
+* Changes should be minimal, additive, and safe for existing clients
+* Avoid unnecessary work on legacy protobuf-based resources that are being deprecated
+* Make KRI reliably available to GUI and tooling
+* Minimize schema churn and developer confusion
+
+## Options Considered
+
+### Option A: Add a `kri` field to all resource/policy structs
+
+**Pros**
+
+* Explicit in types
+* OpenAPI can immediately document it
+
+**Cons**
+
+* KRI is derived, not stored - we must recompute it everywhere
+* Requires touching many code paths
+* High risk of inconsistencies or bugs
+
+### Option B: Compute and inject `kri` during JSON marshalling of `ResourceMeta` (chosen)
+
+**Idea**
+
+* In `pkg/core/resources/model/rest/v1alpha1/ResourceMeta`, override JSON marshalling to insert a computed `kri` based on `type`, `mesh`, `labels` and `name`.
+
+**Pros**
+
+* Minimal code changes
+* Automatically appears wherever `ResourceMeta` is used
+* Clients see it without needing to modify storage
+* No problem with OpenAPI schema generation since our templates can be adjusted to include it
+
+**Cons**
+
+* The field is computed and not settable in code
+
+### Option C: Add a `KRI` field to `ResourceMeta`, but only set it at marshalling time
+
+**Pros**
+
+* Easier to annotate as `readOnly` in schemas
+
+**Cons**
+
+* Confusing to developers (should not be set manually)
+* Still needs extra scaffolding for clarity and safety
+
+## Decision
+
+We choose Option B: compute and inject `kri` during JSON marshalling of `ResourceMeta`. This meets our goal of minimal, safe change, while making KRI available in REST API responses without touching storage or reconciliation. Our OpenAPI templates will be updated to include the field and mark it as `readOnly`. This also aligns with how KRI is used in the Inspect API for per-rule origin.
+
+## Design
+
+```go
+type ResourceMeta struct {
+    Type             string            `json:"type"`
+    Mesh             string            `json:"mesh,omitempty"`
+    Name             string            `json:"name"`
+    CreationTime     time.Time         `json:"creationTime"`
+    ModificationTime time.Time         `json:"modificationTime"`
+    Labels           map[string]string `json:"labels,omitempty"`
+}
+
+func (r ResourceMeta) MarshalJSON() ([]byte, error) {
+    type Alias ResourceMeta
+    return json.Marshal(&struct {
+        Alias
+        KRI string `json:"kri,omitempty"`
+    }{
+        Alias: Alias(r),
+        KRI:   kri.FromResourceMeta(r, core_model.ResourceType(r.Type)).String(),
+    })
+}
+```
+
+* `KRI` is derived and never stored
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/084-add-kri-to-rest-api-resources.md
```diff
@@ -0,0 +1,138 @@
...
+**Pros**
+
+* Explicit in types
+* OpenAPI can immediately document it
+
+**Cons**
+
+* KRI is derived, not stored - we must recompute it everywhere
+* Requires touching many code paths
+* High risk of inconsistencies or bugs
```
lahabana, line 42: Doesn't work on k8s right?

bartsmykla: It doesn’t apply when talking about Custom Resource Definitions and the Kubernetes API. From what I understand, this task is mainly about the GUI and our REST API.

bartsmykla:
>  If we think it's unecessary I'd like this to be covered in "ctx and pb stmt" as non goal.

addressed

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
bartsmykla: I have addressed it by expanding the `Context and Problem Statement` section

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
```diff
@@ -0,0 +1,140 @@
...
+}
+```
+
+* `KRI` is derived and never stored
+* Anywhere `ResourceMeta` is rendered in a REST response, `meta.kri` will be included
+* No changes to storage, reconciliation, or resource logic
+
+## Scope and Legacy Resources
+
+* We will not retrofit full KRI support into legacy protobuf-based resources beyond what [Inspect API](./075-inspect-api-redesign.md) already covers
```
johncowen: From what I understand `Dataplanes` are protobuf-based (but not considered legacy), just wanted to check whether we are planning on adding a `kri` field to `Dataplanes` (and also inbounds)

bartsmykla: I clarified this in the section: `Dataplane`, `Mesh`, `ZoneIngress`, and `ZoneEgress` are resources defined in protobufs, and they are not deprecated or legacy.

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
```diff
@@ -0,0 +1,141 @@
...
+
+* Resources defined in protobuf that are part of the core data model, such as `Dataplane`, `Mesh`, `ZoneIngress`, and `ZoneEgress`, are **not** considered deprecated or legacy. When returned by the REST API, they will also include the `kri` field
+* We will not retrofit full KRI support into truly legacy protobuf-based policies that are being phased out. Those remain covered only by what the [Inspect API](./075-inspect-api-redesign.md) already provides
+* Adding KRI to deprecated policies would require defining short names and introducing additional logic, which we plan to remove in version 3.0.0
+* Only the new Inspect API endpoints can return an origin resource by its KRI for use in the GUI. Deprecated policy types will not appear as origins
+
+## Compatibility
+
+* The change is additive in JSON. Clients ignoring unknown fields continue to work
+* No change to storage, hashing, reconciliation, or system behavior
```
lahabana: Anything special to add to this field to not confuse terraform for example?

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
```diff
@@ -0,0 +1,141 @@
...
+
+## API Schema and OpenAPI
```
lahabana: I'd put this in the  design section

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
```diff
@@ -0,0 +1,141 @@
...
+
+* OpenAPI templates will be updated so `meta.kri` is documented as `string` and `readOnly: true`
+* Clients will see it in generated SDKs
```
lahabana: This should go in `compatibility`

lahabana (approved):
Solid MADR. I only have formatting nits.

You probably also want to mention the issue which adds to the API the ability to retrieve by KRI so someone reading this MADR can go to the other issue

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
johncowen: Perfect! Thanks @bartsmykla 👍

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
bartsmykla: I updated this section and added example changes with `readOnly: true`, which I believe is the only field needed to avoid confusing Terraform.

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
bartsmykla: Addressed

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
bartsmykla: Addressed

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
bartsmykla: @johncowen the only limitation is that KRIs won’t appear in the inbounds of the Dataplane, since this work only covers Resource Meta changes. Updating the `Dataplane` resource itself would be required to adjust its structure.

johncowen: Ah! Glad you mentioned. Ok well that piece of my question was a "nice to have". Is that totally out of the question then, or just much more effort?

bartsmykla: I’d say it’s out of scope for this MADR, but not completely off the table. After this one, I’ll open a small follow-up MADR for those changes to get a clearer idea of the effort involved and to build consensus around it. Wdyt?

johncowen: Gotcha, sounds great, thanks Bart!

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
```diff
@@ -0,0 +1,205 @@
...
+ default: default
+ {{- end}}
++  {{- if ne .ShortName ""}}
++  kri:
++    description: 'A unique identifier for this resource instance, derived from
++      resource attributes such as type, mesh, and name. Used by internal tooling
++      and integrations for cross-references or indexing'
++    type: string
++    readOnly: true
++    example: 'kri_{{ .ShortName }}_default__kuma-system_{{ .ShortName }}123_'
```
Automaat, line 125: Why `{{ .ShortName }}` is duplicated here? the second is actual name of the resource not short name of resource type

Automaat (approved):

docs/madr/decisions/084-add-kri-to-rest-api-resources.md
bartsmykla: We don’t have a `.Name` in this context since it refers to the type descriptor, not the resource. The second value is used just to mimic a name, like `mtp123` or `mal123`.

slonka (approved):

PR #14572, opened 2025-10-01
docs(MADR): document switch to Renovate preset for active branches

bartsmykla:
## Motivation

We want one source of truth for active branches and the ability for Renovate to use it directly. The current file is a plain array that scripts can read, but Renovate cannot extend it as a preset. This limits Renovate features on release branches, like OSV alerts and controlled package updates.

## Implementation information

This PR adds a MADR that formalizes the change from an array to an object with `baseBranchPatterns`. Renovate can then extend the preset with `kumahq/kuma:active-branches`. The document covers:

- The new JSON shape and how Renovate reads it
- The benefits of enabling OSV alerts on release branches
- A low risk migration plan that supports both shapes
- `jq` snippets to keep existing scripts working during rollout
- An example Renovate config to limit updates on release branches to OSV and Envoy patch bumps

Alternatives considered included keeping the array and adding a second preset file, or keeping both files during a longer migration. We chose a single file to avoid duplication and make the change immediately useful.

### DIFF
--- a/docs/madr/decisions/085-active-branches-renovate-preset.md
+++ b/docs/madr/decisions/085-active-branches-renovate-preset.md
@@ -2,8 +2,6 @@
 
 * Status: Accepted
 
-Technical Story: TBA
-
 ## Context and Problem Statement
 
 `active-branches.json` is a raw JSON array of branch patterns:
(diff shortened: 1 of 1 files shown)


slonka (approved):
sounds good 👍

Issue #14469, opened 2025-09-12
Inbound `proxyResourceName` doesn't exactly match with envoy stats


schogges:
### Kuma Version

2.12

### Describe the bug

With `features-unified-resource-naming` enabled on a DP and using the `/_layout` endpoint to fetch the traffic layout of the DP we receive a list of `inbounds` and `outbounds`. The new layout simplifies the mapping of envoy stats with `inbounds` and `outbounds` a lot, but there is a tiny detail that we still have to fix in the data-layer of the GUI.
Given an inbound with a `proxyResourceName` of `self_inbound_http` and a port of `5050`, we have to replace the section name `http` with `5050` (=> `self_inbound_5050`) to correctly map the respective envoy stats to the specific inbound. It would be great if we'd just receive the `proxyResourceName` that exactly matches with the envoy stats.

### To Reproduce

```sh
kumactl install control-plane --set dataPlane.features.unifiedResourceNaming=true | kubectl apply -f -
kubectl apply -f https://raw.githubusercontent.com/kumahq/kuma-counter-demo/refs/heads/main/k8s/000-with-kuma.yaml
```

Check the inbound stats of the DPs (demo-app and kv) and search/filter for `self_inbound_5050` and `self_inbound_http`

PR #14581, opened 2025-10-02
fix(envoy-naming): rename inbound Envoy resource names and stats to use `sectionName` instead of port value when using unified naming

Automaat:
## Motivation
Right now we are having issue with `inbounds[].proxyResourceName` in `_layout` endpoint being different then actual inbound name in Envoy config and stats. Which makes it impossible for GUI to simply match them. They need to have the same name. 

According to the [MADR](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/077-migrating-to-consistent-and-well-defined-naming-for-non-system-envoy-resources-and-stats.md#option-1-use-the-contextual-format-defined-for-proxy-local-resources) This should be `sectionName` not port value, so we need to align Envoy config and stats with API. This change should be only applied to unified resource naming

## Implementation information

Updating how we build inbound name for Envoy config and stats by extending `InboundInterface` with `sectionName`

Fix https://github.com/kumahq/kuma/issues/14469

### DIFF
--- a/pkg/core/faultinjections/matcher_test.go
+++ b/pkg/core/faultinjections/matcher_test.go
@@ -131,6 +131,7 @@ var _ = Describe("Match", func() {
 				mesh_proto.InboundInterface{
 					WorkloadIP:   "1.2.3.4",
 					WorkloadPort: 8080,
+					InboundName:  "0",
 				}: {
 					policyWithDestinationsFunc("fi2", time.Unix(1, 0), []*mesh_proto.Selector{
 						{
@@ -180,6 +181,7 @@ var _ = Describe("Match", func() {
 				mesh_proto.InboundInterface{
 					WorkloadIP:   "1.2.3.4",
 					WorkloadPort: 8081,
+					InboundName:  "0",
 				}: {
 					policyWithDestinationsFunc("fi1", time.Unix(1, 0), []*mesh_proto.Selector{
 						{
@@ -244,6 +246,7 @@ var _ = Describe("Match", func() {
 				mesh_proto.InboundInterface{
 					WorkloadIP:   "1.2.3.4",
 					WorkloadPort: 8080,
+					InboundName:  "0",
 				}: {
 					policyWithDestinationsFunc("fi1", time.Unix(1, 0), []*mesh_proto.Selector{
 						{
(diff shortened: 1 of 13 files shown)


Automaat:
@lahabana updated description, also I looked into MADR and we should use sectionName not port value, so I've rewritten this pr from scratch to align with it

pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin.go
```diff
@@ -284,7 +284,7 @@ func configureListener(
 	getName := naming.GetNameOrFallbackFunc(unifiedNaming)
 
 	inboundID := kri.WithSectionName(kri.From(proxy.Dataplane), iface.WorkloadPort).String()
-	inboundContextualID := naming.MustContextualInboundName(proxy.Dataplane, iface.WorkloadPort)
+	inboundContextualID := naming.MustContextualInboundName(proxy.Dataplane, inbound.GetSectionName())
```
bartsmykla: shouldn't we use `iface.InboundName` here?
```suggestion
	inboundContextualID := naming.MustContextualInboundName(proxy.Dataplane, iface.InboundName)
```

pkg/core/faultinjections/matcher_test.go
```diff
@@ -131,6 +131,7 @@ var _ = Describe("Match", func() {
 				mesh_proto.InboundInterface{
 					WorkloadIP:   "1.2.3.4",
 					WorkloadPort: 8080,
+					InboundName:  "0",
```
bartsmykla: Shouldn't we fall back to `WorkloadPort` when there’s no explicit `InboundName`? Is that scenario even possible?

bartsmykla (approved):
Two questions, but lgtm overall

pkg/core/faultinjections/matcher_test.go
Automaat: you cannot start dataplane without port value: https://github.com/kumahq/kuma/blob/87d0b7ec8e78838b9aeb529f21363ac46e053679/pkg/core/resources/apis/mesh/validators.go#L214-L222 We have a validation for this

pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin.go
Automaat: makes sense, updated

pkg/core/faultinjections/matcher_test.go
bartsmykla: shouldn't we also then adjust the tests so then InboundName is the same as WorkloadPort?

Automaat: I've updated tests to make it more accurate

bartsmykla (approved):

Issue #13882, opened 2025-07-07
Create an endpoint to retrieve a resource by its KRI


lukidzi:
### Description

Since we have introduced KRI in kuma code base and we want to adopt it now into the configuration, it might make sense to have a endpoint to retrieve it by the KRI. We should design how to achieve this
* `/resource/<KRI>` endpoint which either redirect to original endpoint or return the resource

lahabana:
Couple of points:

- API should be typed (this is required for frontend)
- Should we strip the sectionName automatically or ask clients to strip it?
- perf should be acceptable? Should we rebuild the hash to be able to do id lookup instead of per label lookup which will be expensive.
- Can we do KRI even for old resources? If not what's the condition on the type descriptor to figure if the API exists?
- Do all questions

It should be easy to integrate in our generic generation:

https://github.com/kumahq/kuma/blob/46b6807f56c08c79f34db20cfc452b8f5203bf90/tools/openapi/templates/endpoints.yaml#L9

And then add the endpoints in a generic like we do for everything.

https://github.com/kumahq/kuma/blob/b219aa49f11c7f7c38957d79a6c0c57d1b5f8d58/pkg/api-server/server.go#L338

So I think we could quite easily create a:

```yaml
/{{ .Path }}/_kri/{kri}:
  get:
     operationId: getByKri{{.Name}}
      summary: Returns {{ .Name }} entity
      tags: [ "{{ .Name }}" ]
      parameters:
        - in: path
          name: kri
          schema:
            type: string
          required: true
          description: KRI of the {{ .Name }}
      responses:
        '200':
          $ref: "#/components/responses/{{.Name}}Item"
        '404':
          $ref: "/specs/base/specs/common/error_schema.yaml#/components/responses/NotFound"
```

Implementing this this way would make it pretty straight forward and work out of the box for all resources.

johncowen:
> API should be typed (this is required for frontend)

Not totally sure how this will need to be done if we use `/resource/<KRI>` considering as OpenAPI/restful things probably won't be able to infer the type the the KRI.

We might need to do `/resource/<shortName>/<KRI>` so that OpenAPI can give different types/shapes depending on the shortName, or maybe some has other ideas (or maybe OpenAPI can cope with KRIs, I've no idea myself)

The other nitpick is: should it be `/kri/<shortName>/<KRI>` a.k.a. "The KRI HTTP API" and hence prefixed with `/kri`? Super nitpick that one tho!

[...]

PR #14582, opened 2025-10-03
docs(kri-rest): start working on the MADR

slonka:
**PLEASE DO NOT REVIEW YET**

## Supporting documentation

Fix #XX

### DIFF
--- a/docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
+++ b/docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
@@ -1,0 +1,175 @@
+# Add REST Endpoint for KRI-based Resource Lookup
+
+* Status: Accepted
+
+Technical Story: [https://github.com/kumahq/kuma/issues/13882](https://github.com/kumahq/kuma/issues/13882)
+
+## Context and Problem Statement
+
+We want to enable clients (GUI, SDKs, CLI) to fetch any resource using its **Kuma Resource Identifier (KRI)**.
+Currently, clients construct REST paths using mesh and resource name (`/meshes/{mesh}/{resourceType}/{name}`).
+Because of this, it is not possible to simply link between KRI resources: if you have a KRI in one place, you cannot directly navigate to that resource.
+This complicates GUI/frontend development.
+
+A dedicated endpoint for retrieving a resource by its KRI would:
+
+* Provide a canonical API surface for resource lookups.
+* Simplify frontend logic.
+* Ensure KRIs become first-class citizens in the system.
+
+Use cases:
+
+* GUI fetching a resource by clicking a link derived from a KRI.
+* CLI retrieving resources directly by KRI without constructing full paths.
+* SDKs needing stable and consistent access patterns across resource types.
+
+## Decision Drivers
+
+* Ensure stable, predictable API for GUI and SDKs.
+* Keep API design minimal, avoid unnecessary complexity.
+* Maintain performance (avoid costly label lookups).
+* Avoid breaking existing clients; additive change only.
+* Align with OpenAPI/Typescript code generation while balancing developer ergonomics.
+
+## Alternatives Considered
+
+### Backend
+
+#### Option A: Parse KRI and reuse existing store logic (not possible)
+
+**Idea**: Backend parses KRI into `mesh` + `name` and calls existing `store.GetByKey` logic.
+
+* Not possible because the name in KRI is derived using `GetDisplayName` which relies on the `kuma.io/display-name` label.
+
+```go
+return Identifier{
+    ResourceType: resourceType,
+    Mesh:         rm.GetMesh(),
+    Zone:         rm.GetLabels()[mesh_proto.ZoneTag],
+    Namespace:    rm.GetLabels()[mesh_proto.KubeNamespaceTag],
+    Name:         core_model.GetDisplayName(rm),
+}
+```
+
+[source](https://github.com/kumahq/kuma/blob/86e7375e52abbe933f3f9d2d99dc09f76e0cb1a8/pkg/core/kri/kri.go#L54-L60)
+
+DisplayName is a standard label used to recognize policy names without namespace/hash suffixes.
+
+```go
+// prefer display name as it's more predictable
+if labels := rm.GetLabels(); labels != nil && labels[mesh_proto.DisplayName] != "" {
+    return labels[mesh_proto.DisplayName]
+}
+return rm.GetName()
+```
+
+[source](https://github.com/kumahq/kuma/blob/9e9ad8aadf73f240763c30174cfed7ea7ef416eb/pkg/core/resources/model/resource.go#L502-L510)
+
+#### Option B: Introduce hash/index-based lookup
+
+**Idea**: Backend would fetch resources and filter them by labels extending the label filtering logic [we have now](https://github.com/kumahq/kuma/blob/39cd2e670bbb56a0134fcdfce15c46cdf0a308eb/pkg/api-server/filters/filtering.go#L31). A mapping from KRI to resources would then be stored in a hash map to optimize performance.
+
+**Pros**
+
+* Potential performance gain.
+* Avoid repeated expensive label lookups.
+
+**Cons**
+
+* Adds complexity and new indexing logic.
+
+### Frontend & OpenAPI
+
+#### Option A: Single endpoint `/_kri/{kri}` with runtime type guards (chosen)
+
+**Idea**: Provide a single generic endpoint for all resources.
+Clients rely on `type: <ResourceType>` inside the response and infer type at runtime using type guards. OpenAPI schema describes the endpoint generically with `oneOf`s.
+
+**Pros**
+
+* Minimal API surface.
+* Pure KRI-driven design, consistent with intent of KRIs.
+* Flexible for new resource types without adding new endpoints.
+* Frontend can infer type using runtime type guards.
+
+**Cons**
+
+* Weaker compile-time safety compared to OpenAPI  typing.
+* Requires additional runtime type guard implementation in frontend.
+* Slightly more effort for GUI developers to validate resource shape
+
+#### Option B: Typed endpoint with `shortName` segment (not chosen)
+
+**Idea**: Use a separate endpoint per resource type including `shortName` in the path, e.g., `/_kri/{shortName}/{kri}`.
+This would allow OpenAPI/Typescript SDKs to generate more precise types for each resource.
+
+**Pros**
+
+* Stronger typing in generated SDKs.
+* Frontend can leverage OpenAPI type information directly.
+
+**Cons**
+
+* Increased API surface.
+* Redundant endpoints for each resource type.
+* More maintenance overhead as new resource types are added..
+
+## Design
+
+We introduce a new endpoint:
+
+```yaml
+/_kri/{kri}:
+  get:
+    operationId: getByKri
+    summary: Returns resource by KRI
+    tags: [ "KRI" ]
+    parameters:
+      - in: path
+        name: kri
+        schema:
+          type: string
+        required: true
+        description: KRI of the resource
+    responses:
+      '200':
+        description: The resource
+        content:
+          application/json:
+            schema:
+              oneOf: ... # TODO: add schema
+      '400':
+        $ref: "/specs/base/specs/common/error_schema.yaml#/components/responses/BadRequest"
+      '404':
+        $ref: "/specs/base/specs/common/error_schema.yaml#/components/responses/NotFound"
+```
+
+## Reliability implications
+
+* Endpoint is additive.
+* No changes to existing flows.
+* Runtime type guard approach ensures frontend object mapping is correct.
+
+## Implications for Kong Mesh
+
+* None
+
+## Decision
+
+We choose "Option A: Single endpoint `_kri/{kri}`"
+This keeps the API minimal.
+It avoids redundant typed endpoints.
+It lets the frontend enforce typing at runtime.
+Backend parses the KRI and uses existing store lookup routines.
+SectionName is stripped as needed.
+
+This approach ensures KRIs are first-class in the system.
+It avoids adding unnecessary complexity to the API surface.
+
+## Notes
+
+* Typed endpoints were considered but rejected due to added complexity and redundancy.
+
+* Scope is limited to single-resource fetch.
+
+* Frontend will prototype type guard approach to validate developer experience.


docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,175 @@
...
+
+**Idea**: Provide a single generic endpoint for all resources.
+Clients rely on `type: <ResourceType>` inside the response and infer type at runtime using type guards. OpenAPI schema describes the endpoint generically with `oneOf`s.
+
+**Pros**
+
+* Minimal API surface.
+* Pure KRI-driven design, consistent with intent of KRIs.
+* Flexible for new resource types without adding new endpoints.
+* Frontend can infer type using runtime type guards.
```
johncowen: Assuming by `type` you mean "Typescript type", then this is incorrect. We cannot **infer** a Typescript type at runtime. Typescript is a build-time language.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Alternatives Considered
+
+### Backend
+
+#### Option A: Parse KRI and reuse existing store logic (not possible)
+
+**Idea**: Backend parses KRI into `mesh` + `name` and calls existing `store.GetByKey` logic.
+
+* Not possible because the name in KRI is derived using `GetDisplayName` which relies on the `kuma.io/display-name` label.
```
johncowen: If you are _parsing_ a KRI I don't understand why this has anything to do with display-name specifically. 

KRIs have a straightforwards `_` separator and ordered parameters i.e. `kri_<type/shortName>_<mesh>_<zone>_<ns>_<name>_<sectionName>`

Is this a reference to the fact that currently resources are stored using a `<name>-<black-box-hash-containing-at-least-zone-and-possibly-other-things>`? Therefore you can't only use the <name> part of the parsed KRI, you need `<name>` plus `<hash>`.

I'm just guessing here as to what the problem could be. Let me know!

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
johncowen: Just an additional note here, the only way we can add Typescript types to a very generic HTTP API (with a large amount of differently shaped responses) is for us to add our own proxy layer to these requests i.e.

```ts
`/kri/msvc/:kri`: () => {
  return http.GET(`/_kri/:kri`) as Promise<MeshService>
},
`/kri/dp/:kri`: () => {
  return http.GET(`/_kri/:kri`) as Promise<DataPlane>
}
// In our GUI we call

<DataLoader src="/kri/dp/:kri" />
<DataLoader src="/kri/msvc/:kri" />

```

Note: we are smoothing over the sharp edges here by adding the resource URL segment ourselves and providing the type based on that. If this very generic option is chosen, this is the solution we will go with i.e. add the extra type segment ourselves.

---

Typeguards themselves would add additional layer of runtime validation, but I'd argue that this validation is unnecessary, and in order to use typeguard you still have to manually specify/cast the type. **This is not automatic inference.**

Not only that but we need a large amount of small functions (the typeguards themselves) that do nothing more than validate something the backend is (hopefully) already guaranteeing/validating.

The following is kinda-pseudo code, but hopefully it helps explain:

```ts

const isMeshService = (item: MeshService) is MeshService => item.type === "MeshService"
const isDataplane = (item: Dataplane) is Dataplane => item.type === "Dataplane"

///...
`/kri/msvc/:kri`: () => {
  const res = await http.GET(`/_kri/:kri`) as Promise<MeshService>

  // this is pointless if we have a guarantee from the backend
  // that if we ask for a msvc we will always receive a MeshService
  if(isMeshService(res)) {
    return res;
  } else {
    throw new Error('I would be very surprised if this error ever gets thrown')
  }
 
},
`/kri/dp/:kri`: () => {
  const res = await http.GET(`/_kri/:kri`) as Promise<Dataplane>

  // pointless if we have guarantees from the backend
  if(isDataplane(res)) {
    return res;
  } else {
    throw new Error('I would be very surprised if this error ever gets thrown')
  }
}
```

Note we cannot do dynamic inference using a "big switch statement" as suggested in earlier conversations because we end up with a big union i.e. **you cannot use runtime HTTP responses to automatically infer build-time types**.

The following will return a type of `MeshService | Dataplane`, which is no use to us and I guess the same thing as what using OpenAPI `oneOf` will give us

```ts
`/kri/:kri`: () => {
  const res = await http.GET(`/_kri/:kri`)
  switch(true) {
     case isDataPlane(res):
        return res
      case isMeshService(res):
        return res
      
   }
}
```

johncowen:
I thought of another option:

TDLR; `/_kri/:kri` _and_ `/_kri/:shortName/:kri` endpoints.

Don't want/need typing, fine use `/_kri/:kri`.
Want/need typing, fine use `/_kri/:shortName/:kri`

lahabana:
The title of the PR must be `docs(MADR)`

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: Please check out the repo and this part in particular: https://github.com/slonka/ts-oapi-type-inference/blob/c0ec7d7634a762c6737b7376b3597d3e0c8c2751/main.ts#L28-L32 

would something like that work?

johncowen: Just adding a x-ref here https://github.com/kumahq/kuma/issues/13882#issuecomment-3370749724

TLDR; There is no issue here really, we just continue to do what we always do: use OpenAPI, the only difference is the params used in OpenAPI aren't split by `/` they are instead split by `_`

johncowen:
I've added a commit on the end of here after chatting with @slonka 08926fdb6fb756bc9493290f5ddfe943a24abe2d, importantly there is a new Frontend Option C which is now our chosen approach. My commit is very much a "rough draft addition" and will probably need refining during review/together.

It adds two things:

1. Continuing to use OpenAPI for specifying the API (instead of both OpenAPI and Typescript).
2. Includes both a set of automatically generated specs/endpoints for static build time usage, plus a generic endpoint backed by a very wide shape/type for more dynamic/escape-hatch usage.

In regards to point 2, not only does this give consumers who don't care about types a simple approach to retrieving KRI based records, I believe we (GUI) will use/need this for requesting policy data in the frontend (theoretically specific policy types are only known at runtime)

I think the things we should be clear on by now are:

1. We keep specification in OpenAPI
2. We statically specify the resourceTypes/shortNames we know about.
5. We keep an "escape hatch specification" for resourceTypes/shortNames we don't know about.

More specific implementation questions/unknowns remain:

- `/_kri/kri_msvc_{data}` vs `/_kri/kri_msvc_{mesh}_{zone}_{ns}_{name}_{sectionName}`
  - The latter feels like a safer approach but how would optional KRI segments work? Are optional path params possible in OpenAPI?
- `go` implementation / `operationId`: Use `get{.ShortName}ByKri` or is `getByKri(shortName, ...rest)` or better just `getByKri(kri)` possible?

If anything isn't clear please ask.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: > Note we cannot do dynamic inference using a "big switch statement" as suggested in earlier conversations because we end up with a big union i.e. you cannot use runtime HTTP responses to automatically infer build-time types.

In order to no end up with a big union we'd need another path param which could also be dynamic, i.e. `/kri/:resourceType/:kri`. Probably also with a generic param i.e. `/kri/:resourceType/:kri: <T>(params) => Promise<T>` 🤔

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -114,15 +114,44 @@ This would allow OpenAPI/Typescript SDKs to generate more precise types for each
...
 * More maintenance overhead as new resource types are added..
 
+#### Option C: Single endpoint with multiple statically defined variants (chosen)
+
+**Idea**: Provide a single endpoint rooted under `/_kri` but provide statically defined "variants" for all resources based on the KRI shortname.
+
+- `/_kri/kri_msvc_{...remainingKRI}`
+- `/_kri/kri_zi_{...remainingKRI}`
+
+Note: this option differs to Option A in that we are statically specifying _each resource type_ in OpenAPI instead of relying on a single generic specification (`/_kri/{kri}`) for very differently shaped responses. The key difference here is that we are using `_` separated segments in the KRI instead of `/` separated segments in the URL.
```
schogges: Did we try if this works when the provided `kri` is of type `string`? From a typing perspective within the API layer this is great, but the actual usage is not clear to me yet 🤔

johncowen: > Did we try if this works when the provided kri is of type string?

Not sure if I follow the question?

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
johncowen: > Probably also with a generic param i.e. `/kri/:resourceType/:kri: <T>(params) => Promise<T>` 🤔

In this case how would `T` be defined?

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: In the templates we just provide the kri as type `string` when calling the sources via `DataLoader`. But I guess we just keep doing this and parse the KRI in the API layer?

```typescript
    'kri/mi/:kri': async (params) => {
      const { kri } = params
      const { mesh = '', zone = '', namespace = '', name = '', sectionName = '' } = Kri.fromString(kri)
      const res = await http.GET(`/_kri/kri_mi_{mesh}_{zone}_{ns}_{name}_{sectionName}`, {
        params: {
          path: {
            mesh,
            zone,
            ns: namespace,
            name,
            sectionName,
          },
        },
      })
      return MeshIdentity.fromObject(res.data!)
    }
```

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: I tried and failed, so this is not an option.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
johncowen: Yep exactly, for MeshIdentity (and other static/non-pluggable resources) at least anyway.

For policies we don't always know each specific shortName, so this is where we use the "escape hatch"

johncowen:
```typescript
const res = await http.GET(`/_kri/kri_mi_{mesh}_{zone}_{ns}_{name}_{sectionName}`, {
  params: { path: KRI.parse(params.kri) },
})
```

Bit of a nit at this point, but stylistically I'd probably try to do something like this to get rid of as much boilerplate as possible.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -114,15 +114,44 @@ This would allow OpenAPI/Typescript SDKs to generate more precise types for each
...
+* Pure KRI-driven design, consistent with intent of KRIs.
+* Stronger typing in generated SDKs.
+* Frontend can leverage OpenAPI type information directly.
+* Uses existing standards and application patterns.
+* Maintains existing engineering boundaries
+* Static escape hatch for dynamic retrieval (in the case of policies), exchanging dynamism for a less narrow type.
+
+**Cons**
+
+* More maintenance overhead as new resource types are added (mostly offset by the fact that these specifications are automatically generated)
```
slonka: By that you mean you have to manually get the shortname?

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: > Is this a reference to the fact that currently resources are stored using a <name>-<black-box-hash-containing-at-least-zone-and-possibly-other-things>? Therefore you can't only use the part of the parsed KRI, you need <name> plus <hash>.

More or less yes but apparently Ilya has some way of working around this. I'll sync up with him about this today.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
johncowen: First first part of this sentence I took from your original MADR

> More maintenance overhead as new resource types are added

Which I assumed refers to the fact that instead of just one OpenAPI definition for `/_kri/:kri` we need one per resource `/_kri/kri_msvc_{... etc}`, `/_kri/kri_zi_{... etc}`.

The second part of the sentence in brackets I added because even though we will have multiple endpoints in the OpenAPI schema that essentially do the same thing (but specify different shapes/types), these multiple endpoints are automatically generated (from what I understand?), so this point about "maintenance overhead" isn't really an issue/con.

Does that help explain?

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Design
+
+We introduce new automatically generated endpoints for each new resource (note: gotemplate variables begin with `.`, OpenAPI params don't):
+
+```yaml
+/_kri/kri_{.ShortName}_{mesh}_{zone}_{ns}_{name}_{sectionName}:
```
schogges: I think it would be good to stick to one naming pattern, i.e. `ns` vs `namespace`. Looking at previous madrs ([070](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/070-resource-identifier.md#what-characters-we-can-use-as-a-delimiter), [083](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/083-finalized-schemes-for-unified-envoy-resources-and-stats-naming-formats.md#shape)) we usually call name it `namespace`. In the case of `resourceType` I've seen it being named `shortName` too (i.e. `/_resources` endpoint).
Especially for param names it makes sense to stay consistent.

johncowen: I changed `ns` to `namespace`

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,203 @@
...
+    operationId: get{.Shortname}ByKri
+    summary: Returns a {.ShortName} resource by KRI
+    tags: [ "KRI" ]
+    parameters:
+      - in: path
+        name: kri
+        schema:
+          type: string
+        required: true
+        description: KRI of the resource
```
schogges: We have to specify each parameter separately and can't use one `kri` parameter. Also I think we need to have support for the `format` query parameter on each of these endpoints.

```suggestion
    parameters:
      - in: path
        name: mesh
        schema:
          type: string
        required: true
        description: <mesh> segment of the KRI
      - in: path
        name: zone
        schema:
          type: string
        required: true
        description: <zone> segment of the KRI
      - in: path
        name: namespace
        schema:
          type: string
        required: true
        description: <namespace> segment of the KRI
      - in: path
        name: name
        schema:
          type: string
        required: true
        description: <name> segment of the KRI
      - in: path
        name: sectionName
        schema:
          type: string
        required: true
        description: <sectionName> segment of the KRI
```

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: This can be implemented as one function. Takes a KRI and returns a [ResourceKey](https://github.com/kumahq/kuma/blob/9e9ad8aadf73f240763c30174cfed7ea7ef416eb/pkg/core/resources/model/resource.go#L42-L45).

1. Figure out if resource originated on this CP or not. This can be done via a function like "IsLocallyOriginated"
2. If it's locally originated you just need to adjust for the k8s store because on k8s we have CoreName (which is `${name}.${namespace}`). If it's universal that's not the case.
3. If it's not locally originated you need to compute the hash from the data from KRI. https://github.com/kumahq/kuma/blob/c989d3d842850aa468248e76298b9729af217a3b/pkg/kds/context/context.go#L232 . Here you also need to adjust for CoreName.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
johncowen: Oh totally, but I guess thats down to when this gets implemented. I just didn't add these in the MADR for succinctness.

Also I think one of my "unknowns" was whether we express each individual param in the KRI or we just use a more `kri_{.ShortName}_{...rest}` approach. Although the more i think about it my preference would be to express each individual param.

I guess the other thing that is in my unknowns is that these params can't be `required`, well better said we need to be able to leave them empty (i.e. `''`).

I'm not totally sure whether we can express that exactly in OpenAPI path params, I'm not totally sure whether `required: true` means you have to specify something non-empty. Would be good to check that because it might influence what we do.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: I think I just wanted to write something else.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: I think in OAS using `{...rest}` doesn't work and we have to express each individual segment. I've tried it out with https://editor.swagger.io.

> I guess the other thing that is in my unknowns is that these params can't be required, well better said we need to be able to leave them empty (i.e. '').

Good point, I haven't tried that. Probably we'll do not require any of these 👍

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -35,52 +35,63 @@ Use cases:
...
 
 ### Frontend & OpenAPI
 
-#### Option A: Single endpoint `/_kri/{kri}` with runtime type guards
+#### Option A: Single endpoint with multiple statically defined variants (chosen)
+
+**Idea**: Provide a single endpoint rooted under `/_kri` but provide statically defined "variants" for all resources based on the KRI shortname.
+
+- `/_kri/kri_msvc_{...remainingKRI}`
+- `/_kri/kri_zi_{...remainingKRI}`
```
johncowen: To @schogges point, I guess we should fill these in with actual params so its clear thats what we are doing. We can do this during sync I think just in case we wanted to chat over it.

johncowen: actually lemme do that now real quick

johncowen: note: we need to check that folks can provide empty values for these path params.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
johncowen: > I think in OAS using {...rest} doesn't work

Oh sorry I should have explained better

The other option is:

```
/_kri/kri_msvc_{rest}
```

So then the implementation of the spec would have to manually explode `rest` into our KRI params. If OpenAPI won't let us express empty path params then this is still an option.

slonka: @johncowen are you worried about FE or BE now? Can you check on your end and I'll check on ours?

slonka: Seems like online swagger editor  doesn't like that:

```
openapi: 3.0.0
info:
  title: KRI API
  version: 1.0.0

paths:
  /_kri/kri_mtp_{mesh}_{zone}_{namespace}_{name}_{sectionName}:
    get:
      operationId: getMtpByKri
      summary: Returns a MTP resource by KRI
      tags: [ "KRI" ]
      parameters:
        - in: path
          name: mesh
          schema:
            type: string
          required: true
          description: KRI of the resource
        - in: path
          name: zone
          schema:
            type: string
          required: true
          description: KRI of the resource
        - in: path
          name: namespace
          schema:
            type: string
          required: true
          description: KRI of the resource
        - in: path
          name: name
          schema:
            type: string
          required: true
          description: KRI of the resource
        - in: path
          name: sectionName
          schema:
            type: string
          required: true
          description: KRI of the resource
      responses:
        '200':
          description: The resource
          content:
            application/json:
              schema:
                type: string    
```

image

johncowen: > Can you check on your end and I'll check on ours?

I checked with the example I'd used earlier, using `required: true` in the OpenAPI means you have to set every param but you _can_ use empty string `''` in the implementation, so I think this is fine.

Suggestion: I would consider adding the fact that even though these are `required` you can send (and its expected) empty strings for these params in the OpenAPI descriptions

johncowen: P.S. I could see your screengrab of the swagger editor, but I tried it here myself and it seemed to validate fine? I guess I'm misunderstanding something lmk!

schogges: @slonka I can't see the screengrab and when I tried with online swagger editor it was working fine 😅

slonka: @schogges - please check the screengrab again

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: we have a whole section about this now

johncowen:
@slonka I still need to add something around what we spoke about, but I just saw your update from our chat and whilst the non-empty parameters thing presents downsides a similar approach with `/_kri/kri_{.ShortName}_{ partialKRI }` is a potential option I believe? There are Pros and Cons to this also I believe, but its an option still right?

johncowen:
@lahabana we gone back and forth a little here on several options, I'd be interested to know what your opinion is of this:

https://github.com/kumahq/kuma/pull/14582#issuecomment-3365209511

> I thought of another option:

> TDLR; /_kri/:kri _and_ /_kri/:shortName/:kri endpoints.

> Don't want/need typing, fine use /_kri/:kri.
> Want/need typing, fine use /_kri/:shortName/:kri

I guess if you aren't keen on this, it would be good to get the arguments behind _not_ doing this so we can include those in the MADR. ty!

slonka:
> /_kri/kri_{.ShortName}_{ partialKRI } is a potential option I believe? There are Pros and Cons to this also I believe, but its an option still right?

I think so... but it's the most awkward one. Would you be so kind and add it?

schogges:
> @slonka I still need to add something around what we spoke about, but I just saw your update from our chat and whilst the non-empty parameters thing presents downsides a similar approach with `/_kri/kri_{.ShortName}_{ partialKRI }` is a potential option I believe? There are Pros and Cons to this also I believe, but its an option still right?

It is an option, but I wouldn't go for it. It would require to keep the format of the `partialKRI` (order and separator), which would need extra work and patience depending on the user.

bartsmykla (approved):

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,259 @@
...
+
+* Minimal API surface for consumers.
+* Pure KRI-driven design, consistent with intent of KRIs.
+* Flexible for new resource types without adding new endpoints.
+
+**Cons**
+
+* Weaker compile-time safety compared to OpenAPI typing.
+* Requires additional runtime type guard/casting implementation in frontend.
+* Slightly more effort for GUI developers to validate resource shape.
```
lahabana: This is synonymous of the previous point no?

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,259 @@
...
+
+A dedicated endpoint for retrieving a resource by its KRI would:
+
+* Provide a canonical API surface for resource lookups.
+* Simplify frontend logic.
+* Ensure KRIs become first-class citizens in the system.
+
+Use cases:
+
+* GUI fetching a resource by clicking a link derived from a KRI.
```
lahabana, line 22: Wouldn't the gui need the actual complete url to be able to build the route? Or will all the gui routes exist twice now (with and without kri)

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,259 @@
...
+* CLI retrieving resources directly by KRI without constructing full paths.
+* SDKs needing stable and consistent access patterns across resource types.
```
lahabana, line 24: I think the reverse use case would be useful.
I want to retrive by KRI regardless of types and figure the type at runtime. This is impossible atm and I think your design doesn't enable this either.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,259 @@
...
+
+**Pros**
+
+* Minimal API surface for consumers.
+* Pure KRI-driven design, consistent with intent of KRIs.
+* Flexible for new resource types without adding new endpoints.
+
+**Cons**
+
+* Weaker compile-time safety compared to OpenAPI typing.
```
lahabana, line 82: Actually not really there's a oneOf with the type right?
So this means that modulo bugs the oneOf option will match the KRI so you get static strong typing

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: We would keep the current URL/routing structure and use the KRI as a dynamic parameter in one segment, which we then use to fetch the API. In most cases we know the resource type and can route easily. In some cases we might need to parse the KRI.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: That will be possible with the generic `/_kri/{kri}` endpoint we are planning to add.

>Don't want/need typing, fine use /_kri/:kri.
Want/need typing, fine use /_kri/:shortName/:kri.

Basically the first of the two variants ☝️

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: The problem with the generic endpoint is that the `oneOf` ends up as a big massive union type, which is fine for some use cases. In the GUI at some point we'd need to narrow it down for which we'd use the same pattern of `/_kri/:shortName/:kri` and implement this in a proxy layer. We prefer to have this on the API so not every user of the API needs to do that.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: I think you are right, maybe we can merge the two points? 😄

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,296 @@
...
+* Requires additional runtime type guard/casting implementation in frontend.
+* Slightly more effort for GUI developers to validate resource shape.
+* Moves API specification responsibilities from OpenAPI to Typescript against existing boundaries.
+
+#### Option D: Typed endpoint with `shortName` segment (chosen)
+
+**Idea**: Use a separate endpoint per resource type including `shortName` in the path, e.g., `/_kri/{shortName}/{kri}`.
+This would allow OpenAPI/Typescript SDKs to generate more precise types for each resource.
```
slonka:
```suggestion
This would allow OpenAPI/Typescript SDKs to generate more precise types for each resource.
This can also be hidden behind `x-internal` so it's not published as a public spec.
```

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,297 @@
...
+
+* Minimal API surface for consumers.
+* Pure KRI-driven design, consistent with intent of KRIs.
+* Flexible for new resource types without adding new endpoints.
+
+**Cons**
+
+* Weaker compile-time safety compared to OpenAPI typing.
+* Requires additional runtime type guard/casting implementation in frontend.
+* Slightly more effort for GUI developers to validate resource shape.
```
schogges: ref: https://github.com/kumahq/kuma/pull/14582#discussion_r2415866689

```suggestion
* Requires additional effort of runtime type guard/casting implementation to validate the shape of the requested resource.
```

johncowen:
> Did I get correctly that we're implementing 2 endpoints now?

Almost, you understood correctly that we are suggesting having _more than one endpoint_.

But to be super clear: it's _one endpoint per resource_, plus a generic one i.e:

1. Several of the form `/_kri/{.ShortName}/:kri` (each automatically generated from a single template) which provides types.
2. One generic endpoint of the form `/_kri/:kri`, This can provide a very very (almost pointless) generic type.

The first typed endpoint is for folks that prefer a typed specification for their OpenAPI/HTTP API, which I would imagine this is the vast majority of engineers/our customers, but I could be wrong.

The second generic endpoint are for folks or occasions where the type either isn't important to the person/customer/engineer using the API (I honestly can't imagine an engineer not wanting types nowadays) _or_ whatever it is you are doing does not know the shortName at runtime.

The best example I can give for this second case is if you are requesting a policy but do not know the actual kind of policy at build-time. We have this use-case ourselves in the GUI.


---

All in all it sounds like everyone is onboard with a `/_kri/:kri` generic endpoint, so maybe we say that in the MADR and go with that initially i.e. find where we agree and move forwards with that.

We can continue the discussion on how we design/implement the build-time types as a follow up.

Personally, as someone who is probably the primary user of this API, the person who requested it and one of the people who has put a lot of thought and investigation into this MADR along with others, I would really like to understand your argument over not wanting an autogenerated set of typed endpoints. I need to be convinced and at the moment I'm not. That being said, totally happy to be convinced!

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
lahabana: So are we talking about changing what already exists?

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
lahabana: Agreed but I think this is still a use case we should add here

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
lahabana: > In the GUI at some point we'd need to narrow it down for which we'd use the same pattern of /_kri/:shortName/:kri and implement this in a proxy layer. We prefer to have this on the API so not every user of the API needs to do that.

That's the bit where I'm lost. To me there are 2 cases for using this API:

1. you call the API and you know the type you expect (.e.g you are on the MTP route, you know you are getting an MTP here) so the fact that it's a union type doesn't matter (fwiw you could just cast and that would 100% work as well).
2. you have a KRI but you have no idea what's behind it. So here you have 2 choice resolve the type before (look at the shortName in the KRI) or after using the union type. They technically do the same thing right?

Are there other use-cases here?

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,296 @@
...
+
+#### Endpoint handling
+
+We will add the new endpoint next to the existing endpoints in [addFindEndpoint](https://github.com/kumahq/kuma/blob/4004a7231090fb2786d5ede41b3d95b73188d745/pkg/api-server/service_insight_endpoints.go#L28-L32).
+
+## Reliability implications
+
+* Endpoint is additive.
+* No changes to existing flows.
+* Buildtime static specification approach ensures frontend object mapping is correct.
```
lahabana: I'd also add the Name computation is a pure function so no DB load

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: Ah, we've added this above the options as something that goes along with any of the options we are choosing: https://github.com/kumahq/kuma/pull/14582/files#diff-e7b8d7567d88a88336f9f79474a2bf019fbb2122a80f05d0bf2b9ad32cf9078dR67-R69

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: The problematic bit is that at certain points we can't deal with unions anymore and require that the type is narrowed down to exactly one type. In order to do that we can use type guards and also type cast it to ensure it's the expected type. We were experimenting with two approaches:

1. One handler that handles all requests to `/_kri/:kri` and use type guards: This would be a very big handler and the result would still be a big massive union. Any time we are dealing with unions we have to narrow down the type before we can process the data. Therefore we'd end up with type guards/type casting in the proxy-layer and in the views. We've quickly decided that we want to have individual handlers in the proxy-layer.
2. Use individual handlers for each resource type: Each of these handlers returns one type, which is great. But as `/_kri/:kri` is generic and is typed as a union we'd still have to do the type casting. We can do this, but it feels like we are moving the specs from OpenAPI (build-time) to our proxy-layer (run-time). And if we are doing it, most likely other users who need type safety would do the same. The UX is just not as good as with individual endpoints that are returning a resource of just one type.

I've tried to visualize the data-flow in the GUI and I hope it helps to understand why and where we need the type to be narrowed down:

One generic endpoint and one generic proxy handler:
image

We can't use unions in the template/route and in the data-layer. In both places we are processing the data and therefore it needs to be narrowed down to one type.

---

One generic endpoint and individual proxy handlers for each resource type:
image

In the template we would use the type of the processed data as we have handlers for individual resource-types.

---

Individual endpoints and individual proxy handlers for each resource type:
image

This would be preferred way as at each point we know the type which is all based on OAS, no type casting needed at all and type safety at build time.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
schogges: I think the general layout of the URLs will stay the same. In the places where we link from a KRI to a resource we need to check the KRI and map it to a view name. In the respective route we have to check wether the parameter is a KRI or the old style (i.e. name) to make the call. I think until we have KRIs everywhere we have to support both. At some point we are able to use KRIs only.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
johncowen: Just to follow up on/second @schogges 's points.

> you call the API and you know the type you expect (.e.g you are on the MTP route, you know you are getting an MTP here) so the fact that it's a union type doesn't matter (fwiw you could just cast and that would 100% work as well).

Yes we could totally cast, but as @schogges says:

> We can do this, but it feels like we are moving the specs from OpenAPI (build-time) to our proxy-layer (run-time). And if we are doing it, most likely other users who need type safety would do the same.

OpenAPI is for specifying that if you request a certain URL you get a certain response/shape back i.e. "The Specification"

If we cast/type guard, we have moved the specification of the response/shape to Typescript, for one client (us). "The Specification" is now split across OpenAPI and Typescript.

This is not ideal, but possible. We want to specify the API (including the response/shape/type) only in the OpenAPI specification for _all_ clients, like we do for our other endpoints.

---

> you have a KRI but you have no idea what's behind it. So here you have 2 choice resolve the type before (look at the shortName in the KRI) or after using the union type. They technically do the same thing right?

> look at the shortName

We can only do this once we've received the response, or someone's entered that into the URL right? i.e. at runtime.

Typescript is a build-time language. If we have to do something at runtime its too late for Typescript.

We could check `type: Dataplane`, or `kri_{shortName}` at runtime, but then at build time everything is `any`/`unknown` or a big union.

Maybe its worthwhile underlying that what we want/are discussing here are build-time Typescript types for a better developer experience and safety. Not runtime validation.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,296 @@
...
+We introduce new automatically generated endpoints for each new resource (note: gotemplate variables begin with `.`, OpenAPI params don't).
+These endpoints will be next to existing endpoints in [endpoints.yaml](https://github.com/kumahq/kuma/blob/46b6807f56c08c79f34db20cfc452b8f5203bf90/tools/openapi/templates/endpoints.yaml#L9).
+
+```yaml
+openapi: 3.0.0
+info:
+  title: KRI API
+  version: 1.0.0
+paths:
+  /_kri/kri_mtp_{mesh}_{zone}_{namespace}_{name}_{sectionName}:
```
lobkovilya: Why these variables are in the path? I think `kri` should be opaque to API users

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,296 @@
...
+
+* Weaker compile-time safety compared to OpenAPI typing.
+* Requires additional effort of runtime type guard/casting implementation to validate the shape of the requested resource.
+* Moves API specification responsibilities from OpenAPI to Typescript against existing boundaries.
+
+#### Option D: Typed endpoint with `shortName` segment (chosen)
+
+**Idea**: Use a separate endpoint per resource type including `shortName` in the path, e.g., `/_kri/{shortName}/{kri}`.
+This would allow OpenAPI/Typescript SDKs to generate more precise types for each resource.
+This can also be hidden behind `x-internal` so it's not published as a public spec.
```
lobkovilya: I'd not go with `x-internal` regardless of what way we choose. IMO it doesn't bring anything useful. It tries to hide the fact we have 2 endpoints that do the same thing, but the presence of an internal endpoint is just as much a sign of less-than-ideal API design as having 2 endpoints that do the same thing, if not worse.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,296 @@
...
+* No changes to existing flows.
+* Buildtime static specification approach ensures frontend object mapping is correct.
+
+## Implications for Kong Mesh
+
+* None
+
+## Decision
+
+We choose "Option D: Typed endpoint with `shortName` segment", along with a generic `/_kri/:kri` endpoint.
```
lobkovilya: @schogges does GUI actually have a usecase for the generic endpoint? Or endpoint with `shortName` segment cover 100% of your usecases?

schogges: The only use case I can imagine at the moment is when we just want to display the response as is without processing/altering, i.e. when showing the response in yaml format.

johncowen: @lobkovilya and I, spoke offline about this and just to add here for completeness: we need to use the generic one for policies (as opposed to targetable resource aka non-policy resources)

schogges: It seems like we'd also use it for generic policy resources

Edit: sorry, it seems my comment wasn't submitted - ignore me

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: This is no longer the scope of this MADR, if we need we will have a separate MADR where this option will be discussed.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: This is no longer the scope of this MADR, if we need we will have a separate MADR where this option will be discussed.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: This is no longer the scope of this MADR, if we need we will have a separate MADR where this option will be discussed.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: This is no longer the scope of this MADR, if we need we will have a separate MADR where this option will be discussed.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: This is no longer the scope of this MADR, if we need we will have a separate MADR where this option will be discussed.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: Now it's just one option so 👍

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
slonka: This is a bit controversial on what "weaker" means so let's move this conversation to the second MADR that we mention.

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,154 @@
...
+      responses:
+        '200':
+          description: The resource
+          content:
+            application/json:
+              schema: 
+                oneOf:
+                  - $ref: '#/components/schemas/MeshAccessLog'
+                  - $ref: '#/components/schemas/MeshTrafficPermission'
+                  - $ref: '#/components/schemas/MeshTrace'
```
slonka, line 117:
```suggestion
                  - $ref: '#/components/schemas/MeshTrace'
              discriminator:
                propertyName: type
              mapping: # maybe not needed because name of type == name of schema
                MeshTrafficPermission: '#/components/schemas/MeshTrafficPermission'
```

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,158 @@
...
+              discriminator:
+                propertyName: type
+              mapping: # maybe not needed because name of type == name of schema
+                MeshTrafficPermission: '#/components/schemas/MeshTrafficPermission'
+                  # ... all other resource types
+        '400':
```
slonka:
```suggestion
        '400': # only invalid KRI triggers a 400
```

docs/madr/decisions/086-endpoint-to-retrieve-a-resource-by-kri.md
```diff
@@ -0,0 +1,158 @@
...
+          $ref: "/specs/base/specs/common/error_schema.yaml#/components/responses/BadRequest"
+        '404':
```
slonka:
```suggestion
        '404': # this is triggered if the KRI is valid but the resource is not found
```

PR #14596, opened 2025-10-04
ci(automation): finalize active-branches preset and workflow hardening

bartsmykla:
## Motivation

We need a single source of truth for active branches and safer use of that list across workflows. Renovate behavior on release branches should favor stability while still letting OSV alerts and Envoy patch updates through. We also must keep PR titles stable because extra suffixes break our release notes scripts.

## Implementation information

- Replaced the array `active-branches.json` with an object that holds `baseBranchPatterns`. Existing consumers continue to work during rollout by parsing either shape with `jq`.
- Updated GitHub Actions workflows and inline scripts to consume the file safely:
  - Fetch via the raw content API, normalize to a clean array, and print readable summaries in the job output.
  - Use `set -euo pipefail` and optional debug tracing. Switch to `GH_TOKEN` usage.
  - Support both old and new shapes during migration with tolerant `jq` filters.
  - In `backport.yaml`, pin to `ubuntu-24.04` and refactor matrix generation to slice by the PR base branch.
- Tightened Renovate:
  - `extends: ["kumahq/kuma:active-branches"]` to apply `baseBranchPatterns`.
  - `prTitleStrict: true` so Renovate does not append branch name suffixes to PR titles.
  - `vulnerabilityAlerts: { "commitMessageSuffix": "" }` so Renovate does not add `[SECURITY]` like suffixes to PR titles.
  - Base branch labels for clarity on non main or master targets.
  - Disable regular updates on `release-*` by default and allow only Envoy patch updates.

I tested the updated workflows on my fork. The jobs resolved the active branches correctly, produced summaries in the job output, and the matrices were generated based on the PR base branch as intended.

## Supporting documentation

This work implements the MADR Change `active-branches.json` Into a Renovate Preset. It is proposed in PR https://github.com/kumahq/kuma/pull/14572 and documented at [`docs/madr/decisions/085-active-branches-renovate-preset.md`](https://github.com/kumahq/kuma/blob/7d7381ece11f7377624f4da873d601055217d199/docs/madr/decisions/085-active-branches-renovate-preset.md).

### Related PRs

- https://github.com/Kong/team-mesh/pull/452
- https://github.com/kumahq/ci-tools/pull/116

## Testing

<details>
<summary>Script showing the <code>jq</code> query from the <code>_get-active-branches.json</code> action works as expected</summary>

```bash
#!/usr/bin/env bash

filtered() {
  jq -c \
    --arg default "${DEFAULT_BRANCH:-$1}" \
    --arg exclude "${EXCLUDE_DEFAULT:-$2}" \
    --arg before "${BEFORE_BRANCH:-$3}" '
      def trim: gsub("^[[:space:]]+|[[:space:]]+$";"");
      def as_array: if type=="array" then . else (.baseBranchPatterns // []) end;
      def uniq_preserve(a): reduce a[] as $i ([]; if index($i) then . else . + [$i] end);
      def sort_key($def):
        if test("^release-[0-9]+\\.[0-9]+$") then
          (capture("^release-(?<maj>[0-9]+)\\.(?<min>[0-9]+)$")
          | [0, (.maj|tonumber), (.min|tonumber), ""])
        elif ($def|length)>0 and . == $def then
          [2, 0, 0, .]      # default branch last
        else
          [1, 0, 0, .]      # other branches between releases and default
        end;

      (try as_array catch [])
      | map(select(type=="string") | trim)
      | map(select(length>0))
      | if ($exclude=="true" and ($default|length)>0) then map(select(. != $default)) else . end
      | uniq_preserve(.)
      | (if ($before|length)>0 and (index($before) != null)
           then .[:index($before)]
           else . end)
      | sort_by(sort_key($default))
    ' <<< "$4"
}

filtered master false '' '{ "baseBranchPatterns": ["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"] }'
# Result: ["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"]
filtered master false '' '["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"]'
# Result: ["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"]
filtered master true '' '{ "baseBranchPatterns": ["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"] }'
# Result: ["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12"]
filtered master true '' '["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"]'
# Result: ["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12"]
filtered master true release-2.10 '{ "baseBranchPatterns": ["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"] }'
# Result: ["release-2.7","release-2.9"]
filtered master true release-2.10 '["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"]'
# Result: ["release-2.7","release-2.9"]
filtered master false release-2.9 '{ "baseBranchPatterns": ["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"] }'
# Result: ["release-2.7"]
filtered master false release-2.9 '["release-2.7","release-2.9","release-2.10","release-2.11","release-2.12","master"]'
# Result: ["release-2.7"]
```
</details>

### DIFF
--- a/.github/workflows/ci-stability-release-branches.yaml
+++ b/.github/workflows/ci-stability-release-branches.yaml
@@ -20,49 +20,15 @@ jobs:
     permissions:
       contents: read
     outputs:
-      branches: ${{ steps.get-branches.outputs.branches }}
+      branches: ${{ steps.active-branches.outputs.out }}
     steps:
       - name: "Get active branches"
-        id: get-branches
-        env:
-          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
-          REPOSITORY: ${{ github.repository }}
-          GH_TOKEN: ${{ github.token }}
-          GH_DEBUG: ${{ runner.debug == '1' }}
-        run: |
-          [[ "${GH_DEBUG}" == "1" ]] && set -x
-          set -euo pipefail
-          
-          # Fetch raw file content to avoid manual base64 decoding
-          if ! content="$(gh api -H 'Accept: application/vnd.github.raw' "repos/${REPOSITORY}/contents/active-branches.json" 2>/dev/null)"; then
-            echo "active-branches.json not found in ${REPOSITORY}"
-            exit 1
-          fi
-          
-          # Support both formats:
-          # New: {"baseBranchPatterns": ["release-...","master"]}
-          # Old: ["release-...","master"]
-          branches="$(echo "$content" | jq -c --arg DEFAULT_BRANCH "$DEFAULT_BRANCH" '
-            try (if type == "array" then . else (.baseBranchPatterns // []) end) catch []
-            | map(select(type == "string"))
-            | map(gsub("^[[:space:]]+|[[:space:]]+$"; ""))
-            | map(select(. != "" and . != $DEFAULT_BRANCH))
-            | sort | unique
-          ')"
-          
-          if [[ -z "${branches}" || "${branches}" == "[]" ]]; then
-            echo "No active branches to process"
-            exit 1
-          fi
-          
-          {
-            echo "Active branches:"
-            echo '```json'
-            echo "${branches}"
-            echo '```'
-          } >> "$GITHUB_STEP_SUMMARY"
-          
-          echo "branches=${branches}" >> "$GITHUB_OUTPUT"
+        id: active-branches
+        uses: ./.github/workflows/_get-active-branches.yaml
+        secrets: inherit
+        with:
+          excludeDefault: true
+          summary: true
 
   trigger-build-test-distribute:
     needs: get-active-release-branches
--- a/.github/workflows/_get-active-branches.yaml
+++ b/.github/workflows/_get-active-branches.yaml
@@ -1,0 +1,136 @@
+on:
+  workflow_call:
+    inputs:
+      excludeDefault:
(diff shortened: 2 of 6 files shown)


.github/workflows/ci-stability-release-branches.yaml
```diff
@@ -27,19 +27,42 @@ jobs:
...
+          # Old: ["release-...","master"]
+          branches="$(echo "$content" | jq -c --arg DEFAULT_BRANCH "$DEFAULT_BRANCH" '
+            try (if type == "array" then . else (.baseBranchPatterns // []) end) catch []
+            | map(select(type == "string"))
+            | map(gsub("^[[:space:]]+|[[:space:]]+$"; ""))
+            | map(select(. != "" and . != $DEFAULT_BRANCH))
+            | sort | unique
+          ')"
+          
+          if [[ -z "${branches}" || "${branches}" == "[]" ]]; then
```
slonka: I know it was copy pasted like this before but since we're already modifying this should we just make it a composite action that outputs branches? So that it's not in 4 places?

PR #14604, opened 2025-10-05
refactor(kds): remove callbacks

lobkovilya:
## Supporting documentation

Fix #XX

### DIFF
--- a/pkg/kds/context/context.go
+++ b/pkg/kds/context/context.go
@@ -30,7 +30,6 @@ import (
 	"github.com/kumahq/kuma/pkg/core/resources/store"
 	"github.com/kumahq/kuma/pkg/kds"
 	"github.com/kumahq/kuma/pkg/kds/hash"
-	"github.com/kumahq/kuma/pkg/kds/mux"
 	"github.com/kumahq/kuma/pkg/kds/service"
 	"github.com/kumahq/kuma/pkg/kds/util"
 	reconcile_v2 "github.com/kumahq/kuma/pkg/kds/v2/reconcile"
@@ -49,7 +48,7 @@ type Context struct {
 	TypesSentByGlobal     []core_model.ResourceType
 	GlobalProvidedFilter  reconcile_v2.ResourceFilter
 	ZoneProvidedFilter    reconcile_v2.ResourceFilter
-	GlobalServerFiltersV2 []mux.FilterV2
+	GlobalServerFiltersV2 []FilterV2
 
 	GlobalResourceMapper reconcile_v2.ResourceMapper
 	ZoneResourceMapper   reconcile_v2.ResourceMapper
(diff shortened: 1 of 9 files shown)


.golangci.yml
```diff
@@ -4,7 +4,6 @@ run:
 linters:
   enable:
     - bodyclose
-    - contextcheck
```
lobkovilya, line 7: This one didn't like the following code
```go
func foo(ctx context.Context, stream Stream) {
  stream := client.GRPCStream(ctx)
  
  go func() {
    doSomething(stream.Context())
  }()
  
  // ...
}
```
TBH I think it's okay to turn off this linter

pkg/kds/mux/error_recorder_stream.go
```diff
@@ -1,4 +1,4 @@
-package server
+package mux
```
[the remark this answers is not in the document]
lobkovilya: Not sure I understand it

pkg/kds/context/context.go
```diff
@@ -60,6 +59,11 @@ type Context struct {
 	CreateZoneOnFirstConnect bool
 }
 
+type FilterV2 interface {
+	InterceptServerStream(stream grpc.ServerStream) error
+	InterceptClientStream(stream grpc.ClientStream) error
+}
+
```
Icarus9913, line 66: I noticed that we don't have any `FilterV2` implementations, so I was wondering if we should discard it directly?

pkg/kds/global/components.go
```diff
@@ -147,37 +85,13 @@ func Setup(rt runtime.Runtime) error {
...
-			rt.KDSContext().GlobalServerFiltersV2,
-			rt.Extensions(),
-			rt.EventBus(),
-			rt.ResourceManager(),
-			rt.Config().Store.Upsert,
-			rt.GetInstanceId(),
+			rt,
+			deltaServer,
+			resourceSyncer,
+			rt.Config().Store.Type == store_config.KubernetesStore,
```
Icarus9913: As you already passed `rt` into the function, so I think you can set the `k8sStore` property within the function. We don't need an extra parameter `k8sStore bool` for this `NewKDSSyncServiceServer` function.

pkg/kds/v2/server/components.go
```diff
@@ -66,7 +67,7 @@ func New(
 		syncTracker,
 		status.DefaultStatusTracker(rt, log),
 	}
-	return NewServer(cache, callbacks, log), nil
+	return delta.NewServer(context.Background(), cache, callbacks, delta.WithDistinctResourceTypes(1000)), nil
```
Icarus9913, line 73: Please supplement the comments to explain the `delta.WithDistinctResourceTypes(1000)` for better tracking previous kds dead lock issue.

```text
	// Default resource types are length of XDS types. With KDS we have much more types.
	// If we don't adjust this value, we can hit KDS deadlock.
	// We could count exactly how many types we have, but overhead of larger map size is negligible for potential mistake here.
```

pkg/kds/context/context.go
lobkovilya: We have filters in Kong Mesh

pkg/kds/v2/server/components.go
lobkovilya: Ah right, restored the comment

lobkovilya:
It's not the first time the test failed, I created an issue https://github.com/kumahq/kuma/issues/14710 (and restart the CI)

PR #14639, opened 2025-10-07
feat(policies): add proto oneOf support to OAS generator

lobkovilya:
## Supporting documentation

Fix #XX


tools/resource-gen/pkg/generator/main.go
```diff
@@ -556,45 +543,191 @@ func openApiGenerator(pkg string, resources []ResourceInfo) error {
...
+			if key == "RSAbits" {
+				return "rsaBits"
+			}
+			return lowerFirst(snakeToCamel(key))
+		},
+		Mapper: func(t reflect.Type) *jsonschema.Schema {
+			s, err := r.mapper(t, withBackendCheck)
+			if err != nil {
+				// log
+				fmt.Printf("error occured during mapping: %v\n", err)
```
lobkovilya:
```suggestion
				fmt.Printf("error occurred during mapping: %v\n", err)
```

tools/resource-gen/pkg/generator/protoreflect.go
```diff
@@ -0,0 +1,108 @@
...
+		return protoreflect.ValueOfFloat64(0), nil
+	case protoreflect.StringKind:
+		return protoreflect.ValueOfString(""), nil
+	case protoreflect.BytesKind:
+		return protoreflect.ValueOfBytes(nil), nil
+	case protoreflect.EnumKind:
+		evs := fd.Enum().Values()
+		var num protoreflect.EnumNumber
+		if evs.Len() > 0 {
+			num = evs.Get(0).Number()
```
slonka, line 97: why for empty enum we get something at 0 index - I don't get this line 🤔

slonka (approved):
Just one thing I don't understand

tools/resource-gen/pkg/generator/protoreflect.go
lobkovilya: It's the first enum value, i.e. if enum is
```
    enum Type {
      internal = 0;
      external = 1;
      gateway_delegated = 2;
      gateway_builtin = 3;
    }
```
then "zero" value for it is `internal` right?

slonka: 👍🏻

PR #14674, opened 2025-10-13
feat(oapi): expose KRI in REST resource meta and OpenAPI schema

bartsmykla:
## Motivation

Make `KRI` available in REST responses so clients and tooling can link and cross reference resources without extra calls. This is already aligned with how Inspect API exposes `KRI` and follows the agreed decision in [MADR 084](https://github.com/kumahq/kuma/blob/2410a1002fce4abac9a9e10396ec9f8c143b0a2d/docs/madr/decisions/084-add-kri-to-rest-api-resources.md).

Closes: https://github.com/kumahq/kuma/issues/14462

## Implementation information

- Compute KRI during JSON marshalling of ResourceMeta and surface it as `kri` in REST responses. This keeps storage and reconciliation untouched and avoids new fields that must be written or validated
- Add a safer helper `FromResourceMetaE` that returns an error if the descriptor cannot be resolved and returns empty when a resource has no short name. Keep the old `FromResourceMeta` as a wrapper for convenience
- Extend OpenAPI generator templates to include `kri` as a `readOnly` string with examples that reflect mesh and zone presence. Update the shared common resource schema to include the same `readOnly` field so SDKs document it correctly
- Implement a custom `MarshalJSON` for `Resource` to merge `meta`, `spec`, and `status` while preserving the injected `kri` from meta marshalling
- Regeneration of resources and golden files has been split into separate commits to make the review easier and isolate generated changes from functional logic

## Supporting documentation

- MADR: [`docs/madr/decisions/084-add-kri-to-rest-api-resources.md`](https://github.com/kumahq/kuma/blob/2410a1002fce4abac9a9e10396ec9f8c143b0a2d/docs/madr/decisions/084-add-kri-to-rest-api-resources.md)
- Prior PR agreement: https://github.com/kumahq/kuma/pull/14571


pkg/api-server/testdata/resources/crud/put_meshaccesslog_01.golden.json
```diff
@@ -23,7 +23,7 @@
        "file": {
         "format": {
          "type": "Plain",
-         "plain": "[%START_TIME%] %KUMA_SOURCE_SERVICE% -> %KUMA_DESTINATION_SERVICE%",
+         "plain": "[%START_TIME%] %KUMA_SOURCE_SERVICE% -\u003e %KUMA_DESTINATION_SERVICE%",
```
bartsmykla, line 26: ? 🤔

api/openapi/types/common/zz_generated.resource.go
```diff
@@ -102,6 +102,9 @@ type InspectRule struct {
 
 // Meta defines model for Meta.
 type Meta struct {
+	// KRI Kuma Resource Identifier (KRI) of the given resource
+	KRI *string `json:"kri,omitempty"`
```
lobkovilya, line 106: What happens if user tries to set `kri` when creates/updates resources through the API? Is there validation?

pkg/core/kri/kri.go
```diff
@@ -47,17 +47,33 @@ func From(r core_model.Resource) Identifier {
...
 func FromResourceMeta(rm core_model.ResourceMeta, resourceType core_model.ResourceType) Identifier {
-	if rm == nil {
+	id, err := FromResourceMetaE(rm, resourceType)
+	if err != nil {
 		return Identifier{}
 	}
+	return id
+}
+
+func FromResourceMetaE[T ~string](rm core_model.ResourceMeta, resourceType T) (Identifier, error) {
```
lobkovilya, line 57: Did you introduce this generic here just to avoid wrapping the `string` like
```go
kri.FromResourceMetaE(r, core_model.ResourceType(r.Type))
```
If that's the only reason then I'd just stick with 
```go
func FromResourceMeta(rm core_model.ResourceMeta, resourceType core_model.ResourceType) Identifier {}
```

pkg/core/kri/kri.go
```diff
@@ -47,17 +47,33 @@ func From(r core_model.Resource) Identifier {
...
+	if rm == nil {
+		return Identifier{}, nil
+	}
+
+	rType := core_model.ResourceType(resourceType)
+
+	if desc, err := registry.Global().DescriptorFor(rType); err != nil {
+		return Identifier{}, err
+	} else if desc.ShortName == "" {
```
lobkovilya, line 66: What types are we skipping here? Should we consider adding `ShortName` to these types?

api/openapi/types/common/zz_generated.resource.go
```diff
@@ -102,6 +102,9 @@ type InspectRule struct {
 
 // Meta defines model for Meta.
 type Meta struct {
```
lobkovilya, line 104: Can we use this type instead of our own `rest.v1alpha1.ResourceMeta`? It's a bit concerning we add `kri` field "on the fly" when marshalling but `rest.v1alpha1.ResourceMeta` doesn't have the actual `kri` field. At the same time here we have actual `kri` field, but no code reads/writes this field because it's purely for OAS

pkg/core/resources/model/rest/v1alpha1/resource.go
```diff
@@ -10,6 +12,38 @@ type Resource struct {
 	Status core_model.ResourceStatus `json:"status,omitempty"`
 }
 
+var _ json.Marshaler = (*Resource)(nil)
+
+func (r *Resource) MarshalJSON() ([]byte, error) {
```
lobkovilya, line 17: ugh, I hate we have to do this because `ResourceMeta` is embedded 😞

pkg/core/kri/kri.go
bartsmykla: It was considered and rejected in the MADR

pkg/core/resources/model/rest/v1alpha1/resource.go
bartsmykla: I'm not a fan of doing it this way either, but it's the simplest option for now

pkg/core/kri/kri.go
bartsmykla: `FromResourceMetaE` was added mainly to handle the check `if desc, err := registry.Global().DescriptorFor(rType); err != nil {`, with the generic parameter serving mostly as syntactic sugar

pkg/core/kri/kri.go
slonka: should we debug log on this?

bartsmykla: We shouldn't, since legacy resources can legitimately lack short names, and enabling debug would create excessive noise

api/openapi/types/common/zz_generated.resource.go
bartsmykla: In the OpenAPI schema, this field is marked as `readOnly`. Currently, we don’t perform OpenAPI validation on incoming requests. If we decide to enable such validation, it should be handled as a separate task.

api/openapi/types/common/zz_generated.resource.go
bartsmykla: This isn’t trivial since `rest.v1alpha1.ResourceMeta` is currently used in multiple places. If we decide to proceed, it should be done in follow-up PRs.

PR #14675, opened 2025-10-13
feat(kuma-cp): implementation of get by KRI endpoint

slonka:
## Motivation

Implementation of https://github.com/kumahq/kuma/issues/13882

## Implementation information

* create a new `kriEndpoint` structure that only holds information needed for KRI endpoint (resourceEndpoint has a bunch more that we don't need)
* fix a bug in KRI and `shortName`
* Create `FetchResourceByKri` and dry up `FetchResource`

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/13882

### DIFF
--- a/pkg/api-server/resource_endpoints.go
+++ b/pkg/api-server/resource_endpoints.go
@@ -254,7 +254,7 @@ func (r *resourceEndpoints) findResource(withInsight bool) func(request *restful
 	}
 }
 
-func formatResource(resource core_model.Resource, format string, k8sMapper k8s.ResourceMapperFunc, namespace string) (interface{}, error) {
+func formatResource(resource core_model.Resource, format string, k8sMapper k8s.ResourceMapperFunc, namespace string) (any, error) {
 	switch format {
 	case "k8s", "kubernetes":
 		res, err := k8sMapper(resource, namespace)
--- a/pkg/core/kri/kri.go
+++ b/pkg/core/kri/kri.go
@@ -46,10 +46,9 @@ func (i Identifier) IsLocallyOriginated(isGlobal bool, zone string) bool {
 	if isGlobal {
 		// In Global CP, resources without a zone are considered locally originated.
 		return i.Zone == ""
-	} else {
-		// In Zone CP, resources are treated as locally originated if KRI zone matches the current CP zone.
-		return i.Zone == zone
 	}
+	// In Zone CP, resources are treated as locally originated if KRI zone matches the current CP zone.
+	return i.Zone == zone
 }
 
 func From(r core_model.Resource) Identifier {


pkg/core/kri/kri.go
```diff
@@ -69,7 +82,7 @@ func FromString(s string) (Identifier, error) {
 		return Identifier{}, errors.Errorf("identifier must start with 'kri': %q", s)
 	}
 	ds := registry.Global().ObjectDescriptors(core_model.TypeFilterFn(func(d core_model.ResourceTypeDescriptor) bool {
-		return d.ShortName == parts[1]
+		return d.ShortName == parts[1] && d.ShortName != ""
```
slonka, line 81: This is a bug, otherwise if you pass an empty `shortName` (and there are policies that don't have `shortName`) it will match with a random policy.

test/framework/api/fetch.go
```diff
@@ -7,11 +7,29 @@ import (
...
 
+	"github.com/kumahq/kuma/pkg/core/kri"
 	core_model "github.com/kumahq/kuma/pkg/core/resources/model"
 	"github.com/kumahq/kuma/pkg/core/resources/model/rest"
 	"github.com/kumahq/kuma/test/framework"
 )
 
+// fetchResourceFromPath performs the HTTP GET and unmarshalling for a given API path.
+func fetchResourceFromPath(g gomega.Gomega, cluster framework.Cluster, out core_model.Resource, path string) int {
+	r, err := http.Get(cluster.GetKuma().GetAPIServerAddress() + path)
```
bartsmykla, line 20: This should be marked as GinkgoHelper I think

pkg/api-server/resource_endpoints.go
```diff
@@ -242,26 +242,34 @@ func (r *resourceEndpoints) findResource(withInsight bool) func(request *restful
...
+			rest_errors.HandleError(request.Request.Context(), response, err, "Could not retrieve a resource")
+			return
 		}
 		if err := response.WriteAsJson(res); err != nil {
 			log.Error(err, "Could not write the find response")
 		}
 	}
 }
 
+func formatResource(resource core_model.Resource, format string, k8sMapper k8s.ResourceMapperFunc, namespace string) (interface{}, error) {
```
bartsmykla: Nit:

```suggestion
func formatResource(resource core_model.Resource, format string, k8sMapper k8s.ResourceMapperFunc, namespace string) (any, error) {
```

pkg/core/kri/kri.go
```diff
@@ -42,6 +42,16 @@ func (i Identifier) IsEmpty() bool {
...
 }
 
+func (i Identifier) IsLocallyOriginated(isGlobal bool, zone string) bool {
+	if isGlobal {
+		// In Global CP, resources without a zone are considered locally originated.
+		return i.Zone == ""
+	} else {
+		// In Zone CP, resources are treated as locally originated if KRI zone matches the current CP zone.
+		return i.Zone == zone
+	}
```
bartsmykla: Nit:
```suggestion
	}
	// In Zone CP, resources are treated as locally originated if KRI zone matches the current CP zone.
	return i.Zone == zone
```

PR #14726, opened 2025-10-17
feat(meshidentity): support migration from one trustDomain to another

lukidzi:
## Motivation

~Needs https://github.com/kumahq/kuma/pull/14387~

We want to migrate from one SpiffeID to another without a downtime. Currently the problem was that MeshService was missing a new identifier.

## Implementation information

* Change API so provider is not required
* Update MeshService updater to generate SpiffeID once there is no provider
* Don't add SAN validator when using MeshTrust as SPIFFE validator has a trust domain validation
* Added new conditions options in a MeshIdentity which points to providing SpiffeID but not identity
* Added a E2E test with a constants request - flake attempt as seems that other tests are impacting cross zone (needs more investigation why sometimes there is one/two failures on cross zone request)

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/14388


pkg/core/resources/apis/meshidentity/api/v1alpha1/helpers.go
```diff
@@ -20,6 +20,24 @@ const (
...
+		}
+		if mi.Status.IsInitialized() || mi.Status.IsSpiffeIDProviderMode() {
+			matches = append(matches, mi)
+		}
+	}
+
+	return matches
+}
+
 func Matched(
```
lobkovilya: Probably `Matched` should be implemented using `AllMatched` and also renamed to `BestMatched` or something like that

pkg/core/resources/apis/meshidentity/api/v1alpha1/helpers.go
```diff
@@ -20,6 +20,24 @@ const (
 	defaultPathTemplate        = "/ns/{{ .Namespace }}/sa/{{ .ServiceAccount }}"
 )
 
+// we want all matching MeshIdentities that are Ready or in SpiffeIDProvider mode
+func AllMatched(
```
lobkovilya, line 24: It's a public func, please add a proper go doc comment `// AllMatched ...`

pkg/core/resources/apis/meshidentity/api/v1alpha1/helpers.go
```diff
@@ -140,3 +158,12 @@ func (s *MeshIdentityStatus) IsInitialized() bool {
 	}
 	return false
 }
+
+func (s *MeshIdentityStatus) IsSpiffeIDProviderMode() bool {
```
lobkovilya: TBH the name of the method doesn't reflect what it's doing, it just checks that at least one of conditions is equal to
```
type: Ready
status: False
reason: PartiallyReady
```
nothing about SPIFFE.

Probably the correct name would be `IsPartiallyReady`?

lobkovilya: or maybe there is typo and the condition.Type should be checked against `SpiffeIDProviderConditionType`?

pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin_test.go
```diff
@@ -180,6 +182,41 @@ var _ = Describe("MeshTLS", func() {
 				},
 			},
 		}),
+		Entry("strict with MeshTrust", testCase{
```
lobkovilya, line 185: Just checking my understanding, these tests don't check the stuff introduced by this PR? IIUC MeshTLS is not affected right?

pkg/xds/envoy/tls/v3/tls.go
```diff
@@ -84,6 +95,8 @@ func createCommonTlsContext(ownMesh core_xds.IdentityCertRequest, targetMeshCa c
 		),
 	)
 	if useMeshTrust {
+		// spiffe validator has trust domain validation
+		matchers = []*envoy_tls.SubjectAltNameMatcher{}
```
lobkovilya, line 99: Same, is it related to the PR? IIUC the feature is just generate more spiffeIds for MeshService, so envoy generators shouldn't really be touched. I also don't see PR description to mention something about it

test/e2e_env/multizone/meshidentity/migration.go
```diff
@@ -136,16 +221,19 @@ spec:
 			Install(YamlUniversal(yaml)).
 			Setup(multizone.Global)).To(Succeed())
 
-		hashedName := hash.HashedName(meshName, "identity")
+		hashedName = hash.HashedName(meshName, "identity-migration")
 		Expect(WaitForResource(meshidentity_api.MeshIdentityResourceTypeDescriptor, model.ResourceKey{Mesh: meshName, Name: fmt.Sprintf("%s.%s", hashedName, Config.KumaNamespace)}, multizone.KubeZone1, multizone.KubeZone2)).To(Succeed())
 
+		// wait 5 seconds for MeshIdentity to be reconcile
+		time.Sleep(5 * time.Second)
```
lobkovilya: There is status with `conditions`, can't we use `Eventually` and wait for the specific condition?

test/e2e_env/multizone/meshidentity/migration.go
```diff
@@ -111,11 +143,64 @@ func Migration() {
...
+`, meshName)
+		Expect(NewClusterSetup().
+			Install(YamlUniversal(onlyIdentity)).
+			Setup(multizone.Global)).To(Succeed())
+
+		hashedName := hash.HashedName(meshName, "only-identity-migration")
+		Expect(WaitForResource(meshidentity_api.MeshIdentityResourceTypeDescriptor, model.ResourceKey{Mesh: meshName, Name: fmt.Sprintf("%s.%s", hashedName, Config.KumaNamespace)}, multizone.KubeZone1, multizone.KubeZone2)).To(Succeed())
+
+		// wait 5 seconds for MeshIdentity to be reconcile
+		time.Sleep(5 * time.Second)
```
lobkovilya: Can we fetch MeshService inside `Eventually` until it gets correct `spiffeId`?

pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin_test.go
lukidzi: It’s related to a previous PR, but the test wasn’t added at that time. I’ve included it here now.

pkg/xds/envoy/tls/v3/tls.go
lukidzi: I missed this in a previous PR, as it wasn’t testing the full migration. Since we now create a SPIFFE validator that includes the trust domains, it’s no longer needed.

pkg/core/resources/apis/meshidentity/api/v1alpha1/helpers.go
lukidzi: Changed it

pkg/core/resources/apis/meshidentity/api/v1alpha1/helpers.go
```diff
@@ -32,11 +54,7 @@ func Matched(
 
 	var matches []scoredMatch
 
-	for _, mi := range meshIdentities {
-		if mi.Spec.Selector == nil || mi.Spec.Selector.Dataplane == nil || !mi.Spec.Selector.Dataplane.Matches(labels) {
-			continue
-		}
-
+	for _, mi := range AllMatched(labels, meshIdentities) {
```
lobkovilya, line 54: I just noticed, previous implem didn't check `mi.Status` but now in `AllMatched` you're also checking 
```go
if mi.Status != nil && (mi.Status.IsInitialized() || mi.Status.IsPartiallyReady()) {
    matches = append(matches, mi)
}
```
so IIUC `BestMatched` now won't return MeshIdentity unless it's either initialized or partially ready

test/e2e_env/multizone/meshidentity/migration.go
```diff
@@ -219,26 +348,7 @@ spec:
...
-		Eventually(func(g Gomega) {
-			resp, err := client.CollectEchoResponse(
-				multizone.KubeZone1, "demo-client", "test-server.meshidentity-migration.svc.kuma-2.mesh.local",
-				client.FromKubernetesPod(namespace, "demo-client"),
-			)
-			g.Expect(err).ToNot(HaveOccurred())
-			g.Expect(resp.Instance).To(Equal("kube-test-server-zone-2"))
-		}, "30s", "1s").MustPassRepeatedly(5).Should(Succeed())
+		time.Sleep(5 * time.Second) // let the goroutine execute more requests
+		Expect(reqError.Load()).To(BeNil())
```
lobkovilya: this seems like something to use `Consistently` for

pkg/core/resources/apis/meshidentity/api/v1alpha1/helpers.go
lukidzi: I've changed it to validate later if resource is initialised

PR #14743, opened 2025-10-21
fix(oapi): add required fields on schemas that are used in one of

slonka:
## Motivation

Unfortunately none of our oneOf subschemas objects have required fields, so the golang deserializer in speakeasy generator, which has a default strategy of trying to fit one-by-one, can effectively "fit" any of the subschemas in.
There was a correctness change, if the count of required fields was the same, to prefer to OAS ordering when deserializing. This causes the regression in https://github.com/Kong/terraform-provider-konnect-beta/issues/104.
In general, deserialization strategy in speakeasy for oneOf is pretty primitive; it effectively tries all of them and if all of them fit, the order in the OAS is preferred. Ideally we'd look at non-required fields too with some kinda scoring mechanism. The fix is ideally an OAS update to set at least one required field on these blobs so that each oneOf subschema is distinct. 

Unfortunately speakeasy can't revert the behaviour because it's a correctness change and appears sound. They usually hide these behind some kind of feature flag to avoid accidentally regressing live code, but that got missed here.

## Implementation information

Change the `AdditionalProtoTypes` array to be a map of type to required fields.

## Supporting documentation

xrel https://github.com/Kong/terraform-provider-konnect-beta/issues/104

> Changelog: skip

### DIFF
--- a/api/mesh/v1alpha1/datadogtracingbackendconfig/schema.yaml
+++ b/api/mesh/v1alpha1/datadogtracingbackendconfig/schema.yaml
@@ -16,6 +16,9 @@ components:
             get service names like `backend_INBOUND`, `backend_OUTBOUND_db1`, and
             `backend_OUTBOUND_db2` in Datadog. Default: false
           type: boolean
+      required:
+      - address
+      - port
       type: object
 info:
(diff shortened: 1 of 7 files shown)


slonka:
This was fixed another way

PR #14873, opened 2025-11-05
feat(MeshMetric): add `kuma.workload` attribute to metrics

Automaat:
## Motivation

  This PR implements support for unified resource naming in MeshMetric attributes as part of https://docs.google.com/document/d/1NOiIikp-zZ_qtBYC3zoW_X4JQRVVUwuiH-uSUxczk5Q/edit?usp=sharing. When the unified naming feature flag is enabled, metrics should no longer include the legacy mesh, dataplane, and service attributes to reduce metric cardinality and
  align with the new standardized naming scheme for Envoy resources and stats.

###  Metrics Format Mapper Changes

  Modified the metrics format conversion pipeline to conditionally exclude legacy attributes:

  - Updated FromPrometheusMetrics in app/kuma-dp/pkg/dataplane/metrics/metrics_format_mapper.go:24 to accept a new unifiedResourceNamingEnabled parameter
  - Modified extraAttributesFrom function in app/kuma-dp/pkg/dataplane/metrics/metrics_format_mapper.go:186 to skip adding mesh, dataplane, and service attributes when unified naming is enabled
  - These attributes are replaced with custom labels passed through the attributes map parameter

###  Metrics Producer Changes

  Updated the AggregatedProducer in app/kuma-dp/pkg/dataplane/metrics/metrics_producer.go:26:
  - Added unifiedNamingEnabled field to track the feature flag state
  - Propagated this flag through NewAggregatedMetricsProducer and fetchStats to the format mapper
  - Wired the flag from the dataplane runtime configuration in app/kuma-dp/cmd/run.go:508

###  MeshMetric Plugin Changes

  Enhanced the MeshMetric plugin at pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go:
  - Added WorkloadAttributeKey constant (kuma.workload) for future workload identification in plugin.go:46
  - Modified createDynamicConfig to accept the mesh resource and conditionally generate different label sets based on unified naming enablement in plugin.go:270
  - When unified naming is enabled, uses kuma.workload attribute instead of legacy dataplane labels (currently placeholder value "TODO")

Fix https://github.com/kumahq/kuma/issues/14900


app/kuma-dp/pkg/dataplane/metrics/metrics_format_mapper.go
```diff
@@ -183,18 +183,20 @@ func toOpenTelemetryQuantile(prometheusQuantiles []*io_prometheus_client.Quantil
 	return otelQuantiles
 }
 
-func extraAttributesFrom(mesh string, dataplane string, service string, extraLabels map[string]string) []attribute.KeyValue {
+func extraAttributesFrom(mesh string, dataplane string, service string, attributes map[string]string, unifiedResourceNamingEnabled bool) []attribute.KeyValue {
 	var extraAttributes []attribute.KeyValue
-	if mesh != "" {
```
lahabana, line 188: I don't get this, was this just passed for backward compatibility or something? Because it looks to me that these extraLabels are already passed in in the dynconfig no?

lahabana: If that's the case it might actually to do this kind of expansion earlier and stop having to pass through mesh, dp, service, unifiedResourceNamingEnabled in 2 functions no?

lahabana: To me such back compat should happen as close to the parsing as possible .i.e: https://github.com/kumahq/kuma/blob/4b5d411197b10bcc64eaa29d867e4b2cef7b9cc8/app/kuma-dp/pkg/dataplane/meshmetrics/component.go#L86

It also even makes sense to move up the attribute stuff all the way to `ApplicationToScrape` field and avoid recreating these always.

Automaat: it was used to add some stuff we were adding in old metrics, for example tags from dataplane, we can now use it to add workload, and potentially some extra labels which could be configured from policy in the future

Automaat: I went one step further and left only extraAttributes and added old stuff at the policy level. It should not break backward compatibility as dataplane with old version will just override these

app/kuma-dp/pkg/dataplane/metrics/metrics_producer.go
```diff
@@ -26,11 +26,12 @@ type AggregatedProducer struct {
...
 	applicationsToScrapeMutex *sync.Mutex
+	unifiedNamingEnabled      bool
 }
 
 var _ sdkmetric.Producer = &AggregatedProducer{}
 
-func NewAggregatedMetricsProducer(mesh string, dataplane string, service string, applicationsToScrape []ApplicationToScrape, isUsingTransparentProxy bool, kumaVersion string) *AggregatedProducer {
+func NewAggregatedMetricsProducer(mesh string, dataplane string, service string, applicationsToScrape []ApplicationToScrape, isUsingTransparentProxy bool, kumaVersion string, unifiedNamingEnabled bool) *AggregatedProducer {
 	return &AggregatedProducer{
 		mesh:                      mesh,
```
lahabana: don't think you need these fields anymore right?

Automaat: forgot to remove, this should be fine now

pkg/plugins/policies/meshmetric/plugin/v1alpha1/testdata/basic.listeners.golden.yaml
```diff
@@ -31,19 +31,19 @@ resources:
                   headers:
                   - name: If-None-Match
                     stringMatch:
-                      exact: 385b16329ac07fc4fd859a249a471e20a6c9f594896933deb2a699f4a3d1ac59
+                      exact: 96c816e7f8ba1de6f051b44d6254fd02ca055c8b558f18d8b77aeddc628cb799
                   path: /meshmetric
               - directResponse:
                   body:
-                    inlineString: '{"observability":{"metrics":{"applications":[{"path":"/metrics","port":8080,"address":""}],"backends":[{"type":"Prometheus","name":null}],"sidecar":{"includeUnused":false},"extraLabels":{"kuma_io_service":"backend","kuma_io_services":",backend,"}}}}'
+                    inlineString: '{"observability":{"metrics":{"applications":[{"path":"/metrics","port":8080,"address":""}],"backends":[{"type":"Prometheus","name":null}],"sidecar":{"includeUnused":false},"extraLabels":{"dataplane":"dp-1","kuma.workload":"TODO","kuma_io_service":"backend","kuma_io_services":",backend,","mesh":"default","service":"backend"}}}}'
```
lahabana, line 38: I'm not following how we end up with both `kuma.workload` and the other labels. Shouldn't it be exclusive?

lahabana: ha gotcha you add it to the old label set too. why not... Was there any specific reason?

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -257,7 +265,18 @@ func createDynamicConfig(conf api.Conf, proxy *core_xds.Proxy, resources xds_con
 	if rawList := resources[core_mesh.MeshGatewayType]; rawList != nil {
 		gateways = rawList.(*core_mesh.MeshGatewayResourceList).Items
 	}
-	extraLabels := mads.DataplaneLabels(proxy.Dataplane, gateways)
+
+	var extraLabels map[string]string
```
lahabana: To make my later comment unecessary because it'd be obvious I'd do:

```
extraLabels = map[string]string{
   WorkloadAttributeKey : "TODO"
}
if !unified_naming.Enabled(proxy.Metadata, mesh) {
   // add legacy label
}
```

This also makes cleaning the backward compatibility easier (just delete the if block)

pkg/plugins/policies/meshmetric/plugin/v1alpha1/testdata/otel_and_prometheus.listeners.golden.yaml
```diff
@@ -75,15 +74,15 @@ resources:
             virtualHosts:
             - domains:
               - '*'
-              name: system_meshmetric_otel_otel-collector-observability-svc-4317
+              name: _kuma:metrics:opentelemetry:otel-collector.observability.svc-4317
```
lahabana, line 77: Why are these changing suddenly?

lahabana (dismissed):
LGTM got a few minor questions

pkg/plugins/policies/meshmetric/plugin/v1alpha1/testdata/otel_and_prometheus.listeners.golden.yaml
Automaat: I've changed old test to not use unified naming and added specific case for unified naming

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
Automaat: updated

pkg/plugins/policies/meshmetric/plugin/v1alpha1/testdata/basic.listeners.golden.yaml
Automaat: It was in madr, I guess we can have this here for old stuff, we could start using this before people migrate to unified naming

lahabana (approved):
LGTM (btw changed your PR title as it was wrong)

app/kuma-dp/pkg/dataplane/meshmetrics/component.go
```diff
@@ -276,6 +278,14 @@ func (m *Manager) Shutdown() error {
 	return nil
 }
 
+func mapToAttributes(extraLabels map[string]string) []attribute.KeyValue {
+	var extraAttributes []attribute.KeyValue
+	for k, v := range extraLabels {
```
lobkovilya, line 283: Unstable order, need to sort keys

Automaat: Do we? I don't think it matters for otel

lobkovilya: It looks like `Manager` is missing test coverage, however for example the code that maps `dpapi.Application` to `metrics.Application` could be easily tested. I'm not saying you need to add tests in this PR, but that's why stable order is usually important

Automaat: I will think about this tests in separate pr

Issue #14744, opened 2025-10-21
MeshCircuitBreaker envoyconfig seems broken


lobkovilya:
### Description

[Input](https://github.com/kumahq/kuma/blob/master/test/e2e_env/universal/envoyconfig/testdata/sidecars/meshcircuitbreaker/inbound-rules-dpp.input.yaml) clearly sets the circuit breaker for test-server:
```yaml
type: MeshCircuitBreaker
name: mcb-1
mesh: envoyconfig
labels:
  kuma.io/effect: shadow
spec:
  targetRef:
    kind: Dataplane
    name: test-server
  rules:
    - default:
        connectionLimits:
          maxConnectionPools: 5
          maxConnections: 10
          maxPendingRequests: 2
          maxRetries: 1
          maxRequests: 32
```
But [golden file](https://github.com/kumahq/kuma/blob/master/test/e2e_env/universal/envoyconfig/testdata/sidecars/meshcircuitbreaker/inbound-rules-dpp.test-server.golden.json) has empty diff:
```json
{
  "diff": []
}
```

PR #14875, opened 2025-11-05
fix(MeshCircuitBreaker): properly configure inbounds with servicePort set

Automaat:
## Motivation

MeshCircuitBreaker policy was not being applied correctly to inbound clusters when a service port was configured. The plugin was using only the dataplane port to look up the cluster name, without considering the service port, which caused it to fail to find and configure the correct inbound cluster.

This resulted in circuit breaker rules being silently ignored for inbound traffic when service ports were specified.

  ## Implementation information

  This fix centralizes the inbound cluster name resolution logic:

  1. Created a new `GetInboundClusterName(servicePort, listenerPort)` function in `pkg/xds/envoy/names/resource_names.go` that encapsulates the logic for determining the correct cluster name:
     - Uses `servicePort` if non-zero
     - Falls back to `listenerPort` otherwise

  2. Updated `meshcircuitbreaker` plugin to use `GetInboundClusterName()` instead of directly calling `GetLocalClusterName(iface.DataplanePort)` in `pkg/plugins/policies/meshcircuitbreaker/plugin/v1alpha1/plugin.go:95`

  3. Updated `meshtimeout` plugin to use the centralized function and removed its duplicate `createInboundClusterName()` helper function from `pkg/plugins/policies/meshtimeout/plugin/v1alpha1/plugin.go:267-272`

  4. Updated golden test files to reflect that circuit breaker configuration is now properly applied

  The alternative considered was to duplicate the logic in each plugin, but centralizing it ensures consistency across all policies that need to resolve inbound cluster names and makes future maintenance easier.

Fix: https://github.com/kumahq/kuma/issues/14744

### DIFF
--- a/pkg/plugins/policies/meshcircuitbreaker/plugin/v1alpha1/plugin_test.go
+++ b/pkg/plugins/policies/meshcircuitbreaker/plugin/v1alpha1/plugin_test.go
@@ -268,7 +268,7 @@ var _ = Describe("MeshCircuitBreaker", func() {
 				{
 					Name:     "inbound",
 					Origin:   metadata.OriginInbound,
-					Resource: test_xds.ClusterWithName(fmt.Sprintf("localhost:%d", builders.FirstInboundPort)),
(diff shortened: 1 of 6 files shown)


bartsmykla (dismissed):
LGTM

Issue #14896, opened 2025-11-06
Create new Workload resource type


Automaat:
## Motivation

  Need a first-class API object to represent logical groups of Dataplanes. This enables stable KRI, zone-to-global sync, and listing workloads without scanning all DPPs.

##  Implementation information

  Create new resource type with schema:
```
  kind: Workload
  metadata:
    namespace: ns-1
    name: frontend
  spec: {}
  status:
    dataplaneProxies:
      connected: 100
      healthy: 99
      total: 100
```

  ResourceDescriptor configuration:
```
  var WorkloadResourceTypeDescriptor = model.ResourceTypeDescriptor{
      Name:                         WorkloadType,
      Scope:                        model.ScopeMesh,
      KDSFlags:                     model.ZoneToGlobalFlag,
      WsPath:                       "workloads",
      KumactlArg:                   "workload",
      KumactlListArg:               "workloads",
      AllowToInspect:               false,
      IsPolicy:                     false,
      IsDestination:                false,
      IsExperimental:               false,
      SingularDisplayName:          "Workload",
      PluralDisplayName:            "Workloads",
      IsPluginOriginated:           true,
      IsTargetRefBased:             false,
      HasToTargetRef:               false,
      HasFromTargetRef:             false,
      HasRulesTargetRef:            false,
      HasStatus:                    true,
      AllowedOnSystemNamespaceOnly: false,
      IsReferenceableInTo:          false,
      ShortName:                    "wl",
      IsFromAsRules:                false,
  }
```

  - Name can be up to 253 characters (same as ServiceAccount limit)
  - Empty spec for now
  - Status field shows DPP statistics

PR #14908, opened 2025-11-06
feat(workload): add Workload resource

Automaat:
## Motivation

  Introduces new Workload resource to logically group Dataplane Proxies for observability and metrics aggregation. Needed to move away from kuma.io/service tag grouping which  incorrectly assumes one-to-one service-to-workload mapping.

  ## Implementation information

  - Bootstrapped new plugin-generated resource using policy-gen
  - Added Workload v1alpha1 API with status field tracking DPP counts
  - Configured as mesh-scoped resource with ZoneToGlobal KDS sync
  - Generated CRDs, K8s types, REST/OpenAPI schemas, validators
  - Updated helm charts, RBAC, completion scripts, test fixtures
  - Resource descriptor configured: scope=Mesh, has_status=true, short_name=wl

Fix https://github.com/kumahq/kuma/issues/14896


lahabana:
🚢  it!

PR #14921, opened 2025-11-07
ci(lint): block experimental packages using gomodguard

lahabana:
Replace exptostd linter with gomodguard to block usage of experimental packages that have standard library equivalents. This approach is more widely adopted and provides explicit control over blocked packages.

Blocked packages:
- golang.org/x/exp/maps → use maps (stdlib)
- golang.org/x/exp/slices → use slices (stdlib)

Note: golang.org/x/exp/constraints is not blocked as it has no stdlib equivalent yet.


lahabana:
Superseeds #13810

bartsmykla (approved):

Issue #14901, opened 2025-11-06
Store kuma.io/workload as annotation on Kubernetes and add webhook validation


Automaat:
## Motivation

  ServiceAccount names can exceed 63 characters (up to 253), which exceeds label length limits. Need to store as annotation similar to kuma.io/display-name. Also need to prevent users from manually setting this field.

##  Implementation information

  - Store kuma.io/workload as Pod annotation instead of label
  - Add webhook validation to deny Pod creation if kuma.io/workload label is set manually
  - Webhook should check the label is unset and deny resource creation if present
  - Follow same pattern as kuma.io/display-name annotation

PR #14928, opened 2025-11-07
feat(workload): handle workload labels on kubernetes

Automaat:
## Motivation

  Enable automatic workload identification on Kubernetes pods via labels for better resource tracking and management. Previously no automated workload labeling existed.

  ## Implementation information

  - Added `kuma.io/workload` label as computed label, automatically set by Kuma based on pod metadata
  - Implemented pod validator webhook preventing manual setting of `kuma.io/workload` label
  - Webhook validates CREATE/UPDATE operations on pods in namespaces with `kuma.io/sidecar-injection` label
  - Extended pod converter to populate workload label from pod annotations
  - Added label to k8s store for dataplane resource creation
  - Updated all helm/kumactl install golden files with new webhook configuration

Fix https://github.com/kumahq/kuma/issues/14901

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/pod_converter.go
+++ b/pkg/plugins/runtime/k8s/controllers/pod_converter.go
@@ -104,7 +104,6 @@ func (p *PodConverter) PodToIngress(ctx context.Context, zoneIngress *mesh_k8s.Z
 		return err
 	}
 	// we need to validate if the labels have changed
-	workloadName := computeWorkloadName(pod.Labels, p.WorkloadLabels, pod.Spec.ServiceAccountName)
 	labels, err := model.ComputeLabels(
 		core_mesh.ZoneIngressResourceTypeDescriptor,
 		currentSpec,
@@ -115,7 +114,6 @@ func (p *PodConverter) PodToIngress(ctx context.Context, zoneIngress *mesh_k8s.Z
 		model.WithK8s(true),
 		model.WithZone(p.Zone),
 		model.WithServiceAccount(pod.Spec.ServiceAccountName),
-		model.WithWorkload(workloadName),
 	)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


lobkovilya:
Why is there `ci/skip-test` label?

pkg/plugins/runtime/k8s/controllers/pod_converter.go
```diff
@@ -101,6 +104,7 @@ func (p *PodConverter) PodToIngress(ctx context.Context, zoneIngress *mesh_k8s.Z
 		return err
 	}
 	// we need to validate if the labels have changed
+	workloadName := computeWorkloadName(pod.Labels, p.WorkloadLabels, pod.Spec.ServiceAccountName)
```
lobkovilya: I don't think we need workload label on zone ingress

Automaat:
@lobkovilya I don't know I must have put it on wrong pr or something

pkg/plugins/runtime/k8s/controllers/testdata/30.dataplane.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
     k8s.kuma.io/service-account: test-sa
     kuma.io/mesh: default
     kuma.io/proxy-type: sidecar
+    kuma.io/workload: test-sa
```
lobkovilya, line 9: It should be an annotation, no?

Automaat: I think this is because how we serialize dataplane in tests, when you look at other examples, there are no annotations for `display-name` for example. I was testing this on k8s and I got annotation with workload

pkg/plugins/runtime/k8s/controllers/pod_converter.go
Automaat: Right, I was to hasty with adding stuff, removed now

lobkovilya (approved):

PR #14963, opened 2025-11-12
feat(workload): automatically create Workload resource on kubernetes

### LINKED ISSUE #14897: Implement automatic Workload resource creation from DPP labels
## Motivation

  Workload resources need to be automatically created when DPPs with kuma.io/workload labels connect. This eliminates manual resource management.

##  Implementation information

  - Monitor DPP creation/updates for kuma.io/workload label
  - Automatically create/update corresponding Workload resource
  - Resource name matches the label value
  - Clean up Workload resources when no DPPs reference them

Automaat:
## Motivation

  Workload resources need automatic lifecycle management. When DPPs with `kuma.io/workload` labels connect, corresponding Workload resources should be created. When
  all DPPs referencing a workload disconnect, the Workload resource, should be cleaned up automatically.

  ## Implementation information

  - Created `WorkloadReconciler` in `pkg/plugins/runtime/k8s/controllers/workload_controller.go`
    - Watches Dataplane resources for `kuma.io/workload` label
    - Creates Workload resources in same namespace/mesh as Dataplane
    - Cleans up orphaned Workloads when no DPPs reference them
    - Follows MeshServiceReconciler pattern with Namespace/Mesh watches
  - Registered controller after PodStatusReconciler in `plugin.go` (DPPs must exist first)
  - Added table-driven unit tests with golden files in `testdata/workload/`
  - Added e2e test verifying creation/deletion lifecycle

  ## Supporting documentation

  Fixes #14897

### DIFF
--- a/test/e2e_env/kubernetes/workload/workload.go
+++ b/test/e2e_env/kubernetes/workload/workload.go
@@ -57,7 +57,7 @@ func Workload() {
 		Eventually(func(g Gomega) {
 			podName, err = PodNameOfApp(kubernetes.Cluster, appName, namespace)
 			g.Expect(err).ToNot(HaveOccurred())
-		}, "30s").Should(Succeed())
+		}).Should(Succeed())
 
 		// and verify dataplane has workload annotation set to service account
 		Eventually(func(g Gomega) {
@@ -77,7 +77,7 @@ func Workload() {
 			// (stored as annotation in k8s, accessible through labels API)
 			workloadLabel := dp.Meta.GetLabels()[metadata.KumaWorkload]
 			g.Expect(workloadLabel).To(Equal(serviceAccount), "workload should equal service account name")
-		}, "30s").Should(Succeed())
+		}).Should(Succeed())
 	})
 
 	It("should use pod label as workload when workload labels configured", func() {
@@ -103,7 +103,7 @@ func Workload() {
 		Eventually(func(g Gomega) {
 			podName, err = PodNameOfApp(kubernetes.Cluster, appName, namespace)
 			g.Expect(err).ToNot(HaveOccurred())
-		}, "30s").Should(Succeed())
+		}).Should(Succeed())
 
 		// and verify dataplane has workload annotation set to "test-server" (from app.kubernetes.io/name label)
 		Eventually(func(g Gomega) {
@@ -122,7 +122,7 @@ func Workload() {
 			// verify workload is set from app label, not service account
 			workloadLabel := dp.Meta.GetLabels()[metadata.KumaWorkload]
 			g.Expect(workloadLabel).To(Equal(appLabel), "workload should equal app label value")
-		}, "30s").Should(Succeed())
+		}).Should(Succeed())
 	})
 
 	It("should reject pod creation with manual kuma.io/workload label", func() {
@@ -180,7 +180,7 @@ spec:
 		Eventually(func(g Gomega) {
 			_, err = PodNameOfApp(kubernetes.Cluster, appName, namespace)
 			g.Expect(err).ToNot(HaveOccurred())
-		}, "30s").Should(Succeed())
+		}).Should(Succeed())
 
 		// and verify Workload resource is created
 		Eventually(func(g Gomega) {


pkg/plugins/runtime/k8s/controllers/testdata/workload/06.resources.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: kuma.io/v1alpha1
+kind: Dataplane
+metadata:
+  name: other-dp
+  namespace: demo
+  annotations:
+    kuma.io/workload: persistent-workload
```
slonka, line 7: what is a persistent workload?

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, workload := range workloads.Items {
+		workloadName := workload.Name
+		meshName := workload.GetMesh()
+
+		if referencedWorkloads[meshName] == nil || !referencedWorkloads[meshName][workloadName] {
+			log.Info("deleting orphaned workload", "workload", workloadName, "mesh", meshName)
+			if err := r.Delete(ctx, &workload); err != nil && !kube_apierrs.IsNotFound(err) {
+				return errors.Wrapf(err, "failed to delete orphaned Workload %s", workloadName)
+			}
+		}
```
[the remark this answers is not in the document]
slonka: That seems correct, WDYT @Automaat ?

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,213 @@
...
+		if kube_apierrs.IsNotFound(err) {
+			// Dataplane deleted, check if we should cleanup Workload
+			log.V(1).Info("dataplane not found, checking for orphaned workloads")
+			return kube_ctrl.Result{}, r.cleanupOrphanedWorkloads(ctx, req.Namespace)
+		}
+		return kube_ctrl.Result{}, errors.Wrapf(err, "unable to fetch Dataplane %s", req.Name)
+	}
+
+	// If Dataplane is being deleted, trigger cleanup instead of create/update
+	if !dp.GetDeletionTimestamp().IsZero() {
```
lobkovilya: Do we set deletion timestamp for the DPP anywhere in the code? I'm pretty sure we don't do that

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,213 @@
...
+			Namespace: namespace,
+		},
+	}
+
+	result, err := kube_controllerutil.CreateOrUpdate(ctx, r.Client, workload, func() error {
+		// Set mesh label
+		if workload.Labels == nil {
+			workload.Labels = map[string]string{}
+		}
+		workload.Labels[mesh_proto.MeshTag] = meshName
```
lobkovilya, line 92: shouldn't this be covered by the webhook? https://github.com/kumahq/kuma/blob/b97acb2fbb367a258f823fc101f701f72a3085bb/pkg/plugins/runtime/k8s/webhooks/defaulter.go#L69-L69

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return nil
+}
+
+func (r *WorkloadReconciler) SetupWithManager(mgr kube_ctrl.Manager) error {
+	return kube_ctrl.NewControllerManagedBy(mgr).
+		Named("kuma-workload-controller").
+		For(&mesh_k8s.Dataplane{}).
+		Watches(&kube_core.Namespace{}, kube_handler.EnqueueRequestsFromMapFunc(NamespaceToDataplanesMapper(r.Log, mgr.GetClient())), builder.WithPredicates(predicate.LabelChangedPredicate{})).
+		Watches(&mesh_k8s.Mesh{}, kube_handler.EnqueueRequestsFromMapFunc(MeshToAllDataplanesMapper(r.Log, mgr.GetClient())), builder.WithPredicates(CreateOrDeletePredicate{})).
```
lobkovilya: why do you watch Namespace and Mesh?

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (r *WorkloadReconciler) SetupWithManager(mgr kube_ctrl.Manager) error {
+	return kube_ctrl.NewControllerManagedBy(mgr).
+		Named("kuma-workload-controller").
+		For(&mesh_k8s.Dataplane{}).
```
lobkovilya: I think you need to create controller for `&workload_k8s.Workload` instead of `&mesh_k8s.Dataplane{}` and watch Dataplanes. In that case `Reconcile` will be invoked with name/namespace of the workload, so you'll be able to: 
* list all DPPs for the workload (and that's how potentially you implement status later)
* delete workload when there are no DPPs for the workload (instead of constantly calling `cleanupOrphanedWorkloads` which is a rather expensive method

pkg/plugins/runtime/k8s/controllers/testdata/workload/06.resources.yaml
Automaat: this is just a random name for testing

pkg/plugins/runtime/k8s/controllers/workload_controller.go
Automaat: I checked it and this is needed because on k8s we take mesh from labels, so if this is not set defaulter will always set default mesh

pkg/plugins/runtime/k8s/controllers/workload_controller.go
Automaat: kubernetes is setting it, but it turns out it is not that usable, as datplane is removed immediately

pkg/plugins/runtime/k8s/controllers/workload_controller.go
Automaat: Good idea i didn't thought about it earlier. Updated

pkg/plugins/runtime/k8s/controllers/workload_controller.go
Automaat: it does not make much sense in the end, removed

pkg/plugins/runtime/k8s/controllers/workload_controller.go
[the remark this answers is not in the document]
Automaat: addressed

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,172 @@
...
+	// CreatedWorkloadReason is added to an event when
+	// a new Workload is successfully created.
+	CreatedWorkloadReason = "CreatedWorkload"
+	// UpdatedWorkloadReason is added to an event when
+	// an existing Workload is successfully updated.
+	UpdatedWorkloadReason = "UpdatedWorkload"
+	// DeletedWorkloadReason is added to an event when
+	// a Workload is successfully deleted.
+	DeletedWorkloadReason = "DeletedWorkload"
+)
```
lobkovilya: seems like they're unused?

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,172 @@
...
+
+// WorkloadReconciler reconciles Workload resources based on Dataplane labels
+type WorkloadReconciler struct {
+	kube_client.Client
+	kube_record.EventRecorder
```
lobkovilya: do we need it?

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,172 @@
...
+	Log               logr.Logger
+	Scheme            *kube_runtime.Scheme
+	ResourceConverter k8s_common.Converter
```
lobkovilya: do we use it?

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,172 @@
...
+	l = l.WithName("dataplane-to-workload-mapper")
+	return func(ctx context.Context, obj kube_client.Object) []kube_reconcile.Request {
+		dp, ok := obj.(*mesh_k8s.Dataplane)
+		if !ok {
+			l.Error(nil, "unexpected object type", "type", obj.GetObjectKind())
+			return nil
+		}
+
+		workloadName, ok := dp.GetAnnotations()[metadata.KumaWorkload]
+		if !ok || workloadName == "" {
```
lobkovilya: checking only `workloadName == ""` is enough

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		workload.Labels[mesh_proto.MeshTag] = meshName
+		workload.Labels[mesh_proto.ManagedByLabel] = "k8s-controller"
+
+		// Ensure spec is initialized (empty struct is fine)
+		if workload.Spec == nil {
+			workload.Spec = &workload_api.Workload{}
+		}
+
+		// Don't set status - leave it nil
```
lobkovilya: you don't need to comment every thing you're not doing here :D

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	// Find Dataplanes that reference this workload
+	var referencingDPs []mesh_k8s.Dataplane
+	var meshName string
+	for _, dp := range dataplanes.Items {
+		if workloadName, ok := dp.GetAnnotations()[metadata.KumaWorkload]; ok && workloadName == req.Name {
+			referencingDPs = append(referencingDPs, dp)
+			if meshName == "" {
+				meshName = dp.Mesh
```
lobkovilya, line 52: Today it's still possible to have DPPs from different meshes in a single namespace and this will cause Workload collision on Kubernetes. Discussed offline, I'll create a small doc to see what we can do (probably we shouldn't allow that). It should block this PR

test/e2e_env/kubernetes/workload/workload.go
```diff
@@ -157,4 +157,56 @@ spec:
...
+		err = k8s.RunKubectlE(
+			kubernetes.Cluster.GetTesting(),
+			kubernetes.Cluster.GetKubectlOptions(namespace),
+			"delete", "deployment", appName,
+		)
+		Expect(err).ToNot(HaveOccurred())
+
+		// then verify Workload resource is deleted
+		Eventually(func(g Gomega) {
+			// verify Workload resource exists and has correct content
```
lobkovilya: comment is wrong

test/e2e_env/kubernetes/workload/workload.go
```diff
@@ -157,4 +157,56 @@ spec:
...
+			workloadK8sName := fmt.Sprintf("%s.%s", workloadName, namespace)
+			_, err := GetWorkload(kubernetes.Cluster, workloadK8sName, mesh)
+			g.Expect(err).To(HaveOccurred(), "workload should be deleted")
+			g.Expect(err.Error()).To(ContainSubstring("No resources found in workloads mesh"))
+		}, "2m", "3s").Should(Succeed())
```
lobkovilya: why this `Eventually` has 3s and in the previous has 1s? Should we just omit it and keep the default?

pkg/plugins/runtime/k8s/controllers/workload_controller.go
Automaat: removed

pkg/plugins/runtime/k8s/controllers/workload_controller.go
Automaat: removed

pkg/plugins/runtime/k8s/controllers/workload_controller.go
Automaat: removed

pkg/plugins/runtime/k8s/controllers/workload_controller.go
Automaat: fixed

pkg/plugins/runtime/k8s/controllers/workload_controller.go
Automaat: yep :P

test/e2e_env/kubernetes/workload/workload.go
Automaat: removed

test/e2e_env/kubernetes/workload/workload.go
Automaat: removed

Automaat: in the end it needs longer timeout to pass when deleting workload

lobkovilya: My question was about the interval, why the previous Eventually was requesting workload every 1s but this does that every 3s

Automaat: since I extended the timeout I thought I don't need to check every second. To reduce number or operations, But I can change it to 1s if you want

lobkovilya: the default is 500ms https://github.com/kumahq/kuma/blob/719105eda57f365c8fefb4581f1a1e142e210e66/pkg/test/ginkgo.go#L41

unless you have good reasons to increase it 6 times I'd stick with the default

pkg/plugins/runtime/k8s/controllers/workload_controller.go
```diff
@@ -0,0 +1,148 @@
...
+	workload_k8s "github.com/kumahq/kuma/v2/pkg/core/resources/apis/workload/k8s/v1alpha1"
+	mesh_k8s "github.com/kumahq/kuma/v2/pkg/plugins/resources/k8s/native/api/v1alpha1"
+	"github.com/kumahq/kuma/v2/pkg/plugins/runtime/k8s/metadata"
+)
+
+// WorkloadReconciler reconciles Workload resources based on Dataplane labels
+type WorkloadReconciler struct {
+	kube_client.Client
+	Log    logr.Logger
+	Scheme *kube_runtime.Scheme
```
lobkovilya: is it used?

Automaat: it is not, fixed

test/e2e_env/kubernetes/workload/workload.go
Automaat: updated

test/e2e_env/kubernetes/workload/workload.go
```diff
@@ -57,7 +57,7 @@ func Workload() {
 		Eventually(func(g Gomega) {
 			podName, err = PodNameOfApp(kubernetes.Cluster, appName, namespace)
 			g.Expect(err).ToNot(HaveOccurred())
-		}, "30s", "1s").Should(Succeed())
+		}, "30s").Should(Succeed())
```
lobkovilya: Sorry for being annoying but 30s is the default :) https://github.com/kumahq/kuma/blob/719105eda57f365c8fefb4581f1a1e142e210e66/pkg/test/ginkgo.go#L42

pkg/plugins/runtime/k8s/controllers/workload_controller.go
lobkovilya: FYI I opened a PR with the [design](https://github.com/kumahq/kuma/pull/15050) suggestion and [issue to fix it](https://github.com/kumahq/kuma/issues/15049)

test/e2e_env/kubernetes/workload/workload.go
Automaat: no problem, it should be all good now

PR #14976, opened 2025-11-13
chore(tools): add proto oneOf support to OAS generator

bartsmykla:
## Motivation

OpenAPI schemas in `release-2.12` are missing proper `oneOf` support for protobuf `oneOf` fields, resulting in schemas with `Type: true` instead of proper `oneOf: [$ref: ...]` structures. Additionally, field descriptions are missing because `AddGoComments` doesn't work correctly with the `/v2` module path.

This causes two issues:

1. **Schema structure**: Fields like `DataSource.Type` show as `Type: true` instead of proper `oneOf` references
2. **Missing descriptions**: Go comment descriptions don't appear in generated schemas

These issues affect downstream projects that depend on these schemas.

## Implementation information

This PR backports the `oneOf` support from `master` to `release-2.12`:

**Commit 1: Cherry-pick oneOf support** (`cabf5ace54`)

- Backported from `master` commit `4ee924aa13` (PR #14639)
- Adds `tools/resource-gen/pkg/generator/protoreflect.go` with `oneOf` detection
- Rewrites OpenAPI generator to use custom `reflector` with `handleOneOf()` method
- Generates separate schema files for each `oneOf` variant
- Resolved import path conflicts (`v1` → `v2`)

**Commit 2: Integrate AddGoComments fix** (`cd63f4bc9a`)

- Integrates fixes from `master` commits: `cd4eb77780`, `2173b2ae6a`, `8a927d7d18`, `3cdbd24fd2`, `8c7d52b585`
- Hardcodes `/v2` module path instead of concatenating with `base` variable
- Adds symlink resolution and directory handling for cross-project calls
- Ensures working directory is restored after `AddGoComments`

**Result:**

- Schemas now use `oneOf: [$ref: ...]` instead of `Type: true`
- Field descriptions from Go comments appear in generated schemas
- Works correctly when called from downstream projects via symlinks

## Supporting documentation

Related to master PR #14639

Fixes OpenAPI schema issues in `release-2.12` for `/v2` module path migration.

> Changelog: skip

### DIFF
--- a/tools/resource-gen/pkg/generator/main.go
+++ b/tools/resource-gen/pkg/generator/main.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


bartsmykla:
Replaced by #14978 which removes the oneOf changes and only includes the AddGoComments symlink fix needed for /v2 migration in downstream projects.

PR #15035, opened 2025-11-18
feat(helm): support cert-manager CA injection for webhook configurations

slonka:
## Motivation

Enable dynamic CA bundle injection for Kuma webhook configurations using cert-manager. This eliminates hardcoded CA bundles in helm values, allowing seamless certificate rotation and simpler deployments across environments.

## Implementation information

Added `controlPlane.tls.general.certManager.enabled` configuration option that:

**Helm chart changes:**
- Creates `Certificate` and `Issuer` resources when enabled (`cp-certs-cert-manager.yaml`)
- Adds `cert-manager.io/inject-ca-from` annotation to webhook configurations
- Removes `caBundle` from webhook `clientConfig` (cert-manager injects it dynamically)
- Validates that `secretName` and `caBundle` are empty when cert-manager is enabled

**Testing:**
- Helm render test validates template output
- E2E test deploys Kuma with cert-manager and verifies CA injection into both validating and mutating webhooks

**Implementation follows a similar implementation in the Kong Operator:**
- [Configuration](https://github.com/Kong/kong-operator/blob/378bd036120af98ac5f5bab84ed96dcccee0b19b/charts/kong-operator/values.yaml#L125-L133)
- [ValidatingWebhookConfiguration](https://github.com/Kong/kong-operator/blob/6c00f36eb9d2cbe6b39d3ea0ad59dbfdcd97d4cf/charts/kong-operator/templates/validating-webhook.yaml#L22)
- [cert-manager annotation](https://github.com/Kong/kong-operator/blob/6c00f36eb9d2cbe6b39d3ea0ad59dbfdcd97d4cf/charts/kong-operator/templates/validating-webhook.yaml#L49-L52)
- [caBundle](https://github.com/Kong/kong-operator/blob/6c00f36eb9d2cbe6b39d3ea0ad59dbfdcd97d4cf/charts/kong-operator/templates/validating-webhook.yaml#L66-L68)
- [Certificate & Issuer Resources](https://github.com/Kong/kong-operator/blob/73deb41a0d7a0d6faaaca3dff32442c287fe20b6/charts/kong-operator/templates/certs-cert-manager.yaml#L21-L38)
- [Self-signed Issuer](https://github.com/Kong/kong-operator/blob/73deb41a0d7a0d6faaaca3dff32442c287fe20b6/charts/kong-operator/templates/certs-cert-manager.yaml#L40-L52)

###  Key Pattern Summary

1. Single boolean flag controls everything
2. Conditional Secret: Only generate self-signed when cert-manager is disabled
3. Conditional annotation: Only add cert-manager.io/inject-ca-from when enabled
4. Conditional caBundle: Only include when cert-manager is disabled
5. Auto-create Certificate + Issuer: Chart manages cert-manager resources
6. Annotation references Certificate name: Format is <namespace>/<certificate-name>

## Supporting documentation

https://github.com/Kong/kong-mesh/issues/8607

### DIFF
--- a/deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
+++ b/deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
@@ -10,10 +10,10 @@ Helper template for webhook clientConfig caBundle
 Omits caBundle when cert-manager is enabled (will be injected by cert-manager)
 */}}
 {{- define "kuma.webhook.caBundle" -}}
-{{- if not .Values.controlPlane.tls.general.certManager.enabled }}
+{{- if not .Values.controlPlane.tls.general.certManager.enabled -}}
 caBundle: {{ .caBundle }}
-{{- end }}
-{{- end }}
+{{- end -}}
+{{- end -}}
 {{/*
 Generate certificates
 see: https://masterminds.github.io/sprig/crypto.html
--- a/app/kumactl/cmd/install/testdata/install-control-plane.cni-enabled.golden.yaml
+++ b/app/kumactl/cmd/install/testdata/install-control-plane.cni-enabled.golden.yaml
@@ -684,7 +684,7 @@ spec:
     metadata:
       annotations:
         checksum/config: fd9d1d8386f97f2bd49e50f476520816168a1c9f60bbc43dec1347a64d239155
-        checksum/tls-secrets: 22984ab190b42402cb4f97dbe810c792fcf05493e0174e80e1280eda492d9308
+        checksum/tls-secrets: 2c0ab72d70a85dc951f5547748e9708715259a4d372172e37dc45c8c0bfd4051
       labels: 
         app: kuma-control-plane
         app.kubernetes.io/name: kuma
@@ -868,7 +868,6 @@ webhooks:
           values:
           - kube-system
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
@@ -930,7 +929,6 @@ webhooks:
           values:
           - kube-system
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
@@ -1001,7 +999,6 @@ webhooks:
           - kube-system
           - kuma-system
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
@@ -1033,7 +1030,6 @@ webhooks:
       matchLabels:
         kuma.io/sidecar-injection: enabled
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
@@ -1072,7 +1068,6 @@ webhooks:
           values:
           - kube-system
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
@@ -1143,7 +1138,6 @@ webhooks:
           - kuma-system
     failurePolicy: Ignore
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
@@ -1174,7 +1168,6 @@ webhooks:
           - kuma-system
     failurePolicy: Fail
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
--- a/app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
+++ b/app/kumactl/cmd/install/testdata/install-control-plane.defaults.golden.yaml
@@ -11606,7 +11606,7 @@ spec:
     metadata:
       annotations:
         checksum/config: fd9d1d8386f97f2bd49e50f476520816168a1c9f60bbc43dec1347a64d239155
-        checksum/tls-secrets: 22984ab190b42402cb4f97dbe810c792fcf05493e0174e80e1280eda492d9308
+        checksum/tls-secrets: 2c0ab72d70a85dc951f5547748e9708715259a4d372172e37dc45c8c0bfd4051
       labels: 
         app: kuma-control-plane
         app.kubernetes.io/name: kuma
@@ -11784,7 +11784,6 @@ webhooks:
           values:
           - kube-system
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
@@ -11846,7 +11845,6 @@ webhooks:
           values:
           - kube-system
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
@@ -11917,7 +11915,6 @@ webhooks:
           - kube-system
           - kuma-system
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
@@ -11949,7 +11946,6 @@ webhooks:
       matchLabels:
         kuma.io/sidecar-injection: enabled
     clientConfig:
-      
       caBundle: XYZ
       service:
         namespace: kuma-system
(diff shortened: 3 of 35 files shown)


deployments/charts/kuma/templates/cp-certs-cert-manager.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if and .Values.controlPlane.tls.general.certManager.enabled (eq .Values.controlPlane.environment "kubernetes") }}
```
bartsmykla, line 1: Add CRD validation before creating cert-manager resources:

```yaml
{{- if and .Values.controlPlane.tls.general.certManager.enabled (eq .Values.controlPlane.environment "kubernetes") }}
{{- $certManagerCRDExists := false }}
{{- range $crd := (lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "").items }}
  {{- if eq $crd.metadata.name "certificates.cert-manager.io" }}
    {{- $certManagerCRDExists = true }}
  {{- end }}
{{- end }}
{{- if not $certManagerCRDExists }}
  {{- fail "cert-manager CRDs not found. Please install cert-manager before enabling controlPlane.tls.general.certManager.enabled" }}
{{- end }}
{{- end }}
```

Without this, if cert-manager isn't installed, the `Certificate`/`Issuer` resources get created but never reconciled, leading to webhook failures with no clear indication of the root cause.

deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
```diff
@@ -1,6 +1,9 @@
 {{- if not (eq (empty .Values.controlPlane.tls.general.caBundle) (empty .Values.controlPlane.tls.general.secretName)) }}
   {{ fail "You need to send both or neither of controlPlane.tls.general.caBundle and controlPlane.tls.general.secretName"}}
 {{- end }}
+{{- if and .Values.controlPlane.tls.general.certManager.enabled (or (not (empty .Values.controlPlane.tls.general.secretName)) (not (empty .Values.controlPlane.tls.general.caBundle))) }}
+  {{ fail "When controlPlane.tls.general.certManager.enabled is true, secretName and caBundle must be empty"}}
+{{- end }}
```
bartsmykla, line 6: Simplify the double negation:

```yaml
{{- if and .Values.controlPlane.tls.general.certManager.enabled (or .Values.controlPlane.tls.general.secretName .Values.controlPlane.tls.general.caBundle) }}
```

Helm's truthiness check handles empty strings, so `(not (empty ...))` is unnecessary.

test/e2e/webhooks/cert_manager.go
```diff
@@ -0,0 +1,128 @@
...
+				cluster.GetTesting(),
+				cluster.GetKubectlOptions(kumaNamespace),
+				"get", "certificate", "kuma-tls-cert",
+			)
+			g.Expect(err).ToNot(HaveOccurred())
+		}, "60s", "1s").Should(Succeed(), "Certificate resource should exist")
+
+		// Verify Issuer resource exists
+		Eventually(func(g Gomega) {
+			_, err := k8s.RunKubectlAndGetOutputE(
```
bartsmykla: Add functional webhook test to verify webhooks actually work (not just that CA bundles are injected):

```go
It("should successfully validate resources using the webhook", func() {
    yaml := `
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: test-mesh
spec:
  mtls:
    enabledBackend: ca-1
    backends:
    - name: ca-1
      type: builtin
`
    err := k8s.KubectlApplyFromStringE(
        cluster.GetTesting(),
        cluster.GetKubectlOptions(Config.KumaNamespace),
        yaml,
    )
})
```

Also add a test for sidecar injection and certificate rotation.

test/e2e/webhooks/cert_manager.go
```diff
@@ -0,0 +1,128 @@
...
+		// Verify CA bundle is injected into validating webhook configuration
+		Eventually(func(g Gomega) {
+			output, err := k8s.RunKubectlAndGetOutputE(
+				cluster.GetTesting(),
+				cluster.GetKubectlOptions(),
+				"get", "validatingwebhookconfiguration", "kuma-validating-webhook-configuration",
+				"-o", "jsonpath={.webhooks[0].clientConfig.caBundle}",
+			)
+			g.Expect(err).ToNot(HaveOccurred())
+			// CA bundle should be non-empty base64-encoded certificate
```
bartsmykla: Validate the CA bundle is actually a valid X.509 certificate, not just that it's long enough:

```go
import "encoding/base64"

// Verify it's valid base64
decoded, err := base64.StdEncoding.DecodeString(output)

// Verify it contains PEM certificate
g.Expect(string(decoded)).To(ContainSubstring("BEGIN CERTIFICATE"))
g.Expect(len(decoded)).To(BeNumerically(">", 100))
```

test/e2e/webhooks/cert_manager.go
```diff
@@ -0,0 +1,128 @@
...
+		kumaNamespace := Config.KumaNamespace
+
+		// Verify Certificate resource exists
+		Eventually(func(g Gomega) {
+			_, err := k8s.RunKubectlAndGetOutputE(
+				cluster.GetTesting(),
+				cluster.GetKubectlOptions(kumaNamespace),
+				"get", "certificate", "kuma-tls-cert",
+			)
+			g.Expect(err).ToNot(HaveOccurred())
```
bartsmykla, line 69: Check the `Certificate` is in Ready status, not just that it exists:

```go
output, err := k8s.RunKubectlAndGetOutputE(
    cluster.GetTesting(),
    cluster.GetKubectlOptions(kumaNamespace),
    "get", "certificate", "kuma-tls-cert",
    "-o", "jsonpath={.status.conditions[?(@.type=='Ready')].status}",
)
g.Expect(output).To(Equal("True"))
```

Same for the `Issuer` resource.

test/framework/deployments/certmanager/kubernetes.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	// Wait for cert-manager pods to be ready
+	err = t.isPodReady(cluster, "app.kubernetes.io/name=cert-manager")
+	if err != nil {
+		return err
+	}
+	err = t.isPodReady(cluster, "app.kubernetes.io/name=cainjector")
+	if err != nil {
+		return err
```
bartsmykla: Simplify with a loop:

```go
podSelectors := []string{
    "app.kubernetes.io/name=cert-manager",
    "app.kubernetes.io/name=cainjector",
    "app.kubernetes.io/name=webhook",
}
for _, selector := range podSelectors {
    if err := t.isPodReady(cluster, selector); err != nil {
        return err
    }
}
```

test/framework/deployments/certmanager/kubernetes.go
```diff
@@ -0,0 +1,77 @@
...
+		metav1.ListOptions{
+			LabelSelector: selector,
+		},
+		1,
+		framework.DefaultRetries,
+		framework.DefaultTimeout)
+	if err != nil {
+		return err
+	}
+	return nil
```
bartsmykla: Add error context:

```go
if err != nil {
    return errors.Wrapf(err,
        "cert-manager pod with selector %q in namespace %q failed to become ready",
        selector, t.namespace)
}
```

Also simplify the function to just `return k8s.WaitUntilNumPodsCreatedE(...)` instead of the if/return/return pattern.

bartsmykla (approved):

PR #15037, opened 2025-11-18
chore(mesh): add test for invalid CA backend type validation

Automaat:
## Motivation

Invalid CA backend type (e.g., "bulitin" typo) needs test coverage to prevent 500 error regressions

## Implementation information

Added test case in mesh_manager_test.go validating that invalid CA backend types are properly caught. Refactored validateMtls() loop structure for cleaner validation flow.

## Supporting documentation

Related to https://github.com/Kong/kong-mesh/issues/8520

### DIFF
--- a/pkg/core/resources/apis/mesh/mesh_validator.go
+++ b/pkg/core/resources/apis/mesh/mesh_validator.go
@@ -13,11 +13,6 @@ import (
 	"github.com/kumahq/kuma/v2/pkg/util/proto"
 )
 
-const (
-	CaBuiltinType  = "builtin"
-	CaProvidedType = "provided"
-)
-
 var AllowedMTLSBackends = 1
 
 func (m *MeshResource) Validate() error {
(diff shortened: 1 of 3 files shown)


pkg/core/resources/apis/mesh/mesh_validator.go
```diff
@@ -75,18 +80,23 @@ func validateMtls(mtls *mesh_proto.Mesh_Mtls) validators.ValidationError {
...
-	}
-	if mtls.GetEnabledBackend() != "" && !usedNames[mtls.GetEnabledBackend()] {
-		verr.AddViolation("enabledBackend", "has to be set to one of the backends in the mesh")
-	}
-	for _, backend := range mtls.Backends {
+
+		switch backend.GetType() {
+		case CaBuiltinType, CaProvidedType:
+		default:
+			verr.AddViolationAt(validators.RootedAt("backends").Index(i).Field("type"), fmt.Sprintf("unknown backend type. Available backends: %q, %q", CaBuiltinType, CaProvidedType))
```
lobkovilya: what about Kong Mesh where we have other backend types

Automaat: right, that is probably why we didn't have proper validation in first place. I will look for some workaround

Automaat: I found the issue in kong mesh, I will just add unit test with this pr

PR #15042, opened 2025-11-19
fix(kds): server Send blocks when client doesn't call Recv for some time

lobkovilya:
## Motivation

If the client does not call `Recv` frequently enough, the server call to `Send` can block on the stream. This can be reproduced by creating around 200 policies for each of five different types and ensuring the client sends `DiscoveryRequest` messages for those types first.

## Implementation information

Change the implementation of `DeltaKDSStream` to use dedicated goroutines for sending and receiving, with each one reading from or writing to the appropriate channels.

This approach addresses the issue because, regardless of how large the DiscoveryResponse messages are, the client can always accommodate at least one response per distinct type sent over the KDS stream. Well behaving KDS client and KDS server don't send 2 consecutive DiscoveryRequests/DiscoveryResponses for the same typeURL.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/14486


lukidzi (approved):
🎉

pkg/kds/v2/client/stream.go
```diff
@@ -32,21 +33,102 @@ type stream struct {
...
+
+func (s *stream) sendLoop() {
+	for {
+		select {
+		case <-s.ctx.Done():
+			return
+		case req := <-s.sendCh:
+			if err := s.streamClient.Send(req); err != nil {
+				s.cancel(err)
+				return
```
Icarus9913, line 96: There's might be a case leading to goroutine leak. Once if there's a `grpc.Send` error here, you only cancel the wrapped `Context` rather than the `grpc stream/conn` level ctx. Then, in another `recvLoop` we assume the `grpc.Recv` still works well and be stuck by network IO.

Although in `recvLoop` we have the `s.ctx.Done` check later and get returned from thhe function. I'm thinking if we can cancel the `grpc stream/conn` level ctx.

And my idea is pass the `grpc.conn.cancel` on the `Client` and `grpc.server.stop` on the `Server` from the top to here as the cancel function, because we do restart the connection with the current implementation once hit error.

Icarus9913: Or we can wait until the `sendLoop` and `recvLoop` finished, then let the `WatchDog` restart again.

slonka: @lobkovilya WDYT?

lobkovilya: Once SendMsg receives error it [aborts](https://github.com/grpc/grpc-go/blob/cdbafd304983c6c2a903c801d05e3ca531d49888/stream.go#L113-L116) the stream. RecvMsg returns when the [stream is done](https://github.com/grpc/grpc-go/blob/cdbafd304983c6c2a903c801d05e3ca531d49888/stream.go#L137-L140). 

Also, they mention how to [ensure resources are not leaking](https://github.com/grpc/grpc-go/blob/cdbafd304983c6c2a903c801d05e3ca531d49888/stream.go#L148-L160):

> To ensure resources are not leaked due to the stream returned, one of the following actions must be performed:
> 
> 1. Call Close on the ClientConn.
> 2. Cancel the context provided.
> 3. Call RecvMsg until a non-nil error is returned. A protobuf-generated client-streaming RPC, for instance, might use the helper function CloseAndRecv (note that CloseSend does not Recv, therefore is not guaranteed to release all resources).
> 4. Receive a non-nil, non-io.EOF error from Header or SendMsg.

In this PR it always ensures non-nil err is returned either from RecvMsg or SendMsg (so 3 or 4 are always performed).

Additionally, when Send/Recv return error we put it as a ctx's "cause" error. This means all the following calls to `DeltaDiscoveryRequest`, `ACK`, `NACK` will return `context.Cause(s.ctx)`. When these method return error this error is propagated all the way back and we call cancel on the stream's context anyway https://github.com/kumahq/kuma/blob/9dbe37ffdbadc028103429f7d2927822aa04c4f1/pkg/kds/mux/client.go#L148

PR #15045, opened 2025-11-19
feat(api): introduce extension of api validators

lukidzi:
## Motivation

Currently, resource validation in Kuma is limited to built-in validators defined within each resource type. There is no mechanism to register additional validators dynamically for resource types from external code or plugins. This limitation prevents:

  - Custom validation logic from being added to core resource types without modifying the core codebase
  - Plugin or fork-specific validation rules from being enforced on existing resource types
  - Runtime validation extensions based on deployment-specific requirements

  The resource type registry already provides a central place for registering resource types, but lacks the infrastructure to register and invoke additional validators beyond the built-in validate() method that some resources implement.

## Implementation information

This PR adds a validator registration and invocation system to the resource type registry.

* Added AdditionalValidator interface that accepts any resource and returns structured validation errors.
* Extended TypeRegistry interface with methods to register validators per resource type and retrieve them dynamically at validation time.
* Added global helper RegisterTypeValidator() for convenient registration against the global registry.
* Template Updates (tools/policy-gen/generator/cmd/core_resource.go)

Modified the generated Validate() method to:
  1. Run the built-in validate() method if present
  2. Fetch additional validators from the global registry
  3. Execute all validators and accumulate violations
  4. Return all errors together in a single pass
This ensures all validation feedback is collected before returning to the caller.

### Generated Resources

All zz_generated.resource.go files across core resources and policy plugins now include the updated validation logic with registry integration.

Key Design Decisions
  - Dynamic lookup: Validators are retrieved at validation time, preventing race conditions between RegisterType() and RegisterValidator() calls
  - Error accumulation: All validators run to completion before returning, providing comprehensive feedback
  - Backward compatible: Existing built-in validators continue working unchanged
  - Type-safe: Interface-based design ensures type safety while allowing flexible validation logic

## Usage Example
```golang
  type CustomValidator struct{}

  func (v *CustomValidator) Validate(res model.Resource) validators.ValidationError {
      // ... custom validation logic ...
      return verr
  }

  // Register for a specific resource type
  registry.RegisterTypeValidator(
      meshaccesslog.MeshAccessLogResourceTypeDescriptor,
      &CustomValidator{},
  )
```

Or real example: https://github.com/Kong/kong-mesh/pull/8719/commits/fa5736354838d8a2be1af5f07d64e9e587224a5c

## Why global registry? 

Since on Kubernetes we don’t create resources through the ResourceStore but directly via the Kubernetes API, and validation happens in the admission webhook, we can’t simply add validation to the ResourceManager. In the case of the global registry, however, we can attach a validator to all stores.

### DIFF
--- a/pkg/core/resources/validator/validate.go
+++ b/pkg/core/resources/validator/validate.go
@@ -31,5 +31,5 @@ func Validate(resource model.Resource) error {
 			}
 		}
 	}
-	return nil
+	return verr.OrNil()
 }


lukidzi:
Please take a look at: https://github.com/kumahq/kuma/pull/15128 as it is alternative approach

pkg/core/resources/apis/donothingresource/api/v1alpha1/zz_generated.resource.go
```diff
@@ -101,11 +103,23 @@ func (t *DoNothingResourceResource) Descriptor() model.ResourceTypeDescriptor {
...
+		if err := v.validate(); err != nil {
+			if validationErr, ok := err.(*validators.ValidationError); ok {
+				verr.Add(*validationErr)
+			} else {
+				verr.AddViolationAt(validators.Root(), err.Error())
+			}
+		}
+	}
+	// Run additional registered validators from global registry
+	validators := registry.Global().GetValidators(DoNothingResourceType)
```
lobkovilya: you can probably add this to https://github.com/kumahq/kuma/blob/7ddae55f76ddf5164a2f0d9788497629def1ae1f/pkg/core/resources/model/resource.go#L125-L130 instead of updating every generated resource

pkg/core/resources/registry/registry.go
```diff
@@ -116,6 +125,14 @@ func (t *typeRegistry) RegisterType(res model.ResourceTypeDescriptor) error {
 	return nil
 }
 
+func (t *typeRegistry) RegisterValidator(res model.ResourceTypeDescriptor, validator AdditionalValidator) {
+	t.validators[res.Name] = append(t.validators[res.Name], validator)
```
lobkovilya, line 128: can we guarantee all the writes to this map happen before all the reads? if not, then we probably need some form of synchronization

lukidzi: RegisterValidator should be called on bootstrap - but we can add sync to avoid a potential issue

pkg/core/resources/apis/donothingresource/api/v1alpha1/zz_generated.resource.go
lukidzi: I think it could be possible, but the issue is the circular dependency, since it requires `model.ResourceTypeDescriptor`, which is defined in the resource itself. This would require a refactor and reorganisation.

pkg/core/resources/registry/registry.go
lukidzi: After reviewing the code, there is no possibility of a race condition because we initialize the bootstrap phase before starting the servers. This ensures that runtime is fully initialized before serving begins.

lobkovilya (approved):

PR #15050, opened 2025-11-20
docs(MADR): disallow multiple meshes per kubernetes namespace

lobkovilya:
## Motivation

Having multiple meshes in a single namespace causes collisions in workload controller. See `Context and Problem Statement` for more details.

### DIFF
--- a/docs/madr/decisions/093-disallow-multiple-meshes-per-k8s-ns.md
+++ b/docs/madr/decisions/093-disallow-multiple-meshes-per-k8s-ns.md
@@ -30,14 +30,15 @@ which cannot behave correctly when a single namespace is spread across multiple 
 
 Add detection to `workload_controller.go` for multiple meshes per namespace and:
 - Skip Workload resource generation when multiple meshes are detected in a namespace
-- Emit a Kubernetes event to inform the user about the configuration issue
+- Emit a Kubernetes event to inform the user about the configuration issue and log the CP error
 
 ### Add Runtime Flag with Future Default Change
 
 Add a configuration flag `runtime.kubernetes.disallowMultipleMeshesPerNamespace` that:
 - Disabled by default (current behavior)
 - When enabled, prevents pods from using `kuma.io/mesh` label that would result in multiple meshes in a namespace
-- Flip the default to enabled in the next major Kuma release, making it a breaking change
+- Flip the default to enabled in the next major Kuma release to persist this behaviour,
+remove the feature flag, document it's a breaking change
 
 ### Pros and Cons
 


docs/madr/decisions/093-disallow-multiple-meshes-per-k8s-ns.md
```diff
@@ -0,0 +1,52 @@
...
+In practice the main blocker is Workload resource generation,
+which cannot behave correctly when a single namespace is spread across multiple meshes.
+
+## Design
+
+### Warn and Skip Workload Generation
+
+Add detection to `workload_controller.go` for multiple meshes per namespace and:
+- Skip Workload resource generation when multiple meshes are detected in a namespace
+- Emit a Kubernetes event to inform the user about the configuration issue
```
lahabana: CP error as well right?

docs/madr/decisions/093-disallow-multiple-meshes-per-k8s-ns.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Add Runtime Flag with Future Default Change
+
+Add a configuration flag `runtime.kubernetes.disallowMultipleMeshesPerNamespace` that:
+- Disabled by default (current behavior)
+- When enabled, prevents pods from using `kuma.io/mesh` label that would result in multiple meshes in a namespace
+- Flip the default to enabled in the next major Kuma release, making it a breaking change
```
lahabana: I wouldn't flip the default. I would just remove the ability altogether :) So flip the default and remove the FF.

docs/madr/decisions/093-disallow-multiple-meshes-per-k8s-ns.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Pros and Cons
+
+- Good, because it avoids breaking changes for existing users.
+- Good, because it gradually enforces a clean namespace to mesh relationship.
+- Bad, because users who do not check logs or events may be confused why a Workload was not generated.
```
lahabana, line 47: Would it make sense to generate a workload with the info in the Status?

lobkovilya: In my opinion this does not make sense. If there are two meshes in the same namespace, `m1` and `m2`, and we arbitrarily pick `m1` mesh to still generate a Workload, then the info message in that Workload is only useful for `m2` users. However, they will never see it, because the message lives in the `m1` mesh.

lahabana: makes sense

lahabana (approved):

PR #15057, opened 2025-11-20
test(e2e): refactor dataplane yaml building to use a go-template

lukidzi:
## Motivation

Refactors test framework dataplane YAML generation from `fmt.Sprintf` with positional string parameters to Go's text/template package with structured template data. This makes the test framework easier to extend, more maintainable, and less error-prone.

## Implementation information

- Introduced template data structures:
  - DataplaneTemplateData - Unified dataplane configuration
  - ZoneIngressTemplateData - Zone ingress configuration
  - ZoneEgressTemplateData - Zone egress configuration
- Removed old template constants that are now replaced by the template system
- Kept only constants still in use by other parts of the codebase
- Changed deprecated `--time` to `--timeout`
- Introduced a new field labels for additional labels

### DIFF
--- a/test/framework/interface.go
+++ b/test/framework/interface.go
@@ -600,6 +600,7 @@ func WithAppDockerRunOptions(options []string) AppDeploymentOption {
 
 func WithLabels(labels map[string]string) AppDeploymentOption {
 	return AppOptionFunc(func(o *appDeploymentOptions) {
+		o.labels = make(map[string]string)
 		for k, v := range labels {
 			o.labels[k] = v
 		}


test/framework/interface.go
```diff
@@ -597,6 +598,14 @@ func WithAppDockerRunOptions(options []string) AppDeploymentOption {
 	})
 }
 
+func WithLabels(labels map[string]string) AppDeploymentOption {
+	return AppOptionFunc(func(o *appDeploymentOptions) {
+		for k, v := range labels {
+			o.labels[k] = v
+		}
+	})
+}
```
bartsmykla, line 608: potential nil map panic - the function assigns to `o.labels[k]` without initializing the map first. Need to add `o.labels = make(map[string]string)` or check if it's nil before assignment

bartsmykla (approved):

Issue #14975, opened 2025-11-13
Auto generate Workload resource on Universal


Automaat:
## Motivation

  Workload resources are currently auto-generated on Kubernetes via the WorkloadReconciler
  (pkg/plugins/runtime/k8s/controllers/workload_controller.go:42), which watches Dataplanes and creates Workload resources based on the `kuma.io/workload`
   annotation. However, on Universal deployments, there's no equivalent mechanism to generate Workload resources from Dataplanes.

  This creates an inconsistency between K8s and Universal environments - K8s users automatically get Workload resources, while Universal users don't.  Since Workloads are part of the mesh resource model and may be referenced by policies or used for observability/management, Universal deployments need automatic Workload generation similar to how MeshService generation works (pkg/core/resources/apis/meshservice/generate/generator.go:39).

  ## Implementation information

  Follow the MeshService generator pattern at pkg/core/resources/apis/meshservice/generate/:

  ### Core components needed:

  1. **Generator struct** (`pkg/core/resources/apis/workload/generate/generator.go`):
     - Periodic ticker (configurable interval, similar to MeshService's generateInterval)
     - Deletion grace period support
     - Prometheus metrics for generation duration
     - ResourceManager and MeshCache integration
     - Zone information for cross-zone scenarios

  2. **Generation logic**:
     - Iterate through Dataplanes per mesh (similar to generator.go:174-304)
     - Extract workload identifier from Dataplane labels
     - Generate Workload resource for each unique workload identifier
     - Set `kuma.io/managed-by: workload-generator` label to track generated resources
     - Set appropriate labels: mesh, zone, environment (universal), resource-origin

  3. **Lifecycle management**:
     - Create new Workloads when Dataplanes appear
     - Update Workloads if spec changes
     - Deletion with grace period when no Dataplanes reference the workload (generator.go:225-254)
     - Handle orphaned Workloads (similar to controller's cleanupOrphanedWorkloads)

  4. **Setup/registration** (`pkg/core/resources/apis/workload/generate/component.go`):
     - Only run on Zone CP (skip Global mode)
     - Only run for Universal (skip K8s store type)
     - Check if workloads are enabled in CoreResources config
     - Register as ResilientComponent
     - Configuration: generation interval, deletion grace period

  5. **Configuration** (add to runtime config):
     ```go
     type WorkloadConfig struct {
         GenerationInterval   config_types.Duration
         DeletionGracePeriod  config_types.Duration
     }

  Key differences from MeshService generator:

[...]

PR #15058, opened 2025-11-20
feat(workload): auto generate Workload resource on Universal

Automaat:
## Motivation

  Auto-generate Workload resources from Dataplanes on Universal deployments. Enables gradual migration from Dataplane to Workload resource model, similar to MeshService generation.

  ## Implementation information

  Added generator component that:
  - Runs only on Universal (zone) CP
  - Watches Dataplanes with `kuma.io/workload` label
  - Creates Workload resources with `kuma.io/managed-by: workload-generator` label
  - Implements deletion grace period (default 1h) to avoid flapping
  - Configurable via `KUMA_WORKLOAD_GENERATION_INTERVAL` (default 2s) and `KUMA_WORKLOAD_DELETION_GRACE_PERIOD`

  Follows same pattern as MeshService generator.

  ## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/14975

### DIFF
--- a/pkg/api-server/resource_endpoints.go
+++ b/pkg/api-server/resource_endpoints.go
@@ -453,9 +453,14 @@ func (r *resourceEndpoints) createResource(
 	res.SetMeta(resRest.GetMeta())
 
 	// Validate workload label on Universal Zone dataplanes
-	if r.descriptor.Name == core_mesh.DataplaneType {
+	if r.descriptor.Name == core_mesh.DataplaneType && !r.isK8s {
 		if resLabels := res.GetMeta().GetLabels(); resLabels != nil {
 			if workloadName, ok := resLabels[metadata.KumaWorkload]; ok && workloadName != "" {
+				if r.mode == config_core.Global {
+					err := rest_errors.NewBadRequestError("labels[\"kuma.io/workload\"]: not allowed on Global control plane")
+					rest_errors.HandleError(ctx, response, err, "Invalid workload label")
+					return
+				}
 				if validationErrs := apimachineryvalidation.NameIsDNS1035Label(workloadName, false); len(validationErrs) != 0 {
 					err := rest_errors.NewBadRequestError(fmt.Sprintf("labels[\"kuma.io/workload\"]: must be a valid DNS-1035 label (at most 63 characters, matching regex [a-z]([-a-z0-9]*[a-z0-9])?): %s", strings.Join(validationErrs, "; ")))
 					rest_errors.HandleError(ctx, response, err, "Invalid workload label")
--- a/pkg/core/resources/apis/workload/generate/component.go
+++ b/pkg/core/resources/apis/workload/generate/component.go
@@ -14,7 +14,7 @@ func Setup(rt runtime.Runtime) error {
 	if rt.GetMode() == config_core.Global {
 		return nil
 	}
-	if storeType := rt.Config().Store.Type; storeType == store.KubernetesStore {
+	if rt.Config().Store.Type == store.KubernetesStore {
 		return nil
 	}
 	logger := core.Log.WithName("workload").WithName("generator")
--- a/pkg/core/resources/apis/workload/generate/generator.go
+++ b/pkg/core/resources/apis/workload/generate/generator.go
@@ -129,7 +129,7 @@ func (g *Generator) createWorkload(ctx context.Context, log logr.Logger, workloa
 		log.Error(err, "couldn't create Workload")
 		return
 	}
-	log.Info("created Workload")
+	log.V(1).Info("created Workload")
 }
 
 func (g *Generator) NeedLeaderElection() bool {
@@ -155,6 +155,7 @@ func (g *Generator) Start(stop <-chan struct{}) error {
 			if err != nil {
 				// Error from AggregateMeshContexts indicates a fundamental issue
 				// with resource manager or mesh cache, so we terminate the component
+				g.logger.Error(err, "fundamental issue with resource manager or mesh cache, terminating component")
 				return err
 			}
 			for mesh, meshCtx := range aggregatedMeshCtxs.MeshContextsByName {


pkg/config/app/kuma-cp/kuma-cp.defaults.yaml
```diff
@@ -868,3 +868,8 @@ meshService:
   generationInterval: 2s # ENV: KUMA_MESH_SERVICE_GENERATION_INTERVAL
   # How long we wait before deleting a MeshService if all Dataplanes are gone
   deletionGracePeriod: 1h # ENV: KUMA_MESH_SERVICE_DELETION_GRACE_PERIOD
+workload:
```
lobkovilya: Shouldn't this be under the `runtime.universal`, it has no effect on k8s right?

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,202 @@
...
+				return err
+			}
+			for mesh, meshCtx := range aggregatedMeshCtxs.MeshContextsByName {
+				dataplanes := meshCtx.Resources.Dataplanes()
+				workloads := meshCtx.Resources.Workloads()
+				g.generate(ctx, mesh, dataplanes.Items, workloads.Items)
+			}
+			g.metric.Observe(float64(time.Since(start).Milliseconds()))
+
+		case <-stop:
```
lobkovilya, line 168: you can speed up stop by calling it from different goroutine with ctx `cancel` func, i.e. https://github.com/kumahq/kuma/blob/9dbe37ffdbadc028103429f7d2927822aa04c4f1/pkg/intercp/tls/defaults.go#L29-L32

otherwise you can spend some time building the AggregatedMeshContext without knowing the component is stopped

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,202 @@
...
+	}, nil
+}
+
+// workloadForDataplane extracts and validates the workload identifier from a dataplane
+func (g *Generator) workloadForDataplane(dataplane *core_mesh.DataplaneResource) (string, bool) {
+	workloadName, ok := dataplane.GetMeta().GetLabels()[metadata.KumaWorkload]
+	if !ok || workloadName == "" {
+		return "", false
+	}
+	allErrs := apimachineryvalidation.NameIsDNS1035Label(workloadName, false)
```
lobkovilya, line 74: Do we need to check it as well when DPP is created?

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	return workloadName, true
+}
+
+func (g *Generator) generate(ctx context.Context, mesh string, dataplanes []*core_mesh.DataplaneResource, workloads []*workload_api.WorkloadResource) {
+	log := g.logger.WithValues("mesh", mesh)
+	workloadsByName := map[string]bool{}
+	for _, dataplane := range core_mesh.SortDataplanes(dataplanes) {
+		if workloadName, ok := g.workloadForDataplane(dataplane); ok {
+			log.V(1).Info("will generate Workload for dataplane", "dataplane", dataplane.GetMeta().GetName(), "workload", workloadName)
```
lobkovilya: It's too noisy even for debug level, no? It basically runs every 2sec for every DPP with `kuma.io/workload` label

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,202 @@
...
+			workloadsByName[workloadName] = true
+		}
+	}
+	log.V(1).Info("processing workloads", "workloadCount", len(workloadsByName), "dataplaneCount", len(dataplanes))
```
lobkovilya: same

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,202 @@
...
+
+	for _, workload := range workloads {
+		if managedBy, ok := workload.GetMeta().GetLabels()[mesh_proto.ManagedByLabel]; !ok || managedBy != managedByValue {
+			continue
+		}
+		log := log.WithValues("Workload", workload.GetMeta().GetName())
+		workloadName := workload.GetMeta().GetName()
+		gracePeriodStartedAtText, hasGracePeriodLabel := workload.GetMeta().GetLabels()[mesh_proto.DeletionGracePeriodStartedLabel]
```
lobkovilya: Do we need something like that on Kubernetes? I think if you scale the number of replicas to zero and back then workload will be deleted and recreated as well

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,202 @@
...
+			newLabels := maps.Clone(workload.GetMeta().GetLabels())
+			delete(newLabels, mesh_proto.DeletionGracePeriodStartedLabel)
+
+			if err := g.resManager.Update(ctx, workload, store.UpdateWithLabels(newLabels)); err != nil {
+				log.Error(err, "couldn't update Workload")
+				continue
+			}
+			log.Info("updated Workload", "reason", "no longer scheduled for deletion")
+		} else if !stillExists {
+			gracePeriodStartedAt, err := time.Parse(time.RFC3339, gracePeriodStartedAtText)
```
lobkovilya: you shouldn't call `time.Parse` when you know `hasGracePeriodLabel` can be false at that point and then use `err` to control the program flow

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,202 @@
...
+				newLabels[mesh_proto.DeletionGracePeriodStartedLabel] = string(nowText)
+				if err := g.resManager.Update(ctx, workload, store.UpdateWithLabels(newLabels)); err != nil {
+					log.Error(err, "couldn't update Workload")
+					continue
+				}
+				log.Info("Workload deletion grace period started", "period", g.deletionGracePeriod.String())
+			}
+		}
+		delete(workloadsByName, workloadName)
+	}
```
lobkovilya, line 100: please do something with this loop it's just unreadable and hard to follow what's going on with this amount of nested if/else

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,202 @@
...
+	for _, dataplane := range core_mesh.SortDataplanes(dataplanes) {
+		if workloadName, ok := g.workloadForDataplane(dataplane); ok {
+			log.V(1).Info("will generate Workload for dataplane", "dataplane", dataplane.GetMeta().GetName(), "workload", workloadName)
+			workloadsByName[workloadName] = true
+		}
+	}
+	log.V(1).Info("processing workloads", "workloadCount", len(workloadsByName), "dataplaneCount", len(dataplanes))
+
+	for _, workload := range workloads {
+		if managedBy, ok := workload.GetMeta().GetLabels()[mesh_proto.ManagedByLabel]; !ok || managedBy != managedByValue {
```
lobkovilya: why are we doing this? there is no flow when user would need to create Workload manually

pkg/config/app/kuma-cp/kuma-cp.defaults.yaml
Automaat: changed

pkg/core/resources/apis/workload/generate/generator.go
Automaat: updated

pkg/core/resources/apis/workload/generate/generator.go
Automaat: removed debug logs

pkg/core/resources/apis/workload/generate/generator.go
Automaat: removed

pkg/core/resources/apis/workload/generate/generator.go
Automaat: removed

pkg/core/resources/apis/workload/generate/generator.go
Automaat: fixed

pkg/core/resources/apis/workload/generate/generator.go
Automaat: removed whole grace period logic

pkg/core/resources/apis/workload/generate/generator.go
Automaat: removed enitrely

pkg/core/resources/apis/workload/generate/generator.go
Automaat: moved to dpp creation

pkg/api-server/resource_endpoints.go
```diff
@@ -450,6 +452,19 @@ func (r *resourceEndpoints) createResource(
 	_ = res.SetSpec(resRest.GetSpec())
 	res.SetMeta(resRest.GetMeta())
 
+	// Validate workload label on Universal Zone dataplanes
+	if r.mode != config_core.Global && r.descriptor.Name == core_mesh.DataplaneType {
```
lobkovilya: Do we allow DPP creating on Global?

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,181 @@
...
+func (g *Generator) generate(ctx context.Context, mesh string, dataplanes []*core_mesh.DataplaneResource, workloads []*workload_api.WorkloadResource) {
+	log := g.logger.WithValues("mesh", mesh)
+
+	workloadsByName := g.collectWorkloadsByName(dataplanes)
+
+	for _, workload := range workloads {
+		workloadName := workload.GetMeta().GetName()
+		stillExists := workloadsByName[workloadName]
+
+		g.handleExistingWorkload(ctx, log, workload, stillExists)
```
lobkovilya: Func name `handleExistingWorkload` already implies `workload` exists, passing `stillExists` just to return is a bit odd.

I'd just inline the call to `handleExistingWorkload`.

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (g *Generator) createWorkload(ctx context.Context, log logr.Logger, workloadName, mesh string) {
+	log = log.WithValues("Workload", workloadName)
+	workload := workload_api.NewWorkloadResource()
+	// Workload spec is intentionally empty here - the actual spec is computed
+	// separately by the workload status controller based on dataplane state
+	workload.Spec = &workload_api.Workload{}
+	if err := g.resManager.Create(ctx, workload, store.CreateByKey(workloadName, mesh), store.CreateWithLabels(map[string]string{
+		metadata.KumaMeshLabel:         mesh,
+		mesh_proto.DisplayName:         workloadName,
```
lobkovilya: isn't it set automatically by the ResourceManager?

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,181 @@
...
+		mesh_proto.ManagedByLabel:      managedByValue,
+		mesh_proto.EnvTag:              mesh_proto.UniversalEnvironment,
+		mesh_proto.ZoneTag:             g.zone,
+		mesh_proto.ResourceOriginLabel: string(mesh_proto.ZoneResourceOrigin),
```
lobkovilya: same, not sure if we need to explicitly set these labels

pkg/core/resources/apis/workload/generate/component.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/kumahq/kuma/v2/pkg/core"
+	"github.com/kumahq/kuma/v2/pkg/core/runtime"
+	"github.com/kumahq/kuma/v2/pkg/core/runtime/component"
+)
+
+func Setup(rt runtime.Runtime) error {
+	if rt.GetMode() == config_core.Global {
+		return nil
+	}
+	if storeType := rt.Config().Store.Type; storeType == store.KubernetesStore {
```
slonka: Why do we need this weird assignment? can't we just do:


```suggestion
	if rt.Config().Store.Type == store.KubernetesStore {
```

pkg/core/resources/apis/workload/generate/component.go
```diff
@@ -0,0 +1,42 @@
...
+		return nil
+	}
+	logger := core.Log.WithName("workload").WithName("generator")
+	if !slices.Contains(rt.Config().CoreResources.Enabled, "workloads") {
```
slonka, line 21: I assume we have no intention of making it possible to enable resources in runtime?

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,172 @@
...
+	if err := g.resManager.Create(ctx, workload, store.CreateByKey(workloadName, mesh), store.CreateWithLabels(map[string]string{
+		metadata.KumaMeshLabel:    mesh,
+		mesh_proto.ManagedByLabel: managedByValue,
+		mesh_proto.EnvTag:         mesh_proto.UniversalEnvironment,
+		mesh_proto.ZoneTag:        g.zone,
+	})); err != nil {
+		log.Error(err, "couldn't create Workload")
+		return
+	}
+	log.Info("created Workload")
```
slonka: isn't this going to be too verbose?

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -0,0 +1,172 @@
...
+	}()
+
+	for {
+		select {
+		case <-ticker.C:
+			start := time.Now()
+			aggregatedMeshCtxs, err := xds_context.AggregateMeshContexts(ctx, g.resManager, g.meshCache.GetMeshContext)
+			if err != nil {
+				// Error from AggregateMeshContexts indicates a fundamental issue
+				// with resource manager or mesh cache, so we terminate the component
```
slonka, line 157: shouldn't we log something then? if this is fundamental issue? is this a "developer issue" i.e. a bug or is this something wrong with the environment?

PR #15065, opened 2025-11-21
feat(actions): stop rebuilding kuma in every e2e job

Automaat:
## Motivation

CI runs take 40+ min per PR due to duplicate binary builds. Each of 10-12 E2E jobs rebuilds all binaries (`make build` + `make build/distributions`), wasting ~40 min total. The `check` job already builds binaries but they're discarded.

## Implementation information

**Artifact reuse pattern:**
1. `check` job: After `make check`, run `make build` and upload artifacts (1-day retention)
2. E2E jobs: Download pre-built artifacts, skip `make build` and `make build/distributions`  
3. `build-binaries` job: Download artifacts, only rebuild if `FULL_MATRIX=true` (for arm64)

**Changes:**
- [build-test-distribute.yaml](build-test-distribute.yaml#L111-L120): Build + upload after check
- [_e2e.yaml](_e2e.yaml#L38-L41): Download artifacts, removed build steps (saves ~4 min/job)
- [_build_publish.yaml](_build_publish.yaml#L56-L63): Download artifacts, conditional rebuild

**Performance impact:**
- E2E jobs: 10-12 jobs × 4 min = **40-48 min saved**
- build-binaries: ~3 min saved (PRs skip rebuild)
- check job: +3 min (one-time build)
- **Net: ~40-45 min faster per PR**

Alternative considered: Build in separate job before test. Discarded - adds dependency chain latency, check job already has build context.

### DIFF
--- a/.github/workflows/_build_publish.yaml
+++ b/.github/workflows/_build_publish.yaml
@@ -57,6 +57,7 @@ jobs:
         uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
         with:
           name: build-artifacts-${{ runner.os }}-${{ runner.arch }}
+          path: build
       - name: "Build binaries for other architectures if needed"
         if: ${{ fromJSON(inputs.FULL_MATRIX) }}
         run: |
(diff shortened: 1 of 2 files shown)


Automaat:
does not work

Issue #15081, opened 2025-11-21
Add runtime flag runtime.kubernetes.disallowMultipleMeshesPerNamespace


Automaat:
## Description:

  Add configuration flag to prevent pods from using kuma.io/mesh label that would result in multiple meshes in a namespace.

##  Requirements:

  - Add flag runtime.kubernetes.disallowMultipleMeshesPerNamespace
  - Default: disabled (preserves current behavior)
  - When enabled: prevent pods from creating multi-mesh namespace configuration
  - Validation should happen during pod admission/reconciliation

##  Context:

  Allows gradual enforcement of single mesh per namespace constraint without breaking existing deployments.

##  Reference:

  MADR: [docs/madr/decisions/093-disallow-multiple-meshes-per-k8s-ns.md](https://github.com/kumahq/kuma/blob/4ade29f83a9c1706bdae062c48db83c08a7007da/docs/madr/decisions/093-disallow-multiple-meshes-per-k8s-ns.md#L37)

PR #15102, opened 2025-11-25
feat(kuma-cp): add disallowMultipleMeshesPerNamespace to block multi-mesh namespaces

slonka:
## Motivation

Adds `runtime.kubernetes.disallowMultipleMeshesPerNamespace` flag to prevent pods from being created in namespaces that already have pods in different meshes, addressing Workload resource collision issues.

## Implementation information

**Webhook validation:** When flag enabled, admission webhook rejects pod create/update if namespace already contains dataplanes in different mesh. Uses label selector (`kuma.io/mesh != podMesh`) with `Limit(1)` for efficient API-level filtering instead of client-side iteration.

**Config:** Adds `runtime.kubernetes.disallowMultipleMeshesPerNamespace` boolean config (default: false) for backward compatibility.

**Note:** Workload controller multi-mesh detection already in master - this PR adds webhook-based prevention.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/15081

### DIFF
--- a/pkg/plugins/runtime/k8s/plugin.go
+++ b/pkg/plugins/runtime/k8s/plugin.go
@@ -348,7 +348,7 @@ func addValidators(mgr kube_ctrl.Manager, rt core_runtime.Runtime, converter k8s
 	mgr.GetWebhookServer().Register("/validate-v1-secret", &kube_webhook.Admission{Handler: secretValidator})
 
 	if rt.Config().Mode != config_core.Global {
-		podValidator := k8s_webhooks.NewPodValidatorWebhook(admissionDecoder)
+		podValidator := k8s_webhooks.NewPodValidatorWebhook(admissionDecoder, client, rt.Config().Runtime.Kubernetes.DisallowMultipleMeshesPerNamespace)
 		mgr.GetWebhookServer().Register("/validate-v1-pod", &kube_webhook.Admission{Handler: podValidator})
 	}
 


pkg/plugins/runtime/k8s/webhooks/pod_validator.go
```diff
@@ -40,5 +49,41 @@ func (h *PodValidator) ValidatePod(_ context.Context, req admission.Request) adm
...
+	if err := h.client.Get(ctx, kube_client.ObjectKey{Name: pod.Namespace}, ns); err != nil {
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed to get namespace: %w", err))
+	}
+
+	// Get the mesh for this pod using proper precedence
+	podMesh := util.MeshOfByLabelOrAnnotation(logr.Discard(), pod, ns)
+
+	// List all dataplanes in the namespace to check for different meshes
+	dataplanes := &mesh_k8s.DataplaneList{}
+	if err := h.client.List(ctx, dataplanes, kube_client.InNamespace(pod.Namespace)); err != nil {
```
Automaat: what is the performance impact of this? can this be somehow improved?

Automaat: changed to only selecting pod with different mesh then current

PR #15113, opened 2025-11-27
feat(config): disable virtual probes by default

Automaat:
## Motivation

Virtual Probes deprecated in 2.9.x in favor of Application Probe Proxy, which supports more probe types (HTTPGet, TCPSocket, gRPC). Setting default to false completes deprecation path.

## Implementation information

- Changed default: `virtualProbesEnabled: true` → `false` in config defaults and Go code
- Fixed probe proxy logic: Application Probe Proxy now defaults to enabled when no virtual probes annotation present
- Added UPGRADE.md section for 2.13.x with migration guidance
- Updated 93 test/golden files to reflect new default

**Probe Proxy Fix Details:**
When virtual probes were disabled by default, the probe proxy logic was incorrectly disabling Application Probe Proxy too. Modified `pkg/plugins/runtime/k8s/probes/pod_probe_overrider.go` to only disable Application Probe Proxy when virtual probes are **explicitly** disabled via annotation (`kuma.io/virtual-probes: false`). If no annotation exists (the common case now), Application Probe Proxy defaults to enabled on port 9001.

This maintains backward compatibility while enabling the recommended probe solution by default.

## Supporting documentation

- Virtual Probes deprecated in 2.9.x
- Application Probe Proxy is replacement (enabled by default on port 9001)

### DIFF
--- a/pkg/plugins/runtime/k8s/probes/pod_probe_overrider.go
+++ b/pkg/plugins/runtime/k8s/probes/pod_probe_overrider.go
@@ -135,12 +135,13 @@ func SetApplicationProbeProxyPortAnnotation(annotations metadata.Annotations, po
 	}
 
 	// scenarios of switching between virtual probes (vp) and application probe proxy (proxy):
-	// vp   proxy    	result
-	// Y    Y      	     proxy
-	// Y    N            vp
-	// N    N      	     proxy (default)
-	// N    Y - config   proxy
-	// N    Y - pod      proxy
+	// vp           proxy    	result
+	// Y            Y      	     proxy
+	// Y            N            proxy (default)
+	// disabled     N      	     N (disabled)
+	// not set      N      	     proxy (default)
+	// not set      Y - config   proxy
+	// not set      Y - pod      proxy
 
 	proxyPortAnno, proxyPortAnnoExists, err := metadata.Annotations(podAnnotations).GetUint32(metadata.KumaApplicationProbeProxyPortAnnotation)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


bartsmykla (dismissed):

pkg/plugins/runtime/k8s/probes/pod_probe_overrider.go
```diff
@@ -179,13 +173,10 @@ func GetApplicationProbeProxyPort(
 	// vp   proxy    	result
 	// Y    Y      	     proxy
 	// Y    N            vp
-	// N    N      	     N
-	// N    Y - config   N
+	// N    N      	     proxy (default)
```
lobkovilya: I'm not sure I fully understand this part. Is it okay now there is no way to have `result: N`?

Automaat: Good catch! The table was misleading. There **is** a way to disable (result: N) - by setting `kuma.io/virtual-probes: disabled`. The table just didn't show it.

I've updated the table to be clearer:
```
// vp           proxy    	result
// Y            Y      	     proxy
// Y            N            proxy (default)
// disabled     N      	     N (disabled)  ← this was missing
// not set      N      	     proxy (default)
// not set      Y - config   proxy
// not set      Y - pod      proxy
```

So users can disable app probe proxy by:
1. Gateway mode (automatically disabled)
2. Setting `kuma.io/virtual-probes: disabled` annotation

slonka (approved):

PR #15128, opened 2025-11-28
feat(api): introduce possibility to register custom validators

lukidzi:
## Motivation

Usage example: https://github.com/Kong/kong-mesh/commit/214be5ed2451e2be880e5b33090a1bb468c4323a

### DIFF
--- a/pkg/core/bootstrap/bootstrap.go
+++ b/pkg/core/bootstrap/bootstrap.go
@@ -404,7 +404,8 @@ func initializeAPIServerAuthenticator(builder *core_runtime.Builder) error {
 
 func initializeResourceManager(cfg kuma_cp.Config, builder *core_runtime.Builder) error {
 	defaultManager := core_manager.NewResourceManager(builder.ResourceStore())
-	customizableManager := core_manager.NewCustomizableResourceManager(defaultManager, nil)
+	validatingManager := core_manager.NewValidationManager(defaultManager, builder.CustomValidators())
+	customizableManager := core_manager.NewCustomizableResourceManager(validatingManager, nil)
 
 	customizableManager.Customize(
 		mesh.MeshType,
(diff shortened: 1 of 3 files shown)


pkg/core/bootstrap/bootstrap.go
```diff
@@ -404,7 +404,8 @@ func initializeAPIServerAuthenticator(builder *core_runtime.Builder) error {
 
 func initializeResourceManager(cfg kuma_cp.Config, builder *core_runtime.Builder) error {
 	defaultManager := core_manager.NewResourceManager(builder.ResourceStore())
-	customizableManager := core_manager.NewCustomizableResourceManager(defaultManager, nil)
+	validatingManager := core_manager.NewValidationManager(defaultManager, builder.CustomValidators())
+	customizableManager := core_manager.NewCustomizableResourceManager(validatingManager, nil)
```
lukidzi, line 408: We can also do not introduce a new interface and instead pass `builder.CustomValidators()` to `NewResourceManager` and call it in `Create` and `Update`

lobkovilya: But custom resource managers are bypassed on Kubernetes, no?

lukidzi: I don't think so, based on the code seems we use ResourceManager with all Stores

lukidzi: After discussing with @lobkovilya, we realized that this approach won’t work on Kubernetes because the validation logic is executed inside the validation [webhook](https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/webhooks/validation.go#L76), and the resource is not created via the ResourceStore.

lukidzi:
After discussing with @lobkovilya, we realized that this approach won’t work on Kubernetes because the validation logic is executed inside the validation [webhook](https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/webhooks/validation.go#L76), and the resource is not created via the ResourceStore.

PR #15163, opened 2025-12-03
fix(xds): restrict inbound traffic to only inbound ports

lukidzi:
## Motivation

When transparent proxy is enabled, the sidecar previously accepted inbound traffic on any port, not just those explicitly defined in the dataplane configuration. This created a security risk where services could be accessed on unintended ports. This change introduces strict port filtering for inbound traffic to improve security by ensuring that only explicitly configured inbound ports accept traffic. This prevents unauthorized access to services through undeclared ports.

## Implementation information

  - Added `StrictInboundPortsEnabled` configuration flag to `kuma-dp` config (defaults to `true`)
  - Introduced `FeatureStrictInboundPorts` feature flag to communicate the setting from `kuma-dp` to control plane
  - Modified transparent proxy generator to create inbound passthrough listeners with filter chain matches for specific destination ports when the feature is enabled
  - Updated MeshTLS plugin to handle TLS mode mismatches by reconfiguring inbound passthrough listeners
  - When enabled, the inbound passthrough listener only accepts traffic on ports explicitly defined in the dataplane configuration
  - Added comprehensive test coverage for the new behavior
  - Introduced a no-destination cluster which handles the case where there is inbound - we cannot have empty filterChain

### Notes
I was worried about case of Metrics, Admin API and HealthProbes but they are not a concern anymore, as we exclude port for HealthProbes in transparent proxy, Metrics creates a real listener and AdminAPI also has a real listener. If you are aware of any inbound listener that might be virtual (other than inbounds defined in dataplane) please take a look.

fix: https://github.com/Kong/kong-mesh/issues/6516

### DIFF
--- a/test/e2e_env/universal/strictinbounds/strict_inbound.go
+++ b/test/e2e_env/universal/strictinbounds/strict_inbound.go
@@ -90,7 +90,7 @@ func StrictInboundPorts() {
 		// not secured DPP can be accessed
 		notSecuredDPPInboundAddress := net.JoinHostPort(universal.Cluster.GetApp("test-server-not-secure").GetIP(), "80")
 		notSecuredServiceAddress := net.JoinHostPort(universal.Cluster.GetApp("test-server-not-secure").GetIP(), "8080")
-Eventually(func(g Gomega) {
+		Eventually(func(g Gomega) {
 			resp, err := client.CollectEchoResponse(
 				universal.Cluster, "demo-client", "test-server-not-secure.mesh",
 			)
@@ -121,7 +121,6 @@ Eventually(func(g Gomega) {
 			g.Expect(err).ToNot(HaveOccurred())
 			g.Expect(resp.Instance).To(Equal("test-server-not-secure"))
 		}, "30s", "1s").Should(Succeed())
-
 	})
 
 	It("should allow all traffic when permissive mode", func() {
@@ -176,7 +175,7 @@ Eventually(func(g Gomega) {
 		// not secured DPP can be accessed
 		notSecuredDPPInboundAddress := net.JoinHostPort(universal.Cluster.GetApp("test-server-not-secure").GetIP(), "80")
 		notSecuredServiceAddress := net.JoinHostPort(universal.Cluster.GetApp("test-server-not-secure").GetIP(), "8080")
-Eventually(func(g Gomega) {
+		Eventually(func(g Gomega) {
 			resp, err := client.CollectEchoResponse(
 				universal.Cluster, "demo-client", "test-server-not-secure.mesh",
 			)


test/e2e_env/universal/strictinbounds/strict_inbound.go
```diff
@@ -0,0 +1,217 @@
...
+	AfterEachFailure(func() {
+		DebugUniversal(universal.Cluster, meshName)
+	})
+
+	E2EAfterAll(func() {
+		Expect(universal.Cluster.DeleteMeshApps(meshName)).To(Succeed())
+		Expect(universal.Cluster.DeleteMesh(meshName)).To(Succeed())
+	})
+
+	It("should allow all traffic when there is no tls", func() {
```
lobkovilya, line 52: I don't understand this test. Why `:80` is an inbound address, but `:8080` is a service address? Also shouldn't at least one of `eventually` check that instance is `test-server-not-secure`?

test/e2e_env/universal/strictinbounds/strict_inbound.go
```diff
@@ -0,0 +1,217 @@
...
+			g.Expect(resp.Instance).To(Equal("test-server"))
+		}, "30s", "1s").Should(Succeed())
+
+		// and
+		// the service port cannot be accessed
+		Eventually(func(g Gomega) {
+			_, err := client.CollectEchoResponse(
+				universal.Cluster, "demo-client-not-in-mesh", dppInboundAddress,
+			)
+			g.Expect(err).To(HaveOccurred())
```
lobkovilya: you should check errors with https://github.com/kumahq/kuma/blob/f6c7ece0ca5096fc1dc8002709a490bc27d173d9/test/framework/client/collect.go#L432

Checking `err` is incorrect because it's possible util code is failed, not the actual failure between client and server.

UPGRADE.md
```diff
@@ -8,6 +8,73 @@ does not have any particular instructions.
...
+Then set the annotation `kuma.io/container-patches` on deployment where it should be disabled
+
+```yaml
+"kuma.io/container-patches":"disable-strict-inbounds-ports"
+```
+
+or for all Deployments by setting control-plane configuration:
+
+```
+KUMA_RUNTIME_KUBERNETES_INJECTOR_CONTAINER_PATCHES="disable-strict-inbounds-ports"
```
Automaat: I would reorg this, first example of global disable and then specific one with ContainerPatch

test/e2e_env/universal/strictinbounds/strict_inbound.go
lukidzi: > Why :80 is an inbound address, but :8080 is a service address?

Our test server uses port 80 as the Dataplane inbound port, and port 8080 as the port where the application service is running. The goal of this test is to verify:
* that the inbound port of the Dataplane (:80) is reachable, and
* that the application port (:8080) cannot be accessed by bypassing the sidecar. This behavior should be allowed in no-TLS and permissive modes, but blocked in strict TLS mode.

pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin.go
```diff
@@ -274,6 +296,51 @@ func configureTLSParams(conf api.Conf, cluster *envoy_cluster.Cluster) error {
...
+	proxy *core_xds.Proxy,
+	xdsCtx xds_context.Context,
+	conf api.Conf,
+	ipv6 bool,
+) (envoy_common.NamedResource, error) {
+	tpCfg := proxy.GetTransparentProxy()
+	caBackend := xdsCtx.Mesh.Resource.GetEnabledCertificateAuthorityBackend()
+	if tpCfg == nil && caBackend == nil && proxy.WorkloadIdentity == nil && !proxy.Metadata.HasFeature(xds_types.FeatureStrictInboundPorts) {
+		return nil, nil
+	}
```
Automaat, line 312: **Critical: Nil pointer dereference**

Guard clause uses AND logic - returns early ONLY if ALL conditions true. When `tpCfg == nil` BUT (`caBackend != nil` OR `workloadIdentity != nil` OR feature enabled), code proceeds to lines 331/339 which dereference nil `tpCfg`:

```go
tpCfg.Redirect.Inbound.Port.Uint32()  // PANIC
```

**Fix**: Separate nil check:
```go
if tpCfg == nil {
    return nil, nil
}
if caBackend == nil && proxy.WorkloadIdentity == nil && !proxy.Metadata.HasFeature(xds_types.FeatureStrictInboundPorts) {
    return nil, nil
}
```

Impact: With feature now default-enabled, crash likely when transparent proxy disabled but feature flag set.

pkg/xds/generator/transparent_proxy_generator_test.go
```diff
@@ -206,5 +221,134 @@ var _ = Describe("TransparentProxyGenerator", func() {
 			meshServicesMode: mesh_proto.Mesh_MeshServices_Exclusive,
 			expected:         "05.envoy.golden.yaml",
 		}),
+		Entry("transparent_proxying=true,unified_naming=true,inbound_filter,strict", testCase{
```
Automaat, line 259: **Code duplication: 380+ lines can be reduced to ~50**

These 3 test entries (lines 224-351) are nearly identical, differing only in:
- Number of inbound ports (1 vs 2)
- TLS mode (STRICT/PERMISSIVE)
- Expected golden file

**Suggested refactor**:
```go
func strictInboundPortsProxy(inboundPorts []uint32) *model.Proxy {
    inbounds := make([]*mesh_proto.Dataplane_Networking_Inbound, len(inboundPorts))
    for i, port := range inboundPorts {
        inbounds[i] = &mesh_proto.Dataplane_Networking_Inbound{Port: port}
    }
    
    return &model.Proxy{
        Metadata: &model.DataplaneMetadata{Features: map[string]bool{
            types.FeatureUnifiedResourceNaming: true,
            types.FeatureStrictInboundPorts:    true,
        }},
        Id: *model.BuildProxyId("", "side-car"),
        Dataplane: &core_mesh.DataplaneResource{
            Meta: &test_model.ResourceMeta{Version: "v1"},
            Spec: &mesh_proto.Dataplane{
                Networking: &mesh_proto.Dataplane_Networking{
                    Inbound: inbounds,
                    TransparentProxying: &mesh_proto.Dataplane_Networking_TransparentProxying{
                        IpFamilyMode:         mesh_proto.Dataplane_Networking_TransparentProxying_DualStack,
                        RedirectPortOutbound: 15001,
                        RedirectPortInbound:  15006,
                    },
                },
            },
        },
        APIVersion: envoy_common.APIV3,
        Policies: model.MatchedPolicies{
            TrafficLogs: map[model.ServiceName]*core_mesh.TrafficLogResource{
                "some-service": {Spec: &mesh_proto.TrafficLog{Conf: &mesh_proto.TrafficLog_Conf{Backend: "file"}}},
            },
        },
        InternalAddresses: DummyInternalAddresses,
    }
}

// Tests become 4 lines each:
Entry("strict inbound filter with strict mTLS", testCase{
    proxy:    strictInboundPortsProxy([]uint32{8080}),
    tlsMode:  mesh_proto.CertificateAuthorityBackend_STRICT.Enum(),
    expected: "06.envoy.golden.yaml",
}),
Entry("strict inbound filter with permissive mTLS (2 ports)", testCase{
    proxy:    strictInboundPortsProxy([]uint32{8080, 9000}),
    tlsMode:  mesh_proto.CertificateAuthorityBackend_PERMISSIVE.Enum(),
    expected: "07.envoy.golden.yaml",
}),
```

**Impact**: 87% line reduction, improved maintainability

pkg/plugins/policies/meshtls/plugin/v1alpha1/plugin_test.go
```diff
@@ -218,6 +226,35 @@ var _ = Describe("MeshTLS", func() {
 				},
 			},
 		}),
+		Entry("strict with strict mTLS on the mesh (no mismatch, no passthrough listeners)", testCase{
```
Automaat: **Entry names could be more concise and behavior-focused**

Current names mix implementation details with expected outcomes. Suggestion:

```go
Entry("strict mode + strict mesh = no passthrough listeners", testCase{
    caseName:    "strict-with-strict-mtls",
    meshBuilder: samples.MeshMTLSBuilder(),
}),
Entry("permissive mode + strict mesh = passthrough listeners", testCase{
    caseName:    "permissive-with-strict-mtls",
    meshBuilder: samples.MeshMTLSBuilder(),
}),
Entry("workload identity without CA = passthrough listeners", testCase{
    caseName:    "strict-with-workload-identity-no-ca",
    // ...
}),
Entry("strict inbound ports feature = port filtering", testCase{
    caseName:    "strict-with-feature-strict-inbound-ports",
    // ...
}),
```

Clearer pattern: `condition = expected_behavior` makes test intent scannable

test/e2e_env/universal/strictinbounds/strict_inbound.go
Automaat, line 52: **Use Ginkgo DescribeTable for repetitive test scenarios**

Current: 4 identical `Eventually` blocks repeated across 3 test cases (~200 lines of duplication)

**Refactor using DescribeTable/Entry**:
```go
It("should allow all traffic when there is no tls", func() {
    app := universal.Cluster.GetApp("test-server")
    
    type testCase struct {
        from  string
        to    string
        works bool
    }
    
    DescribeTable("connectivity scenarios",
        func(tc testCase) {
            Eventually(func(g Gomega) {
                resp, err := client.CollectEchoResponse(universal.Cluster, tc.from, tc.to)
                if tc.works {
                    g.Expect(resp.Instance).To(ContainSubstring("test-server"))
                } else {
                    g.Expect(err).To(HaveOccurred())
                }
        },
        Entry("mesh client → service name", testCase{
            from: "demo-client", to: "test-server.mesh", works: true,
        }),
        Entry("external client → DPP port", testCase{
            from: "demo-client-not-in-mesh", 
            to: net.JoinHostPort(app.GetIP(), "80"), 
            works: true,
        }),
        Entry("mesh client → service port", testCase{
            from: "demo-client",
            to: net.JoinHostPort(app.GetIP(), "8080"),
            works: true,
        }),
        Entry("external client → service port", testCase{
            from: "demo-client-not-in-mesh",
            to: net.JoinHostPort(app.GetIP(), "8080"),
            works: true,
        }),
    )
})
```

**Benefits**:
- Native Ginkgo pattern (idiomatic)
- Each Entry runs independently  
- Better test output (shows which Entry failed)
- Parallel execution support
- ~200 lines → ~60 lines (70% reduction)

Same pattern for the strict mode test with 8 Entry cases instead of manually duplicating assertions.

lukidzi: Added check of `test-server-not-secure`

Automaat (approved):

lobkovilya (approved):

Issue #14163, opened 2025-08-06
Validate on api and k8s controller if user set origin for MeshTrust


lukidzi:
### Description

We cannot validate it in validator since it runs on a create/update by our component which reconcile. To avoid this we need to add a check on Api and Controller

Implementation:
* Api - universal https://github.com/kumahq/kuma/blob/master/pkg/api-server/resource_endpoints.go#L432
* K8s webhook https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/webhooks/validation.go#L67

lukidzi:
Should we add a status field to the MeshTrust resource? The initial design, which places the origin field in the spec, might not be ideal: it requires mapping objects and then checking whether they have an origin. Instead, we could introduce a status field. While the API wouldn’t allow users to modify this field directly, the MeshIdentity controller could update it internally.

If we go this route, we’d need to deprecate or remove the existing field, since the feature is still experimental.
```yaml
status:
  origin:
    kri: kri_...
```

PR #15165, opened 2025-12-05
feat(meshtrust): move origin field from spec to status

slonka:
## Motivation

In https://github.com/kumahq/kuma/issues/14163

## Implementation information

- Add has_status=true to MeshTrust resource
- Update MeshIdentity controller to set status.origin
- Deprecate spec.origin field with migration helper
- Update tests and golden files

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/14163

### DIFF
--- a/pkg/core/resources/apis/meshidentity/status/updater_test.go
+++ b/pkg/core/resources/apis/meshidentity/status/updater_test.go
@@ -320,6 +320,7 @@ var _ = Describe("Updater", func() {
 			g.Expect(trust.Status.Origin.KRI).ToNot(BeNil())
 			g.Expect(*trust.Status.Origin.KRI).ToNot(BeEmpty())
 			// Spec.origin should NOT be set (deprecated field)
+			//nolint:staticcheck // SA1019 Backward compatibility check
 			g.Expect(trust.Spec.Origin).To(BeNil())
 		}, "10s", "100ms").Should(Succeed())
 	})


pkg/core/resources/apis/meshtrust/api/v1alpha1/helpers.go
```diff
@@ -7,3 +7,20 @@ func GetAllTrustDomains(meshTrusts MeshTrustResourceList) map[string]struct{} {
 	}
 	return trustDomains
 }
+
+// MigrateOriginToStatus migrates the origin field from spec to status
+// for backward compatibility during the transition period.
+// Returns true if migration was performed, false otherwise.
+func (r *MeshTrustResource) MigrateOriginToStatus() bool {
```
Automaat, line 14: This returns bool, but result is not used

pkg/core/resources/apis/meshidentity/status/updater.go
```diff
@@ -218,6 +218,12 @@ func (i *IdentityProviderReconciler) createOrUpdateMeshTrust(ctx context.Context
 			return err
 		}
 	}
+
+	// Migrate deprecated spec.origin to status.origin for backward compatibility
+	if update {
```
Automaat, line 223: can we just move this to original update block?

deployments/charts/kuma/crds/kuma.io_meshidentities.yaml
```diff
@@ -265,3 +265,5 @@ spec:
         type: object
     served: true
     storage: true
+    subresources:
```
Automaat, line 268: why this is needed?

deployments/charts/kuma/crds/kuma.io_meshtrusts.yaml
```diff
@@ -71,8 +71,11 @@ spec:
                 minItems: 1
                 type: array
               origin:
-                description: Origin specifies whether the resource was created from
-                  a MeshIdentity.
+                description: |-
```
Automaat, line 74: this is actually experimental and noone uses it, we can just break compatibility

Issue #15127, opened 2025-11-28
Properties of HTTP responses are alphabetised


johncowen:
### Kuma Version

master

### Describe the bug

Properties of HTTP responses are alphabetised, whereas they used to be ordered in order of "importance"

<img width="1401" height="628" alt="Image" src="https://github.com/user-attachments/assets/491479ef-5e26-43f7-ae07-6454e3c68969" />

For example the above screenshot shows the `type` at the very bottom of the response whereas it should really be at the top.

Automaat:
> Triage it might be related to recent changes in marshalling around KRI

lahabana:
@Automaat let's look at why there's a map and why we can't just directly serialize the object. If it's needed should we use a struct instead of a map to preserve field order and avoid manually building json objects.

PR #15168, opened 2025-12-05
fix(api-server): preserve field order in HTTP responses

Automaat:
## Motivation

HTTP response properties were being alphabetized instead of ordered by importance, causing important fields like `type` to appear at the bottom of responses (issue #15127).

## Implementation information

Changed `MarshalJSON` in `pkg/core/resources/model/rest/v1alpha1/resource.go` to manually concatenate JSON bytes instead of using a map. This preserves field order: type, spec, status appear in importance order rather than alphabetically.

## Supporting documentation

Fixes https://github.com/kumahq/kuma/issues/15127

### DIFF
--- a/pkg/api-server/testdata/resources/crud/put_meshmultizoneservice_01.golden.json
+++ b/pkg/api-server/testdata/resources/crud/put_meshmultizoneservice_01.golden.json
@@ -22,5 +22,6 @@
     "appProtocol": "tcp"
    }
   ]
- }
+ },
+ "status": {}
 }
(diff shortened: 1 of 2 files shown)


Automaat:
Applied review comment fixes in ff61fcae06:

✓ Line 45: Use bytes.Equal instead of string conversion for specBytes  
✓ Line 50: Use bytes.Equal instead of string conversion for statusBytes

Line 44 (defensive validation): json.Marshal on struct guarantees valid JSON ending with }. Error already handled at lines 19-21. Check for impossible condition would add unreachable code.

PR #15171, opened 2025-12-05
fix(meshservice): skip MeshService for delegated gateway Services

### LINKED ISSUE #12746: MeshService for delegated gateway shows unavailable
### Kuma Version

master

### Describe the bug

I followed the [Delegated Gateway guide](https://kuma.io/docs/2.9.x/guides/gateway-delegated/) and successfully deployed a delegated gateway. However, after deployment, I checked the UI and noticed that the MeshService status is marked as Unavailable. I understand that the service does not directly communicate with the gateway, but this status is misleading and might suggest that the gateway is not deployed correctly.

traffic works:
```
~ $ curl -XPOST 172.21.3.1:80/increment
{"counter":7,"zone":"local","err":null}~ $
```

```
type: MeshService
mesh: default
name: kong-gateway-proxy.kong
labels:
  app.kubernetes.io/instance: kong
  app.kubernetes.io/managed-by: Helm
  app.kubernetes.io/name: gateway
  app.kubernetes.io/version: '3.8'
  enable-metrics: 'true'
  helm.sh/chart: gateway-2.46.0
  k8s.kuma.io/is-headless-service: 'false'
  k8s.kuma.io/namespace: kong
  k8s.kuma.io/service-name: kong-gateway-proxy
  kuma.io/display-name: kong-gateway-proxy
  kuma.io/env: kubernetes
  kuma.io/managed-by: k8s-controller
  kuma.io/mesh: default
  kuma.io/origin: zone
  kuma.io/zone: default
spec:
  state: Unavailable
  selector:
    dataplaneTags:
      app.kubernetes.io/component: app
      app.kubernetes.io/instance: kong
      app.kubernetes.io/name: gateway
      k8s.kuma.io/namespace: kong
  ports:
    - name: kong-proxy
      port: 80
      targetPort: 8000
      appProtocol: tcp
    - name: kong-proxy-tls
      port: 443
      targetPort: 8443
      appProtocol: tcp
status:
  vips:
    - ip: 10.43.99.250
  tls:
    status: Ready
  dataplaneProxies: {}
```

Image

### To Reproduce

Follow the guide https://kuma.io/docs/2.9.x/guides/gateway-delegated/ 

### Expected behavior

The gateway is shown as a healthy or MeshService for gateway is not generated since it should be an entry point and doesn't need to be tracked (cross mesh is not supported with MS for now)

_No response_

Automaat:
## Motivation

Delegated gateways deployed following the official guide show as "Unavailable" in UI even though they work correctly. This is misleading and suggests incorrect deployment. Root cause: `kuma.io/gateway` annotation exists on Pod/Deployment but not Service, so MeshService controller cannot detect it's a gateway.

## Implementation information

Enhanced MeshService controller to check backing Pods via EndpointSlices:
- Added `isServiceForGateway()` helper that queries EndpointSlices for Service
- Checks first Pod for `kuma.io/gateway` annotation
- Skips MeshService creation if gateway detected (deletes existing if any)
- Integrated check after line 135 in reconciliation logic
- Added test case for delegated gateway with annotation on Pod

## Supporting documentation

Fixes #12746


pkg/plugins/runtime/k8s/controllers/meshservice_controller.go
```diff
@@ -207,7 +219,11 @@ func (r *MeshServiceReconciler) Reconcile(ctx context.Context, req kube_ctrl.Req
 					endpoint.TargetRef.APIVersion != "") {
 				continue
 			}
-			servicePodEndpoints[kube_types.NamespacedName{Name: endpoint.TargetRef.Name, Namespace: endpoint.TargetRef.Namespace}] = endpoint
+			namespace := endpoint.TargetRef.Namespace
+			if namespace == "" {
+				namespace = svc.Namespace
+			}
```
slonka: I think we have ValueOrDefault or something similar - this could be simplified like that.

pkg/plugins/runtime/k8s/controllers/meshservice_controller.go
```diff
@@ -269,6 +285,62 @@ func (r *MeshServiceReconciler) Reconcile(ctx context.Context, req kube_ctrl.Req
...
+		}),
+	); err != nil {
+		return false, errors.Wrap(err, "unable to list EndpointSlices for Service")
+	}
+
+	// Check first Pod for gateway annotation
+	for _, slice := range endpointSlices.Items {
+		for _, endpoint := range slice.Endpoints {
+			if endpoint.TargetRef == nil ||
+				endpoint.TargetRef.Kind != "Pod" ||
```
slonka, line 324: Isn't there a const for "Pod"?

pkg/plugins/runtime/k8s/controllers/meshservice_controller.go
```diff
@@ -440,6 +516,7 @@ func (r *MeshServiceReconciler) SetupWithManager(mgr kube_ctrl.Manager) error {
 		Watches(&kube_core.Namespace{}, kube_handler.EnqueueRequestsFromMapFunc(NamespaceToServiceMapper(r.Log, mgr.GetClient())), builder.WithPredicates(predicate.LabelChangedPredicate{})).
 		Watches(&v1alpha1.Mesh{}, kube_handler.EnqueueRequestsFromMapFunc(MeshToAllMeshServices(r.Log, mgr.GetClient())), builder.WithPredicates(CreateOrDeletePredicate{})).
 		Watches(&kube_discovery.EndpointSlice{}, kube_handler.EnqueueRequestsFromMapFunc(EndpointSliceToServicesMapper(r.Log, mgr.GetClient()))).
+		Watches(&kube_core.Pod{}, kube_handler.EnqueueRequestsFromMapFunc(PodToServicesMapper(r.Log, mgr.GetClient())), builder.WithPredicates(GatewayAnnotationChangedPredicate{})).
```
slonka, line 529: Is the perf impact of this going to be big?

pkg/plugins/runtime/k8s/controllers/testdata/meshservice/06.meshservice.yaml
```diff
@@ -0,0 +1,2 @@
+items: []
```
slonka, line 3: Why is it empty?

pkg/plugins/runtime/k8s/controllers/meshservice_controller.go
Automaat: we have somehing similar in tests, and pointer.OrDefault, but nothing for this in production code

pkg/plugins/runtime/k8s/controllers/testdata/meshservice/06.meshservice.yaml
Automaat: Empty because no MeshService should be created for delegated gateways when the kuma.io/gateway annotation is on the Pod (not the Service)

pkg/plugins/runtime/k8s/controllers/meshservice_controller.go
Automaat: I could not find anything

pkg/plugins/runtime/k8s/controllers/meshservice_controller.go
Automaat: added util

pkg/plugins/runtime/k8s/controllers/meshservice_controller.go
Automaat: I dont think so, query there is specific and we stop iteration an first pod

PR #15172, opened 2025-12-05
feat(meshmultizoneservice): add status condition for empty selector

### LINKED ISSUE #14433: Make clear that MeshMultiZoneService doesn't select any MeshService
### Description

When creating a MeshMultiZoneService we create a selector to select specific MeshServices. Sometimes selector might not select any MeshService and user might not be aware that the MMZS is not having any endpoint. We should check if we can improve it

Automaat:
## Motivation

When creating MeshMultiZoneService with a selector, users aren't aware when selector doesn't match any MeshService. MMZS has no endpoints but no indication of the problem.

## Implementation information

Added `MeshServicesMatched` condition to `MeshMultiZoneServiceStatus`:
- Condition set to `False` (reason: `NoMatchesFound`) when selector matches 0 MeshServices
- Condition set to `True` (reason: `MatchesFound`) when matches exist, with count in message
- StatusUpdater (zone-side) computes and updates condition during reconciliation
- Added 3 test cases covering: no matches, matches found, matches disappear

Pattern follows MeshIdentity's condition-based status reporting.

## Supporting documentation

Fixes #14433

### DIFF
--- a/pkg/core/resources/apis/meshmultizoneservice/status_updater_test.go
+++ b/pkg/core/resources/apis/meshmultizoneservice/status_updater_test.go
@@ -2,7 +2,6 @@ package meshmultizoneservice_test
 
 import (
 	"context"
-	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -26,8 +25,7 @@ import (
 )
 
 var _ = Describe("Updater", func() {
-	var ch chan struct{}
-	var updaterDone *sync.WaitGroup
+	var stopCh chan struct{}
 	var resManager manager.ResourceManager
 	var metrics core_metrics.Metrics
 
@@ -36,28 +34,20 @@ var _ = Describe("Updater", func() {
 		Expect(err).ToNot(HaveOccurred())
 		metrics = m
 		resManager = manager.NewResourceManager(memory.NewStore())
-		updaterDone = &sync.WaitGroup{}
 
 		updater, err := meshmultizoneservice.NewStatusUpdater(logr.Discard(), resManager, resManager, 50*time.Millisecond, m)
 		Expect(err).ToNot(HaveOccurred())
-		ch = make(chan struct{})
-		updaterDone.Add(1)
-		go func() {
+		stopCh = make(chan struct{})
+		go func(stopCh chan struct{}) {
 			defer GinkgoRecover()
-			defer updaterDone.Done()
-			Expect(updater.Start(ch)).To(Succeed())
-		}()
+			Expect(updater.Start(stopCh)).To(Succeed())
+		}(stopCh)
 
 		Expect(samples.MeshDefaultBuilder().Create(resManager)).To(Succeed())
 	})
 
 	AfterEach(func() {
-		if ch != nil {
-			close(ch)
-		}
-		if updaterDone != nil {
-			updaterDone.Wait()
-		}
+		close(stopCh)
 	})
 
 	matchedMeshServices := func() ([]meshmzservice_api.MatchedMeshService, error) {


pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/rest.yaml
```diff
@@ -221,6 +221,46 @@ components:
...
+                  reason:
+                    description: |-
+                      reason contains a programmatic identifier indicating the reason for the condition's last transition.
+                      Producers of specific condition types may define expected values and meanings for this field,
+                      and whether the values are considered a guaranteed API.
+                      The value should be a CamelCase string.
+                      This field may not be empty.
+                    maxLength: 1024
+                    minLength: 1
+                    pattern: ^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$
```
slonka, line 243: why do we need this pattern when our current available options options are: `NoMatchesFound` and `MatchesFound`?

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/rest.yaml
```diff
@@ -221,6 +221,46 @@ components:
...
+                  status:
+                    description: status of the condition, one of True, False, Unknown.
+                    enum:
+                      - "True"
+                      - "False"
+                      - Unknown
+                    type: string
+                  type:
+                    description: type of condition in CamelCase or in foo.example.com/CamelCase.
+                    maxLength: 316
```
slonka, line 254: why is it 316?

pkg/core/resources/apis/meshmultizoneservice/status_updater.go
```diff
@@ -115,10 +118,18 @@ func (s *StatusUpdater) updateStatus(ctx context.Context) error {
 			return matched[i].FullyQualifiedName() < matched[j].FullyQualifiedName()
 		})
 
-		if !reflect.DeepEqual(mzSvc.Status.MeshServices, matched) {
+		// Build condition based on whether matches were found
+		condition := buildMeshServicesMatchedCondition(matched)
+
+		// Check if status needs updating (either meshServices or conditions changed)
+		statusChanged := !reflect.DeepEqual(mzSvc.Status.MeshServices, matched) ||
```
slonka, line 122: does it make sense to make an early check for `len` or something similar to not do deep equal right away? or does deep equal do that as a first step?

pkg/core/resources/apis/meshmultizoneservice/status_updater.go
```diff
@@ -115,10 +118,18 @@ func (s *StatusUpdater) updateStatus(ctx context.Context) error {
...
+			!conditionEquals(mzSvc.Status.Conditions, condition)
+
+		if statusChanged {
 			log := s.logger.WithValues("meshmultizoneservice", mzSvc.Meta.GetName())
 			mzSvc.Status.MeshServices = matched
-			log.Info("updating matched mesh services", "matchedMeshServices", matched)
+			mzSvc.Status.Conditions = updateConditions(mzSvc.Status.Conditions, condition)
+			log.Info("updating matched mesh services", "matchedMeshServices", matched, "condition", condition.Type)
```
slonka:
```suggestion
			log.V(1).Info("updating matched mesh services", "matchedMeshServices", matched, "condition", condition.Type)
```

pkg/core/resources/apis/meshmultizoneservice/status_updater_test.go
```diff
@@ -33,17 +35,20 @@ var _ = Describe("Updater", func() {
 
 		updater, err := meshmultizoneservice.NewStatusUpdater(logr.Discard(), resManager, resManager, 50*time.Millisecond, m)
 		Expect(err).ToNot(HaveOccurred())
-		stopCh = make(chan struct{})
+		ch := make(chan struct{})
+		stopCh = ch
```
slonka: no need for this `ch` variable

pkg/core/resources/apis/meshmultizoneservice/status_updater.go
Automaat: deepEqual starts by checking len

pkg/core/resources/apis/meshmultizoneservice/status_updater.go
Automaat: updated

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/rest.yaml
Automaat: This is coming from our common API for condition https://github.com/kumahq/kuma/blob/29d62590fdf1b203ec729192b03144e98aca3570/api/common/v1alpha1/status.go#L18-L51

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/rest.yaml
Automaat: This is coming from our common API for condition https://github.com/kumahq/kuma/blob/29d62590fdf1b203ec729192b03144e98aca3570/api/common/v1alpha1/status.go#L18-L51

pkg/core/resources/apis/meshmultizoneservice/status_updater_test.go
Automaat: updated

pkg/core/resources/apis/meshmultizoneservice/status_updater_test.go
```diff
@@ -23,7 +24,8 @@ import (
 )
 
 var _ = Describe("Updater", func() {
-	var stopCh chan struct{}
+	var ch chan struct{}
+	var updaterDone *sync.WaitGroup
```
slonka: do we need both of them?

PR #15227, opened 2025-12-09
fix(api-server): simplify and fix hostname inspection for multi-zone …

lukidzi:
## Summary

In multi-zone setups, hostname generators may use origin-based selectors (e.g., `kuma.io/origin` label). Resources synced across zones can appear with different origins depending on perspective. The global control plane now tests both perspectives to ensure all possible hostnames are discovered.

## Changes

  - Extracted duplicate hostname generation logic into `generateAndRecord` helper function
  - Simplified origin testing logic from duplicated code blocks to a clean loop-based approach
  - Added `isGlobal` parameter to hostname inspection endpoints
  - Global control planes now test both Zone and Global origin perspectives to capture all hostname matches
  - Zone control planes only test the actual service origin for efficiency

### Testing
  - Added dedicated test suite for global mode in `endpoints_table_test.go`
  - Organized test data into `zone/` and `global/` subdirectories

### DIFF
--- a/pkg/api-server/inspect_mesh_service.go
+++ b/pkg/api-server/inspect_mesh_service.go
@@ -143,9 +143,10 @@ func matchingHostnames(resManager manager.ResourceManager, isGlobal bool) restfu
 				origins = []mesh_proto.ResourceOrigin{mesh_proto.ZoneResourceOrigin, mesh_proto.GlobalResourceOrigin}
 			}
 
+			originalMeta := svc.GetMeta()
 			for _, origin := range origins {
 				overridden := ResourceMetaWithOverriddenOrigin{
-					ResourceMeta: svc.GetMeta(),
+					ResourceMeta: originalMeta,
 					origin:       origin,
 				}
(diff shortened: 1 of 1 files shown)


Automaat (commented):
.

pkg/api-server/inspect_mesh_service.go
```diff
@@ -118,35 +133,28 @@ func matchingHostnames(resManager manager.ResourceManager) restful.RouteFunction
```
Automaat, line 133: The `svc` object is mutated in the loop via `SetMeta()`. On second iteration of `origins` loop, `svc.GetMeta()` returns the `ResourceMetaWithOverriddenOrigin` from previous iteration, not the original meta.

**Fix:** Store original meta before the loop:
```go
originalMeta := svc.GetMeta()
for _, origin := range origins {
    overridden := ResourceMetaWithOverriddenOrigin{
        ResourceMeta: originalMeta,
        origin:       origin,
    }
    svc.SetMeta(overridden)
    ...
}
```

PR #15233, opened 2025-12-10
refactor(api-server): use auxiliary struct for JSON marshaling

Automaat:
## Motivation

PR #15168 fixed field ordering in HTTP responses by replacing map-based marshaling with manual byte concatenation. While functional, byte manipulation is error-prone and harder to maintain.

## Implementation information

Replaced manual byte concatenation in `Resource.MarshalJSON` with explicit auxiliary struct approach:

**Before:** Manual byte slicing and appending
```go
result := metaBytes[:len(metaBytes)-1] // Remove closing }
if len(specBytes) > 0 {
    result = append(result, []byte(`,"spec":`)...)
    result = append(result, specBytes...)
}
```

**After:** Standard json.Marshal with ordered struct
```go
aux := &struct {
    Type string `json:"type"`
    // ... explicit field order ...
    Spec   json.RawMessage `json:"spec,omitempty"`
    Status json.RawMessage `json:"status,omitempty"`
}{...}
return json.Marshal(aux)
```

Benefits:
- Uses standard Go JSON encoding (no manual byte manipulation)
- Type-safe with compiler-checked field assignments
- Preserves field order through struct definition
- More maintainable and less error-prone
- Easier to extend with new fields

Trade-off: Inlines KRI logic from ResourceMeta.MarshalJSON, but explicit field ordering is worth the duplication.

## Supporting documentation

Related: #15168, #15127

### DIFF
--- a/pkg/core/resources/model/rest/v1alpha1/resource.go
+++ b/pkg/core/resources/model/rest/v1alpha1/resource.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"encoding/json"
+	"time"
 
 	"github.com/kumahq/kuma/v2/pkg/core/kri"
 	core_model "github.com/kumahq/kuma/v2/pkg/core/resources/model"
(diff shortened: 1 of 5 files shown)


pkg/core/resources/model/rest/v1alpha1/resource.go
```diff
@@ -15,46 +17,54 @@ type Resource struct {
...
+	if id, _ := kri.FromResourceMetaE(r.ResourceMeta, r.Type); !id.IsEmpty() {
+		kriStr = id.String()
 	}
 
-	if len(statusBytes) > 0 {
-		result = append(result, []byte(`,"status":`)...)
-		result = append(result, statusBytes...)
+	// Explicit struct with all fields in desired order
+	// Uses standard json.Marshal to avoid manual byte manipulation
+	aux := &struct {
```
lobkovilya, line 45: Isn't it basically type `Resource` (defined at the beginning of this file)?

Automaat: we need to add KRI here during marshalling which is not in resource

pkg/core/resources/model/rest/v1alpha1/resource.go
```diff
@@ -15,46 +17,54 @@ type Resource struct {
...
+	}{
+		Type:             r.Type,
+		Mesh:             r.Mesh,
+		Name:             r.Name,
+		CreationTime:     r.CreationTime,
+		ModificationTime: r.ModificationTime,
+		Labels:           r.Labels,
+		KRI:              kriStr,
+		Spec:             specJSON,
+		Status:           statusJSON,
```
slonka: Can't we do:

```suggestion
	aux := &struct {
		Resource
		KRI string `json:"kri,omitempty"`
	}{
		Resource: r,
		KRI:      kriStr,
```

?

Automaat: fixed

pkg/core/resources/model/rest/v1alpha1/resource.go
Automaat: updated

slonka (approved):

PR #15241, opened 2025-12-10
feat(meshtrust): add status field and deprecate origin in spec

slonka:
## Motivation

MeshTrust's spec.origin field is informational (tracks if resource was created from MeshIdentity) and should not be user-settable. Moving it to status.origin better reflects its read-only nature.

## Implementation information

- Added Status.Origin field to MeshTrust
- Deprecated Spec.Origin field (marked with deprecation notice)
- Added validation webhook warning when users attempt to set status.origin
- Fixed status handling to prevent nil pointer errors and validation failures with empty status objects
- Updated MeshIdentity status updater to populate new status.origin field

##  Supporting documentation

- Validation tests: pkg/plugins/runtime/k8s/webhooks/validation_test.go
- API tests with both status configurations: pkg/api-server/resource_ws_test.go
- Migration path: spec.origin continues to work but emits deprecation warning

fix https://github.com/kumahq/kuma/issues/14163

### DIFF
--- a/pkg/test/resources/builders/meshtrust_builder.go
+++ b/pkg/test/resources/builders/meshtrust_builder.go
@@ -48,6 +48,7 @@ c9Vh+kJkmPUJZQIgQBr2GkV8uSfq/5ZKHD6jz6MJvKsg06dMBdvZBIA2ujg=
 					},
 				},
 			},
+			Status: &meshtrust_api.MeshTrustStatus{},
 		},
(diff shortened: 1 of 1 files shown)


slonka:
How are golden files failing since `/golden_files` resulted in no change - https://github.com/kumahq/kuma/actions/runs/20121055512/job/57741126040 🤔

PR #15265, opened 2025-12-12
fix(meshidentity): nil pointer check

lukidzi:
## Motivation

There was a nil pointer
## Implementation information

Check if the struct is defined and then Deref

### DIFF
--- a/pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
+++ b/pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
@@ -83,7 +83,7 @@ func NewBundledIdentityProvider(roSecretManager manager.ReadOnlyResourceManager,
 
 func (b *bundledIdentityProvider) Validate(ctx context.Context, identity *meshidentity_api.MeshIdentityResource) error {
 	if !pointer.DerefOr(identity.Spec.Provider.Bundled.InsecureAllowSelfSigned, false) {
-		if pointer.DerefOr(identity.Spec.Provider.Bundled.Autogenerate.Enabled, false) {
+		if identity.Spec.Provider.Bundled.Autogenerate != nil && pointer.DerefOr(identity.Spec.Provider.Bundled.Autogenerate.Enabled, false) {
 			return errors.Errorf("self-signed certificates are not allowed")
 		}
 		ca, err := b.GetRootCA(ctx, identity)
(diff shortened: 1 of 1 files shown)


pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
```diff
@@ -83,7 +83,7 @@ func NewBundledIdentityProvider(roSecretManager manager.ReadOnlyResourceManager,
 
 func (b *bundledIdentityProvider) Validate(ctx context.Context, identity *meshidentity_api.MeshIdentityResource) error {
 	if !pointer.DerefOr(identity.Spec.Provider.Bundled.InsecureAllowSelfSigned, false) {
-		if pointer.DerefOr(identity.Spec.Provider.Bundled.Autogenerate.Enabled, false) {
+		if identity.Spec.Provider.Bundled.Autogenerate != nil && pointer.DerefOr(identity.Spec.Provider.Bundled.Autogenerate.Enabled, false) {
```
slonka, line 86: where do we check that `Spec`, `Provider` and `Bundled` are not nil?

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
```diff
@@ -237,7 +240,10 @@ func (b *bundledIdentityProvider) CreateIdentity(ctx context.Context, identity *
 	if err != nil {
 		return nil, err
 	}
-	certValidity := pointer.DerefOr(identity.Spec.Provider.Bundled.CertificateParameters.Expiry, DefaultWorkloadCertValidityPeriod)
+	certValidity := DefaultWorkloadCertValidityPeriod
+	if identity.Spec.Provider.Bundled != nil && identity.Spec.Provider.Bundled.CertificateParameters != nil {
```
slonka, line 244: don't we have `bundled` from line [173](https://github.com/kumahq/kuma/pull/15265/changes#diff-f497431b90a1d64ad9ddd0187861e19ae272625a36d5e92a92b98362b9926595R173)?

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
lukidzi: We check here: https://github.com/kumahq/kuma/blob/master/pkg/core/resources/apis/meshidentity/api/v1alpha1/validator.go

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
lukidzi: It a different method

PR #15266, opened 2025-12-12
refactor(policy-gen): introduce single `model.Res[T]` type instead of autogenerated `*Resource` types

lobkovilya:
## Supporting documentation

Fix #XX


lukidzi:
Closing as it was a PoC

PR #15294, opened 2025-12-16
ci(pr-coments): handle partial successes in pr-comments workflow

slonka:
## Motivation

If there is a partial failure like this https://github.com/kumahq/kuma/actions/runs/20259928702 it still might be valid to push format changes.

## Implementation information

Made it possible to submit partial success, added 👎 when failure.

### DIFF
--- a/.github/workflows/pr-comments.yaml
+++ b/.github/workflows/pr-comments.yaml
@@ -54,19 +54,19 @@ jobs:
         continue-on-error: true
         run: make clean/generated check
 
-      - name: "Commit and push format changes"
+      - name: "Commit format changes"
         if: contains(github.event.comment.body, '/format') && steps.format.outcome == 'success'
+        id: commit-format
         continue-on-error: true
-        env:
-          GITHUB_TOKEN: ${{ steps.github-app-token.outputs.token }}
         run: |
           if git diff --exit-code --stat; then
             echo "No format changes detected"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-comments.yaml
```diff
@@ -50,31 +50,84 @@ jobs:
       # Automatically update code formatting if /format is in the comment
       - name: "Auto-format code"
         if: contains(github.event.comment.body, '/format')
+        id: format
+        continue-on-error: true
         run: make clean/generated check
+
+      - name: "Commit and push format changes"
```
Automaat: should we just commit and push as last step? after finishing all

.github/workflows/pr-comments.yaml
```diff
@@ -50,31 +50,84 @@ jobs:
...
             git config user.name "${GH_USER}"
             git config user.email "${GH_EMAIL}"
-            git commit -s -m "fix(ci): format files" .
+            git commit -s -m "fix(ci): update transparent proxy golden files" .
             git push
+            echo "TPROXY_GOLDEN_FILES_PUSHED=true" >> "$GITHUB_ENV"
           fi
-      - run: gh api --method POST -f content='hooray' ${{ github.event.comment.url }}/reactions
+
+      - name: "Add success reaction"
```
Automaat, line 171: can we have reaction per step?  to see which failed immediately? Or maybe just add review comment with summary

Issue #12931, opened 2025-02-24
MeshRetry do not allow retry GRPC connection errors


tony-bondarenko:
### Description

All MeshRetry retryOn options available for grpc protocol (Canceled, DeadlineExceeded, Internal, ResourceExhausted, Unavailable) only trigger retry in case http headers are returned from the grpc  service. 
If request failure is related to connection issues (reset, timeout), no retry is triggered.

Such options are available for http protocol: Reset, ConnectFailure. Also 5xx is a combination of these.
But http MeshRetry policy is filtered out if service port has appProtocol = grpc and is not being applied.

Icarus9913:
I did a test for that and got your point, will talk to the team about the design.
Thanks for raising it.

Icarus9913:
In `2.10.*` version, the MeshRetry policy has a strong isolation about the `retryOn` conditions between `http` and `grpc` protocols on each 'Service'. 
> For GRPC, we can only use `retryOn: Canceled, DeadlineExceeded, Internal, ResourceExhausted, Unavailable`

For example, here's my MeshRetry policy which is set the retryOn property explicitly and k8s service configuration:

```text
apiVersion: kuma.io/v1alpha1
kind: MeshRetry
metadata:
  annotations:
    kuma.io/display-name: mesh-retry-all-default
  name: mesh-retry-all-default
  namespace: kuma-system
spec:
  targetRef:
    kind: Mesh
  to:
  - default:
      grpc:
        backOff:
          baseInterval: 25ms
          maxInterval: 250ms
        numRetries: 5
        perTryTimeout: 16s
        retryOn:                                                       // here!
        - Canceled
        - DeadlineExceeded
        - Internal
      http:
        backOff:
          baseInterval: 25ms
          maxInterval: 250ms
        numRetries: 5
        perTryTimeout: 16s
        retryOn:                                                     // here!
        - Reset
        - ConnectFailure
      tcp:
        maxConnectAttempt: 5
    targetRef:
      kind: Mesh
---------------------------
apiVersion: v1
kind: Service
spec:
...
  ports:
  - appProtocol: grpc                                      // here!
    name: name-of-service-port
    port: 80
    protocol: TCP
    targetPort: http-web-svc
```

Then I'll get my xds configurations:

```text
 "route": {
    "cluster": "nginx-service_kuma-demo_svc_80",
    "timeout": "15s",
    "retry_policy": {
      "retry_on": "canceled,deadline-exceeded,internal",                               // here!
[...]

PR #15353, opened 2026-01-11
feat(MeshRetry): support http retryOn for grpc services

kxiong0:
## Motivation

Envoy `RetryPolicy` supports retrying on conditions under both `x-envoy-retry-on` and `x-envoy-retry-grpc-on`. Therefore Kuma should allow all conditions, including http `retryOn` conditions, for grpc services.

## Implementation information

- `MeshRetry` policy validator updated to allow input in `grpc.retryOn` if it is a valid http `retryOn` value
- Removed enum constraint on `MeshRetry` CRD to accept status code entries
- `MeshRetry` configurer updated to support http and grpc conditions
   - Moved logic for separating `retryOn` values and retriable status codes and request headers to function `HttpRetriables`
   - Renamed `GrpcRetryOn` to `GrpcRetriables`, which now returns `retryOn` values and status codes / headers

## Supporting documentation

Addresses https://github.com/kumahq/kuma/issues/12931


sardarmscs (commented):
## Code Review: feat(MeshRetry): support http retryOn for grpc services

### Summary

This PR extends `MeshRetry` gRPC retry policy to accept HTTP `retryOn` values (e.g., `5xx`, `ConnectFailure`, `Reset`, HTTP status codes) in addition to the existing gRPC-specific conditions. Since Envoy combines `x-envoy-retry-on` and `x-envoy-retry-grpc-on` into a single `retry_on` field, and gRPC runs over HTTP/2, connection-level failures are only catchable via HTTP retry conditions. The existing Kuma model was unnecessarily restrictive.

---

### Correctness

**Mostly correct, with one critical concern.**

1. **Envoy retry_on semantics**: The mapping is sound. Mixing `canceled,connect-failure,retriable-status-codes` in a single string is exactly what Envoy expects. ✓

2. **`retriable-status-codes` injection**: When HTTP status codes appear in `grpc.retryOn`, the code correctly adds the `retriable-status-codes` trigger via `ensureRetriableStatusCodes`. ✓

3. **`ensureRetriableStatusCodes` empty string fix**: The guard for empty strings is a legitimate fix — `strings.Split("", ",")` returns `[""]` which would produce a leading comma. ✓

4. **HTTP method conditions for gRPC**: `HttpMethodGet` in a gRPC context is effectively a no-op since gRPC always uses POST. This won't break anything, but could confuse users. Consider a validation warning or documentation note.

---

### Critical Issue: Kubebuilder Enum Marker

The `GRPCRetryOn` type still has:
```go
// +kubebuilder:validation:Enum=Canceled;DeadlineExceeded;Internal;ResourceExhausted;Unavailable
type GRPCRetryOn string
```

The CRD YAML removes the enum constraint manually, but **the next `make generate` will regenerate the CRD from Go types and revert this change**. The kubebuilder marker must be removed or updated for this to persist.

---

### Completeness

A few gaps:
- **No E2E test** demonstrating gRPC retry on HTTP-level conditions
- **No `UPGRADE.md` update** — this is a user-facing behavior change (CRD schema relaxation)
- **Missing edge case tests**: gRPC retryOn with *only* HTTP values (no gRPC-native conditions) — verify no leading/trailing commas in the `retryOn` string

---

### Testing

Unit tests and golden file updates are solid. The validator tests cover both positive (HTTP values accepted in gRPC context) and negative (invalid values rejected) cases. The golden file verifies correct Envoy config generation.

**Missing tests:**
- gRPC retryOn with only HTTP status codes and no gRPC conditions
- gRPC retryOn with only HTTP string conditions (e.g., just `["5xx"]`)
- `ensureRetriableStatusCodes` empty string fix in isolation

---

### Breaking Changes

Low risk:
- **Backward compatible** for existing policies — gRPC-only values continue to work identically
- **CRD schema relaxation** means Kubernetes API server won't reject invalid values at admission — validation deferred to Kuma webhook (acceptable tradeoff, but should be documented)
- **Exported function rename** `GrpcRetryOn` → `GrpcRetriables` — breaking for downstream consumers

---

### Recommendations

1. **Critical: Remove the kubebuilder enum marker** on `GRPCRetryOn` — this will cause the CRD change to be lost on next code generation
2. Add test cases for gRPC retryOn with only HTTP values
3. Update `UPGRADE.md` to document the expanded `grpc.retryOn` values
4. Consider validation warning for HTTP method conditions in gRPC context

---

### Verdict

Well-motivated change that correctly addresses a real limitation. The implementation approach — reusing HTTP retry parsing — is clean. However, the **kubebuilder enum marker issue must be resolved** before merge, as it will cause the CRD change to be lost on the next generation run. With that fixed and edge case tests added, this is ready to merge.

kxiong0:
Hi @sardarmscs 
Thanks for the reveiw!
I've removed the kubebuilder marker and updated the crd / testdata yamls to reflect the change, and added more tests as you suggested.

> Consider validation warning for HTTP method conditions in gRPC context

Is there a recommended way for implementing this validation warning? I've been reading through other plugins but I can't find a good example.

Issue #11294, opened 2024-09-03
Use resource builders in e2e tests


michaelbeaumont:
### Description

We have a lot of static YAMLs and `Install` functions for various permutations resource fields. We should use the builders we've started using in unit tests.

> nit: not the biggest fan of this. We may then need metrics and we'll end up with
`MTLSMeshWithMeshServicesAndMetricsKubernetes`. I'd do

```
samples.MeshMTLSBuilder().WithName(meshName).WithEnabledMeshServices("Exclusive").KubeYaml()
```

_Originally posted by @jakubdyszkiewicz in https://github.com/kumahq/kuma/pull/11279#discussion_r1742057213_

slonka:
Please take https://github.com/kumahq/kuma/pull/11383#discussion_r1758627010 into account when doing this.

lukidzi:
I think we started using it recently. Does it address the issue?https://github.com/kumahq/kuma/blob/master/test/e2e_env/kubernetes/meshexternalservice/meshexternalservice.go#L31

lukidzi:
> lets check if we still have these methods

lukidzi:
We still are using some of them `MTLSMeshWithMeshServicesUniversal`, `MTLSMeshWithMeshServicesKubernetes`

NawafSwe:
Hello @michaelbeaumont , @slonka 
Is this issue still valid? 
I would like to contribute 🙏🏻

slonka:
hi @NawafSwe :) happy New Year! :) please check the codebase for usages of functions mentioned here https://github.com/kumahq/kuma/issues/11294#issuecomment-2981499205 and if there are any the issue is still valid. I'll assign it to you. Let us know if you have any questions.

NawafSwe:
Thanks! @slonka I will check it out by today

NawafSwe:
Hey @slonka, hope you’re doing well!
I’ve gone through the codebase thoroughly and I have a good understanding of what’s required from my side.

I just wanted to clarify one thing: in the e2e tests, I noticed there are some hardcoded YAMLs, but there don’t seem to be any builder functions for them.

Should I go ahead and create builders for these as part of this work, or would you prefer that to be done in a separate PR and, for now, only use builders where they already exist or can be easily replaced?

NawafSwe:
Hello @slonka, I have opened a PR to replace hardcoded YAMLs for mesh and multi-zone mesh resources. 
[...]

PR #15376, opened 2026-01-13
Chore/11294 build resources using builder functions

NawafSwe:
## Motivation

Currently, many e2e tests use hardcoded YAML strings and `fmt.Sprintf`-based helper functions to create test resources. This approach has several drawbacks:

1. **Function name bloat**: As noted in [#11279](https://github.com/kumahq/kuma/issues/11279#issuecomment-xxx), this leads to unwieldy function names like `MTLSMeshWithMeshServicesAndMetricsKubernetes` when we need to support various field combinations.

2. **Maintainability**: Static YAML strings scattered throughout test files are harder to maintain and refactor when resource schemas change.

3. **Type safety**: String-based YAML generation lacks compile-time validation and IDE support.

This PR addresses these issues by replacing static YAML strings with the existing builder pattern that's already used in unit tests, following the approach suggested in the issue comments.

## Changes

This PR replaces hardcoded YAML strings with builder functions in e2e tests:

- **`test/e2e_env/multizone/localityawarelb/meshmultizoneservice.go`**: Replaced static `MeshMultiZoneService` YAML with `builders.MeshMultiZoneService()` builder pattern using method chaining.

- **`test/e2e_env/multizone/externalservices/externalservices_multizone_universal.go`**: Fixed incorrect usage of `ResourceUniversal()` to properly use the builder pattern.

## Benefits

- **Better maintainability**: Changes to resource schemas are easier to propagate through tests
- **Type safety**: Compile-time validation and IDE autocomplete support
- **Readability**: Fluent builder API is more readable than YAML strings
- **Consistency**: Aligns e2e tests with the builder pattern already used in unit tests
- **Extensibility**: Easy to add new field combinations without creating new function variants

## Future work

This PR focuses on `MeshMultiZoneService` as a starting point. Additional resources (such as `MeshLoadBalancingStrategy`) still use static YAML and can be migrated in follow-up PRs as builders become available.

## Related issues

Addresses [#11294](https://github.com/kumahq/kuma/issues/11294) and the concerns raised in [#11279](https://github.com/kumahq/kuma/issues/11279).

> Changelog: skip


NawafSwe:
@slonka appreicate your review, 
I will address the copilot comments 🙌

NawafSwe:
Hey @bartsmykla , @lukidzi appreicate your review 🙏

NawafSwe:
Hey @bartsmykla , @lukidzi appreicate your review 🙏

test/e2e/zoneegress/externalservices/zoneegress_externalservices_hybrid.go
```diff
@@ -4,35 +4,28 @@ import (
...
-    type: builtin
-networking:
-  outbound:
-    passthrough: %s
-routing:
-  zoneEgress: %s
-`
-
+	meshMTLOnFn := func(name string, enableNetworkOutBound, enableZoneIngress bool) *core_mesh.MeshResource {
+		return builders.Mesh().WithName(nonDefaultMesh).WithBuiltinMTLSBackend("ca-1").
```
Automaat, line 23: **blocking:** mTLS is never enabled on this mesh. `WithBuiltinMTLSBackend("ca-1")` calls `AddBuiltinMTLSBackend`, which appends to `Spec.Mtls.Backends` but never sets `EnabledBackend`. The original YAML had `enabledBackend: ca-1`. Because the chain also sets `zoneEgress: true`, `Build()` runs `Validate()` and panics: `mtls: has to be set when zoneEgress enabled` (reproduced locally). The `meshMTLOnFn(_, _, true)` call sites at lines 82 and 188 crash the suite in setup. Use `WithEnabledMTLSBackend("ca-1")` instead.

test/e2e_env/multizone/zoneegress/internal_services.go
```diff
@@ -19,22 +20,15 @@ func InternalServices() {
...
-  backends:
-  - name: ca-1
-    type: builtin
-routing:
-  zoneEgress: true
-`
-
+	mesh := builders.Mesh().
+		WithName(meshName).
+		WithBuiltinMTLSBackend("ca-1").
```
Automaat, line 25: **blocking:** Same root cause. `WithBuiltinMTLSBackend("ca-1")` leaves `EnabledBackend` empty while `WithRoutingZoneEgress(true)` is set, so `Build()` panics with `mtls: has to be set when zoneEgress enabled`. Switch to `WithEnabledMTLSBackend("ca-1")`.

test/e2e_env/multizone/meshtrafficpermission/meshtrafficpermission.go
```diff
@@ -30,18 +31,7 @@ networking:
...
-mtls:
-  enabledBackend: ca-1
-  backends:
-    - name: ca-1
-      type: builtin
-routing:
-  zoneEgress: true
-`, name)
-	return YamlUniversal(mesh)
+	return ResourceUniversal(builders.Mesh().WithName(name).WithBuiltinMTLSBackend("ca-1").WithRoutingZoneEgress(true).Build())
```
Automaat, line 34: **blocking:** Same root cause. `WithBuiltinMTLSBackend("ca-1")` plus `WithRoutingZoneEgress(true)` makes `Build()` panic at `mtlsAndEgressMeshUniversal` setup. Use `WithEnabledMTLSBackend("ca-1")`.

test/e2e_env/multizone/localityawarelb/locality_multizone_hybrid.go
```diff
@@ -15,21 +17,13 @@ import (
...
-networking:
-  outbound:
-    passthrough: false
-routing:
-  zoneEgress: %s
-`, mesh, zoneEgress)
+func MeshMTLSOnAndZoneEgressAndNoPassthrough(mesh string, zoneEgress bool) *mesh.MeshResource {
+	return builders.Mesh().
+		WithName(mesh).
+		WithBuiltinMTLSBackend("ca-1").
```
Automaat, line 23: **blocking:** Same root cause in this shared helper. The `(mesh, true)` caller (line 74) panics at `Build()` because `zoneEgress` is on with no enabled backend. The `false` callers (line 75 here, and `meshloadbalancingstrategy.go:36`) don't panic but silently run with mTLS off despite the `MTLSOn` name. Use `WithEnabledMTLSBackend("ca-1")`.

test/e2e_env/multizone/externalservices/externalservices_multizone_universal.go
```diff
@@ -20,20 +21,11 @@ import (
...
-networking:
-  outbound:
-    passthrough: true
-routing:
-  localityAwareLoadBalancing: %s
-`
+	meshDefaultMTLSOn := builders.
+		Mesh().
+		WithName("default").
+		WithBuiltinMTLSBackend("ca-1").
```
Automaat, line 27: **blocking:** Same root cause, silent variant. There's no `zoneEgress` here, so no panic, but `WithBuiltinMTLSBackend("ca-1")` leaves `EnabledBackend` empty, so `meshDefaultMTLSOn` (applied at lines 93 and 265) runs with mTLS disabled even though the original YAML and the variable name say mTLS is on. Use `WithEnabledMTLSBackend("ca-1")`.

test/e2e_env/multizone/meshservice/connectivity.go
```diff
@@ -38,55 +46,34 @@ func Connectivity() {
...
-spec:
-  template: '{{ .DisplayName }}.svc.{{ .Zone }}.not-my-mesh-specific.mesh.local'
-  selector:
-    meshService:
-      matchLabels:
-        kuma.io/env: universal
-        kuma.io/mesh: non-existent-mesh
-`)).
+			Install(ResourceUniversal(hostGen("kube-mesh-specific-msconnectivity",
+				`'{{ .DisplayName }}.{{ .Namespace }}.svc.{{ .Zone }}.mesh-specific.mesh.local'`,
```
Automaat, line 50: **blocking:** The HostnameGenerator templates are wrapped in literal single quotes. In the original YAML, `template: '{{ ... }}'` used the quotes as a YAML scalar delimiter, so the stored value had none. As Go raw strings (lines 50, 57, 70), the quotes are part of the value. `WithTemplate` stores them verbatim and `text/template` parsing accepts them, so `Build()` doesn't reject it. At generation time the rendered hostname carries leading and trailing `'`, and the positive assertion at line 313 queries the unquoted hostname, so the lookup never resolves. Drop the surrounding quotes from all three templates.

test/e2e_env/multizone/meshservice/connectivity.go
```diff
@@ -38,55 +46,34 @@ func Connectivity() {
...
+				}))).
+			Install(ResourceUniversal(hostGen("kube-not-my-mesh-specific-msconnectivity",
+				`'{{ .DisplayName }}.{{ .Namespace }}.svc.{{ .Zone }}.not-my-mesh-specific.mesh.local'`,
+				map[string]string{
+					"kuma.io/env":                     "kubernetes",
+					"kuma.io/mesh":                    meshName,
+					"k8s.kuma.io/is-headless-service": "false",
+				}))).
+			Install(ResourceUniversal(hostGen("uni-mesh-specific-msconnectivity",
+				`{{ .DisplayName }}.{{ .Namespace }}.svc.{{ .Zone }}.mesh-specific.mesh.local`,
```
Automaat, line 64: **blocking:** This template gained a `.{{ .Namespace }}` segment the original lacked. The original was `{{ .DisplayName }}.svc.{{ .Zone }}.mesh-specific.mesh.local`. Universal services have an empty namespace, so this renders `test-server..svc.kuma-5.mesh-specific.mesh.local` (double dot), while the assertion at line 331 queries the single-dot form. Remove the `.{{ .Namespace }}` segment.

pkg/test/resources/builders/mesh_builder.go
```diff
@@ -67,6 +68,41 @@ func (m *MeshBuilder) WithBuiltinMTLSBackend(name string) *MeshBuilder {
...
+		m.res.Spec.Networking = &mesh_proto.Networking{
+			Outbound: &mesh_proto.Networking_Outbound{
+				Passthrough: &wrapperspb.BoolValue{Value: b},
+			},
+		}
+	}
+	return m
+}
+
+func (m *MeshBuilder) WithRoutingZoneEgress(b bool) *MeshBuilder {
```
Automaat, line 82: **suggestion:** The four new boolean setters introduce a second vocabulary for concepts the builder already names. `WithRoutingZoneEgress(true)` duplicates `WithEgressRoutingEnabled()` (20+ callers), and `WithRoutingDefaultForbidMeshExternalServiceAccess(true)` duplicates `WithMeshExternalServiceTrafficForbidden()`. Two names for one field erode the builder's vocabulary, and the new style will get copied. Prefer the existing intention-revealing methods; add only the genuinely missing `false`-case and locality-aware setters.

pkg/test/resources/builders/mesh_builder.go
```diff
@@ -67,6 +68,41 @@ func (m *MeshBuilder) WithBuiltinMTLSBackend(name string) *MeshBuilder {
 	return m.AddBuiltinMTLSBackend(name)
 }
 
+func (m *MeshBuilder) WithNetworkingPassThrough(b bool) *MeshBuilder {
```
Automaat, line 71: **suggestion:** `WithNetworkingPassThrough` only assigns when `Spec.Networking == nil` and silently does nothing otherwise. Its sibling `WithoutPassthrough` nil-guards both `Networking` and `Outbound` and always assigns. No current caller is bitten (all start from a fresh `Mesh()`), but a setter that drops its argument under invisible conditions is a footgun for the next caller. Mirror the `WithoutPassthrough` structure and always set the value.

test/e2e_env/multizone/externalservices/externalservices_multizone_universal.go
```diff
@@ -270,7 +262,7 @@ networking:
 		)
 
 		// when locality-aware lb is enabled
-		Expect(YamlUniversal(fmt.Sprintf(meshDefaulMtlsOn, "true"))(global)).To(Succeed())
+		Expect(ResourceUniversal(meshDefaultMTLSOn.WithRoutingLocalityAwareLoadBalancing(true).Build())(global)).To(Succeed())
```
Automaat, line 265: **suggestion:** `meshDefaultMTLSOn` is a shared `*MeshBuilder`, and `Build()` returns the same pointer with no copy. It's mutated to `LocalityAwareLoadBalancing(false)` and applied at line 93, then mutated to `true` and re-applied at 265. This works only because each install serializes eagerly before the next mutation; it reads as two independent meshes but aliases one. Build two distinct resources instead.

test/e2e_env/multizone/meshservice/sync.go
```diff
@@ -19,95 +21,50 @@ import (
...
+			WithName("this-service-has-a-name-thats-the-exact-length-allowed-for-svcs").
+			WithLabels(map[string]string{"kuma.io/mesh": "meshservice"}).
+			WithDataplaneTagsSelector(map[string]string{"kuma.io/service": "some-service"}).
+			AddIntPortWithName(80, 80, core_meta.ProtocolHTTP, "80").
+			Build()
 
 		NewClusterSetup().
 			Install(NamespaceWithSidecarInjection(namespace)).
-			Install(YamlK8s(veryLongNamedService)).
+			Install(ResourceUniversal(veryLongNamedService)).
```
Automaat, line 67: **issue:** `veryLongNamedService` switched from a Kubernetes resource (`YamlK8s`, `metadata.namespace: meshservice`, applied to `KubeZone2`) to `ResourceUniversal`, which applies via kumactl and carries no `k8s.kuma.io/namespace` label. The assertion at line 159 expects `hash.HashedName(meshName, "this-service...", "kuma-2", "meshservice")`, where the fourth component is the namespace. The synced resource now hashes without it, so the "should sync MeshServices with a long name" lookup fails. The test's purpose, a K8s name-length boundary, is also K8s-specific and no longer exercised. Keep this one on `YamlK8s` or build it as a K8s resource with the namespace set.

test/e2e_env/multizone/meshservice/connectivity.go
```diff
@@ -38,55 +46,34 @@ func Connectivity() {
...
+				map[string]string{
+					"kuma.io/env":                     "kubernetes",
+					"kuma.io/mesh":                    meshName,
+					"k8s.kuma.io/is-headless-service": "false",
+				}))).
+			Install(ResourceUniversal(hostGen("kube-not-my-mesh-specific-msconnectivity",
+				`'{{ .DisplayName }}.{{ .Namespace }}.svc.{{ .Zone }}.not-my-mesh-specific.mesh.local'`,
+				map[string]string{
+					"kuma.io/env":                     "kubernetes",
+					"kuma.io/mesh":                    meshName,
```
Automaat, line 60: **issue:** This generator's selector changed from `kuma.io/mesh: non-existent-mesh` to `meshName`. It was deliberately pointed at a non-existent mesh so it would match nothing; it now matches the real mesh. The paired negative assertion still passes only by accident (the literal-quote bug above makes its hostname unresolvable anyway), so the test no longer verifies its stated intent. Restore `non-existent-mesh`. Note: fixing the quote bug will expose this as a real failure, so fix both together.

test/e2e/zoneegress/externalservices/zoneegress_externalservices_hybrid.go
```diff
@@ -4,35 +4,28 @@ import (
...
+			WithEgressRoutingEnabled().
```
Automaat, line 24: **suggestion:** A few rough edges in this helper. The typo `meshMTLOnFn` is missing an `S`. The `name` parameter is ignored, since the body hardcodes `WithName(nonDefaultMesh)`. The third parameter `enableZoneIngress` actually drives zone *egress*, which is misleading in a zone-egress test. And `.WithEgressRoutingEnabled()` is dead, because the next line `.WithRoutingZoneEgress(enableZoneIngress)` overwrites the same field. Drop the dead call and the unused parameter, and rename for accuracy.

pkg/test/resources/builders/mesh_builder.go
```diff
@@ -67,6 +68,41 @@ func (m *MeshBuilder) WithBuiltinMTLSBackend(name string) *MeshBuilder {
...
+
+func (m *MeshBuilder) WithRoutingLocalityAwareLoadBalancing(b bool) *MeshBuilder {
+	if m.res.Spec.Routing == nil {
+		m.res.Spec.Routing = &mesh_proto.Routing{}
+	}
+	m.res.Spec.Routing.LocalityAwareLoadBalancing = b
+	return m
+}
+
+func (m *MeshBuilder) WithRoutingDefaultForbidMeshExternalServiceAccess(b bool) *MeshBuilder {
```
Automaat, line 98: **issue:** `WithRoutingDefaultForbidMeshExternalServiceAccess` has zero callers anywhere in the repo. It's speculative and duplicates `WithMeshExternalServiceTrafficForbidden`. Drop it; it's six lines to re-add when something needs it.

PR #6973, opened 2023-06-08
feat(kuma-cp): improve BuildRules algorithm

lobkovilya:
While implementing `targetRef.Kind: MeshHTTPRoute` I've encountered the problem when `BuildRules` returns excessive results because I'm adding an artificial tag `__route_hash__` to the Subset. 

But the problem could be noticed even for the given MeshTrafficPermission policy:

```yaml
type: MeshTrafficPermission
spec:
  targetRef:
    kind: Mesh
  from:
    - targetRef:
        kind: MeshService
        name: frontend
      default:
        action: Allow
    - targetRef:
        kind: MeshServiceSubset
        name: backend
        tags:
          version: v1
      default:
        action: Deny
```

the algorithm is going to produce 2 excessive rules:

```yaml
rules:
  - subset:
      kuma.io/service: frontend
      version: v1
    conf:
      action: Allow
  - subset:
      kuma.io/service: frontend
      version: !v1 # not 'v1'
    conf:
      action: Allow
```

The algorithm splits frontend service into `v1` and `not v1` but there is absolutely no reason for that. This happens because the current algorithm assumes that all provided rules could have an intersection. The new algorithm takes this into account. As a result, we're going to produce fewer rules, for example:

```yaml
type: MeshTrafficPermission
spec:
  targetRef:
    kind: Mesh
  from:
    - targetRef:
        kind: MeshService
        name: frontend_kuma-demo_svc_8080
      default:
        action: Allow
    - targetRef:
        kind: MeshServiceSubset
        name: backend_kuma-demo_svc_3001
        tags:
          team: mesh
          zone: east
          env: dev
      default:
        action: Deny
```

Old algorithm Inspect API:

image

New algorithm Inspect API:

image

### Checklist prior to review

- [X] [Link to relevant issue][1] as well as docs and UI issues --
- [X] This will not break child repos: it doesn't hardcode values (.e.g "kumahq" as a image registry) and it will work on Windows, system specific functions like `syscall.Mkfifo` have equivalent implementation on the other OS --
- [X] Tests (Unit test, E2E tests, manual test on universal and k8s) --
- [X] Do you need to update [`UPGRADE.md`](../blob/master/UPGRADE.md)? --
- [X] Does it need to be backported according to the [backporting policy](../blob/master/CONTRIBUTING.md#backporting)? --
[...]

PR #15412, opened 2026-01-22
feat(MeshTrafficPermission): use cliques instead of connected components as an optimization when building rules

lobkovilya:
## Motivation

The API that we've created for `spec.from` policies can only be implemented by enumerating all possible groups of clients. For example, given the policy:

```yaml
spec:
  from:
    - targetRef:
        kind: MeshSubset
        tags:
          zone: zone-1
      default: $conf1
    - targetRef:
        kind: MeshSubset
        tags:
          namespace: ns-1
      default: $conf2
```

we'd have to generate configs for 4 groups of clients:

* zone=zone-1,namespace=ns-1 -> merge($conf1,$conf2)
* zone!=zone-1,namespace=ns-1 -> $conf2
* zone=zone-1,namespace!=ns-1 -> $conf1
* zone!=zone-1,namespace!=ns-1 -> none

the size of the output depends on the number of unique `key=value` pairs as O(2^N).

Some time ago I noticed we can reduce the size of the output if we consider the fact that not every 2 `spec.from` items have an intersection. Instead of collecting all `key=value` pairs and iterating over all 2^N permutations, we can build a graph where `spec.from` items are nodes and edge exists only when 2 `spec.from` item can have an intersection. After that we take all connected components of the resulting graph and iterate on all possible `key=value` pairs within each individual component. That optimisation was implemented in https://github.com/kumahq/kuma/pull/6973 and significantly reduced the size of the output for cases when `spec.from` items don't have intersection.

The problem with this approach appears if we look at the following example:

```yaml
spec:
  from:
    - targetRef:
        kind: Mesh
      default:
        action: AllowWithShadowDeny
    - targetRef:
        kind: MeshSubset
        tags:
          app: app-1
          namespace: ns-1
      default:
        action: Allow
    - targetRef:
        kind: MeshSubset
        tags:
          app: app-2
          namespace: ns-2
      default:
        action: Allow
    - targetRef:
        kind: MeshSubset
        tags:
          app: app-3
          namespace: ns-3
      default:
        action: Allow
```

Item 1 to 3 in `spec.from` have no intersections with each other. However, item `spec.from[0]` has intersection with every other item. The graph we're building is going to have only 1 connected component:

```mermaid
graph TD;
    Mesh-->app-1,ns-1;
    Mesh-->app-2,ns-2;
    Mesh-->app-3,ns-3;
```

And despite `app-1,ns-1` is not connected with `app-2,ns-2` or `app-3,ns-3` the algorithm is still going to iterate over all possible combinations of pairs, i.e. we'll see in the output:

* app=app-1,namespace=ns-1
* app=app-1,namespace=ns-2
* app=app-1,namespace=ns-3
* app=app-1,namespace!=ns-1,namespace!=ns-2,namespace!=ns-3
* app=app-2,namespace=ns-1
* ...
* app!=app-1,app!=app-2,app!=app-3,namespace=ns-3

even though there are no rules that'd specify behaviour for example for `app-1,ns-2`, etc.

The solution to this problem is actually pretty simple. Instead of taking connected components from the graph we should take [cliques](https://en.wikipedia.org/wiki/Clique_(graph_theory))! 

For the example above, the algorithm will find 3 cliques:

* {`Mesh`, `app-1,ns-1`}
* {`Mesh`, `app-2,ns-2`}
* {`Mesh`, `app-3,ns-3`}

and iterating over all possible pairs within cliques produces a smaller output:

* app=app-1,namespace=ns-1
* app!=app-1,namespace=ns-1
* app=app-1,namespace!=ns-1
* app!=app-1,namespace!=ns-1
* app=app-2,namespace=ns-2
* ...
* app!=app-3,namespace!=ns-3

## Implementation information

I'm using [Bron–Kerbosch algorithm](https://en.wikipedia.org/wiki/Bron%E2%80%93Kerbosch_algorithm) provided by the gonum library to find a set of maximal cliques in the graph. It's considered the most efficient algorithm for the job. For sparse graphs it's basically linear, see [worst-case analysis](https://en.wikipedia.org/wiki/Bron%E2%80%93Kerbosch_algorithm#Worst-case_analysis).

## Supporting documentation

N/A

### DIFF
--- a/pkg/plugins/policies/core/rules/rules.go
+++ b/pkg/plugins/policies/core/rules/rules.go
@@ -40,7 +40,7 @@ var useCliques = true
 
 func init() {
 	// TODO: remove ability to opt-out for the next major version https://github.com/kumahq/kuma/issues/15440
-	if v, ok := os.LookupEnv("KUMA_MESH_TRAFFIC_PERMISSION_DISSABLE_CLIQUES_ALGORITHM"); ok && v == "true" {
+	if v, ok := os.LookupEnv("KUMA_MESH_TRAFFIC_PERMISSION_DISABLE_CLIQUES_ALGORITHM"); ok && v == "true" {
 		useCliques = false
 	}
 }


pkg/plugins/policies/core/rules/rules.go
```diff
@@ -468,14 +468,18 @@ func BuildRules(list []PolicyItemWithMeta, withNegations bool) (Rules, error) {
 		}
 	}
 
-	// 3. Construct rules for all connected components of the graph independently
-	components := topo.ConnectedComponents(g)
+	var nodeGroups [][]graph.Node
+	if useCliques {
+		nodeGroups = topo.BronKerbosch(g)
```
lobkovilya, line 492: todo: need to add comments

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -400,7 +400,7 @@ func BuildSingleItemRules(matchedPolicies []core_model.Resource) (SingleItemRule
 // which has empty subset or kuma.io/service.
 //
 // See the detailed algorithm description in docs/madr/decisions/007-mesh-traffic-permission.md
-func BuildRules(list []PolicyItemWithMeta, withNegations bool) (Rules, error) {
+func BuildRules(list []PolicyItemWithMeta, withNegations bool, useCliques bool) (Rules, error) {
```
lahabana: Is the only reason to have both variants testing? If that is we should probably just remove this before merging no?

pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -910,4 +913,348 @@ var _ = Describe("Rules", func() {
 			}),
 		)
 	})
+
+	Describe("Clique vs Connected Components optimization", func() {
```
lukidzi, line 917: Do we need to keep both? Can't we just switch everyone to use new if it's better and doesn't have an impact on the output?

pkg/plugins/policies/core/rules/rules.go
lobkovilya: I’m leaning towards making this feature opt-out. 

I have concerns about the complexity of the Bron–Kerbosch algorithm. According to the analysis [here](https://en.wikipedia.org/wiki/Bron%E2%80%93Kerbosch_algorithm#Worst-case_analysis), for sparse graphs it is effectively linear, but the worst case is still O(3^(n/3)). I don't know if it's possible to have a format proof that the optimization improves performance for _all_ graphs. What if there is a user out there who has a pathological graph where CPU usage increases significantly, even if the number of rules has never been a concern.

That's why I think it'd be nice to have a way to switch it back to connected components.

pkg/plugins/policies/core/rules/rules_test.go
lobkovilya: Answered [here](https://github.com/kumahq/kuma/pull/15412#discussion_r2727401205), a bit unsure on the CPU if the graph is too dense

pkg/plugins/policies/core/rules/rules.go
slonka: A stupid idea but maybe not that stupid, maybe we could run every 10th computation on the other algorithm, compare the time it took and dynamically switch to the other algorithm if we detect that Bron–Kerbosch is slower? 🤔

pkg/plugins/policies/core/matchers/testdata/egressmatchedpolicies/fromrules/01.golden.yaml
```diff
@@ -55,7 +55,18 @@ Rules:
     Subset:
     - Key: kuma.io/service
       Not: true
-      Value: backend
+      Value: orders
+  - Conf:
+      action: Deny
```
slonka, line 60: maybe this will become apparent while reviewing but why is this changing? 🤔

pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -910,4 +913,348 @@ var _ = Describe("Rules", func() {
...
+					continue
+				}
+				Expect(confCliques).ToNot(BeNil(), "cliques rule should match element %v", elem)
+				Expect(confComponents).ToNot(BeNil(), "components rule should match element %v", elem)
+				Expect(confCliques.Conf).To(Equal(confComponents.Conf),
+					"configuration should be the same for element %v", elem)
+			}
+		}
+
+		It("should produce semantically equivalent results for disjoint subsets connected via Mesh", func() {
```
slonka, line 949: Just so it's easier to visualize for me:

```mermaid
graph TD
    Mesh[Mesh<br/>AllowWithShadowDeny]
    App1[app-1<br/>Allow]
    App2[app-2<br/>Deny]
    App3[app-3<br/>Allow]
    
    Mesh --- App1
    Mesh --- App2
    Mesh --- App3
    
    style Mesh fill:#e1f5ff,stroke:#0066cc,stroke-width:3px
    style App1 fill:#d4edda,stroke:#28a745,stroke-width:2px
    style App2 fill:#f8d7da,stroke:#dc3545,stroke-width:2px
    style App3 fill:#d4edda,stroke:#28a745,stroke-width:2px
```

pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -910,4 +913,348 @@ var _ = Describe("Rules", func() {
...
+				{"app": "app-1"},
+				{"app": "app-2"},
+				{"app": "app-3"},
+				{"app": "app-4"}, // not in any specific rule, should match Mesh
+				{"app": "app-1", "version": "v1"},
+			}
+			verifySemanticEquivalence(rulesCliques, rulesComponents, testElements)
+		})
+
+		It("should handle overlapping subsets that form actual cliques", func() {
```
slonka, line 994:
```mermaid
graph TD
    A[zone: us-east<br/>Allow]
    B[zone: us-east<br/>env: prod<br/>Deny]
    C[env: prod<br/>AllowWithShadowDeny]
    
    A --- B
    B --- C
    A --- C
    
    style A fill:#d4edda,stroke:#28a745,stroke-width:2px
    style B fill:#f8d7da,stroke:#dc3545,stroke-width:2px
    style C fill:#fff3cd,stroke:#ffc107,stroke-width:2px
```

pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -910,4 +913,348 @@ var _ = Describe("Rules", func() {
...
+			testElements := []subsetutils.Element{
+				{"zone": "us-east", "env": "prod"},
+				{"zone": "us-east", "env": "dev"},
+				{"zone": "us-west", "env": "prod"},
+				{"zone": "us-west", "env": "dev"},
+			}
+			verifySemanticEquivalence(rulesCliques, rulesComponents, testElements)
+		})
+
+		It("should handle chain of subsets (A-B-C where A and C don't intersect)", func() {
```
slonka, line 1029:
```mermaid
graph TD
    A[app: app-1<br/>version: v1<br/>Allow]
    B[app: app-1<br/>Deny]
    C[app: app-1<br/>version: v2<br/>AllowWithShadowDeny]
    
    A --- B
    B --- C
    
    style A fill:#d4edda,stroke:#28a745,stroke-width:2px
    style B fill:#f8d7da,stroke:#dc3545,stroke-width:2px
    style C fill:#fff3cd,stroke:#ffc107,stroke-width:2px
```

pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -910,4 +913,348 @@ var _ = Describe("Rules", func() {
...
+			testElements := []subsetutils.Element{
+				{"app": "app-1", "namespace": "ns-1"},
+				{"app": "app-2", "namespace": "ns-2"},
+				{"app": "app-1", "namespace": "ns-2"}, // mix of tags - should match Mesh
+				{"app": "other"},                      // should match Mesh
+			}
+			verifySemanticEquivalence(rulesCliques, rulesComponents, testElements)
+		})
+
+		It("should handle completely disjoint subsets (no edges in graph)", func() {
```
slonka, line 1161:
```mermaid
graph TD
    A[app: app-1<br/>Allow]
    B[app: app-2<br/>Deny]
    
    style A fill:#d4edda,stroke:#28a745,stroke-width:2px
    style B fill:#f8d7da,stroke:#dc3545,stroke-width:2px
```

pkg/plugins/policies/core/rules/rules_test.go
```diff
@@ -910,4 +913,348 @@ var _ = Describe("Rules", func() {
...
+			testElements := []subsetutils.Element{
+				{"zone": "us-east", "env": "prod", "team": "platform"},
+				{"zone": "us-west", "env": "dev", "team": "product"},
+				{"zone": "us-east", "env": "dev"},
+				{"zone": "other"},
+			}
+			verifySemanticEquivalence(rulesCliques, rulesComponents, testElements)
+		})
+
+		It("should handle partial overlap between subsets", func() {
```
slonka, line 1218:
```mermaid
graph TD
    A[zone: us-east<br/>Allow]
    B[zone: us-east<br/>env: prod<br/>Deny]
    C[env: prod<br/>AllowWithShadowDeny]
    D[env: dev<br/>Allow]
    
    A --- B
    B --- C
    A --- D
    
    style A fill:#d4edda,stroke:#28a745,stroke-width:2px
    style B fill:#f8d7da,stroke:#dc3545,stroke-width:2px
    style C fill:#fff3cd,stroke:#ffc107,stroke-width:2px
    style D fill:#d4edda,stroke:#28a745,stroke-width:2px
```

pkg/plugins/policies/core/matchers/testdata/egressmatchedpolicies/fromrules/01.golden.yaml
lobkovilya: It changed from:
```yaml
rules:
  - conf: Allow
    subset:
      - kuma.io/service=backend
  - conf: AllowWithShadowDeny
    subset:
      - kuma.io/service=orders
  - conf: Deny
    subset:
      - kuma.io/service!=backend
      - kuma.io/service!=orders
```
to
```yaml
rules:
  - conf: AllowWithShadowDeny
    subset:
      - kuma.io/service=orders
  - conf: Allow
    subset:
      - kuma.io/service=backend
  - conf: Deny
    subset:
      - kuma.io/service!=orders
  - conf: Deny
    subset:
      - kuma.io/service!=backends
```

The graph is

```mermaid
graph TD
    Mesh[Mesh<br/>Deny]
    App1[kuma.io/service=backend<br/>Allow]
    App2[kuma.io/service=orders<br/>AllowWithShadowDeny]
    
    Mesh --- App1
    Mesh --- App2
    
    style Mesh fill:#e1f5ff,stroke:#0066cc,stroke-width:3px
    style App1 fill:#d4edda,stroke:#28a745,stroke-width:2px
    style App2 fill:#f8d7da,stroke:#dc3545,stroke-width:2px
```

Previously we took these 3 nodes as a single connected component, combined tags from each node:
* kuma.io/service=backend
* kuma.io/service=orders

and generated all possible combinations with these 2 tags (excluding impossible subsets):
* kuma.io/service=backend,kuma.io/service=orders (exclude as an impossible subset)
* kuma.io/service=backend,kuma.io/service!=orders 
* kuma.io/service!=backend,kuma.io/service=orders
* kuma.io/service!=backend,kuma.io/service!=orders

so the final output contained 3 rules.

Now, instead of connected components we're getting cliques. The graph above contains 2:
1. Mesh <--> kuma.io/service=backend
2. Mesh <--> kuma.io/service=orders

For each clique we're generating all possible combinations of tags. 

For clique 1:
* kuma.io/service=backend
* kuma.io/service!=backend

For clique 2:
* kuma.io/service=orders
* kuma.io/service!=orders

So as a result we're getting 4 rules.

pkg/plugins/policies/core/rules/rules.go
lobkovilya: Talked offline with @slonka, it'd make sense to consider such approach if both algos would have the same output, but it's not the case for us.

lobkovilya: Discussed the solution offline with everyone. Passing the opt-out flag that deep is going to affect a lot of function signatures. We've agreed to read the env right in the `rules` package, the flag won't be advertised to users in docs and exists just as a safety feature for patch releases. 

I added TODO comment with the reference to the issue https://github.com/kumahq/kuma/issues/15440, so we don't forget to remove it for 3.0

pkg/plugins/policies/core/rules/rules_test.go
lobkovilya: See https://github.com/kumahq/kuma/pull/15412#discussion_r2732500043

pkg/plugins/policies/core/rules/rules.go
```diff
@@ -30,6 +31,20 @@ import (
...
+// useCliques controls the algorithm for grouping subsets when building rules.
+// Cliques (BronKerbosch) finds maximal fully-connected subgraphs, producing fewer
+// but more precise groups. Connected components finds all reachable nodes, which
+// may over-group subsets that don't directly relate. Cliques is the default as it
+// generates more accurate rules at the cost of slightly more computation.
+var useCliques = true
+
+func init() {
+	// TODO: remove ability to opt-out for the next major version https://github.com/kumahq/kuma/issues/15440
+	if v, ok := os.LookupEnv("KUMA_MESH_TRAFFIC_PERMISSION_DISSABLE_CLIQUES_ALGORITHM"); ok && v == "true" {
```
lahabana: typo

slonka:
```suggestion
	if v, ok := os.LookupEnv("KUMA_MESH_TRAFFIC_PERMISSION_DISABLE_CLIQUES_ALGORITHM"); ok && v == "true" {
```

slonka (dismissed):

pkg/plugins/policies/core/rules/rules.go
lobkovilya: ah indeed, thanks

PR #15460, opened 2026-01-29
docs(MADR): 094 zone proxies new deployment model

slonka:
## Motivation

Zone proxies are becoming mesh-scoped to enable MeshIdentity and per-mesh policy support (see MADR 090). This requires revisiting the deployment tooling — how users deploy zone proxies via Helm, Konnect UI, and Terraform.

## Implementation information

MADR 094 covers deployment tooling decisions for mesh-scoped zone proxies:

- **`meshes` list** in `values.yaml` with per-mesh entries (single-mesh is one entry)
- **Per-mesh templates** render Deployment(s), Service(s), HPA, PDB, ServiceAccount per mesh
- **Per-mesh Services** for mTLS isolation (each mesh gets its own Service/LoadBalancer)
- **Namespace placement**: `kuma-system`
- **`ingress`/`egress` or `combinedProxies`** (mutually exclusive deployment options)
- **`createMesh` + `createPolicies`** for unfederated zones
- **Raw K8s spec passthrough** (`podSpec`, `resources`, `hpa`, `pdb`) — no chart changes needed for new fields
- **Additive migration** alongside existing `ingress`/`egress` keys (deprecated, removed in future major release)
- **MADR 093 relaxation** — allow multiple meshes per namespace, handle Workload collisions in controller
- **Keep `kuma.io/ingress-public-address`** as escape hatch for complex network topologies
- **Default**: `meshes: []` — explicit opt-in, no zone proxies deployed by default

Resource model (Dataplane representation, labels, tokens, workload identity) is deferred to a separate MADR.

## Supporting documentation

- docs/madr/decisions/090-zone-egress-identity.md
- docs/madr/decisions/094-zone-proxy-deployment-model.md

fix #9030


docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,998 @@
...
+
+##### Recommendation for Self-Hosted
+
+**Option A** with good error messaging:
+
+1. User specifies mesh names in values.yaml
+2. Helm deploys zone proxy
+3. Zone proxy logs clear message if mesh doesn't exist:
+   ```
+   WARN: Mesh 'payments-mesh' not found. Zone proxy waiting for mesh creation.
```
lobkovilya: Discussed offline: is it printed by `kuma-dp` when bootstrap fails?

lobkovilya: And envoy probe is probably going to fail as well which is good

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,998 @@
...
+        role: both
+
+  # Optional:
+  meshes:
+    - name: default
+      mtls:
+        enabled: true
+        backends:
+          - name: builtin
+            type: builtin
```
lobkovilya: We don't have to account for mTLS anymore. We need to think what we're going to do with `skipDefaultPolicies` field

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,998 @@
...
+| Tool | Can Validate Mesh? | Strategy |
+|------|-------------------|----------|
+| **Konnect UI** | Yes (API access) | Pre-populate mesh list, validate before generating YAML |
+| **Helm (federated)** | No (offline install) | Accept names, fail gracefully at runtime with clear logs |
+| **Helm (unfederated)** | Yes (creates mesh) | Cross-reference in templates, fail at install if mismatch |
+| **Terraform** | Yes (API access) | Validate at plan/apply, encourage resource references |
+
+#### Design Decisions
+
+##### 1. Mesh Deletion Handling
```
lobkovilya: Discussed offline: probably works out of the box since zone proxy is a regular DPP and we don't allow removing the Mesh that contains online DPPs.

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,998 @@
...
+
+For Helm, we could create one release per mesh:
+```bash
+helm install zone-proxy-payments kuma/kuma-zone-proxy -f payments-values.yaml
+helm install zone-proxy-orders kuma/kuma-zone-proxy -f orders-values.yaml
+```
+
+Then cleanup is: `helm uninstall zone-proxy-payments`
+
+##### 2. Per-Mesh Services (Not Shared)
```
lobkovilya: same as we do today but per mesh https://github.com/kumahq/kuma/blob/master/deployments/charts/kuma/templates/egress-service.yaml

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,998 @@
...
+#### Namespace Placement Options
+
+##### Option A: kuma-system Namespace (Centralized)
+
+```yaml
+# All zone proxies in kuma-system
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: zone-proxy-payments-mesh
```
lobkovilya: are we okay with mesh name being 63 chars at max?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,998 @@
...
+
+##### Option B: Per-Mesh Namespace
+
+```yaml
+# Zone proxy in mesh's namespace
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: zone-proxy
+  namespace: payments-mesh  # or payments-system
```
lobkovilya: probably can be a variable in values if user is not happy with the default being `kuma-system`

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,998 @@
...
+
+```yaml
+labels:
+  kuma.io/proxy-type: zoneproxy
+  kuma.io/ingress-enabled: "true"
+  kuma.io/egress-enabled: "true"
+```
+
+The `kuma.io/zone-proxy-role` label controls which listeners the control plane generates:
+- `both`: Generate both ingress and egress listeners (default)
```
lobkovilya: nit: `both` probably should be `all` or go with separate labels like `kuma.io/zone-egress: enabled`

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,998 @@
...
+- Advantages:
+  - Clear operational model - zone proxies are infrastructure
+  - Independent scaling and lifecycle management
+  - Familiar pattern for infrastructure operators
+  - Lower resource overhead
+  - Simpler debugging and monitoring
+- Disadvantages:
+  - Different pattern from application sidecars (minor)
+
+**Option C: Direct pod without injection**
```
lobkovilya: does it work with metrics hijacker?

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: > Discussed offline: is it printed by kuma-dp when bootstrap fails?
yes, just like other bootstrap failures - will add more detail

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: added notes, not sure how to handle this

bartsmykla (dismissed):

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,999 @@
...
+      role: all  # or: ingress, egress
+      replicas: 2
+    - name: orders-mesh
+      role: all
+      replicas: 1
+```
+
+Note: `meshes[]` only configures zone proxy deployment per mesh. Mesh creation (mTLS, backends, etc.) is managed separately on the Global CP.
+
+**Why this works for Konnect:**
```
lukidzi, line 180: Does it mean you need to know all meshes before even creating them? What if you create a Mesh later after the deployment?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,999 @@
...
+  - Simpler configuration
+  - Zone proxies are infrastructure, not application workloads
+  - Targeted by the `kuma.io/proxy-type` label (e.g., `zoneegress`) or extended `proxyTypes` in targetRef
+  - The Deployment itself is the workload - no parent correlation needed
+- Disadvantages:
+  - Slight inconsistency with application sidecar pattern (acceptable)
+
+#### Recommendation
+
+**Option B: Not required** - Zone proxies should not require `kuma.io/workload` annotation.
```
lukidzi: On universal `kuma.io/workload` is used to create identity for a Dataplane. In case of egress, either we automatically add `kuma.io/workload: app-name` or we can require it to be set, since user needs to generate a token for this sidecar.

lukidzi (commented):
I think the structure of the madr is not easy to track, is it possible to change it so it's clear which options we have and which we want to go? So far I saw 4 different deployment models and each presenting different types which isn't obvious to me

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: For konnect no, for on-prem yes (or we could have this hook option - https://github.com/kumahq/kuma/pull/15460/changes#diff-e1658531d0b23a0ad796f31bc5c31b19d0d5b092712cb84bf0d625b85bf4c771R180 )

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+└─────────────────────────────────────────────────────────┘
+```
+
+Generates `values.yaml` with `ingress.enabled: true` / `egress.enabled: true`
+
+##### Proposed Flow
+
+**Step 1: UI Enhancement**
+
+Replace simple checkboxes with a mesh-aware configuration:
```
slonka: How to handle this flow:
1. The user creates global cp on konnect
2. They create a zone without meshes
3. They have a zone without any ingress or egress or any sidecars.
4. They create a mesh at some time in the future on the global CP
5. They restart some deployments and that makes the app pick up a DPP
(let's focus on egress for now because there is a slight chance that we're not going to touch ingress)
6. They create a MES and they can't reach it (it would be the same situation even with "global egress" - when you decide you need egress later not on zone creation). It would be nice if there was "edit zone" that would regenerate values / helm command.
7. They have to go and do `helm upgrade` with egress for "global egress" and for "mesh scoped egress" they have to do this for every mesh they create.

Side things:
1. Maybe we need a flag called "egressForAllMeshes" or something like that, but it's not possible without having a list of meshes - maybe `kumactl` could spit it out.
2. Maybe we need egress instance controller? We thought that we might convince everyone they just need one mesh but that did not happen (we already have people firm on having multiple meshes).

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+    enabled = true
+    backend {
+      name = "builtin"
+      type = "builtin"
+    }
+  }
+}
+
+# Zone proxy explicitly depends on mesh
+resource "kuma_zone_proxy" "payments_proxy" {
```
slonka: that would be a module that underneath is a k8s deployment

or

just a k8s deployment resource that's configured the right way and documented the right way

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+| Tool | Can Validate Mesh? | Strategy |
+|------|-------------------|----------|
+| **Konnect UI** | Yes (API access) | Pre-populate mesh list, validate before generating YAML |
+| **Helm (federated)** | No (offline install) | Accept names, fail gracefully at runtime with clear logs |
+| **Helm (unfederated)** | Yes (creates mesh) | Cross-reference in templates, fail at install if mismatch |
+| **Terraform** | Yes (API access) | Validate at plan/apply, encourage resource references |
+
+#### Design Decisions
+
+##### 1. Mesh Deletion Handling
```
slonka: If we go with controller we need to make sure we don't end up in cyclic dependencies. On DPP we need some kind of indication that it's managed by controller so we can skip it on the "online dpps check".

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+| **Failure blast radius** | ✅ Issues isolated to single mesh |
+| **Cleanup** | ✅ `helm uninstall zp-payments` removes just that mesh |
+| **GitOps** | ⚠️ Multiple files/releases to track |
+| **Scaling** | ✅ Different resource profiles per mesh |
+
+**Best for**: Large deployments, multi-team environments, meshes with different SLAs.
+
+#### Namespace Placement Options
+
+**K8s Naming Constraint**: Kubernetes resource names (Deployments, Services) are limited to 63 characters.
```
slonka: If we go with controller then we need another resource, like MeshEgressInstance and the resources will be in that namespace. If we were to create that in one namespace (e.g. system namespace) then we need to solve for collisions.

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+**Best for**: Teams wanting separation from CP but not full per-mesh isolation.
+
+##### Recommendation
+
+**Default to kuma-system**, make namespace configurable:
+
+```yaml
+meshes:
+  - name: payments-mesh
+    namespace: kuma-system  # default, configurable via Helm values
```
slonka: if you want labels you need to add labels here (and then anything else needs to be defined here, so it's copy pasted everything)

slonka: 1. We could have a subchart (that would have values files), current egress has a bunch of switches we could take that
2. Subchart has this nice thing that helm require namespace so that would be on helm

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+  - name: orders-mesh
+    namespace: orders-system  # override for this mesh
+```
+
+This allows gradual migration to per-mesh namespaces without breaking existing setups.
+
+### Question 1: Unified vs Separate Zone Proxies
```
slonka: since we might not be touching ingress let's not talk about this yet, the other MADR will address this

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+Option B is still recommended because zone proxies are infrastructure that should update with the control plane.
+The rolling update strategy provides sufficient control over the upgrade process.
+
+#### Advantages and Disadvantages
+
+**Option A: Sidecar to fake container**
+- Advantages:
+  - Consistent with application sidecar pattern
+  - Could leverage existing injection infrastructure
+  - Manual upgrade control - operator decides when to roll pods
```
slonka:
```suggestion
  - Manual upgrade control - operator decides when to roll pods
  - You don't need special `limit / resources` for egress / ingress you can just use containerpatches for regular sidecars.
```

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+  - Familiar pattern for infrastructure operators
+  - Lower resource overhead
+  - Simpler debugging and monitoring
+  - **Prometheus metrics support**: Unified zone proxy as Dataplane uses `DefaultProxyProfile` which includes `PrometheusEndpointGenerator` (see [`pkg/xds/generator/proxy_template.go#L78-L92`](https://github.com/kumahq/kuma/blob/master/pkg/xds/generator/proxy_template.go#L78-L92)).
+    This enables Prometheus metrics scraping via the metrics hijacker.
+    Current ZoneIngress/ZoneEgress profiles don't include this generator.
+- Disadvantages:
+  - Different pattern from application sidecars (minor)
+
+**Option C: Direct pod without injection**
```
slonka: remove this

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+```
+
+#### New Mesh-Scoped Deployment
+
+With the move to mesh-scoped zone proxies and the unified model, the deployment changes:
+
+1. Zone proxies become `Dataplane` resources with specific labels
+2. Each resource must specify a `mesh` field
+3. Deploy one zone proxy instance **per mesh** (not one per zone for all meshes)
+4. The `kuma.io/zone-proxy-role` label determines ingress/egress/all capabilities
```
slonka: let's just have the specifics of this in the other MADR

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+3. Scale down old global-scoped proxy after verification
+4. Remove legacy ZoneIngress/ZoneEgress resources
+
+#### Tooling Updates Required
+
+- `kumactl` commands for zone proxy management
+- Documentation for Universal deployment
+- Migration guide from global to mesh-scoped
+
+### Question 4: Require kuma.io/workload Annotation
```
slonka: todo: what tokens do we use for universal for ingress / egress - we should do the same + workload

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+
+**Auto-generated workload identity**: On Universal, `kuma.io/workload` is used to create identity for a Dataplane (required for token generation).
+For zone proxies, the workload will be auto-generated with the pattern: `zone-proxy-<mesh>-<role>`.
+Example: `zone-proxy-payments-mesh-all`.
+This enables token generation on Universal without manual configuration.
+
+Since zone proxies become mesh-scoped Dataplane resources, targeting a zone proxy for a specific mesh is straightforward: policies themselves are mesh-scoped, so a policy in `payments-mesh` automatically only applies to zone proxies in that mesh.
+The `kuma.io/zone-proxy-role` label then filters to specific roles (ingress, egress, or all) within that mesh.
+
+### Question 5: Support kuma.io/ingress-public-address
```
slonka: if we don't touch ingress then this is not a problem so let's see what's the outcome of the other madr

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,1042 @@
...
+  - Forces workarounds in complex network topologies
+
+#### Recommendation
+
+**Option A: Keep annotation support** but document it as an escape hatch:
+- Primary method: Configure Service (LoadBalancer/NodePort) correctly
+- Annotation: Use only when Service address is not accessible from other zones
+- Consider adding a deprecation warning in logs when annotation is used
+
+### Question 6: Default Helm Installation Behavior
```
slonka: product question:
- when you start global CP do you want default mesh

maybe this should be separate MADR?

this MADR show how we put stuff into values but what we put there is a different question

slonka:
Edited:

> 1. Let's have some indication on which parts of this doc are ingress or egress specific

I think that only `kuma.io/ingress-public-address` is ingress only other's are both.

> 2. Let's have a PoC of the subchart approach.

In pocs folder.

docs/madr/decisions/094-zone-proxy-deployment-model.md
lahabana: We could rework the flow to direct folks towards creating a mesh first (and create a default mesh for non konnect install).
Then in the multi mesh case I think this has to be for advanced users so I see 2 options:
1. Have a list of meshes in the helm chart with creates each deployment (there's an issue as we said that namespaces are dedicated to meshes)
2. Don't support more than 1 mesh in Helm and have folks manage zoneProxies themselves.

> They create a MES and they can't reach it (it would be the same situation even with "global egress" - when you decide you need egress later not on zone creation). It would be nice if there was "edit zone" that would regenerate values / helm command.

This we should most likely to make sure the GUI has good feedback.

> Maybe we need egress instance controller? We thought that we might convince everyone they just need one mesh but that did not happen (we already have people firm on having multiple meshes).

I'd prefer we stay away from this TBH

As a side note this becomes a lot easier if we used KO instead of Helm

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: this is a token for zone proxy (egress, same for "ingress" just different scope):

```
{
  "Zone": "us-east",
  "Scope": [
    "egress" 
  ],
  "exp": 1772885592,
  "nbf": 1770293587,
  "iat": 1770293587,
  "jti": "2df65330-361f-4222-8f28-7c6429d41253"
}
```

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: > As a side note this becomes a lot easier if we used KO instead of Helm

Do you mean https://github.com/Kong/kong-operator?

> We could rework the flow to direct folks towards creating a mesh first (and create a default mesh for non konnect install).

Can you elaborate on what you mean by "creating a mesh first"? For federated zones the mesh is created on Global CP and synced via KDS, so the zone Helm install can't create it. For unfederated zones, the MADR already describes Helm creating Mesh resources locally (line 264-276). Do you mean something like: always create a default mesh on zone install (even federated) and deploy a zone proxy for it, so the single-mesh case works out of the box?

> 1. Have a list of meshes in the helm chart
> 2. Don't support more than 1 mesh in Helm and have folks manage zoneProxies themselves.

Both of these are already covered:
- (1) is the `meshes[]` list used throughout the Tooling flows (line 132-139) and Helm Release Options 1-3 (lines 422-508)
- (2) would be `Option 3` (line 483) where each mesh gets its own Helm release, or just manual kubectl - the MADR doesn't currently frame it as "Helm only handles 1 mesh" explicitly though. Worth calling out as a simplification?

slonka: While chatting with Ilya - I'm adding more Qs:

> direct folks towards creating a mesh first

By this do you mean in the GUI we should have some indication after creating global that the user should create a mesh as a next step?

> default mesh

Whenever you deploy global there would be a default mesh and since there is no spec it won't cause problems for terraform.

1. What do we do with `skipMeshCreation`?  
We keep the flag.

2. What about terraform and default value for mesh parameter?
Can we set a default there? If not maybe we could have tf provider expose a var/const.

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,616 @@
...
+The mesh validator returns an error: `"unable to delete mesh, there are still some dataplanes attached"`.
+See [`pkg/core/managers/apis/mesh/mesh_validator.go#L70-L88`](https://github.com/kumahq/kuma/blob/master/pkg/core/managers/apis/mesh/mesh_validator.go#L70-L88).
+
+**Mesh-Scoped Zone Proxy Benefit**: With zone proxies represented as Dataplane resources, this protection applies automatically.
+No additional implementation is needed for mesh deletion handling - the existing safeguard covers the new deployment model.
+
+**Note**: Current ZoneIngress/ZoneEgress resources are NOT covered by this protection (they're global-scoped).
+The move to mesh-scoped Dataplanes resolves this gap.
+
+**K8s Deployment cleanup**: When a mesh is deleted (after removing its Dataplanes), the K8s Deployment for zone proxies remains.
```
lobkovilya: I think user removes Dataplanes by removing the deployment, so no additional cleanup is required

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,616 @@
...
+For single-mesh, cleanup is straightforward: `helm uninstall <release-name>`.
+
+##### 2. Per-Mesh Services (Not Shared)
+
+**Decision**: Each mesh gets its own Service/LoadBalancer.
+
+**Rationale**: With mesh-scoped zone proxies:
+- Each mesh has **different mTLS CA certificates**
+- Zone ingress must present the correct mesh's certificate
```
lobkovilya: ZI doesn't need the CA, only ZE

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,616 @@
...
+**Service naming**: Follows the existing pattern from [`deployments/charts/kuma/templates/egress-service.yaml`](https://github.com/kumahq/kuma/blob/master/deployments/charts/kuma/templates/egress-service.yaml) but with per-mesh naming: `<release>-<mesh>-zoneproxy` (e.g., `kuma-payments-mesh-zoneproxy`).
+This prevents name collisions when multiple meshes are deployed.
+
+**Name length validation**: Helm templates should validate that the combined name fits within the 63-character Kubernetes limit:
+
+```
+{{ if gt (add (len .Values.zoneProxy.mesh) (len $prefix)) 63 }} {{ fail "zone proxy service name exceeds 63 characters; use a shorter mesh name" }} {{ end }}
+```
+
+This catches the issue at install time rather than silently truncating.
```
lobkovilya: but then what's left to do for the user? we can't expect they're going to rename the mesh right?

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka:
```suggestion
```

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka:
```suggestion
```

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: we need to have an easy way to access the name from the template and ask the user to set their own name

slonka: for deployment it's 255 so mesh name + suffix should be ok - but let's check that

slonka:
General Q:
- check what resources we need - do we need RBAC?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,614 @@
...
+
+**Cost implication**: More LoadBalancers = higher cloud cost.
+Users can use NodePort or Ingress controllers to reduce LB count if needed.
+
+##### 3. Migration Path from Global Zone Proxies
```
slonka: For 2.14 when user does `helm upgrade` they should use global scope zone proxies (no change). When you create a new zone in konnect you'll have the new zone proxies.

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,614 @@
...
+   See [MADR 090](090-zone-egress-identity.md) for detailed analysis.
+
+2. **Cannot apply policies on zone proxies**: Kuma policies (MeshTrafficPermission, MeshTimeout, etc.) are mesh-scoped.
+   A global zone proxy cannot be targeted by mesh-specific policies, limiting observability and traffic control for cross-zone communication.
+
+To resolve these limitations, zone proxies are being changed to **mesh-scoped** resources represented as Dataplane resources with specific tags.
+This architectural change requires revisiting the deployment model for zone proxies.
+
+**Scope of this document**: This MADR focuses on **deployment tooling** — how users deploy zone proxies via Helm, Konnect UI, and Terraform.
+The resource model (Dataplane representation, labels, tokens, workload identity) is covered in a separate MADR. [^1]
```
slonka:
```suggestion
```

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,621 @@
+# Zone Proxy Deployment Model
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/9030
+
+## Context and Problem Statement
+
+Currently, zone proxies are **global-scoped** resources, meaning a single ZoneIngress or ZoneEgress instance handles traffic for all meshes in a zone.
```
Automaat: wouldnt cluser scope be better term here?

lobkovilya: that's the term we're using in the code https://github.com/kumahq/kuma/blob/73f6aa6f30d5ee35a78987a771de93af19e04e3c/pkg/core/resources/model/resource.go#L51

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,621 @@
...
+```
+
+**Mesh creation in unfederated zones**: Unlike federated zones (where the mesh is synced from the Global CP), unfederated zones will create the mesh locally. There are two mechanisms:
+
+1. **CP auto-creation**: When `skipMeshCreation: false` (the default), the CP creates the `default` mesh at startup (`EnsureDefaultMeshExists` in `pkg/defaults/mesh.go`). This only creates a mesh named `default` — non-default mesh names are not auto-created.
+
+2. **Helm template**: For non-default mesh names, or when `skipMeshCreation: true`, Helm can create the Mesh resource:
+
+```yaml
+{{- if and .Values.zoneProxy.enabled (not .Values.controlPlane.kdsGlobalAddress) }}
```
lobkovilya: should we also check `not skipMeshCreation` here? otherwise user might end up with 2 meshes without expecting that

Automaat (dismissed):

bartsmykla (dismissed):

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,621 @@
...
+  enabled: false  # Disable global
+egress:
+  enabled: false
+
+zoneProxy:
+  enabled: true
+  mesh: default
+```
+
+#### Helm Release Structure Options
```
lobkovilya: so what option is selected here?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,621 @@
...
+| **Failure blast radius** | ✅ Bad zone proxy config doesn't break CP |
+| **GitOps** | ✅ Clear separation of concerns |
+
+**Best for**: Production environments wanting CP stability isolation.
+
+##### Helm Chart Structure: Pod Spec Passthrough
+
+**Problem**: Current Helm charts enumerate Kubernetes fields one by one — the ingress chart alone is 292 lines for ~15 fields. Users are blocked when they need an unsupported field (e.g., `shareProcessNamespace`, `initContainers`, `topologySpreadConstraints`), creating a cat-and-mouse problem where the chart never fully covers the Kubernetes API surface.
+
+**Solution**: The zone proxy chart accepts raw `podSpec` and `containers` sections. Helm's `merge` overlays user values onto sensible defaults. Adding any PodSpec or container field requires zero template changes.
```
lobkovilya: do we need service spec as well?

bartsmykla (dismissed):

Automaat (dismissed):

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+1. Should we continue supporting `kuma.io/ingress-public-address` annotation?
+2. What should be the default Helm installation behavior for zone proxies?
+
+Note: Whether zone ingress and egress share a single deployment is addressed in a separate MADR. [^2]
+
+### Decision Summary
+
+| Tooling Decision | Choice |
+|------------------|--------|
+| Per-mesh Services | **Yes** - each mesh gets its own Service/LoadBalancer for mTLS isolation |
```
lukidzi, line 38: What means Per-mesh Services? What kind of services ?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+| Namespace placement | **kuma-system** |
```
lukidzi, line 39: Does it need to be system namespace? In case of mulitple meshes is it possible to change it or we want to stick to the one only?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+egress:
+  enabled: true
+
+# After (mesh-scoped)
+ingress:
+  enabled: false  # Disable global
+egress:
+  enabled: false
+
+zoneProxy:
```
lukidzi: What if I don't need egress but only ingress?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+**Subchart approach**: single release for the default mesh (simple, one `helm install`); standalone release of the same `kuma-zone-proxy` chart for additional meshes.
+
+##### Helm Chart Structure: Passthrough Values
+
+**Problem**: Current Helm charts enumerate Kubernetes fields one by one — the ingress chart alone is 292 lines for ~15 fields. Users are blocked when they need an unsupported field (e.g., `shareProcessNamespace`, `initContainers`, `topologySpreadConstraints`), creating a cat-and-mouse problem where the chart never fully covers the Kubernetes API surface.
+
+**Solution**: The zone proxy chart groups passthrough values by Kubernetes resource. Helm's `merge` overlays user values onto sensible defaults. Adding any PodSpec, container, or Service field requires zero template changes.
+
+```yaml
+# values.yaml — resource-grouped passthrough
```
lukidzi: Are we going to distinguish somehow between ingress and egress?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+#### Helm Release Structure: Subchart Approach
+
+The zone proxy is packaged as a `kuma-zone-proxy` subchart of the main `kuma` chart. This single packaging naturally supports both single-mesh and multi-mesh deployment modes.
+
+##### Single Release (Default — Single-Mesh)
+
+The `kuma-zone-proxy` subchart is a dependency of the main `kuma` chart. A single `helm install` deploys everything:
+
+```bash
+helm install kuma kuma/kuma -f values.yaml
```
lukidzi: Q: Does an install creates to helm releases or just one?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+
+HPA and PDB settings use dedicated fields (`autoscaling.*`, `pdb.*`) since they have few, well-defined options.
+
+This reduces template code from 292 lines to 71 lines while providing unlimited field coverage. See the [POC repo](https://github.com/slonka/poc-094-helm-zone-proxy) for a working demonstration.
+
+#### Namespace Placement Options
+
+**K8s Naming Constraint**: Service names are limited to 63 characters (DNS label, RFC 1123). Other resources (Deployments, HPAs, PDBs) use DNS subdomain names (253 chars) and are not a concern. With the naming pattern `<release>-<mesh>-zoneproxy`, long mesh names may exceed the Service name limit. Users can override via `zoneProxy.service.name` (see Per-Mesh Services above).
+
+##### Option A: kuma-system Namespace (Centralized) — Recommended for Single-Mesh
```
lukidzi: What if we want to deploy it in a different namespace? Would that be possible, or should we limit it to the system namespace?

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Kubernetes Services - specifically the `LoadBalancer` or `NodePort` Service that exposes the zone proxy to other zones. Today there's one global Service for zone ingress and one for zone egress. With mesh-scoped zone proxies, each mesh gets its own Service (e.g., `kuma-payments-mesh-zoneproxy`) because each mesh has a different mTLS CA, so they can't share a single LoadBalancer without complex SNI-based cert selection. This is detailed in the "Per-Mesh Services (Not Shared)" section below.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: For single-mesh (this MADR's scope), `kuma-system` keeps everything together with the CP. For multi-mesh, standalone `kuma-zone-proxy` releases can target any namespace — the Helm `--namespace` flag controls that. We could add a dedicated namespace option per mesh later, but the single-mesh default of `kuma-system` keeps things simple. I'll clarify this in the namespace section.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Whether ingress and egress can be deployed/enabled independently is addressed in a separate MADR (the "Zone Proxy Deployment Topology" MADR referenced in footnote [^2]). This MADR uses `zoneProxy` as a unified concept — the topology MADR will define how users opt into ingress-only, egress-only, or both. The distinction will likely be driven by labels/tags on the Dataplane resource rather than separate Helm flags.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Yes — the topology MADR [^2] will cover this. The current direction is that the ingress vs egress role will be dictated by labels on the Dataplane resource, rather than being entirely separate deployment types. The passthrough values structure (`service.*`, `deployment.*`) applies regardless of how the topology shakes out. This MADR intentionally defers that decision.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Just one. The `kuma-zone-proxy` is a subchart (dependency) of the main `kuma` chart, so `helm install kuma kuma/kuma` creates a single Helm release that includes both the CP and the zone proxy. A second Helm release is only needed for multi-mesh — installing the `kuma-zone-proxy` chart standalone for additional meshes.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: For the single-mesh default, `kuma-system` is recommended. For standalone releases (multi-mesh), users can install into any namespace via `helm install --namespace <ns>`. We don't need to restrict this — Helm handles it natively. I'll add a note clarifying that standalone releases support custom namespaces.

docs/madr/decisions/094-zone-proxy-deployment-model.md
lukidzi: We need to ensure there is transparent proxy disabled for a zoneproxy - even if it's enabled on the namespace we should have it disabled on the deployment

docs/madr/decisions/094-zone-proxy-deployment-model.md
lukidzi: It's just kubernetes specific what about universal not sure if naming `Service` is here a good option

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+Currently, zone proxies are **global-scoped** resources, meaning a single ZoneIngress or ZoneEgress instance handles traffic for all meshes in a zone.
+This global nature creates fundamental limitations:
+
+1. **Cannot issue MeshIdentity for zone egress**: MeshIdentity is mesh-scoped, but a global zone egress serves multiple meshes.
+   This creates identity conflicts and prevents proper mTLS certificate issuance.
+   See [MADR 090](090-zone-egress-identity.md) for detailed analysis.
+
+2. **Cannot apply policies on zone proxies**: Kuma policies (MeshTrafficPermission, MeshTimeout, etc.) are mesh-scoped.
+   A global zone proxy cannot be targeted by mesh-specific policies, limiting observability and traffic control for cross-zone communication.
+
```
lahabana, line 18: 3. Observability story is tricky (how do you share to a mesh a subset of all the observability)

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+To resolve these limitations, zone proxies are being changed to **mesh-scoped** resources represented as Dataplane resources with specific tags.
+This architectural change requires revisiting the deployment model for zone proxies.
+
+**Scope of this document**: This MADR focuses on **deployment tooling** — how users deploy zone proxies via Helm, Konnect UI, and Terraform.
+
+**Single-mesh focus**: This document assumes **single-mesh-per-zone as the default** deployment pattern.
+For multi-mesh scenarios, deploy additional zone proxies using separate Helm releases with a dedicated `kuma-zone-proxy` chart. A multi-mesh deployment guide will be provided separately.
```
lahabana: No let's not as we're just creating maintainance for a path we don't recommend.

I think for folks that NEED multi-mesh they can:

- Just write their own deployment
- Push for Mesh to be supported by KO

Our goal for K3 is really to streamline what folks can do. Adding a new helm short is a complete no go.

docs/madr/decisions/094-zone-proxy-deployment-model.md
lahabana: This is not good because that assumes `kuma-system` namespace will be marked as a "meshed" namespace which we don't want right?

lahabana: Also kuma-system is sort of meant to be admin realm here there's almost a Mesh operator persona comes in no?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+
+```
+┌─────────────────────────────────────────────────────────┐
+│ Connect zone                                            │
+├─────────────────────────────────────────────────────────┤
+│                                                         │
+│ Mesh: [default ▼]                                       │
+│ (if multiple meshes detected there will be info here on │
+│ how to handle that)                                     │
+└─────────────────────────────────────────────────────────┘
```
lahabana: What if no mesh?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,630 @@
...
+    mode: zone
+    zone: zone-1
+    kdsGlobalAddress: grpcs://us.mesh.sync.konghq.tech:443
+
+  zoneProxy:
+    enabled: true
+    # mesh defaults to "default"
+```
+
+If the user selected a non-default mesh name:
```
lahabana: Why not always do it?

lahabana (commented):
Is there a PoC for this?

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Good point. With a global zone proxy serving all meshes, observability (metrics, access logs, traces) can't be scoped to a specific mesh — everything is mixed. Mesh-scoped zone proxies make per-mesh observability possible via policies like MeshAccessLog, MeshMetric, etc. Added this as a third limitation.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Makes sense — we shouldn't commit to a separate chart or guide for a path we don't recommend. For users who need multi-mesh, they can write their own deployment or push for Mesh to be supported by KO. Simplified the multi-mesh mention and removed the subchart/standalone-release framing throughout.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Right — `kuma-system` has sidecar injection disabled and should stay that way. The zone proxy doesn't need sidecar injection since it *is* the proxy — it runs as a standalone Deployment with kuma-dp directly, connecting to the CP the same way current ZoneIngress/ZoneEgress do (via bootstrap, not sidecar injection). So `kuma-system` doesn't need to be "meshed" for this to work.

Added a clarification note in the Namespace section.

slonka: In our RBAC model, admin and mesh operator are the same persona/role (see [RBAC docs](https://developer.konghq.com/mesh/rbac/#kong-mesh-operator-admin-role)). Zone proxies are infrastructure managed by this single operator persona, so IMO `kuma-system` is the right place.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: If no mesh exists, the UI would show an empty state and ask whether to generate a \`values.yaml\` that creates a \`default\` mesh. The zone proxy can also be deployed targeting a mesh name that doesn't exist yet — it will wait and retry until the mesh is created.

Added empty-state handling to the Konnect UI flow.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: You're right, always including \`mesh:\` explicitly is clearer. Updated all examples to always include \`mesh: default\`.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Good point — "Per-mesh Services" is Kubernetes-specific. On Universal, the equivalent is per-mesh listeners on the zone proxy (different ports or addresses). Universal deployment specifics are deferred to the resource model MADR (listed in Out of Scope), but I can rename this to something more generic like "Per-mesh network endpoints" if you prefer. The core principle (separate endpoints per mesh for mTLS CA isolation) applies to both platforms.

slonka (commented):
@lahabana yes - the passthrough values Helm chart PoC is at https://github.com/slonka/poc-094-helm-zone-proxy (linked in the doc at line 437). - or do you mean something else?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,558 @@
...
+This architectural change requires revisiting the deployment model for zone proxies.
+
+**Scope of this document**: This MADR focuses on **deployment tooling** — how users deploy zone proxies via Helm, Konnect UI, and Terraform.
+
+**Single-mesh focus**: This document assumes **single-mesh-per-zone as the default** deployment pattern.
+Multi-mesh scenarios are out of scope for this document. Users who need multiple meshes per zone can write their own zone proxy deployment.
+
+This document addresses the following questions:
+
+1. Should we continue supporting `kuma.io/ingress-public-address` annotation?
```
lahabana: What's this annotation for?

docs/madr/decisions/094-zone-proxy-deployment-model.md
lahabana: I'm worried of the implication of this as it's this "one" special snowflake when the general effort here is to make it more "normal". Let's keep this in mind here.

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,558 @@
...
+```
+┌─────────────────────────────────────────────────────────┐
+│ Connect zone                                            │
+├─────────────────────────────────────────────────────────┤
+│                                                         │
+│ Mesh: [default ▼]                                       │
+│ If multiple meshes detected - there will be info here   │
+│ how to handle that                                      │
+│                                                         │
+│ If no meshes exist:                                     │
```
lahabana: How about we just don't allow creating a zone before we have a mesh? Sounds just simpler to me

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,558 @@
...
+```yaml
+kuma:
+  controlPlane:
+    mode: zone
+    zone: zone-1
+    kdsGlobalAddress: grpcs://us.mesh.sync.konghq.tech:443
+
+  zoneProxy:
+    enabled: true
+    mesh: default
```
lahabana: Isn't an option to shift the problem around.

Go with:

```
meshes:
- name: <x>
   zoneEgress: ...
   zoneIngress: ...
```

Also IIRC there's a bunch of possible params on the zone proxies we should consider.

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,558 @@
...
+##### Single-Mesh (Default)
+
+```yaml
+kuma:
+  controlPlane:
+    mode: zone
+    zone: zone-1
+    kdsGlobalAddress: grpcs://global-cp:5685
+
+  zoneProxy:
```
lahabana: I think that's indirectly the advantage of going with "mesh" because in this case you'll be able to figure that it's a Global CP or a non-federated Zone and create the mesh object

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Charly just reminded me that we can't have multiple meshes in the same namespace so this must be configurable

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Zone ingress advertises a public address so other zones can route traffic to it. By default, it auto-detects this from the K8s Service (LoadBalancer IP/hostname, or NodePort + node address). The `kuma.io/ingress-public-address` annotation overrides this when auto-detection does not match reality — e.g., behind NAT gateways, split DNS, or external LBs not integrated with K8s. See Question 1 section for the full priority chain and the decision to keep it.

Added inline context where the annotation is first mentioned.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Noted. The MADR 093 relaxation is general (not zone-proxy-specific) — it allows multiple meshes per namespace everywhere, with collision handling in the Workload controller. Zone proxies are infrastructure (like the CP), and the naming pattern `zone-proxy-<mesh>-<role>` avoids collisions. The `kuma-system` namespace does not need sidecar injection for zone proxies since they run kuma-dp directly. Will keep this in mind as the implementation progresses.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: For Konnect, agree — the UI can require at least one mesh to exist before generating zone proxy Helm values. The "deploy before mesh exists" behavior is just an operational reality of Helm (the user types a mesh name, it may not exist yet on the Global CP), not something we need to design for as a first-class flow. Simplified the Konnect UI to remove the empty-state prompt.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: The `meshes` list structure has been adopted (see latest revision). Re: `zoneIngress`/`zoneEgress` naming — currently using `ingress`/`egress` for brevity since they are already nested under a mesh entry. Open to renaming if the team prefers explicit naming. Re: additional params — the schema includes `podSpec`, `hpa`, `pdb`, `service`, `resources` mirroring the current chart.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Exactly - the mesh subchart can check for `kdsGlobalAddress` in the parent chart values to determine if it is federated or not, and conditionally render the Mesh resource. This is already described in the "Conditional rendering" section under Flow 3 (Unfederated Zone).

bartsmykla (dismissed):
lgtm, reviewed on a call, together with @slonka and @Automaat

Automaat (dismissed):

bartsmykla (dismissed):

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,533 @@
...
+    egress:                       # Separate egress deployment.
+      enabled: true
+      podSpec: {}
+      hpa: {}
+      pdb: {}
+      service:
+        name: ""                  # Optional override for auto-generated Service name.
+      resources: {}
+
+    combinedProxies:              # Single deployment running both ingress + egress roles.
```
lobkovilya, line 99: nit: maybe `combinedZoneProxies`

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,533 @@
...
+**Step 1: UI Enhancement**
+
+```
+┌─────────────────────────────────────────────────────────┐
+│ Connect zone                                            │
+├─────────────────────────────────────────────────────────┤
+│                                                         │
+│ Detected meshes:                                        │
+│                                                         │
+│ ☑ default     Ingress: [Yes ▼]  Egress: [Yes ▼]         │
```
lobkovilya: probably also can have `combined` here to mirror the values.yaml structure

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,533 @@
...
+
+**`createPolicies`** controls which default policies are created (see schema above). Derived from `skipCreatingInitialPolicies` in `mesh.proto`.
+
+**`combinedProxies`** merges ingress and egress into a single Deployment. Mutually exclusive with `ingress`/`egress` — Helm validates and errors if both are defined. References the separate deployment topology MADR [^2] for analysis.
+
+**Subchart implementation**: The subchart lives within the main `kuma` chart under `charts/mesh/` (or as a dependency). It is not published separately — users interact with it purely through the `meshes` list in `values.yaml`.
+
+```bash
+# Install with mesh subchart entries
+helm install kuma kumahq/kuma -n kuma-system -f values.yaml
```
lobkovilya, line 329: That's just how the regular installation looks like today, no? What's the difference with the subchart?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,533 @@
...
+|--------|----------|
+| **Simplicity** | One `helm install` deploys everything; `meshes` list is declarative |
+| **Isolation** | Each mesh subchart renders independent resources; no cross-mesh interference |
+| **GitOps** | Single `values.yaml` is the source of truth for all meshes in a zone |
+
+**`createPolicies`** controls which default policies are created (see schema above). Derived from `skipCreatingInitialPolicies` in `mesh.proto`.
+
+**`combinedProxies`** merges ingress and egress into a single Deployment. Mutually exclusive with `ingress`/`egress` — Helm validates and errors if both are defined. References the separate deployment topology MADR [^2] for analysis.
+
+**Subchart implementation**: The subchart lives within the main `kuma` chart under `charts/mesh/` (or as a dependency). It is not published separately — users interact with it purely through the `meshes` list in `values.yaml`.
```
lobkovilya: I don't understand the subchart thing. What it gives us? Can we just put the `meshes` array to the `values.yaml` of the main Kuma chart?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,533 @@
...
+For K8s naming constraints and `service.name` overrides, see Per-Mesh Services above.
+
+| Aspect | Analysis |
+|--------|----------|
+| **Operations** | All Kuma components in one place |
+| **RBAC** | Simple - one namespace to grant access |
+| **Monitoring** | Single namespace to scrape metrics |
+| **Multi-mesh** | MADR 093 relaxation allows coexistence; naming pattern avoids collisions |
+
+**Note on sidecar injection**: Zone proxies run `kuma-dp` directly as standalone Deployments (the same way current ZoneIngress/ZoneEgress work). They connect to the CP via bootstrap, not via sidecar injection. The `kuma-system` namespace does **not** need sidecar injection enabled for zone proxies to function.
```
lobkovilya, line 412: There is a question which is more related to my [MADR](https://github.com/kumahq/kuma/pull/15566).

We need a way to mark a Pod to be reconciled by the Pod controller, using either a label or an annotation. 

Do you think reusing `kuma.io/sidecar-injection: enabled` here is acceptable, even though there is no actual sidecar involved? If not, we probably need to introduce a dedicated Pod annotation/label like `kuma.io/zone-proxy: enabled` so that zone proxy pods in `kuma-system` are recognized by the pod controller.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Keeping `combinedProxies` for consistency with `ingress`/`egress` — all three use short names since they're already nested under a mesh entry. Adding `Zone` to just one would be inconsistent. Open to renaming all three (`zoneIngress`/`zoneEgress`/`combinedZoneProxies`) if the team prefers explicit naming.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Good point. Updated the mockup to include a deployment mode selector (Separate vs Combined) per mesh.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Exactly — the install/upgrade command is identical to today. The only user-facing difference is the `values.yaml` content: a `meshes` list instead of flat `ingress.enabled`/`egress.enabled`. The "subchart" in the previous revision was about internal template organization (see reply on the subchart comment below) — users never interact with it directly. Clarified the wording.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Yes, `meshes` goes directly in the main kuma chart's `values.yaml`. The "mesh subchart" was an internal template organization pattern, not a separate chart. Updated the MADR to use "per-mesh templates" instead.

The reason it exists: each mesh entry needs a full set of K8s resources (Deployment, Service, HPA, PDB, SA) per role (ingress/egress/combined). In Helm, traditional subcharts render exactly once — you can't say "render this subchart N times for N meshes." So internally, the chart uses named templates (e.g., `mesh.deployment`, `mesh.service`) that get called in a `range .Values.meshes` loop. A Helm library chart (`type: library`) is the idiomatic way to organize these reusable templates, but they could also just be named templates in `_helpers.tpl`.

From the user's perspective: `helm install kuma kumahq/kuma -f values.yaml` with a `meshes` list. The internal template structure is invisible. Clarified the MADR wording to deemphasize this — it now says the implementation can be a library chart, subchart, or named templates, and is left to the implementation.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Not really a fan of reusing `kuma.io/sidecar-injection`. A dedicated pod annotation like `kuma.io/zone-proxy: enabled` is the right approach. It fits the existing annotation-based dispatch pattern in pod_controller.go (ingress annotation → reconcileZoneIngress, egress → reconcileZoneEgress,
 gateway → reconcileBuiltinGatewayDataplane). We'd add a new branch: `kuma.io/zone-proxy: enabled` → reconcileDataplane, where the controller uses `kuma.io/zone-proxy-type` Service labels (from your MADR 095) to determine which listeners to generate.

slonka (commented):
(removed - posted in error)

docs/madr/decisions/094-zone-proxy-deployment-model.md
lobkovilya: If you just look at the path `meshes[].combinedProxies` it's a bit unclear what proxies are we talking about here. This naming works only if you see `ingress` and `egress` next to it. 

Adding `zone` to all three works for me as well. In my MADR I'm currently planning to have a listener type with`ZoneIngress` and `ZoneEgress` values.

docs/madr/decisions/094-zone-proxy-deployment-model.md
lobkovilya: I think a Service labeled with `kuma.io/zone-proxy-type` that targets regular sidecars with the `kuma.io/sidecar-injection` annotation should also result in `ZoneIngress` or `ZoneEgress` listeners being generated.

Because of that, from the code path perspective there is no real difference whether the Pod has `kuma.io/sidecar-injection` or a hypothetical `kuma.io/zone-proxy` annotation. That is why I am a bit hesitant to introduce a new one.

We also generally follow the rule that users set labels and the CP sets annotations. In this case, however, `kuma.io/zone-proxy: enabled` would be an annotation that users must set themselves. So maybe this should be a label instead?

lobkovilya: Let's move this discussion to my MADR. IMO current PR is ready to be merged.

lobkovilya (dismissed):

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: I think we can talk about this later / your PR (this is not strictly related to this PR) - if the flows are the same then I'm 60/40 in favor of reusing.

bartsmykla (dismissed):

Automaat (dismissed):

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,539 @@
...
+
+### Decision Summary
+
+| Tooling Decision | Choice |
+|------------------|--------|
+| Per-mesh Services | **Yes** - each mesh gets its own Service/LoadBalancer for mTLS isolation |
+| Namespace placement | **kuma-system** |
+| Deployment mechanism | **Helm-managed** (current pattern extended for mesh-scoped zone proxies) |
+| Helm release structure | **Per-mesh templates** (each mesh entry rendered by per-mesh templates) |
+| MADR 093 relaxation | **Yes** - allow multiple meshes per namespace, handle Workload collisions in controller |
```
lahabana: It's even reverting MADR-093 right?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,539 @@
...
+| Additive migration | **Yes** - `meshes` config alongside existing `ingress`/`egress` keys |
+
+| Question | Decision |
+|----------|----------|
+| 1. Support ingress-public-address? | **Yes** - keep as escape hatch |
```
lahabana: I'm sorry but I still don't get what this is for?
Would this still exist as a annotation on the pod or will it become an annotation on the k8s service?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,539 @@
...
+```yaml
+meshes:
+  - name: <mesh-name>            # Required. Name of the mesh this entry targets.
+    createMesh: false             # Optional. Render a Mesh resource (unfederated zones only).
+    createPolicies:               # Optional. Map of default policies to create with the mesh.
+      MeshCircuitBreaker: true    # Default circuit breaker for all traffic.
+      MeshRetry: true             # Default retry config (TCP + HTTP/GRPC).
+      MeshTimeout: true           # Default timeouts for sidecars + gateways.
+      TrafficPermission: true     # Allow-all traffic permission (requires createMeshDefaultRoutingResources=true).
+      TrafficRoute: true          # Default load-balancing route (requires createMeshDefaultRoutingResources=true).
```
lahabana: Let's not add the legacy ones (let's think as if we're 100% Kuma 3.0 already)

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,539 @@
...
+
+For single-mesh deployments, the `meshes` list has one entry. Multi-mesh deployments add additional entries.
+
+**Full mesh entry schema**:
+
+```yaml
+meshes:
+  - name: <mesh-name>            # Required. Name of the mesh this entry targets.
+    createMesh: false             # Optional. Render a Mesh resource (unfederated zones only).
+    createPolicies:               # Optional. Map of default policies to create with the mesh.
```
lahabana: should we take a list instead?

Also worth saying explicitly that this is off by default and that `createMesh and createPolicies` would fail when going federated or global no?).

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,539 @@
...
+For K8s naming constraints and `service.name` overrides, see Per-Mesh Services above.
+
+| Aspect | Analysis |
+|--------|----------|
+| **Operations** | All Kuma components in one place |
+| **RBAC** | Simple - one namespace to grant access |
+| **Monitoring** | Single namespace to scrape metrics |
+| **Multi-mesh** | MADR 093 relaxation allows coexistence; naming pattern avoids collisions |
+
+**Note on sidecar injection**: Zone proxies run `kuma-dp` directly as standalone Deployments (the same way current ZoneIngress/ZoneEgress work). They connect to the CP via bootstrap, not via sidecar injection. The `kuma-system` namespace does **not** need sidecar injection enabled for zone proxies to function.
```
lahabana, line 412: I guess we'll do this as a follow up MADR to just track this 1 change?

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,539 @@
...
+
+### Question 1: Support kuma.io/ingress-public-address
+
+From `pkg/plugins/runtime/k8s/controllers/ingress_converter.go:24-71`, the address resolution follows this priority:
+
+1. Pod annotations: `kuma.io/ingress-public-address` + `kuma.io/ingress-public-port`
+2. Service LoadBalancer IP/Hostname
+3. NodePort with node address (ExternalIP > InternalIP)
+
+**Decision: Keep the annotation** as an escape hatch for NAT gateways, split DNS, non-standard LB topologies, and on-premises environments with external load balancers. Primary method remains proper Service (LoadBalancer/NodePort) configuration. Consider adding a deprecation warning in logs when the annotation is used.
```
lahabana: Can the annotation move to the Service instead though?

lahabana (commented):
Looks good! I think it'd be worth showing what the mesh object would look like when `createMesh: true`. IMO it should be a mesh with all the v3 stuff turned on and no flexibility so this makes us future proof.

docs/madr/decisions/094-zone-proxy-deployment-model.md
lobkovilya: I think the decision on this annotation fits well the next MADR https://github.com/Kong/kong-mesh/issues/9151. So I'd just remove it from here

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Yes, you're right — it's reverting MADR 093, not just relaxing it. Updated the wording throughout.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Today it's a Pod annotation — `ingress_converter.go:78` reads from `pod.Annotations`. Moving it to the Service would be more natural since the Service is what exposes the address. This MADR just decides "keep it as an escape hatch" — whether it lives on Pod or Service is an implementation detail we can address when building it.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Agreed — removed `TrafficPermission` and `TrafficRoute`.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Good call — changed `createPolicies` to a list. Also added explicit defaults (`createMesh: false`, `createPolicies: []`) and a note that both are ignored when `kdsGlobalAddress` is set (federated zones).

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Agreed — I'll create a GitHub issue and a follow-up MADR that supersedes 093.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Agreed — added a recommendation in Question 1 to move the annotation from the Pod to the Service for mesh-scoped zone proxies, since the Service is where the address is exposed and we're writing new converter code anyway.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: Agreed — removed question 1 entirely. The annotation decision will be addressed in the resource model MADR.

docs/madr/decisions/094-zone-proxy-deployment-model.md
slonka: https://github.com/kumahq/kuma/issues/15631

Automaat (approved):

docs/madr/decisions/094-zone-proxy-deployment-model.md
```diff
@@ -0,0 +1,522 @@
...
+│                                                                          │
+│ Detected meshes:                                                         │
+│                                                                          │
+│ ☑ default     Mode: [Separate ▼]  Ingress: [Yes ▼]  Egress: [Yes ▼]     │
+│ ☑ payments    Mode: [Combined ▼]  Ingress: [--- ▼]  Egress: [--- ▼]     │
+│ ☐ backend     Mode: [--------- ▼]  Ingress: [--- ▼]  Egress: [--- ▼]    │
+│                                                                          │
+│ Mode options:                                                            │
+│   Separate — deploy ingress and egress as independent Deployments        │
+│   Combined — deploy a single Deployment running both roles               │
```
johncowen, line 152: Q: I'm guessing that "combined" here means that you don't need to select yes/no for Ingress and Egress right? 

Also, how many meshes do we expect users to have, like as a maximum? Do we need some sort of alternative "apply the same to all meshes" option or something?

slonka: > combined

combined is both ingress / egress, need to check if that was implemented

> Also, how many meshes do we expect users to have
just one, I think if there are more than 3 (?) we can just give the example how to set up one and have a comment for the user to fill out the rest - but happy to hear other's thoughts

slonka: Quoting Ilya:

> we decided we don’t need it as a special support in helm charts, user can always update ingress and egress services to select the same pods

PR #15499, opened 2026-02-02
feat(dataplane): skip zone tag on empty inbounds in Exclusive mode

### LINKED ISSUE #15435: Universal dataplanes don't generate inbound tags
Depends on #15430, #15431

## What

Stop generating `kuma.io/service` and `kuma.io/protocol` tags on Universal inbounds

## Why

- Tags no longer needed with label-based matching + protocol field
- Reduces data stored per inbound
- Cleaner dataplane resources
- Consistent with K8s approach (#15432)

## Requirements

- MeshService mode: `Exclusive`
- MeshIdentity: enabled

## Outcome

- Universal inbounds have empty tags when MeshService Exclusive mode + MeshIdentity enabled
- In other modes, tags still generated (backward compat)
- System works correctly without inbound tags on Universal

Automaat:
## Motivation

Don't force zone tag on Universal inbounds when tags are empty and MeshService mode is `Exclusive`. This allows users to submit dataplanes with empty inbound tags for skip inbound tag generation.

## Implementation information

Modified `setInboundsClusterTag()` in dataplane manager to:
- Accept mesh parameter to check MeshServices mode
- Skip adding zone tag when `len(inbound.Tags) == 0` AND mode is `Exclusive`

Added unit tests covering:
- Zone tag not added when empty tags + Exclusive mode
- Zone tag added when inbound has tags even in Exclusive mode
- Zone tag added to empty inbound when not in Exclusive mode
- Zone tag not added on update when empty tags + Exclusive mode

## Supporting documentation

Fixes #15435

### DIFF
--- a/pkg/core/managers/apis/dataplane/dataplane_manager_test.go
+++ b/pkg/core/managers/apis/dataplane/dataplane_manager_test.go
@@ -229,4 +229,166 @@ var _ = Describe("Dataplane Manager", func() {
 		Expect(err).ToNot(HaveOccurred())
 		Expect(actual.Spec.Networking.Inbound[0].State).To(Equal(mesh_proto.Dataplane_Networking_Inbound_NotReady))
 	})
+
+	It("should not add zone tag when inbound tags empty and MeshServices mode is Exclusive", func() {
+		// setup
+		s := memory.NewStore()
+		manager := dataplane.NewDataplaneManager(s, "zone-1", config_core.Zone, false, "", dataplane.NewMembershipValidator())
+		mesh := &core_mesh.MeshResource{
+			Spec: &mesh_proto.Mesh{
+				MeshServices: &mesh_proto.Mesh_MeshServices{
+					Mode: mesh_proto.Mesh_MeshServices_Exclusive,
+				},
+			},
+		}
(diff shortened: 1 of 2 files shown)


bartsmykla (approved):

pkg/core/managers/apis/dataplane/dataplane_manager_test.go
```diff
@@ -229,4 +229,166 @@ var _ = Describe("Dataplane Manager", func() {
...
+
+		actual := core_mesh.NewDataplaneResource()
+		err = s.Get(context.Background(), actual, store.GetByKey("dp1", "default"))
+		Expect(err).ToNot(HaveOccurred())
+
+		// then
+		Expect(actual.Spec.Networking.Inbound[0].Tags).To(BeEmpty())
+	})
+
+	It("should add zone tag when inbound has tags even in Exclusive mode", func() {
```
slonka, line 274: nitpick: these could be parametrized I think, only 3 lines are different between this and the above test (not something that's needed now I guess).

slonka (commented):
just for consideration: should `setInboundsClusterTag` be taken to some module and made public so we can test it more easily? 🤔

PR #15513, opened 2026-02-03
fix(kuma-dp): add logging for MeshMetric application scraping failures

Automaat:
## Motivation

MeshMetric application scraping failures are silent - when configured apps don't return metrics, there's no indication of why. This makes debugging issues like network failures, non-200 responses, or parsing errors difficult.

## Implementation information

Added logging at key failure points in the metrics collection pipeline:

| File | Log Level | Condition |
|------|-----------|-----------|
| `metrics_producer.go` | Debug | Start of collection (lists apps being scraped) |
| `metrics_producer.go` | Info | Non-200 HTTP status from app |
| `metrics_producer.go` | Info | Empty metrics after parsing |
| `metrics_producer.go` | Info | Zero total metrics despite configured apps |
| `server.go` | Info | Non-200 HTTP status |
| `server.go` | Debug | Empty response body |
| `metrics_format_mapper.go` | Debug | Empty input to format converter |

Log levels chosen:
- **Info**: Actionable issues users should investigate
- **Debug (V(1))**: Operational details for deeper troubleshooting

## Supporting documentation

N/A

### DIFF
--- a/app/kuma-dp/pkg/dataplane/metrics/server.go
+++ b/app/kuma-dp/pkg/dataplane/metrics/server.go
@@ -364,7 +364,8 @@ func (s *Hijacker) getStats(ctx context.Context, initReq *http.Request, app Appl
 		return nil, ""
 	}
 	if resp.StatusCode != http.StatusOK {
-		logger.Info("application returned non-200 status", "name", app.Name, "status", resp.StatusCode, "path", app.Path, "port", app.Port)
+		bodyBytes, _ := io.ReadAll(io.LimitReader(resp.Body, 1024))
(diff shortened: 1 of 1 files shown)


app/kuma-dp/pkg/dataplane/metrics/server.go
```diff
@@ -381,6 +385,9 @@ func (s *Hijacker) getStats(ctx context.Context, initReq *http.Request, app Appl
 			return nil, ""
 		}
 	}
+	if len(bodyBytes) == 0 {
```
lukidzi, line 389: In case of status in 4xx/5xx you never print a response body as it quits on status different than 200, should we check if there is a body to get more context?

Automaat: updated

PR #15537, opened 2026-02-05
feat(meshidentity): introduce an extension to the MeshIdentity

lukidzi:
## Motivation

We want to introduce ability to extends with customs plugin

## Implementation information

Introduce a custom type and json config

### DIFF
--- a/pkg/core/resources/apis/meshidentity/api/v1alpha1/meshidentity.go
+++ b/pkg/core/resources/apis/meshidentity/api/v1alpha1/meshidentity.go
@@ -36,8 +36,8 @@ type ProviderType string
 
 const (
 	BundledType   ProviderType = "Bundled"
-	ExtensionType ProviderType = "Extension"
 	SpireType     ProviderType = "Spire"
+	ExtensionType ProviderType = "Extension"
 )
 
 type Provider struct {
--- a/pkg/core/resources/apis/meshidentity/api/v1alpha1/validator.go
+++ b/pkg/core/resources/apis/meshidentity/api/v1alpha1/validator.go
@@ -112,6 +112,9 @@ func validateSpire(path validators.PathBuilder, b *Spire) validators.ValidationE
 	return verr
 }
 
+// validateExtension performs base validation for the Extension provider type.
+// Extension-specific config validation (e.g. parsing config fields) is delegated to the
+// extension implementation via registry.RegisterTypeValidator.
 func validateExtension(path validators.PathBuilder, b *Extension) validators.ValidationError {
 	var verr validators.ValidationError
 	if b == nil {
--- a/pkg/core/resources/apis/meshidentity/api/v1alpha1/testdata/spec/valid-extension-with-config.input.yaml
+++ b/pkg/core/resources/apis/meshidentity/api/v1alpha1/testdata/spec/valid-extension-with-config.input.yaml
@@ -1,0 +1,8 @@
+provider:
+  type: Extension
+  extension:
+    name: my-custom-provider
+    config:
+      key: value
+      nested:
+        setting: true
--- a/pkg/core/resources/apis/meshidentity/api/v1alpha1/testdata/spec/valid-extension.input.yaml
+++ b/pkg/core/resources/apis/meshidentity/api/v1alpha1/testdata/spec/valid-extension.input.yaml
@@ -1,0 +1,4 @@
+provider:
+  type: Extension
+  extension:
+    name: my-custom-provider
--- a/pkg/core/resources/apis/meshidentity/providers/extension/extension.go
+++ b/pkg/core/resources/apis/meshidentity/providers/extension/extension.go
@@ -1,0 +1,86 @@
+package extension
+
+import (
+	"context"
+	"fmt"
+	"sort"
+
+	"github.com/kumahq/kuma/v2/pkg/core"
+	meshidentity_api "github.com/kumahq/kuma/v2/pkg/core/resources/apis/meshidentity/api/v1alpha1"
+	"github.com/kumahq/kuma/v2/pkg/core/resources/apis/meshidentity/providers"
+	"github.com/kumahq/kuma/v2/pkg/core/xds"
+)
+
+var log = core.Log.WithName("identity-provider").WithName("extension")
+
+// Dispatcher routes MeshIdentity Extension requests to handlers registered by
+// Extension.Name. It implements providers.IdentityProvider and is registered as
+// the single "Extension" provider in the identity manager.
+type Dispatcher struct {
+	handlers map[string]providers.IdentityProvider
+}
+
+var _ providers.IdentityProvider = &Dispatcher{}
+
+// NewDispatcher creates a dispatcher that routes to extension sub-handlers by name.
+func NewDispatcher(handlers map[string]providers.IdentityProvider) *Dispatcher {
+	return &Dispatcher{handlers: handlers}
+}
+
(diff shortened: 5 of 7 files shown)


pkg/core/resources/apis/meshidentity/providers/interface.go
```diff
@@ -13,6 +13,7 @@ type IdentityProvider interface {
 	Initialize(context.Context, *meshidentity_api.MeshIdentityResource) error
 	CreateIdentity(context.Context, *meshidentity_api.MeshIdentityResource, *xds.Proxy) (*xds.WorkloadIdentity, error)
 	GetRootCA(context.Context, *meshidentity_api.MeshIdentityResource) ([]byte, error)
+	ShouldCreateMeshTrust(*meshidentity_api.MeshIdentityResource) (bool, error)
```
slonka: `ShouldCreateMeshTrust` feels like a leaky abstraction on this interface. MeshTrust creation is an orchestration concern (status updater), not a provider concern. The bundled implementation just reads `mid.Spec.Provider.Bundled.MeshTrustCreation` - something the updater was already doing directly. Adding this to the interface forces every provider to implement a method that most return `false, nil` for.

A simpler approach: keep the decision in the updater. The updater already switches on provider type and has full access to the spec. Alternatively, the updater could just check if `GetRootCA` returns non-nil CA bytes (which the refactored code already does with the `if ca != nil` guard).

pkg/core/resources/apis/meshidentity/status/updater.go
```diff
@@ -177,30 +177,51 @@ func (i *IdentityProviderReconciler) initialize(ctx context.Context, mid *meshid
...
+	if err != nil {
+		conditions = append(conditions, common_api.Condition{
+			Type:    meshidentity_api.MeshTrustConditionType,
+			Status:  kube_meta.ConditionFalse,
+			Reason:  "MeshTrustCreationError",
+			Message: err.Error(),
+		})
+		return conditions
+	}
+	if ca != nil {
```
slonka: Subtle behavior change: if the extension provider returns `ShouldCreateMeshTrust=true` but `GetRootCA` returns `nil` (like Spire does), this block is silently skipped and **no MeshTrust condition is set at all** - neither success nor failure. The caller checks conditions to determine overall readiness.

In the original code, a nil CA inside `createOrUpdateMeshTrust` would have stored an empty PEM. Was the silent skip intentional? Should there be an explicit condition when `ca == nil` after `ShouldCreateMeshTrust` returned `true`?

pkg/core/resources/apis/meshidentity/api/v1alpha1/validator.go
```diff
@@ -47,6 +47,8 @@ func validateProvider(provider Provider) validators.ValidationError {
 		verr.Add(validateBundled(validators.RootedAt("bundled"), provider.Bundled))
 	case SpireType:
 		verr.Add(validateSpire(validators.RootedAt("spire"), provider.Spire))
+	case ExtensionType:
+		verr.Add(validateExtension(validators.RootedAt("extension"), provider.Extension))
 	default:
 		verr.AddError("type", validators.MakeFieldMustBeOneOfErr(string(provider.Type), string(BundledType), string(SpireType)))
```
slonka: This error message still only lists `BundledType` and `SpireType` but should now also include `ExtensionType`:
```suggestion
		verr.AddError("type", validators.MakeFieldMustBeOneOfErr(string(provider.Type), string(BundledType), string(SpireType), string(ExtensionType)))
```

pkg/core/resources/apis/meshidentity/api/v1alpha1/validator.go
```diff
@@ -110,6 +112,21 @@ func validateSpire(path validators.PathBuilder, b *Spire) validators.ValidationE
...
+func validateExtension(path validators.PathBuilder, b *Extension) validators.ValidationError {
+	var verr validators.ValidationError
+	if b == nil {
+		verr.AddViolationAt(path, "configuration needs to be defined")
+		return verr
+	}
+	if b.Type == "" {
+		verr.AddViolationAt(path.Field("type"), "extension type needs to be defined")
+	}
+	if b.Config == nil {
```
slonka: Is requiring `config` to be non-nil intentional? The CRD schema marks `config` as optional (not in the `required` list) and uses `x-kubernetes-preserve-unknown-fields`. Some extensions might legitimately need only a `type` string and no additional config. Having the CRD accept it but the validator reject it is surprising.

pkg/core/resources/apis/meshidentity/api/v1alpha1/meshidentity.go
```diff
@@ -98,6 +102,13 @@ type Spire struct {
 	Agent *SpireAgent `json:"agent,omitempty"`
 }
 
+type Extension struct {
+	// Type of the extension.
+	Type string `json:"type"`
```
slonka: nit: Having `Provider.Type = "Extension"` and then `Extension.Type` is a two-level type discrimination that could be confusing to read.

Current:
```yaml
provider:
  type: Extension
  extension:
    type: my-custom-provider
    config:
      key: value
```

Suggested - rename `type` to `name` to differentiate from `provider.type`:
```yaml
provider:
  type: Extension
  extension:
    name: my-custom-provider
    config:
      key: value
```

pkg/core/resources/apis/meshidentity/api/v1alpha1/meshidentity.go
```diff
@@ -98,6 +102,13 @@ type Spire struct {
...
+	// Config freeform configuration for the extension.
```
slonka: nit: Godoc reads oddly - "Config freeform configuration for the extension." Should be `// Config is a freeform configuration for the extension.`

pkg/core/resources/apis/meshidentity/providers/bundled/bundled.go
```diff
@@ -164,6 +164,10 @@ func (b *bundledIdentityProvider) GetRootCA(ctx context.Context, identity *meshi
 	return cert, err
 }
 
+func (s *bundledIdentityProvider) ShouldCreateMeshTrust(mid *meshidentity_api.MeshIdentityResource) (bool, error) {
```
slonka: nit: Receiver name is `s` here but all other methods in this file use `b` for `bundledIdentityProvider`.

pkg/core/resources/apis/meshidentity/api/v1alpha1/testdata/spec/invalid-extension.input.yaml
```diff
@@ -0,0 +1,8 @@
+spiffeID:
+  trustDomain: "{{ .Mesh }}.{{ .Zone }}.mesh.local"
+  path: "/ns/.Namespace/sa/{{ .ServiceAccount }}"
+provider:
+  type: Extension
+  extension:
+    type: ""
+    config:
```
slonka: nit: Missing newline at end of file.

pkg/core/resources/apis/meshidentity/api/v1alpha1/validator.go
```diff
@@ -110,6 +112,18 @@ func validateSpire(path validators.PathBuilder, b *Spire) validators.ValidationE
 	return verr
 }
 
+func validateExtension(path validators.PathBuilder, b *Extension) validators.ValidationError {
```
lobkovilya, line 118: Is there a way to pass validation func for a given extension from the KM?

lukidzi: Yes, added comment about it.

PR #15566, opened 2026-02-09
docs(MADR): mesh-scoped zone proxies representation

lobkovilya:
## Motivation

We want to replace global-scoped ZoneIngress/ZoneEgress with mesh-scoped Dataplane (with new fields for zone proxy specific info).

## Supporting documentation

Fix https://github.com/Kong/kong-mesh/issues/9028

### DIFF
--- a/docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
+++ b/docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
@@ -358,6 +358,12 @@ Out of scope for this MADR. See https://github.com/Kong/kong-mesh/issues/9151.
 
 ## Implementation
 
+### Prerequisites
+
+New zone proxy listeners require unified resource naming and `meshServices.mode: Exclusive`.
+Legacy `kuma.io/service` labels and `ExternalService` resources are not supported.
+This constraint enables a cleaner implementation without legacy compatibility code.
+
 ### Avoiding excessive Envoy config for zone-proxy-only Dataplanes
 
 Dataplanes with only zone proxy listeners (no inbounds) should not have outbound listeners and clusters generated.


docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,188 @@
...
+        name: route-1
+      default: $conf # will be ignored on DPPs that don't have `route-1`
+```
+
+We should follow the same approach with policies targeting zone ingress and zone egress.
+
+```
+spec:
+  targetRef:
+    kind: Dataplane
```
lobkovilya, line 305: check if we allow kind Dataplane without labels?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,188 @@
...
+           - spiffeId:
+               type: Prefix
+               value: "spiffe://default/"
+```
+
+Field `targetRef.sectionName` can be used to select only zone egress or zone ingress listener. 
+
+#### Labels on Zone Ingress/Egress Dataplanes
+
+Currently we have `kuma.io/proxy-type: sidecar | gateway` label that can be set on the DPP.
```
lobkovilya: Check if we can remove it?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,188 @@
...
+    targetPort: 6739
+    appProtocol: tcp
+  - name: some-port
+    port: 16739
+    targetPort: target-port-from-container
+    appProtocol: tcp
+  zoneIngress: # different name? maybe 'multizone'?
+    address: 192.168.0.1 # advertised address of zone ingress
+    port: 10002 # advertised port of zone ingress
+    sni: "ae10a8071b8a8eeb8.backend.8080.demo.ms"
```
lobkovilya: todo: we probably need `zoneEgress` section in MES, similar with `sni` to avoid scanning through the list of DPPs when configuring MES outbound

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: @lukidzi yes, we do allow that https://github.com/kumahq/kuma/blob/5b70b39f0b9c6342a8cf9fe959c34a217b1f9bb4/pkg/core/resources/apis/mesh/validators_test.go#L167-L176

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: We can https://github.com/kumahq/kuma/issues/15567

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: done

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,229 @@
...
+      name: zi-port # optional, user should be able to set name since `port` can be the same when `addresses` are different
+    zoneEgress:
+      address: 10.0.0.2 # required
+      port: 10002 # required
+      name: ze-port # optional
+```
+
+##### Open Questions
+
+1. Should `networking.zoneIngress` and `networking.zoneEgress` be an array?
```
lahabana: Why?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,229 @@
...
+2. There is also `networking.advertisedAddress` field (but no `networking.advertisedPort`). 
+It was a very niche Universal-only field [contributed by community](https://github.com/kumahq/kuma/pull/2116) that I think we should remove in v3. 
+I think `networking.zoneIngress.advertisedAddress` should be `required` so it never falls back to `networking.advertisedAddress`.
```
lahabana: Sounds good. So in a more concrete manner we disallow using `networking.advertisedAddress` always in 2.14? Or do we warn in 2.14 and remove later?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,229 @@
...
+
```
lahabana: Can we be at the same time:

ZoneEgress, ZoneIngress, Regular DP?

lahabana: How do we filter in the API to get just the Ingress? Just the Egress?

Does that mean you can have multiple Egress/Ingress?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,229 @@
...
+* Bad, because items in `listeners` need to have the same schema so we'd have to validate `advertisedAddress` based on `type`
+
+### Mixing sidecars and zone proxies
+
+Dataplanes that represent zone proxies don't need `transparentProxying` even on Kubernetes.
+This means on Kubernetes we can't mix `inbounds` and zone proxy listeners on the same DPP.
+However, on Universal we can have both sidecar and zone proxy listeners on the same DPP since transparent proxying is not required.
+But the question is do we want to allow that? It might be confusing for users and doesn't allow us having schema level validation.
+
+### Policy Targeting for Zone Ingress and Zone Egress
```
lahabana, line 282: This should be its own MADR IMO

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,229 @@
...
+* Good, because `name` doesn't look awkward, it's obvious it has to be unique within the `listeners` array
+* Good, because in the future if we want clean DPP spec we can deprecate `networking.inbound` and add listeners with `type: Inbound`
+* Bad, because items in `listeners` need to have the same schema so we'd have to validate `advertisedAddress` based on `type`
+
+### Mixing sidecars and zone proxies
+
+Dataplanes that represent zone proxies don't need `transparentProxying` even on Kubernetes.
+This means on Kubernetes we can't mix `inbounds` and zone proxy listeners on the same DPP.
+However, on Universal we can have both sidecar and zone proxy listeners on the same DPP since transparent proxying is not required.
+But the question is do we want to allow that? It might be confusing for users and doesn't allow us having schema level validation.
```
lahabana: OK let's not allow it but we can at least be both Ingress and Egress.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,229 @@
...
+   rules:
+     - matches: [] # select specific FilterChain
+       default: # ignored if DPP doesn't have the requested FilterChain
+         deny:
+           - spiffeId:
+               type: Prefix
+               value: "spiffe://default/"
+```
+
+Field `targetRef.sectionName` can be used to select only zone egress or zone ingress listener.
```
lahabana: Can you show an actual example here?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,229 @@
...
+#### Labels on Zone Ingress/Egress Dataplanes
+
+Currently we have `kuma.io/proxy-type: sidecar | gateway` label that is automatically set on the DPP.
+The only purpose of this label was to let users select proxy types in `spec.targetRef.labels`.
+
+However, in v3 we're going to remove `gateway`.
+New `zoneEgress` and `zoneIngress` sections have `name` field and can be selected with `spec.targetRef.sectionName`.
+There are no reasons to keep `kuma.io/proxy-type` label and it should be removed in v3, see [#15567](https://github.com/kumahq/kuma/issues/15567).
+
+### Syncing Zone Ingress Addresses via MeshService
```
lahabana: Same other MADR

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,229 @@
...
+      k8s.kuma.io/namespace: kuma-demo
+  ports:
+  - port: 6739
+    targetPort: 6739
+    appProtocol: tcp
+  - name: some-port
+    port: 16739
+    targetPort: target-port-from-container
+    appProtocol: tcp
+  zoneIngress: # different name? maybe 'multizone'?
```
lahabana: We need a better name.

Also I feel like we should keep the same field names and on Ingress (`advertised`).

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,229 @@
...
+```yaml
+type: Dataplane
+mesh: default
+name: zone-ingress-1
+spec:
+  networking:
+    zoneIngress:
+      address: 10.0.0.1 # required, address listener binds to
+      port: 10001 # required, port listener binds to
+      advertisedAddress: 192.168.1.100  # required, address visible to other zones
```
lahabana: There's something wrong about this because advertisedAddress is common to all ZoneIngress...

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: We discussed this offline yesterday. It should never be an array. I missed that this question was still in the doc.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: There will be a separate MADR for policies. Here, I only want to cover how the top-level targetRef will work with the extended Dataplane resource. Since this MADR focuses on the new Dataplane API, I think this part belongs here.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: Not sure. This MADR explains how we’re going to convert the globally scoped ZoneIngress resource, which is synced everywhere, into a mesh-scoped Dataplane resource that is synced only to Global. IMO answering the syncing question here makes the design more complete. The entire doc is still only about 200 lines.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: It’s the same today, right? Multiple ZoneIngress resources in the same zone can have the same advertisedAddress when they're behind the LB

lahabana: The fact that it is like this today doesn't mean it's right :)

lobkovilya: TODO: 

There are 2 flows possible:
1. meshservice_controller.go watches `Service` which is labels as "ingress" and reconciles MS (or . In that case we don't need the `advertisedAddress` in DPP
2. pod_controller.go reconciles 

We'd like to get rid of the advertisedAddress in the DPP.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: TODO: think if we can make less exceptions if nothing wrong with inbound and zoneIngress/zoneEgress listeners working together then we probably shouldn't limit it

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: todo: single MeshZoneAddress resource per zone?

```yaml
spec:
  address: 192.168.0.1
  port: 10001
status:
  origin: # k8s service ???
```

todo: write why can't we use zone object?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: Created an issue to track the other MADR https://github.com/Kong/kong-mesh/issues/9151.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: topic of the next madr

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: Updated the MADR, no more advertisedAddress in the DPP

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: Yeah, I think we should deprecate in 2.14 and remove in 3.0. I'll open an issue

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: > ZoneEgress, ZoneIngress, Regular DP?

Yes, updated the MADR

> How do we filter in the API to get just the Ingress? Just the Egress?

We can add DPP labels like `kuma.io/zone-proxy-ingress: enabled`. But does our CP need to filter DPPs to get just ingress/egress? I think outbound generator shouldn't depend on specific DPPs and instead get addresses through the dedicated resources (i.e. MeshZoneAddress for ingress and MeshExternalService for egress, but that's for the next MADR).

> Does that mean you can have multiple Egress/Ingress?

Yes, DPPs are created per ingress/egress replica pods.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,279 @@
...
+```yaml
+type: Dataplane
+mesh: default
+name: zone-ingress-1
+spec:
+  networking:
+    zoneIngress:
+      address: 10.0.0.1 # required, address listener binds to
+      port: 10001 # required, port listener binds to
+      name: zi-port # optional, user should be able to set name since `port` can be the same when `addresses` are different
```
slonka: The `name` field's purpose isn't clear until much later (line 244) where `sectionName` is introduced. The pros/cons about `name` semantics (lines 58, 81) don't fully land without that context.

**Suggestion:** Add a brief mention here, e.g. `name: zi-port # optional, used for policy targeting via sectionName`

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,279 @@
...
+      name: ze-port # optional
+```
+
+Note: There is also a `networking.advertisedAddress` field (but no `networking.advertisedPort`).
+It was a niche Universal-only field [contributed by the community](https://github.com/kumahq/kuma/pull/2116) that should be removed in v3.
+Otherwise, it may appear that `advertisedAddress` would be used for `zoneIngress`, which is not the case.
+
+##### Pros and Cons
+
+* Bad, because `name` looks awkward—it needs to be unique across `networking.zoneIngress` and `networking.zoneEgress` (and potentially `networking.inbound` if we allow mixing)
```
slonka, line 54: This con references "mixing" (`potentially networking.inbound if we allow mixing`) but the "Mixing inbound and zone proxy listeners" section doesn't appear until line 203, after the entire K8s section.

**Suggestion:** Move "Mixing inbound and zone proxy listeners" (lines 203-209) up, right after the schema "Decision" (line 88), before "Marking Pods and Services". The schema decision and mixing are both about the Dataplane spec shape. The K8s section is an implementation detail that should come after all spec-level decisions.

Proposed order:
1. New Schema (Options 1 & 2 + Decision) — keep as-is
2. Mixing inbound and zone proxy listeners — move here
3. Policy Targeting — move here
4. Marking Pods and Services as zone proxies
5. Syncing Zone Ingress Addresses via MeshService

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,279 @@
...
+          path: /tmp/zone-ingress-access.log
+```
+
+#### Labels on Zone Ingress/Egress Dataplanes
+
+Currently, the `kuma.io/proxy-type: sidecar | gateway` label is automatically set on the DPP.
+The only purpose of this label is to let users select proxy types via `spec.targetRef.labels`.
+
+However, in v3 we are removing `gateway`.
+Listeners in the new `listeners` array have a `name` field and can be selected with `spec.targetRef.sectionName`.
```
slonka, line 340: This repeats what was already explained at line 244: `targetRef.sectionName` selects a specific listener.

**Suggestion:** Remove the repeated sentence. This subsection can simply say the label is no longer needed and should be removed in v3, with the link to #15567.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,279 @@
...
+      address: 10.0.0.2 # required
+      port: 10002 # required
+      name: ze-port # optional
+```
+
+Note: There is also a `networking.advertisedAddress` field (but no `networking.advertisedPort`).
+It was a niche Universal-only field [contributed by the community](https://github.com/kumahq/kuma/pull/2116) that should be removed in v3.
+Otherwise, it may appear that `advertisedAddress` would be used for `zoneIngress`, which is not the case.
+
+##### Pros and Cons
```
slonka, line 52: Option 1 has no "Good" points — only one "Bad". This makes the comparison lopsided.

**Suggestion:** Add at least one pro or just mention that this was considered.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,279 @@
...
+      - type: ZoneEgress
+        address: 10.0.0.2
+        port: 10002
+        name: ze-port
+```
+
+##### Pros and Cons
+
+* Good, because `name` semantics are clear—it is obviously unique within the `listeners` array
+* Good, because in the future if we want a cleaner DPP spec, we can deprecate `networking.inbound` and add listeners with `type: Inbound`
```
slonka, line 149: These two pros are both speculative future benefits:
- "in the future if we want a cleaner DPP spec..."
- "we can organically add more listener types in the future..."

**Suggestion:** Merge into one: "Good, because it's extensible — new listener types can be added without schema changes (e.g., moving `inbound` into the array later)."

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,279 @@
...
+      address: 10.0.0.1 # required, address listener binds to
+      port: 10001 # required, port listener binds to
+      name: zi-port # optional, user should be able to set name since `port` can be the same when `addresses` are different
+    zoneEgress:
+      address: 10.0.0.2 # required
+      port: 10002 # required
+      name: ze-port # optional
+```
+
+Note: There is also a `networking.advertisedAddress` field (but no `networking.advertisedPort`).
```
slonka: This note about `advertisedAddress` interrupts the flow between Option 1's YAML example and its pros/cons.

**Suggestion:** Move to a "Notes" section at the bottom, or fold it into Option 2's con on line 84 (since it's more relevant there — Option 2 is where the schema validation concern lives).

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: https://github.com/kumahq/kuma/issues/15629

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: It's important to mention here that we still have a way to select even if `kuma.io/proxy-type` is removed.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: We had solid "good" points when the ZoneIngress and ZoneEgress sections used different schemas. Now that they’ve converged to the same schema, this option no longer brings any real advantages. So yes, it is a bit lopsided now, and that’s fine :)

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: These are different, moving existing inbounds to `listeners` is one thing, introducing new listener type (for delegated gateway) is another

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: Since I've opened an [issue](https://github.com/kumahq/kuma/issues/15629) I believe we don't need this section in the MADR

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,329 @@
...
+### Mixing inbound and zone proxy listeners
+
+Initially, we assumed zone proxy listeners would be incompatible with transparent proxying.
+However, testing proved otherwise.
+A [PoC](https://github.com/kumahq/kuma/pull/15619/changes) demonstrated this by modifying Kuma to add a "fake" zone egress listener,
+showing that zone proxy listeners can coexist with regular inbound listeners without conflicts.
+
+**Decision:** We will not restrict this functionality.
+Since there are no technical limitations preventing zone proxy listeners from running alongside regular inbounds,
+imposing artificial constraints would add unnecessary complexity.
```
lahabana, line 165: And there's no risk with other entities because they are all uniquely named right?

For example If I have a service Foo will add a XDS Cluster for Foo. 
If I'm also a ZoneIngress I guess this XDS Cluster Foo will exist as well but for the zone ingress, is there any risk of collision?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,329 @@
...
+
+### Marking Pods and Services as zone proxies
+
+Dataplane resources are generated by [pod_controller.go](https://github.com/kumahq/kuma/blob/0049139023aae3b102495403b6031282f90b565f/pkg/plugins/runtime/k8s/controllers/pod_controller.go), one per pod in the mesh.
+
+According to [MADR-094](TODO-add-ref) zone proxies are special kinds of deployments in `kuma-system` namespace
+that use the `kuma-dp` image as the main container.
```
lahabana: MADR-094 purposefully doesn't cover that right @slonka ?

lahabana: Also that will mean that doing it this way causes helm upgrade to automatically update the zone proxies which we said was problematic no?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,329 @@
...
+##### Pros and Cons of introducing `kuma.io/zone-proxy: enabled` label
+
+* Good, because less confusing, since the label explicitly indicates that the pod is a zone proxy
+* Good, because pod labels are set as dataplane labels,
+so we're getting dataplane filtering (i.e top-level targetRef) for free.
+But we need to remember to compute this label automatically on Universal.
+
+##### Decision
+
+We chose to introduce a dedicated `kuma.io/zone-proxy: enabled` label to mark Pods as zone proxies.
```
lahabana: should it be egress/ingress/both as the value?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,329 @@
...
+      app: my-app
+  template:
+    metadata:
+      labels:
+        app: my-app
+        kuma.io/zone-proxy: enabled
+    spec:
+      containers:
+        - name: app
+          image: my-app:latest
```
lahabana, line 251: This is unclear how this fits with what you mentioned about having kuma-dp and "main" container

lobkovilya: forgot to update here, image should be `kuma-dp`

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: It's mentioned here https://github.com/kumahq/kuma/blob/c80e0486d30f6ac58b2c54ead635fcadd61d5d84/docs/madr/decisions/094-zone-proxy-deployment-model.md?plain=1#L412

that's why I updated my MADR

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: I don't think we need it. `kuma.io/zone-proxy` is meant to be set manually by user on a deployment's pod template, so we can't automatically keep it consistent with the services that select the pods.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya:
```suggestion
          image: kuma-dp:latest
```

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: Yes, there is no risk with unified naming https://github.com/kumahq/kuma/blob/c80e0486d30f6ac58b2c54ead635fcadd61d5d84/docs/madr/decisions/089-naming-envoy-resources-and-stats-from-multiple-kuma-policies.md?plain=1#L21

Inbound listener and cluster are named like `self_inbound_dp_<port-name>`. I'll update this MADR to mention what naming convention we're going to use for new `listeners`.

Probably `self_zoneingress_dp_<port-name>` and `self_zoneegress_dp_<port-name>`

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,329 @@
...
+Regardless of whether sidecar injection is used,
+users must set the `kuma.io/zone-proxy: enabled` label on the Pod.
+This applies both when mixing inbounds with zone proxy listeners
+and when running a dedicated Deployment with `kuma-dp` as the main container.
+
+If this label is not present, Services labeled with `k8s.kuma.io/zone-proxy-type` will be ignored and no zone proxy listeners will be generated.
+
+#### Open Questions
+
+* What if multiple `k8s.kuma.io/zone-proxy-type: ingress` services select the same DPP?
```
lukidzi, line 180: One more question. What will happen when `kuma.io/sidecar-injection` is enabled and the `kuma.io/zone-proxy: enabled`?

lobkovilya: It's covered by the few sentences above:

> Regardless of whether sidecar injection is used, users must set the `kuma.io/zone-proxy: enabled` label on the Pod. This applies both when mixing inbounds with zone proxy listeners and when running a dedicated Deployment with `kuma-dp` as the main container.

So even if sidecar injection is enabled user still needs to add `kuma.io/zone-proxy` label to enable zone proxy listeners mixed with inbounds.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: Updated the document with the naming convention

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -81,38 +77,99 @@ spec:
...
+
+We can't reuse `kuma.io/egress` and `kuma.io/ingress`
+as they still will be used in 2.14 to generate global-scoped ZoneIngress and ZoneEgress resources.
+
+We could, however, reuse the `kuma.io/sidecar-injection: enabled` annotation.
+In this case it would be treated purely as a signal for the pod controller to reconcile the Pod
+and generate the Dataplane object, even though no actual sidecar injection via the webhook took place.
+
+Alternatively, we can introduce a pod label `kuma.io/zone-proxy: enabled`
+so the pod controller can quickly decide whether a pod needs reconciliation.
```
slonka: I'm leaning towards the `kuma.io/zone-proxy: enabled` label. Reusing `sidecar-injection` for something that isn't a sidecar injection feels semantically misleading  `kuma-dp` is the main container, not a sidecar. That kind of overloading could lead to confusion in docs and debugging down the line.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
slonka: Maybe this should have been stated more clearly that I am not changing anything in this MADR, and if anyone plans to change anything, they need to open up another MADR to address this.

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,341 @@
...
+| Zone Egress   | `self_zoneegress_dp_<port-name>`     | `self_zoneegress_dp_ze-port`  |
+
+### Syncing Zone Ingress Addresses via MeshService
+
+Out of scope for this MADR. See https://github.com/Kong/kong-mesh/issues/9151.
+
+## Implications for Kong Mesh
+
+None
+
```
lobkovilya, line 378: TODO: if DPP has only zone proxy listeners we don't need to generate outbounds (maybe put reachableBackends to an empty array or don't set transparentProxying section)

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
slonka: the decision was made to go with this https://github.com/Kong/kong-mesh/issues/9182

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,297 @@
...
+type: Dataplane
+mesh: default
+name: my-app-xyz123
+labels:
+  app: my-app
+spec:
+  networking:
+    address: 10.0.0.1
+    inbound:
+      - port: 8080
```
lahabana, line 268: For coherency in 3.0 should we just go with `listeners: type: Inbound` ? Or is this too massive of a change?

Can we at least double write maybe?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
```diff
@@ -0,0 +1,297 @@
...
+    listeners:
+      - type: ZoneIngress
+        address: 10.0.0.1
+        port: 10001
```
lahabana, line 272: there can be a name here as well right?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: I checked what's the current gap to add `type: Inbound`, we are missing:

* [state](https://github.com/kumahq/kuma/blob/111e24c8076424c833f5044f37112d49b0d609ed/api/mesh/v1alpha1/dataplane.proto#L154), we actually need this for ZoneIngress/ZoneEgress, I'm going to update the doc.
* [serviceAddress/servicePort](https://github.com/kumahq/kuma/blob/111e24c8076424c833f5044f37112d49b0d609ed/api/mesh/v1alpha1/dataplane.proto#L73-L80), Universal only, doesn't make sense for ZoneIngress/ZoneEgress
* [serviceProbe](https://github.com/kumahq/kuma/blob/111e24c8076424c833f5044f37112d49b0d609ed/api/mesh/v1alpha1/dataplane.proto#L140), Universal only, doesn't make sense for ZoneIngress/ZoneEgress

So we need to make a choice, do we just add these fields as optional, or do we create a proper oneOf discriminator field, for example:
```yaml
listeners:
  - type: Inbound
    inbound: {}
  - type: ZoneIngress
    zoneIngress: {}
```
or maybe we can find a way to place `servicePort` and `serviceAddress` somewhere else.

We can spend time thinking about it now, but I’m not sure it will bring much practical benefit in the end. 

@lukidzi @slonka what do you think?

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lobkovilya: yes, added `name` to the example

docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md
lahabana: Ok thanks for looking. I'll differ to you 3 to take a decision.

lobkovilya:
The PR had three approvals before I added the new section, and we aligned on its content offline, so I'm merging the design.

PR #15571, opened 2026-02-11
feat(meshexternalservice): allow to define priority for endpoints

lukidzi:
## Motivation

All `MeshExternalService` endpoints currently have equal priority - there's no way to express that traffic should prefer one endpoint and failover to another when the primary becomes unhealthy.

## Implementation information

Add an optional priority field to MeshExternalService.Endpoint that maps directly to https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/priority. Lower values = higher priority, 0 is the default.

```yaml
endpoints:
    - address: primary.us-east-2.com
      port: 5432
      priority: 0
    - address: secondary.us-west-2.com
      port: 5432
      priority: 1
```

  Changes:
  - Added Priority `*uint32` field to Endpoint struct
  - Updated `createMeshExternalServiceEndpoint()` to always set `Locality.Priority` from the endpoint priority (defaults to 0)
  - Removed unused zone parameter from `createMeshExternalServiceEndpoint` since MES endpoints don't use zone-based locality
  - Added validator and topology tests covering priority-based failover

## Supporting documentation

Fix #https://github.com/Kong/kong-mesh/issues/9123

### DIFF
--- a/pkg/core/resources/apis/meshexternalservice/api/v1alpha1/testdata/endpoints-with-priority-valid.input.yaml
+++ b/pkg/core/resources/apis/meshexternalservice/api/v1alpha1/testdata/endpoints-with-priority-valid.input.yaml
@@ -11,4 +11,4 @@ endpoints:
     priority: 1
   - address: 3.3.3.3
     port: 443
-    priority: 255
+    priority: 2


pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
```diff
@@ -69,6 +69,12 @@ type Endpoint struct {
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=65535
 	Port int32 `json:"port"`
+	// Priority maps to Envoy's priority levels to enable endpoint failover.
+	// Lower values have higher priority (0 is the default/primary).
+	// When the primary endpoints become unhealthy, traffic fails over to the next priority level.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=0
+	Priority *uint32 `json:"priority,omitempty"`
```
lobkovilya, line 78: Does it work together with FailoverThreshold from MLBS? https://github.com/kumahq/kuma/blob/c777dba9b10faa137b250326e2dbd956b45e7da4/pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/meshloadbalancingstrategy.go#L79

lobkovilya:
In Envoy [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#envoy-v3-api-field-config-endpoint-v3-localitylbendpoints-priority) it says:

> Priorities should range from 0 (highest) to N (lowest) without skipping.

Does it mean Envoy is going to reject the configuration if priorities are `0` and `2` for example? If so then we should probably validate this as well

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
lukidzi: No, we don't support MLBS on the MES or Egress at this moment

lukidzi:
> In Envoy [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#envoy-v3-api-field-config-endpoint-v3-localitylbendpoints-priority) it says:
> 
> > Priorities should range from 0 (highest) to N (lowest) without skipping.
> 
> Does it mean Envoy is going to reject the configuration if priorities are `0` and `2` for example? If so then we should probably validate this as well

It is possible to define like this:

image

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
```diff
@@ -69,6 +69,12 @@ type Endpoint struct {
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=65535
 	Port int32 `json:"port"`
+	// Priority maps to Envoy's priority levels to enable endpoint failover.
+	// Lower values have higher priority (0 is the default/primary).
+	// When the primary endpoints become unhealthy, traffic fails over to the next priority level.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=0
```
slonka: Do we want to add a `+kubebuilder:validation:Maximum` here? Priority values above ~10-20 are meaningless in practice - would something like 255 make sense as an upper bound to prevent misconfiguration?

pkg/xds/topology/outbound.go
```diff
@@ -646,13 +645,16 @@ func createMeshExternalServiceEndpoint(
 		if i == 0 && es.ServerName == "" && govalidator.IsDNSName(endpoint.Address) && tls != nil && tls.Enabled {
 			es.ServerName = endpoint.Address
 		}
+		locality := &core_xds.Locality{
+			Priority: pointer.DerefOr(endpoint.Priority, 0),
+		}
```
slonka: Priority-based failover won't work as-is. `LocalityLbEndpointsMap.append()` in `pkg/xds/envoy/endpoints/v3/endpoints.go:73-98` groups endpoints by `ep.LocalityString()`, which returns `Zone:SubZone` (`pkg/core/xds/types.go:419`). Since all MES endpoints have empty Zone and SubZone, they all get the same key `":"` and collapse into a single `LocalityLbEndpoints` group - only the first endpoint's priority is used.

`LocalityString()` is only used for this grouping key, so including Priority there would be the simplest fix:
```go
func (e Endpoint) LocalityString() string {
    if e.Locality == nil {
        return ""
    }
    return fmt.Sprintf("%s:%s:%d", e.Locality.Zone, e.Locality.SubZone, e.Locality.Priority)
}
```
I'd also add an xDS-level test (e.g. in the egress generator tests or `endpoints_test.go`) that asserts endpoints with different priorities end up in separate `LocalityLbEndpoints` groups.

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
slonka, line 76: Do we want to add a `+kubebuilder:validation:Maximum` here? Priority values above ~10-20 are meaningless in practice - would something like 255 make sense as an upper bound to prevent misconfiguration?

pkg/xds/topology/outbound.go
slonka, line 652: I think this feature doesn't work as-is. The data flow:
1. `outbound.go` correctly sets `Locality.Priority` per endpoint (0, 1, etc.)
2. But `pkg/xds/envoy/endpoints/v3/endpoints.go:74` groups endpoints by `ep.LocalityString()` which returns `Zone:SubZone` (`pkg/core/xds/types.go:419`) - both empty for MES
3. All endpoints get key `":"`, so they're merged into one `LocalityLbEndpoints` group
4. The priority is only read from the first endpoint that creates the group (`endpoints.go:84`)
5. Envoy sees all endpoints at priority 0 - no failover happens

`LocalityString()` is only used for this grouping key, so including Priority there would be the simplest fix:
```go
func (e Endpoint) LocalityString() string {
    if e.Locality == nil {
        return ""
    }
    return fmt.Sprintf("%s:%s:%d", e.Locality.Zone, e.Locality.SubZone, e.Locality.Priority)
}
```
I'd also add an xDS-level test (e.g. in the egress generator tests or `endpoints_test.go`) that asserts endpoints with different priorities end up in separate `LocalityLbEndpoints` groups.

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
lobkovilya: Ah right it's going to be on the egress. We probably need to make sure it works as it's important threshold once you're using priorities

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/testdata/endpoints-with-priority-valid.input.yaml
```diff
@@ -0,0 +1,14 @@
...
+endpoints:
+  - address: 1.1.1.1
+    port: 443
+    priority: 0
+  - address: 2.2.2.2
+    port: 443
+    priority: 1
+  - address: 3.3.3.3
+    port: 443
+    priority: 255
```
slonka: This test uses `priority: 255` but the CRD has `maximum: 128`. The REST path validates against the OpenAPI schema so this would be rejected on both K8s and Universal. The "valid" test case is actually invalid - should be `<= 128`.

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/validator_test.go
```diff
@@ -84,6 +84,10 @@ var _ = Describe("MeshExternalService", func() {
 				name: "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
 				file: "name-length-63",
 			}),
+			Entry("endpoints with valid priority", testCase{
+				name: "external-service",
+				file: "endpoints-with-priority-valid",
```
slonka, line 89: Just FYI: the priority bounds are validated via the OpenAPI schema on the REST path, so both K8s and Universal are covered.

pkg/xds/topology/outbound.go
```diff
@@ -646,13 +645,18 @@ func createMeshExternalServiceEndpoint(
 		if i == 0 && es.ServerName == "" && govalidator.IsDNSName(endpoint.Address) && tls != nil && tls.Enabled {
 			es.ServerName = endpoint.Address
 		}
+		priority := pointer.DerefOr(endpoint.Priority, 0)
```
slonka, line 648: This drops zone-based locality for MES endpoints entirely. Is this intentional?

lukidzi: Yes, as you cannot define zone for MES.

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/testdata/endpoints-with-priority-valid.input.yaml
lukidzi:
```suggestion
    priority: 2
```

Issue #13015, opened 2025-03-05
Potential KDS syncing problem by the resource name length


Icarus9913:
### Kuma Version

master

### Describe the bug

We decide to use [RFC 1035 Label Names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names) for the name validation of our resources like `Mesh`, `MeshService`, `MeshExternalService`, `MeshMultizoneService`, `Zone`. Then we'll have the length restriction for them `at most 63 characters`.

In our KDS system, like you create a `MeshService` resource with length 63, but the name of new syncing MS to zones could be 80 (63 + 17) which breaks the validation.  

### To Reproduce

1. Build a multi-zone environment (1 global, 3 zones)
2. Create a MeshService resource with 63-length name on zone1(k8s).

```text
apiVersion: kuma.io/v1alpha1
kind: MeshService
metadata:
  name: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  namespace: kuma-system
  labels:
    kuma.io/mesh: default
    kuma.io/origin: zone
spec:
  selector:
    dataplaneTags:
      app: redis
  ports:
  - port: 6739
    targetPort: 6739
    appProtocol: tcp
```

3. Then you'll notice the MeshService resource is synced to Global(k8s), zone2(k8s) and zone3(universal) with name `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-dww29994z4bd2292`

<details>

Global:

```text
apiVersion: kuma.io/v1alpha1
kind: MeshService
metadata:
  annotations:
    kuma.io/display-name: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  creationTimestamp: "2025-03-11T06:46:35Z"
  generation: 3
  labels:
    k8s.kuma.io/namespace: kuma-system
    kuma.io/env: kubernetes
    kuma.io/mesh: default
    kuma.io/origin: zone
    kuma.io/zone: kuma-2
  name: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-dww29994z4bd2292
  namespace: kuma-system
  ownerReferences:
[...]

PR #15579, opened 2026-02-13
fix(workload): limit workloads to 63 characters

lahabana:
This enables using labels and use indexes when retrieving workload controllers.

The only danger there is that serviceAccount with more than 64 characters will fail dataplane generation. This is a breaking change but workload is a beta feature so it should be fine.

## Supporting documentation

Fix #XX

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/pod_converter.go
+++ b/pkg/plugins/runtime/k8s/controllers/pod_converter.go
@@ -63,7 +63,7 @@ func (p *PodConverter) PodToDataplane(
 	// we need to validate if the labels have changed
 	workloadName := computeWorkloadName(pod.Labels, p.WorkloadLabels, pod.Spec.ServiceAccountName)
 	if len(workloadName) > 63 {
-		return fmt.Errorf("workload name extracted from serviceAccountName %q is too long", pod.Spec.ServiceAccountName)
+		return fmt.Errorf("workload name %q (derived from serviceAccountName %q) is too long: must be at most %d characters", workloadName, pod.Spec.ServiceAccountName, 63)
 	}
 	labels, err := model.ComputeLabels(
 		core_mesh.DataplaneResourceTypeDescriptor,


lukidzi:
Should we change then `k8s.kuma.io/service-account` that we use a for `spiffeID` to be the label as well? WDYT?

lahabana:
> Should we change then k8s.kuma.io/service-account that we use a for spiffeID to be the label as well? WDYT?

I think not as this is for indentity which is different no? A user can set an alternative label for workload id

lobkovilya:
I don’t mind it. I actually like the idea of having `kuma.io/workload` as a label everywhere, including Kubernetes. I’m just not sure it’s worth it.

Reasons not to merge this PR:

1. It doesn't improve the performance. There is an informer cache backing `.List()` calls from the controller. If I understand the code correctly [the filtering is still happening on the client side](https://github.com/kubernetes-sigs/controller-runtime/blob/80d56e741406b2de8fca0de04deb9cd5569e2a21/pkg/cache/internal/cache_reader.go#L163-L168). In my opinion, the only filtering performance we really need to optimize for is on Global, and that’s backed by Postgres anyway.

2. We previously discussed adding `workloadSelector` to MeshTrafficPermission. That’s only feasible if the workload is tied to the service account. By limiting the workload name to 63 characters, we make things harder for users, since they would need to rename their service accounts to comply.

3. We need to carefully validate that migration to 2.14 works when workload names exceed 63 characters, both on Kubernetes and Universal. That's fine, just extra work.

lahabana:
> 2. We previously discussed adding workloadSelector to MeshTrafficPermission. That’s only feasible if the workload is tied to the service account. By limiting the workload name to 63 characters, we make things harder for users, since they would need to rename their service accounts to comply.

Let's think realistically when would a service account name be more than 63 characters. Most of the time these are the same as deployment names and folks keep their deployment name short anyway no?

> 3. We need to carefully validate that migration to 2.14 works when workload names exceed 63 characters, both on Kubernetes and Universal. That's fine, just extra work.

I don't think we do workload is experimental no?

lobkovilya:
> Let's think realistically when would a service account name be more than 63 characters. Most of the time these are the same as deployment names and folks keep their deployment name short anyway no?

Probably extremely rare, I guess only when people have a system to automatically construct SA names out of other components (i.e. include `zone`, `clusterName` etc). But probably you're right, 63 is a safe assumption.

> I don't think we do workload is experimental no?

AFAIK they're generated by default. So we need to be careful with limits on Global when 2.13 is connected.

lahabana:
@lukidzi @lobkovilya what's your opinion about this in the end?

1. On the one hand I think having this as label probably simplifies some stuff
2. I get that there's a risk attached to the change

lobkovilya:
I just remembered we still have a KDS hashing issue that needs to be resolved before we limit workloads to 63 characters.

Right now the problem only shows up for MeshServices, after we reduced the limit from 253 to 63: https://github.com/kumahq/kuma/issues/13015.

TLDR `addSuffix` func has length limit 253 hardcoded (however it should take it from the type descriptor): https://github.com/kumahq/kuma/blob/6c583b6dae6a5cd87007dfbac4db2ff4c42fa8e1/pkg/kds/hash/hash.go#L16-L21

lahabana:
We should check in our data whether we can do this.

lahabana:
@lobkovilya to dig in for the data when he's back

lobkovilya:
We can proceed, no workload names exceed 63 char.

lobkovilya:
Even though we label this feature as experimental, workload generation is enabled by default. As a result, we risk breaking upgrades to future minor versions for on-prem users who may not even be aware that this feature exists. 

That's why I think we still need:

1. A section in UPGRADE.md
2. Manual upgrade test of this PR. We need to make sure only Workload related functionality is broken (i.e. KDS sync for other resources is not affects, logs are not spamming with errors, etc).
    1. Deploy Global and Zone 2.13 version
    2. Create Workload with the name longer than 63
    3. Upgrade Global to the version in this PR
    4. Upgrade Zone

@lahabana let me know if you're going to work on this or I should do the testing

lahabana:
AS the perf improvement that was expected isn't there and that this is a breaking change let's just close

PR #15620, opened 2026-02-16
chore(renovate): add range strategy to test if golang upgrade works

lukidzi:
## Motivation

Renovate opens PR only with mise upgrade of golang

## Implementation information

Lets see if that helps


lukidzi:
Might works https://github.com/lukidzi/kuma/pull/24/changes

PR #15627, opened 2026-02-17
chore(ci): disable Docker 29 provenance to fix manifest creation

lukidzi:
## Motivation

Docker 29 (rolled out to GHA runners on Feb 9) enables provenance attestations by default, which turns pushed single-platform images into manifest lists, breaking `docker manifest create --amend`

## Implementation information

disable provenance via `DOCKER_BUILD_OPTS=--provenance=false` on all docker build calls


bartsmykla (approved):
lgtm

PR #15644, opened 2026-02-19
docs(MADR): syncing zone ingress addresses across zones

lobkovilya:
## Motivation

https://github.com/Kong/kong-mesh/issues/9151

## Supporting documentation

Fix https://github.com/Kong/kong-mesh/issues/9151

### DIFF
--- a/docs/madr/decisions/096-ingress-address-sync.md
+++ b/docs/madr/decisions/096-ingress-address-sync.md
@@ -53,7 +53,7 @@ var MeshZoneAddressResourceTypeDescriptor = model.ResourceTypeDescriptor{
 }
 ```
 
-We can run a controller on the Zone CP that watches Service resources labeled `k8s.kuma.io/zone-proxy-type: ingress`
+We can run a controller on the Kubernetes Zone CP deployment that watches Service resources labeled `k8s.kuma.io/zone-proxy-type: ingress`
 and creates a corresponding MeshZoneAddress resource.
 
 ```yaml
@@ -75,6 +75,18 @@ we guarantee there are no naming collisions within the Kubernetes cluster.
 
 This means if a user has multiple Kubernetes Services labeled with `k8s.kuma.io/zone-proxy-type: ingress`,
 the CP is going to generate multiple `MeshZoneAddress` for that zone.
+
+`MeshZoneAddress` resource must be generated only when there is at least one **ready** endpoint in the `EndpointSlice`.
+This guarantees:
+
+1. Safe startup. `MeshZoneAddress` is synced to other zones only when at least one zone ingress is up and running.
+
+2. Safe shutdown. The user must follow a graceful shutdown sequence: first scale down zone ingress replicas, then remove the service and deployment.
+`MeshZoneAddress` resource is removed once the `EndpointSlice` has no ready endpoints, 
+leaving a time window for KDS to propagate the changes to other zones.
+
+On Universal, `MeshZoneAddress` resource is managed manually by the user.
+They need to create it providing the LB address and port.
 
 #### Outbound configuration
 


docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,113 @@
...
+Mesh-scoped zone proxies are an opt-in feature in Kuma 2.14.
+Once enabled, both the legacy `ZoneIngress` and the new mesh-scoped zone proxy resources are supported during a transition period to ensure a smooth migration.
+
+When the CP detects both a legacy `ZoneIngress` and a new `MeshZoneAddress` for a given zone,
+it prioritizes `MeshZoneAddress` and shifts all traffic to the new zone proxies.
+
+## Implications for Kong Mesh
+
+None
+
```
lobkovilya, line 235: TODO: what if service type LB has multiple ports? Do we need to validate that?

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,185 @@
...
+- reduce duplication of `advertisedAddress`,
+changing the public zone address should not trigger a mesh wide reconciliation storm
+
+## Design
+
+### Decision
+
+Chosen option: Option 1.
+
+The downside of introducing a new resource is mitigated by the fact that creation of new resource type is already automated in Kuma.
```
slonka, line 23: The MADR should mention that MeshZoneAddress needs KDS sync flags (similar to MeshService's `SyncedAcrossZonesFlag`) to actually reach other zones.

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,185 @@
...
+We can run a controller on the Zone CP that watches Service resources labeled `k8s.kuma.io/zone-proxy-type: ingress`
+and creates a corresponding MeshZoneAddress resource.
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshZoneAddress
+metadata:
+  name: zone-ingress-service-name # name of the service
+  namespace: kuma-system # namespace of the service
+  labels:
```
slonka, line 65: How will IPv6 addresses be represented in `spec.address`? If it's a bare string field, consumers will need to know whether to wrap it in brackets for `[host]:port` notation. Worth mentioning the expected format or at least acknowledging IPv6 support.

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,185 @@
...
+Follow the existing `ZoneIngress` behavior: always use index 0 for both `status.loadBalancer.ingress` and `service.Spec.Ports`.
+This is simple and consistent with existing behavior.
+
+* Good, because it maintains consistency with existing ZoneIngress behavior
+* Good, because it's simple to implement
+* Bad, because silently ignoring additional addresses/ports may be confusing to users
+
+### Option 2: Validate and reject multiple addresses/ports
+
+The `MeshZoneAddress` controller validates the Service and emits an error (via status condition or event)
```
slonka: Using `spec.externalIPs[0]` as the highest priority is a deliberate departure from the legacy `ZoneIngress` behavior (which uses Pod annotations as highest priority). The rationale makes sense (MeshZoneAddress watches Services, not Pods), but I think this deserves a callout in the "Key differences" section at the bottom alongside the other two differences. It's a meaningful behavioral change for users who rely on the annotation-based override today - they'll need to switch to `externalIPs` on the Service.

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,185 @@
...
+The downside of introducing a new resource is mitigated by the fact that creation of new resource type is already automated in Kuma.
+
+### Option 1: Create new MeshZoneAddress resource
+
+We can run a controller on the Zone CP that watches Service resources labeled `k8s.kuma.io/zone-proxy-type: ingress`
+and creates a corresponding MeshZoneAddress resource.
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshZoneAddress
```
slonka: The example shows `kuma.io/mesh: default` on MeshZoneAddress, making it mesh-scoped. But zone ingresses today serve traffic for all meshes in a zone (a single ZoneIngress handles cross-zone routing regardless of mesh).

With the new mesh-scoped zone proxies, does each mesh get its own zone ingress? If so, does that mean each mesh gets its own `MeshZoneAddress` resources pointing to potentially different Services? Or do all meshes share the same zone ingress Service and we'd have duplicate `MeshZoneAddress` resources per mesh pointing to the same address?

Clarifying this would help reviewers understand the expected cardinality of these resources in a multi-mesh setup.

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,185 @@
...
+Mesh-scoped zone proxies are an opt-in feature in Kuma 2.14.
+Once enabled, both the legacy `ZoneIngress` and the new mesh-scoped zone proxy resources are supported during a transition period to ensure a smooth migration.
+
+When the CP detects both a legacy `ZoneIngress` and a new `MeshZoneAddress` for a given zone,
+it prioritizes `MeshZoneAddress` and shifts all traffic to the new zone proxies.
+
+## Implications for Kong Mesh
+
+None
+
```
slonka, line 235: What's the rollback story? If a user enables mesh-scoped zone proxies and something breaks, can they disable it and fall back to ZoneIngress cleanly?

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,185 @@
...
+The `MeshZoneAddress` controller derives address and port from the Service using the following algorithm:
+
+1. **`spec.externalIPs[0]`** (highest priority): If the Service has `externalIPs` defined, use the first one.
+   Port is taken from `service.Spec.Ports[0].Port`.
+
+2. **LoadBalancer**: If Service type is LoadBalancer, use `status.loadBalancer.ingress[0]`.
+   If both `ip` and `hostname` are present, `hostname` takes precedence (more stable, e.g., AWS ELB IPs can change).
+   If the LoadBalancer is not yet ready (no ingress entries), do not generate `MeshZoneAddress` yet.
+   Port is taken from `service.Spec.Ports[0].Port`.
```
slonka, line 196: The legacy `ZoneIngress` uses Pod annotations (`kuma.io/ingress-public-address/port`) as the highest priority override. Here, `externalIPs[0]` on the Kubernetes Service takes that role. This makes sense since MeshZoneAddress watches Services not Pods, but it's a user-facing migration requirement - users relying on Pod annotation overrides today will need to switch to setting `externalIPs` on the Service. I'd add this to the "Key differences" section at the bottom alongside the other two.

docs/madr/decisions/096-ingress-address-sync.md
lukidzi: I think there is no really difference between IPv6 and IPv4, address is a string. We can mention that it's a string field and supports IPv4 and IPv6

docs/madr/decisions/096-ingress-address-sync.md
lobkovilya: But it was already added in the initial version of the MADR, see my commit https://github.com/kumahq/kuma/pull/15644/changes/b703e81b8f78891eb4c1e39ddaa420207dc5b7d9

lobkovilya: If it's about this
> **Migration requirement**: Users currently relying on Pod annotation overrides must switch to setting `spec.externalIPs` on the Kubernetes Service when adopting `MeshZoneAddress`.

Then it's kind of obvious no? Entire Deployment is going to be different (i.e. it's going to have `kuma-sidecar` and some sleep main container) so there is no situation when user would keep using `kuma.io/ingress-public-address`.

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,223 @@
...
+	HasRulesTargetRef:            false,
+	HasStatus:                    false,
+	AllowedOnSystemNamespaceOnly: false,
+	IsReferenceableInTo:          false,
+	ShortName:                    "mza",
+	IsFromAsRules:                false,
+}
+```
+
+We can run a controller on the Zone CP that watches Service resources labeled `k8s.kuma.io/zone-proxy-type: ingress`
```
lahabana: Might be worth doing this in 2 part:

- Universal: resource is managed by user, they need to create it as they manage the LB in front of the proxies
- Kubernetes: a new controller (with more or less what you say here).

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,223 @@
...
+spec:
+  address: 192.168.0.1 # string field, can support either IPv4 or IPv6
+  port: 10001
+```
+
+By generating `MeshZoneAddress` in a one-to-one relationship with the Kubernetes Service,
+we guarantee there are no naming collisions within the Kubernetes cluster.
+
+This means if a user has multiple Kubernetes Services labeled with `k8s.kuma.io/zone-proxy-type: ingress`,
+the CP is going to generate multiple `MeshZoneAddress` for that zone.
```
lahabana, line 77: We'll need to be especially careful to delete things properly :)

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,223 @@
...
+
+We can run a controller on the Zone CP that watches Service resources labeled `k8s.kuma.io/zone-proxy-type: ingress`
+and creates a corresponding MeshZoneAddress resource.
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshZoneAddress
+metadata:
+  name: zone-ingress-service-name # name of the service
+  namespace: kuma-system # namespace of the service
```
lahabana, line 64: IIRC there's nothing that forces the ingress/egress to be in the system namespace but @slonka can confirm.
It might be worth making this explicit just to make sure we don't make mistakes in the implem

docs/madr/decisions/096-ingress-address-sync.md
```diff
@@ -0,0 +1,223 @@
...
+  `kuma.io/ingress-public-port`) for overrides. `MeshZoneAddress` uses the native Kubernetes `spec.externalIPs`
+  field on the Service instead, since it watches Services directly.
+  **Migration requirement**: Users currently relying on Pod annotation overrides must switch to setting
+  `spec.externalIPs` on the Kubernetes Service when adopting `MeshZoneAddress`.
+- **Hostname over IP**: `ZoneIngress` prefers `ip` over `hostname` for LoadBalancer ingress entries.
+  `MeshZoneAddress` prefers `hostname` over `ip` for better stability (e.g., AWS ELB IPs can change).
+
+## Migration
+
+Mesh-scoped zone proxies are an opt-in feature in Kuma 2.14.
```
lahabana, line 218: For guaranteed service transition can we ensure we don't create MeshZoneAddress until there's at least 1 endpoint in the endpointSlice of the backing service? This will prevent new ZoneProxy deployment that fail to start which break cross zone traffic.

docs/madr/decisions/096-ingress-address-sync.md
slonka: Can you elaborate on what scenario you're worried about? Standard controller reconciliation should handle this, right?

docs/madr/decisions/096-ingress-address-sync.md
slonka: Confirmed.

docs/madr/decisions/096-ingress-address-sync.md
slonka: Yes, this is feasible on Kubernetes - the controller can watch `EndpointSlices` and only create `MeshZoneAddress` when there's at least 1 ready endpoint, right?

lahabana: That's my guess as well. If we go this way we should write it in this MADR

docs/madr/decisions/096-ingress-address-sync.md
lahabana: Maybe, my memory is blurry here :) 
Btw should we think a little more about the flow for no traffic loss replacement of service? Does a deletion of the service ensure no traffic will be lost if there's another service?

docs/madr/decisions/096-ingress-address-sync.md
lobkovilya: MZA is in 1-to-1 relationship with Service, so the MZA namespace is the name of the Service

docs/madr/decisions/096-ingress-address-sync.md
lobkovilya: If we watch EndpointSlice and require at least one ready endpoint, we create a path for graceful shutdown. The user should first scale the ingress replicas down to zero, and only then remove the Service.

I have updated the document accordingly.

docs/madr/decisions/096-ingress-address-sync.md
lobkovilya: Yes, we need to watch EndpointSlice, I updated the doc.

docs/madr/decisions/096-ingress-address-sync.md
lobkovilya: right, updated the doc

PR #15677, opened 2026-02-23
docs(MADR): add MeshOpenTelemetryBackend

bartsmykla:
## Motivation

Three observability policies (MeshMetric, MeshTrace, MeshAccessLog) each define OTel collector endpoints independently. In practice most deployments point all three at the same collector, so the endpoint string is duplicated 3x per mesh. When the collector moves, you update three places. Connection settings like TLS and auth would need to be added to all three policies separately.

## Implementation information

This PR adds MADR 095 proposing a new mesh-scoped resource `MeshTelemetryBackend` that holds the shared OTel collector endpoint config. Policies reference it via `backendRef` instead of inlining the endpoint.

The resource uses the standard `type` discriminator pattern (same as MeshMetric/MeshTrace backends). Only `type: OpenTelemetry` is supported. Each policy's OTel backend struct gets a new optional `backendRef` field, mutually exclusive with the existing inline `endpoint`.

### DIFF
--- a/docs/madr/decisions/095-mesh-opentelemetry-backend.md
+++ b/docs/madr/decisions/095-mesh-opentelemetry-backend.md
@@ -62,7 +62,7 @@ spec:
 
 This default is still explicit. Kuma only uses it when the operator creates a MeshOpenTelemetryBackend and points a policy at it. There is no hidden mesh-wide collector. If the operator already has a stable Service or DNS name, they should set `endpoint.address` and skip the default path. The empty backend exists for the node-local collector case where a second backend shape would add complexity without adding control.
 
-No `type` discriminator - the resource name itself is the type. If other telemetry backend types are needed later (Zipkin, Datadog), they become separate resources (`MeshZipkinBackend`, etc.). In this design, `backendRef.kind` is required and must be `MeshOpenTelemetryBackend`, `backendRef.name` is required, and this MADR does not define defaulting for `kind`. The admission webhook enforces that contract and rejects any `backendRef` where `kind` does not match the enclosing OTel backend block.
+No `type` discriminator - the resource name itself is the type. If other telemetry backend types are needed later (Zipkin, Datadog), they become separate resources (`MeshZipkinBackend`, etc.). In this design, `backendRef.kind` is required and must be `MeshOpenTelemetryBackend`, and exactly one of `backendRef.name` or `backendRef.labels` must be set. This MADR does not define defaulting for `kind`. The admission webhook enforces that contract and rejects any `backendRef` where `kind` does not match the enclosing OTel backend block.
 
 If `endpoint` is present, each field is still optional. `address` uses the same runtime defaulting rule. The control plane defaults `port` to `4317` and `path` to empty. This covers DaemonSet plus `hostPort` deployments, empty backends, and partially specified backends without adding a second field with different rules.
 
@@ -80,7 +80,7 @@ When `protocol: http`, the `path` field is a base path prefix. The CP appends th
 
 #### Policy reference
 
-Each policy's OTel backend gets an optional `backendRef` field. When set, the endpoint comes from the referenced MeshOpenTelemetryBackend. The `kind` + `name` pair is just a typed reference to that mesh-scoped resource, and KDS syncs the resource itself to zones. Signal-specific fields remain inline. In 3.0, `backendRef` becomes required and the inline `endpoint` is removed.
+Each policy's OTel backend gets an optional `backendRef` field. When set, the endpoint comes from the referenced MeshOpenTelemetryBackend. The reference works the same way as `BackendRef` in MeshHTTPRoute: use `name` to reference a resource in the same cluster, use `labels` to reference a resource synced from another cluster. KDS syncs the resource itself to zones. Signal-specific fields remain inline. In 3.0, `backendRef` becomes required and the inline `endpoint` is removed.
 
 ##### MeshMetric
 
@@ -164,24 +164,36 @@ None of the current policy OTel backend structs have a `backendRef` field. Today
 
 Each struct needs a new optional `backendRef` field. The inline `endpoint` remains supported but is deprecated starting in 2.14 and will be removed in 3.0. Validation enforces mutual exclusivity: either `endpoint` or `backendRef`, not both. When inline `endpoint` is used, the implementation should also emit a deprecation warning so operators see the migration path before 3.0.
 
-Use `common_api.BackendResourceRef`, not `TargetRef`. This is a typed resource reference (`kind` + `name`), not policy matching. `backendRef.name` is the operator-facing resource name (what the operator typed when creating the resource). KDS syncs MeshOpenTelemetryBackend to zones, but the synced resource name gets a hash suffix (same as all mesh-scoped resources without `SkipKDSHash`). The `kuma.io/display-name` label preserves the original name through sync, so the CP resolves `backendRef.name` by matching against `kuma.io/display-name`, falling back to direct name match for locally-created resources. This follows the same pattern as `GetDisplayName()` used throughout the codebase for KDS-synced resources.
+Use `common_api.BackendResourceRef`, not `TargetRef`. This is a typed resource reference, not policy matching. `BackendResourceRef` has three fields: `kind`, `name`, and `labels`. Either `name` or `labels` must be set, not both. `backendRef.name` always means `metadata.name` - no implicit display-name conversion. For cross-zone references where KDS adds a hash suffix to the synced name, use `backendRef.labels` instead (e.g., `kuma.io/display-name: my-collector`). This follows the same pattern as `BackendRef` in MeshHTTPRoute: `name` for same-cluster, `labels` for cross-zone.
+
+When `labels` is set and multiple resources match, the oldest by creation time wins - same strategy as `DestinationIndex.resolveResourceIdentifier` uses for MeshService.
 
 ```go
 // In each policy's OTel backend struct
 type OpenTelemetryBackend struct {
     // Inline endpoint (existing, deprecated in 2.14, removed in 3.0)
     Endpoint string `json:"endpoint,omitempty"`
-    // Reference to shared backend (kind + name). Weight/port don't apply here.
+    // Reference to shared backend. Weight/port don't apply here.
     BackendRef *common_api.BackendResourceRef `json:"backendRef,omitempty"`
     // ... signal-specific fields unchanged
 }
+```
+
+```go
+type BackendResourceRef struct {
+    Kind   BackendResourceKind `json:"kind"`
+    Name   string              `json:"name,omitempty"`
+    Labels map[string]string   `json:"labels,omitempty"`
+}
+// Validation: exactly one of Name or Labels must be set.
 ```
 
 #### Resolution during xDS generation
 
 1. CP loads all MeshOpenTelemetryBackend resources into MeshContext (same as MeshService, MeshExternalService).
 2. During policy plugin's `Apply()`:
-   - If `backendRef` is set: resolve by matching `backendRef.name` against the `kuma.io/display-name` label of all MeshOpenTelemetryBackend resources in `ctx.Mesh.Resources`. A direct name match (locally-created resource) takes priority over a display-name match (KDS-synced resource). If multiple resources share the same display name, the oldest by creation time wins - same strategy as `DestinationIndex.resolveResourceIdentifier` uses for MeshService. This means a zone-local resource with the same name as a Global-synced one will take priority because the direct name match short-circuits. Operators who want the Global resource to apply uniformly should avoid creating zone-local resources with the same name.
+   - If `backendRef` is set with `name`: look up the MeshOpenTelemetryBackend by `metadata.name` from `ctx.Mesh.Resources`. This works for locally-created resources and same-cluster references.
+   - If `backendRef` is set with `labels`: match labels against all MeshOpenTelemetryBackend resources. If multiple match, the oldest by creation time wins (same as `DestinationIndex.resolveResourceIdentifier`). This is the only way to reference resources synced from another cluster, because KDS adds a hash suffix to `metadata.name`.
    - If `spec.endpoint.address` is set: use it.
    - If `spec.endpoint.address` is omitted: leave the address empty in the `/otel` plan and let `kuma-dp` resolve the node-local default address at runtime.
    - If `spec.endpoint.port` is omitted: default it to `4317`.
@@ -333,6 +345,7 @@ spec:
     port: 4317
   protocol: grpc
 # Policies also created on Global CP, synced to all zones.
+# Since both MOTB and policies originate on Global, backendRef.name works.
 # Each zone's DNS resolves the address to its local collector.
 ```
 
@@ -371,7 +384,7 @@ Backends can also be created directly on a zone CP instead of the Global CP. KDS
 
 This MADR does not introduce a new zone-local resource model. It relies on the existing KDS visibility rules for mesh-scoped resources. When zones need different collectors, use distinct backend names plus zone-targeted policies rather than relying on implicit zone-local lookup or same-name overrides.
 
-Name collision across zones: if Global has a backend named `my-collector` and a zone also has a locally-created backend named `my-collector`, the zone CP sees both - the local one under its original name and the synced one under a hashed name with `kuma.io/display-name: my-collector`. Resolution prefers the direct name match (local resource), so the local backend silently wins. This matches how other KDS-synced resources behave but can be surprising. Operators should use distinct names to avoid this. A future enhancement could surface a status warning when a local resource shadows a synced one.
+Cross-zone references: when a policy on a zone CP needs to reference a MeshOpenTelemetryBackend synced from Global, the operator must use `backendRef.labels` because KDS adds a hash suffix to `metadata.name`. For example, `labels: { "kuma.io/display-name": "main-collector" }` matches the synced resource regardless of the hashed name. This is the same pattern as referencing a synced MeshService from a MeshHTTPRoute. `backendRef.name` only works for resources created in the same cluster.
 
 When zones need different collector addresses (separate cloud regions, different infrastructure), use zone-specific backend names:
 
@@ -536,7 +549,7 @@ Accept the endpoint duplication. Each policy manages its own OTel backend config
 ### Option A
 
 - Dangling references: If a MeshOpenTelemetryBackend is deleted while policies reference it, policies lose their backend config. Unlike routing where a dropped backend is one of many weighted destinations, a dropped telemetry backend means the signal is entirely lost. The CP logs at Info level when a referenced backend is not found during xDS generation. The referencing policies' status conditions surface the unresolved backendRef so operators can detect missing backends (user story 5) rather than silently losing telemetry. The MeshOpenTelemetryBackend's `HasStatus: true` lets the CP track references while the resource exists. An empty backend is still a valid resolved backend because the defaults are intentional. No cross-reference validation webhooks - Kuma's existing pattern is to not block deletion of referenced resources.
-- Resource sync: MeshOpenTelemetryBackend syncs via KDS. If sync is delayed, newly created policies in a zone may not find their backend. Same behavior as MeshExternalService references today. KDS applies a hash suffix to the resource name during sync (`HashSuffixMapper`), so the synced name differs from the original. The `kuma.io/display-name` label preserves the original name, and the resolution logic matches against it. This is the standard mechanism for all KDS-synced mesh-scoped resources.
+- Resource sync: MeshOpenTelemetryBackend syncs via KDS. If sync is delayed, newly created policies in a zone may not find their backend. Same behavior as MeshExternalService references today. KDS applies a hash suffix to `metadata.name` during sync (`HashSuffixMapper`), so `backendRef.name` cannot reference a synced resource directly. Use `backendRef.labels` for cross-zone references (e.g., `kuma.io/display-name`). This is the same pattern as MeshHTTPRoute referencing a synced MeshService.
 
 ### Option B
 
@@ -569,7 +582,7 @@ The initial implementation should cover:
 
 1. MeshOpenTelemetryBackend resource with optional `endpoint` (`address`, `port`, `path`), `protocol` (`grpc`/`http`), and `HasStatus: true`
 2. `backendRef` field added to all three policy OTel backends
-3. Validation: exactly one of `endpoint` or `backendRef` on each policy backend (mutual exclusivity enforced at admission); non-empty `path` rejected when `protocol: grpc`; `endpoint.port` must be in range when set; omitted endpoint fields use defaults
+3. Validation: exactly one of `endpoint` or `backendRef` on each policy backend (mutual exclusivity enforced at admission); within `backendRef`, exactly one of `name` or `labels` must be set; non-empty `path` rejected when `protocol: grpc`; `endpoint.port` must be in range when set; omitted endpoint fields use defaults
 4. Resolution in each policy's `Apply()`: use `endpoint.address` when it is set. If `endpoint.address` is omitted, leave the address empty in the `/otel` plan and let `kuma-dp` resolve the node-local default address at runtime. If `endpoint.port` is omitted, default it to `4317`. If `endpoint.path` is omitted, use an empty path.
 5. On Kubernetes, the sidecar injector always injects `HOST_IP` into every `kuma-dp` container, unconditionally, using the Downward API `status.hostIP`. `kuma-dp` uses it locally for default node-local backend resolution. On other runtimes, the same default flow falls back to `127.0.0.1`. There is no need to send `HOST_IP` through bootstrap metadata, and this default does not depend on OpenTelemetry environment variables.
 6. A status updater component sets `Referenced` or `NotReferenced` on MeshOpenTelemetryBackend resources and sets resolved or unresolved `backendRef` conditions on MeshMetric, MeshTrace, and MeshAccessLog (user story 5). An empty MeshOpenTelemetryBackend is still a valid resolved backend because the defaults are intentional. There should be no warning just because `endpoint.address` is omitted. To debug that path, the operator can see it directly in the resource shape and in the generated `/otel` plan, where the address stays empty while port and path are already defaulted.


docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,410 @@
+# Shared telemetry backend resource for observability policies
+
+* Status: accepted
+
+Technical Story: https://github.com/Kong/kong-mesh/issues/9163
+
+## Context and problem statement
```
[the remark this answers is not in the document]
bartsmykla: Intentional - we use sentence case headings across the project.

docs/madr/decisions/095-mesh-telemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+#### Resolution during xDS generation
+
+1. CP loads all MeshTelemetryBackend resources into MeshContext (same as MeshService, MeshExternalService)
+2. During policy plugin's `Apply()`:
+   - If `backendRef` is set: look up MeshTelemetryBackend by name from `ctx.Mesh.Resources.MeshLocalResources`
+   - Extract endpoint from the type-specific config (e.g., `spec.openTelemetry.endpoint`)
+   - Convert to the same `*core_xds.Endpoint` struct used today
+3. Proceed with existing cluster/listener creation - no changes downstream
+
+This follows the same pattern as BackendRef resolution in routing policies (`pkg/plugins/policies/core/rules/resolve/backendref.go`).
```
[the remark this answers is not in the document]
bartsmykla: Fixed in 73b1f0fb70 - changed to targetref.go since the field type is *common_api.TargetRef.

docs/madr/decisions/095-mesh-telemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+
+The first implementation should be minimal:
+
+1. MeshTelemetryBackend resource with `type: OpenTelemetry` and `endpoint` (address + port) only
+2. `backendRef` field added to all three policy OTel backends
+3. Validation: `endpoint` XOR `backendRef` (mutual exclusivity)
+4. Resolution in each policy's Apply()
+5. Inline `endpoint` remains fully supported
+
+TLS, auth, and protocol fields are follow-up work.
```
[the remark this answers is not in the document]
bartsmykla: Fixed in 73b1f0fb70 - MVP scope now says address + port + path to match the YAML.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+
+These would need to be added to all three policies individually, tripling the work and the surface area.
+
+### User stories
+
+- As a mesh operator, I want to deploy one OTel collector (DaemonSet or Deployment) and point MeshMetric, MeshTrace, and MeshAccessLog at it without duplicating the endpoint in three places.
+- As a mesh operator, I want to update my collector address in one place when the observability team moves it to a different namespace or changes the port.
+- As a mesh operator using a managed OTel backend (Grafana Cloud, Datadog, etc.), I want to configure a bearer token for collector auth.
+- As a mesh operator running multi-zone, I want each zone to have its own collector config without duplicating endpoints across zone-scoped policies.
+- As a mesh operator, I want to roll out signals incrementally - start with metrics, verify it works, then add tracing and access logging against the same collector.
```
lahabana: I think we're missing something like: "As a mesh operator, I want to have Otel configured directly from std otel env vars injected in my pods"

docs/madr/decisions/095-mesh-telemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+
+#### Required policy changes
+
+None of the current policy OTel backend structs have a `backendRef` field. Today they only have an inline `endpoint` string:
+
+- MeshMetric `OpenTelemetryBackend`: `endpoint` + `refreshInterval`
+- MeshTrace `OpenTelemetryBackend`: `endpoint`
+- MeshAccessLog `OtelBackend`: `endpoint` + `attributes` + `body`
+
+Each struct needs a new optional `backendRef` field. The inline `endpoint` remains supported. Validation enforces mutual exclusivity: either `endpoint` or `backendRef`, not both.
```
lahabana: Do we deprecate `endpoint` ?

docs/madr/decisions/095-mesh-telemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+- As a mesh operator, I want to update my collector address in one place when the observability team moves it to a different namespace or changes the port.
+- As a mesh operator using a managed OTel backend (Grafana Cloud, Datadog, etc.), I want to configure a bearer token for collector auth.
+- As a mesh operator running multi-zone, I want each zone to have its own collector config without duplicating endpoints across zone-scoped policies.
+- As a mesh operator, I want to roll out signals incrementally - start with metrics, verify it works, then add tracing and access logging against the same collector.
+
+## Design
+
+### Option A: New shared telemetry backend resource (recommended)
+
+Introduce a new mesh-scoped resource `MeshTelemetryBackend` that defines an OTel collector endpoint with connection settings. Observability policies reference it via a `backendRef` field.
```
lahabana: Any reason to not name this `MeshOpenTelemetryBackend` and then not have a type? After all this can only be referenced in the otel part of the policy...

docs/madr/decisions/095-mesh-telemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+metadata:
+  name: main-collector
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  # Only OpenTelemetry is supported.
+  type: OpenTelemetry
+  openTelemetry:
+    # Collector endpoint - supports gRPC and HTTP/HTTPS
```
lahabana: How do I select gRPC or HTTP/HTTPs?

docs/madr/decisions/095-mesh-telemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+    endpoint:
+      # Address of the collector (hostname or IP)
+      address: otel-collector.observability
+      # Port number
+      port: 4317
+      # Path for HTTP/HTTPS endpoints (e.g., /v1/traces, /v1/metrics, /v1/logs).
```
lahabana: It's not clear to me how I can specify each path when there's a single string

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+          value: "%KUMA_MESH%"
+```
+
+#### Required policy changes
+
+None of the current policy OTel backend structs have a `backendRef` field. Today they only have an inline `endpoint` string:
+
+- MeshMetric `OpenTelemetryBackend`: `endpoint` + `refreshInterval`
+- MeshTrace `OpenTelemetryBackend`: `endpoint`
+- MeshAccessLog `OtelBackend`: `endpoint` + `attributes` + `body`
```
lahabana, line 163: What happens with this `body` field btw? It's not super clear if it should be kept of not.

See docs: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-body

Also should we add attributes to `MeshMetric` and `MeshTrace` ? Does it make sense to add attributes in our Backend as well with a rule for merging them?

docs/madr/decisions/095-mesh-telemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+
+### Options C & D
+
+- No new security concerns.
+
+## Reliability implications
+
+### Option A
+
+- Dangling references: If a MeshTelemetryBackend is deleted while policies reference it, policies lose their backend config. Mitigation: validation webhook that prevents deletion of in-use backends, or status field listing referencing policies.
```
lahabana: We shouldn't do cross reference webhooks. It's not what k8s webhooks are for. I know we do it but we shouldn't add more of them.

status and or logs should be sufficient here.

docs/madr/decisions/095-mesh-telemetry-backend.md
```diff
@@ -0,0 +1,410 @@
...
+
+TLS, auth, and protocol fields are follow-up work.
+
+### Type mismatch between resource and policy
+
+Since OpenTelemetry is the only supported type, this MADR does not cover what happens when a MeshTelemetryBackend's type doesn't match the referencing policy's backend type (e.g., a MeshTrace with `type: Zipkin` pointing at a MeshTelemetryBackend with `type: OpenTelemetry`). That's a separate design decision for when additional types are introduced.
+
+### Naming
+
+`MeshTelemetryBackend` was chosen to avoid vendor-locking the resource name to OpenTelemetry. The `type` discriminator makes it clear that current config is OTel-specific (nested under `openTelemetry`), while the resource name stays generic.
```
lahabana: But if we were to support different types, wouldn't we just support different resources.

We a generic MeshTelemetryBackend you won't be able to enforce that the reference and the values match no?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Added the user story + a Notes section on the app-vs-proxy telemetry split. Env var auto-detection is out of scope for MVP but the path forward is documented.

docs/madr/decisions/095-mesh-telemetry-backend.md
bartsmykla: Deprecated in 2.14, removed in 3.0.

docs/madr/decisions/095-mesh-telemetry-backend.md
bartsmykla: Renamed to MeshOpenTelemetryBackend, dropped the type discriminator. backendRef.kind enforces type matching at schema level.

docs/madr/decisions/095-mesh-telemetry-backend.md
bartsmykla: Added explicit protocol field (grpc/http, default grpc). Port-based inference breaks for non-standard setups.

docs/madr/decisions/095-mesh-telemetry-backend.md
bartsmykla: path is now a base prefix - CP appends /v1/traces, /v1/metrics, /v1/logs for HTTP, matching OTEL_EXPORTER_OTLP_ENDPOINT semantics. Ignored for gRPC.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: body stays in MeshAccessLog (OTel LogRecord body with Envoy command operators). Attributes use different Envoy mechanisms per signal, so no shared abstraction works. Shared resource attributes are follow-up work. Added a Notes section on this.

docs/madr/decisions/095-mesh-telemetry-backend.md
bartsmykla: Agreed - rewrote to match existing Kuma pattern. Silently drop unresolvable refs at xDS time, log at Info, no webhooks.

docs/madr/decisions/095-mesh-telemetry-backend.md
bartsmykla: Renamed to MeshOpenTelemetryBackend. Different backend types = separate resources, backendRef.kind enforces type matching.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+
+Each struct needs a new optional `backendRef` field. The inline `endpoint` remains supported but is deprecated starting in 2.14 and will be removed in 3.0. Validation enforces mutual exclusivity: either `endpoint` or `backendRef`, not both.
+
+```go
+// In each policy's OTel backend struct
+type OpenTelemetryBackend struct {
+    // Inline endpoint (existing, backward compatible)
+    Endpoint string `json:"endpoint,omitempty"`
+    // OR reference to shared backend
+    BackendRef *common_api.TargetRef `json:"backendRef,omitempty"`
```
slonka: The Go snippet uses `*common_api.TargetRef` for the `backendRef` field. The existing `BackendRef` type in routing policies (`api/common/v1alpha1/targetref.go`) is a different struct that extends `TargetRef` with `Weight` and `Port` fields.

Using `TargetRef` directly makes sense here since weight/port are irrelevant for a telemetry backend reference - but it's worth being explicit about this in the MADR since readers familiar with the routing `BackendRef` might assume that's what's being used. A one-line note like "This uses `TargetRef` directly (not the routing `BackendRef` struct) since weight/port don't apply" would prevent confusion.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+    # +optional
+    path: ""
+  # Transport protocol. Drives whether Envoy uses GrpcService or
+  # HttpService in the generated xDS config.
+  # Values: grpc, http. Default: grpc.
+  # +optional
+  protocol: grpc
+```
+
+No `type` discriminator - the resource name itself is the type. If other telemetry backend types are needed later (Zipkin, Datadog), they become separate resources (`MeshZipkinBackend`, etc.). The `backendRef.kind` field enforces type matching at the schema level - a policy's OTel backend can only reference a `MeshOpenTelemetryBackend`, not a hypothetical `MeshZipkinBackend`.
```
slonka: What does "at the schema level" mean here exactly? Is this CRD-level validation (kubebuilder enum restricting `kind` values on the field) or runtime validation in the admission webhook? The distinction matters for the implementation - a kubebuilder enum gives you kubectl validation and IDE autocomplete for free, while webhook validation is more flexible but only catches mismatches at apply time.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+
+The `endpoint` is structured (address + port + path) instead of a raw string so we validate each component separately. Today MeshMetric/MeshAccessLog split the endpoint string on `:`, and MeshTrace has an unreleased URL parser for HTTP endpoints that's being removed. A structured format avoids these inconsistencies.
```
slonka: The structured `endpoint` (address + port + path) is a nice improvement over the raw `host:port` strings the existing policies use. Any plans to adopt this structured format for other fields across the codebase that currently take URLs or host:port strings, or is this a one-off for the new resource?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+These would need to be added to all three policies individually, tripling the work and the surface area.
+
+### User stories
+
+- As a mesh operator, I want to deploy one OTel collector (DaemonSet or Deployment) and point MeshMetric, MeshTrace, and MeshAccessLog at it without duplicating the endpoint in three places.
+- As a mesh operator, I want to update my collector address in one place when the observability team moves it to a different namespace or changes the port.
+- As a mesh operator using a managed OTel backend (Grafana Cloud, Datadog, etc.), I want to configure a bearer token for collector auth.
+- As a mesh operator running multi-zone, I want each zone to have its own collector config without duplicating endpoints across zone-scoped policies.
+- As a mesh operator, I want to roll out signals incrementally - start with metrics, verify it works, then add tracing and access logging against the same collector.
+- As a mesh operator, I want the data plane's OTel export configured from standard `OTEL_EXPORTER_OTLP_*` env vars injected into my pods so I don't duplicate collector config between application instrumentation and the mesh.
```
lahabana:
```suggestion
- As a mesh operator, I want the data plane's OTel export configured from standard [`OTEL_EXPORTER_OTLP_*`](https://opentelemetry.io/docs/specs/otel/protocol/exporter/) env vars injected into my pods so I don't duplicate collector config between application instrumentation and the mesh.
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+    # signal-specific suffix (/v1/traces, /v1/metrics, /v1/logs)
+    # automatically, matching OTEL_EXPORTER_OTLP_ENDPOINT semantics.
+    # Ignored for gRPC.
+    # +optional
+    path: ""
+  # Transport protocol. Drives whether Envoy uses GrpcService or
+  # HttpService in the generated xDS config.
+  # Values: grpc, http. Default: grpc.
+  # +optional
+  protocol: grpc
```
lahabana: What about format in http IIRC there's http/json and http/protobuf?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+
+- TLS settings for the collector connection (custom CA, client certs)
+- Authentication headers (bearer tokens for managed OTel backends)
+- Protocol preference (gRPC vs HTTP) - policies are gRPC-only today
+
+These would need to be added to all three policies individually, tripling the work and the surface area.
+
+### User stories
+
+- As a mesh operator, I want to deploy one OTel collector (DaemonSet or Deployment) and point MeshMetric, MeshTrace, and MeshAccessLog at it without duplicating the endpoint in three places.
```
lahabana: you should number your user stories. It's easier to refer to them in the doc.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+
+| | Application telemetry | Proxy telemetry |
+|---|---|---|
+| Config source | OTel env vars on app container | Control plane via xDS |
+| Who exports | OTel SDK in app code | Envoy sidecar |
+| What it captures | App-level spans, custom metrics | L4/L7 request metrics, access logs, mesh traces |
+
+A mesh operator might want both paths pointing at the same collector without configuring each separately. The kuma-dp process already has a `DynamicMetadata` transport (`map[string]string`) that carries key-value pairs from the data plane to the control plane during bootstrap. This transport could carry `OTEL_EXPORTER_OTLP_*` values from the pod environment to the CP, which would then generate matching xDS config.
+
+Env var auto-detection is out of scope for the MVP. MeshOpenTelemetryBackend is the explicit configuration path for proxy telemetry. Env var integration would build on top of it as follow-up work - the resource is still the canonical config for a backend's connection settings, whether set manually or populated from env vars.
```
lahabana: If you put this as out of scope I'd like to see how you intend to fullfill user story 1 and 5

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
+# Shared telemetry backend resource for observability policies
+
+* Status: accepted
+
+Technical Story: https://github.com/Kong/kong-mesh/issues/9163
```
slonka: nit/FYI: This links to a private repo (Kong/kong-mesh)

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+- Authentication headers (bearer tokens for managed OTel backends)
+- Protocol preference (gRPC vs HTTP) - policies are gRPC-only today
+
+These would need to be added to all three policies individually, tripling the work and the surface area.
+
+### User stories
+
+- As a mesh operator, I want to deploy one OTel collector (DaemonSet or Deployment) and point MeshMetric, MeshTrace, and MeshAccessLog at it without duplicating the endpoint in three places.
+- As a mesh operator, I want to update my collector address in one place when the observability team moves it to a different namespace or changes the port.
+- As a mesh operator using a managed OTel backend (Grafana Cloud, Datadog, etc.), I want to configure a bearer token for collector auth.
```
slonka: This story also covers TLS (custom CA, client certs) and protocol switching (gRPC vs HTTP) for the collector connection - both are key motivations for the shared resource mentioned in the context section but not reflected in the user stories.

```suggestion
- As a mesh operator, I want to configure connection settings (TLS, auth, protocol) for my collector in one place rather than across three policies.
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+- As a mesh operator running multi-zone, I want each zone to have its own collector config without duplicating endpoints across zone-scoped policies.
```
slonka: So how does that work? You would want to have one policy and the backend to be different in different zones? Is there gonna be some special ref?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+- As a mesh operator, I want to roll out signals incrementally - start with metrics, verify it works, then add tracing and access logging against the same collector.
```
slonka: This is a bit repetitive of #1/#2.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: What does "at the schema level" mean here exactly? Is this CRD-level validation (kubebuilder enum restricting `kind` values on the field) or runtime validation in the admission webhook?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: The existing `BackendRef` type in routing policies is a different struct that extends `TargetRef` with `Weight` and `Port` fields. Using `TargetRef` directly makes sense since weight/port don't apply - worth making that explicit.

```suggestion
    // Uses TargetRef directly, not the routing BackendRef struct (weight/port don't apply)
    BackendRef *common_api.TargetRef `json:"backendRef,omitempty"`
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+
+This follows the same pattern as TargetRef resolution in policies (`pkg/plugins/policies/core/rules/resolve/targetref.go`).
+
+#### Resource characteristics
+
+| Property        | Value                                                                                |
+|-----------------|--------------------------------------------------------------------------------------|
+| `IsPolicy`      | false (it's a resource, not a policy)                                                |
+| `Scope`         | Mesh                                                                                 |
+| `HasStatus`     | false                                                                                |
```
slonka: Unlike routing where a dropped backend is one of many weighted destinations and traffic still flows, a dropped telemetry backend means the signal is entirely lost with no user-visible feedback (only a CP Info log). Should `HasStatus` be `true` so the resource or referencing policies can surface unresolved backendRefs?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: nit/FYI: This links to a private repo (Kong/kong-mesh)

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+
+The endpoint is the common denominator. Signal-specific fields (`refreshInterval`, `attributes`, `body`) are unique to each policy and don't belong in a shared resource.
+
+Beyond endpoint duplication, there's a missing abstraction for collector connection settings. Today none of the policies support configuring:
+
+- TLS settings for the collector connection (custom CA, client certs)
+- Authentication headers (bearer tokens for managed OTel backends)
+- Protocol preference (gRPC vs HTTP) - policies are gRPC-only today
+
+These would need to be added to all three policies individually, tripling the work and the surface area.
```
slonka, line 27: Are we sure that all policies support these fields? How is it gonna work if one of the policy doesn't support it? Are we gonna have the same schema but different validators? Isn't this going to be problematic for the open API schema?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+
+### User stories
+
+- As a mesh operator, I want to deploy one OTel collector (DaemonSet or Deployment) and point MeshMetric, MeshTrace, and MeshAccessLog at it without duplicating the endpoint in three places.
+- As a mesh operator, I want to update my collector address in one place when the observability team moves it to a different namespace or changes the port.
```
slonka: So if someone wanted to test something, what would be the workflow? Creating a new backend and then changing the URL in that backend and then rolling that out to original backend?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: This story also covers TLS (custom CA, client certs) and protocol switching (gRPC vs HTTP) for the collector connection - both are key motivations for the shared resource mentioned in the context section but not reflected in the user stories.

```suggestion
- As a mesh operator using a managed OTel backend (Grafana Cloud, Datadog, etc.), I want to configure connection settings (TLS, auth, protocol) for my collector in one place rather than across three policies.
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: So how does that work? You would want to have one policy and the backend to be different in different zones? Is there gonna be some special ref?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: This is a bit repetitive of 1 / 2

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: This is explicitly out of scope for MVP (noted in the Notes section) - should it be a user story here?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: What does "at the schema level" mean here exactly? Is this CRD-level validation (kubebuilder enum restricting `kind` values on the field) or runtime validation in the admission webhook?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: The existing `BackendRef` type in routing policies is a different struct that extends `TargetRef` with `Weight` and `Port` fields. Using `TargetRef` directly makes sense since weight/port don't apply - worth making that explicit.

```suggestion
    // Uses TargetRef directly, not the routing BackendRef struct (weight/port don't apply)
    BackendRef *common_api.TargetRef `json:"backendRef,omitempty"`
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: Unlike routing where a dropped backend is one of many weighted destinations and traffic still flows, a dropped telemetry backend means the signal is entirely lost with no user-visible feedback (only a CP Info log). Should `HasStatus` be `true` so the resource or referencing policies can surface unresolved backendRefs?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,436 @@
...
+- Separates infrastructure config (where the collector is) from policy config (what to send)
+
+#### Disadvantages
+
+- New resource type: CRD, REST API, kumactl support, documentation
+- Three policy plugins need modification (add backendRef + resolution logic)
+- Users learn a new concept
+- More indirection (policy -> resource -> endpoint)
+
+### Option B: Named backends on the Mesh resource (rejected)
```
slonka, line 500: Options B and C are quite verbose for rejected alternatives. Consider trimming the YAML examples - the core reasoning is what matters and it can be stated in a few sentences each.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: Missing user story worth adding:

```suggestion
- As a mesh operator, I want the data plane's OTel export configured from standard `OTEL_EXPORTER_OTLP_*` env vars injected into my pods so I don't duplicate collector config between application instrumentation and the mesh.
- As a mesh operator, I want to know when a policy references a backend that no longer exists so I don't silently lose telemetry.
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Removed the Technical Story line entirely.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Yes, all three policies have OTel backends today. Each gets its own backendRef field in its own struct with its own validator. OpenAPI is generated per-policy so no schema conflicts.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Two options: update the one MeshOpenTelemetryBackend and all referencing policies follow, or create a second backend and switch policies to it one by one. The second approach gives you a rollback path.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Accepted, updated story 3 to cover TLS, auth, and protocol.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: MeshOpenTelemetryBackend is mesh-scoped and syncs via KDS like MeshExternalService. Each zone can have its own backend resource. Policies reference by name, same resolution as any other mesh-scoped resource. No special ref needed.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Agreed, merged into story 1.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Moved env var auto-detection into initial scope (item 7 in the scope list). It is user story 6 now.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Reworded to "via the admission webhook". The validator rejects backendRef.kind values that don't match the enclosing backend type.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Accepted, added the comment.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Good point. Changed HasStatus to true and added a user story for it (story 5). Also updated the dangling refs section in Reliability implications.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Trimmed both. Removed YAML examples, kept the core reasoning.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Added as story 5. Also set HasStatus to true to support surfacing unresolved backendRefs.

bartsmykla: Accepted, added the link on story 6.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Added a paragraph after the protocol field description. Envoy's built-in OTel extensions only support protobuf encoding over HTTP - JSON encoding would need a custom filter. So protocol: http means OTLP/HTTP with protobuf. If JSON is needed later, a separate encoding field can be added.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Done, numbered 1-6.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Moved env var auto-detection into initial scope (item 7 in the scope list). Also expanded the section with DynamicMetadata transport details, K8s/Universal paths, priority semantics, and supported env vars.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,700 @@
...
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  endpoint:
+    address: collector.eu-west.internal
+    port: 4317
+  protocol: grpc
+---
+# Zone-targeted policy
```
lahabana, line 407: Can this be just defined in the zone?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,700 @@
...
+          kind: MeshOpenTelemetryBackend
+          name: main-collector    # no longer exists
+        refreshInterval: 30s
+```
+
+CP behavior when the referenced backend is missing:
+
+- Logs at Info level: `MeshOpenTelemetryBackend "main-collector" not found, referenced by MeshMetric "all-metrics"`
+- Skips OTel backend config in xDS for that signal (no telemetry export)
+- Status condition on the resource surfaces the unresolved reference so the operator can detect it via `kubectl` or the REST API
```
lahabana: The status is on the policy ? Because I thought the story is about the missing resource right?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,700 @@
...
+
+Unlike routing, where a dropped backend is one of many weighted destinations, a dropped telemetry backend means the signal is entirely lost. This is why `HasStatus: true` matters here.
+
+##### Story 6: Env var auto-detection
+
+The OTel Operator or a pod annotation sets `OTEL_EXPORTER_OTLP_*` env vars on the sidecar. kuma-dp reads them at startup and sends them to the CP via DynamicMetadata. Policies without an explicit endpoint or backendRef use them as fallback.
```
lahabana: There shouldn't be a need for this. The otel exporter runs on kuma-dp right? So it have everything available already it doesn't need to wire in the CP here.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,700 @@
...
+
+```yaml
+# Pod annotation injects env vars into kuma-sidecar
```
lahabana: How does this really work in the wild.

A quick claude chat returned: https://claude.ai/share/0f4bd1b7-d038-4e42-ba45-a060cb4b7a15

I think if we want to build something useful to many we have to dig and make sure that whatever we build we play nicely with the ecosystem.
To do so we should understand the ecosystem in depth, especially the env var injection possibly

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: I meant the underlying envoy plugins that handle the policy.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: After looking into this, neither Istio nor Linkerd reads `OTEL_EXPORTER_OTLP_*` env vars for proxy telemetry:

- **Istio** configures proxy tracing via `MeshConfig.extensionProviders` + the `Telemetry` CRD. The proxy doesn't read OTel env vars. ([docs](https://istio.io/latest/docs/tasks/observability/distributed-tracing/opentelemetry/))
- **Linkerd** configures it via Helm values (`proxy.tracing.collector.endpoint`). Same - no env var reading. ([docs](https://linkerd.io/2-edge/tasks/distributed-tracing/))

Both treat app SDK telemetry and proxy telemetry as separate concerns, configured through separate mechanisms.

There are also documented ecosystem conflicts when the OTel Operator and a service mesh sidecar injector both mutate the same pod:
- Container ordering - the Operator injects env vars into `Containers[0]`, but sidecar injection can reorder containers, so env vars land on the proxy instead of the app ([open-telemetry/opentelemetry-operator#2158](https://github.com/open-telemetry/opentelemetry-operator/issues/2158))

Given this, should we drop the env var user story? MeshOpenTelemetryBackend already provides an explicit configuration path for proxy telemetry, and auto-detecting from app env vars would conflate two concerns that no other mesh combines.

Re [this comment](https://github.com/kumahq/kuma/pull/15677#discussion_r2846014370) - the user stories that matter (single collector config, incremental rollout) would be fulfilled by MeshOpenTelemetryBackend itself, not by env var auto-detection.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Yes, zone CPs can create backends directly. KDS syncs them via ZoneToGlobalFlag. Updated the example to show both paths.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: The wording was unclear, fixed it. When the MOTB gets deleted, the broken reference shows up on the referencing policy's status (the MOTB itself is gone). MOTB's HasStatus tracks which policies reference it while it still exists.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Yes, after rethinking this I agree. No mesh reads OTEL_* env vars for proxy telemetry, the OTel Operator targets the app container not the sidecar, and 2 out of 3 signals export from Envoy which doesn't read env vars anyway. More problems than usefulness. Dropped story 6 and replaced the env var section with a "considered and dropped" summary that documents the reasoning.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: This is true for MeshMetric - kuma-dp runs the OTel gRPC exporter. But MeshTrace and MeshAccessLog export from Envoy directly (`envoy.tracers.opentelemetry`, `envoy.access_loggers.open_telemetry`), configured through xDS. kuma-dp has no role in those export paths, so env vars on kuma-dp don't help for 2 out of 3 signals. We dropped env var auto-detection entirely - see the "considered and dropped" section.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: We dug into the ecosystem: the OTel spec puts `OTEL_EXPORTER_OTLP_*` env vars under [Language APIs & SDKs](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/) - they're for app instrumentation, not infrastructure proxies. The OTel Operator injects into `Containers[0]` (the app container) and has `container-names` specifically to exclude sidecars (open-telemetry/opentelemetry-operator#2158). No service mesh reads these for proxy telemetry. We dropped env var auto-detection and documented the reasoning in the MADR.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Each policy uses a different export mechanism. MeshMetric doesn't use an Envoy plugin - kuma-dp exports via the OTel gRPC SDK through a Unix socket that Envoy proxies to the collector. MeshTrace uses `envoy.tracers.opentelemetry` which has both `grpc_service` and `http_service`. MeshAccessLog uses `envoy.access_loggers.open_telemetry` - currently gRPC only in our go-control-plane v1.36.0, but v1.37.0 adds `HttpService` and `GrpcService` fields ([envoy#42445](https://github.com/envoyproxy/envoy/pull/42445)). MeshOpenTelemetryBackend's `endpoint` and `protocol` fields work across all three - gRPC today, HTTP after the go-control-plane upgrade for MeshAccessLog.

Automaat (approved):

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: Yes, and do all of these support the adding of authentication header, different protocols, TLS and everything we want to support? Or are we gonna need to cherry peak different properties for different mechanisms?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,501 @@
...
+metadata:
+  name: main-collector
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  endpoint:
+    address: otel-collector.observability
+    port: 4317
+    path: ""        # +optional, base path prefix for HTTP (ignored for gRPC)
```
slonka: shouldn't this be rejected by validation and not ignored?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,501 @@
...
+| `ShortName`     | `motb`                                                                               |
+| `IsDestination` | false                                                                                |
+
+Placed in `pkg/core/resources/apis/meshopentelemetrybackend/` following the same directory structure as MeshExternalService.
+
+#### Advantages
+
+- Purpose-built API for the use case
+- Single place for endpoint + connection settings
+- Extensible: TLS, auth, protocol preferences added once, used by all policies
```
slonka, line 254: this is what I've asked here https://github.com/kumahq/kuma/pull/15677#discussion_r2851939268

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,501 @@
...
+| `HasStatus`     | true (surfaces unresolved backendRef conditions to referencing policies)             |
+| `KDSFlags`      | <code>GlobalToZonesFlag &#124; ZoneToGlobalFlag</code> (same as MeshExternalService) |
+| `ShortName`     | `motb`                                                                               |
+| `IsDestination` | false                                                                                |
+
+Placed in `pkg/core/resources/apis/meshopentelemetrybackend/` following the same directory structure as MeshExternalService.
+
+#### Advantages
+
+- Purpose-built API for the use case
```
slonka, line 252: to me it's too vague

```suggestion
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,501 @@
...
+- Extensible: TLS, auth, protocol preferences added once, used by all policies
+- Follows established resource patterns (tooling, KDS, REST API, kumactl come for free)
+- Multi-zone: syncs via KDS, can be zone-scoped
+- Backward compatible with inline endpoints
+- Separates infrastructure config (where the collector is) from policy config (what to send)
+
+#### Disadvantages
+
+- New resource type: CRD, REST API, kumactl support, documentation
+- Three policy plugins need modification (add backendRef + resolution logic)
```
slonka, line 263: is this really a disadvantage? I don't think we can do without any modification, right? 


```suggestion
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: TLS and protocols are consistent across all three - cluster-level TLS via `ClientSideTLS` already, and all three use the same `GrpcService` / `HttpService` Envoy types.

Auth headers are follow-up work. All three go through Envoy - `GrpcService.initial_metadata` and `HttpService.request_headers_to_add` cover MeshTrace and MeshAccessLog, and MeshMetric routes through an HCM proxy listener where headers can be set at the route level. No cherry-picking from the operator's side, the CP handles the translation.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,501 @@
...
+- Users learn a new concept
```
slonka, line 264: Can we specify this is about the first time we do cross policies references?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,501 @@
...
+
+#### OTel env var auto-detection (considered and dropped)
+
+We considered having kuma-dp read `OTEL_EXPORTER_OTLP_*` env vars and forward them to the CP via the existing DynamicMetadata bootstrap transport, so policies without an explicit endpoint or backendRef could use them as fallback. After investigation, this creates more problems than it solves:
+
+- `OTEL_EXPORTER_OTLP_*` env vars are designed for application OTel SDKs, not infrastructure proxies. The OTel spec places them under "Language APIs & SDKs." The OTel Collector itself doesn't read them for its own export config.
+- No service mesh reads these env vars for proxy telemetry. Istio uses `MeshConfig.extensionProviders` + the `Telemetry` CRD. Linkerd uses Helm values. Both treat app SDK telemetry and proxy telemetry as separate concerns.
+- The OpenTelemetry Operator injects env vars into `Containers[0]` (the app container). Mesh sidecars are appended, so they don't get the vars. The `container-names` annotation exists specifically to exclude sidecars ([open-telemetry/opentelemetry-operator#2158](https://github.com/open-telemetry/opentelemetry-operator/issues/2158)).
```
slonka: The `instrumentation.opentelemetry.io/container-names` annotation is for targeting specific containers for injection, not for excluding sidecars. The linked issue (https://github.com/open-telemetry/opentelemetry-operator/issues/2158) actually shows the annotation didn't work initially with Istio. Consider rewording.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,501 @@
...
+- `MeshObservabilityBackend` - too abstract, doesn't add clarity over the specific protocol name
+- `MeshCollectorBackend` - not all backends are "collectors" (e.g., managed SaaS endpoints)
+
+#### Signal-specific fields stay in policies
+
+MeshAccessLog's `body` field is logs-only (OTel LogRecord body). It holds Envoy access log command operators (`%START_TIME%`, `%UPSTREAM_HOST%`) that control what each log record says. This is per-record content formatting, not connection config - it stays in MeshAccessLog.
+
+MeshAccessLog's `attributes` are also logs-only (mapped to OTel `KeyValueList` with command operator substitution). MeshTrace has `Tags` (mapped to Envoy `CustomTag` on the HCM tracing config). MeshMetric has no OTel attributes (uses kuma-dp `ExtraLabels`). Each signal uses a completely different Envoy mechanism for attributes, so there's no shared abstraction that works across all three.
+
+#### OTel env var auto-detection (considered and dropped)
```
slonka, line 485: Is there a low-hanging fruit for aligning OTel resource attributes (`service.name`, `service.namespace`) between proxy and app telemetry? If both report the same resource attributes to the same collector, correlation comes for free in any OTel-native backend.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: agreed on zoom

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: Agreed that this can be achieved on xds envoy level

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: agreed that too vague to add

```suggestion
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
slonka: agreed to revisit

bartsmykla: Updated in the latest push - removed the https://github.com/open-telemetry/opentelemetry-operator/issues/2158 reference; env vars in Kubernetes are container-scoped, so setting `OTEL_EXPORTER_OTLP_ENDPOINT` on the app container doesn't propagate to kuma-sidecar automatically. Getting it there requires either `container-names: <app>,kuma-sidecar` per pod (which has a webhook ordering race - OTel Operator may run before Kuma injects the sidecar) or explicit `ContainerPatch`, both needed for every workload individually. There's no automatic benefit over creating one MOTB; and even then, in its current form, only MeshMetric would benefit since MeshTrace and MeshAccessLog use Envoy's built-in OTel exporters and still need xDS config from the CP regardless.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Updated - the MADR now says non-empty `path` is rejected by validation when `protocol: grpc`, not silently ignored. Scope item 3 updated to match.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Removed.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Removed.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Not low-hanging. For MeshMetric (kuma-dp OTel SDK), we could inject `service.name` and `service.namespace` as OTel resource attributes using kuma-dp's own `DataplaneMetadata` - that part is straightforward. For MeshTrace and MeshAccessLog (Envoy's built-in exporters), the Envoy OTel exporter config doesn't currently have a place for resource attributes in the xDS API we use.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,532 @@
...
+
+### Scope
+
+The first implementation covers:
+
+1. MeshOpenTelemetryBackend resource with `endpoint` (address + port + path) or `nodeEndpoint` (port + path), `protocol` (grpc/http), and `HasStatus: true`
+2. `backendRef` field added to all three policy OTel backends
+3. Validation: exactly one of `endpoint` or `nodeEndpoint` on the resource; exactly one of `endpoint` or `backendRef` on each policy backend (mutual exclusivity enforced at admission); non-empty `path` rejected when `protocol: grpc`
+4. Resolution in each policy's Apply(): `endpoint` uses address directly; `nodeEndpoint` resolves host IP from `BootstrapDynamicMetadata["HOST_IP"]`, falls back to `127.0.0.1` on Universal
+5. Sidecar injector injects `HOST_IP` env var (Downward API `status.hostIP`) into the kuma-dp container to support `nodeEndpoint`
```
slonka: Is `HOST_IP` always injected into every kuma-dp container, or only when a `nodeEndpoint`-type MeshOpenTelemetryBackend exists in the mesh? Always injecting is simpler, conditional injection requires the injector to watch/cache MOTB resources - can you specify the tradeoff and the chosen approach here?

bartsmykla: always injected, unconditionally - same as INSTANCE_IP and other standard env vars in factory.go. No MOTB watching needed. clarified in latest push

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,546 @@
...
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshOpenTelemetryBackend
+metadata:
+  name: main-collector
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  endpoint:             # +optional, mutually exclusive with nodeEndpoint
```
lahabana: May I suggest defaulting to OTEL env vars if any of the fields are not set. I feel it just makes everything a lot simpler no?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,546 @@
...
+
+The OTLP/HTTP spec defines two content encodings: binary protobuf (`application/x-protobuf`) and JSON (`application/json`). Envoy's built-in OTel extensions (stats sink, access logger, tracer) only support protobuf encoding over HTTP. JSON encoding is not available in Envoy without a custom filter. So `protocol: http` means OTLP/HTTP with protobuf encoding. If JSON encoding support is needed later, a separate `encoding` field can be added to the spec.
+
+When `protocol: http`, the `path` field is a base path prefix. The CP appends the signal-specific suffix (`/v1/traces`, `/v1/metrics`, `/v1/logs`) during xDS generation, matching how the OTel SDK handles `OTEL_EXPORTER_OTLP_ENDPOINT`. An empty path means the standard paths are used directly. For gRPC, paths are irrelevant - gRPC routes by protobuf service name.
+
+#### Policy reference
+
+Each policy's OTel backend gets an optional `backendRef` field. When set, the endpoint comes from the referenced MeshOpenTelemetryBackend. Signal-specific fields remain inline.
```
lahabana: In v3 the backendRef will become required. Might be worth clarifying this

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,546 @@
...
+
+```go
+// In each policy's OTel backend struct
+type OpenTelemetryBackend struct {
+    // Inline endpoint (existing, backward compatible)
+    Endpoint string `json:"endpoint,omitempty"`
+    // OR reference to shared backend.
+    // Uses TargetRef directly, not the routing BackendRef struct
+    // (weight/port don't apply here).
+    BackendRef *common_api.TargetRef `json:"backendRef,omitempty"`
```
lahabana: I don't think it's a `TargetRef` right? It has similarity but it's not the same. What are we gaining from sharing Datastructure?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,546 @@
...
+   - If `spec.nodeEndpoint` is set: resolve node IP from `proxy.Metadata.DynamicMetadata["HOST_IP"]`; fall back to `127.0.0.1` on Universal/VM.
+3. Inline `endpoint` path: CP generates a standard static Envoy cluster pointing directly at the collector. MeshTrace and MeshAccessLog use Envoy's built-in OTel exporters (`envoy.tracers.opentelemetry`, `envoy.access_loggers.open_telemetry`) over gRPC. Unchanged from current behavior.
+4. `backendRef` path - kuma-dp pipe mode: when `backendRef` is set and the proxy advertises `FeatureOtelViaKumaDp`, MeshTrace and MeshAccessLog route through kuma-dp via Unix sockets:
+   ```
+   Envoy --gRPC--> Unix socket --> kuma-dp otelreceiver --> gRPC or HTTP --> OTel Collector
+   ```
+   CP configures a `pipe`-type Envoy cluster per signal (socket paths from `OtelTraceSocketName()` / `OtelLogSocketName()`). kuma-dp runs gRPC receivers on those sockets and forwards to the collector using the protocol and address from the MOTB spec. kuma-dp reads backend config from dynconf routes (`/meshtrace`, `/meshaccesslog`), which return JSON with `socketPath`, `endpoint`, `useHTTP`, and `path`. ETag caching avoids reconnects on unchanged config.
```
lahabana: Does it need different dyn routes? Seems to me like there's a single on that's `otelbackend` no?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,546 @@
...
+5. MeshMetric path: unchanged - kuma-dp reads Envoy stats via its own Unix socket and exports via OTel SDK. Not affected by `FeatureOtelViaKumaDp`.
```
lahabana: It does right? It needs to support not having endpoint for the exporter right? I believe right now the exporter is set with the dynroute of MeshMetric no?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Right, this is not a `TargetRef`. I changed the design to use `common_api.BackendResourceRef`, since this is a typed resource reference, not policy matching.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Clarified this. `backendRef` stays optional in 2.14, then becomes required in 3.0 when the inline `endpoint` is removed.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Clarified this to use a single `/otel` dynconf route for `backendRef` backends. The per-signal route split is no longer part of the design.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Clarified the split here. Inline `endpoint` backends stay on the direct Envoy path, while `backendRef` backends go through the unified `/otel` route. `/meshmetric` stays only for the stats pipeline and `refreshInterval`.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Covered by: https://github.com/kumahq/kuma/pull/15807

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,593 @@
...
+
+The `protocol` field selects gRPC or HTTP transport. Envoy's OTel extensions use separate `grpc_service` and `http_service` fields in their protobuf config - the CP needs to know which one to generate. Port-based inference (4317 = gRPC, 4318 = HTTP) would break for non-standard setups, so an explicit field is safer. Default is `grpc` for backward compatibility with existing Kuma behavior.
+
+The OTLP/HTTP spec defines two content encodings: binary protobuf (`application/x-protobuf`) and JSON (`application/json`). Envoy's built-in OTel extensions (stats sink, access logger, tracer) only support protobuf encoding over HTTP. JSON encoding is not available in Envoy without a custom filter. So `protocol: http` means OTLP/HTTP with protobuf encoding. If JSON encoding support is needed later, a separate `encoding` field can be added to the spec.
```
lahabana: Any reason to not go with `grpc` `http/protobuf` `http/json` and not ahve `http/json` at the start. Would avoid future impossibility like grpc with json in the future

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,593 @@
...
+
+When `protocol: http`, the `path` field is a base path prefix. The CP appends the signal-specific suffix (`/v1/traces`, `/v1/metrics`, `/v1/logs`) during xDS generation, matching how the OTel SDK handles `OTEL_EXPORTER_OTLP_ENDPOINT`. An empty path means the standard paths are used directly. For gRPC, paths are irrelevant - gRPC routes by protobuf service name.
```
lahabana:
```suggestion
When `protocol: http`, the `path` field is a base path prefix. The CP appends the signal-specific suffix (`/v1/traces`, `/v1/metrics`, `/v1/logs`) during xDS generation, matching how the OTel SDK handles `OTEL_EXPORTER_OTLP_ENDPOINT`. An empty path means the standard paths are used directly. For gRPC, paths are irrelevant and setting it will fail validation - gRPC routes by protobuf service name.
```

docs/madr/decisions/095-mesh-opentelemetry-backend.md
lahabana: I'm not sure why envoy plays a role here. Doesn't envoy to dp always send the stuff the same way (.i.e: grpc)? It's only the dp that does format transformation no?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,593 @@
...
+
+#### Policy reference
+
+Each policy's OTel backend gets an optional `backendRef` field. When set, the endpoint comes from the referenced MeshOpenTelemetryBackend. Signal-specific fields remain inline. In 3.0, `backendRef` becomes required and the inline `endpoint` is removed.
```
lahabana: Can we add a warning for the deprecation? I think that's something we support already no?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,593 @@
...
+5. MeshMetric path: MeshMetric still keeps its `/meshmetric` dynconf route, but only for scrape-side config such as applications, sidecar stats settings, extra labels, and Prometheus state. OTEL backend runtime for metrics, including socket selection and OTEL export `refreshInterval`, comes from `/otel` together with traces and logs. In short, `/meshmetric` answers "what should I scrape" and `/otel` answers "where and how should I export". Inline `endpoint` backends stay on the direct Envoy path and do not use `/otel`.
+
```
lahabana, line 204: Are we clear that using a policy with a backendRef on a dataplane that doesn't have `FeatureOtelViaKumaDp` leads to an error and the policy just doesn't apply. Correct?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,593 @@
...
+spec:
+  targetRef:
+    kind: Mesh
+  default:
+    backends:
+      - type: OpenTelemetry
+        openTelemetry:
+          backendRef:
+            kind: MeshOpenTelemetryBackend
+            name: main-collector
```
lahabana, line 127: Need @lobkovilya to check this. What's up with stuff synced around? Is this display name? Something else?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,593 @@
...
+10. kuma-dp should treat `/otel` as one shared source of truth for OTEL backend runtime. One fetch and decode step should feed both the OTEL receiver side and the metrics pipeline so the three signals cannot drift apart.
+
+#### kuma-dp pipe mode versioning
+
+In 2.14.0, pipe mode should be on by default. The CP should use it whenever `backendRef` is set and the data plane proxy advertises `FeatureOtelViaKumaDp`.
+
+Compatibility note: the unified `/otel` route is only used when the CP selects pipe mode and the data plane proxy advertises `FeatureOtelViaKumaDp`. Older data planes without that feature should keep using direct clusters. Older control planes that do not write `/otel` should keep using the existing direct-cluster path.
+
+To opt out, set `dataPlane.features.otelPipe: false` in the Helm chart. This maps to `DataplaneRuntime.OtelPipeEnabled` in kuma-dp config. When disabled, kuma-dp omits `FeatureOtelViaKumaDp` from its bootstrap feature list; the CP then generates direct static Envoy clusters for all backends, including `backendRef` ones. The behavior is identical to 2.13 and earlier.
```
lahabana: Do we need this? Seems to me like the UDS is only used and present when using a MOTB so it's basically already behind a feature flag. Users that don't want to use it should stop using `backendRef` no?

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: I tightened that up. The Envoy limitation only applies to the direct inline endpoint path. The backendRef pipe path is separate because kuma-dp handles the local OTLP traffic.

bartsmykla: I left protocol at grpc or http and treated encoding separately. Right now Kuma only needs the transport choice. If we ever need non-protobuf support, I would add an encoding field instead of values like http/json.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Clarified. backendRef.name is just the mesh-scoped resource name, and KDS syncs the MeshOpenTelemetryBackend resource itself to zones.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Agreed. I dropped the extra pipe-mode opt-out. backendRef already gates the new path, so the versioning section now sticks to feature advertisement and the 3.0 end state.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Applied. The text now says a non-empty path is invalid with protocol: grpc.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: No. I made that explicit: if FeatureOtelViaKumaDp is missing, backendRef still resolves and the proxy falls back to the pre-pipe path. It does not fail the policy.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
bartsmykla: Added. The MADR now says inline endpoint usage in 2.14 should raise a deprecation warning.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
lahabana: After reading your update I'm still unclear why we bother making this complicated thing.
What this means is that a user is starting to use a new Feature available in 2.14 before updating their DP. To me that's a user error and we shouldn't bother trying to gracefully fallback. People should update their DP and only then they can use MOTB

bartsmykla: Misread your earlier comment, you're right. Updated the MADR: CP skips the backendRef backend and logs a warning when the proxy doesn't have FeatureOtelViaKumaDp. No fallback.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
lobkovilya: You can't use `name` to reference resources that were synced from another cluster (zone or global). The only way to do that is by using `labels`. You can't really guarantee only one MeshOpenTelemetryBackend is referenced at the validation time, but we have a rule to pick the oldest resource when multiple are matched, see https://github.com/kumahq/kuma/blob/a5239a5de79f09c770cdeaa9ac6dc730bbf34b14/pkg/xds/context/destination_index.go#L127-L147

bartsmykla: Right, KDS applies `HashSuffixMapper` so the synced name differs from the original. Updated the MADR to document the resolution strategy: match `backendRef.name` against `kuma.io/display-name` label (preserves original name through sync), direct name match as fast path for local resources, oldest wins on collision. Also documented the name collision scenario when Global and Zone both have a backend with the same name - local wins silently via direct match.

docs/madr/decisions/095-mesh-opentelemetry-backend.md
```diff
@@ -0,0 +1,593 @@
...
+- MeshAccessLog `OtelBackend`: `endpoint` + `attributes` + `body`
+
+Each struct needs a new optional `backendRef` field. The inline `endpoint` remains supported but is deprecated starting in 2.14 and will be removed in 3.0. Validation enforces mutual exclusivity: either `endpoint` or `backendRef`, not both. When inline `endpoint` is used, the implementation should also emit a deprecation warning so operators see the migration path before 3.0.
+
+Use `common_api.BackendResourceRef`, not `TargetRef`. This is a typed resource reference (`kind` + `name`), not policy matching. `backendRef.name` is the operator-facing resource name (what the operator typed when creating the resource). KDS syncs MeshOpenTelemetryBackend to zones, but the synced resource name gets a hash suffix (same as all mesh-scoped resources without `SkipKDSHash`). The `kuma.io/display-name` label preserves the original name through sync, so the CP resolves `backendRef.name` by matching against `kuma.io/display-name`, falling back to direct name match for locally-created resources. This follows the same pattern as `GetDisplayName()` used throughout the codebase for KDS-synced resources.
```
lobkovilya: > so the CP resolves `backendRef.name` by matching against `kuma.io/display-name`

That's not what we should do. `backendRef.name` should always refer to `metadata.name`, we don't want to convert name to display-name under the hood.

You can check how backendRef works in MeshHTTPRoutes. If MeshService and MeshHTTPRoute originate in the same cluster, you can use `backendRef.name` and `backendRef.namespace` to reference. However, if for example MeshService is synced from Global and MeshHTTPRoute is created on zone, you **must** use `backendRef.labels`, that's the only way.

bartsmykla: Agreed. `BackendResourceRef` now has `kind`, `name`, and `labels`. `name` means `metadata.name` only, `labels` for cross-zone. Same pattern as MeshHTTPRoute.

PR #15695, opened 2026-02-24
feat(meshtrace): inject workload KRI in traces

Automaat:
## Motivation

Traces from MeshTrace carry only a bare service name with no zone or mesh context. In multi-zone deployments this makes it impossible to correlate traces with metrics or filter by zone/workload. MeshMetric already injects `kuma.mesh`, `kuma.zone`, and `kuma.workload` — traces should too.

## Implementation information

Added `Mesh`, `Zone`, and `WorkloadKRI` fields to `plugin/xds/Configurer`. In `Configure()`, after user-defined tags are mapped, the three identity tags are appended as literal custom tags — skipping any that are already defined by the user (so user tags are never overridden).

In `plugin.go` `configureListener()`, workload KRI is computed from dataplane labels (`kuma.io/workload`, `kuma.io/zone`, `k8s.kuma.io/namespace`) using the same `kri.Identifier` pattern as MeshMetric. `Mesh` and `Zone` are passed directly from proxy metadata.

Tags are injected on all listener types (inbound, outbound, gateway) for all three backends (Zipkin, Datadog, OpenTelemetry).

> Changelog: feat(meshtrace): inject kuma.mesh/zone/workload span tags

### DIFF
--- a/.github/workflows/build-test-distribute.yaml
+++ b/.github/workflows/build-test-distribute.yaml
@@ -108,7 +108,7 @@ jobs:
         with:
           args: --fix=false --verbose
           version: ${{ env.GOLANGCI_LINT_VERSION }}
-          only-new-issues: ${{ github.event_name == 'pull_request' }}
+          only-new-issues: false
 
       - run: |
           make clean
--- a/app/cni/pkg/cni/utils_linux.go
+++ b/app/cni/pkg/cni/utils_linux.go
@@ -67,7 +67,7 @@ func getPids(name string) []int {
 				continue
 			}
 
-			cmdline, err := os.ReadFile(filepath.Join("/proc", entry.Name(), "cmdline"))
+			cmdline, err := os.ReadFile(filepath.Join("/proc", strconv.Itoa(pid), "cmdline")) // #nosec G703 -- pid is validated integer, path is safe
 			if err != nil {
 				continue
 			}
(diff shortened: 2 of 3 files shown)


pkg/plugins/policies/meshtrace/plugin/v1alpha1/plugin.go
```diff
@@ -159,13 +162,28 @@ func configureListener(ctx xds_context.Context, rules core_rules.SingleItemRules
...
 	configurer := plugin_xds.Configurer{
 		Conf:                  conf,
 		Service:               serviceName,
 		TrafficDirection:      listener.TrafficDirection,
 		Destination:           destination,
 		IsGateway:             proxy.Dataplane.Spec.IsBuiltinGateway(),
 		UnifiedResourceNaming: unified_naming.Enabled(proxy.Metadata, ctx.Mesh.Resource),
+		Mesh:                  proxy.Dataplane.GetMeta().GetMesh(),
+		Zone:                  proxy.Zone,
+		WorkloadKRI:           workloadKRI,
```
lahabana, line 186: Shouldn't that be the Service?

Why are we still using:`serviceName := proxy.Dataplane.Spec.GetIdentifyingService()` ? AFAIK this uses the kuma.io/service tag which will no longer exist

Automaat: This should actually cover this issue: https://github.com/kumahq/kuma/pull/15703

PR #15698, opened 2026-02-25
fix(lint): address golangci-lint 2.10.1 issues

Automaat:
## Motivation

golangci-lint was bumped from 2.9.0 to 2.10.1 which introduced new
gosec rules (G101, G117, G602, G702, G703, G704, G705) and exposed
one existing staticcheck SA5008 issue, resulting in 51 new lint
failures.

## Implementation information

Fixed 5 real issues:
- **G704 (SSRF)**: validate URLs before use in `wait.go`,
  `apply.go`, `remote_bootstrap.go`; constrain HTTP scheme in
  `probes/component.go` to only HTTP/HTTPS
- **G602 (slice OOB)**: fix loop bounds in `tags/match.go` from
  `i < len(s)` to `i+1 < len(s)` to prevent potential panic

For the remaining 46 false positives:
- Added `// #nosec GXXX` inline suppressions on prod code
- Added global gosec exclusion for test paths (`_test.go`, `test/`)
  in `.golangci.yml` — avoids hundreds of suppressions in test files
- Added targeted SA5008 per-file exclusion for `targetref.go`
  (`//nolint` is blocked by repo hooks, per-file config is the only
  option)
- Removed now-unused `//nolint:gosec` directives from test files

> Changelog: skip


slonka:
@Automaat should we backport the things that are real issues?

pkg/xds/generator/testdata/transparent-proxy/09.envoy.golden.yaml
```diff
@@ -58,6 +58,7 @@ resources:
           cluster: self_transparentproxy_passthrough_inbound_ipv4
           statPrefix: self_transparentproxy_passthrough_inbound_ipv4
     name: self_transparentproxy_passthrough_inbound_ipv4
+    statPrefix: self_transparentproxy_passthrough_inbound_ipv4
```
slonka, line 61: why did this change?

slonka: let's leave it in this PR but @bartsmykla could you take a look why this was not there or maybe it shouldn't be there?

bartsmykla: it should be there, I don't know why it wasn't there before

PR #15703, opened 2026-02-25
feat(xds): use workload label when tags disabled

Automaat:
## Motivation

When `InboundTagsDisabled=true` (experimental flag, precursor to Kuma 3.0
tag removal), inbound tags are empty. Reading `kuma.io/service` from inbound
tags returns `\"unknown\"`, breaking observability across all paths: access
logs, traces, metrics labels, MADS monitoring assignments, and Envoy
bootstrap node cluster name.

## Implementation information

Thread `InboundTagsDisabled bool` through `ControlPlaneContext` (set from
`rt.Config().Experimental.InboundTagsDisabled` in the server components).

Introduce `IdentifyingName(inboundTagsDisabled bool)` on `DataplaneResource`
that returns the `kuma.io/workload` label when the flag is true, falling back
to the service tag otherwise. `Dataplane.GetIdentifyingService()` is removed
in favour of this method.

At each observability call site the flag is threaded through:

- `pkg/xds/context/context.go` — new `InboundTagsDisabled` field on
  `ControlPlaneContext`
- `pkg/xds/server/components.go` — sets field from runtime config
- `pkg/plugins/policies/meshaccesslog` — `dataplaneServiceName` helper +
  threaded through internal helpers
- `pkg/plugins/policies/meshtrace` — inline lookup in `configureListener`
- `pkg/plugins/policies/meshmetric` — `service` extra-label skipped when
  workload KRI is set (workload identity already captured via
  `WorkloadAttributeKey`); label set via `IdentifyingName` otherwise
- `pkg/xds/generator/{outbound,transparent,direct_access}` — inline with
  nil guard (tests pass nil `ControlPlane`)
- `pkg/xds/bootstrap/generator.go` — struct field + inline lookup
- `pkg/mads/v1/generator/{assignments,meshmetrics}` — struct field / param
- `pkg/mads/v1/reconcile/snapshot_generator.go` — stores + passes flag
- `pkg/mads/v1/service/service.go` and `pkg/mads/server/server.go` — thread
  from runtime config down to generators
- `pkg/plugins/runtime/gateway` — replaced `GetIdentifyingService()` calls
  with `IdentifyingName(false)` (gateways always identify by service tag)

Default behaviour (`InboundTagsDisabled=false`) is unchanged.


pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -288,7 +293,7 @@ func createDynamicConfig(
 		maps.Copy(extraLabels, mads.DataplaneLabels(proxy.Dataplane, gateways))
 		extraLabels["mesh"] = mesh.GetMeta().GetName()
 		extraLabels["dataplane"] = proxy.Dataplane.GetMeta().GetName()
-		extraLabels["service"] = proxy.Dataplane.Spec.GetIdentifyingService()
+		extraLabels["service"] = proxy.Dataplane.IdentifyingName(inboundTagsDisabled)
```
lahabana: is service right here? Should this just be skipped when we're have workloads?

pkg/core/resources/apis/mesh/dataplane_helpers.go
```diff
@@ -174,6 +175,17 @@ func (d *DataplaneResource) Hash() []byte {
...
 
+// IdentifyingName returns the workload label when inbound tags are disabled,
+// falling back to the identifying service name.
+func (d *DataplaneResource) IdentifyingName(inboundTagsDisabled bool) string {
+	if inboundTagsDisabled {
+		if workload := d.GetMeta().GetLabels()[k8s_metadata.KumaWorkload]; workload != "" {
+			return workload
+		}
+	}
+	return d.Spec.GetIdentifyingService()
```
lahabana: Should you just kill this method?

pkg/plugins/policies/meshaccesslog/plugin/v1alpha1/plugin.go
```diff
@@ -378,7 +386,7 @@ func applyToRealResource(
 	defaultFormat := DefaultFormat(r.Protocol)
 
 	kumaValues := listeners_v3.KumaValues{
-		SourceService:      proxy.Dataplane.Spec.GetIdentifyingService(),
+		SourceService:      proxy.Dataplane.IdentifyingName(inboundTagsDisabled),
```
lahabana, line 389: Does it need to be the inbound KRI instead (.e.g: with a sectionName for the port receiving traffic)?

pkg/plugins/policies/meshtrace/plugin/v1alpha1/plugin.go
```diff
@@ -155,7 +155,7 @@ func applyToRealResources(ctx xds_context.Context, rules core_rules.SingleItemRu
 }
 
 func configureListener(ctx xds_context.Context, rules core_rules.SingleItemRules, proxy *xds.Proxy, listener *envoy_listener.Listener, destination string) error {
-	serviceName := proxy.Dataplane.Spec.GetIdentifyingService()
+	serviceName := proxy.Dataplane.IdentifyingName(ctx.ControlPlane != nil && ctx.ControlPlane.InboundTagsDisabled)
```
lahabana, line 158: Same as before. I have the feel that this may need the sectionName

pkg/core/resources/apis/mesh/dataplane_helpers.go
Automaat: removed

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
Automaat: updated

pkg/plugins/policies/meshtrace/plugin/v1alpha1/plugin.go
Automaat: I dont think we should use kri in tracing, users will get telemetry data wit it, and it wont be easy to correllate it with their other data. We should use just workload by default, which matches using app label in other kubernetes tools

pkg/plugins/policies/meshaccesslog/plugin/v1alpha1/plugin.go
Automaat: changed

bartsmykla (approved):

PR #15708, opened 2026-02-26
docs(MADR): zone proxy as a sidecar

lukidzi:
## Motivation

We want zone proxy to follow the same patter as a regular dataplane

## Implementation information

MADR

## Supporting documentation

Fix #https://github.com/Kong/kong-mesh/issues/9182

### DIFF
--- a/docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
+++ b/docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
@@ -40,9 +40,8 @@ is injected into regular workload pods.
 
 ```yaml
 containers:
-  - name: pause          # dummy main container, never changes
-    image: kumahq/kuma-dp:v2.13.2
-    args: ["pause"]      # dedicated subcommand — sleeps forever, no-op
+  - name: pause          # dummy main container; image rarely changes across Kuma upgrades
+    image: registry.k8s.io/pause:3.10  # override via Helm if image is not available
   # kuma-dp injected here by webhook:
   - name: kuma-sidecar
     image: kumahq/kuma-dp:v2.13.2
@@ -95,18 +94,42 @@ the regular inbound/outbound configuration.
 
 #### Waiting container image
 
-The dummy main container uses `kumahq/kuma-dp` with a dedicated `pause` subcommand added to the
-binary. The subcommand sleeps indefinitely and exits cleanly on SIGTERM.
+**Decision: `registry.k8s.io/pause:3.10`**
 
-**Decision: `kumahq/kuma-dp pause`**
+The `pause` container is bundled with every Kubernetes cluster
+([kubeadm defaults to `registry.k8s.io` as `DefaultImageRepository`](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/apis/kubeadm/v1beta4/defaults.go#L44)):
+- Very small (514 kB) and purpose-built for doing nothing
+- Version (`3.10`) is pinned and changes independently of Kuma upgrades → the Deployment spec is
+  stable between Kuma releases → no unwanted rollouts on `helm upgrade`
+- Already present on every node (Kubernetes uses it for pod sandboxes)
 
-- Operators already pull and mirror this image — no new image to manage or push to a private registry
-- No dependency on `registry.k8s.io` or any third-party registry
-- Maintained and released on the same cadence as Kuma itself
-- The `pause` subcommand adds negligible binary size (~a few KB of Go code)
+Using `kumahq/kuma-dp:<version>` with a `pause` subcommand was considered but **rejected**:
+because that image is version-tagged, any Kuma upgrade would change the Deployment spec and
+trigger a rollout, negating the main benefit of this design.
 
-The Helm chart uses the same `kuma-dp` image reference already configured for the zone
-(e.g., via `global.image.tag`). No new Helm value is needed for the waiting container image.
+A dedicated `kumahq/pause` image was also considered but rejected — it would be functionally
+identical to `registry.k8s.io/pause` with extra maintenance overhead (separate pipeline, release
+cadence, operators still need to mirror it).
+
+The default registry (`registry.k8s.io`) varies per distribution:
+
+| Distribution | Image reference |
+|:-------------|:----------------|
+| GKE | `gke.gcr.io/pause:3.8` |
+| k3d / Rancher | `docker.io/rancher/mirrored-pause:3.6` |
+| upstream kubeadm | `registry.k8s.io/pause:3.10` |
+| others | ... |
+
+The Helm chart exposes an image override per mesh proxy for environments where
+`registry.k8s.io` is not reachable or images are mirrored under a different path
+
+```yaml
+meshes:
+  - name: default
+    ingress:
+      enabled: true
+      image: registry.k8s.io/pause:3.10  # override if image is mirrored elsewhere
+```
 
 #### Configuring sidecar resources
 
@@ -167,13 +190,31 @@ Resource requests/limits use pod-level resources (`spec.resources`), alpha in K8
 beta (enabled by default) from K8s 1.34. For K8s 1.32–1.33 or clusters with the feature gate
 disabled, `ContainerPatch` is the fallback mechanism.
 
-The dummy container runs `kumahq/kuma-dp pause` — the same image operators already manage,
-with a new `pause` subcommand that sleeps indefinitely. No additional images or registry
-dependencies are introduced.
+The dummy container runs `registry.k8s.io/pause:3.10` — bundled with every Kubernetes cluster,
+with a version that changes independently of Kuma releases. Operators can override the image via
+the `image` Helm value per mesh proxy when `registry.k8s.io` is not reachable. No additional images or registry dependencies are introduced beyond what Kubernetes already uses internally.
 
 The sidecar injection webhook is extended to look up the `k8s.kuma.io/zone-proxy-type` label on
 the associated Service and generate the appropriate `kuma-dp` arguments for zone proxy mode
 instead of regular sidecar mode.
+
+#### reachableBackends
+
+Because the injected `kuma-dp` sidecar is a regular Dataplane, it would by default receive
+outbound configuration for all backends in the mesh. Zone proxies must not get populated
+outbounds. The Deployment is annotated with an empty `reachableBackends` list so the webhook
+instructs `kuma-dp` to request no outbound routes:
+
+```yaml
+spec:
+  template:
+    metadata:
+      annotations:
+        kuma.io/reachable-backends: '{"refs":[]}'
+```
+
+This is the same mechanism available to regular workloads. Using an empty list is
+intentional, zone proxies route via the zone ingress/egress path, not direct mesh outbounds.
 
 ### Update behavior
 
@@ -204,5 +245,5 @@ capabilities and exposing no network ports).
 ## Notes
 
 - MADR 094: Zone Proxy Deployment Model (Helm schema, per-mesh templates)
-- Mesh-scoped zone proxies (Dataplane `listeners` array)
-- Syncing zone ingress address across zones
+- MADR 095: Mesh-scoped zone proxies (Dataplane `listeners` array, `k8s.kuma.io/zone-proxy-type` Service label)
+- MADR 096: Syncing zone ingress address across zones


docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,210 @@
...
+
+The Deployment spec is owned by Helm and contains only the dummy container.
+When Kuma is upgraded (new `kuma-dp` image), the Deployment spec is unchanged → **no rollout**.
+
+The Deployment is labeled to opt into sidecar injection:
+```yaml
+metadata:
+  labels:
+    kuma.io/sidecar-injection: enabled
+    k8s.kuma.io/zone-proxy-type: ingress  # or: egress
```
lahabana: Can it be both?

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,210 @@
...
+
+The webhook detects the `k8s.kuma.io/zone-proxy-type` label and configures `kuma-dp` with the
+appropriate zone proxy arguments instead of the regular inbound/outbound configuration.
+
+* Good, because Helm upgrades no longer restart zone proxies
+* Good, because consistent deployment model with regular Dataplanes
+* Good, because reuses existing sidecar injection and bootstrap machinery
+* Good, because aligns with the mesh-scoped zone proxy model (Dataplanes with `listeners`)
+* Bad, because requires a dummy container, which is non-obvious to users inspecting the pod
+* Bad, because zone proxy update cadence decouples from Kuma upgrade cadence
```
lahabana, line 88: They need to do this for the rest of the dataplanes already right?

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,210 @@
...
+  (operators must trigger restarts explicitly to pick up new `kuma-dp`)
+* Bad, because configuring sidecar resources requires `ContainerPatch` or CP injector config
+  rather than a direct `resources:` field in the pod spec (see "Configuring sidecar resources" below)
```
lahabana: For resources anyway we should look at leveraging Pod resources when we can: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#example-2 it's a better way to run these things in the future

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,210 @@
...
+  be adjusted to enable injection in the control-plane namespace
+
+#### Waiting container image
+
+The dummy main container must run indefinitely without consuming resources.
+The following candidates are evaluated against three criteria: **image size**, **pull overhead**, and **security surface**.
+
+| Image | Compressed size | Pull overhead | Shell | Notes |
+|---|---|---|---|---|
+| `registry.k8s.io/pause:3.10` | ~300 KB | Usually none (pre-pulled by kubelet for pod sandboxes) | No | Purpose-built for this use case; used by K8s itself for pod sandboxes |
```
lahabana: If we're 100% all our users will already have this container then it sounds ok but we gotta make this as easy as possible.

At first I thought we were going to do a kuma-dp pause (as the kuma-dp image is already used anyway...)

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,210 @@
...
+      value: '{"requests":{"cpu":"100m","memory":"128Mi"},"limits":{"cpu":"500m","memory":"256Mi"}}'
+```
+
+```yaml
+# Deployment pod template annotation
+annotations:
+  kuma.io/container-patches: zone-ingress-resources
+```
+
+##### 2. Zone-wide default: CP injector configuration
```
lahabana: Do we do both these? Why? I thought we really just wanted to prefer containerPatch as the way to do these things.

Also if we use pod resources when we can this even simpler no?

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
lobkovilya: Why do we need `k8s.kuma.io/zone-proxy-type` on deployment? In [MADR-095](https://github.com/kumahq/kuma/pull/15566) I mentioned only Service needs to be labeled with `k8s.kuma.io/zone-proxy-type`, for deployment just `kuma.io/sidecar-injection` is sufficient

lahabana:
Btw if this is a regular sidecar, how does it work with backends? My guess is that it will get all backends by default. Can we do something WRT reachableBackends (hopefully without making up a new mechanism)

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,210 @@
...
+    image: kumahq/kuma-dp:v2.13.2
+    args: ["run", "--cp-address=..."]
+```
+
+The Deployment spec is owned by Helm and contains only the dummy container.
+When Kuma is upgraded (new `kuma-dp` image), the Deployment spec is unchanged → **no rollout**.
+
+The Deployment is labeled to opt into sidecar injection:
+```yaml
+metadata:
```
lobkovilya: this is `spec.template.metadata`, not just deployment's metadata right?

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
lukidzi: Fixed. It shouldn’t be here, as it’s part of MADR-095 and should be on the Service.

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
lukidzi: I thought about it a bit more, and I think using `kuma-dp` will be a better option for users, since they won’t need to download and republish the image if it’s not in their repository. It will also be easier to explain to users why it’s used.

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
lukidzi: For users on versions 1.32–1.33, we need a fallback (in this case, ContainerPatch), since PodResources are enabled by default starting from 1.34.

lukidzi:
> Btw if this is a regular sidecar, how does it work with backends? My guess is that it will get all backends by default. Can we do something WRT reachableBackends (hopefully without making up a new mechanism)

~Yes we want to use `reachableBackends` but I think it was covered in MADR 095~

Added section as it wasn't there

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,208 @@
...
+binary. The subcommand sleeps indefinitely and exits cleanly on SIGTERM.
+
+**Decision: `kumahq/kuma-dp pause`**
+
+- Operators already pull and mirror this image — no new image to manage or push to a private registry
+- No dependency on `registry.k8s.io` or any third-party registry
+- Maintained and released on the same cadence as Kuma itself
+- The `pause` subcommand adds negligible binary size (~a few KB of Go code)
+
+The Helm chart uses the same `kuma-dp` image reference already configured for the zone
```
lahabana: But this causes helm upgrade to restart the pod right... :( We're going in circle I'm afraid :(

lahabana: There might be a way around this (not including the version maybe but this might be not fine either?)...

For the original pause suggested are we sure users to always bring in the image?

lukidzi: The pause image is bundled with every Kubernetes cluster - [kubeadm](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/images/images.go#L129-L131) defaults to `registry.k8s.io/pause`, but each distribution ships it under a different registry:

  | Distribution     | Image                                |
  |------------------|--------------------------------------|
  | GKE              | gke.gcr.io/pause:3.8                 |
  | k3d / Rancher    | docker.io/rancher/mirrored-pause:3.6 |
  | upstream kubeadm | registry.k8s.io/pause:3.10           |

We could use the `registry.k8s.io/pause` image with a specific version. That would ensure the Deployment is not rolled on upgrades, since the image version would remain unchanged.

Another option is to create a dedicated `kumahq/pause` image, which would basically do the same thing as the `pause` image provided by Kubernetes. The downside is that we would need to maintain an additional image and a pipeline to build it. Even if it’s simple, it still adds extra maintenance overhead.

If we are okay with using the Kubernetes pause image, I would prefer that option because it is very small and already well known. It would only require some clarification in the documentation for users why we download it.

```
registry.k8s.io/pause         3.10                               afb61768ce38   21 months ago   514kB
```

lahabana: Ok this makes sense thanks for looking into it

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,249 @@
...
+
+The Deployment spec is owned by Helm and contains only the dummy container.
+When Kuma is upgraded (new `kuma-dp` image), the Deployment spec is unchanged → **no rollout**.
+
+The pod template is labeled to opt into sidecar injection:
+```yaml
+spec:
+  template:
+    metadata:
+      labels:
```
lobkovilya, line 59: btw currently it's pod annotation, not label https://github.com/kumahq/kuma/blob/5873eeb771c2fcd288a9b547793057c8536e9327/pkg/plugins/runtime/k8s/controllers/pod_controller.go#L99-L106

lukidzi: Step 1
`kuma.io/sidecar-injection: enabled` is a label on the pod template (or namespace). The webhook precheck reads it from pod.Labels:

https://github.com/kumahq/kuma/blob/5873eeb771c2fcd288a9b547793057c8536e9327/pkg/plugins/runtime/k8s/webhooks/injector/precheck.go#L100

Note: Annotations() is just a map[string]string wrapper: the same helper is used to read both labels and annotations depending on what map you pass in.

Step 2

After injecting `kuma-dp`, the webhook sets `kuma.io/sidecar-injected: true` as an annotation on the pod:

https://github.com/kumahq/kuma/blob/5873eeb771c2fcd288a9b547793057c8536e9327/pkg/plugins/runtime/k8s/webhooks/injector/injector.go#L744-L750

Step 3

https://github.com/kumahq/kuma/blob/5873eeb771c2fcd288a9b547793057c8536e9327/pkg/plugins/runtime/k8s/controllers/pod_controller.go#L99-L106 reads `kuma.io/sidecar-injected` from `pod.Annotations` to decide whether to create a Dataplane:

* `kuma.io/sidecar-injection` is a label because the user (or Helm) sets it to express intent, and webhook selectors can match on it
*  `kuma.io/sidecar-injected` is an annotation because the webhook sets it as a runtime status marker after it acts

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,249 @@
...
+The zone proxy Deployment contains only `kuma-dp` as the main container.
+The Helm chart manages the full pod spec, including the `kuma-dp` image and arguments.
+
+* Good, because no changes required
+* Good, because simple - no injection machinery involved
+* Good, because it's easier to adjust resources for a ZoneProxy
+* Bad, because any Helm upgrade that touches the Deployment spec triggers a zone proxy restart
+* Bad, because inconsistent with the regular Dataplane deployment model
+
+### Option 2: Sidecar injection with a dummy main container
```
Automaat, line 35: did we analyse any other approaches? For example:  
- Webhook helm image mutation
- CP-managed Deployments

It feels like MADR is incomplete without comparing this to other alternatives

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,249 @@
...
+
+The zone proxy Deployment contains a **dummy main container** (a no-op process) as the main container.
+`kuma-dp` is injected as a sidecar by the existing Kuma sidecar injection webhook, the same way it
+is injected into regular workload pods.
+
+```yaml
```
Automaat, line 41: it is missing from MADR but for better UX we could add `kubectl.kubernetes.io/default-container: "kuma-sidecar"` annotation

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,249 @@
...
+
+* Good, because no changes required
+* Good, because simple - no injection machinery involved
+* Good, because it's easier to adjust resources for a ZoneProxy
+* Bad, because any Helm upgrade that touches the Deployment spec triggers a zone proxy restart
+* Bad, because inconsistent with the regular Dataplane deployment model
+
+### Option 2: Sidecar injection with a dummy main container
+
+The zone proxy Deployment contains a **dummy main container** (a no-op process) as the main container.
```
Automaat, line 37: do we have a plan for naming this container to make sure it is easy to understand that it does nothing? Like: `zone-proxy-anchor` or something similar?

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,249 @@
...
+  template:
+    metadata:
+      annotations:
+        kuma.io/reachable-backends: '{"refs":[]}'
+```
+
+This is the same mechanism available to regular workloads. Using an empty list is
+intentional, zone proxies route via the zone ingress/egress path, not direct mesh outbounds.
+
+### Update behavior
```
Automaat, line 219: Can we write more about migration? I would like to see actual migration steps when going from Helm managed zone proxy to sandalone deployments. Also it would be nice to see how rollback would look like

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,249 @@
...
+containers:
+  - name: pause          # dummy main container; image rarely changes across Kuma upgrades
+    image: registry.k8s.io/pause:3.10  # override via Helm if image is not available
+  # kuma-dp injected here by webhook:
+  - name: kuma-sidecar
+    image: kumahq/kuma-dp:v2.13.2
+    args: ["run", "--cp-address=..."]
+```
+
+The Deployment spec is owned by Helm and contains only the dummy container.
```
Automaat, line 51: is this behind some flag in helm? how does it work with old zone proxies?

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
lukidzi: This was already covered in https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/094-zone-proxy-deployment-model.md, where we chose this option and documented it. We wanted just to clarify details

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
lukidzi: The name of the container could be `pause` (as it's clear) but we can name it better if you think it gives more value

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,249 @@
...
+
+Technical Story: https://github.com/Kong/kong-mesh/issues/9182
+
+## Context and Problem Statement
+
+Zone proxies (zone ingress and zone egress) are currently deployed as standalone Kubernetes Deployments
+where `kuma-dp` is the **main container** in the pod. This differs from how Kuma deploys proxies for
+regular workloads, where `kuma-dp` is injected as a sidecar alongside the application container.
+
+This standalone model creates two problems:
```
Automaat, line 13: Which problem is more important to us, Helm restarting proxies or inconsistency in deployment mode? 

In my opinion the importance can drastically change optics on how we look at the MADR

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
lukidzi: This was covered by https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/094-zone-proxy-deployment-model.md. and here https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md#marking-services-as-zone-proxies

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Option 1: Keep standalone Deployment (status quo)
+
+The zone proxy Deployment contains only `kuma-dp` as the main container.
+The Helm chart manages the full pod spec, including the `kuma-dp` image and arguments.
+
+* Good, because no changes required
+* Good, because simple - no injection machinery involved
+* Good, because it's easier to adjust resources for a ZoneProxy
+* Bad, because any Helm upgrade that touches the Deployment spec triggers a zone proxy restart
```
Automaat, line 32: We've lived with this approach for a really long time, was there actually any complaints about this, or do you just want to do some cleanup before 3.0? I think I am missing some context for this MADR

lukidzi: It's problematic as it might introduce unexpected restart (it's not expected as can potentially introduce a small interruption of traffic). That was one of the thing mentioned before

docs/madr/decisions/097-zone-proxy-sidecar-deployment.md
lukidzi: I think both are not desired and if we can address both it's even better.

PR #15748, opened 2026-03-03
feat(helm): mesh zone proxy

### LINKED ISSUE #15681: Implement zone proxy deployment model from MADR 094
## Summary

Implement the deployment tooling decisions from MADR 094 (#15460) for mesh-scoped zone proxies.

## Scope

**Helm chart changes**
- Add `meshes` list to `values.yaml` (default: `[]`)
- Per-mesh templates rendering: Deployment(s), Service(s), HPA, PDB, ServiceAccount, optional Mesh resource and default policies
- Validate Service name length (63-char limit) with `service.name` override
- Validate `ingress`/`egress` and `combinedProxies` mutual exclusivity
- Raw K8s spec passthrough (podSpec, resources, hpa, pdb)

## References

- MADR 094: docs/madr/decisions/094-zone-proxy-deployment-model.md
- PR: #15460

slonka:
## Motivation

Zone proxies are currently global-scoped - a single instance handles all meshes in a zone.
[MADR-094](https://github.com/kumahq/kuma/blob/7c7f26b42ec552d76162702027b9ba3686038729/docs/madr/decisions/094-zone-proxy-deployment-model.md) moves them to mesh-scoped resources to enable per-mesh policies, mTLS identity, and observability.
This PR implements the Helm deployment tooling for mesh-scoped zone proxies (Phase 1, additive migration).

## Implementation information

Adds a `meshes` list to `values.yaml`.
Each entry renders per-mesh Deployment(s), Service(s), HPA, PDB, and ServiceAccount.
Supports separate `ingress`/`egress` deployments.
Existing top-level `ingress`/`egress` keys remain functional.

## Supporting documentation

MADR 094: https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/094-zone-proxy-deployment-model.md

Fixes #15681


deployments/charts/kuma/templates/mesh-zoneproxy-deployment.yaml
```diff
@@ -0,0 +1,163 @@
...
+    metadata:
+      annotations:
+        kuma.io/reachable-backends: '{"refs":[]}'
+      labels:
+        {{- include "kuma.mesh.zoneproxy.labels" (dict "root" $ "meshName" $meshName "role" $role) | nindent 8 }}
+        kuma.io/sidecar-injection: enabled
+    spec:
+      {{- if $cfg.resources }}
+      resources: {{ toYaml $cfg.resources | nindent 8 }}
+      {{- end }}
```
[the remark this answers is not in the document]
lobkovilya: https://kubernetes.io/docs/tasks/configure-pod-container/assign-pod-level-resources/

deployments/charts/kuma/README.md
```diff
@@ -223,6 +223,7 @@ A Helm chart for the Kuma Control Plane
...
...Percentage: 80       # -- Pod Disruption Budget settings.       pdb:         enabled: false         maxUnavailable: 1      egress:       # -- Deploy a zone egress for this mesh.       enabled: true       replicas: 1       # -- Pause container image (dummy main container for sidecar injection).       # Override if registry.k8s.io is not reachable or mirrored elsewhere.       image: registry.k8s.io/pause:3.10       # -- Override the auto-generated Service name (max 63 chars).       # Auto-generated: <name>-<mesh>-egress (where <name> is the chart name or nameOverride)       service:         name: ""       # -- Resource requests and limits for the pod (pod-level resources).       # Applied to all containers in the pod (pause + injected kuma-sidecar).       resources: {}       # -- Subset of Kubernetes PodSpec fields applied to the pod template       # (nodeSelector, tolerations, affinity, topologySpreadConstraints,       #  priorityClassName, securityContext, containerSecurityContext).       podSpec: {}       hpa:         enabled: false         minReplicas: 2         maxReplicas: 5         targetCPUUtilizationPercentage: 80       pdb:         enabled: false         maxUnavailable: 1 |
```
lukidzi: This looks kinda bad, could we get a better structure here?

deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
lukidzi, line 237: Should we mention somewhere that namespace `kuma-system` needs to have `kuma.io/sidecar-injection: disabled` as we don't want to inject a sidecar to the control-plane and have a circular dependency

deployments/charts/kuma/templates/post-install-rollout-zoneproxy-job.yaml
```diff
@@ -0,0 +1,142 @@
...
+  {{- include "kuma.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "kuma.name" . }}-rollout-zoneproxy-job
+subjects:
+  - kind: ServiceAccount
+    name: {{ $serviceAccountName }}
+    namespace: {{ .Release.Namespace }}
+---
```
lukidzi, line 72: Should this be a RoleBinding and Role with namespace where the zone-proxy is deployed?

app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
```diff
@@ -757,6 +757,71 @@ egress:
...
+#   - name: default
+#
+#     ingress:
+#       # -- Deploy a zone ingress for this mesh.
+#       enabled: true
+#       # -- Number of replicas. Ignored when hpa.enabled is true.
+#       replicas: 1
+#       # -- Pause container image (dummy main container for sidecar injection).
+#       # Override if registry.k8s.io is not reachable or mirrored elsewhere.
+#       image: registry.k8s.io/pause:3.10
```
lukidzi: Should this have the same format as we have in values and be able to override everything in a one place: 
`Values.global.registry`

app/kumactl/cmd/install/testdata/install-cp-helm/certManagerCustomIssuer.golden.yaml
lukidzi, line 1: Should we add a new test case which deploys mesh-proxy in 2 meshes?

deployments/charts/kuma/templates/post-install-rollout-zoneproxy-job.yaml
```diff
@@ -0,0 +1,142 @@
+{{- $hasZoneProxies := false }}
```
lukidzi, line 9: Should we add a comment why we need this job?

app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
lobkovilya: I can change the format, but for non-kuma images I don't think the override works, see https://github.com/kumahq/kuma/blob/a71d0f97d901d9cb894bdfd2584518d2aefb03c4/deployments/charts/kuma/templates/pre-install-patch-namespace-job.yaml#L105

deployments/charts/kuma/templates/post-install-rollout-zoneproxy-job.yaml
lobkovilya: Fixed

deployments/charts/kuma/README.md
lobkovilya: Found a way to display it nicer, basically I keep settings for default mesh with disabled zone proxies. WDYT?

deployments/charts/kuma/templates/post-install-rollout-zoneproxy-job.yaml
lobkovilya: Added

app/kumactl/cmd/install/testdata/install-cp-helm/certManagerCustomIssuer.golden.yaml
lobkovilya: added a test

deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
lobkovilya: I don't know, we have a helm pre-install job that automatically patches the system namespace to have `kuma.io/sidecar-injection: false` https://github.com/kumahq/kuma/blob/4f41c74ae87a40c3b1257d7f7abd3d7a41d148b2/deployments/charts/kuma/templates/pre-install-patch-namespace-job.yaml#L123

deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
```diff
@@ -241,15 +241,13 @@ webhooks:
           operator: NotIn
           values:
           - kube-system
-          - {{ .Release.Namespace }}
-      {{- if gt (len (default list .Values.namespaceAllowList)) 0 }}
         - key: kubernetes.io/metadata.name
           operator: In
           values:
+          - {{ .Release.Namespace }}
```
slonka: Making the `In` selector unconditional (previously gated on non-empty `namespaceAllowList`) changes the default behavior of `pods-kuma-injector`.

**Before** (empty `namespaceAllowList`): no `In` selector present, so the webhook matches any namespace with `kuma.io/sidecar-injection` label (except kube-system and kuma-system) - enabling per-pod opt-in across all labeled namespaces.

**After**: `In [kuma-system]` is always present, restricting the webhook to only kuma-system. Per-pod sidecar opt-in (pod labeled `kuma.io/sidecar-injection: enabled` in a namespace where the label is set to a non-enabled value like `disabled`) stops working for all namespaces except kuma-system.

This is needed for mesh zone proxy injection in kuma-system, but it's a breaking change for the per-pod opt-in pattern in other namespaces.

**How it manifests on upgrade:** After `helm upgrade`, existing pods keep running with sidecars - nothing breaks immediately. But when a pod is later recreated (deployment rollout, node drain, crash), the new pod silently starts without a sidecar because the webhook no longer matches that namespace. No error - the pod works from Kubernetes' perspective but is no longer in the mesh (no mTLS, no traffic routing, no observability). This is a delayed, silent failure that could surface days/weeks after the upgrade.

The blast radius is limited to users who use per-pod opt-in (namespace label exists but is NOT `enabled`/`true`). The common namespace-level injection pattern (`kuma.io/sidecar-injection: enabled` on the namespace) uses `namespace-kuma-injector` and is unaffected.

Should this be documented in UPGRADE.md, or is there a way to preserve both behaviors (e.g. keeping the conditional and adding kuma-system separately)?

deployments/charts/kuma/templates/mesh-zoneproxy-deployment.yaml
```diff
@@ -0,0 +1,115 @@
...
+      {{ toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with (and $cfg.podSpec $cfg.podSpec.tolerations) }}
+      tolerations:
+      {{ toYaml . | nindent 8 }}
+      {{- end }}
+      terminationGracePeriodSeconds: {{ default 40 $cfg.terminationGracePeriodSeconds }}
+      containers:
+        - name: pause
+          image: {{ $cfg.image.registry }}/{{ $cfg.image.repository }}:{{ $cfg.image.tag }}
```
slonka, line 88: The pause image uses `$cfg.image.registry` directly without falling back to `$.Values.global.image.registry`. Other chart components (kuma-cp, kuma-dp, etc.) respect `global.image.registry` for private registry / air-gapped setups. Should this also respect `global.image.registry` as a fallback? (Same point as @lukidzi's comment on the values file.)

lobkovilya: Same reply as for Lukasz's comment https://github.com/kumahq/kuma/pull/15748#discussion_r2996016332 :) 

We don't have fallback for `kubectl` images, see the [pre-install-patch-namespace-job.yaml](https://github.com/kumahq/kuma/blob/a71d0f97d901d9cb894bdfd2584518d2aefb03c4/deployments/charts/kuma/templates/pre-install-patch-namespace-job.yaml#L105), so I'm not sure if we need to add the fallback here.

lukidzi (dismissed):

deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
lobkovilya: That's a very good catch! The behaviour shouldn't change, so the best way I found to solve this is handling sidecars in `kuma-system` in a separate webhook `mesh-zoneproxy-kuma-injector.kuma.io`

deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
```diff
@@ -269,6 +269,39 @@ webhooks:
         resources:
           - pods
     sideEffects: None
+  - name: mesh-zoneproxy-kuma-injector.kuma.io
```
slonka, line 272: Nice webhook split. Namespace disjointness is correct and the dual-label `objectSelector` means `/inject-sidecar` needs no code changes 🎉

deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
```diff
@@ -269,6 +269,39 @@ webhooks:
...
+    admissionReviewVersions: ["v1"]
+    failurePolicy: {{ .Values.controlPlane.injectorFailurePolicy }}
```
slonka, line 274: This webhook is always registered even with `meshes: []`. Was gating on non-empty `meshes` considered?

deployments/charts/kuma/templates/mesh-zoneproxy-deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+      {{- end }}
+      {{- with (and $cfg.podSpec $cfg.podSpec.priorityClassName) }}
+      priorityClassName: {{ . }}
+      {{- end }}
+      {{- if and $cfg.podSpec $cfg.podSpec.securityContext }}
+      securityContext:
+      {{- toYaml $cfg.podSpec.securityContext | trim | nindent 8 }}
+      {{- else }}
+      securityContext:
+        runAsUser: 5678
```
slonka, line 69: Why did we drop `runAsNonRoot`?

slonka:
Approved (since Ilya took it over I can't actually approve) 👍

deployments/charts/kuma/templates/mesh-zoneproxy-deployment.yaml
lobkovilya: It's a pod level security context, `runAsNonRoot: true` here doesn't allow the sidecar to run as root.

deployments/charts/kuma/templates/cp-webhooks-and-secrets.yaml
lobkovilya: No, but it probably makes sense, let me try

lobkovilya: I tried, but it's makes the following flow a bit broken:

1. Install with `meshes[0].ingress.enabled: false` (no webhook is installed)
2. Upgrade with `meshes[0].ingress.enabled: true`

Since webhook and deployment are installed at the same time the deployment doesn't have the sidecar. Usually `post-install` webhook rollouts the deployments as a workaround, but it's not registered for `upgrades`. However, if we register it for upgrade then zoneproxies are going to restart on every upgrade and we don't want that.

slonka: fair enough, doesn't make sense to complicate this

lukidzi (approved):

Issue #15631, opened 2026-02-18
Create follow-up MADR to supersede MADR 093 (disallow multiple meshes per K8s namespace)


slonka:
MADR 094 (#15460) reverts the decision made in MADR 093 by allowing multiple meshes per namespace to support mesh-scoped zone proxy deployments in `kuma-system`. A follow-up MADR should be created to formally supersede MADR 093 and document the new approach to handling Workload collisions across meshes.

Context: https://github.com/kumahq/kuma/pull/15460#discussion_r2821659183

PR #15750, opened 2026-03-04
docs(MADR): supersede madr-093 by reverting the decision

lobkovilya:
## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/15631

### DIFF
--- a/docs/madr/decisions/099-allow-multiple-meshes-per-k8s-ns.md
+++ b/docs/madr/decisions/099-allow-multiple-meshes-per-k8s-ns.md
@@ -1,0 +1,59 @@
+# Allow Multiple Meshes per Kubernetes Namespace
+
+* Status: accepted
+
+* Supersedes: [MADR 093](093-disallow-multiple-meshes-per-k8s-ns.md)
+
+Technical Story: https://github.com/kumahq/kuma/issues/15631
+
+## Context and Problem Statement
+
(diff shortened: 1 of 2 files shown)


docs/madr/decisions/099-allow-multiple-meshes-per-k8s-ns.md
```diff
@@ -0,0 +1,59 @@
...
+### Allow Multiple Meshes per Namespace
+
+Multiple meshes are allowed in a single Kubernetes namespace.
+This is a general change — not scoped only to `kuma-system`.
+
+### Handle Workload Collisions in Controller
+
+Instead of preventing multiple meshes per namespace,
+the Workload controller will fail with a clear error message if a Workload name collision occurs across meshes.
+This provides explicit feedback to users while not blocking valid use cases like mesh-scoped zone proxies.
```
lahabana, line 31: It would be good for this to be a bit more precise of what will be done here.

I think it's worth stating that the policy will be "first win". If we try to change the mesh of workload that already exist it indicates a "collision" and it will fail with an error and backoff ?

lobkovilya: It's already handled, see https://github.com/kumahq/kuma/pull/15097. We are just going to remove `disallowMultipleMeshesPerNamespace` flag that's `false` by default.

Issue #15672, opened 2026-02-23
Create MADR for zone proxy deployment topology (shared vs separate ingress/egress)


slonka:
MADR 094 (#15460) defers the zone proxy deployment topology decision to a separate MADR (footnote [^2]). This should cover:

- Whether zone ingress and zone egress share a single Deployment or remain separate
- The `combinedProxies` option in the `meshes` Helm values
- How users opt into ingress-only, egress-only, or both

**Note:** MADR 095 (PR #15566) proves the resource model supports combined listeners on a single Dataplane, but the deployment topology recommendation is not addressed there.

**Context:** footnote [^2] in MADR 094, review comments on PR #15460 (lines 628, 810, 915)

PR #15752, opened 2026-03-04
docs(MADR): provide reasoning for each deployment model

lukidzi:
## Motivation

Close: https://github.com/kumahq/kuma/issues/15672

## Implementation information

Added madr if we should go for combined or ingress/egress separate

### DIFF
--- a/docs/madr/decisions/098-zone-proxy-deployment-topology.md
+++ b/docs/madr/decisions/098-zone-proxy-deployment-topology.md
@@ -88,7 +88,7 @@ Pros:
 - Good, because different resource limits and Kubernetes security contexts can be applied per role.
 
 Cons:
-- Bad, because it requires managing two Deployments per mesh instead of one, which also increases cost and complexity.
+- Bad, because it requires managing two Deployments(K8s)/Running applications(Universal) per mesh instead of one, which also increases cost and complexity.
 
 ### Option 2: Combined Deployment (`combinedProxies`)
 


docs/madr/decisions/098-zone-proxy-deployment-topology.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+Pros:
+- Good, because it preserves the existing operational model operators are familiar with.
+- Good, because each role scales independently — a zone that primarily serves traffic can scale ingress without scaling egress.
+- Good, because ingress and egress are in separate failure domains; a crash in one does not affect the other (e.g.: incorrect xDS configuration, identity provisioning error).
+- Good, because different resource limits and Kubernetes security contexts can be applied per role.
+
+Cons:
+- Bad, because it requires managing two Deployments per mesh instead of one, which also increases cost.
```
lahabana:
```suggestion
- Bad, because it requires managing two Deployments per mesh instead of one, which also increases cost and complexity.
```

docs/madr/decisions/098-zone-proxy-deployment-topology.md
```diff
@@ -0,0 +1,175 @@
...
+      - type: ZoneEgress
+        address: 10.0.0.1
+        port: 10002
+        name: ze-port
+        state: Ready
+```
+
+`combinedProxies` is mutually exclusive with `ingress`/`egress` — Helm validates and fails if both are set.
+
+Even in combined mode, a Service is created only for the zone ingress port so its address can be advertised to other zones for cross-zone routing.
```
lahabana:
```suggestion
Even in combined mode, a Service type load balancer is created only for the zone ingress port so its address can be advertised to other zones for cross-zone routing.
```

docs/madr/decisions/098-zone-proxy-deployment-topology.md
```diff
@@ -0,0 +1,175 @@
...
+
+Pros:
+- Good, because it preserves the existing operational model operators are familiar with.
+- Good, because each role scales independently — a zone that primarily serves traffic can scale ingress without scaling egress.
+- Good, because ingress and egress are in separate failure domains; a crash in one does not affect the other (e.g.: incorrect xDS configuration, identity provisioning error).
+- Good, because different resource limits and Kubernetes security contexts can be applied per role.
+
+Cons:
+- Bad, because it requires managing two Deployments per mesh instead of one, which also increases cost.
+
```
lahabana, line 92: I guess another bad is that running stuff in universal is sometimes painful so less instances is better :)

lukidzi:
> Sounds good! Feature wise this is pretty simple right?

Yes, it doesn't really change many changes. We can support in helm combined proxy but by the default we are going to make them separate.

lobkovilya (approved):
Also going from separate to combined should be pretty straightforward, you just deploy ingress (or egress) and apply new k8s service with `k8s.kuma.io/zone-proxy-type: egress` label. This should give you a combined proxy.

PR #15753, opened 2026-03-05
docs(MADR): add tool to list madrs and update to 2 approvers

lahabana:
Move the 2 approvers because the team is smaller.
The README expected a list which we never update add a script to do it instead


lahabana:
@copilot can you fix the issues you pointed?

bartsmykla (approved):

PR #15759, opened 2026-03-05
feat(zoneproxy): added fields for new zoneproxy model to dataplane

lukidzi:
## Motivation

ZoneIngress and ZoneEgress are currently global-scoped resources, which prevents mesh-scoped mTLS identity and policy application for zone proxy traffic. This is the first step toward MADR-095, which proposes replacing them by embedding zone proxy listeners directly inside the `Dataplane` resource. This PR introduces the data model foundation: a new Listener message in the `Dataplane.Networking` proto, along with helpers and validation.

## Implementation information

* added proto fields
* added validators

## Supporting documentation

Fix #https://github.com/Kong/kong-mesh/issues/9248


api/mesh/v1alpha1/dataplane.proto
```diff
@@ -328,6 +328,51 @@ message Dataplane {
...
+      string name = 4;
+
+      // State describes the current health state of the listener.
+      // The control plane sets this based on the readiness of the underlying
+      // pod and sidecar container.
+      State state = 5;
+    }
+
+    // Listeners describes zone proxy listeners embedded in this Dataplane.
+    // When listeners are present and no inbound or gateway is defined
```
lahabana: Even when inbound are present right?

api/mesh/v1alpha1/dataplane.proto
```diff
@@ -328,6 +328,51 @@ message Dataplane {
...
+
+      // Type determines the role of this listener: ZoneIngress for inbound
+      // cross-zone traffic or ZoneEgress for outbound external traffic.
+      Type type = 1;
+
+      // Address on which the listener will be exposed.
+      string address = 2;
+
+      // Port on which the listener will be exposed.
+      uint32 port = 3 [(validate.rules).uint32 = {lte: 65535}];
```
lobkovilya: we don't generate proto validators, you can remove `validate.rules`

api/mesh/v1alpha1/dataplane.proto
```diff
@@ -328,6 +328,51 @@ message Dataplane {
...
+      // State describes the current health state of the listener.
+      // The control plane sets this based on the readiness of the underlying
+      // pod and sidecar container.
+      State state = 5;
+    }
+
+    // Listeners describes zone proxy listeners embedded in this Dataplane.
+    // Listeners may coexist with inbounds; when listeners are defined without
+    // any inbound or gateway (zone-proxy-only mode), the data plane proxy acts
+    // exclusively as a zone ingress, egress, or both.
```
lobkovilya: > when listeners are defined without any inbound or gateway (zone-proxy-only mode), the data plane proxy acts exclusively as a zone ingress, egress, or both.

I feel like this sentence doesn't provide any new info :)

pkg/core/resources/apis/mesh/dataplane_validator_test.go
```diff
@@ -1355,6 +1400,199 @@ var _ = Describe("Dataplane", func() {
...
+                      tags:
+                        kuma.io/service: backend
+                  listeners:
+                    - type: ZoneIngress
+                      address: 192.168.0.1
+                      port: 10001`,
+			expected: `
+                violations:
+                - field: networking.listeners[0].name
+                  message: name can't be empty`,
```
lobkovilya: Name can be empty for the inbound, we should probably make it optional for listeners as well

api/mesh/v1alpha1/dataplane.proto
```diff
@@ -328,6 +328,53 @@ message Dataplane {
...
+
+      // Address on which the listener will be exposed.
+      string address = 2;
+
+      // Port on which the listener will be exposed.
+      uint32 port = 3;
+
+      // Name uniquely identifies this listener within the Dataplane and is
+      // used to reference it via sectionName in policies.
+      string name = 4;
```
lobkovilya, line 368: Please add a comment that `name` is optional if if unset we're using the port value as a string.

pkg/core/resources/apis/mesh/dataplane_validator_test.go
```diff
@@ -378,6 +378,51 @@ var _ = Describe("Dataplane", func() {
...
+              listeners:
+                - type: ZoneIngress
+                  address: 192.168.0.1
+                  port: 10001
+                  name: zi-a
+                - type: ZoneIngress
+                  address: 192.168.0.1
+                  port: 10003
+                  name: zi-b`,
+		),
```
lobkovilya, line 425: please add valid entry where listener doesn't have `name`

lobkovilya (approved):

PR #15791, opened 2026-03-10
fix(matchers): match delegated gw dpps

bartsmykla:
## Motivation

Policies with `kind: Dataplane` as top-level targetRef didn't match delegated gateway proxies. The label matching succeeded, but the policy was never applied because `DppSelectedByPolicy` hardcoded `delegatedGateway=false` in the `Dataplane` case. This blocked customers from applying zone-scoped routing policies (like MeshHTTPRoute) to delegated gateways from the global control plane.

## Implementation

- In `DppSelectedByPolicy` (`pkg/plugins/policies/core/matchers/dataplane.go`), the `Dataplane` case returned `delegatedGateway=false` unconditionally. Builtin gateways early-return before this branch, so only sidecars and delegated gateways reach it. Changed `false` to `dpp.Spec.IsDelegatedGateway()`.
- Added a `delegated-gateway-select-by-labels` test case using MeshHTTPRoute (the policy type from the customer report) with golden files covering all zone/origin combinations.


Automaat (dismissed):

pkg/plugins/policies/core/matchers/testdata/matchedpolicies/dataplane-kind/delegated-gateway-select-by-labels.policies.yaml
```diff
@@ -0,0 +1,28 @@
+# policies using kind Dataplane selecting delegated gateway by labels
+type: MeshTrafficPermission
+mesh: mesh-1
+name: mtp-1
+spec:
+  targetRef:
+    kind: Dataplane
+    labels:
+      app: kong-gateway
+  from:
```
lobkovilya: please don't use old `spec.from` policies when `spec.rules` is supported

bartsmykla: addressed

PR #15793, opened 2026-03-10
chore(lint): fix gosec 2.11.2 regressions

bartsmykla:
## Motivation

The golangci-lint 2.11.2 bump (#15780) introduced new gosec rules that flag existing code. G118 catches `context.WithCancel` calls where the cancel function isn't deferred, and G703 uses taint analysis to flag file path operations on variable inputs.

## Implementation information

Three categories of fixes across 12 files:

1. Added `defer cancel()` where context cancel functions weren't being cleaned up on all exit paths (G118). These are mostly `Start()` methods that create a context and spawn a goroutine to cancel on shutdown signal - the defer ensures cleanup if the function returns before the goroutine fires.

2. Added `//nolint:gosec` with explanations where cancel is intentionally stored and called from a different code path (stream close callbacks, disconnect handlers). Also wrapped watchdog goroutines with `defer cancel()` so the context gets cleaned up if the watchdog exits on its own.

3. Applied `filepath.Clean` to satisfy G703 taint analysis on file paths that come from config or CLI args. The paths are already validated upstream but gosec can't trace that.

## Supporting documentation

Follow-up to #15780

> Changelog: skip


Automaat (approved):

pkg/core/resources/apis/workload/generate/generator.go
```diff
@@ -141,6 +141,7 @@ func (g *Generator) Start(stop <-chan struct{}) error {
 	defer ticker.Stop()
 	ctx := user.Ctx(context.Background(), user.ControlPlane)
 	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
lahabana, line 146: Would it be clear to cancel just where the return is? Feels like this might be a little error prone.

bartsmykla: Both are intentional. `defer cancel()` covers every return path including early errors, while the `<-stop` goroutine cancels on shutdown so in-flight work exits right away. `cancel()` is idempotent, so calling it from both places is safe. I can add a comment if you want that explicit in the code.

PR #15794, opened 2026-03-10
fix(injector): decouple probe ports from admin

Automaat:
## Motivation

the Envoy admin interface (port 9901) is accessible from any container in a meshed pod via localhost with no authentication. Setting `KUMA_BOOTSTRAP_SERVER_PARAMS_ADMIN_PORT=0` to disable admin breaks pod creation with native sidecar containers because the injector hardcodes probe ports to the admin port value — port 0 is rejected by Kubernetes (must be 1-65535).

The kuma-dp readiness server (port 9902, #11107) already exists and serves `/ready` with the same semantics, but the injector never uses it for probes.

## Implementation information

When `DefaultAdminPort == 0` (admin disabled on CP), sidecar probes (liveness, readiness, startup) and the `waitForDataplaneReady` lifecycle hook now fall back to the kuma-dp readiness port (9902) instead of the admin port. The readiness port is referenced from `kumadp_config.DefaultConfig().Dataplane.ReadinessPort` to stay in sync with kuma-dp defaults.

When admin is enabled (default, port 9901), behavior is unchanged.

> Changelog: fix(injector): decouple sidecar probe ports from Envoy admin port


Automaat:
closing in favour of https://github.com/kumahq/kuma/pull/15795

PR #15795, opened 2026-03-10
feat(bootstrap): admin API UDS support

Automaat:
## Motivation

The Envoy admin API (`127.0.0.1:9901`) is a zero-auth, full-control interface. In Kubernetes, all containers in a pod share the same network namespace, so a compromised app container can reach the admin API and:

- Kill the sidecar via `/quitquitquit` (bypassing mTLS enforcement)
- Extract service mesh topology via `/config_dump` and `/clusters`
- Modify runtime behavior via `/runtime_modify`
- Disable logging via `/logging` to hide activity

This PR adds support for binding the Envoy admin API to a Unix domain socket instead of TCP. UDS is not accessible from the app container unless the socket directory is explicitly mounted, eliminating the shared-network-namespace attack vector.

## Implementation information

Follows the existing `FeatureReadinessUnixSocket` pattern:

- **Config**: `KUMA_EXPERIMENTAL_ADMIN_UNIX_SOCKET` on the CP (default `false`, opt-in for now)
- **Bootstrap**: generates Envoy `admin.address.pipe` instead of `admin.address.socket_address`
- **Admin proxy generator**: envoy admin cluster connects via UDS when feature enabled
- **Readiness**: separate readiness reporter binds to all interfaces (IPv4/IPv6-aware) when admin is on UDS, so K8s probes and PostStart hooks still work

The admin proxy listener on `podIP:adminPort` remains TCP-based for K8s probes and CP mTLS access. Only the Envoy-internal admin binding changes.

**Opt-in rationale**: `kuma-dp wait` postStart hook currently uses `http://localhost:adminPort/ready`. Enabling UDS by default would break this. A follow-up can make it default after updating the K8s factory to pass the socket path.

> Changelog: feat(bootstrap): add UDS support for Envoy admin API

### DIFF
--- a/app/kuma-dp/pkg/dataplane/readiness/component.go
+++ b/app/kuma-dp/pkg/dataplane/readiness/component.go
@@ -191,7 +191,7 @@ func (r *Reporter) adminProxy() http.Handler {
 		ErrorLog: adapter.ToStd(logger),
 	}
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		// Block endpoints that terminate or destabilise Envoy.
+		// Block endpoints that terminate or destabilize Envoy.
 		// Read endpoints (/stats, /config_dump, /clusters, etc.) accept
 		// both GET and POST in Envoy's admin API, so we restrict by path
 		// rather than method to preserve operator and test-framework access.


pkg/plugins/runtime/k8s/containers/factory.go
```diff
@@ -120,6 +126,15 @@ func (i *DataplaneProxyFactory) NewContainer(
 		adminPort = i.DefaultAdminPort
 	}
 
+	// When admin UDS is enabled, Envoy admin listens on a Unix socket
+	// instead of TCP. Use the dedicated readiness port for probes since
+	// K8s probes only support TCP/HTTP.
+	probePort := adminPort
+	var probeHost string
```
lahabana: This is never set no?

Automaat: yes, fixed

app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
```diff
@@ -941,6 +941,8 @@ experimental:
   sidecarContainers: false
   # -- If true, inbound tags are not generated for dataplanes. Used with label-based MeshService matching.
   inboundTagsDisabled: false
+  # -- If true, Envoy admin API binds to a Unix domain socket instead of TCP.
+  adminUnixSocket: true
```
bartsmykla: Nit: I think the name should be `envoyAdminUnixSocket`

pkg/config/app/kuma-cp/config.go
```diff
@@ -551,6 +551,12 @@ func (c Config) GetEnvoyAdminPort() uint32 {
 	return c.BootstrapServer.Params.AdminPort
 }
 
+const DefaultReadinessPort = 9902
```
bartsmykla: I don't like the placement of this const. Please check if we can't use some existing const, I would prefer  consts to be defined at the top of the file

pkg/config/app/kuma-cp/config.go
```diff
@@ -551,6 +551,12 @@ func (c Config) GetEnvoyAdminPort() uint32 {
...
+
+func (c Config) GetEnvoyReadinessPort() uint32 {
+	return DefaultReadinessPort
```
bartsmykla: default suggests this value shouldn't be static, which means that this is a bug as the value should be taken from the config or if it's always 9902 instead of using const we should just return this value directly as the const doesn't seem to be reused anywhere else

pkg/xds/bootstrap/generator.go
```diff
@@ -140,6 +143,14 @@ func (b *bootstrapGenerator) Generate(ctx context.Context, request types.Bootstr
 		}
 	}
 
+	if b.adminUnixSocket {
+		if request.Workdir != "" {
+			params.AdminSocketPath = core_xds.AdminSocketName(request.Workdir)
+		} else {
+			log.Info("[WARNING] admin UDS enabled but workdir is empty, falling back to TCP admin", "proxyId", proxyId.String())
```
bartsmykla: Nit: I'm not sure the UDS abbreviation is popular enough so everyone would understand what it stands for, so I would probably use the full unix domain socket instead

test/e2e_env/universal/meshhealthcheck/policy.go
```diff
@@ -556,3 +562,14 @@ spec:
 		})
 	}, Ordered)
 }
+
+func filterLines(s, substr string) string {
+	GinkgoHelper()
```
bartsmykla: Nit: as far as I remember this is unnecessary if you are not using any ginkgo stuff inside the function

PR #15809, opened 2026-03-12
feat(meshzoneaddress): create API definition

lukidzi:
## Motivation

we need a new resource to enable cross zone traffic

## Implementation information

Create API definition

## Supporting documentation

Fix https://github.com/Kong/kong-mesh/issues/9263


lobkovilya (commented):
I think it's missing validation, i.e. address can't be empty, port value has to be in range

lobkovilya (approved):

PR #15811, opened 2026-03-12
feat(meshzoneaddress): create a controller which translate Service

lukidzi:
## Motivation

We introduced https://github.com/kumahq/kuma/pull/15809 api specification based on MADR-096. Based on the MADR we need to create a controller which map Service into MeshZoneAddress to enable cross zone traffic.

## Implementation information

- created K8s controller
  - Watches Services labeled `k8s.kuma.io/zone-proxy-type: ingress`
  - Gates on at least one ready endpoint in EndpointSlice (safe startup/shutdown)
  - Resolves public address in priority order per MADR-096: spec.externalIPs[0] → LoadBalancer (hostname > IP) →
  NodePort (nodes.Items[0], ExternalIP > InternalIP)
  - Sets OwnerReference to the Service for automatic cleanup

## Supporting documentation

Docs: https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/096-ingress-address-sync.md
Fix https://github.com/Kong/kong-mesh/issues/9264


pkg/plugins/runtime/k8s/controllers/meshzoneaddress_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	return func(ctx context.Context, _ kube_client.Object) []kube_ctrl.Request {
+		svcs := &kube_core.ServiceList{}
+		if err := c.List(ctx, svcs, kube_client.MatchingFields{
+			zoneProxyTypeLabelIndex: KumaZoneProxyTypeIngress,
+		}); err != nil {
+			r.Log.Error(err, "failed to list zone-proxy Services on node event")
+			return nil
+		}
+		reqs := make([]kube_ctrl.Request, 0, len(svcs.Items))
+		for _, svc := range svcs.Items {
```
lahabana, line 288: Do this only neeed to happen if the svc in of type NodePort then?

lukidzi: good catch, yes we need only for NodePort. Changed logic to only enqueue services with NodePort Type

lobkovilya: Did you push? I don't see NodePort filtering

lobkovilya (approved):

PR #15843, opened 2026-03-16
feat(k8s): pod to dataplane mapping when using mesh-scoped zone proxy

lukidzi:
## Motivation

As part of MADR-098, the Kubernetes controller needs to translate pods running zone-proxy ingress/egress containers into Dataplane resources with listeners instead of inbound entries. Without this, the new zone-proxy deployment model has no K8s-side automation.

## Implementation information

* New service label: `k8s.kuma.io/zone-proxy-type`
*   `ListenersForService(pod, svc`) mirrors inboundForService but produces listener objects instead of Inbound entries:
    - Resolves container port via FindPort
    - Sets listener State to NotReady if the app container or kuma-sidecar is not ready, or if the pod has a DeletionTimestamp
    - Supports both ZoneIngress and ZoneEgress listener types
* don't generate MeshService for Ingress and Egress

## Supporting documentation

Fix #https://github.com/Kong/kong-mesh/issues/9301


pkg/plugins/runtime/k8s/controllers/listener_converter.go
```diff
@@ -0,0 +1,96 @@
...
+		if cs := util_k8s.FindContainerOrInitContainerStatus(
+			util_k8s.KumaSidecarContainerName,
+			pod.Status.ContainerStatuses,
+			pod.Status.InitContainerStatuses,
+		); cs != nil && !cs.Ready {
+			state = mesh_proto.Dataplane_Networking_Listener_NotReady
+		}
+
+		if pod.DeletionTimestamp != nil {
+			state = mesh_proto.Dataplane_Networking_Listener_NotReady
```
lobkovilya, line 62: the logic that computes the state of the listener probably can be shared with the inbound's logic?

pkg/plugins/runtime/k8s/controllers/testdata/39.services-for-pod.yaml
```diff
@@ -0,0 +1,28 @@
+---
```
lobkovilya, line 1: I'd add few more test cases:
1. service with `k8s.kuma.io/zone-proxy-type` has different values for `port` and `targetPort`
2. port on the pod also has `name`, what will be the final name?
3. 2 services with exactly the same ports select the same pod (check if we handle duplicates correctly)
4. 2 services with exactly the same ports but different zone-proxy type select the same pod (check if we handle conflicts correctly)

pkg/plugins/runtime/k8s/controllers/listener_converter_test.go
```diff
@@ -0,0 +1,140 @@
...
+	kube_core "k8s.io/api/core/v1"
+	kube_meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kube_intstr "k8s.io/apimachinery/pkg/util/intstr"
+
+	mesh_proto "github.com/kumahq/kuma/v2/api/mesh/v1alpha1"
+	. "github.com/kumahq/kuma/v2/pkg/plugins/runtime/k8s/controllers"
+	"github.com/kumahq/kuma/v2/pkg/plugins/runtime/k8s/metadata"
+)
+
+var _ = Describe("ListenersForService", func() {
```
lobkovilya: do you need these tests if you already have golden file based tests (that are easier to review IMO)

pkg/plugins/runtime/k8s/controllers/pod_converter.go
```diff
@@ -299,20 +299,54 @@ func (p *PodConverter) dataplaneFor(
...
+			} else {
+				ifaces, err = p.InboundConverter.LegacyInboundInterfacesFor(ctx, p.Zone, pod, regularServices)
+				if err != nil {
+					return nil, err
+				}
+			}
+			dataplane.Networking.Inbound = ifaces
+		}
+
+		for _, zpSvc := range zoneProxyServices {
```
lobkovilya: we probably need to check if MS are enabled, otherwise the generator will probably produce the config that won't work

pkg/plugins/runtime/k8s/controllers/testdata/43.dataplane.yaml
```diff
@@ -0,0 +1,17 @@
+mesh: default
```
lobkovilya: should be an error in this case https://github.com/kumahq/kuma/blob/247286580a10175b37fc4fa753ea942163e25bac/docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md?plain=1#L184-L185

lobkovilya (approved):

PR #15807, opened 2026-03-11
docs(MADR): otel env bootstrap design

bartsmykla:
## Motivation

The `MeshOpenTelemetryBackend` MADR in #15677 defines the shared backend for traces, logs, and metrics. This MADR answers the next question: how should Kuma reuse standard `OTEL_EXPORTER_OTLP_*` env vars that already exist on the dataplane, from the OTel Operator, systemd, container runtime, or similar sources, without sending secrets through the control plane?

## Implementation information

MADR 100 proposes:

- `kuma-dp` reports detected OTEL env vars to the CP during bootstrap as typed non-secret inventory
- CP merges that inventory with `MeshOpenTelemetryBackend` fields and sends a typed `/otel` runtime plan back to `kuma-dp`
- `env.mode` lets operators set `Disabled`, `Optional`, or `Required`
- `env.precedence` lets operators choose `ExplicitFirst` or `EnvFirst`
- `allowSignalOverrides` controls whether signal-specific env vars such as `OTEL_EXPORTER_OTLP_TRACES_*` may diverge from shared env vars
- Headers, client keys, and certificates stay local to `kuma-dp`
- The model works the same on Kubernetes and Universal
- Status explains whether a signal is ready, blocked, missing, or ambiguous, plus blocked reasons and missing fields

## Supporting documentation

Builds on the `MeshOpenTelemetryBackend` MADR in #15677.

> Changelog: skip

PR #15865, opened 2026-03-17
feat(motb): enable and add status updater

bartsmykla:
## Motivation

PR #15863 added the `MeshOpenTelemetryBackend` resource type with `skip_registration`. This PR enables it and wires up the runtime: observability policies can reference a MOTB via `backendRef` instead of inlining collector addresses, and the CP tracks whether those references resolve.

## Implementation information

Removes `skip_registration` from MOTB so the resource gets registered, CRDs deployed, and the API exposed.

Adds `BackendRef` field to `MeshAccessLog`, `MeshTrace`, and `MeshMetric`. The inline `Endpoint` field is deprecated. Validators enforce that exactly one of `endpoint` or `backendRef` is set, and that `backendRef` has either `name` or `labels` (not both). Deprecation warnings fire when inline endpoints are used.

Status updater runs in the Zone CP on a 5s interval:
- Sets `ReferencedByPolicies` condition on each MOTB (how many policies point at it)
- Sets `BackendRefsResolved` condition on each policy (whether its MOTB references exist)

On conflict, it logs and retries next interval - same as `MeshMultiZoneService` and `Workload` updaters.

Added `.golangci.yml` exclusion for the deprecated `Endpoint` field since existing plugin code still uses it (updated in follow-up xDS PR).

## Supporting documentation

MADR 095: #15677
MADR 100: #15807

> Changelog: feat(kuma-cp): add MeshOpenTelemetryBackend for shared policy-based OpenTelemetry backends


pkg/core/validators/common_validators.go
```diff
@@ -263,6 +265,49 @@ func ValidateLength(path PathBuilder, maxLength int, v string) ValidationError {
...
+	if endpoint == "" && backendRef == nil {
+		verr.AddViolation("", MustHaveExactlyOneOf("openTelemetry", "endpoint", "backendRef"))
+		return verr
+	}
+	if backendRef != nil {
+		verr.AddErrorAt(RootedAt("backendRef"), ValidateBackendResourceRef(backendRef))
+	} else if strings.ContainsAny(endpoint, "/?#") {
+		verr.AddViolationAt(RootedAt("endpoint"), "must be in host:port format, not a URL")
+	} else if _, _, err := net.SplitHostPort(endpoint); err != nil {
+		verr.AddViolationAt(RootedAt("endpoint"), "must be in host:port format")
```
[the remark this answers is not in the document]
bartsmykla: Dropped net.SplitHostPort from the shared validator. MeshTrace and MeshAccessLog accept host-only endpoints (runtime defaults port to 4317). MeshMetric keeps the host:port requirement via an extra validator callback for backward compatibility - it enforced URL format before backendRef existed.

pkg/plugins/policies/meshtrace/api/v1alpha1/deprecated.go
```diff
@@ -2,8 +2,16 @@ package v1alpha1
...
 import (
 	"github.com/kumahq/kuma/v2/pkg/plugins/policies/core/jsonpatch/validators"
+	"github.com/kumahq/kuma/v2/pkg/util/pointer"
 )
 
 func (t *MeshTraceResource) Deprecations() []string {
-	return validators.TopLevelTargetRefDeprecations(t.Spec.TargetRef)
+	var deprecations []string
+	for _, backend := range pointer.Deref(t.Spec.Default.Backends) {
+		if backend.OpenTelemetry != nil && backend.OpenTelemetry.Endpoint != "" && backend.OpenTelemetry.BackendRef == nil {
```
Automaat: we should add deprecations after we release this version, as this will be seen by konnect customers immedaitely, whithout possibility to use new backend

Automaat (commented):
## Critical Issue: Cached resource mutation

The status updater mutates resources fetched from `roResManager.List()` in-place. Per codebase convention (`pkg/core/resources/manager/cache.go`), cached resources are **shared instances that must never be modified**.

**Affected locations in `status_updater.go`:**

1. MOTB status update (~line 140):
```go
motb.Status.Conditions = updatedConditions
```

2. Policy status updates (~lines 290-330):
```go
if mm.Status == nil {
    mm.Status = &meshmetric_api.MeshMetricStatus{}
}
```
(same pattern for MeshTrace and MeshAccessLog)

These mutations corrupt the shared cache and can cause **data races** with other goroutines reading the same cache entries.

**Fix:** Deep-copy the resource before mutation, or re-fetch via `resManager.Get()` before updating status fields.

Note: existing status updaters (MeshMultiZoneService, MeshService) may have the same issue — worth auditing, but shouldn't block fixing it here.

bartsmykla:
MeshMultiZoneService, MeshService, and Workload updaters all do the same - mutate from `roResManager.List()` and pass to `resManager.Update()`, with `store.IsConflict` handling the race. If we fix this it should be done across all updaters at once, not only here.

Automaat (approved):

pkg/plugins/policies/meshtrace/api/v1alpha1/deprecated.go
bartsmykla: Removed all three deprecation warnings in 9ff9898d83. They will be added back in a separate PR after the full MOTB feature lands.

PR #15870, opened 2026-03-18
feat(zoneproxy): implement xds changes and identity 

lukidzi:
## Motivation

This PR implements xDS configuration support for mesh-scoped zone proxies: zone ingress and egress listeners embedded directly in a regular Dataplane resource.

## Implementation information

* New ZoneProxyListenerGenerator (pkg/xds/generator/zone_proxy_listener_generator.go)
  - Generates Envoy listeners, clusters, and endpoints for DataplaneZoneListeners embedded in a regular Dataplane
  - Ingress path: exposes local MeshService/MeshMultiZoneService endpoints via a TLS-inspecting listener
  - Egress path: exposes MeshExternalService endpoints with downstream mTLS (WorkloadIdentity only): no-ops when WorkloadIdentity is absent
* Endpoint map helpers (pkg/xds/topology/outbound.go)
  - `BuildDataplaneZoneIngressEndpointMap`: local MeshService/MeshMultiZoneService endpoints for the ingress listener
  - `BuildDataplaneZoneEgressEndpointMap`: MeshExternalService endpoints reachable from the zone for the egress listener
  - BuildEdsEndpointMap / BuildIngressEndpointMap now accept []xds.ZoneEgress (resolved with SANs) alongside legacy []*ZoneEgressResource; MeshZoneAddress routes take priority over legacy ZoneIngress for zones that have them 
* Generate workload identity and deliver it for Egress
* MeshContext has information about zone egresses and their SAN since we can have multiple egresses and each can have different SAN 
 
## Supporting documentation

Close: https://github.com/Kong/kong-mesh/issues/9265
Close: https://github.com/Kong/kong-mesh/issues/9215
Close: https://github.com/Kong/kong-mesh/issues/9249


pkg/core/naming/contextual.go
```diff
@@ -25,6 +25,14 @@ func MustContextualInboundName[T sectionName](r core_model.Resource, sectionName
 	return name
 }
 
+func ContextualZoneIngressListenerName(listenerName string) string {
```
lobkovilya: nit: arg should be called `portName` not `listenerName` otherwise it's confusing since the function that builds listener name expects the listener name as an input

pkg/core/resources/apis/mesh/dataplane_overview_helpers.go
```diff
@@ -37,6 +37,19 @@ func (t *DataplaneOverviewResource) Status() (Status, []string) {
...
+	if t.Spec.Dataplane.GetNetworking().HasZoneProxyListeners() {
+		networking := t.Spec.Dataplane.GetNetworking()
+		readyListeners := len(networking.GetReadyZoneIngressListeners()) + len(networking.GetReadyZoneEgressListeners())
+		totalListeners := 0
+		for _, l := range networking.GetListeners() {
+			if l.Type == mesh_proto.Dataplane_Networking_Listener_ZoneIngress || l.Type == mesh_proto.Dataplane_Networking_Listener_ZoneEgress {
+				totalListeners++
+			}
+		}
+		allInboundsOffline = readyListeners == 0
```
lobkovilya: you shouldn't reuse `allInboundsOffline` and `allInboundsOnline` variable, you need to create new like `allListeners...` and use it later to compute the DP state

pkg/core/xds/types.go
```diff
@@ -322,15 +333,34 @@ type ZoneEgressProxy struct {
...
+// DataplaneIngressListener holds xDS context for a single embedded ZoneIngress listener.
+type DataplaneIngressListener struct {
+	Listener      *mesh_proto.Dataplane_Networking_Listener
+	MeshResources *MeshProxyResources
+}
+
+// DataplaneEgressListener holds xDS context for a single embedded ZoneEgress listener.
+type DataplaneEgressListener struct {
+	Listener      *mesh_proto.Dataplane_Networking_Listener
+	MeshResources *MeshProxyResources
```
lobkovilya: these types are exactly the same, just keep one `DataplaneListener`

pkg/core/xds/types.go
```diff
@@ -32,6 +32,15 @@ type APIVersion string
 // StreamID represents a stream opened by XDS
 type StreamID = int64
 
+// ZoneEgress holds the resolved bind address, port and workload-identity SAN
+// of a zone egress proxy. SAN is the SPIFFE ID when WorkloadIdentity is
+// enabled: empty for legacy mTLS deployments.
+type ZoneEgress struct {
```
lobkovilya: nit: maybe `ZoneEgressInstance` is a better name (or `ZoneEgressEndpoint`)

pkg/xds/context/mesh_context_builder.go
```diff
@@ -594,6 +600,49 @@ func inferServiceProtocol(endpoints []xds.Endpoint) core_meta.Protocol {
 	return serviceProtocol
 }
 
+func resolveZoneEgresses(
```
lobkovilya, line 625: Maybe it'll be easier to cleanup in the future if we have 2 separate methods:
* `resolveZoneEgresses`
* `resolveLegacyZoneEgresses`

If `resolveZoneEgresses` returns non-empty result then we can skip calling `resolveLegacyZoneEgresses`.

pkg/xds/context/mesh_context_builder.go
```diff
@@ -594,6 +600,49 @@ func inferServiceProtocol(endpoints []xds.Endpoint) core_meta.Protocol {
...
+) []xds.ZoneEgress {
+	var dpEgresses []xds.ZoneEgress
+	for _, dp := range dataplanes {
+		for _, l := range dp.Spec.GetNetworking().GetListeners() {
+			if l.GetType() != mesh_proto.Dataplane_Networking_Listener_ZoneEgress {
+				continue
+			}
+			if l.GetState() != mesh_proto.Dataplane_Networking_Listener_Ready {
+				continue
+			}
```
lobkovilya: you already have `GetReadyZoneEgressListeners`, probably can use it instead

pkg/xds/context/mesh_context_builder.go
```diff
@@ -594,6 +600,49 @@ func inferServiceProtocol(endpoints []xds.Endpoint) core_meta.Protocol {
...
+				if trustDomain, err := identity.Spec.GetTrustDomain(identity.GetMeta(), zone); err != nil {
+					logger.Error(err, "failed to compute trust domain for zone egress", "dataplane", dp.GetMeta().GetName())
+				} else if spiffeID, err := identity.Spec.GetSpiffeID(trustDomain, dp.GetMeta(), env); err != nil {
+					logger.Error(err, "failed to compute SPIFFE ID for zone egress", "dataplane", dp.GetMeta().GetName())
+				} else {
+					ze.SAN = spiffeID
+				}
+			}
+			dpEgresses = append(dpEgresses, ze)
+		}
```
lobkovilya: Do we need to do it here because there is no `MeshService` alternative for zone egress?

pkg/xds/context/mesh_context_builder.go
```diff
@@ -594,6 +600,49 @@ func inferServiceProtocol(endpoints []xds.Endpoint) core_meta.Protocol {
...
+	for _, dp := range dataplanes {
+		for _, l := range dp.Spec.GetNetworking().GetListeners() {
+			if l.GetType() != mesh_proto.Dataplane_Networking_Listener_ZoneEgress {
+				continue
+			}
+			if l.GetState() != mesh_proto.Dataplane_Networking_Listener_Ready {
+				continue
+			}
+			ze := xds.ZoneEgress{Address: l.GetAddress(), Port: l.GetPort()}
+			if identity, ok := meshidentity_api.BestMatched(dp.GetMeta().GetLabels(), identities); ok {
```
lobkovilya: IIUC identity is per DPP (not per listener) so you can move the code to the loop over the `dataplanes` (not loop over the listeners)

pkg/xds/topology/outbound.go
```diff
@@ -74,6 +75,48 @@ func BuildEgressEndpointMap(
...
 
+// BuildDataplaneZoneEgressEndpointMap builds endpoints only for MeshExternalServices reachable from the zone.
+// Used for embedded zone egress listeners in a Dataplane resource.
+func BuildDataplaneZoneEgressEndpointMap(
+	ctx context.Context,
+	mesh *core_mesh.MeshResource,
+	zone string,
+	meshExternalServices []*meshexternalservice_api.MeshExternalServiceResource,
+	loader datasource.Loader,
+	unifiedNaming bool,
```
lobkovilya: can probably remove `unifiedNaming` arg

pkg/xds/topology/outbound.go
```diff
@@ -74,6 +75,48 @@ func BuildEgressEndpointMap(
...
+) core_xds.EndpointMap {
+	outbound := core_xds.EndpointMap{}
+	for _, mes := range meshExternalServices {
+		if mes.IsReachableFromZone(zone) {
```
lobkovilya: but how can you have MES in the zone that's not reachable from that zone...

pkg/xds/sync/dataplane_proxy_builder.go
```diff
@@ -86,6 +90,57 @@ func (p *DataplaneProxyBuilder) Build(ctx context.Context, key core_model.Resour
...
+		EndpointMap: xds_topology.BuildDataplaneZoneEgressEndpointMap(
+			ctx,
+			meshContext.Resource,
+			p.Zone,
+			meshContext.Resources.MeshExternalServices().Items,
+			meshContext.DataSourceLoader,
+			unifiedNaming,
+		),
+		Resources: meshContext.Resources.MeshLocalResources,
+	}
```
lobkovilya: can't you store these maps in the meshcontext? why do we need to build them for every DPP?

pkg/xds/context/mesh_context_builder.go
lukidzi: Yes, as we don't have a MeshService for egress to hold it

pkg/xds/context/mesh_context_builder.go
```diff
@@ -188,6 +189,10 @@ func (m *meshContextBuilder) BuildIfChanged(ctx context.Context, meshName string
 	zoneIngresses := resources.ZoneIngresses().Items
 	zoneEgresses := resources.ZoneEgresses().Items
 	externalServices := resources.ExternalServices().Items
+	zoneEgressList := resolveZoneEgresses(dataplanes, resources.MeshIdentities().Items, m.zone)
+	if len(zoneEgressList) == 0 {
```
slonka, line 193: I'm assuming we're not mixing legacy and new because mixing them in one pool would cause TLS failures - right? Is it worth a comment here?

test/server/cmd/echo.go
```diff
@@ -81,9 +80,7 @@ func newEchoHTTPCmd() *cobra.Command {
 				return err
 			}
 			sdkmetric.NewMeterProvider(sdkmetric.WithReader(promExporter))
-			promHandler := promhttp.HandlerFor(prom_client.DefaultGatherer, promhttp.HandlerOpts{
-				ErrorHandling: promhttp.ContinueOnError,
-			})
+			promHandler := promhttp.Handler()
```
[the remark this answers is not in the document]
lobkovilya: @lukidzi do you remember if there was a reason to change it?

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -54,18 +54,34 @@ func GenerateClusters(
...
-							false,
-						))
+					sni := SniForBackendRef(service.BackendRef().RealResourceBackendRef(), meshCtx, systemNamespace)
+					if proxy.WorkloadIdentity != nil {
+						// MeshExternalService is only routable through zone egress in workload-identity mode.
+						// Skip cluster generation if no zone egress SANs are available.
+						egressSANs := meshCtx.ZoneEgressSANs()
+						if len(egressSANs) == 0 {
+							continue
+						}
```
slonka, line 64: If WorkloadIdentity is on but `ZoneEgressSANs()` returns empty, every MeshExternalService cluster gets silently skipped. `resolveZoneEgresses` already logs errors for trust domain / SPIFFE ID failures. On the client side I don't think there's anything connecting those CP logs to the missing clusters. Is it worth adding a warning here?

pkg/core/resources/apis/mesh/dataplane_overview_helpers.go
```diff
@@ -37,6 +37,26 @@ func (t *DataplaneOverviewResource) Status() (Status, []string) {
...
+
+		switch {
+		case !proxyOnline || readyListeners == 0:
+			return Offline, errs
+		case readyListeners < totalListeners:
+			return PartiallyDegraded, errs
+		default:
+			return Online, nil
+		}
+	}
```
slonka, line 51: When a Dataplane has both inbounds and zone proxy listeners, the `HasZoneProxyListeners()` block returns early based on listener readiness alone. Inbound errors accumulated in the loop above get ignored. All listeners Ready + some inbounds NotReady == Online (instead of maybe PartiallyDegraded).

Is this intended?

pkg/xds/generator/zone_proxy_listener_generator.go
```diff
@@ -0,0 +1,350 @@
...
+		Configure(envoy_listeners.TLSInspector())
+
+	downstreamTLS, err := meshIdentityDownstreamTLS(proxy)
+	if err != nil {
+		return nil, err
+	}
+
+	addedFilterChains := 0
+	for _, dst := range destinations {
+		esPort := dst.GetPorts()[0]
```
lukidzi: Can it be empty list? Do we need to do a check here?

pkg/xds/generator/zone_proxy_listener_generator.go
```diff
@@ -0,0 +1,350 @@
...
+		Configure(envoy_listeners.StatPrefix(listenerName)).
+		Configure(envoy_listeners.TLSInspector())
+
+	meshResources := xds_context.Resources{MeshLocalResources: xdsCtx.Mesh.Resources.MeshLocalResources}
+
+	// Only expose services local to this zone.
+	localMS := &meshservice_api.MeshServiceResourceList{}
+	for _, ms := range meshResources.MeshServices().GetItems() {
+		if lbls := ms.GetMeta().GetLabels(); lbls == nil || lbls[mesh_proto.ZoneTag] == "" || lbls[mesh_proto.ZoneTag] == cp.Zone {
+			_ = localMS.AddItem(ms)
```
lukidzi: Do we need to handle the error?

pkg/xds/generator/zone_proxy_listener_generator.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	rs.AddSet(eds)
+
+	for _, cluster := range clusters {
+		listenerBuilder.Configure(envoy_listeners.FilterChain(zoneproxy.CreateFilterChain(proxy, cluster)))
+	}
+
+	if len(clusters) == 0 {
+		return nil, nil
+	}
```
lukidzi: Maybe we can move 
```
	if len(clusters) == 0 {
		return nil, nil
	}
```
we can move it after `clusters := services.Clusters()` so we don't need to generate everything if clusters are empty

pkg/xds/generator/zone_proxy_listener_generator.go
```diff
@@ -0,0 +1,350 @@
...
+		Configure(envoy_listeners.InboundListener(address, port, core_xds.SocketAddressProtocolTCP)).
+		Configure(envoy_listeners.StatPrefix(listenerName)).
+		Configure(envoy_listeners.TLSInspector())
+
+	meshResources := xds_context.Resources{MeshLocalResources: xdsCtx.Mesh.Resources.MeshLocalResources}
+
+	// Only expose services local to this zone.
+	localMS := &meshservice_api.MeshServiceResourceList{}
+	for _, ms := range meshResources.MeshServices().GetItems() {
+		if lbls := ms.GetMeta().GetLabels(); lbls == nil || lbls[mesh_proto.ZoneTag] == "" || lbls[mesh_proto.ZoneTag] == cp.Zone {
```
slonka: MeshService already has `IsLocalMeshService()` but it checks origin, not zone. Could we add something like `IsLocalToZone(zone string) bool` that combines both checks? Would make this condition more readable. WDYT?

pkg/xds/context/mesh_context_builder.go
```diff
@@ -594,6 +619,47 @@ func inferServiceProtocol(endpoints []xds.Endpoint) core_meta.Protocol {
...
+		listeners := dp.Spec.GetNetworking().GetReadyZoneEgressListeners()
+		if len(listeners) == 0 {
+			continue
+		}
+		var san string
+		if identity, ok := meshidentity_api.BestMatched(dp.GetMeta().GetLabels(), identities); ok {
+			env := config_core.UniversalEnvironment
+			if _, isK8s := dp.GetMeta().GetLabels()[mesh_proto.KubeNamespaceTag]; isK8s {
+				env = config_core.KubernetesEnvironment
+			}
```
lukidzi, line 641: Should we pass env directly from config since user potentially could set this label (unless we override it)

pkg/xds/generator/zone_proxy_listener_generator_test.go
```diff
@@ -0,0 +1,358 @@
...
+						},
+					},
+					WorkloadIdentity:  testWorkloadIdentity,
+					InternalAddresses: DummyInternalAddresses,
+				},
+				meshContext: xds_context.MeshContext{
+					Resource: builders.Mesh().WithName("default").WithMeshServicesEnabled(mesh_proto.Mesh_MeshServices_Exclusive).Build(),
+					DataplaneZoneEgressEndpointMap: core_xds.EgressEndpointMap{
+						unifiedSvcName: core_xds.EgressEndpointGroup{
+							Protocol:      core_meta.ProtocolHTTP,
```
slonka, line 284: are we testing TCP anywhere?

test/e2e_env/multizone/meshproxy/connectivity.go
```diff
@@ -0,0 +1,367 @@
...
+			Expect(NewClusterSetup().Install(YamlUniversal(yaml)).Setup(multizone.Global)).To(Succeed())
+		}
+
+		trustZone1 := getMeshTrust(multizone.KubeZone1.Name(), Config.KumaNamespace)
+		installTrustToGlobal(trustZone1, multizone.KubeZone1.Name())
+
+		trustZone2 := getMeshTrust(multizone.KubeZone2.Name(), Config.KumaNamespace)
+		installTrustToGlobal(trustZone2, multizone.KubeZone2.Name())
+
+		trustZone4 := getMeshTrust(multizone.UniZone1.Name())
```
slonka, line 236: did we miss zone3?

test/e2e_env/multizone/meshproxy/connectivity.go
```diff
@@ -0,0 +1,367 @@
...
+		Expect(multizone.KubeZone1.TriggerDeleteNamespace(namespace)).To(Succeed())
+		Expect(multizone.KubeZone1.TriggerDeleteNamespace(externalNamespace)).To(Succeed())
+		Expect(multizone.KubeZone2.TriggerDeleteNamespace(namespace)).To(Succeed())
+		Expect(multizone.UniZone1.DeleteMeshApps(meshName)).To(Succeed())
+		Expect(multizone.Global.DeleteMesh(meshName)).To(Succeed())
+	})
+
+	It("should route cross-zone traffic via new zone ingress proxies and record metrics on zone-proxy-ingress", func() {
+		uniIngressFilter := fmt.Sprintf(
+			"cluster.kri_msvc_%s_%s__test-server_80.upstream_rq_active",
```
slonka: `upstream_rq_active` is a gauge - it'll be 0 if the request finishes before the stats check. The egress test uses `upstream_rq_200` (a counter), which won't have this problem. Should ingress use a counter too?

test/framework/k8s_cluster.go
```diff
@@ -255,6 +255,20 @@ func (c *K8sCluster) GetZoneIngressEnvoyTunnel() envoy_admin.Tunnel {
 	return tnl
 }
 
+// GetEnvoyAdminTunnel creates or returns an Envoy admin tunnel for any named
+// app in a given namespace.
+func (c *K8sCluster) GetEnvoyAdminTunnel(appName, namespace string) envoy_admin.Tunnel {
```
slonka, line 244: `GetZoneEgressEnvoyTunnel` and `GetZoneIngressEnvoyTunnel` right above do the same thing with hardcoded values. Could be one-liners wrapping this. Could be a separate PR.

test/framework/universal_cluster.go
```diff
@@ -345,6 +345,29 @@ func (c *UniversalCluster) CreateZoneEgress(
...
+	}
+	c.apps[name] = app
+	c.networking[name] = app.universalNetworking
+	c.createEnvoyTunnel(name)
+	return app.dpApp.Start()
+}
+
+// GetAppEnvoyTunnel returns the Envoy admin tunnel for a named app (e.g. a
+// zone proxy deployed via CreateDataplaneProxy).
+func (c *UniversalCluster) GetAppEnvoyTunnel(name string) envoy_admin.Tunnel {
```
slonka, line 363: probably similar situation as in `test/framework/k8s_cluster.go`

test/server/cmd/echo.go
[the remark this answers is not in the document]
slonka: I have the same question.

slonka (commented):
This is quite a big PR, I tried to review this as carefully as I can. Nothing obviously wrong stands out maybe apart from https://github.com/kumahq/kuma/pull/15870#discussion_r3009124453 .

Mostly questions / small things, if you *we* conclude that there's nothing that needs to change I'll change this to 👍

test/server/cmd/echo.go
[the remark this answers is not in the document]
lobkovilya: Reverted the change

pkg/xds/generator/zone_proxy_listener_generator.go
lobkovilya: It can't because of the API validation, but I added a check with the comment that it can't happen

pkg/xds/generator/zone_proxy_listener_generator.go
lobkovilya: Moved

pkg/xds/context/mesh_context_builder.go
lobkovilya: Added a comment

test/e2e_env/multizone/meshproxy/connectivity.go
lobkovilya: Good point, we shouldn't check gauge. We can't check `upstream_rq_200` either because ingress is a TCP proxy. But I've changed it to `rq_total`.

pkg/xds/generator/zone_proxy_listener_generator.go
lobkovilya: yep, why not

pkg/xds/context/mesh_context_builder.go
lobkovilya: I don't want to pass it, I'll make sure user can't set it in another issue I'm working on https://github.com/kumahq/kuma/issues/16049

pkg/core/resources/apis/mesh/dataplane_overview_helpers.go
lobkovilya: Right, I refactored the `Status` func to handle these cases.

pkg/xds/generator/zone_proxy_listener_generator.go
lobkovilya: We can use `IsLocalMeshService()` as is, no need to check the zone. I updated the comment for `IsLocalMeshService()` to explain why we can do it.

pkg/xds/generator/zone_proxy_listener_generator_test.go
lobkovilya: added tcp unit test

test/e2e_env/multizone/meshproxy/connectivity.go
lobkovilya: No, it's just weird naming, kuma-1 and kuma-2 are the names of k3d clusters, kuma-3 we're using for global and kuma-4 and kuma-5 are universal zones.

pkg/plugins/policies/core/xds/meshroute/clusters.go
lobkovilya: We can't log from generator, it'll be very noisy. It probably requires similar handling in controller as https://github.com/kumahq/kuma/pull/16014, I'll open an issue

slonka (approved):

PR #15963, opened 2026-03-22
build(make): harden target graph

bartsmykla:
> [!WARNING]
> WIP

## Motivation

Fix the GNU make issues by making prerequisite edges explicit and by removing parse-time side effects from dev targets

## Implementation information

- move proto exports, Docker save/load, docs generation, e2e startup, and transparent proxy setup onto explicit prerequisites or file targets
- preserve legacy target aliases and the legacy shared kubeconfig path on top of tool-specific kubeconfigs
- normalize the k3d deployment volume path for absolute `KUMA_DIR` callers and keep existing local workflows working


lahabana:
I'm a little torn on that one. Do we have evidence of something not working in the makefile?
I feel like these have been like this for years so I'm wary of changing this.

The PR description mention "GNU make issues", where did we encounter these? Are they causing flakes in CI?

PR #15997, opened 2026-03-25
fix(xds): use listener name as vhost fallback

Automaat:
## Motivation

When deploying on Universal without transparent proxy and without
defining `kuma.io/service` on an HTTP inbound, the control plane
fails with:

```
virtual host name is required, but it was not provided
```

`FilterChainBuilder` derives `virtualHostName` from the service tag.
With no tag present and unified naming disabled, the name is an empty
string, which the virtual host builder rejects.

Fixes #15580

## Implementation information

When `virtualHostName` is empty (service tag absent, unified naming
off), fall back to `inbound:IP:port` — the inbound listener name —
for both `routeConfigName` and `virtualHostName`. This is always
unique and non-empty.

Added test case 10: HTTP inbound without `kuma.io/service` tag,
with golden file confirming the listener is generated correctly.

> Changelog: fix(xds): use listener name as vhost fallback when kuma.io/service absent

PR #16001, opened 2026-03-25
test(e2e): use e2e tests for legacy tests

lukidzi:
## Motivation

Seems we were incorrectly using Kube tests for E2E target instead of e2e package tests.

## Implementation information

Tests fixes:
* use ContainerPatch to remove CPU limit on kuma-init. Noticed that CPU is throttled and later kuma cannot install transparent proxy. Removing the limit helps in a low resource env to start everything and makes test much more stable
* added isCA as was missing and causing problem since the cert couldn't be created for a DP based on the provided Cert 
* increased clusters retries and timeout
* remove cert-manager test which didn't make sense
* added KUMA_RUNTIME_KUBERNETES_INJECTOR_BUILTIN_DNS_EXPERIMENTAL_PROXY to corefile test as we don't use CoreDNS by the default
* `test/e2e/helm/kuma_helm_upgrade_multizone.go` added fetching ZoneIngresses from the endpoint for version < 2.11 since kumactl since 2.11 doesn't support old endpoint
* `skipinboundtags` test was incorrect as it didn't work with legacy mTLS - moved to use MeshIdentity
* moved CNI test to multizone setup as we run Kuma2 cluster with CNI
* removed ExternalService test as this functionality seems to be tested in zoneegress/externalservice


mk/e2e.new.mk
```diff
@@ -127,7 +127,7 @@ test/e2e: $(E2E_DEPS_TARGETS) $(E2E_K8S_BIN_DEPS) ## Run slower e2e tests (slowe
 	$(MAKE) test/e2e/k8s/start/cluster/kuma-1
 	$(MAKE) test/e2e/k8s/wait/kuma-1
 	$(MAKE) test/e2e/k8s/load/images/kuma-1
-	$(E2E_ENV_VARS) $(GINKGO_TEST_E2E) $(KUBE_E2E_PKG_LIST) || (ret=$$?; $(MAKE) test/e2e/k8s/stop/cluster/kuma-1 && exit $$ret)
+	$(E2E_ENV_VARS) $(GINKGO_TEST_E2E) $(E2E_PKG_LIST) || (ret=$$?; $(MAKE) test/e2e/k8s/stop/cluster/kuma-1 && exit $$ret)
 	$(MAKE) test/e2e/k8s/stop/cluster/kuma-1
```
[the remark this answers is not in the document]
bartsmykla: Fixed - switched to test/e2e/k8s/start and test/e2e/k8s/stop which handle all K8SCLUSTERS.

mk/e2e.new.mk
```diff
@@ -127,7 +127,7 @@ test/e2e: $(E2E_DEPS_TARGETS) $(E2E_K8S_BIN_DEPS) ## Run slower e2e tests (slowe
 	$(MAKE) test/e2e/k8s/start/cluster/kuma-1
 	$(MAKE) test/e2e/k8s/wait/kuma-1
 	$(MAKE) test/e2e/k8s/load/images/kuma-1
-	$(E2E_ENV_VARS) $(GINKGO_TEST_E2E) $(KUBE_E2E_PKG_LIST) || (ret=$$?; $(MAKE) test/e2e/k8s/stop/cluster/kuma-1 && exit $$ret)
+	$(E2E_ENV_VARS) $(GINKGO_TEST_E2E) $(E2E_PKG_LIST) || (ret=$$?; $(MAKE) test/e2e/k8s/stop/cluster/kuma-1 && exit $$ret)
```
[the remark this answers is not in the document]
bartsmykla: Switched from ls to wildcard to only match directories.

lahabana:
Please add `--fail-on-empty` [docs](https://onsi.github.io/ginkgo/#avoiding-filtering-out-all-tests) to https://github.com/kumahq/kuma/blob/a71d0f97d901d9cb894bdfd2584518d2aefb03c4/mk/test.mk#L23  so this doesn't happen again.

mk/e2e.new.mk
```diff
@@ -123,12 +123,17 @@ test/e2e/debug-universal:
 
 .PHONY: test/e2e
 test/e2e: $(E2E_DEPS_TARGETS) $(E2E_K8S_BIN_DEPS) ## Run slower e2e tests (slower as they start and stop Kuma for each tests). Use DEBUG=1 to more easily find issues
+	@# Can't use --fail-on-empty here: test/e2e runs multiple suites with
+	@# --label-filter and suites that have no specs for a given label (e.g.
+	@# skipinboundtags on job-0) would fail even though that's expected.
```
lahabana, line 129: Are they a lot of these? If that's the only case what about moving one of the skipinbound to job-0?

test/e2e/helm/kuma_helm_upgrade_multizone.go
```diff
@@ -144,8 +144,11 @@ spec:
 				Setup(zoneUniversal)
 			Expect(err).ToNot(HaveOccurred())
 
+			// Query global (already upgraded) instead of zoneK8s (still old version)
+			// because old CPs don't have the /zoneingresses endpoint that the
+			// current kumactl uses (renamed from /zone-ingresses in 2.11.0).
 			Eventually(func(g Gomega) (int, error) {
-				return NumberOfResources(zoneK8s, mesh.ZoneIngressResourceTypeDescriptor)
+				return NumberOfResources(global, mesh.ZoneIngressResourceTypeDescriptor)
 			}, "30s", "1s").Should(Equal(2), "have remote and local zoneIngress")
```
lahabana: This defeats the test a little bit no? Because what' this is checking is that stuff got propagated no?
Can't we just do a if on the version instead?

lahabana:
> ginkgo --procs 2
> runs them in parallel causing helm release and namespace collisions.

How did this used to work? AFAIR these always required to be run 1 by 1. Or is this is only the case for a few tests?

mk/e2e.new.mk
lukidzi: We have a lot of thems, each job runs different subset of tests

test/e2e/webhooks/cert_manager.go
```diff
@@ -246,88 +246,6 @@ spec:
 		)
 		Expect(err).ToNot(HaveOccurred())
 	})
-
-	It("should continue working after certificate rotation", func() {
```
lukidzi, line 250: This test makes no sense, as we don't live reload CP cert

test/e2e/skipinboundtags/e2e_suite_test.go
```diff
@@ -17,24 +17,31 @@ func TestE2E(t *testing.T) {
...
+	if skipinboundtags.KubeCluster == nil {
+		return
+	}
 	Expect(skipinboundtags.KubeCluster.DeleteKuma()).To(Succeed())
 	Expect(skipinboundtags.KubeCluster.DismissCluster()).To(Succeed())
 })
 
 var (
 	_ = ReportAfterSuite("report suite", report.DumpReport)
-	_ = Describe("Skip Inbound Tags on Kubernetes", Label("job-3"), skipinboundtags.SkipInboundTags, Ordered)
```
lukidzi, line 39: Needs https://github.com/kumahq/kuma/pull/15997

test/e2e/helm/kuma_helm_upgrade_multizone.go
lukidzi: Fixed this

PR #16019, opened 2026-03-27
fix(kri): fix lookup for empty zone/ns KRIs

### LINKED ISSUE #15544: Generation of KRI does not use default fallbacks
### Kuma Version

### Describe the bug

### Without a namespace

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  name: without-namespace
  labels:
    kuma.io/mesh: default
spec:
  targetRef:
    kind: Mesh
  rules:
    - default:
        allow:
          - spiffeID:
              type: Prefix
          - spiffeID:
              type: Prefix
```

### With a namespace

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTrafficPermission
metadata:
  name: with-namespace
  namespace: kuma-demo
  labels:
    kuma.io/mesh: default
spec:
  targetRef:
    kind: Mesh
  rules:
    - default:
        allow:
          - spiffeID:
              type: Prefix
          - spiffeID:
              type: Prefix
```

### To Reproduce

### Expected behavior

_No response_

Automaat:
## Motivation

KRI lookup via `/_kri/{kri}` returns 404 for resources without zone/namespace labels (e.g. `kri_mtp_default___name_`). The previous fix (#15921) worked around this by injecting CP zone and system namespace as defaults into KRI strings in REST API responses. Per KRI spec, empty slots must remain stable — this approach was incorrect and caused phantom resources to appear in GUIs when defaults referenced non-existent zones.

Fixes #15544

## Implementation information

Two-part fix:

**1. Revert #15921 KRI defaults injection**
- Remove `WithDefaultZone`, `WithDefaultNamespace`, `Option` from `pkg/core/kri/kri.go`
- Remove `SetKRIDefaults`, `kriDefaultZone`, `kriDefaultNamespace`, `kriOptions` from `ResourceMeta`
- Remove `ResourceWithKRIDefaults`, `ResourceListWithKRIDefaults`, `setKRIDefaults` from converter
- Revert `formatResource` signature and all list/single-resource endpoints to use `rest.From.Resource`/`ResourceList`
- KRI strings in REST responses now correctly use empty slots per spec

**2. Fix the actual bug — hash mismatch in `getCoreName`**

Root cause: `getCoreName` computed hashes using `hash.HashedName(mesh, name, zone, namespace)` always passing zone and namespace as arguments, even when empty. `HashSuffixMapper` in KDS sync only includes label values that actually exist on the resource (skipping absent labels). While FNV hash treats empty string writes as no-ops (so `hash(mesh, name, "", ns)` == `hash(mesh, name, ns)`), the code now explicitly matches `HashSuffixMapper` behavior: only non-empty values are passed to `HashedName`.

`zone=""` always means the resource originated on Global CP, regardless of which CP is reading it. There is no ambiguity — on Zone CP, locally created resources always have `zone==cpZone`.

> Changelog: fix(kri): fix KRI lookup for resources without zone/namespace labels

### DIFF
--- a/pkg/api-server/kri_endpoint.go
+++ b/pkg/api-server/kri_endpoint.go
@@ -49,28 +49,20 @@ func (k *kriEndpoint) findByKriRoute() restful.RouteFunction {
 			return
 		}
 
-		var resource core_model.Resource
-		for _, name := range k.getCoreNames(identifier) {
-			if err := k.resourceAccess.ValidateGet(
-				request.Request.Context(),
-				core_model.ResourceKey{Mesh: identifier.Mesh, Name: name},
-				*descriptor,
-				user.FromCtx(request.Request.Context()),
-			); err != nil {
-				rest_errors.HandleError(request.Request.Context(), response, err, "Access Denied")
-				return
-			}
-			r := descriptor.NewObject()
-			if err := k.resManager.Get(request.Request.Context(), r, store.GetByKey(name, identifier.Mesh)); err == nil {
-				resource = r
-				break
-			} else if !store.IsNotFound(err) {
-				rest_errors.HandleError(request.Request.Context(), response, err, "Could not retrieve a resource")
-				return
-			}
+		name := k.getCoreName(identifier)
+		if err := k.resourceAccess.ValidateGet(
+			request.Request.Context(),
+			core_model.ResourceKey{Mesh: identifier.Mesh, Name: name},
+			*descriptor,
+			user.FromCtx(request.Request.Context()),
+		); err != nil {
+			rest_errors.HandleError(request.Request.Context(), response, err, "Access Denied")
+			return
 		}
-		if resource == nil {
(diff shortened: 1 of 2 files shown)


pkg/api-server/kri_endpoint.go
```diff
@@ -49,25 +49,32 @@ func (k *kriEndpoint) findByKriRoute() restful.RouteFunction {
...
+				request.Request.Context(),
+				core_model.ResourceKey{Mesh: identifier.Mesh, Name: name},
+				*descriptor,
+				user.FromCtx(request.Request.Context()),
+			); err != nil {
+				rest_errors.HandleError(request.Request.Context(), response, err, "Access Denied")
+				return
+			}
+			r := descriptor.NewObject()
+			if err = k.resManager.Get(request.Request.Context(), r, store.GetByKey(name, identifier.Mesh)); err == nil {
```
bartsmykla: Nit
```suggestion
			if err := k.resManager.Get(request.Request.Context(), r, store.GetByKey(name, identifier.Mesh)); err == nil {
```

Automaat:
Split into two PRs: kumahq/kuma#16046 (revert) and kumahq/kuma#16047 (fix)

PR #16024, opened 2026-03-27
feat(universal): allow disabling Workload and MeshService generators

lahabana:
## Motivation

Users may want to manage them manually. This wasn't possible. You can now set `KUMA_MESH_SERVICE_GENERATION_INTERVAL=0` or `KUMA_RUNTIME_UNIVERSAL_WORKLOAD_GENERATION_INTERVAL=0` which turns these off.

## Implementation information

Log that these are manually manage and simply don't start the component.
Not starting the component is the behaviour we follow when running global or Kubernetes so it's the best suited way to do this

### DIFF
--- a/pkg/config/app/kuma-cp/config.go
+++ b/pkg/config/app/kuma-cp/config.go
@@ -564,8 +564,8 @@ func (i MeshMultiZoneServiceIPAM) Validate() error {
 }
 
 type MeshServiceConfig struct {
-	// How often we check whether MeshServices need to be generated from
-	// Dataplanes
+	// How often we check whether MeshServices need to be
+	// generated from Dataplanes set to 0 to disable automatic workload management (only applies to universal)
 	GenerationInterval config_types.Duration `json:"generationInterval" envconfig:"KUMA_MESH_SERVICE_GENERATION_INTERVAL"`
 	// How long we wait before deleting a MeshService if all Dataplanes are gone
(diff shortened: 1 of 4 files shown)


lahabana:
xref #16020

pkg/config/app/kuma-cp/config.go
```diff
@@ -564,8 +564,8 @@ func (i MeshMultiZoneServiceIPAM) Validate() error {
 }
 
 type MeshServiceConfig struct {
-	// How often we check whether MeshServices need to be generated from
-	// Dataplanes
+	// How often we check whether MeshServices need to be
+	// generated from Dataplanes set to 0 to disable automatic workload management (only applies to universal)
```
Automaat: Wouldn't it be better to just have a separate config flag for disabling generators? with this we have a config which can be set on both k8s and univesal and will take effect only on universal

What will happen on k8s if i set this to 0?

lahabana: This just doesn't run on k8s right?

Automaat: Right, these are separate paths, so i Think this is ok then

lahabana:
```suggestion
	// generated from Dataplanes set to 0 to disable automatic meshservice management (only applies to universal)
```

Automaat (approved):

PR #15962, opened 2026-03-22
chore(ci): fix errorf and use-errors-new issues from revive

mmorel-35:
## Motivation

fix [errorf](https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md#errorf) and [use-errors-new](https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md#use-errors-new) issues from revive

## Supporting documentation

Fix #XX

PR #16033, opened 2026-03-29
chore(test): use fmt.Errorf instead of errors.Errorf part 1

mmorel-35:
## Motivation

use fmt.Errorf instead of errors.Errorf in test

Related to https://github.com/kumahq/kuma/pull/15962#pullrequestreview-3997348533

## Supporting documentation

Fix #XX


lahabana:
Sorry @mmorel-35 but these PRs are creating more noise than they are bringing value. Closing them.

PR #16052, opened 2026-03-30
feat(metrics): add cp metrics OTLP push

Automaat:
## Motivation

The CP has internal metrics (xDS generation, store ops, KDS sync, resource counts, gRPC, Go runtime) exposed via Prometheus `/metrics` on port 5680. OTel-collector-based observability stacks require a Prometheus scraper to consume them. This adds native OTLP push so CP metrics flow directly to an OTel collector.

Resolves https://github.com/Kong/kong-mesh/issues/9159

## Implementation information

Uses the OTel prometheus bridge (`go.opentelemetry.io/contrib/bridges/prometheus`) to read from the existing CP `prometheus.Gatherer` and push via `otlpmetricgrpc` with a `PeriodicReader`. The pusher runs as a `ResilientComponent` in the existing OTel runtime plugin.

- Auto-enabled when `OTEL_EXPORTER_OTLP_ENDPOINT` is set — no Kuma-specific config needed
- Endpoint/auth/TLS: standard `OTEL_EXPORTER_OTLP_*` env vars
- Export interval: `OTEL_METRIC_EXPORT_INTERVAL` (default 60s, read by OTel SDK)
- Export timeout: `OTEL_METRIC_EXPORT_TIMEOUT` (default 30s, read by OTel SDK)
- Existing `/metrics` scraping unchanged
- Handles collector unavailability gracefully (resilient component with backoff)

> Changelog: feat(kuma-cp): improve control-plane observability with dashboards, histograms, and operational metrics


pkg/config/app/kuma-cp/config.go
```diff
@@ -116,6 +117,21 @@ type ControlPlaneMetrics struct {
 	ReportResourcesCount bool `json:"reportResourcesCount" envconfig:"kuma_metrics_control_plane_report_resources_count"`
 }
 
+type MetricsOpenTelemetry struct {
+	config.BaseConfig
+
+	// Enabled if true will push control plane metrics to an OTel collector via OTLP.
+	// Configure the collector endpoint via standard OTEL_EXPORTER_OTLP_* environment variables.
+	// Default: false
+	Enabled bool `json:"enabled" envconfig:"kuma_metrics_open_telemetry_enabled"`
```
lahabana: Do we need an enabled flag? Can we just rely on the env vars?

Automaat: changed to use env vars

pkg/plugins/runtime/opentelemetry/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+
+	m.log.Info("OTLP metrics push started")
+	<-stop
+	m.log.Info("stopping OTLP metrics push")
+
+	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer shutdownCancel()
+
+	if shutdownErr := provider.Shutdown(shutdownCtx); shutdownErr != nil {
+		return fmt.Errorf("shutting down OTLP metrics provider: %w", shutdownErr)
```
slonka: **issue:** `Start()` returns shutdown errors, which causes ResilientComponent to restart the pusher after the stop signal. It's inconsistent with the tracer right next to it. The tracer (plugin.go:50-51) logs shutdown errors and returns nil.

I wrote a [POC branch](https://github.com/automaat/kuma/compare/feat/cp-otel-metrics-push...slonka:kuma:worktree-poc-metrics-leak-16052) with tests confirming this. Each restart cycle churns through a new exporter, PeriodicReader, and MeterProvider.

```suggestion
		m.log.Error(shutdownErr, "shutting down OTLP metrics provider")
```

pkg/plugins/runtime/opentelemetry/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	defer cancel()
+
+	exporter, err := otlpmetricgrpc.New(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create OTLP metric exporter: %w", err)
+	}
+
+	bridge := prombridge.NewMetricProducer(prombridge.WithGatherer(m.gatherer))
+
+	reader := sdkmetric.NewPeriodicReader(exporter, sdkmetric.WithProducer(bridge))
```
slonka, line 40: **issue**: Export failures bypass structured logging. When the collector is unreachable, `PeriodicReader` calls `otel.Handle(err)`, which defaults to `log.Printf` on stderr. No JSON format, no component name. Log aggregation won't pick these up. Diagnosing "metrics aren't arriving" at 3am requires knowing this OTel SDK detail. Set a custom `otel.SetErrorHandler` that routes to the CP logger.

pkg/plugins/runtime/opentelemetry/plugin.go
```diff
@@ -56,16 +59,25 @@ func (t *tracer) NeedLeaderElection() bool {
...
+			return err
+		}
 	}
 
-	t := tracer{
-		config: otel,
-	}
-	if err := rt.Add(component.NewResilientComponent(core.Log.WithName("otel-tracer"), &t, rt.Config().General.ResilientComponentBaseBackoff.Duration, rt.Config().General.ResilientComponentMaxBackoff.Duration)); err != nil {
-		return err
+	if os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT") != "" {
```
slonka: **question:** Feature activation via raw `os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")` bypasses the config system. The tracing side already went through this: it originally read env vars directly, then got a config field - should metrics push should follow the same path?

pkg/plugins/runtime/opentelemetry/plugin.go
```diff
@@ -56,16 +59,25 @@ func (t *tracer) NeedLeaderElection() bool {
...
+		metricsLog := core.Log.WithName("otel-metrics-pusher")
+		mp := &metricsPusher{
```
slonka, line 74: **question:** Users who already set `OTEL_EXPORTER_OTLP_ENDPOINT` for tracing (following the deprecation guidance on line 97 of this file) will silently start pushing all CP Prometheus metrics to that same endpoint after upgrading. No opt-out, no warning.

should we consider a note in `UPGRADE.md` and a clear log message when metrics push auto-activates?

Also: what is the intended opt-out mechanism? `OTEL_METRICS_EXPORTER=none` won't work here because `otlpmetricgrpc.New()` is called directly.

pkg/plugins/runtime/opentelemetry/metrics.go
slonka: @Automaat what about this one?

slonka (dismissed):

pkg/plugins/runtime/opentelemetry/metrics.go
Automaat: Done in 650b977449

slonka (approved):

PR #16066, opened 2026-03-30
fix(postgres): handle conn closed err

Automaat:
## Motivation

pgxpool can hand out connections whose TCP socket was
silently closed by the network (Aurora failover, Envoy
idle timeout, cross-region routing drops). pgx v5 calls
`deallocateInvalidatedCachedStatements()` on the dead
socket, producing:

```
failed to deallocate cached statement(s): conn closed
```

Observed 60+ errors over 4 days in production (Kong
Mesh / mink), including goroutine panics causing 5xx
responses. Reducing `MaxConnectionIdleTime` did not
help — connections go stale from active TCP disruptions,
not long idle periods.

## Implementation information

Two complementary fixes:

**1. `BeforeAcquire` hook** (`connection.go`)

Rejects connections where `PgConn().IsClosed()` returns
true before handing them out from the pool. Cheap
in-memory check (no network I/O). Catches connections
the OS has already marked dead via TCP keepalive.

**2. `SafeToRetry` in Upsert** (`manager.go`)

Adds retry for errors implementing `SafeToRetry() == true`
(pgx marks `conn closed` as safe to retry since no data
was sent). Uses `errors.As` with anonymous interface to
avoid pgx dependency in manager package. Handles the
race where a connection dies between `BeforeAcquire` and
actual query execution.

## Supporting documentation

- pgx issue: https://github.com/jackc/pgx/issues/2100
- Kong/mink-charts#1202

> Changelog: skip

### DIFF
--- a/pkg/core/resources/manager/manager.go
+++ b/pkg/core/resources/manager/manager.go
@@ -209,10 +209,6 @@ func Upsert(ctx context.Context, manager ResourceManager, key model.ResourceKey,
 		if store.IsAlreadyExists(err) || store.IsConflict(err) {
 			return retry.RetryableError(err)
 		}
-		var safeErr interface{ SafeToRetry() bool }
-		if errors.As(err, &safeErr) && safeErr.SafeToRetry() {
-			return retry.RetryableError(err)
-		}
 		return err
 	})
 }
--- a/pkg/core/resources/manager/manager_test.go
+++ b/pkg/core/resources/manager/manager_test.go
@@ -2,9 +2,6 @@ package manager_test
 
 import (
 	"context"
-	"fmt"
-	"sync/atomic"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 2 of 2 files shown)


pkg/core/resources/manager/manager.go
```diff
@@ -209,6 +209,10 @@ func Upsert(ctx context.Context, manager ResourceManager, key model.ResourceKey,
 		if store.IsAlreadyExists(err) || store.IsConflict(err) {
 			return retry.RetryableError(err)
 		}
+		var safeErr interface{ SafeToRetry() bool }
+		if errors.As(err, &safeErr) && safeErr.SafeToRetry() {
```
lobkovilya: Are you sure we need all of this? Reading the docs for the `PrepareConn`:

> PrepareConn is called before a connection is acquired from the pool. If this function returns true, the connection is considered valid, otherwise the connection is destroyed. If the function returns a non-nil error, the instigating query will fail with the returned error.
> 
> Specifically, this means that:
> 	- If it returns true and a nil error, the query proceeds as normal.
> 	- If it returns true and an error, the connection will be returned to the pool, and the instigating query will fail with the returned error.
> 	- If it returns false, and an error, the connection will be destroyed, and the query will fail with the returned error.
> 	- If it returns false and a nil error, the connection will be destroyed, and **the instigating query will be retried** on a new connection.

It seems like it's going to handle the retry automatically

Automaat: Good catch — you're right.  causes pgxpool to automatically retry the query on a new connection, so adding another retry layer in `Upsert()` is redundant. Removed the `SafeToRetry` block and the corresponding tests.

PR #16071, opened 2026-03-31
fix(dns): bind proxy to loopback

### LINKED ISSUE #13802: bind new dns server to localhost
> is it the case with Internal DNS? I think we bind now to `0.0.0.0` 

 _Originally posted by @lukidzi in [#8064](https://github.com/kumahq/kuma/issues/8064#issuecomment-2983755223)_

bartsmykla:
## Motivation

The embedded DNS proxy binds to `0.0.0.0`, exposing it to the network. With transparent proxy, iptables REDIRECT sends DNS queries to loopback addresses, so binding to all interfaces is not needed.

A previous attempt (#15568) bound to `127.0.0.1` only, but was reverted (#15760) because `ip6tables REDIRECT` sends traffic to `::1`, not `127.0.0.1`, breaking IPv6-only clusters.

## Implementation information

DNS proxy bind addresses are now chosen based on transparent proxy config:

- No VNet, dual-stack: `127.0.0.1` and `::1` - OUTPUT chain REDIRECT sends to loopback
- No VNet, IPv4-only: `127.0.0.1`
- VNet, dual-stack: `0.0.0.0` and `::` - PREROUTING chain redirects to interface IP
- VNet, IPv4-only: `0.0.0.0`
- No transparent proxy: `0.0.0.0` for backward compatibility

The DNS proxy `Server` now takes multiple bind addresses, creating one UDP listener per address. All listeners share the same handler and DNS map.

`DataplaneConfig` gets a `VNet` field to deserialize VNet networks from the transparent proxy config already serialized in pod annotations. No injector or annotation changes needed.

## Supporting documentation

Fix #13802

### DIFF
--- a/app/kumactl/cmd/install/testdata/install-cp-helm/meshZoneProxyMultipleMeshes.golden.yaml
+++ b/app/kumactl/cmd/install/testdata/install-cp-helm/meshZoneProxyMultipleMeshes.golden.yaml
@@ -651,6 +651,8 @@ spec:
               value: "false"
             - name: KUMA_API_SERVER_READ_ONLY
               value: "true"
+            - name: KUMA_BOOTSTRAP_SERVER_PARAMS_ENVOY_ADMIN_UNIX_SOCKET
+              value: "true"
             - name: KUMA_DEFAULTS_SKIP_MESH_CREATION
               value: "false"
             - name: KUMA_DP_SERVER_HDS_ENABLED


pkg/transparentproxy/config/dataplane/config_dataplane.go
```diff
@@ -45,10 +45,15 @@ func DataplaneTrafficFlowFromPortLike[T PortLike](port T) DatalpaneTrafficFlow {
 	return NewDataplaneTrafficFlow(port > 0, port)
 }
 
+type DataplaneVNet struct {
```
Automaat, line 52: **suggestion:** `DataplaneVNet` duplicates the `VNet` struct from `pkg/transparentproxy/config/config.go` -- both are `Networks []string`. The separation is probably intentional (dataplane config is a stripped-down serialization subset), but the relationship is implicit. Embed or type-alias the existing `VNet` to prevent drift, or add a comment explaining why the duplication exists.

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -151,31 +150,52 @@ func (s *Server) Handler(res dns.ResponseWriter, req *dns.Msg) {
...
+		servers[i] = &dns.Server{Addr: addr, Handler: dns.HandlerFunc(s.Handler), Net: "udp"}
 	}
-	go func() {
-		err := srv.ListenAndServe()
-		serverDone <- err
-	}()
+
+	errCh := make(chan error, len(servers))
+	var readyCount atomic.Int32
+	total := int32(len(servers))
```
Automaat, line 161: **issue:** If one listener fails to bind, `readyCount` never reaches `total`, so `s.ready` never closes. `WaitForReady()` is only used in tests right now, not the component manager, so this won't hang production. But it's a latent bug: if anything starts gating health checks on `s.ready`, it will deadlock on partial failure. Close `s.ready` in the error path or switch to context-based cancellation.

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -151,31 +150,52 @@ func (s *Server) Handler(res dns.ResponseWriter, req *dns.Msg) {
...
+			}
+		}
+		go func() {
+			errCh <- srv.ListenAndServe()
+		}()
+	}
+
+	remaining := len(servers)
+	var firstErr error
+
```
Automaat, line 179: **issue:** Calling `Shutdown()` on a `dns.Server` that failed `ListenAndServe` returns `"server not started"`, which gets logged via `log.Error`. Operators will see a spurious error alongside the real failure. Skip shutdown for servers that never started, or log at debug level.

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -151,31 +150,52 @@ func (s *Server) Handler(res dns.ResponseWriter, req *dns.Msg) {
...
+	for _, srv := range servers {
+		srv.NotifyStartedFunc = func() {
+			if readyCount.Add(1) == total {
+				close(s.ready)
+			}
+		}
+		go func() {
+			errCh <- srv.ListenAndServe()
+		}()
```
Automaat, line 174: **suggestion:** The log message `[WARNING] server stopped with shutdown never called` omits the failed address and the error. With dual-stack listeners, operators cannot tell which one failed without correlating other logs. Add `"address"` and `"error"` fields to the structured log.

app/kuma-dp/cmd/run.go
```diff
@@ -49,6 +49,37 @@ import (
...
+// chain REDIRECT sends to 127.0.0.1 (IPv4) and ::1 (IPv6).
+func dnsProxyAddresses(tpCfg *tproxy_dp.DataplaneConfig, port string) []string {
+	if tpCfg == nil {
+		return []string{net.JoinHostPort("0.0.0.0", port)}
+	}
+
+	dualStack := tpCfg.IPFamilyMode != tproxy_config.IPFamilyModeIPv4
+	vnet := tpCfg.HasVNet()
+
+	switch {
```
Automaat, line 69: Binding to `::` for VNet dual-stack assumes the Linux kernel's `net.ipv6.bindv6only` sysctl defaults to `0`. If a node sets `bindv6only=1`, the `::` socket won't accept IPv4 traffic, silently breaking IPv4 DNS in dual-stack VNet mode. Either set `IPV6_V6ONLY=0` explicitly via `ListenConfig`, or bind to both `0.0.0.0` and `::` separately for VNet dual-stack (matching the loopback dual-stack approach).

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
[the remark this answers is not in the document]
bartsmykla: This is Go 1.26 - loop variable capture was fixed in Go 1.22. Each iteration gets its own variable now.

pkg/transparentproxy/config/dataplane/config_dataplane.go
```diff
@@ -45,10 +45,15 @@ func DataplaneTrafficFlowFromPortLike[T PortLike](port T) DatalpaneTrafficFlow {
...
+	Networks []string `json:"networks,omitempty"`
+}
+
 type DataplaneRedirect struct {
 	Inbound  DatalpaneTrafficFlow `json:"inbound"`
 	Outbound DatalpaneTrafficFlow `json:"outbound"`
 	DNS      DatalpaneTrafficFlow `json:"dns"`
+	VNet     DataplaneVNet        `json:"vnet,omitempty"`
```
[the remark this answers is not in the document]
bartsmykla: Removed omitempty, the tag now matches actual serialization behavior.

app/kuma-dp/cmd/run.go
```diff
@@ -49,6 +49,37 @@ import (
...
+	case vnet:
+		return []string{net.JoinHostPort("0.0.0.0", port)}
+	case dualStack:
+		return []string{
+			net.JoinHostPort("127.0.0.1", port),
+			net.JoinHostPort("::1", port),
+		}
+	default:
+		return []string{net.JoinHostPort("127.0.0.1", port)}
+	}
```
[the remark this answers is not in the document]
bartsmykla: Added tests for all four branches: nil tpCfg, IPv4-only non-VNet, dual-stack non-VNet, and IPv4-only VNet.

pkg/transparentproxy/config/dataplane/config_dataplane.go
bartsmykla: Added a comment explaining the intentional separation from config.VNet.

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
bartsmykla: Fixed - s.ready is now closed in the error path so WaitForReady never deadlocks on partial failure.

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
bartsmykla: Fixed - tracking which servers actually started via NotifyStartedFunc, skipping Shutdown on the rest.

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
bartsmykla: Added error field to the log line.

app/kuma-dp/cmd/run.go
bartsmykla: Yes - changed VNet dual-stack to bind 0.0.0.0 and :: separately instead of relying on :: alone.

Automaat (approved):

PR #16083, opened 2026-04-01
docs(MADR): add MADR 101 for status syncing rules in multizone

lahabana:
Consolidate conventions scattered from MADRs 014, 039, 044, 051, 056, 072, and 096 into explicit rules for resource status ownership and syncing in multizone deployments.


docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+1. Resources are authored on one control plane (zone or global) and synced to
+   others via KDS.
+2. Status contains information that is inherently local to the zone where the
+   resource is consumed (VIPs, hostnames, proxy counts, availability).
+3. Syncing status cross-zone would overwrite locally-computed values, causing
+   traffic interruptions and inconsistencies.
+4. Computing status on global would be costly and scale in O(number of entities globally)
+   which we try to avoid.
+
+Previous MADRs addressed this:
```
slonka, line 26: I think the list above and below is missing `014` and `072`.

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+**Global CP MUST NOT compute or populate status fields on any resource.** Global
+CP does not have the zone-local context required to produce correct values.
+
+This applies to all resource types, including:
+- `MeshService`
+- `MeshMultiZoneService`
+- `MeshExternalService`
+- Any future resource with a `status` sub-resource
+
+### Rule 2: Status MUST NOT be synced cross-zone
```
slonka: The heading says "MUST NOT be synced cross-zone" but the first paragraph says status *does* flow zone-to-global. On first read that looks like a contradiction. Something like "Status MUST NOT propagate from one zone to another" would capture the actual invariant more precisely, since zone-to-global for visibility is expected.

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+| Field | Writer | Example |
+|-------|--------|---------|
+| `spec` | The originating control plane (zone or global) | Zone creates MeshService spec; global creates MeshMultiZoneService spec |
+| `status` | The consuming zone CP | Each zone computes VIPs, hostnames, proxy stats independently |
+
+"Single writer" here means single CP deployment, not single component. Within a zone CP,
+multiple components may write to different parts of status (e.g., the VIP
+allocator writes `status.VIPs`, the hostname generator writes
+`status.Addresses`, the status updater writes `status.TLS` and
+`status.DataplaneProxies`). These are coordinated via optimistic concurrency
```
slonka: Is "optimistic concurrency (conflict retries)" referring to Kubernetes `resourceVersion` conflicts, or is there separate retry logic in the status updater? A pointer to the mechanism would help someone debugging a conflict loop.

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+(conflict retries). The key invariant is that no two CP deployments write to the same
+resource instance's status.
+
+Note that computed spec fields like `Spec.Identities` and `Spec.State` (see
```
slonka: "Status updater writes spec fields" might confuse someone new to the codebase. Worth noting that "status updater" is a component name, not a description of its scope - it writes both computed spec and status fields.

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+served by a dedicated read-only computed API endpoint that recomputes the
+information on demand rather than storing it.
+
+Note: zone-originated resources arrive on global with their status intact (see
+Rule 2). Passively storing this zone-synced status is expected — it enables
+visibility via the global API. The prohibitions below concern global CP
+**independently** computing or aggregating status:
+
+Global CP MUST NOT:
+- Independently compute or populate status fields on resources
```
slonka, line 143:
```suggestion
- Independently compute or populate status fields on resources (passively storing zone-synced data via KDS is fine)
```

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+- No dependency on global CP for status computation
+- Zone can operate autonomously for status even when disconnected from global
+
+## Implications for Kong Mesh
+
+None. These rules apply uniformly to all deployments.
+
+## Decision
+
+Status is always zone-specific and MUST only be computed on zone CPs.
```
slonka: These rules are framed for multizone, but standalone (non-federated) mode has a single CP acting as both originator and consumer. The Insight exception mentions "non-federated zone CPs" but there's no general statement about how these rules map to standalone. A sentence or two early in the Design section would help.

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+Global CP MUST NOT compute status. Cross-zone status visibility requires a
+dedicated API. Data that needs to travel across zones belongs in `spec`.
+
+These rules consolidate and strengthen existing conventions from MADRs 014, 039,
+044, 051, 056, 072, and 096.
```
slonka, line 245: What does enforcement look like? The guards already exist in code but they're ad-hoc per component. Is code review sufficient or should there be a systematic check?

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+for how resource `status` is handled in multizone deployments. These conventions
+are scattered and implicit. This MADR consolidates the rules into a single
+reference and strengthens them with an explicit prohibition on computing status
+on Global CP.
+
+The core tension is:
+
+1. Resources are authored on one control plane (zone or global) and synced to
+   others via KDS.
+2. Status contains information that is inherently local to the zone where the
```
slonka, line 19: These examples (VIPs, hostnames, proxy counts, availability) seem `MeshService`-specific - what about the other resources with `HasStatus: true` (MeshIdentity, MeshOpenTelemetryBackend, Workload, MeshTrust, MeshTrace, MeshMetric, MeshAccessLog, HostnameGenerator)? Should the framing be broader to cover those cases or is this MADR only about `Mesh*Service`?

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+
+Technical Story: none
+
+## Context and Problem Statement
+
+Across multiple MADRs (039, 044, 051, 056, 096) we have established conventions
+for how resource `status` is handled in multizone deployments. These conventions
+are scattered and implicit. This MADR consolidates the rules into a single
+reference and strengthens them with an explicit prohibition on computing status
+on Global CP.
```
slonka, line 13: The context says these conventions are "scattered and implicit" - are there implicit ones in the code that aren't captured by any of the referenced MADRs? 

Looking at the codebase:
1. Every status updater and VIP allocator is gated by [`rt.GetMode() == config_core.Zone`](https://github.com/kumahq/kuma/blob/master/pkg/insights/components.go#L16) (and [`pkg/ipam/components.go`](https://github.com/kumahq/kuma/blob/master/pkg/ipam/components.go#L18)). 
2. [`IsLocalMeshService()`](https://github.com/kumahq/kuma/blob/master/pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go#L28) filtering in the status updater

Is it worth mentioning these as existing enforcement mechanisms, or at least acknowledging them so future contributors know the guards already exist?

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+merged across zones.
+
+However, when global syncs that resource onward to other zones, the status is
+stripped by the `RemoveStatus()` mapper (`pkg/kds/context/context.go`), which is
+applied as a blanket transformation to all resources with `HasStatus: true` sent
+from global to zones. On the receiving zone, the `IgnoreStatusChange` sync option
+(`pkg/kds/v2/store/sync.go`) ensures that locally-computed status is preserved
+and not overwritten by the empty status arriving from global.
+
+**Implication: global-originated resources have no status on global.** Since
```
slonka, line 72: Does it make sense to compute it for some resources? For `MeshService`-style status (VIPs, hostnames) the answer is clear - global lacks the zone-local inputs.

Is the same true for `MeshOpenTelemetryBackend`, `MeshIdentity` etc.?

Was this considered?

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+"Single writer" here means single CP deployment, not single component. Within a zone CP,
+multiple components may write to different parts of status (e.g., the VIP
+allocator writes `status.VIPs`, the hostname generator writes
+`status.Addresses`, the status updater writes `status.TLS` and
+`status.DataplaneProxies`). These are coordinated via optimistic concurrency
+(conflict retries). The key invariant is that no two CP deployments write to the same
+resource instance's status.
+
+Note that computed spec fields like `Spec.Identities` and `Spec.State` (see
+Rule 4) are also written by the local zone's status updater, not by the user.
```
slonka, line 100: A bit related: should we consider rename since it's no longer "status updater"? computed fields reconciler?

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+This is consistent with the single-CP-writer model: for a local MeshService, the
+originating zone writes both user-authored and computed spec fields.
+
+This separation eliminates race conditions between spec updates from the origin
+and status updates on the destination (the scenario described in MADR 056).
+
+### Rule 4: Data that must cross zone boundaries belongs in spec
```
slonka, line 107: Rule 4 formalizes "computed spec fields" as a concept but there's no structural way to distinguish them from user-authored spec fields. `Spec.Identities` and `Spec.State` sit next to user-authored fields. Should we come up with a naming pattern for this?

docs/madr/decisions/101-status-syncing-rules.md
lahabana: I think this MADR only tries to persist what the state of things are. I'm purposefully not trying to expand to new things to keep it focused.

docs/madr/decisions/101-status-syncing-rules.md
lahabana: Yes we could but didn't want to create work out of this MADR

docs/madr/decisions/101-status-syncing-rules.md
lahabana: Probably, though the goal of the MADR is to persist existing state not create new work.

docs/madr/decisions/101-status-syncing-rules.md
lahabana: Isn't it the 3rd point?

docs/madr/decisions/101-status-syncing-rules.md
lahabana: I think I was mostly trying to focus on persisting this. For what it's worth I got this MADR and the code reviewed by agents a few times with current use-cases for completeness. Which worked very well. I think as our usage of AI grows I think having skill to help self-reviews with targeted agents will probably get us a bigger effort/outcome than code enforcements.

docs/madr/decisions/101-status-syncing-rules.md
slonka: Sort of is, the "synced data into stored resources" part indicates that something is ok to be synced, wanted to make it clear. If no one else feels the need then feel free to drop this.

docs/madr/decisions/101-status-syncing-rules.md
slonka: It's adding more restrictions (e.g. "... strengthens them with an explicit prohibition ...") but I understand that we want to keep scope minimal.

docs/madr/decisions/101-status-syncing-rules.md
slonka: We should think about whether code enforcements are feasible.

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+
+Previous MADRs addressed this:
+
+| MADR | Decision |
+|------|----------|
+| 039 (MeshService API) | Status holds VIPs, hostnames, proxy counts, availability. Managed by CP. |
+| 051 (MeshService multizone) | Status is NOT synced cross-zone. Each zone computes its own. |
+| 056 (Identity sync) | Identity placed in `spec` (not `status`) to avoid partial status syncing. Motivated the single-writer model for spec vs status. |
+| 044 (Zone-to-global policy sync) | Zone-originated policies sync to global for visibility only. |
+| 096 (Ingress address sync) | `MeshZoneAddress` is status-less; address is in `spec`. |
```
Automaat, line 34: **suggestion:** The Context intro and precedent table list five MADRs (039, 044, 051, 056, 096), but the title and Decision claim seven (adding 014 and 072). Add 014 and 072 rows here so the precedent table matches the stated scope; both are already cited in Rule 5.

docs/madr/decisions/101-status-syncing-rules.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Design
+
```
Automaat, line 37: **praise:** Every code reference in this MADR resolves to the real symbol at the cited location — RemoveStatus(), IgnoreStatusChange, the _hostnames handler, the resyncer, the proto RPCs, and the computed spec fields all check out. Consolidation docs usually rot on exactly these details.

Automaat (approved):
Approve — a clean consolidation MADR whose every code reference (mappers, sync options, the _hostnames endpoint, the resyncer, proto RPCs, spec fields) resolves to the real symbol at the cited location. One minor gap: the precedent table lists five MADRs but the title/Decision claim seven — add the 014 and 072 rows. No blockers.

PR #16097, opened 2026-04-02
feat(log): per-component log level control via API

### LINKED ISSUE #16095: Per-component log level control via control plane API
## Problem

When debugging issues in a production Kuma control plane, the only option is to set the global log level to `debug` via `--log-level`. This floods logs with debug output from **every** component (xds, kds, insights, bootstrap, etc.), making it hard to find relevant information and potentially impacting performance.

For example, if I'm investigating a KDS synchronization issue between global and zone, I have to enable debug logging for the entire control plane just to see KDS-related debug logs. This generates massive amounts of noise from unrelated components like xds, insights resyncers, etc.

## Desired Behavior

I would like to be able to change the log level for individual components at runtime, without restarting the control plane. Ideally via an HTTP API:

```bash
# See current log levels for all components
curl http://localhost:5681/logging

# Enable debug logging only for KDS
curl -X PUT http://localhost:5681/logging -d '{"component": "kds", "level": "debug"}'

# Reset back to default
curl -X DELETE http://localhost:5681/logging/kds
```

This would allow targeted debugging in production without the noise and performance impact of global debug logging.

## Additional Context

- Kuma already has 180+ named component loggers via `core.Log.WithName("component")` — the infrastructure for component identification exists
- Similar capability exists in Envoy (`/logging` admin endpoint) and Istio (`/scopez` endpoint)
- MADR 003 (logging guidelines) notes on line 206: _"controlling the log level on logger name is not yet available"_
- Changes should be in-memory only (lost on restart) — this is a runtime debugging tool, not persistent configuration

Automaat:
## Motivation

When debugging issues in production, the only option is setting the global log level to `debug`, flooding logs from all 188+ components. This makes it hard to find relevant info and impacts performance. Operators need targeted debugging for specific components (e.g. only KDS or XDS) without the noise.

Resolves #16095

## Implementation information

**Registry-aware LogSink wrapper** — zero changes to existing `core.Log.WithName()` call sites.

- `ComponentLevelRegistry` (`pkg/log/component_level.go`) — thread-safe map of component name → log level with hierarchical lookup (setting `xds` to debug also applies to `xds.server`, `xds.secrets`, etc.)
- `componentAwareSink` (`pkg/log/component_sink.go`) — custom `logr.LogSink` wrapping zapr. Intercepts `WithName()` to track dot-separated component names, `Enabled()` to check registry overrides. Inner zap logger runs at max verbosity; all level filtering happens in the sink
- HTTP API (`pkg/api-server/logging_endpoint.go`):
  - `GET /logging` — list global level + all component overrides
  - `PUT /logging` — set component or global level
  - `DELETE /logging/{component}` — reset one override
  - `DELETE /logging` — reset all overrides
- Overrides are in-memory only (lost on restart) — this is a runtime debugging tool

### DIFF
--- a/pkg/diagnostics/server.go
+++ b/pkg/diagnostics/server.go
@@ -3,6 +3,7 @@ package diagnostics
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/http/pprof"
@@ -17,6 +18,7 @@ import (
 	config_types "github.com/kumahq/kuma/v2/pkg/config/types"
 	"github.com/kumahq/kuma/v2/pkg/core"
 	"github.com/kumahq/kuma/v2/pkg/core/runtime/component"
+	kuma_log "github.com/kumahq/kuma/v2/pkg/log"
 	"github.com/kumahq/kuma/v2/pkg/metrics"
 	kuma_srv "github.com/kumahq/kuma/v2/pkg/util/http/server"
 )
@@ -24,10 +26,11 @@ import (
 var diagnosticsServerLog = core.Log.WithName("xds-server").WithName("diagnostics")
 
 type diagnosticsServer struct {
-	isReady func() bool
-	config  *diagnostics_config.DiagnosticsConfig
-	metrics metrics.Metrics
-	ready   atomic.Bool
+	isReady  func() bool
+	config   *diagnostics_config.DiagnosticsConfig
+	metrics  metrics.Metrics
+	registry *kuma_log.ComponentLevelRegistry
+	ready    atomic.Bool
 }
 
 func (s *diagnosticsServer) NeedLeaderElection() bool {
@@ -59,6 +62,7 @@ func (s *diagnosticsServer) Start(stop <-chan struct{}) error {
 		mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
 		mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
 	}
+	AddLoggingHandlers(mux, s.registry)
 	var tlsConfig *tls.Config
 	if s.config.TlsEnabled {
 		cert, err := tls.LoadX509KeyPair(s.config.TlsCertFile, s.config.TlsKeyFile)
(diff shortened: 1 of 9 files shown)


pkg/api-server/server.go
```diff
@@ -180,6 +180,7 @@ func NewApiServer(
 		return nil, errors.Wrap(err, "could not create index webservice")
 	}
 	addWhoamiEndpoints(ws)
+	addLoggingEndpoints(ws, rt.Access().ControlPlaneMetadataAccess)
```
lahabana: We can't put this in a api-server for 2 reasons:

1. Anything in the public api is usually meant to do something in the DB. Here when you have multiple instances things just land in random places.
2. Anything in this api should be documented with openAPI now and be done as openAPI first.

This is more diagnostic stuff and we already have a listener for this: https://github.com/kumahq/kuma/blob/fa7b39e3e912364f5323df9b5dab40aeb4c86a7a/pkg/diagnostics/server.go#L50

Should we move this here?

pkg/log/component_level.go
```diff
@@ -0,0 +1,91 @@
...
+// It checks for an exact match first, then walks up the hierarchy
+// (e.g. "xds.server" → "xds"). Returns the level and true if an
+// override was found, or zero value and false otherwise.
+func (r *ComponentLevelRegistry) GetEffectiveLevel(component string) (LogLevel, bool) {
+	if component == "" {
+		return 0, false
+	}
+
+	r.mu.RLock()
+	defer r.mu.RUnlock()
```
lahabana: looks like we're doing a read mutex lock with every single log message. This tends to create contention at high load. Should we consider a lockless data structure like an atomic reference to a map? AFAI the expectation is that this will be read way more than it will be written so the case for duplicating the Datastructure on write makes sense no?

pkg/log/component_level.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		idx := strings.LastIndex(name, ".")
+		if idx < 0 {
+			break
+		}
+		name = name[:idx]
+	}
+
+	return 0, false
+}
```
lahabana, line 132: This loop in the critical path isn't great because we do entire splits for each log messages right? Looking at your sink you could do the split there already no?

pkg/log/component_level.go
```diff
@@ -0,0 +1,148 @@
...
+	result := make(map[string]LogLevel, len(prev))
+	maps.Copy(result, prev)
+	return result
+}
+
+// GetEffectiveLevel returns the effective log level for a component.
+// It checks for an exact match first, then walks up the hierarchy
+// (e.g. "xds.server" → "xds"). Returns the level and true if an
+// override was found, or zero value and false otherwise.
+func (r *ComponentLevelRegistry) GetEffectiveLevel(component string) (LogLevel, bool) {
```
lahabana: Do we still need this it doesn't like it's used anywhere no?

Automaat: removed

pkg/diagnostics/server.go
```diff
@@ -17,17 +18,19 @@ import (
...
 
 type diagnosticsServer struct {
-	isReady func() bool
-	config  *diagnostics_config.DiagnosticsConfig
-	metrics metrics.Metrics
-	ready   atomic.Bool
+	isReady  func() bool
+	config   *diagnostics_config.DiagnosticsConfig
+	metrics  metrics.Metrics
+	registry *kuma_log.ComponentLevelRegistry
```
lobkovilya: nit: maybe `logRegistry` or something like that, otherwise the name is too generic

pkg/diagnostics/server.go
```diff
@@ -102,3 +106,92 @@ func (s *diagnosticsServer) Start(stop <-chan struct{}) error {
...
+			http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+			return
+		}
+	})
+	mux.HandleFunc("/logging/", func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodDelete {
+			http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+			return
+		}
+		component := r.URL.Path[len("/logging/"):]
```
lobkovilya: since Go 1.22 you can use PathValue https://pkg.go.dev/net/http#Request.PathValue, see [examples](https://go.dev/blog/routing-enhancements#enhancements)

pkg/log/component_level.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if !validComponentName.MatchString(name) {
+		return fmt.Errorf("component name %q contains invalid characters", name)
+	}
+	return nil
+}
+
+// SetLevel sets a log level override for the given component.
+// Returns an error if the maximum number of overrides would be exceeded.
+func (r *ComponentLevelRegistry) SetLevel(component string, level LogLevel) error {
```
lobkovilya, line 65: Small efficiency improvement is in methods like `SetLevel`, `ResetLevel` and `ResetAll` check if the snapshot is already in desired state before locking mutex and performing a copy.

pkg/diagnostics/server.go
Automaat: fixed

pkg/diagnostics/server.go
Automaat: fixed

pkg/log/component_level.go
Automaat: fixed

bartsmykla (approved):

PR #16101, opened 2026-04-02
fix(api): add KRI support for Zone resource

### LINKED ISSUE #16000: Zone is missing KRI support
### Description

When fetching the `/_resources` endpoint `Zone` is listed in the response, but it has no `shortName`. Trying to fetch the generic `/_kri/{kri}` API with i.e. `/_kri/kri_z____zone-1_` is not supported and returns with 400. When fetching a zone via `/zones/{zone}/_overview` the response is missing a `kri` entry.

Automaat:
## Motivation

Zone resource is missing KRI (Kuma Resource Identifier) support (#16000). The `/_resources` endpoint lists Zone with an empty `shortName`, the `/_kri/{kri}` endpoint returns 400 for Zone KRIs, and zone overview responses omit the `kri` field.

Root cause: the `short_name` proto option was not set on the Zone message.

## Implementation information

Added `short_name = "z"` to `api/system/v1alpha1/zone.proto` and regenerated all dependent files. This follows the existing pattern used by other resources (Mesh=`m`, Dataplane=`dp`, ZoneEgress=`ze`, ZoneIngress=`zi`).

All existing KRI infrastructure (overview endpoints, REST marshaling, KRI parsing) already handles Zone correctly once the short name is present — no code changes beyond the proto option were needed.

## Supporting documentation

Closes #16000

> Changelog: fix(api): add KRI support for Zone resource

### DIFF
--- a/api/system/v1alpha1/zone.pb.go
+++ b/api/system/v1alpha1/zone.pb.go
@@ -75,11 +75,11 @@ var File_api_system_v1alpha1_zone_proto protoreflect.FileDescriptor
 
 const file_api_system_v1alpha1_zone_proto_rawDesc = "" +
 	"\n" +
-	"\x1eapi/system/v1alpha1/zone.proto\x12\x14kuma.system.v1alpha1\x1a\x16api/mesh/options.proto\x1a\x1egoogle/protobuf/wrappers.proto\"\xa4\x01\n" +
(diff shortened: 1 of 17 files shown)


Automaat:
> This is a great catch. I don't know of any k8s resources that have a single letter shortname. May I suggest we go for `zn` instead?

@lahabana we use `m` for `Mesh`, so I think `z` for `Zone` is fine

PR #16105, opened 2026-04-02
chore(build): migrate to use oidc token

lukidzi:
## Motivation

We want to switch to using OIDC instead of token

## Implementation information

Switch workflow to use OIDC instead of token
* created 2 envs one for preview and one for release

### DIFF
--- a/.github/workflows/_build_publish.yaml
+++ b/.github/workflows/_build_publish.yaml
@@ -84,7 +84,7 @@ jobs:
         uses: cloudsmith-io/cloudsmith-cli-action@18665afcce9f859312b61989671af9af7402e4fb # v2.0.2
         with:
           oidc-namespace: kong
-          oidc-service-slug: ${{ vars.CLOUDSMITH_SERVICE_ACCOUNT }}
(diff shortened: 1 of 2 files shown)


.github/workflows/_build_publish.yaml
```diff
@@ -76,11 +79,18 @@ jobs:
             ./build/distributions/out/*.sha256
             !./build/distributions/out/*.tar.gz.sha256
           retention-days: ${{ github.event_name == 'pull_request' && 1 || 10 }}
+      - name: Get Cloudsmith OIDC token
+        id: cloudsmith_token
+        uses: cloudsmith-io/cloudsmith-cli-action@18665afcce9f859312b61989671af9af7402e4fb # v2.0.2
+        with:
+          oidc-namespace: kong
+          oidc-service-slug: ${{ vars.CLOUDSMITH_SERVICE_ACCOUNT }}
```
lahabana: We've create 2 service accounts remember? One for dev releases and one for prod releases so I think this needs 2 vars and a if

lukidzi: Fixed, what were the exact names of them? I would like to add them to the Secrets but I don't have access to cloudsmith to verify that information. Also, is `oidc-namespace: kong` correct?

lahabana: namespace correct and saw you updated all this

lahabana (approved):

PR #16107, opened 2026-04-03
fix(api): consistently override kuma.io/display-name

### LINKED ISSUE #16049: name / display-name validation doesn't always seem to be working
### Kuma Version

latest

### Describe the bug

It seems to me that when creating and/or updating an existing resource it is possible for the name / display-name to get out of sync, or maybe better said "mismatched"

The Control Plane should be either overwriting the resource with the correct value, or better returning a validation error when these values are "mismatched"

Note: I'm sure I've seen a validation error sometimes, but I've also just tried to briefly reproduce and cannot see this at the moment and its possible to save resources with mismatched name / display-name

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

lobkovilya:
## Motivation

`kuma.io/display-name` is meant to hold the human-readable name of a resource. On Kubernetes, resource names get hashed during KDS sync (e.g. `my-policy` → `my-policy-abc123`), so `display-name` is the only way to recover the original name for display in the UI and for policy matching. However, `ComputeLabels` never enforced this label — a user could supply any arbitrary value via a REST API label or a `kubectl apply` annotation and it would be persisted as-is, creating mismatches between `name` and `display-name`.

## Implementation information

Added a `WithDisplayName(name string)` option to `ComputeLabels`. When provided, it unconditionally overrides the `display-name` label (bypassing the usual "keep existing" behaviour). The callers decide whether to pass it:

- **REST API (Universal):** `createResource` always passes `WithDisplayName(name)` from the URL path. `updateResource` reads the current `display-name` from the stored resource and passes it, making the field immutable across updates.
- **K8s admission webhook:** passes `WithDisplayName(obj.Name)` only when the request is from a **non-privileged user** (i.e. a real `kubectl` interaction). Privileged CP service accounts — which perform KDS sync writes — are excluded, so the already-correct `display-name` set by the KDS hash mapper is preserved.

`mesh_proto.DisplayName` was also registered in `AllComputedLabels`, which is required for `set()` inside `ComputeLabels` not to panic.

## Supporting documentation

Fix #16049

### DIFF
--- a/pkg/api-server/testdata/resources/crud/put_circuitbreaker_01.golden.json
+++ b/pkg/api-server/testdata/resources/crud/put_circuitbreaker_01.golden.json
@@ -5,6 +5,7 @@
  "creationTime": "0001-01-01T00:00:00Z",
  "modificationTime": "0001-01-01T00:00:00Z",
  "labels": {
+  "kuma.io/display-name": "web-to-backend",
   "kuma.io/mesh": "default"
  },
  "sources": [
--- a/pkg/api-server/testdata/resources/crud/put_faultinjection_01.golden.json
+++ b/pkg/api-server/testdata/resources/crud/put_faultinjection_01.golden.json
@@ -5,6 +5,7 @@
  "creationTime": "0001-01-01T00:00:00Z",
  "modificationTime": "0001-01-01T00:00:00Z",
  "labels": {
+  "kuma.io/display-name": "web-to-backend",
   "kuma.io/mesh": "default"
  },
  "sources": [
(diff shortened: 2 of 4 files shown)


lahabana:
@lobkovilya the issue suggests returning an update error. Is there a reason this option wasn't picked?

lobkovilya:
@lahabana the issue doesn't really suggest the solution:

> The Control Plane should be either overwriting the resource with the correct value, or better returning a validation error when these values are "mismatched"

So I've picked the same approach we have for other auto-computed labels, when user tries to specify `kuma.io/policy-role`, `kuma.io/proxy-type`, `kuma.io/workload`, etc, the CP just overrides the value with the correct one without the error.

From a UX perspective, validation errors might be nicer for GUI CRUD forms. But if we want to switch to that behaviour, I think we should do it consistently across all auto-computed labels.

lobkovilya:
I noticed there was already work done in this area https://github.com/kumahq/kuma/pull/15164, however the PR touched only the update path, that's why it's still possible to set arbitrary labels on create. 

I'm going to close this one and open new PR that continues the direction started in https://github.com/kumahq/kuma/pull/15164

PR #16108, opened 2026-04-04
fix(api-server): include HostnameGenerator

### LINKED ISSUE #15993: Missing `HostnameGenerator` in global-insight
### Kuma Version

2.13

### Describe the bug

The `/global-insight` endpoint returns the amounts of several resources, but it is missing the amount of `HostnameGenerator`s.

**Feature request**
Additionally it would be great if we could add an entry for each individual resource by its `type` to the `resources` entry. This would make it a lot easier to extract information in the GUI.

### To Reproduce

_No response_

### Expected behavior

_No response_

Example response:

```json
{
 "createdAt": "2026-03-25T10:56:18.966390461Z",
 "dataplanes": {
  "gatewayBuiltin": {
   "offline": 0,
   "online": 0,
   "partiallyDegraded": 0,
   "total": 0
  },
  "gatewayDelegated": {
   "offline": 0,
   "online": 0,
   "partiallyDegraded": 0,
   "total": 0
  },
  "standard": {
   "offline": 0,
   "online": 2,
   "partiallyDegraded": 0,
   "total": 2
  }
 },
 "meshes": {
  "total": 1
 },
 "policies": {
  "total": 9
 },
 "resources": {
  "Dataplane": {
   "total": 2
  },
  "MeshCircuitBreaker": {
   "total": 3
  },
  "MeshIdentity": {
   "total": 1
  },
  "MeshRetry": {
   "total": 2
  },
  "MeshService": {
   "total": 4
  },
  "MeshTimeout": {
   "total": 4
  },
  "MeshTrust": {
   "total": 1
  },
  "Secret": {
   "total": 3
  },
  "Workload": {
   "total": 2
  }
 },
 "services": {
  "external": {
   "total": 0
  },
  "gatewayBuiltin": {
   "offline": 0,
   "online": 0,
   "partiallyDegraded": 0,
   "total": 0
  },
  "gatewayDelegated": {
   "offline": 0,
   "online": 0,
   "partiallyDegraded": 0,
   "total": 0
  },
  "internal": {
   "offline": 0,
   "online": 3,
   "partiallyDegraded": 0,
   "total": 3
  }
 },
 "zones": {
  "controlPlanes": {
   "online": 1,
   "total": 1
  },
  "zoneEgresses": {
   "online": 0,
   "total": 0
  },
  "zoneIngresses": {
   "online": 1,
   "total": 1
  }
 }
}
```

aviralgarg05:
## Motivation

`/global-insight` is meant to give a quick summary of the state of the control plane, including resource totals that the GUI can consume directly.

At the moment, those resource totals are built only from `MeshInsight.Resources`. That works for mesh-scoped resources, but it misses global-scoped resources entirely. As a result, `HostnameGenerator` does not appear in the `/global-insight` response even though it is a valid resource type and is already visible through the admin API.

This change fixes that gap so the endpoint reports `HostnameGenerator` correctly.

## Implementation information

The fix keeps the existing aggregation from `MeshInsight.Resources` for mesh-scoped entries and extends it by also collecting counts for relevant global-scoped resources from the registered resource descriptors.

A few resource types are still intentionally excluded from the `resources` map:
- insight resources, since they are internal aggregation artifacts rather than user-facing resource totals
- resources that already have dedicated top-level sections in the response, such as meshes and zones
- config entries that are not meant to be exposed here

This approach avoids special-casing `HostnameGenerator` and instead fixes the underlying mismatch between mesh-scoped and global-scoped resource counting.

The test coverage was updated to add `HostnameGenerator` fixtures to the global insight service test, and the golden output was refreshed accordingly.

## Supporting documentation

Issue: #15993

Tests run:
- `go test ./pkg/insights/globalinsight -count=1`
- `go test ./pkg/insights/... -count=1`
- `go test -race ./pkg/insights/globalinsight -count=1`

Fix #15993


lahabana:
This is nice! Does it make sense to refactor the different methods that act on global into a single one? This may make the code much easier to read.

aviralgarg05:
> This is nice! Does it make sense to refactor the different methods that act on global into a single one? This may make the code much easier to read.

Good call. I folded the mesh-scoped and global-scoped resource counting into a single `aggregateResources` path so the flow is easier to follow while keeping the behavior unchanged.

pkg/insights/globalinsight/global_insight_service.go
```diff
@@ -100,20 +104,49 @@ func (gis *defaultGlobalInsightService) aggregatePolicies(
...
+		core_model.HasScope(core_model.ScopeGlobal),
+		core_model.HasWsEnabled(),
+		core_model.Not(core_model.IsInsight()),
+		core_model.Not(core_model.Named(
+			mesh.MeshType,
+			system.ZoneType,
+			mesh.ZoneIngressType,
+			mesh.ZoneEgressType,
+			system.ConfigType,
+		)),
```
lukidzi: I think we can initialize it in `NewDefaultGlobalInsightService` and store it on the object, so we don’t have to filter it every time.

pkg/insights/globalinsight/global_insight_service.go
```diff
@@ -100,20 +104,49 @@ func (gis *defaultGlobalInsightService) aggregatePolicies(
...
-			}
-			stats.Total += int(resStat.Total)
-			globalInsight.Resources[resName] = stats
+			addResourceTotal(globalInsight.Resources, resName, int(resStat.Total))
+		}
+	}
+
+	for _, descriptor := range registry.Global().ObjectDescriptors(
+		core_model.HasScope(core_model.ScopeGlobal),
+		core_model.HasWsEnabled(),
```
lukidzi: I don't think this one is correct, but I think we need to exclude admin only resources: `model.TypeFilterFn(func(d model.ResourceTypeDescriptor) bool { return !d.AdminOnly })`

pkg/insights/globalinsight/global_insight_service.go
```diff
@@ -100,20 +104,49 @@ func (gis *defaultGlobalInsightService) aggregatePolicies(
...
+		if err := gis.resourceStore.List(ctx, list); err != nil {
+			return err
+		}
+
+		count := len(list.GetItems())
+		if count == 0 {
+			continue
 		}
+
+		addResourceTotal(globalInsight.Resources, string(descriptor.Name), count)
```
lukidzi: I don't think we need that logic as we have only one call for all resources, we don't have separate meshes.

```golang
globalInsight.Resources[string(desc.Name)] = api_types.ResourceStats{Total: len(list.GetItems())}
```

lukidzi (changes requested):

lukidzi (approved):

lukidzi:
@aviralgarg05 you need to sign DCO before I can merge

PR #16109, opened 2026-04-05
fix(metrics): count zone resources without insights

### LINKED ISSUE #15895: fix(metrics): zone CP missing resources_count for Dataplane
## Motivation

Zone CPs don't emit `resources_count{resource_type="Dataplane"}` metric. The counter in `pkg/metrics/store/counter.go` derives Dataplane and policy counts from `MeshInsight` resources, which only exist on the global CP. Zone CPs have no `MeshInsight`, so the loop never executes and these resource types are never emitted.

This breaks dashboards and monitoring on zone CPs — the Dataplane Count panel shows "No data".

## Implementation information

`countMeshScopedResources()` in `pkg/metrics/store/counter.go` relies solely on `MeshInsight.Spec.GetDataplanes().GetTotal()`. On a zone CP, it should count `Dataplane` resources directly from the store instead of depending on `MeshInsight`.

officialasishkumar:
## Motivation

Zone control planes do not create `MeshInsight` resources, but the store counter currently relies on `MeshInsight` for all mesh-scoped `resources_count` metrics. That means a zone CP can have local dataplanes and policies while still emitting no `resources_count{resource_type="Dataplane"}` metric at all.

## Implementation information

- keep the existing `MeshInsight` aggregation path unchanged when insights are present
- add a fallback path for environments without `MeshInsight` that counts `Dataplane` and policy resources directly from the store
- add a regression test that covers the zone-CP case where mesh resources exist without any `MeshInsight` entries

## Supporting documentation

Fix #15895

Validation:
- `go test ./pkg/metrics/store -run TestMetrics -count=1 -v`
- `go test -race ./pkg/metrics/store -run TestMetrics -count=1`

> Changelog: fix(metrics): count zone resources without insights

### DIFF
--- a/pkg/metrics/store/counter.go
+++ b/pkg/metrics/store/counter.go
@@ -128,12 +128,11 @@ func (s *storeCounter) countMeshScopedResources(ctx context.Context, resourceCou
 	if err := s.resManager.List(ctx, insights); err != nil {
 		return err
 	}
-	// When no MeshInsights exist, fall back to counting resources directly from the store.
-	// This covers zone CPs (which don't have MeshInsight resources) as well as the rare
-	// case of a global CP with zero meshes, where the fallback is harmless.
+	// Zone CPs do not create MeshInsight resources, so count dataplanes directly
+	// from the store to keep the Dataplane metric available there as well.
 	if len(insights.Items) == 0 {
-		log.V(1).Info("no MeshInsight found, counting mesh-scoped resources from store directly")
-		return s.countMeshScopedResourcesWithoutInsights(ctx, resourceCount)
+		log.V(1).Info("no MeshInsight found, counting dataplanes from store directly")
+		return s.countDataplanesWithoutInsights(ctx, resourceCount)
 	}
 	for _, meshInsight := range insights.Items {
 		resourceCount[string(mesh.DataplaneType)] += meshInsight.Spec.GetDataplanes().GetTotal()
@@ -144,21 +143,11 @@ func (s *storeCounter) countMeshScopedResources(ctx context.Context, resourceCou
 	return nil
 }
 
-func (s *storeCounter) countMeshScopedResourcesWithoutInsights(ctx context.Context, resourceCount map[string]uint32) error {
-	for _, resDesc := range registry.Global().ObjectDescriptors() {
-		if resDesc.Scope != model.ScopeMesh {
-			continue
-		}
-		if resDesc.Name != mesh.DataplaneType && !resDesc.IsPolicy {
-			continue
-		}
-
-		list := resDesc.NewList()
-		if err := s.resManager.List(ctx, list); err != nil {
-			return err
-		}
-		resourceCount[string(resDesc.Name)] += uint32(len(list.GetItems()))
+func (s *storeCounter) countDataplanesWithoutInsights(ctx context.Context, resourceCount map[string]uint32) error {
+	dataplanes := &mesh.DataplaneResourceList{}
+	if err := s.resManager.List(ctx, dataplanes); err != nil {
+		return err
 	}
-
+	resourceCount[string(mesh.DataplaneType)] += uint32(len(dataplanes.GetItems()))
 	return nil
 }
--- a/pkg/metrics/store/counter_test.go
+++ b/pkg/metrics/store/counter_test.go
@@ -138,7 +138,7 @@ var _ = Describe("Counter", func() {
 		}).Should(Succeed())
 	})
 
-	It("should count dataplanes and policies without mesh insights", func() {
+	It("should count dataplanes without mesh insights", func() {
 		// given
 		err := resManager.Create(context.Background(), core_mesh.NewMeshResource(), core_store.CreateByKey("mesh-1", model.NoMesh))
 		Expect(err).ToNot(HaveOccurred())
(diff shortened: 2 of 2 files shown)


pkg/metrics/store/counter.go
```diff
@@ -128,6 +128,9 @@ func (s *storeCounter) countMeshScopedResources(ctx context.Context, resourceCou
 	if err := s.resManager.List(ctx, insights); err != nil {
 		return err
 	}
+	if len(insights.Items) == 0 {
```
Automaat, line 133: `len(insights.Items) == 0` conflates "no meshes exist" with "zone CP without insights." If a global CP temporarily has zero meshes, this fires the expensive per-type List path unnecessarily. A CP-mode check would be more precise and self-documenting. That said, zero-mesh global CPs listing 14 empty types is harmless — this is a clarity concern, not a correctness one.

pkg/metrics/store/counter.go
```diff
@@ -136,3 +139,22 @@ func (s *storeCounter) countMeshScopedResources(ctx context.Context, resourceCou
...
+func (s *storeCounter) countMeshScopedResourcesWithoutInsights(ctx context.Context, resourceCount map[string]uint32) error {
+	for _, resDesc := range registry.Global().ObjectDescriptors() {
+		if resDesc.Scope != model.ScopeMesh {
+			continue
+		}
+		if resDesc.Name != mesh.DataplaneType && !resDesc.IsPolicy {
+			continue
+		}
+
+		list := resDesc.NewList()
```
Automaat: At scale (5k+ Dataplanes on a zone CP), deserializing full Dataplane specs every minute solely to count them produces non-trivial GC pressure. This matches the existing `countGlobalScopedResources` pattern so it's not a regression introduced by this PR. Worth a follow-up issue for a `Count()` store API that avoids full deserialization — it would benefit both code paths.

pkg/metrics/store/counter.go
```diff
@@ -136,3 +139,22 @@ func (s *storeCounter) countMeshScopedResources(ctx context.Context, resourceCou
 	}
 	return nil
 }
```
Automaat: **praise:** Clean adherence to existing patterns. Method naming (`countMeshScopedResourcesWithoutInsights`) follows the file's `count{Scope}Resources` convention. Error handling is identical to `countGlobalScopedResources`. Test uses the same Ginkgo/Gomega/Eventually style.

pkg/metrics/store/counter.go
```diff
@@ -136,3 +139,22 @@ func (s *storeCounter) countMeshScopedResources(ctx context.Context, resourceCou
...
+
+func (s *storeCounter) countMeshScopedResourcesWithoutInsights(ctx context.Context, resourceCount map[string]uint32) error {
+	for _, resDesc := range registry.Global().ObjectDescriptors() {
+		if resDesc.Scope != model.ScopeMesh {
+			continue
```
Automaat: **praise:** The filter logic `resDesc.Name != mesh.DataplaneType && !resDesc.IsPolicy` correctly mirrors what MeshInsight tracks (dataplanes + policies), avoiding inflation from non-policy mesh-scoped resources like DataplaneInsight or ServiceInsight. Subtle detail done right.

pkg/metrics/store/counter_test.go
```diff
@@ -135,4 +135,31 @@ var _ = Describe("Counter", func() {
 			g.Expect(findGauge("TrafficPermission").GetValue()).To(Equal(float64(2)))
 		}).Should(Succeed())
 	})
+
```
Automaat, line 140: Good test coverage — the new test correctly isolates the no-insight scenario by creating mesh resources without running the resyncer, and validates that counts propagate.

pkg/metrics/store/counter.go
Automaat, line 133: Consider adding a debug log when the fallback activates — helps operators understand store load differences between CP modes:
```go
if len(insights.Items) == 0 {
    log.V(1).Info("no MeshInsight found, counting mesh-scoped resources from store directly")
    return s.countMeshScopedResourcesWithoutInsights(ctx, resourceCount)
}
```

pkg/metrics/store/counter_test.go
```diff
@@ -135,4 +135,31 @@ var _ = Describe("Counter", func() {
...
+	It("should count dataplanes and policies without mesh insights", func() {
+		// given
+		err := resManager.Create(context.Background(), core_mesh.NewMeshResource(), core_store.CreateByKey("mesh-1", model.NoMesh))
+		Expect(err).ToNot(HaveOccurred())
+
```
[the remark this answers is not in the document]
officialasishkumar: Switched the test shutdown to `close(stop)` so both goroutines exit, and added `GinkgoRecover()` in each goroutine to match the rest of the test suite. Included in 0dfc3a4.

pkg/metrics/store/counter.go
officialasishkumar: Thanks for the observation. You're right that `len(insights.Items) == 0` conflates both cases. Since a CP-mode check would require threading additional config into the store counter (it currently only has a `ReadOnlyResourceManager`), I've added a comment explaining the trade-off: this covers zone CPs (no MeshInsight resources) and the harmless zero-mesh global CP case.

pkg/metrics/store/counter.go
officialasishkumar: Good point about the GC pressure at scale. Since this matches the existing `countGlobalScopedResources` pattern and isn't a regression from this PR, I think a follow-up issue for a `Count()` store API would be the right approach. Would you like me to open that issue?

pkg/metrics/store/counter_test.go
officialasishkumar: Thanks!

pkg/metrics/store/counter.go
officialasishkumar: Good suggestion -- added the debug log along with a comment explaining the two scenarios covered by this check. See 39d301d.

officialasishkumar:
@lahabana I narrowed the no-`MeshInsight` fallback to dataplanes only and rebased the branch on current `master`. That keeps the fix aligned with #15895 while avoiding the per-policy scan that was adding most of the extra store work.

Commit: 89b803cf7

Issue #13047, opened 2025-03-11
Remove deprecations for resources whose name breaks RFC-1035


Icarus9913:
### Description

In 2.10.x we add deprecation for the resources names that break RFC-1035. And I gonna clean up in 2.12.x

xref: https://github.com/kumahq/kuma/pull/13003

lahabana:
This is stuck because of Mink so removing the 2.13 milestone

lahabana:
Is this any different?

lobkovilya:
@lahabana I think we need to add a flag to 2.14 for v3 readiness. When enabled, instead of deprecations we will return real validation errors.

lahabana:
I'm moving this to 3.0 the deprecation is here already.

lahabana:
#12825 should only warn about deprecation. This should also ensure this deprecation becomes a failure in 3.0.

PR #16127, opened 2026-04-07
docs(MADR): improve SNI format

### LINKED ISSUE #16041: MADR on SNI format considerations
### Description

Current format has a few limitations/issues:
* it does not include the zone, which can lead to collisions when MeshExternalService resources with the same name are both synced from the global control plane and created locally in a zone
* it is opaque to users because it contains a hash, which makes it harder for integrations to construct it reliably
* has special handling for MeshServices (because MS is namespace-scoped and MES/MMZS are cluster-scoped) https://github.com/kumahq/kuma/blob/b7673621b320220ba52eadf5faddd1b4a378f718/pkg/xds/generator/zoneproxy/destinations.go#L58-L64

Given that we have now adopted [the new resource identifier format](https://github.com/kumahq/kuma/blob/48a29af2aa22fcdafe32c16dbf6f5db9d5d98efe/docs/madr/decisions/070-resource-identifier.md), KRI, this is a good opportunity to try to address these issues.

lobkovilya:
## Motivation

Adds MADR 101 proposing a KRI-derived SNI format (`sni.<type>.<mesh>.<zone?>.<namespace?>.<name>.<sectionName>`) to replace the current hash-based format (`<ver><hash>.<name>.<port>.<mesh>.<type>`).

The new format addresses three limitations: zone collisions between global-synced and zone-local `MeshExternalService` resources, the inability to recover a KRI from an observed SNI without a lookup table, and the `SNIName()` special-casing for `MeshService` vs other resource types.

Two options are evaluated — tweaking the hash inputs (Option 2) vs adopting a new format (Option 1) — and Option 1 is chosen on the strength of stateless SNI → KRI parsing, which unblocks using SNI in policy matchers.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/16041

### DIFF
--- a/docs/madr/decisions/101-sni-format-improvements.md
+++ b/docs/madr/decisions/101-sni-format-improvements.md
@@ -1,0 +1,170 @@
+# SNI Format Improvements
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/16041
+
+## Context and Problem Statement
+
+### Background
+
+SNI (Server Name Indication) is used in Kuma for cross-zone communication to route traffic through zone proxies.
+The current SNI format was introduced in MADR 055 with the following structure:
+
+```
+<format-version><hash>.<resource-name>.<port>.<mesh-name>.<resource-type>
+```
+
+For example:
+* `ae10a8071b8a8eeb8.backend.8080.demo.ms` - MeshService named `backend` in `demo` mesh
+* `ae10a8071b8a8eeb8.google.0.default.mes` - MeshExternalService named `google` in `default` mesh
+* `ae10a8071b8a8eeb8.backend.8080.demo.mzms` - MeshMultiZoneService
+
+This format was designed to:
+* Be a valid DNS hostname (avoiding characters like `,{/`)
+* Stay within length limits (max ~157 chars)
+* Support differentiation between resource types via suffixes (`ms`, `mes`, `mzms`)
+
+MADR 065 further decided to store the SNI directly in the resource definition
+rather than computing it dynamically.
+
+### Current limitations
+
+The current SNI format has several issues:
+
+#### 1. Zone is not included in the SNI
+
+The SNI does not include the zone where the resource was created.
+This can lead to naming collisions when `MeshExternalService` resources with the same name
+are both synced from the global control plane and created locally in a zone.
+
+For example,
+if zone `zone-a` and the global CP both have a MeshExternalService named `external-api`,
+their SNIs would be identical,
+causing routing conflicts.
+
+#### 2. Opaque hash makes integration difficult
+
+The current format includes a hash (`<format-version><hash>`) which:
+* Makes the SNI opaque to users who cannot easily construct it from known resource attributes
+* Complicates integrations that need to predict or construct SNIs programmatically
+* Reduces debuggability when troubleshooting routing issues
+
+The hash was originally introduced to handle uniqueness and tags,
+but with the current resource model where SNI is stored in the resource itself,
+this may no longer be necessary.
+
+#### 3. Inconsistent handling between resource types
+
+The code has special handling for MeshService vs MeshExternalService/MeshMultiZoneService
+when computing the resource name for SNI:
+
+```go
+// From pkg/xds/generator/zoneproxy/destinations.go
+switch r := any(dest).(type) {
+case *meshservice_api.MeshServiceResource:
+    rName = r.SNIName(systemNS)
+default:
+    rName = core_model.GetDisplayName(dest.GetMeta())
+}
+```
+
+MeshService uses a special `SNIName()` method that accounts for namespace scoping on Kubernetes
+and various resource origin scenarios:
+* Resources synced from global vs local zone resources
+* Kubernetes namespaced resources vs Universal resources
+* System namespace handling
+
+MeshExternalService and MeshMultiZoneService are cluster-scoped on Kubernetes
+and use a simpler `GetDisplayName()` approach,
+but this inconsistency makes the codebase harder to maintain and reason about.
+
+
+## Decision Drivers
+
+* **Zone-awareness**:
+  SNIs must uniquely identify resources across zones to prevent routing collisions,
+  especially for MeshExternalService resources that may exist both globally and locally
+
+* **Human-readability**:
+  Users and integrations should be able to understand and construct SNIs from known resource attributes
+  without needing to compute hashes
+
+* **Consistency**:
+  The format should handle MeshService, MeshExternalService, and MeshMultiZoneService uniformly
+  without special-case code paths
+
+* **Valid DNS hostname**:
+  SNI must remain a valid DNS hostname per RFC 6066
+
+* **Alignment with KRI**:
+  Consider how the SNI format can align with or leverage the KRI format adopted in MADR 070
+
+## Design
+
+### Option 1: New SNI format derived from KRI
+
+#### Format
+
+Creating a mapping of all possible KRIs to SNI is tricky and doesn't make much sense.
+Main problem is that KRI allows empty segments i.e. `kri_m____default_` is valid KRI,
+however DNS hostnames don't allow empty segments, i.e. `kri.m....default.` is not a valid hostname.
+
+The problem can be solved for a subset of KRIs that satisfy the constraints:
+* resource types are `MeshService`, `MeshExternalService` and `MeshMultiZoneService`
+* `mesh`, `name` and `sectionName` are always non-empty
+* if `namespace` is non-empty then `zone` is non-empty 
+
+In that case, the KRI `kri_<type>_<mesh>_<zone>_<namespace>_<name>_<sectionName>` can be unambiguously mapped to
+`sni.<type>.<mesh>.<zone>.<namespace>.<name>.<sectionName>` omitting the empty segment.
+
+It's unambiguous because different cases have different number of segments:
+* global-originated resource - `sni.<type>.<mesh>.<name>.<sectionName>` (5 segments)
+* zone-originated resource in system namespace - `sni.<type>.<mesh>.<zone>.<name>.<sectionName>` (6 segments)
+* zone-originated resource in custom namespace - `sni.<type>.<mesh>.<zone>.<namespace>.<name>.<sectionName>` (7 segments)
+
+#### DNS hostname limitation
+
+[RFC 1035](https://datatracker.ietf.org/doc/html/rfc1035) enforces limitations:
+* Each segment (DNS label) must be 63 characters or less
+* Total length of the hostname must be 253 characters or less
+
+There is no issue with each segment being 63 or less, however, total length is the issue.
+
+The max length of the SNI would be:
+* sni (3 chars)
+* type (6 chars because of the "extsvc")
+* mesh, zone, namespace, name, section name (63 chars each)
+* dots (6 chars)
+
+Total: 3 + 6 + 63*5 + 6 = 330
+
+So why even consider this option even though it exceeds the DNS hostname limit?
+I think there are few reasons to that:
+
+* Single mesh future. We've discussed multiple times that single mesh should the way to run Kuma. 
+In that case instead of 63 the mesh is likely to be 7 ("default").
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,309 @@
...
+- Makes the SNI opaque — users cannot construct it from known resource attributes
+- Makes it impossible to reverse an SNI back into a KRI, since the hash is a lossy transformation
+- Complicates integrations that need to map between SNIs and resource identifiers programmatically
+- Reduces debuggability when troubleshooting routing issues
+
+A bidirectional SNI <-> KRI mapping would allow components to resolve a resource
+from an observed SNI (e.g. in Envoy logs or access logs) without needing a lookup table.
+
+The hash was originally introduced to handle uniqueness and tags,
+but with the current resource model where SNI is stored in the resource itself,
```
lukidzi: Do we store SNI in the resource? I don't think it's true

docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,309 @@
...
+
+- Each segment (DNS label) must be 63 characters or less
+- Total length of the hostname must be 253 characters or less
+
+There is no issue with each segment being 63 or less, however, total length is the issue.
+
+The max length of the SNI would be:
+
+- sni (3 chars)
+- type (6 chars because of the "extsvc")
```
slonka: not a big deal but maybe good to call out explicitly:

```suggestion
- type (uses short names rather than the current SNI abbreviations `ms`, `mes`, `mzms` - 6 chars because of the "extsvc")
```

docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,309 @@
...
+In that case instead of 63 the mesh is likely to be 7 ("default").
+- 63 char for port name is way too much (even though Service allows port names upto 63, port name on Container is only 15 chars max)
+
+We don't need to limit individual segments, however,
+when user creates `MeshService`, `MeshExternalService` or `MeshMultiZoneService`,
+we can validate the length of `mesh`, `zone`, `namespace`, `name` and `port` + 15 (for "sni", "type" and dots) less than 253.
+
+#### Migration
+
+Only new mesh-scoped zone proxies adopting a new format.
```
slonka, line 169: Is this that simple? What happens during a rolling upgrade when some zone proxies use the old format and others use the new?

docs/madr/decisions/101-sni-format-improvements.md
slonka: Doesn't look like it https://github.com/kumahq/kuma/blame/master/pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go#L55

docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,309 @@
...
+For example,
+if zone `zone-a` and the global CP both have a MeshExternalService named `external-api`,
+their SNIs would be identical,
+causing routing conflicts.
+
+#### 2. Opaque hash prevents SNI <-> KRI conversion
+
+The current format includes a hash (`<format-version><hash>`) which:
+
+- Makes the SNI opaque — users cannot construct it from known resource attributes
```
slonka: nitpick: I think this is a bit overstated. The hash uses public inputs, no secrets. Anyone who reimplements the hash logic can construct the SNI.

docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,309 @@
...
+Main problem is that KRI allows empty segments i.e. `kri_m____default_` is valid KRI,
+however DNS hostnames don't allow empty segments, i.e. `kri.m....default.` is not a valid hostname.
+
+However, the problem can be solved for a subset of KRIs that satisfy the constraints:
+
+- resource types are `MeshService`, `MeshExternalService` and `MeshMultiZoneService`
+- `mesh`, `name` and `sectionName` are always non-empty
+- if `namespace` is non-empty then `zone` is non-empty
+
+These are reasonable constraints that follow naturally from Kuma's resource model.
```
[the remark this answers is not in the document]
slonka: I think this is just the MADR. `sectionName` is never empty in the SNI context because the `Port` interface contract guarantees it:

https://github.com/kumahq/kuma/blob/master/pkg/core/resources/apis/core/interfaces.go#L10-L11

docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,309 @@
...
+
+In that case, the KRI `kri_<type>_<mesh>_<zone>_<namespace>_<name>_<sectionName>` can be unambiguously mapped to
+`sni.<type>.<mesh>.<zone>.<namespace>.<name>.<sectionName>` omitting the empty segment.
+
+It's unambiguous because different cases have different number of segments:
+
+- global-originated resource - `sni.<type>.<mesh>.<name>.<sectionName>` (5 segments)
+- zone-originated resource in system namespace - `sni.<type>.<mesh>.<zone>.<name>.<sectionName>` (6 segments)
+- zone-originated resource in custom namespace - `sni.<type>.<mesh>.<zone>.<namespace>.<name>.<sectionName>` (7 segments)
```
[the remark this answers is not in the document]
slonka: `NameCharacterSet` allows `.` - https://github.com/kumahq/kuma/blob/master/pkg/core/resources/apis/mesh/validators.go#L26

slonka (commented):
Are there any security or reliability implications. Anything specific for Kong Mesh?


I think we need to figure out this https://github.com/kumahq/kuma/pull/16127#discussion_r3044689800 before moving forward.

docs/madr/decisions/101-sni-format-improvements.md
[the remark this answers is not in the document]
lobkovilya: Validation for Mesh*Services resource is different https://github.com/kumahq/kuma/issues/13047, today we just return deprecation messages but in v3 names will be fully DNS label compatible.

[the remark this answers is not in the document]
slonka: Worth calling out this dependency in the MADR though - something like "Option 1 requires DNS 1035-compliant names, enforced via #13047." Otherwise a reader looking at `NameCharacterSet` today would reasonably conclude dots are allowed.

docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,309 @@
...
+construct an SNI from a KRI, and recover a KRI from an observed SNI —
+without requiring a lookup table
+- **Consistency**:
+The format should handle MeshService, MeshExternalService, and MeshMultiZoneService uniformly
+without special-case code paths
+- **Valid DNS hostname**:
+SNI must remain a valid DNS hostname per [RFC 6066](https://datatracker.ietf.org/doc/html/rfc6066)
+- **Alignment with KRI**:
+Consider how the SNI format can align with or leverage the KRI format adopted in MADR 070
+
```
lahabana, line 102: backward compatibility (I think this is why we had a version prefxi)

lahabana (commented):
Intuitively makes sense to me! I haven't read in deep enough details to approve but I also don't want to block you

docs/madr/decisions/101-sni-format-improvements.md
lahabana: nevermind this is explained lower down (no need because it's only for new proxies)

docs/madr/decisions/101-sni-format-improvements.md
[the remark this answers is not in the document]
lobkovilya: Add a note in the doc

docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,309 @@
...
+So why even consider this option even though it exceeds the DNS hostname limit?
+I think there are few reasons to that:
+
+- Single mesh future. We've discussed multiple times that single mesh should the way to run Kuma. 
+In that case instead of 63 the mesh is likely to be 7 ("default").
+- 63 char for port name is way too much (even though Service allows port names upto 63, port name on Container is only 15 chars max)
+
+We don't need to limit individual segments, however,
+when user creates `MeshService`, `MeshExternalService` or `MeshMultiZoneService`,
+we can validate the length of `mesh`, `zone`, `namespace`, `name` and `port` + 15 (for "sni", "type" and dots) less than 253.
```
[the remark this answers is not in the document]
lobkovilya: It's the same principle as with constructing the KRI. If port's name is not specified then numeric value will be used.

docs/madr/decisions/101-sni-format-improvements.md
lobkovilya: Removed the line

docs/madr/decisions/101-sni-format-improvements.md
lobkovilya: New mesh-scoped zone proxies always use the new format. We never mix old and new zone proxies in the same envoy cluster.

lukidzi (dismissed):

docs/madr/decisions/101-sni-format-improvements.md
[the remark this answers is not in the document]
slonka: nitpick: shouldn't it say 


```suggestion
we can validate the length of `mesh`, `zone`, `namespace`, `name` and `sectionName` + 15 (for "sni", "type" and dots) less than 253.
```

?

docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,312 @@
...
+  (no risk of exceeding DNS hostname limits)
+- Bad, because SNI → KRI reversal is impossible —
+  the hash is lossy,
+  so an observed SNI cannot be mapped back to a resource without a lookup
+- Bad, because users and integrations still cannot construct or predict SNIs
+  from resource attributes
+- Bad, because debugging routing issues still requires looking up the stored SNI
+  rather than reasoning about it from the resource name
+
+## Decision
```
slonka, line 293: Are there any security or reliability implications. Anything specific for KM?

slonka (dismissed):

lobkovilya:
Before merging, I’d like to make sure we’re aligned on the [limitations](https://github.com/kumahq/kuma/pull/16127/changes#diff-3ea12a6bdf4569d12c202ded111a25004f566bbc7061244a4ffb30bd103169beR137-R164) of the chosen approach.

It also seems this decision is largely driven by https://github.com/kumahq/kuma/pull/16275. If SNI weren’t part of the policy matchers, we’d likely be less willing to accept these trade-offs.

slonka:
@lahabana please take a look at https://github.com/kumahq/kuma/pull/16127#issuecomment-4287552752

docs/madr/decisions/101-sni-format-improvements.md
```diff
@@ -0,0 +1,312 @@
...
+So why even consider this option even though it exceeds the DNS hostname limit?
+I think there are few reasons to that:
+
+- Single mesh future. We've discussed multiple times that single mesh should be the way to run Kuma. 
+In that case instead of 63 the mesh is likely to be 7 ("default").
+- 63 char for port name is way too much (even though Service allows port names upto 63, port name on Container is only 15 chars max)
+
+We don't need to limit individual segments, however,
+when user creates `MeshService`, `MeshExternalService` or `MeshMultiZoneService`,
+we can validate the length of `mesh`, `zone`, `namespace`, `name` and `sectionName` + 15 (for "sni", "type" and dots) less than 253.
```
lahabana: When would we validate this?

Would be good to clarify it

lobkovilya: Updated the doc

docs/madr/decisions/101-sni-format-improvements.md
lobkovilya: Updated the doc to mention no implication for Kong Mesh

lukidzi (approved):

slonka (approved):

PR #16165, opened 2026-04-08
fix(xds): eliminate TOCTOU race in OnProxyDisconnected

### LINKED ISSUE #16162: fix(xds): TOCTOU race in dataplaneSyncTracker.OnProxyDisconnected
## Motivation

`OnProxyDisconnected` in `pkg/xds/server/callbacks/dataplane_sync_tracker.go:68-81` has a TOCTOU race. It reads the watchdog entry under `RLock`, releases the lock, then later acquires `Lock` to delete. In the gap (lines 71-77), `OnProxyConnected` can replace the entry for the same `dpKey`. The subsequent `delete` then removes the **new** watchdog, leaking its goroutine.

```go
t.RLock()
dpData := t.watchdogs[dpKey]  // reads old entry
t.RUnlock()                    // gap opens

if dpData != nil {
    dpData.cancelFunc()        // cancels old watchdog
    <-dpData.stopped           // waits for old watchdog (OnProxyConnected can run here)
    t.Lock()
    delete(t.watchdogs, dpKey) // deletes NEW entry
}
```

This is a logical race (not a data race), so `-race` won't catch it. Under high DP churn (rapid connect/disconnect), this leaks goroutines and causes CP instability.

## Implementation information

Hold the write lock for the entire disconnect operation instead of upgrading from RLock to Lock:

```go
func (t *dataplaneSyncTracker) OnProxyDisconnected(...) {
    t.Lock()
    dpData := t.watchdogs[dpKey]
    delete(t.watchdogs, dpKey)
    t.Unlock()

    if dpData != nil {
        dpData.cancelFunc()
        <-dpData.stopped
    }
}
```

This eliminates the TOCTOU window. The blocking `<-dpData.stopped` happens after releasing the lock so it doesn't hold up other operations.

Automaat:
## Motivation

`OnProxyDisconnected` in `pkg/xds/server/callbacks/dataplane_sync_tracker.go` had a TOCTOU race. It read the watchdog entry under `RLock`, released the lock, then acquired `Lock` to delete. In the gap, `OnProxyConnected` could replace the entry for the same `dpKey`. The subsequent `delete` then removed the **new** watchdog, leaking its goroutine.

Under high DP churn (rapid connect/disconnect), this leaks goroutines and causes CP instability. The race is a logical race (not data race), so `-race` won't catch it.

## Implementation information

Hold the write lock for the entire read+delete operation instead of upgrading from RLock to Lock:

```go
t.Lock()
dpData := t.watchdogs[dpKey]
delete(t.watchdogs, dpKey)
t.Unlock()
// blocking wait happens outside the lock
```

The `<-dpData.stopped` wait is intentionally outside the lock to avoid stalling other callbacks.

Added a regression test that directly calls `OnProxyConnected`/`OnProxyDisconnected` on the tracker (bypassing `xdsCallbacks`, which serialises connect/disconnect via `activeStreams` and would prevent reproducing the race).

## Supporting documentation

Fix #16162

> Changelog: fix(xds): eliminate TOCTOU race in OnProxyDisconnected

### DIFF
--- a/pkg/xds/server/callbacks/dataplane_sync_tracker.go
+++ b/pkg/xds/server/callbacks/dataplane_sync_tracker.go
@@ -66,16 +66,14 @@ func (t *dataplaneSyncTracker) OnProxyConnected(streamID core_xds.StreamID, dpKe
 }
 
 func (t *dataplaneSyncTracker) OnProxyDisconnected(_ context.Context, streamID core_xds.StreamID, dpKey core_model.ResourceKey) {
-	t.RLock()
+	t.Lock()
 	dpData := t.watchdogs[dpKey]
-	t.RUnlock()
+	delete(t.watchdogs, dpKey)
+	t.Unlock()
 
 	if dpData != nil {
 		dpData.cancelFunc()
 		<-dpData.stopped
 		dataplaneSyncTrackerLog.V(1).Info("watchdog for a Dataplane stopped", "dpKey", dpKey, "streamID", streamID)
-		t.Lock()
-		defer t.Unlock()
-		delete(t.watchdogs, dpKey)
 	}
 }
--- a/pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
+++ b/pkg/xds/server/callbacks/dataplane_sync_tracker_test.go
@@ -192,6 +192,71 @@ var _ = Describe("Sync", func() {
 			Expect(activeWatchdogs.Load()).To(Equal(int32(0)))
 		})
 
+		It("should not delete new watchdog when old disconnect races with new connect (TOCTOU)", test.Within(10*time.Second, func() {
+			// afterCancel is signalled when a watchdog's ctx is cancelled but before it finishes.
+			// barrier controls when the watchdog goroutine is actually allowed to complete.
+			// This lets us force the interleaving: OnProxyConnected for stream 2 runs while
+			// stream 1's OnProxyDisconnected has already released the lock (delete done) but
+			// is still blocking on <-dpData.stopped — exactly the original TOCTOU window.
+			//
+			// Note: this test calls the tracker directly (not through xdsCallbacks) because
+			// xdsCallbacks.activeStreams serialises connect/disconnect and would prevent the
+			// concurrent calls needed to reproduce the race.
+			afterCancel := make(chan struct{}, 1)
+			barrier := make(chan struct{})
+			var activeWatchdogs atomic.Int32
+
+			tracker := NewDataplaneSyncTracker(sync.DataplaneWatchdogFactoryFunc(func(key core_model.ResourceKey, _ *core_xds.DataplaneMetadata) util_xds_v3.Watchdog {
+				return util_xds_v3.WatchdogFunc(func(ctx context.Context) {
+					activeWatchdogs.Add(1)
+					<-ctx.Done()
+					select {
+					case afterCancel <- struct{}{}:
+					default:
+					}
+					<-barrier
+					activeWatchdogs.Add(-1)
+				})
+			}))
+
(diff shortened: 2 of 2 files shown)


pkg/xds/server/callbacks/dataplane_sync_tracker.go
```diff
@@ -66,16 +66,14 @@ func (t *dataplaneSyncTracker) OnProxyConnected(streamID core_xds.StreamID, dpKe
 }
 
 func (t *dataplaneSyncTracker) OnProxyDisconnected(_ context.Context, streamID core_xds.StreamID, dpKey core_model.ResourceKey) {
-	t.RLock()
+	t.Lock()
```
lobkovilya, line 69: Looks like the implementation with RWLock was intentional https://github.com/kumahq/kuma/pull/12886/changes#r1965422081, but I'm not sure yet why is that...

Automaat: @lobkovilya The RLock was intentional to avoid holding a write lock during the blocking `<-dpData.stopped` wait, which would stall all other callbacks.

The fix in this PR doesn't reintroduce that problem. The key difference is the `delete` happens **before** the blocking wait, not after:

```go
t.Lock()
dpData := t.watchdogs[dpKey]
delete(t.watchdogs, dpKey)
t.Unlock()
// blocking wait happens outside the lock
dpData.cancelFunc()
<-dpData.stopped
```

So the write lock is held only for the fast map read+delete, and `<-dpData.stopped` still runs outside the lock — same non-blocking property Icarus intended, but without the TOCTOU race.

lahabana: Isn't it a problem that now the shutdown of the DP may happen while another DP with the same key is starting?

Automaat: This overlap also existed with the RLock approach: between `RUnlock` and the final `Lock+delete`, `OnProxyConnected` could fire and start a new watchdog. Old and new watchdogs are independent (separate `cancelFunc` + `stopped` channels), so the transient overlap is safe.

The current fix is strictly better — it eliminates the overlap AND prevents the TOCTOU where PR 12886's final `delete` would remove the newly connected watchdog's state.

lobkovilya (commented):
**There is no TOCTOU race.** Between RLock and Lock no one can update `t.watchdogs` because it's guaranteed `OnProxyConnected` can be called only once per DPP ResourceKey https://github.com/kumahq/kuma/blob/4c106439271ad9efb026edeaf43c726d4fcdd3bd/pkg/xds/server/callbacks/dataplane_callbacks.go#L157-L161

However, I think your change improves readability. Though, you probably need to replace `syncstd.RWMutex` with `syncstd.Mutex` since we don't call RLock now.

Automaat:
Closing — agreeing with @lobkovilya, no race exists.

`xdsCallbacks.onStreamRequest` (dataplane_callbacks.go:136-161) rejects any new stream for a `dpKey` whose `activeStreams` entry is still set. That entry is only cleared in `onStreamClosed` **after** `OnProxyDisconnected` returns (line 97 vs 100-105). So during the entire RLock→Lock window in `OnProxyDisconnected`, any concurrent connect for the same dpKey gets the "already an active stream, try again later" error before reaching the tracker. `OnProxyConnected(dpKey)` cannot run concurrently with `OnProxyDisconnected(dpKey)`.

The PR's regression test only reproduces the "race" by calling the tracker directly and bypassing `xdsCallbacks` — i.e. bypassing the invariant that prevents the bug. That's not a race, that's an unreachable interleaving.

Closing #16162 as not-a-bug. If anyone wants the readability win (single critical section, `RWMutex`→`Mutex`), happy to take a small refactor PR with honest framing — no test that bypasses the wrapper.

Issue #10952, opened 2024-07-23
Have validation webhook reject applying/modify resources on global


lahabana:
We should be better at this inded.
 
              Yes, you are correct. You should change the object only in the source of truth which is Zone CP.

Let's double-check our validation. We should prevent users from doing this to avoid such issues.

_Originally posted by @jakubdyszkiewicz in https://github.com/kumahq/kuma/issues/10738#issuecomment-2218023694_

bartsmykla:
> triage: to be validated on the next Test Friday session if it's still the issue

Automaat:
Verified this is now handled in current master. Validation exists at both the REST API and the K8s admission webhook.

Reproduced on a global CP:
- PUT a ZoneIngress on global (the original #10738 case) returns 405 with a message telling you to modify it on the zone CP instead.
- Creating a resource with `kuma.io/origin: zone` on global returns 400: the origin label must be set to 'global'.
- A normal global write still works.

The kubectl edit path is covered by the admission webhook (403 when origin is zone). Closing as resolved.

PR #16169, opened 2026-04-08
refactor(kds): use core_model.ZoneOfResource in global filter

### LINKED ISSUE #16161: fix(kds): ZoneTag panics on Dataplane with empty inbound
## Motivation

`ZoneTag()` in `pkg/kds/util/resource.go:126` accesses `GetInbound()[0]` without bounds check. When a Dataplane has no gateway and no inbound listeners, this panics with `index out of range [0] with length 0`, crashing the control plane.

Confirmed via PoC test — a `Dataplane` with empty `Inbound` slice and no `Gateway` triggers the panic.

## Implementation information

Add a length check before accessing `[0]`:

```go
inbounds := res.GetNetworking().GetInbound()
if len(inbounds) == 0 {
    return ""
}
return inbounds[0].GetTags()[mesh_proto.ZoneTag]
```

Alternatively, consider replacing the function with `core_model.ZoneOfResource(r)` as the existing TODO at line 132 suggests.

Automaat:
## Motivation

`ZoneTag()` in `pkg/kds/util/resource.go` accesses `GetInbound()[0]` without a bounds check. When a `Dataplane` has no gateway and no inbound listeners (empty slice), this panics with `index out of range [0] with length 0`, crashing the control plane.

A pre-existing TODO at the only call site already suggested replacing the function with `core_model.ZoneOfResource(r)` (see #10952). `ZoneOfResource` reads `kuma.io/zone` from metadata labels (set during KDS sync) and works uniformly across resource types, so it removes the panic without needing a band-aid bounds check.

## Implementation information

- Replace `util.ZoneTag(r)` with `core_model.ZoneOfResource(r)` at the only call site in `pkg/kds/context/context.go`.
- Drop the now-unused `util.ZoneTag` helper.

## Supporting documentation

- #10952

> Changelog: refactor(kds): replace util.ZoneTag with core_model.ZoneOfResource

### DIFF
--- a/pkg/kds/util/resource.go
+++ b/pkg/kds/util/resource.go
@@ -123,7 +123,11 @@ func ZoneTag(r core_model.Resource) string {
 		if res.GetNetworking().GetGateway() != nil {
 			return res.GetNetworking().GetGateway().GetTags()[mesh_proto.ZoneTag]
 		}
-		return res.GetNetworking().GetInbound()[0].GetTags()[mesh_proto.ZoneTag]
+		inbounds := res.GetNetworking().GetInbound()
+		if len(inbounds) == 0 {
+			return ""
+		}
+		return inbounds[0].GetTags()[mesh_proto.ZoneTag]
 	case *mesh_proto.ZoneIngress:
 		return res.GetZone()
 	case *mesh_proto.ZoneEgress:
--- a/pkg/kds/util/resource.go
+++ b/pkg/kds/util/resource.go
@@ -117,27 +117,6 @@ func ResourceNameHasAtLeastOneOfPrefixes(resName string, prefixes ...string) boo
 	return false
 }
 
-func ZoneTag(r core_model.Resource) string {
-	switch res := r.GetSpec().(type) {
-	case *mesh_proto.Dataplane:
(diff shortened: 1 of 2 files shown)


pkg/kds/util/resource.go
```diff
@@ -123,7 +123,11 @@ func ZoneTag(r core_model.Resource) string {
 		if res.GetNetworking().GetGateway() != nil {
 			return res.GetNetworking().GetGateway().GetTags()[mesh_proto.ZoneTag]
 		}
-		return res.GetNetworking().GetInbound()[0].GetTags()[mesh_proto.ZoneTag]
+		inbounds := res.GetNetworking().GetInbound()
+		if len(inbounds) == 0 {
```
lobkovilya, line 127: I don't really like this fix. `ZoneTag` method shouldn't be called at all when you're running Kuma with disabled tags, so maybe it worth going over `ZoneTag` usage and see if appropriate to replace it with `core_model.ZoneOfResource(r)` and potentially remove or deprecate `ZoneTag` method

lobkovilya (changes requested):

lobkovilya (approved):
please adjust the PR title and description

Issue #14039, opened 2025-07-24
remove kuma-dp readiness reporter `ReadinessUnixSocketDisabled` flag


Icarus9913:
### Description

In 2.12.0, we introduced an environment variable `ReadinessUnixSocketDisabled` and a feature flag `FeatureReadinessUnixSocket` for kuma-dp, which works for users who want to disable the kuma-dp readiness reporter via unix socket to keep the previous behavior for compatibility. 
Also, the Unix socket would be our default behavior for the readiness reporter. 

In 2.15, we should remove the flags:
 - ReadinessUnixSocketDisabled
 - FeatureReadinessUnixSocket
 - ReadinessPort
 
Also, we need to clean up the unit, e2e tests

Icarus9913:
xref: https://github.com/kumahq/kuma/pull/13924

lahabana:
@Automaat can we do this in the end?

Automaat:
@lahabana — yes, with a re-scope. Working through #16174 surfaced enough about the actual wiring that the original ask should be tightened, not abandoned.

### How it's wired today

There are two distinct ports, not a fallback:

- **Admin**: TCP `9901` (or UDS when `EnvoyAdminUnixSocket=true` — default since 2.12, see `pkg/config/xds/bootstrap/config.go:102`)
- **Readiness**: TCP `9902` (`KUMA_BOOTSTRAP_SERVER_PARAMS_READINESS_PORT`, default at `pkg/config/xds/bootstrap/config.go:103`)

`ReadinessPort` is a dedicated port for the kuma-dp readiness reporter, independent of admin. K8s probes always hit `9902`.

`ReadinessUnixSocketDisabled` controls only the *transport* of the readiness reporter itself:

- `true` → readiness on TCP `9902` (K8s mode with admin UDS — the default)
- `false` → readiness on UDS (Universal mode, no kubelet probes)

The K8s injector sets `KUMA_READINESS_UNIX_SOCKET_DISABLED=true` whenever `envoyAdminUnixSocket` is on (`pkg/plugins/runtime/k8s/containers/factory.go:308`).

### What #16174 actually shipped

Original removal commit was reverted mid-PR (`f606048511 fix(dp): restore TCP readiness fallback`); the PR was retitled `fix(dp): make readiness reporter dual-stack` and now ships only the dual-stack listener fix. The deprecation isn't in it.

### Proposed re-scope

[...]

PR #16174, opened 2026-04-08
fix(dp): make readiness reporter dual-stack

Automaat:
## Motivation

K8s pods on dual-stack clusters can have probes hit either the IPv4 or
IPv6 pod address. The kuma-dp readiness reporter previously bound
`tcp6` when the listen address was IPv6, which sets `IPV6_V6ONLY=1`
and refuses IPv4 probes — the probe would fail intermittently
depending on which family the kubelet picked.

## Implementation information

- `readiness/component.go`: bind `tcp` (Linux dual-stack when bound to
  `::`) instead of `tcp6` for the TCP readiness listener. For concrete
  addresses the family is determined by the address itself.
- `cmd/run.go`: create the `SocketDir` up front when an operator-supplied
  path doesn't exist, so component startup doesn't fail trying to bind
  Unix sockets inside it.
- `cmd/wait.go`: drop the `KUMA_READINESS_UNIX_SOCKET_DISABLED` /
  `KUMA_READINESS_PORT` URL-derivation block. The K8s injector PostStart
  hook already passes `--url` explicitly, so the env-var path was
  redundant in cluster. Universal-mode users invoking `kuma-dp wait`
  without flags will hit the default `http://localhost:9901/ready`
  (Envoy admin) — unchanged for them, since they already needed
  `--url` or `--unix-socket` for non-default ports.
- `tools/ci/api-linter/linter/testdata/...`: gofmt fixup picked up by
  `make check`.

The full removal of `ReadinessUnixSocketDisabled`, `ReadinessPort`,
`FeatureReadinessUnixSocket`, `FieldDataplaneReadinessPort`,
`GetReadinessPort()`, and the `KUMA_READINESS_UNIX_SOCKET_DISABLED`
env var is deferred to a separate PR per #14039 (target v2.15).

> Changelog: fix(dp): make readiness reporter dual-stack


app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -80,11 +80,15 @@ func (r *Reporter) Start(stop <-chan struct{}) error {
 	var lis net.Listener
 	var protocol, addr string
 	if r.unixSocketDisabled {
+		// Use "tcp" (not "tcp6") so that when localListenAddr is the IPv6
+		// wildcard "::", Linux gives a dual-stack listener that accepts
+		// both IPv4 and IPv6 probes. "tcp6" sets IPV6_V6ONLY and would
+		// refuse IPv4 probes. For concrete addresses the family is
+		// determined by the address itself.
+		protocol = "tcp"
```
bartsmykla, line 88: The dual-stack switch is the behavioral fix this PR ships, but there's no test exercising it. `component_test.go` already sits next to this file — please add a case that starts the reporter with `localListenAddr="::"` and dials both `127.0.0.1:<port>` and `[::1]:<port>` (skip when the host has no IPv6 loopback). Without coverage the next refactor regresses this silently.

app/kuma-dp/cmd/wait.go
```diff
@@ -29,20 +28,6 @@ func newWaitCmd() *cobra.Command {
...
-				if os.Getenv("KUMA_READINESS_UNIX_SOCKET_DISABLED") == "true" {
-					port := os.Getenv("KUMA_READINESS_PORT")
-					if port == "" {
-						port = "9902"
-					}
-					args.url = fmt.Sprintf("http://localhost:%s/ready", port)
-				}
-			}
-
 			client := httpclient.NewTCPOrUDS(args.unixSocket, args.requestTimeout, args.requestTimeout)
```
bartsmykla, line 31: PR body claims the K8s injector PostStart hook always passes `--url`, so the env-derivation block is redundant in cluster. Worth a quick grep of `pkg/plugins/runtime/k8s/webhooks/injector/` to confirm there's no path that still relies on `KUMA_READINESS_UNIX_SOCKET_DISABLED` / `KUMA_READINESS_PORT`. If any in-cluster caller hits `kuma-dp wait` without `--url`, this is a silent break — gate the old block behind a deprecation log rather than deleting it outright.

bartsmykla (commented):
The dual-stack fix needs test coverage, and the wait.go cleanup needs a quick sanity check on remaining callers

bartsmykla (dismissed):

bartsmykla (approved):

PR #16199, opened 2026-04-09
feat(tools): add CP perf testing scripts

Automaat:
## Motivation

Need a reusable, committed toolkit for running Kuma CP performance tests
against a real cluster and capturing profiles + metrics for offline
analysis. Has been used to drive the perf investigation tracked in #16183
and its subissues.

## Implementation information

Adds `tools/perf-test/` with five driver scripts and a README:

- `deploy-wave.sh` — deploy N namespaces x M services x R replicas of
  fake-service with sidecar injection enabled
- `collect-profiles.sh` — grab cpu, heap, goroutine, mutex, block, trace
  and a `/metrics` snapshot from one or more CP endpoints in parallel
- `run-waves.sh` — progressive load ramp (default: 50 → 2350 DPs),
  stabilize + profile after each wave
- `policy-scenarios.sh` — policy/KDS scenarios (spread, single-DP,
  updates, bulk delete) against an already-deployed fleet
- `dump-prometheus.sh` — pull Go runtime, xDS, KDS, and store metrics as
  Prometheus range queries (JSON) for offline analysis

All scripts are env-var-driven with sensible defaults for local k3d/kind
runs; `ENDPOINTS="zone=... global=..."` enables multi-zone profiling with
a prefix-per-CP on output files. `tools/perf-test/profiles/` is gitignored
so results stay local.

Alternatives considered: keep the scripts in a gist or in a private repo.
Rejected because the perf investigation that produced the umbrella issue
needs to be reproducible by anyone on future releases; a committed,
documented toolkit makes that practical.

## Supporting documentation

- Umbrella issue: #16183
- README with usage and analysis hints: `tools/perf-test/README.md`

> Changelog: skip

### DIFF
--- a/tools/perf-test/.gitignore
+++ b/tools/perf-test/.gitignore
@@ -1,0 +1,1 @@
+profiles/
--- a/tools/perf-test/README.md
+++ b/tools/perf-test/README.md
@@ -1,0 +1,126 @@
+# Perf Test
+
+Scripts for running Kuma control plane performance tests on an existing cluster
+(typically k3d / kind for local runs, GKE / EKS for real scale).
+
+The scripts capture [pprof](https://pkg.go.dev/net/http/pprof) profiles and
+Prometheus metrics so you can compare Kuma CP behaviour across releases, configs,
+or scale points.
+
+## Contents
+
+| Script                 | Purpose                                                           |
+| ---------------------- | ----------------------------------------------------------------- |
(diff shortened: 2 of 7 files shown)


lobkovilya (changes requested):
Unless these scripts are meant to run regularly in CI, I don’t think they belong in the main repo. You could revive `Kong/mesh-perf` and put them there, or create a separate repo under kumahq.

PR #16200, opened 2026-04-09
perf(k8s): cache labels + spec per resourceVersion

Automaat:
## Motivation

Fixes #16191.

`KubernetesMetaAdapter.GetLabels` ran `maps.Clone` on every call plus an annotation-merge into the cloned map. `DestinationIndex` invokes it per destination per reconcile, producing ~68 MB of label-map churn per wave in the referenced profile. The existing `cachingConverter` only cached the unmarshaled spec, so the adapter and its labels were rebuilt on every cache hit.

## Implementation information

- `KubernetesMetaAdapter` gets a `cachedLabels` field + `sync.Once`. First `GetLabels` call computes once; subsequent calls return the same map pointer. Adapter is always used via pointer (verified) so `sync.Once` is safe.
- `cachingConverter` now caches `(spec, labels)` together in a `cachedEntry` struct keyed by the same `namespace:name:resourceVersion:gvk`. On hit, it pre-populates the new adapter's `cachedLabels` so the first `GetLabels` skips the Once closure and returns the shared map directly. The adapter is still reallocated per call (cheap) but the expensive labels map is now shared across reconciles for the lifetime of the cache entry.
- **Status is still refetched on every call** — preserved the contract asserted by the existing test that verifies status subresource mutations surface on cache hits.
- **Read-only labels contract** — documented on `GetLabels`. Audit of call sites in `pkg/xds/context/destination_index.go`, `pkg/xds/topology/outbound.go`, `pkg/xds/generator/{dns,ingress}_generator.go`, `pkg/plugins/policies/core/rules/resolve/targetref.go`, and `pkg/xds/server/callbacks/*` confirms every caller treats the map as read-only. Mutation canary test (60 samples over 60s against a live cluster) observed zero drift.
- No invalidation wiring needed: `resourceVersion` is part of the cache key, so any K8s update mints a fresh miss. Existing `MarshalingCacheExpirationTime` TTL bounds memory.

### Tests

Unit tests extended in `pkg/plugins/resources/k8s/caching_converter_test.go`:
- Adapter memoizes labels across repeat `GetLabels` calls (pointer equality).
- Same labels map pointer across cache hits for the same `resourceVersion`.
- Different `resourceVersion` produces a fresh map; stale label values don't leak.
- Pre-existing status-freshness tests still pass unchanged.

### Manual verification on k3d (kuma-2 cluster)

Ran a 10-test correctness suite covering create/update/delete propagation, display-name override + fallback, status freshness, mutation canary, webhook validation, and dp bootstrap. All passed except T-05 label routing, which surfaced a pre-existing Envoy-rendering issue (empty `weightedClusters`) unrelated to this change — the `_rules` endpoint confirmed label-based resolution itself works correctly. No panics or data races in 15 minutes of CP logs. `go test -race ./pkg/plugins/resources/k8s/...` and `./pkg/xds/context/...` green.

## Supporting documentation

- Issue: #16191
- Parent umbrella: #16183

> Changelog: perf(k8s): cache labels + spec per resourceVersion

### DIFF
--- a/pkg/plugins/resources/k8s/caching_converter.go
+++ b/pkg/plugins/resources/k8s/caching_converter.go
@@ -1,6 +1,7 @@
 package k8s
 
 import (
+	"maps"
 	"strings"
 	"time"
 
@@ -53,12 +54,14 @@ func (c *cachingConverter) ToCoreResource(obj k8s_model.KubernetesObject, out co
 	}, ":")
 	if v, ok := c.cache.Get(key); ok {
 		entry := v.(cachedEntry)
-		// Pre-populate cachedLabels so the adapter's GetLabels skips the
-		// maps.Clone + annotation-merge work for the lifetime of the entry.
+		// Pre-populate cachedLabels with a fresh clone so the adapter's
+		// GetLabels skips the annotation-merge work, while keeping the cached
+		// map isolated from downstream consumers that mutate labels in place
+		// (e.g. removeDisplayNameLabel in the ServiceInsight endpoints).
 		out.SetMeta(&KubernetesMetaAdapter{
 			ObjectMeta:   *obj.GetObjectMeta(),
 			Mesh:         obj.GetMesh(),
-			cachedLabels: entry.labels,
+			cachedLabels: maps.Clone(entry.labels),
 		})
 		if err := out.SetSpec(entry.spec); err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


pkg/plugins/resources/k8s/caching_converter.go
```diff
@@ -33,18 +45,25 @@ func NewCachingConverter(expirationTime time.Duration) k8s_common.Converter {
 }
 
 func (c *cachingConverter) ToCoreResource(obj k8s_model.KubernetesObject, out core_model.Resource) error {
```
bartsmykla, line 48: Can we override `ToCoreList` here too? `KubernetesStore.List` goes through `Converter.ToCoreList`, and `SimpleConverter.ToCoreList` calls `SimpleConverter.ToCoreResource` directly. As written, list reads still rebuild spec and labels for every item, so the `mesh_context_builder.fetchResourceList()` / `DestinationIndex` path never touches this cache.

pkg/plugins/resources/k8s/caching_converter.go
```diff
@@ -75,7 +96,10 @@ func (c *cachingConverter) ToCoreResource(obj k8s_model.KubernetesObject, out co
 	if obj.GetResourceVersion() != "" {
 		// an absence of the ResourceVersion means we decode 'obj' from webhook request,
 		// all webhooks use SimpleConverter, so this is not supposed to happen
-		c.cache.SetDefault(key, out.GetSpec())
+		c.cache.SetDefault(key, cachedEntry{
+			spec:   out.GetSpec(),
+			labels: adapter.GetLabels(), // materialize labels so future hits reuse them
```
bartsmykla: Please clone the labels map before caching it. `rest.From.Resource` forwards `GetLabels()` into REST metadata, and `removeDisplayNameLabel` deletes `kuma.io/display-name` in place for `ServiceInsight` GET/LIST. After a single request, later cache hits for the same `resourceVersion` will see that label as already missing.

bartsmykla (changes requested):
I think this still needs two fixes before it lands. The cache does not cover list reads yet, and the cached labels map is shared with code that mutates it in place.

pkg/plugins/resources/k8s/caching_converter.go
Automaat: Good catch — fixed in ceeb459. Added a `cachingConverter.ToCoreList` override that loops over items and dispatches through `c.ToCoreResource`, so list reads now hit the cache. Confirmed by a regression test that mutates the source labels between two list calls and asserts the second call returns the cached values.

pkg/plugins/resources/k8s/caching_converter.go
Automaat: Fixed in ceeb459, but I went with cloning on **both** sides (write *and* read), not just write — single-side cloning leaves a window where the cache still gets corrupted. Reasoning:

- **Write only**: cache holds map M2 = clone(M1). First consumer of M1 mutates → cache OK ✓. But the next cache hit wires M2 into a new adapter via `cachedLabels: entry.labels`. That consumer mutates M2 → cache corrupted on the second access ✗.
- **Both sides**: cache holds an isolated M2; every cache hit clones M2 → M3 into the new adapter. Cache map is never returned to any caller, so no in-place mutation can ever reach it.

Cost is one extra `maps.Clone` per access, but the annotation-merge work is still skipped, which was the point of the optimization.

Added a regression test that simulates `removeDisplayNameLabel` by deleting `kuma.io/display-name` and `app` from the returned labels across three consecutive conversions, and asserts the third call still sees the pristine entry.

bartsmykla (approved):

PR #16201, opened 2026-04-09
feat(metrics): add operational metrics

Automaat:
## Motivation

Operators have limited visibility into Kuma's internal behavior. This adds
low-cardinality metrics across the key subsystems to enable dashboards and
alerting without index explosion.

Metrics added:

**DNS proxy** (`app/kuma-dp`):

- `dns_queries_total{qtype,source}` — query count by type (A/AAAA/other) and
  resolution path (local/upstream)
- `dns_response_codes_total{rcode}` — response codes (noerror/nxdomain/
  servfail/other)
- `dns_entries_total` — current DNS map size (gauge)

**VIP allocator**:

- `vip_allocation_exhaustion_total{mesh}` — IPAM pool exhaustion events

**xDS**:

- `xds_cert_expiration_remaining{mesh}` — seconds until cert expiry
  (MeshIdentity/WorkloadIdentity path only)
- `xds_snapshot_resources_total{resource_type}` — snapshot resource counts

**KDS** (multi-zone):

- `kds_zone_active_connections{zone_name}` — active KDS streams per zone
- `kds_nack_total{zone_name,resource_type}` — NACK count by zone and type

**Sidecar injection** (K8s):

- `sidecar_injection_total{result,reason}` — injection decisions
  (success/skipped/failed) with reason

## Implementation information

All metrics use label cardinality bounded to single digits. Cert expiry metrics
are scoped strictly to the MeshIdentity/WorkloadIdentity code path.

`kds_server.New` now returns `*Metrics` alongside the delta server so callers
share a single registered instance — avoids duplicate registration across the
GlobalToZone and ZoneToGlobal streams.

`sidecar_injection_total` is wired through `injector.New()` with a nil-safe
metrics parameter to avoid forcing test refactors.

> Changelog: feat(kuma-cp): improve control-plane observability with dashboards, histograms, and operational metrics

### DIFF
--- a/pkg/plugins/runtime/k8s/webhooks/injector/injector.go
+++ b/pkg/plugins/runtime/k8s/webhooks/injector/injector.go
@@ -163,18 +163,30 @@ func (i *KumaInjector) InjectKuma(ctx context.Context, pod *kube_core.Pod) error
 
 	meshName, err := i.preCheck(ctx, pod, logger)
 	if err != nil {
-		if i.metrics != nil {
-			i.metrics.InjectionTotal.WithLabelValues("failed", classifyInjectionError(err)).Inc()
-		}
+		i.recordInjection("failed", classifyInjectionError(err))
 		return err
 	}
 	if meshName == "" {
-		if i.metrics != nil {
-			i.metrics.InjectionTotal.WithLabelValues("skipped", "not_needed").Inc()
-		}
+		i.recordInjection("skipped", "not_needed")
 		return nil
 	}
 
+	if err := i.injectKuma(ctx, pod, meshName, logger); err != nil {
+		i.recordInjection("failed", classifyInjectionError(err))
+		return err
+	}
+	i.recordInjection("success", "")
+	return nil
+}
+
+func (i *KumaInjector) recordInjection(result, reason string) {
+	if i.metrics == nil {
+		return
+	}
+	i.metrics.InjectionTotal.WithLabelValues(result, reason).Inc()
+}
+
+func (i *KumaInjector) injectKuma(ctx context.Context, pod *kube_core.Pod, meshName string, logger logr.Logger) error {
 	logger.Info("injecting Kuma")
 
 	// annotations
@@ -384,9 +396,6 @@ func (i *KumaInjector) InjectKuma(ctx context.Context, pod *kube_core.Pod) error
 		}
 	}
 
-	if i.metrics != nil {
-		i.metrics.InjectionTotal.WithLabelValues("success", "").Inc()
-	}
 	return nil
 }
 
--- a/pkg/plugins/runtime/k8s/webhooks/injector/injector.go
+++ b/pkg/plugins/runtime/k8s/webhooks/injector/injector.go
@@ -22,6 +22,7 @@ import (
 	core_config "github.com/kumahq/kuma/v2/pkg/config"
 	runtime_k8s "github.com/kumahq/kuma/v2/pkg/config/plugins/runtime/k8s"
 	"github.com/kumahq/kuma/v2/pkg/core"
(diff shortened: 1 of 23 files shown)


pkg/plugins/runtime/k8s/webhooks/injector/injector.go
```diff
@@ -144,15 +155,25 @@ type KumaInjector struct {
...
 }
 
 func (i *KumaInjector) InjectKuma(ctx context.Context, pod *kube_core.Pod) error {
 	logger := log.WithValues("pod", pod.GenerateName, "namespace", pod.Namespace)
 
 	meshName, err := i.preCheck(ctx, pod, logger)
-	if meshName == "" || err != nil {
+	if err != nil {
+		if i.metrics != nil {
+			i.metrics.InjectionTotal.WithLabelValues("failed", classifyInjectionError(err)).Inc()
```
bartsmykla: The failure counter only covers `preCheck`. If injection passes this point and later fails on a missing `ContainerPatch`, invalid transparent-proxy annotation, or probe rewrite error, those webhook rejections never increment `sidecar_injection_total{result="failed"}`. Record the final outcome for the full `InjectKuma` call, for example with a deferred result, so the failed series does not undercount rejections.

pkg/xds/sync/dataplane_watchdog.go
```diff
@@ -188,6 +191,11 @@ func (d *DataplaneWatchdog) syncDataplane(ctx context.Context) (SyncResult, erro
 			return SyncResult{}, errors.Wrap(err, "could not get identity")
 		}
 		d.workloadIdentity = identity
+		if d.XdsMetrics != nil && identity != nil && identity.ExpirationTime != nil {
+			d.XdsMetrics.CertExpirationRemaining.
+				WithLabelValues(d.key.Mesh).
+				Set(time.Until(*identity.ExpirationTime).Seconds())
```
bartsmykla: This stores a one-time `time.Until` sample only when `syncIdentity` runs. Later watchdog ticks with the same config and identity can return before this block, and when the cert finally becomes `ExpiringSoon` the identity is renewed before the gauge is set again. The value can stay near the full lifetime instead of counting down, so dashboards and alerts never see remaining lifetime fall. Update it from the current `d.workloadIdentity.ExpirationTime` whenever the identity is available, or export an expiry timestamp instead.

bartsmykla (changes requested):
Two metric paths need changes before this lands: sidecar injection failures after `preCheck`, and the cert-expiration gauge countdown.

pkg/plugins/runtime/k8s/webhooks/injector/injector.go
Automaat: Good catch — addressed in 4078b9cb9a. Extracted the main injection body into a private `injectKuma` helper and added a `recordInjection` method that runs on every return path in the public `InjectKuma` wrapper (preCheck error, skipped, body error, success). Every rejection now increments the counter with the correct `result`/`reason` labels.

pkg/xds/sync/dataplane_watchdog.go
Automaat: Addressed in 4078b9cb9a by switching to a stable **expiration timestamp** metric instead of remaining seconds. Renamed to `xds_cert_expiration_timestamp_seconds` and set it to `identity.ExpirationTime.Unix()`. Since the value is absolute (not a countdown), setting it only when the identity is refreshed is correct — dashboards/alerts compute remaining lifetime in PromQL with `xds_cert_expiration_timestamp_seconds - time()`, which decreases continuously without needing to update the gauge every tick.

bartsmykla (approved):

PR #16207, opened 2026-04-09
feat(k8s): remove CPU limit from init/sidecar container

lukidzi:
## Motivation

CPU limits cause throttling even when CPU is available, which increases latency under load. Removing the limit allows the containers to burst during startup and high-traffic periods.

## Implementation information

The default CPU limit for injected `kuma-init` and `kuma-sidecar` containers has been removed (set to `0`, meaning no limit). Previously the defaults were `100m` and `1000m` respectively. New configuration options were introduced and user is able to set them manually.


pkg/plugins/runtime/k8s/webhooks/injector/testdata/inject.01.golden.yaml
```diff
@@ -142,11 +142,10 @@ spec:
     name: kuma-init
     resources:
       limits:
-        cpu: 100m
-        memory: 50M
+        memory: 50Mi
       requests:
         cpu: 20m
-        memory: 20M
+        memory: 20Mi
```
lobkovilya: 1Mi is slightly larger than 1M. It's not critical but I wouldn't change it in this PR as the main goal is to remove the CPU limit

Automaat (approved):

PR #16213, opened 2026-04-10
fix(xds): skip SNI when BackendRef port not found

### LINKED ISSUE #13842: Investigate SNI behaviour with empty port from BackendRef
### Description

In Kuma there is a function that builds SNI from real resource: 
https://github.com/kumahq/kuma/blob/4eddb73a6f7a4a5b62cdc8114226a66d51eea963/pkg/plugins/policies/core/xds/meshroute/clusters.go#L146-L163

If we don't find port from sectionName port will be 0, and we will use this 0 to build SNI:
https://github.com/kumahq/kuma/blob/8dc904d4a84bc8304784866501721e5153757b09/pkg/xds/envoy/tls/sni.go#L79

We should investigate if this is ok to have 0 port in SNI, and if not we should handle this case properly

Automaat:
## Motivation

When a BackendRef doesn't select any actual port (port not found in destination), `SniForBackendRef` was still building an SNI using port `0`, producing an incorrect SNI value. This could also panic if the destination service itself is nil.

Fixes https://github.com/kumahq/kuma/issues/13842

## Implementation information

Callers of `SniForBackendRef` now resolve the destination and port up-front via `DestinationPortFromRef`, which returns `(dest, port, ok)`. When `ok == false` (destination missing or `SectionName` does not match a port), callers skip cluster generation (`continue` in loops, early `return nil, "", nil` in the gateway generator) instead of falling back to port `0`.

`SniForBackendRef` now takes the already-resolved port as input, so the SNI is always built from a real port value.

> Changelog: fix(xds): skip SNI when BackendRef port not found

### DIFF
--- a/pkg/plugins/policies/core/xds/meshroute/clusters.go
+++ b/pkg/plugins/policies/core/xds/meshroute/clusters.go
@@ -54,7 +54,10 @@ func GenerateClusters(
 			if meshCtx.IsExternalService(serviceName) {
 				switch {
 				case isMeshExternalService(meshCtx.EndpointMap[serviceName]):
-					sni := SniForBackendRef(service.BackendRef().RealResourceBackendRef(), meshCtx, systemNamespace)
+					sni, ok := SniForBackendRef(service.BackendRef().RealResourceBackendRef(), meshCtx, systemNamespace)
+					if !ok {
+						continue
+					}
 					if proxy.WorkloadIdentity != nil {
 						// MeshExternalService is only routable through zone egress in workload-identity mode.
 						// Skip cluster generation if no zone egress SANs are available.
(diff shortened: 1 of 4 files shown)


pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -54,7 +54,10 @@ func GenerateClusters(
 			if meshCtx.IsExternalService(serviceName) {
 				switch {
 				case isMeshExternalService(meshCtx.EndpointMap[serviceName]):
-					sni := SniForBackendRef(service.BackendRef().RealResourceBackendRef(), meshCtx, systemNamespace)
+					sni, ok := SniForBackendRef(service.BackendRef().RealResourceBackendRef(), meshCtx, systemNamespace)
```
lobkovilya: `service.BackendRef()` is already a resolved backend ref, meaning at this point we've managed to find the MeshService. 

The correct fix is probably implement `SniForBackendRef` that way so it doesn't require calling `FindPortByName` and already accepts the resolved port.

bartsmykla (approved):

PR #16215, opened 2026-04-10
fix(plugins): dont panic on removed policy

### LINKED ISSUE #15572: Removing policy from the list of enabled policies might cause panic
### Description

Kuma has a list of enabled policies, removing the policy from the list supposed to disable all the API endpoints related to the policy https://github.com/kumahq/kuma/blob/6a959ad0b5a6955e1e3bab2309519c9bf3f81305/deployments/charts/kuma/values.yaml#L957-L982

However, it might cause panic, because `ordered.Policies` list contains hardcoded policy types https://github.com/kumahq/kuma/blob/6a959ad0b5a6955e1e3bab2309519c9bf3f81305/pkg/api-server/resource_endpoints.go#L1298

Attempting to call `PolicyPlugins` with `ordered.Policies` is going to result in panic https://github.com/kumahq/kuma/blob/6a959ad0b5a6955e1e3bab2309519c9bf3f81305/pkg/core/plugins/registry.go#L137

lukidzi:
## Motivation

PolicyPlugins(`ordered.Policies`) panics when a policy is in `ordered.Policies` but not registered (e.g. disabled via config). Two separate lists had to stay in sync manually:

1. pkg/plugins/policies/core/ordered/ordered.go - hardcoded execution order
2. Runtime plugin registration - which policies are actually enabled

## Implementation information

Each policy now declares its own execution order via a `// +kuma:policy:order=N` marker, stored `ResourceTypeDescriptor.Order` and exposed via `PolicyPlugin.Order() int`. The `PolicyPlugins()` method takes no arguments and returns only registered (enabled) plugins sorted by their order. A disabled policy simply isn't registered, so it never appears.

- `ResourceTypeDescriptor`: new `Order int` field
- `PolicyPlugin` interface: new `Order() int` method (for parent project)

## Supporting documentation

Fix #15572


lobkovilya (changes requested):
I don't think we need to introduce `Order` into the `ResourceTypeDescriptor`. 

You just need a new method that's going to return `registeredPolicies` (i.e. `RegisteredPolicies()`) and update API to call
```go
allPlugins := core_plugins.Plugins().RegisteredPolicies()
```
instead of
```go
allPlugins := core_plugins.Plugins().PolicyPlugins(ordered.Policies)
```

lukidzi:
> I don't think we need to introduce `Order` into the `ResourceTypeDescriptor`.
> 
> You just need a new method that's going to return `registeredPolicies` (i.e. `RegisteredPolicies()`) and update API to call
> 
> ```go
> allPlugins := core_plugins.Plugins().RegisteredPolicies()
> ```
> 
> instead of
> 
> ```go
> allPlugins := core_plugins.Plugins().PolicyPlugins(ordered.Policies)
> ```

How would you ensure the order in this case? We need specific order of policies + ability to modify this order in KongMesh?

lobkovilya:
> How would you ensure the order in this case? We need specific order of policies + ability to modify this order in KongMesh?

In the Inspect API, do you need this specific order or just any order to produce stable responses?

If you need the specific order, then `RegisteredPolicies` can iterate over `ordered.Policies` and add to return value only policies that are present in `registeredPolicies`.

lukidzi:
> > How would you ensure the order in this case? We need specific order of policies + ability to modify this order in KongMesh?
> 
> In the Inspect API, do you need this specific order or just any order to produce stable responses?
> 
> If you need the specific order, then `RegisteredPolicies` can iterate over `ordered.Policies` and add to return value only policies that are present in `registeredPolicies`.

I don't like idea of maintaining 2 separate lists of policies. I would prefer to have only one list

pkg/core/resources/model/resource.go
```diff
@@ -210,6 +210,9 @@ type ResourceTypeDescriptor struct {
 	// IsFromAsRules if true, the entries in the spec.from field should be interpreted as rules.
 	// It's true for policies that allow only kind 'Mesh' in the spec.from.targetRef.
 	IsFromAsRules bool
+	// Order defines the execution order of this policy plugin relative to others.
+	// Lower values run first. Used by PolicyPlugins() to return a sorted list.
+	Order int
```
slonka, line 215: Is `Order` on the descriptor read anywhere at runtime? The registry sorts by `entry.Plugin.Order()` (the method on the plugin instance), not by `Descriptor().Order`.

pkg/core/plugins/registry.go
```diff
@@ -216,10 +205,16 @@ func (r *registry) Register(name PluginName, plugin Plugin) error {
...
 	if policy, ok := plugin.(PolicyPlugin); ok {
-		if old, exists := r.registeredPolicies[name]; exists {
-			return pluginAlreadyRegisteredError(policyPlugin, name, old, policy)
+		for _, existing := range r.registeredPolicies {
+			if existing.Name == name {
+				return pluginAlreadyRegisteredError(policyPlugin, name, existing.Plugin, policy)
+			}
 		}
-		r.registeredPolicies[name] = policy
+		entry := RegisteredPolicyPlugin{Plugin: policy, Name: name}
```
slonka, line 213: When two plugins declare the same order, insertion order depends on `Register` call order - right? Should we at least log something or maybe even fail?

docs/guides/new-policies.md
```diff
@@ -3,7 +3,7 @@
 Use the tool:
 
 ```shell
-go run ./tools/policy-gen/bootstrap/... --name CaseNameOfPolicy --is-policy true
+go run ./tools/policy-gen/bootstrap/... --name CaseNameOfPolicy --is-policy true --order 899
```
slonka: Should the PR lay out an explicit rule for picking order number? Maybe:

> **Picking an `order` value**
>
> - Existing policies use multiples of 100 (`100`, `200`, ..., `1500`) with a gap of 100 between each, so a new policy can be slotted between two neighbors without renumbering anything.
> - `9999` is reserved for policies that must run last (currently only `MeshProxyPatch`).
> - When adding a new policy, pick a value halfway between the two neighbors where it logically fits (e.g. between `MeshTrace` at `700` and `MeshFaultInjection` at `800`, use `750`).
> - If the policy must run first, pick below `100`.
>
> Lower `order` runs first. The value is declared on the API struct via `// +kuma:policy:order=N` and propagates to `ResourceTypeDescriptor.Order` at generation time.

With that rule in place, `--order 899` in the example above would be a poor choice and should probably be `--order 750`


```suggestion
go run ./tools/policy-gen/bootstrap/... --name CaseNameOfPolicy --is-policy true --order 750
```

docs/guides/new-policies.md
```diff
@@ -12,7 +12,7 @@ The output of the tool will tell you where the important files are!
 
 To enable policy you need to adjust configuration of two places:
 * Remove `+kuma:policy:skip_registration=true` from your policy schema.
-* `pkg/plugins/policies/core/ordered/ordered.go`. Plugins name is equals to `KumactlArg` in file `zz_generated.resource.go`. It's important to place the plugin in the correct place because the order of executions is important.
+* Also, define the order of the policy as it is important in configuration generation
```
slonka: Should we mention here `+kuma:policy:order=N`?

pkg/plugins/policies/meshaccesslog/plugin/v1alpha1/plugin.go
```diff
@@ -52,6 +52,8 @@ var _ core_plugins.PolicyPlugin = &plugin{}
 
 type plugin struct{}
 
+func (p plugin) Order() int { return 600 }
```
slonka: should this return `DescriptorForThisType.Order`?

slonka (commented):
2 main things:
1. Picking `order` value could have defined rules
2. `Order` field possibly not used.

pkg/core/resources/model/resource.go
lukidzi: Now it's used

pkg/core/plugins/registry.go
```diff
@@ -129,19 +129,8 @@ func (r *registry) ProxyPlugins() map[PluginName]ProxyPlugin {
...
-		if !ok {
-			panic(fmt.Sprintf("Couldn't find plugin %s", policy))
-		}
-		plugins = append(plugins, RegisteredPolicyPlugin{
-			Plugin: plugin,
-			Name:   policy,
-		})
-	}
-	return plugins
+func (r *registry) PolicyPlugins() []RegisteredPolicyPlugin {
```
lobkovilya, line 133: please add a comment this method returns policies sorted by the order

pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/plugin.go
```diff
@@ -30,7 +30,7 @@ var (
 
 type plugin struct{}
 
-func (p plugin) Order() int { return 1100 }
+func (p plugin) Order() int { return api.MeshTrafficPermissionResourceTypeDescriptor.Order }
```
slonka, line 33: is this generated by the plugin generation logic? or manually entered? I don't see changes in the generation logic so I assume it's manual. We should update that so the next person that runs ` ./tools/policy-gen/bootstrap/... --order 750` doesn't have to put that in. WDYT?

docs/guides/new-policies.md
slonka: In the comment above I suggested writing down the rules for picking order value, do you think it's not very useful?

pkg/core/resources/model/resource.go
```diff
@@ -210,6 +210,9 @@ type ResourceTypeDescriptor struct {
 	// IsFromAsRules if true, the entries in the spec.from field should be interpreted as rules.
 	// It's true for policies that allow only kind 'Mesh' in the spec.from.targetRef.
 	IsFromAsRules bool
+	// Order defines the execution order of this policy plugin relative to others.
+	// Lower values run first. Used by PolicyPlugins() to return a sorted list.
```
lobkovilya:
```suggestion
	// Order defines the execution order of the associated plugin relative to others. It's used only when IsPluginOriginated is true.
	// Lower values run first. Used by PolicyPlugins() to return a sorted list.
```

pkg/core/plugins/registry.go
```diff
@@ -216,10 +205,19 @@ func (r *registry) Register(name PluginName, plugin Plugin) error {
...
+		}
+		entry := RegisteredPolicyPlugin{Plugin: policy, Name: name}
+		pos, found := slices.BinarySearchFunc(r.registeredPolicies, entry, func(a, b RegisteredPolicyPlugin) int {
+			return cmp.Compare(a.Plugin.Order(), b.Plugin.Order())
+		})
+		if found {
+			return errors.Errorf("policy plugin %q has the same order (%d) as plugin %q; each plugin must declare a unique order", name, policy.Order(), r.registeredPolicies[pos].Name)
 		}
-		r.registeredPolicies[name] = policy
+		r.registeredPolicies = slices.Insert(r.registeredPolicies, pos, entry)
```
lobkovilya, line 221: Can we log something on a startup? i.e. these plugins in this order are loaded. I think it'll be helpful for debugging. 

Or even better, can we make `/_resources` endpoint to respect the order?

lobkovilya (commented):
~~It looks like we have some logic duplication. There is a `/_resources` endpoint and I was thinking it would be nice if this endpoint would return the list of resources respecting the `order` field. However, API server is using `ObjectDescriptors()` method from the type registry https://github.com/kumahq/kuma/blob/67ccebd412c34d0ba44310b833c32cfd5a3ff828/pkg/core/resources/registry/registry.go#L30~~

UPD: I think it's fine, one registry is for types and another registry for plugins.

pkg/api-server/mappers/resource_type_description.go
```diff
@@ -9,8 +9,25 @@ import (
 )
 
 func MapResourceTypeDescription(defs []model.ResourceTypeDescriptor, readOnly bool, federatedZone bool) api_types.ResourceTypeDescriptionList {
+	sorted := make([]model.ResourceTypeDescriptor, len(defs))
```
lobkovilya: I'm not sure it makes sense, sorry 😅 It feels like a hidden contract that `/_resources` returns items in the order that associated plugins are going to run.

Could you please just log on CP init that plugins are loaded and will be applied in the following order?

lobkovilya (approved):

PR #16219, opened 2026-04-10
test(e2e): fix multiple flaky e2e tests

Automaat:
## Motivation

Aggregated e2e flakes from ~2 weeks of master CI + the `lukidzi-patch-1` (PR #16001) verify-stability loop. `ci-stability-master` stays 100% green on same-SHA reruns → all of these are non-determinism, not regressions. Each fix below was reproduced locally (where possible) under throttled k3d, root-caused, and verified.

### 1. Gateway connection-limit race (`resources.go:218`)

`Timed out after 60s … Expected <int>: 0 To satisfy [&{52} &{56}]`. After `KillAppPod` + new `keepConnectionOpen` goroutine the test immediately ran the rejection Eventually, racing pod respawn + sidecar init + `kubectl exec ncat` + TCP handshake. Fix: poll gateway Envoy admin (`http.<gw>_<ns>_svc.downstream_cx_active >= 1`) before the rejection check — ground truth instead of timing.

Repro at 1.25 CPU / 3 GB throttle: 3/12 fail before, 12/12 pass after.

### 2. MLBS WaitNumPods (`testserver/kubernetes.go:336`)

`curl: (22) … 404` for full 30 s window on `sidecarContainers`. `WaitNumPods(..., 1, ...)` was hardcoded — `testserver.Install(WithReplicas(3))` returned when only pod 0 existed; gateway's MeshService had incomplete endpoints. One-line fix: `int(k.opts.Replicas)`. Also benefits any test using `WithReplicas(N>1)`.

### 3. Helm upgrade multizone ingress race + 5m budget (`kuma_helm_upgrade_multizone.go`)

Two variants: (a) `Consistently` race where `global` reached 2 ingresses but zone views still showed 3 — extended to wait for **all three views** to agree; (b) outer `Eventually` timing out at 120 s because the rolling upgrade (`maxSurge=1, maxUnavailable=0`) holds 3 ingresses for longer than the nominal budget under CI pressure — bumped to 5 min and added the three-count to the failure annotation.

### 4. Multizone WaitForZonesOnline (`framework/envs/multizone/env.go`)

Root cause of `MeshTrafficPermission [It]` 503→403 + the `InboundPassthroughDisabled [BeforeAll]` family: `SynchronizedBeforeSuite` returned on local readiness probes, but readiness ≠ KDS zone-to-global stream established. Tests raced KDS — Global hadn't synced cross-zone DPs, demo-client's Envoy had no upstream endpoint, first request returned 503 not 403. Fix: new `Eventually` after `wg.Wait()` polling `kumactl inspect zones` on Global until all four zone names appear `Online`. 2 min budget, 2 s poll.

8/8 PASS at 2 CPU / 4 GB throttle (vs 3/5 fail without fix).

### 5. envoyconfig waitConfigStable (`envoyconfig/utils.go`)

`Sidecars` table is `Ordered`, alphabetic — `meshmetric/otel.input.yaml` runs immediately before `meshmetric/prometheus.input.yaml`. Both apply `MeshMetric mm-1`; otel adds an OpenTelemetry cluster. `cleanupAfterTest` deleted `mm-1` but didn't wait for envoy to drop the cluster, so the next prometheus spec saw the leftover otel cluster and the golden compare timed out. Fix: new `waitConfigStable(mesh, dpp)` polls `getConfig` for two consecutive identical snapshots (1 s apart, 30 s budget) per dataplane; called from `cleanupAfterTest`. Parameterized `dpps` per caller (the BuiltinGateway suite uses `gateway-proxy`, not `demo-client`/`test-server`).

### 6. kuma-init ContainerPatch retry on webhook race (`framework/k8s_cluster.go`)

Recurring CI signature across **6+ tests** in 3 distinct files (taint controller, helm Global+Zone, cert-manager, federation, multizone resilience, externalservices hybrid):

```
failed to apply kuma-init-no-cpu-limit ContainerPatch:
failed calling webhook "validator.kuma-admission.kuma.io":
no endpoints available for service "kuma-control-plane"
```

`WaitApp` returns once the kuma-cp pod's readiness probe passes, but the `Endpoints` object is populated by a separate controller and lags by a few seconds — the documented k8s pod-Ready vs Service-Endpoints sync race. Under e2e_4 split (4 parallel ginkgo procs hammering the same API server) this gap is reliably hit.

Can't reproduce locally (single ginkgo proc → idle API server → endpoints land instantly). Fix is the canonical community workaround: wrap the failing `kubectl apply ContainerPatch` in `retry.DoWithRetryE` (60 s budget). Principled, one-line behavior change, clears the entire family.

## Supporting documentation

Investigation notes and per-flake reproduction details in `flake-investigation.md` (worktree-only, not committed).

> Changelog: skip


slonka:
@Automaat / @lukidzi can you confirm which items are still needed?

- Fix 2 (MLBS `WaitNumPods` / test deployments): seems to be addressed in https://github.com/kumahq/kuma/pull/16269?
- Fix 3 (helm upgrade multizone ingress race): seems to be addressed in https://github.com/kumahq/kuma/pull/16281?
- Fix 4 (`WaitForZonesOnline`): seems to be addressed in https://github.com/kumahq/kuma/pull/16269 
- Fix 6 (kuma-init `ContainerPatch` webhook race): seems to be addressed in https://github.com/kumahq/kuma/pull/16207 and https://github.com/kumahq/kuma/pull/16269?

Do we only need fix 1, 5 from this?

PR #16226, opened 2026-04-11
fix(kuma-dp): bridge self metrics to otel path

Automaat:
## Motivation

PR #16201 added new Kuma-DP Prometheus metrics (DNS proxy, config fetcher, etc.) registered on `prometheus.DefaultRegisterer`. Under MeshMetric with an OpenTelemetry backend these metrics never reach the pipeline: the OTel path uses `AggregatedProducer` which only HTTP-scrapes Envoy + user apps, so anything registered on the kuma-dp process itself is silently dropped.

The Prometheus backend path is unaffected — the hijacker HTTP handler already serves `DefaultGatherer` on `/meshmetric`.

Symptom: `kuma_dp_dns_*` series are missing from Prometheus when the mesh is configured with a MeshOpenTelemetryBackend. This also silently affects the existing `kuma_dp_envoyconfigfetcher_*` metrics and the `kuma_dp_dns_request_duration_seconds_*` histogram referenced by the workload-debug dashboard.

## Implementation information

Two commits:

1. **`fix(kuma-dp): bridge self metrics to otel path`** — add a second `sdkmetric.Producer` to the `PeriodicReader` in `startExporter` that wraps `prometheus.DefaultGatherer` via `go.opentelemetry.io/contrib/bridges/prometheus`. The bridge package is already a direct dependency — `pkg/plugins/runtime/opentelemetry/metrics.go` uses the same pattern for the control-plane OTel pusher. Both pipe and direct (Envoy-socket) OTel backends flow through `startExporter`, so one change covers both paths. No new deps, no config surface change.

2. **`fix(dns): fix race in multi-listener shutdown`** — drive-by fix for a race exposed by the existing `returns the first listener error when another listener started` test. When one DNS listener fails immediately (e.g. bind `address already in use`) and another is racing to call `NotifyStartedFunc`, the shutdown loop may read `started[i]=false` and skip `Shutdown()`. The drain-errCh loop then blocks forever waiting for an exit signal that never comes. Fix: retry `Shutdown()` in a goroutine until either the server transitions to started or its per-listener `done` channel fires. Hit this as a flake on the original CI run for this PR; ginkgo `-repeat=5` locally is now stable.

Alternatives considered for (1):
- Adding a synthetic `ApplicationToScrape` pointing at the hijacker's own `/meshmetric` path — creates a self-scrape loop and duplicates parse/serialize work.
- Migrating `kuma_dp_dns_*` to native OTel instruments — larger blast radius, breaks the Prometheus backend path which currently works.

## Supporting documentation

Follow-up to #16201.

> Changelog: fix(kuma-dp): ship kuma-dp self metrics to OpenTelemetry backends

### DIFF
--- a/app/kuma-dp/pkg/dataplane/meshmetrics/component.go
+++ b/app/kuma-dp/pkg/dataplane/meshmetrics/component.go
@@ -9,6 +9,8 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	prombridge "go.opentelemetry.io/contrib/bridges/prometheus"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
@@ -255,8 +257,14 @@ func startExporter(ctx context.Context, target OtelExportTarget, producer *metri
 	if err != nil {
 		return nil, err
 	}
+	// Surface kuma-dp's own Prometheus metrics (DNS proxy, config fetcher, etc.)
+	// through the OTel pipeline. In the Prometheus backend path the hijacker
+	// HTTP handler already serves these via DefaultGatherer; the OTel path
+	// bypasses that handler, so without this bridge the metrics are lost.
+	selfMetricsBridge := prombridge.NewMetricProducer(prombridge.WithGatherer(prometheus.DefaultGatherer))
 	readerOpts := []sdkmetric.PeriodicReaderOption{
 		sdkmetric.WithInterval(target.RefreshInterval),
+		sdkmetric.WithProducer(selfMetricsBridge),
 	}
 	if producer != nil {
 		readerOpts = append(readerOpts, sdkmetric.WithProducer(producer))
--- a/app/kuma-dp/pkg/dataplane/dnsproxy/component.go
+++ b/app/kuma-dp/pkg/dataplane/dnsproxy/component.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"runtime"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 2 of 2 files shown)


app/kuma-dp/pkg/dataplane/meshmetrics/component.go
```diff
@@ -255,8 +257,14 @@ func startExporter(ctx context.Context, target OtelExportTarget, producer *metri
 	if err != nil {
 		return nil, err
 	}
+	// Surface kuma-dp's own Prometheus metrics (DNS proxy, config fetcher, etc.)
+	// through the OTel pipeline. In the Prometheus backend path the hijacker
+	// HTTP handler already serves these via DefaultGatherer; the OTel path
+	// bypasses that handler, so without this bridge the metrics are lost.
+	selfMetricsBridge := prombridge.NewMetricProducer(prombridge.WithGatherer(prometheus.DefaultGatherer))
```
utafrali, line 264: `prometheus.DefaultGatherer` includes the standard Go process and runtime collectors (goroutines, GC pause, process CPU/mem, etc.) that the Prometheus client registers by default. These will now be pushed to every configured OTel backend. That is probably desirable and consistent with what the Prometheus backend already exposes on `/meshmetric`, but it is worth a quick note in the PR description or commit message so reviewers know the scope — it is not just `kuma_dp_dns_*` and `kuma_dp_envoyconfigfetcher_*` that start flowing.

app/kuma-dp/pkg/dataplane/meshmetrics/component.go
```diff
@@ -255,8 +257,14 @@ func startExporter(ctx context.Context, target OtelExportTarget, producer *metri
...
 	readerOpts := []sdkmetric.PeriodicReaderOption{
 		sdkmetric.WithInterval(target.RefreshInterval),
+		sdkmetric.WithProducer(selfMetricsBridge),
```
utafrali, line 267: `startExporter` has no unit test that exercises the bridge wiring. The existing tests in `component_test.go` construct `runningBackend` directly with a pre-shutdown provider, so `startExporter` is never called. Adding even a lightweight check — e.g. registering a counter on a fresh `prometheus.Registry`, passing it via `prombridge.WithGatherer`, and asserting the bridge produces a non-empty metric slice — would catch a regression if the `WithProducer` call is accidentally dropped in a future refactor. That said, wiring up a full gRPC socket in a unit test is non-trivial, so this is a suggestion rather than a blocker.

utafrali (approved):
Correct, minimal fix that follows the established pattern from the CP OTel pusher and uses an already-direct dependency. The two comments are low-severity: one is a scope clarification and the other is a test coverage suggestion for a code path that was already untested before this PR.

bartsmykla (approved):

PR #16260, opened 2026-04-13
feat(meshpassthrough): introduce matching API

lukidzi:
## Motivation

`MeshPassthrough` currently generates filter chain selection using per-filter-chain `FilterChainMatch` fields. This approach has known limitations: Envoy evaluates `FilterChainMatch` fields using a fixed precedence order that does not always align with how `MeshPassthrough` rules need to be prioritized, and it becomes unwieldy when combining protocol, SNI, port, and IP/CIDR matching across many rules. Envoy's Matcher API (`Listener.FilterChainMatcher`, using `xds.type.matcher.v3.Matcher`) provides a declarative, tree-structured alternative that gives explicit control over matching order and combines multiple conditions cleanly.

## Implementation information

This PR migrates `MeshPassthrough` listener generation to use the Envoy Matcher API behind a feature flag (`feature-meshpassthrough-matcher-api`). The flag can be set per-dataplane or enabled globally via the `MeshPassthroughMatcherAPI` experimental config field.

`filter_chain_matcher.go` builds this tree from the existing `FilterChainMatch` slice. The plugin checks for the feature flag and passes `UseMatcherAPI` to the configurer, which then either attaches the matcher to `Listener.FilterChainMatcher` or falls back to the existing per-chain `FilterChainMatch` path.
New builder helpers added to `pkg/envoy/builders/xds/matchers`:
- `MatcherTreeExactMap` / `MatcherTreePrefixMap` — tree node constructors
- `FilterChainNameOnMatch` — action wrapping a `StringValue` (as required by Envoy's filter chain action type)
- `NetworkInputTransportProtocol`, `NetworkInputApplicationProtocol`, `NetworkInputServerName`, `NetworkInputDestinationPort`, `NetworkInputDestinationIP` — typed extension config helpers
- `IPRangeOnMatch` — builds a `MatcherList` node for CIDR matching using `envoy.matching.matchers.ip`


slonka:
This can be reopened for 3.1.x

PR #16209, opened 2026-04-09
docs(MADR): context and user stories for mesh-scoped proxy policies

lukidzi:
## Motivation

We want to support targeting mesh-scoped zone proxies by policies.

## Implementation information

Define context and user stories

> Changelog: skip

PR #16275, opened 2026-04-14
docs(MADR): added policy matching

lukidzi:
## Motivation

We want to enable mesh-scoped proxies to be targeted by policies

## Implementation information

Added document how can we achieve this

## Supporting documentation

Fix https://github.com/Kong/kong-mesh/issues/9029


docs/madr/decisions/101-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -2,114 +2,83 @@
...
-  service limits and exhaust the budget.
-* As a mesh operator I want to inject HTTP headers with a token on the egress for all outgoing requests
-  to an external service so that all clients in the mesh can use the same token without granting access
-  to the token to individual clients.
-* As a mesh operator I want to give access to service owners to a single HTTP endpoint of an external
-  resource so that the system follows the least privilege principle.
+Zone Ingress and Zone Egress were originally dedicated global-scoped resource types.
+Being global-scoped meant they lived outside any mesh context, which created two fundamental blockers:
+
+1. **No mesh identity** — `ZoneEgress` could not participate in mTLS with a
```
slonka, line 12: I'm confused, should this PR be pointed to this branch https://github.com/kumahq/kuma/pull/16209/changes ?

lobkovilya:
@lukidzi please merge master to see a clean diff

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -2,114 +2,83 @@
...
+* As a mesh operator I want to rate limit requests to an external service so that clients don't
+  go over service limits and exhaust the budget.
+* As a mesh operator I want to inject HTTP headers with a token on the egress for all outgoing
+  requests to an external service so that all clients in the mesh can use the same token without
+  granting access to the token to individual clients.
 
 ## Design
 
-### General Policy Model for Zone Proxy
+### Scope of Zone Egress
```
lobkovilya: Since the MADR is focusing on both zone ingress and egress it'd make sense to cover both here

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -122,106 +91,142 @@ type Match struct {
 }
 ```
 
-Usage:
+With the KRI-based SNI format ([MADR-101](101-sni-format-improvements.md)), SNIs follow the
+pattern `sni.<type>.<mesh>.<zone>.<namespace>.<name>[.<sectionName>]` and are human-readable,
```
lobkovilya: SectionName is not optional, it's always present. Without the port we can't forward the traffic.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -122,106 +91,142 @@ type Match struct {
...
 ```
 
-Usage in other inbound policies (`MeshRateLimit`):
+* Good, because users reference the resource by name or labels.
+* Good, because it is consistent with how other Kuma policies reference resources.
+* Good, because label-based matching provides broad scoping (by namespace, zone, team, etc.) without requiring users to know internal SNI formats.
+* Good, because users are already familiar with `targetRef` selectors from existing policies and do not need to learn a new matching mechanism.
+* Good, because policies are insensitive to SNI format changes, if the SNI format evolves, existing policies require no updates; the CP re-resolves them.
+* Bad, because CP must resolve `targetRef` to SNI(s) during policy compilation.
+* Bad, because `targetRef` in `Match` is only meaningful on zone proxy Dataplanes where per-destination filter chains exist. On regular sidecars the field is invalid and must be rejected at admission, meaning the shared `Match` struct carries a field that most consumers can never use.
```
lobkovilya: We shouldn't reject it, just ignore if no such filter chain. Same logic we have with targeting MeshHTTPRoute, you can have MeshTimeout:

```yaml
spec:
  to:
    - targetRef:
        kind: MeshHTTPRoute
        name: route-1
      default: {}
```

and the policy will be ignored on the DPPs that don't have `route-1`

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -122,106 +91,142 @@ type Match struct {
...
+          type: Exact
+          value: "spiffe://default/ns/backend-ns/sa/backend"
+        targetRef:
+          kind: MeshExternalService
+          labels:
+            k8s.kuma.io/namespace: backend-ns
 ```
 
-Usage in other inbound policies (`MeshRateLimit`):
+* Good, because users reference the resource by name or labels.
```
lobkovilya, line 177: Using `labels` is the problem, user can unintentionally select multiple MeshExternalService and open access to resources they didn't mean to

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -122,106 +91,142 @@ type Match struct {
...
+    SpiffeID  *SpiffeIDMatch        `json:"spiffeID,omitempty"`
+    TargetRef *common_api.TargetRef `json:"targetRef,omitempty"`
+}
+```
 
-The `targetRef` field in `Match` is optional. When absent, the match applies to all
-destinations (same as today for regular sidecars). When present, the match applies only
-to the filter chain for the referenced resource.
+- `targetRef` is the primary, user-facing mechanism. The CP resolves it to SNI(s) at
+  compile time. For zone egress, only `kind: MeshExternalService` is valid.
```
lobkovilya: what's "compile time"?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -489,140 +299,79 @@ spec:
...
 mesh: default
-name: egress-outbound-access-log
+name: aurora-combined
 spec:
   targetRef:
     kind: Dataplane
-    labels:
-      k8s.kuma.io/zone-proxy-type: egress
+    sectionName: ze-port
+  rules:
```
lobkovilya, line 296: it's not possible to combine `to` and `rules`

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -489,140 +299,79 @@ spec:
...
+    - matches:
+        - targetRef:
+            kind: MeshExternalService
+            name: aws-aurora
+      default:
+        connectionTimeout: 10s
   to:
```
lobkovilya: Should we keep `spec.to[]` only for sidecars? I think it will be a simpler mental model for users:
* `spec.to[].targetRef.kind: MeshExternalService` always modifies the sidecar's outbound
* `spec.rules[].matches[].targetRef.kind: MeshExternalService` always modifies zone proxy's filter chain

lobkovilya (commented):
I’m not convinced by Option B with targetRef. Option A feels cleaner:

* SNI is objective and observable. You don’t need Kuma to infer whether a policy applies. You can inspect the traffic directly (e.g., in Wireshark) and verify it.


* `targetRef.labels` can unintentionally match multiple MeshExternalServices, whereas `matches[].sni` is inherently singular.


* If the Control Plane still translates targetRef into sni, then targetRef is effectively just syntactic sugar. For a gradual API evolution, it would be better to first introduce `matches[].sni`, and only add `matches[].targetRef` later if users actually need a higher-level abstraction.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+   Because `ZoneIngress`/`ZoneEgress` were not `Dataplane` resources, none of the mesh policies:
+   observability (MeshAccessLog, MeshMetric, MeshTrace), security (MeshTrafficPermission), or
+   traffic management (MeshRateLimit, MeshFaultInjection) — could target them.
+
+[MADR-095](095-mesh-scoped-zone-ingress-egress.md) resolved this by modelling zone proxies as
+mesh-scoped `Dataplane` resources with a `listeners` array. With that foundation in place, this
+MADR establishes the **unified policy model** for zone proxies - how inbound and outbound policies
+are structured, how zone proxy Dataplanes are targeted, and how policies can select a specific
+MeshExternalService destination on zone egress. It resolves all policy placement items deferred
+by [MADR-062](062-meshexternalservice-and-zoneegress.md).
```
slonka: MADR-062 also deferred `MeshFaultInjection` - do we want to covert it here? Should we explicitly mention that it's out of scope? There is also mentions of MFI below.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+
+## User Stories
+
+* As a mesh operator I want to give access to service owners to a specific external resource
```
slonka: nitpick: these could be numbered so it’s easier to refer to them later on

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+  can audit access post-incident.
+* As a mesh operator I want to have observability available on zone proxies so that I can
+  troubleshoot issues and monitor performance.
+* As a mesh operator I want to be able to override Envoy connection and traffic parameters
+  (connectTimeout, maxConnections, etc.) so that I can always provide values suited for my traffic.
+* As a mesh operator I want to rate limit requests to an external service so that clients don't
+  go over service limits and exhaust the budget.
+* As a mesh operator I want to inject HTTP headers with a token on the egress for all outgoing
+  requests to an external service so that all clients in the mesh can use the same token without
+  granting access to the token to individual clients.
```
slonka: Should we highlight that this is a new functionality (compared to other user stories)?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+
+### Destination Selector in Inbound Rules
+
+On zone egress, every inbound connection from a sidecar carries a **destination**:
+the SNI presented in the mTLS handshake identifies the MeshExternalService the sidecar
+wants to reach. When zone egress builds per-MeshExternalService filter chains,
+this destination is available at the filter chain selection level.
+
+Inbound `rules` can select a specific MeshExternalService destination,
+enabling per-destination rate limits, access control, and fault injection on zone egress.
```
slonka: `fault injection` is mentioned here as a use case, but there's no corresponding user story for it

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+### Scope of Zone Egress
+
+Zone egress is **not** a general-purpose L7 gateway.
+It is a transit proxy for outbound MeshExternalService traffic with policy enforcement.
+
+Zone egress is responsible for:
+- Forwarding mTLS-terminated traffic to MeshExternalService endpoints
+- Enforcing inbound access control (MeshTrafficPermission) and rate limits per source/destination
+- Applying outbound policies (timeouts, circuit breakers, health checks) to external endpoints (see policy matrix for prioritization — some are deferred)
+Zone egress is NOT responsible for:
```
slonka, line 59: You need a newline otherwise it looks like this:

Image

```suggestion

Zone egress is NOT responsible for:
```

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+        sni:
+          value: "sni.meshexternalservice.default.zone-1.backend-ns.aws-aurora"
+```
+
+* Good, because it maps directly to Envoy's filter chain match on `server_names`.
+* Good, because it is transparent — what you write is what Envoy matches on.
+* Good, because the KRI-based SNI format (MADR-101) makes SNIs human-readable and
+  predictable — users can construct them from resource attributes without querying the CP.
+* Bad, because SNIs are an internal representation and coupling policy to them creates
+  a dependency on the SNI format stability.
```
slonka: Is it really internal if MADR-101 goes through?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+          type: Exact
+          value: "spiffe://default/ns/backend-ns/sa/backend"
+        targetRef:
+          kind: MeshExternalService
+          name: aws-aurora
+```
+
+No other service can reach `aws-aurora` through zone egress because no other `allow` entry matches.
+
+#### Access control: deny specific source from all external resources
```
slonka: Why does it say **all**? aren't we denying only `compromised-worker` from accessing `aws-aurora`?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Security implications and review
+
+### MeshTrafficPermission Default Behaviour
+
+Zone egress is a security boundary — it is the sole exit path for MeshExternalService traffic.
+The default behaviour when no MeshTrafficPermission targets zone egress must be **deny-all**.
+This MADR proposes changing `mesh.spec.routing.defaultForbidMeshExternalServiceAccess` from its
+current default of `false` (permissive, set in MADR-062 when egress targeting was not yet available)
+to `true` (deny-all). Now that MeshTrafficPermission can target zone egress with per-destination
```
slonka, line 334: This can be really surprising, we should document it in `upgrade.md` in **bold**

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+   struct can be extended in a follow-up without breaking existing policies.
+
+3. **Default RBAC behaviour**: `defaultForbidMeshExternalServiceAccess` defaults to `true`
+   (deny-all). This supersedes MADR-062's permissive default, which was set before egress
+   targeting was available.
+
+4. **Resolution-failure semantics**: Unresolvable `targetRef` in security policies fails closed
+   (match-none). In traffic policies, unresolvable matches are silently dropped.
+
+5. **SNI tombstone on delete**: Deleted MeshExternalService SNIs must not be reused until all
```
slonka: "SNI tombstone on delete" is mentioned here for the first time in the document - there's no section explaining what exactly it is

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+   zone proxies have converged on the new snapshot.
+
+## Notes
+
+* `targetRef` in `Match` is valid only when the policy targets a zone proxy Dataplane.
```
slonka: nitpick: this repeats line 225

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,421 @@
...
+### Audit Trail
+
+MeshAccessLog on zone egress inbound captures the source SPIFFE identity and the original
+destination SNI, providing an audit trail of which workload accessed which external resource.
+
+## Implications for Kong Mesh
+
+Kong Mesh ships its own rate-limiting and header-injection policies. When zone egress gains
+inbound `rules` support and HTTP outbound policies, Kong Mesh policy equivalents should be
+validated against the new filter chain architecture.
```
slonka: should we expand on this section?

slonka (commented):
Did a first pass today, will do another tomorrow.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lukidzi: As of today I think the `MeshTimeout` and `MeshAccessLog` policy would make sense to support so maybe would be good to use it only in this case.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lukidzi: Not required, as there is no impact on KongMesh

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lukidzi: SNI still is internal, and as I remember 101 didn't mention to store it anywhere

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lukidzi: SNI is more intentional but I feel like it's not obvious as well. SNI omits zone/namespace sections in case the resource is created on Global or universal (no namespace).

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+|---------------------------|---------------------------|-----------------------|---------------|-------|
+| MeshTrafficPermission     | Yes                       | No                    | **2.14**      | Required for MeshExternalService access control to be complete |
+| MeshAccessLog             | Yes                       | Yes                   | **2.14**      | |
+| MeshMetric                | Yes                       | Yes                   | **2.14**      | |
+| MeshTrace                 | Yes                       | No                    | **2.14**      | Ingress is TCP-only — tracing not meaningful |
+| MeshProxyPatch            | Yes                       | Yes                   | 3.0           | Independent of SNI/MES designs; customer demand in ask-mesh |
+| MeshTLS                   | Yes                       | Yes                   | 3.0           | |
+| MeshRateLimit             | Maybe                     | No                    | Maybe         | |
+| MeshTimeout               | Maybe                     | No                    | Maybe         | |
+| MeshFaultInjection        | Maybe                     | No                    | Maybe         | |
```
slonka: Five policies are `Maybe / Maybe` with no decision criteria. I think @justindavies / @lahabana  should be aware of these and be able to make the decision - not necessarily now but I think sometime after `2.14`.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+   SNI maps directly to Envoy's `server_names` filter chain match — no CP resolution step.
+   With the KRI-based SNI format (MADR-101), SNIs are human-readable and predictable.
+   `targetRef` in `Match` is deferred to a follow-up.
+
+3. **Default RBAC behaviour**: `defaultForbidMeshExternalServiceAccess` defaults to `true`
+   (deny-all). This supersedes MADR-062's permissive default, which was set before egress
+   targeting was available.
+
+4. **Non-matching SNI semantics**: If no `allow` entry matches the SNI, access is denied
+   (fail-closed). In traffic policies, non-matching `sni` entries are silently skipped.
```
slonka: You removed `SNI tombstone on delete` bullet - was this not needed? Or just because this collision (same name/port/mesh/type → same hash) is now possible so we don't care about improving it?

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+    allow:
+      - spiffeID:
+          type: Exact
+          value: "spiffe://default/ns/backend-ns/sa/backend"
+        sni:
+          value: "sni.meshexternalservice.default.zone-1.backend-ns.aws-aurora"
+```
+
+* Good, because it maps directly to Envoy's filter chain match on `server_names`.
+* Good, because it is transparent — what you write is what Envoy matches on.
```
slonka, line 268: I think this wasn't recorded from Ilya's review https://github.com/kumahq/kuma/pull/16275#pullrequestreview-4139056605

```suggestion
* Good, because it is transparent — what you write is what Envoy matches on.
* Good, because SNI is observable at the wire - policy application can be verified by inspecting traffic directly.
```

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+  targetRef:
+    kind: Dataplane
+    sectionName: ze-port
+  default:
+    allow:
+      - spiffeID:
+          type: Exact
+          value: "spiffe://default/ns/backend-ns/sa/backend"
+        sni:
+          value: "sni.meshexternalservice.default.zone-1.backend-ns.aws-aurora"
```
slonka: nitpick: are we missing sectionName here?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+          value: "spiffe://default/ns/backend-ns/sa/backend"
+        targetRef:
+          kind: MeshExternalService
+          labels:
+            k8s.kuma.io/namespace: backend-ns
+```
+
+* Good, because users reference the resource by name or labels.
+* Good, because it is consistent with how other Kuma policies reference resources.
+* Good, because label-based matching provides broad scoping (by namespace, zone, team, etc.) without requiring users to know internal SNI formats.
```
slonka:
```suggestion
* Good, because label-based matching provides broad scoping (by namespace, zone, team, etc.) without requiring users to know SNI formats.
```

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
slonka: This thread is still open but `c5b5b2a` removed the outbound section entirely - including the MeshTimeout `to:` example.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+
+```yaml
+type: MeshTrafficPermission
+mesh: default
+name: backend-to-aurora
+spec:
+  targetRef:
+    kind: Dataplane
+    sectionName: ze-port
+  default:
```
lobkovilya: all MTP examples are slightly incorrect, it should be
```yaml
spec:
  rules:
     - default: {}
```

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+  rather than in `rules[].matches[]` — this is a consequence of the MADR-081 API design and
+  must be documented prominently for contributors.
+
+## Resources
+
+[MADR-095](095-mesh-scoped-zone-ingress-egress.md) resolved this by modelling zone proxies as
+mesh-scoped `Dataplane` resources with a `listeners` array.
+[MADR-062](062-meshexternalservice-and-zoneegress.md) established which `to` policies apply
+on zone egress for MeshExternalService traffic but deferred egress targeting and inbound policies.
+MADR-101 (pending) introduces a new KRI-based SNI format that makes SNIs human-readable and
```
lobkovilya:
```suggestion
MADR-101 (accepted) introduces a new KRI-based SNI format that makes SNIs human-readable and
```

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+enabling per-destination configuration on zone egress.
+
+#### Option A: SNI string match in `Match`
+
+Extend the `Match` struct with an `sni` field:
+
+```go
+type Match struct {
+    SpiffeID *SpiffeIDMatch `json:"spiffeID,omitempty"`
+    SNI      *SNIMatch      `json:"sni,omitempty"`
```
lobkovilya, line 240: Please add SNIMatch definition. Should we add `type: Exact` just to be future-proof?
```yaml
sni:
  type: Exact
  value: sni.meshexternalservice.default.zone-1.backend-ns.aws-aurora.8443
```

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+- With the KRI-based SNI format (MADR-101), SNIs are human-readable and predictable —
+  users construct them from resource attributes without querying the CP.
+- `sni` in `Match` is valid only when the policy targets a zone proxy Dataplane.
+
+`targetRef` in `Match` may be added as a follow-up once `sni` is proven in production.
+The `Match` struct can be extended without breaking existing policies.
+
+### Policy Examples
+
+#### Access control: allow specific service to reach specific external resource
```
lobkovilya: We need to make sure "deny always wins" principle is still applied, i.e. 

```yaml
type: MeshTrafficPermission
mesh: default
name: deny-aurora
spec:
  rules:
    - default:
        deny:
          - sni:
              value: "sni.extsvc.default.zone-1.backend-ns.aws-aurora.8443"
```

means no one can consume aws-aurora even if they apply MTP with `allow`. 

I think it will work out of the box with the algo, but worth adding a test.

Please add this as an example to the doc.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+  targetRef:
+    kind: Dataplane
+    sectionName: ze-port
+  default:
+    allow:
+      - spiffeID:
+          type: Exact
+          value: "spiffe://default/ns/backend-ns/sa/backend"
+        sni:
+          value: "sni.meshexternalservice.default.zone-1.backend-ns.aws-aurora.8443"
```
lobkovilya: sni format has short names, `extsvc` in this case

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+Priority column indicates planned release milestone.
+
+| Policy                    | Zone Egress               | Zone Ingress          | Priority      | Notes |
+|---------------------------|---------------------------|-----------------------|---------------|-------|
+| MeshTrafficPermission     | Yes                       | No                    | **2.14**      | Required for MeshExternalService access control to be complete |
+| MeshAccessLog             | Yes                       | Yes                   | **2.14**      | |
+| MeshMetric                | Yes                       | Yes                   | **2.14**      | |
+| MeshTrace                 | Yes                       | No                    | **2.14**      | Ingress is TCP-only — tracing not meaningful |
+| MeshProxyPatch            | Yes                       | Yes                   | 3.0           | Independent of SNI/MES designs; customer demand in ask-mesh |
+| MeshTLS                   | Yes                       | Yes                   | 3.0           | |
```
lobkovilya: MeshTLS for Zone Ingress is `No`, isn't it?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+| MeshRateLimit             | Maybe                     | No                    | Maybe         | |
+| MeshTimeout               | Maybe                     | No                    | Maybe         | |
+| MeshFaultInjection        | Maybe                     | No                    | Maybe         | |
+| MeshCircuitBreaker        | Maybe                     | No                    | Maybe         | |
```
lobkovilya: It's worth mentioning that today if you create MeshCircuitBreaker (and MeshHealthCheck, and maybe something else) with `to[].targetRef.kind: MeshExternalService` it will be applied on a legacy ZoneEgress. And that's not something we'd like to do for new zone proxies.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+`targetRef.kind: Dataplane` with `name`, `labels`, or `sectionName`. No special targeting
+mechanism is introduced.
+
+- Use `name` to target a specific zone proxy instance.
+- Use `labels` (e.g. `k8s.kuma.io/zone-proxy-type: egress`) to target all zone proxies of a type.
+  On Kubernetes these labels are set automatically, Universal users set them in their Dataplane template.
+- Use `sectionName` to target a specific listener when a Dataplane mixes zone proxy and regular
+  inbound listeners.
+
+### Destination Selector in Inbound Rules
```
lobkovilya: We need to mention somewhere that from the Kuma's point of view zone proxies don't have outbounds. So if user tries to create something like
```yaml
spec:
  targetRef:
    kind: Dataplane
    labels:
      k8s.kuima.io/zone-proxy-type: egress
  to:
    - targetRef:
        kind: MeshExternalService
        name: httpbin
      default: {}
```
then the policy will be simply ignored.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,391 @@
...
+- Outbound traffic (zone egress handles that)
+
+### Targeting
+
+Zone proxy `Dataplane` resources are targeted the same way as any other `Dataplane` — using
+`targetRef.kind: Dataplane` with `name`, `labels`, or `sectionName`. No special targeting
+mechanism is introduced.
+
+- Use `name` to target a specific zone proxy instance.
+- Use `labels` (e.g. `k8s.kuma.io/zone-proxy-type: egress`) to target all zone proxies of a type.
```
lahabana: Why `k8s.kuma.io` ? egress/ingress are not k8s specific

lahabana: Also can you remind me what's the value of this label when it's both and egress and an ingress?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lahabana: Does that mean we can select multiple services at once?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lahabana: I think it makes sense for these 3 to 3.0. I don't see why it wouldn't be valuable to have them no?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: That's an inaccuracy in the MADR, `k8s.kuma.io/zone-proxy-type` is a Kubernetes Service label, it's not propagated to Pod's labels. https://github.com/kumahq/kuma/blob/eb0eb0d8d77a7c264f7738483a3cf2766bbefdf7/docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md?plain=1#L176

At this moment we don't standardise zone proxy DPP labels at all. It's up to the user to add whatever label they need to their zone proxy deployment and later use it in the policies.

lobkovilya: We need some labels anyway for policy matching and for the further Inspect API work, so I've created a short MADR to standardise these labels https://github.com/kumahq/kuma/pull/16430

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: Added as 3.0

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: Can't answer this one. @lukidzi do you remember?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,551 @@
...
+Extend the `Match` struct with an `sni` field:
+
+```go
+type Match struct {
+    SpiffeID *SpiffeIDMatch `json:"spiffeID,omitempty"`
+    SNI      *SNIMatch      `json:"sni,omitempty"`
+}
+```
+
+With the KRI-based SNI format (MADR-101), SNIs follow the
```
lahabana: link please

lahabana (dismissed):

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,551 @@
...
+    sectionName: ze-port
+  rules:
+    - default:
+        allow:
+          - spiffeID:
+              type: Exact
+              value: "spiffe://default/ns/backend-ns/sa/backend"
+            sni:
+              type: Exact
+              value: "sni.extsvc.default.zone-1.aws-aurora.8443"
```
slonka, line 264: are we missing ns here?

```suggestion
              value: "sni.extsvc.default.zone-1.backend-ns.aws-aurora.8443"
```

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,551 @@
...
+| MeshLoadBalancingStrategy | Zone Egress    | Same  |
+
+The existence of these policies breaks the established rule "top-level targetRef selects what proxy to be modified",
+so we should not apply this approach to new mesh-scoped zone proxies.
+
+This transition can be made non-breaking by keying on how the MeshExternalService cluster is
+generated: when the cluster routes through the old global-scoped Zone Egress listener, the
+`EgressMatchedPolicies` path applies these policies to the egress as before. When the cluster
+routes through a mesh-scoped zone proxy Dataplane, the policies are applied on the sidecar's
+outbound cluster instead. Both paths can coexist during the migration window, so no UPGRADE.md
```
slonka: just to clarify - "no UPGRADE.md entry is required" for 2.14 right? For 3.0 we need to say that:

- `MeshCircuitBreaker` and `MeshHealthCheck` policies rewrite from `to[]` to `rules[].matches[].sni`
- 3.0 drops `MeshLoadBalancingStrategy` support
- `defaultForbidMeshExternalServiceAccess` flag will be removed

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,551 @@
...
+
+It needs to be rewritten to
+
+```yaml
+type: MeshCircuitBreaker
+spec:
+  targetRef:
+    kind: Dataplane
+    labels:
+      zone-egress: enabled
```
slonka: I see we dropped `k8s.kuma.io/zone-proxy-type` - should we have a paragraph about `zone-egress` label? For ingress would that just be `zone-ingress: enabled`? Is the label applied the same way?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,551 @@
...
+  Zone CPs, name-based policies must be applied on the Zone CP, not the Global CP.
+- Use `labels` to target a group of zone proxies (e.g. all proxies in a namespace). 
+  When the policy is applied on the Global CP, label-based matching is the only
+  approach since `name/namespace` matching doesn't work due to KDS hashing.
+- Use `sectionName` to target a specific listener when a Dataplane mixes zone proxy and regular
+  inbound listeners.
+
+Zone proxies have no outbound listeners, so **`spec.to[]` is ignored** when a policy targets a
+zone proxy Dataplane. Policies must use `spec.rules[]` instead. A policy applied with `spec.to[]`
+targeting a zone proxy will be silently skipped during xDS generation.
```
slonka: We're silently skipping because we don't want to create noise for the cases where user doesn't want to target zone proxies, right? Do we forsee this (using `.to`) being something that user's often do by accident? Should we have some feedback? Maybe a warning in status? WDYT?

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
slonka:
```suggestion
With the KRI-based SNI format [MADR-101](101-sni-format-improvements.md), SNIs follow the
```

slonka (commented):
Just a couple of things, looks quite good.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,551 @@
...
+         default:
+           backends:
+             - type: File
+               file:
+                 path: /tmp/access.log
+   ```
+
+   On which dataplanes should this log rule take effect?
+   - Currently, it takes effect only on mesh-scoped zone ingress
+   - Should it also apply to regular sidecars implementing the `backend` MeshService?
```
slonka: I think it's important to answer this question. For 2 (regular sidecars) would we convert SNI to KRI and match that way? ~~I'm leaning towards 1.~~ Actually if `kind: Mesh` should match everything then I'm leaning towards 2.

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: MeshExternalServices can't be in custom namespace, only system, so it's the second template from here https://github.com/kumahq/kuma/blob/45d8beeb9af9b92b9bd2c7ac5076daf57df81b4d/docs/madr/decisions/101-sni-format-improvements.md?plain=1#L124-L128

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: We agreed on new labels yesterday https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/102-zone-proxy-listener-computed-labels.md, I'm going to update the MADR to use new labels, `zone-egress: enabled` was just a placeholder

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: > MeshCircuitBreaker and MeshHealthCheck policies rewrite from to[] to rules[].matches[].sni

This one is required when transitioning from global scoped egress to mesh-scoped. However, "moving to mesh-scoped zone proxies" is definitely in 3.0 UPGRADE.md.

docs/madr/decisions/102-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: We usually skip if user tries to configure something on the DPP that doesn't exist on the DPP, i.e. setting timeout on `route-1` on `dpp-1` will be ignored when there is no such route:
```yaml
type: MeshTimeout
spec:
  targetRef:
    kind: Dataplane
    name: dpp-1
  to:
    - targetRef:
        kind: MeshHTTPRoute
        name: route-1
      default: $conf1
```

So if we want warnings in status it should be solved in general.

slonka (dismissed):

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,678 @@
...
+right fit. The `SNIMatch` type is shared across (1) and (3).
+
+The incomplete `rules[].matches[]` implementation tracked in
+[#16460](https://github.com/kumahq/kuma/issues/16460) must be finished as part of this work.
+
+##### `spec.to[]` is also accepted as an alias for `rules[].matches[].sni`
+
+Per Kuma's "apply where possible" style, `spec.to[]` on a zone proxy is not rejected for
+group (3) policies — it is accepted as an alias for `rules[].matches[].sni`. The CP
+resolves `spec.to[].targetRef → MeshExternalService → SNI`, locates the filter chain
```
slonka: don't we say on line [312](https://github.com/kumahq/kuma/pull/16275/changes#diff-0327bc432ca0ea8ab2952db943587540048dc2c857b1d8d10f065a76cbcafaadR312) that this resolution is bad?

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,678 @@
...
+
+* Good, because it is consistent with group (2) (`MeshCircuitBreaker`, `MeshHealthCheck`) on zone proxies.
+* Good, because users can target a `MeshExternalService` by labels without knowing the SNI format.
+* Bad, because it cannot express source-conditioned per-destination policies like "access log
+  requests from spiffeID X to MeshExternalService Y" — `spec.to[]` selects only the destination
+  filter chain and has no place to constrain the source identity.
+
+##### Decision
+
+Implement Option A: `rules[].matches[].sni` selects the `filterChain` on the zone proxy.
```
slonka, line 547: does **zone ingress + `spiffeID` matching** work? since we don't terminate mTLS it can't right? should we mention it here?

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,678 @@
...
+No per-destination granularity — these configure the entire proxy. Use top-level `targetRef`
+(optionally with `sectionName`) and nothing else.
+
+```yaml
+type: MeshMetric
+spec:
+  targetRef:
+    kind: Dataplane
+    labels:
+      kuma.io/listener-zoneegress: enabled
```
slonka, line 109: nitpick, this could be new to some folks, might be worth linking in the comments to https://github.com/kumahq/kuma/blob/4a69db53d3656d6b8d23ff099b2f5bdfd4817836/docs/madr/decisions/102-zone-proxy-listener-computed-labels.md?plain=1#L36

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,678 @@
...
+}
+```
+
+- `sni` maps directly to Envoy's `server_names` filter chain match — no CP resolution step.
+- `sni` is optional. When absent, the match applies to all destinations (same as today).
+  When present, the match applies only to the filter chain whose SNI matches.
+- When a `Match` contains both `spiffeID` and `sni`, both conditions must hold.
+- With the KRI-based SNI format [MADR-101](101-sni-format-improvements.md), SNIs are human-readable and predictable —
+  users construct them from resource attributes without querying the CP.
+- `sni` in `Match` is not limited to zone proxy Dataplanes. It applies wherever the policy
```
slonka, line 346: by "first supported customer" you mean that it will also work for regular inbouds sometime in the future right? should we explicitly mention that?

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,678 @@
...
+
+## Summary
+
+Builds on [MADR-095](095-mesh-scoped-zone-ingress-egress.md), which turned zone proxies into
+mesh-scoped `Dataplane` resources, and defines how policies target them. Zone proxy Dataplanes
+are selected with the standard `targetRef.kind: Dataplane` (`name`, `labels`, `sectionName`) —
+no special targeting. Each policy group uses a different destination selector based on what its
+data model can express; the four groups are summarized below with a small example for each.
+
+Ships in 2.14 (no upgrade actions required). 3.0 removes the legacy global-scoped zone proxies
```
slonka, line 15: The previous version had a Policy Matrix table that was explicit per-policy: `MeshTrafficPermission` and `MeshAccessLog`/`MeshMetric` in 2.14, `MeshRateLimit`/`MeshTimeout`/`MeshFaultInjection`/`MeshCircuitBreaker`/`MeshHealthCheck` in 3.0. Does this now mean everything ships in 2.14?

lobkovilya: No, I just think this design is orthogonal to the planning. There are no technical limitations that require something to be released in 3.0 instead of 2.14. If something slips to 3.0, it simply means we didn’t have enough time.

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: I don't think just the fact of resolution is bad, I'll update 312

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: Added a note:

> **Zone ingress note**: zone ingress does not terminate mTLS, so no SPIFFE identity is available on its filter chains. A `rules[].matches[].spiffeID` entry on a zone ingress is syntactically valid but will never match traffic.

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: I just removed this part, we should get sidecar and zone proxy support at the same time

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,686 @@
...
+Top-level `targetRef.sectionName` still should work and select the `listener` by its name.
+
+For example,
+
+```yaml
+type: MeshCircuitBreaker
+spec:
+  targetRef:
+    kind: Dataplane
+    sectionName: ze-port
```
lahabana, line 460: If I don't do `sectionName` just `targetRef: kind: Dataplane` (all dataplanes selected) will it still apply this to the proxy?

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,686 @@
...
+care about the source, so reusing the matcher API introduced for `MeshTrafficPermission` is the
+right fit. The `SNIMatch` type is shared across (1) and (3).
+
+The incomplete `rules[].matches[]` implementation tracked in
+[#16460](https://github.com/kumahq/kuma/issues/16460) must be finished as part of this work.
+
+##### `spec.to[]` is also accepted as an alias for `rules[].matches[].sni`
+
+Per Kuma's "apply where possible" style, `spec.to[]` on a zone proxy is not rejected for
+group (3) policies — it is accepted as an alias for `rules[].matches[].sni`. The CP
```
lahabana: Do we need this? Seems like a case of creating 2 ways to do 1 thing no?

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: Yes, it will be applied to all Dataplanes including the ones that have ZoneIngress/ZoneEgress listeners

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
```diff
@@ -0,0 +1,686 @@
...
+        prometheus:
+          port: 5670
+          path: /metrics
+```
+
+### Out of scope
+
+- `MeshRetry` on zone egress — squared retry amplification.
+- `MeshOPA` — was never supported on zone egress; needs its own decision.
+- `MeshGlobalRateLimit` — no action required.
```
lobkovilya, line 122: nit: add MLBS

docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md
lobkovilya: Yes, we don't need it. We discussed offline with the team it'll be easier if we don't support it. The only downside is when user applies MAL (or MT):
```yaml
type: MeshAccessLog
spec:
  targetRef:
    kind: Dataplane
    labels:
       kuma.io/listeners-zoneegress: enabled
  to:
     - targetRef:
         kind: MeshExternalService
         labels: 
           kuma.io/display-name: httpbin
       default: $conf
```
it will be silently ignored on zone egress. But we'll make sure it's covered in the docs.

lukidzi:
👍

slonka (approved):

PR #16278, opened 2026-04-14
fix(config): check if domain starts with dot

lukidzi:
## Motivation

`KUMA_DNS_SERVER_DOMAIN` (default: `mesh`) is used to build DNS entries like `<service>.mesh`. A domain starting with a dot (e.g.`.mesh.local`) would produce invalid hostnames such as `<service>..mesh.local`, causing DNS resolution failures with no clear error at startup.

## Implementation information

Added an early check in `Config.Validate()` (`pkg/config/dns-server/config.go`) that rejects any `Domain` value with a leading dot and returns `"domain must not start with a dot"

### DIFF
--- a/pkg/config/dns-server/config.go
+++ b/pkg/config/dns-server/config.go
@@ -3,6 +3,7 @@ package dns_server
 import (
 	"errors"
 	"net"
+	"strings"
 
 	"github.com/kumahq/kuma/v2/pkg/config"
 )
@@ -22,6 +23,9 @@ type Config struct {
 }
 
 func (g *Config) Validate() error {
+	if strings.HasPrefix(g.Domain, ".") {
+		return errors.New("domain must not start with a dot")
+	}
 	_, _, err := net.ParseCIDR(g.CIDR)
 	if err != nil {
 		return errors.New("CIDR must be valid")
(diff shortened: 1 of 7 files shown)


lobkovilya:
This check is useful, but I'd also add another check to HostnameGenerator to make sure it always produces valid hostnames.

pkg/config/dns-server/config.go
```diff
@@ -22,6 +23,9 @@ type Config struct {
 }
 
 func (g *Config) Validate() error {
+	if strings.HasPrefix(g.Domain, ".") {
```
slonka, line 26: should we use `k8s_validation.IsDNS1123Subdomain(g.Domain)` here as well?

slonka (commented):
Do we need UPGRADE.md note?

pkg/config/dns-server/config.go
lukidzi: No, we can't as it checks if has `.` at the begining

slonka: so why is this using `HasPrefix` and `EvaluateTemplate` uses `IsDNS1123Subdomain`?

lukidzi: Because they are 2 different things, one has only suffix (domain, legacy DNS), hostname generator has a whole resolved address.

slonka (approved):

PR #16281, opened 2026-04-15
test(e2e): fix loadbalancing test

lukidzi:
## Motivation

When the MLBS is applied at line 370, the control plane pushes updated EDS to remove kuma-5 (UniZone2) endpoints from the cluster. Before that push arrives, 37 requests (~18%) hit kuma-5 and return 2xx instead of 5xx. This consistently gives 65 5xx instead of ~100, below the ±25 tolerance.

## Implementation information

Wrapped the traffic generation and metric assertions in Eventually("60s", "10s") with a counter reset each iteration. This retries until EDS has fully converged and the 50/50 split produces the expected failure distribution. The FlakeAttempts(3) decorator can be removed from this test too, but that's a separate cleanup.

## Supporting documentation

> Changelog: skip


test/e2e_env/multizone/localityawarelb/locality_aware_multizone_hybrid.go
lukidzi:
```suggestion

```

bartsmykla (approved):

PR #16282, opened 2026-04-16
feat(helm): allow to customize san

lukidzi:
## Motivation

When using cert-manager to manage control plane TLS certificates, the generated Certificate resource had hardcoded SANs (service.namespace, .svc, .svc.cluster.local). There was no way to include additional DNS names: for example, when the control plane is exposed under a custom domain or behind an ingress with a specific hostname.

## Implementation information

Moved the default dnsNames from hardcoded template logic into values.yaml as a configurable list under `controlPlane.tls.general.certManager.dnsNames`.

### DIFF
--- a/deployments/charts/kuma/templates/cp-certs-cert-manager.yaml
+++ b/deployments/charts/kuma/templates/cp-certs-cert-manager.yaml
@@ -31,7 +31,11 @@ spec:
   {{- end }}
   dnsNames:
     {{- range $dnsNames }}
-    - {{ tpl . $ | quote }}
+    {{- $rendered := tpl . $ }}
+    {{- if eq $rendered "" }}
+      {{- fail "controlPlane.tls.general.certManager.dnsNames must not contain empty strings" }}
+    {{- end }}
+    - {{ $rendered | quote }}
     {{- end }}
   issuerRef:
     kind: {{ $issuerKind }}
--- a/deployments/charts/kuma/templates/cp-certs-cert-manager.yaml
+++ b/deployments/charts/kuma/templates/cp-certs-cert-manager.yaml
@@ -25,9 +25,13 @@ metadata:
   labels: {{ include "kuma.cpLabels" . | nindent 4 }}
 spec:
   isCA: true
+  {{- $dnsNames := .Values.controlPlane.tls.general.certManager.dnsNames }}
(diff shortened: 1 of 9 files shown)


deployments/charts/kuma/templates/cp-certs-cert-manager.yaml
```diff
@@ -26,10 +25,14 @@ metadata:
...
+  {{- $dnsNames := .Values.controlPlane.tls.general.certManager.dnsNames }}
+  {{- if not $dnsNames }}
+    {{- fail "controlPlane.tls.general.certManager.dnsNames must not be empty" }}
+  {{- end }}
   dnsNames:
-    - {{ $serviceName }}.{{ .Release.Namespace }}
-    - {{ $serviceName }}.{{ .Release.Namespace }}.svc
-    - {{ $serviceName }}.{{ .Release.Namespace }}.svc.cluster.local
+    {{- range $dnsNames }}
+    - {{ tpl . $ | quote }}
```
slonka: Should we consider `extraDnsNames` list that is appended to a hardcoded ones? We already use `extra*`/`additional*` convention (`extraLabels`, `extraConfigMaps`, `additionalRules`).

deployments/charts/kuma/templates/cp-certs-cert-manager.yaml
```diff
@@ -26,10 +25,14 @@ metadata:
   labels: {{ include "kuma.cpLabels" . | nindent 4 }}
 spec:
   isCA: true
+  {{- $dnsNames := .Values.controlPlane.tls.general.certManager.dnsNames }}
+  {{- if not $dnsNames }}
```
slonka, line 29: nitpick: what if someone does `[""]` by mistake? won't it render as `- ""` and get rejected by cert-manager at runtime?

slonka (commented):
Looks good overall - 1 Q.

deployments/charts/kuma/templates/cp-certs-cert-manager.yaml
lukidzi: Don't like the idea, if someone runs it in a different way and won't need these addresses won't be able to remove them

slonka (approved):

PR #16294, opened 2026-04-17
feat(kuma-dp): gate /ready on DNS proxy config with 15s timeout

lukidzi:
## Motivation

 The `kuma.io/wait-for-dataplane-ready` annotation tells the kubelet to hold app containers until the sidecar `/ready` returns 200. That endpoint fires as soon as Envoy reports xDS sync, but the embedded DNS proxy (dnsproxy) gets its VIP records through a separate ConfigFetcher poll of Envoy's `/dns` dynamic-config route. The two paths race.
The result: the kubelet marks the pod ready, the app starts, immediately resolves a `.mesh` name, and gets NXDOMAIN because the DNS map is still empty. The `wait_for_envoy` E2E test (`test/e2e_env/kubernetes/graceful/wait_for_envoy.go`) exercises this path with `restartPolicy: Never`, turning the race into a hard failure rather than a transient one.

## Implementation information

- `dnsproxy.Server` gains `ConfigReady() <-chan struct{}`, closed once after the first successful `ReloadMap`. On shutdown the channel closes too, so no caller blocks.
- `readiness.Reporter` accepts an optional `dnsConfigReady` channel. When set, `/ready` returns 503 until the channel closes. After a 15-second deadline the gate is bypassed with a warning log, avoiding indefinite NotReady on misconfigured deployments.
- `run.go` wires the channel only for the embedded proxy path (`cfg.DNS.ProxyPort != 0`). The CoreDNS path and zone proxies pass nil and are unaffected.
- Metrics: `kuma_dp_dns_config_ready_wait_seconds` histogram (time from start to first config)
- Tests: gate blocks → 503; gate opens → 200; past-deadline reporter bypasses → 200 immediately.

## Supporting documentation

Fix #XX

### DIFF
--- a/app/kuma-dp/pkg/dataplane/dnsproxy/component.go
+++ b/app/kuma-dp/pkg/dataplane/dnsproxy/component.go
@@ -288,12 +288,10 @@ func (s *Server) ReloadMap(ctx context.Context, reader io.Reader) error {
 	if err != nil {
 		return err
 	}
-	if configuration.ExtraLabels != nil {
-		s.metricsOnce.Do(func() {
-			m := newMetrics(s.registerer, configuration.ExtraLabels)
-			s.metrics.Store(m)
-		})
-	}
+	s.metricsOnce.Do(func() {
+		m := newMetrics(s.registerer, configuration.ExtraLabels)
+		s.metrics.Store(m)
+	})
 	res := dnsMap{
 		ARecords:    make(map[string]*dnsEntry),
 		AAAARecords: make(map[string]*dnsEntry),
--- a/app/kuma-dp/pkg/dataplane/dnsproxy/component.go
+++ b/app/kuma-dp/pkg/dataplane/dnsproxy/component.go
@@ -68,13 +68,14 @@ func WithRegisterer(r prometheus.Registerer) ServerOption {
 // NewServerWithCustomClient is used for testing purposes
 func NewServerWithCustomClient(addresses []string, upstreamClient func(msg *dns.Msg) (*dns.Msg, error), opts ...ServerOption) *Server {
 	s := Server{
-		addresses:      addresses,
-		componentDone:  make(chan struct{}),
-		dnsMap:         atomic.Pointer[dnsMap]{},
-		ready:          make(chan struct{}),
-		configReady:    make(chan struct{}),
-		registerer:     prometheus.DefaultRegisterer,
-		upstreamClient: upstreamClient,
+		addresses:       addresses,
+		componentDone:   make(chan struct{}),
+		dnsMap:          atomic.Pointer[dnsMap]{},
+		ready:           make(chan struct{}),
+		configReady:     make(chan struct{}),
+		registerer:      prometheus.DefaultRegisterer,
+		upstreamClient:  upstreamClient,
+		configReadyTime: time.Now(),
 	}
 	for _, opt := range opts {
 		opt(&s)
@@ -167,7 +168,6 @@ func (s *Server) Handler(res dns.ResponseWriter, req *dns.Msg) {
 
 func (s *Server) Start(stop <-chan struct{}) error {
 	defer close(s.componentDone)
-	s.configReadyTime = time.Now()
 
 	servers := make([]*dns.Server, len(s.addresses))
 	for i, addr := range s.addresses {
@@ -332,10 +332,8 @@ func (s *Server) ReloadMap(ctx context.Context, reader io.Reader) error {
 	log.V(1).Info("DNS proxy configured", "config", res)
 	s.dnsMap.Store(&res)
 	s.configReadyOnce.Do(func() {
-		if !s.configReadyTime.IsZero() {
-			if m := s.metrics.Load(); m != nil {
-				m.ConfigReadyWaitSeconds.Observe(time.Since(s.configReadyTime).Seconds())
-			}
+		if m := s.metrics.Load(); m != nil {
+			m.ConfigReadyWaitSeconds.Observe(time.Since(s.configReadyTime).Seconds())
 		}
 		log.Info("DNS proxy received first configuration", "records", len(res.ARecords))
 		close(s.configReady)
--- a/app/kuma-dp/pkg/dataplane/dnsproxy/component.go
+++ b/app/kuma-dp/pkg/dataplane/dnsproxy/component.go
@@ -30,8 +30,11 @@ type Server struct {
 	registerer    prometheus.Registerer
 	metricsOnce   sync.Once
 	// upstreamClient is used for testing purposes
-	upstreamClient func(msg *dns.Msg) (*dns.Msg, error)
-	ready          chan struct{}
+	upstreamClient  func(msg *dns.Msg) (*dns.Msg, error)
+	ready           chan struct{}
+	configReady     chan struct{}
+	configReadyOnce sync.Once
+	configReadyTime time.Time
 }
 
 func NewServer(addresses []string) (*Server, error) {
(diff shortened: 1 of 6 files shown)


app/kuma-dp/pkg/dataplane/dnsproxy/metrics.go
```diff
@@ -51,13 +52,21 @@ func newMetrics(registerer prometheus.Registerer, constLabels prometheus.Labels)
 		ConstLabels: constLabels,
 	})
 	registerer.MustRegister(entriesTotal)
+	configReadyWaitSeconds := prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name:        "kuma_dp_dns_config_ready_wait_seconds",
+		Help:        "Time between DNS proxy creation and first configuration received from Envoy.",
+		ConstLabels: constLabels,
+		Buckets:     []float64{0.1, 0.5, 1, 2, 5, 10, 30},
+	})
+	registerer.MustRegister(configReadyWaitSeconds)
```
slonka, line 61: there is `kuma_dp_dns_config_ready_gate_bypassed_total` in the description but I don't see it registered here, is it missing?

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -65,12 +68,14 @@ func WithRegisterer(r prometheus.Registerer) ServerOption {
...
-		registerer:     prometheus.DefaultRegisterer,
-		upstreamClient: upstreamClient,
+		addresses:       addresses,
+		componentDone:   make(chan struct{}),
+		dnsMap:          atomic.Pointer[dnsMap]{},
+		ready:           make(chan struct{}),
+		configReady:     make(chan struct{}),
+		registerer:      prometheus.DefaultRegisterer,
+		upstreamClient:  upstreamClient,
+		configReadyTime: time.Now(),
```
slonka: Nit: `configReadyTime` is captured at `NewServerWithCustomClient` return, not at `Start`. The histogram `kuma_dp_dns_config_ready_wait_seconds` therefore includes the (usually tiny) gap between construction and listen, plus any time the ComponentManager spent queuing this component. Fine in practice, but consider moving to the top of `Start` so the metric matches the help text ("DNS proxy creation") more literally and is less sensitive to changes in component ordering.

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -314,6 +331,13 @@ func (s *Server) ReloadMap(ctx context.Context, reader io.Reader) error {
 	}
 	log.V(1).Info("DNS proxy configured", "config", res)
 	s.dnsMap.Store(&res)
+	s.configReadyOnce.Do(func() {
+		if m := s.metrics.Load(); m != nil {
+			m.ConfigReadyWaitSeconds.Observe(time.Since(s.configReadyTime).Seconds())
+		}
+		log.Info("DNS proxy received first configuration", "records", len(res.ARecords))
+		close(s.configReady)
+	})
```
slonka: If the first `ReloadMap` arrives with `configuration.ExtraLabels == nil`, `s.metrics.Load()` returns nil here (metrics initialization is gated on `ExtraLabels != nil` at line 291). In that case `ConfigReadyWaitSeconds` silently loses its one and only sample - `configReadyOnce.Do` fires, and subsequent reloads can't record it either. Likely fine if the CP always sends labels, but worth either (a) asserting ExtraLabels is always present, or (b) initializing metrics with empty const labels on first reload when none are provided.

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -100,9 +119,9 @@ func (r *Reporter) Start(stop <-chan struct{}) error {
 		ErrorLog:          adapter.ToStd(logger),
 	}
 
-	errCh := make(chan error)
+	errCh := make(chan error, 1)
 	go func() {
-		if err := server.Serve(lis); err != nil {
+		if err := server.Serve(lis); err != nil && !errors.Is(err, http.ErrServerClosed) {
 			errCh <- err
 		}
```
slonka: Nice side-fix worth calling out: buffering `errCh` to capacity 1 plus the `errors.Is(err, http.ErrServerClosed)` filter closes a pre-existing goroutine leak. Before this change, if `stop` fired first and `server.Serve` subsequently returned `ErrServerClosed`, the goroutine would block forever on the unbuffered send. Maybe add a line to the PR description so this fix is visible when the blame lands here later.

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -126,6 +145,26 @@ func (r *Reporter) handleReadiness(writer http.ResponseWriter, req *http.Request
 		return
 	}
 
+	// Gate readiness on DNS proxy receiving its first config from Envoy.
+	// This ensures mesh-generated DNS names, resolve before app containers start.
```
slonka: nitpick: I don't think you need a comma there

```suggestion
	// This ensures mesh-generated DNS names resolve before app containers start.
```

app/kuma-dp/pkg/dataplane/dnsproxy/component.go
```diff
@@ -314,6 +331,13 @@ func (s *Server) ReloadMap(ctx context.Context, reader io.Reader) error {
 	}
 	log.V(1).Info("DNS proxy configured", "config", res)
 	s.dnsMap.Store(&res)
+	s.configReadyOnce.Do(func() {
+		if m := s.metrics.Load(); m != nil {
```
slonka, line 333: can this be nil? do we need that check on ExtraLabels on line [291](https://github.com/kumahq/kuma/pull/16294/changes#diff-af8f2fd7111fd43e40e5f18e84c366ac36c0f3617216698854b4aba23d194655R291) ? if this is somehow nil we miss the metric recording

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -100,9 +119,9 @@ func (r *Reporter) Start(stop <-chan struct{}) error {
 		ErrorLog:          adapter.ToStd(logger),
 	}
 
-	errCh := make(chan error)
+	errCh := make(chan error, 1)
 	go func() {
-		if err := server.Serve(lis); err != nil {
+		if err := server.Serve(lis); err != nil && !errors.Is(err, http.ErrServerClosed) {
```
slonka, line 127: seems a bit unrelated, could you add a comment why this was changed?

app/kuma-dp/pkg/dataplane/readiness/component_test.go
```diff
@@ -47,6 +47,84 @@ var _ = Describe("Readiness Reporter", func() {
...
+			dnsReady = make(chan struct{})
+			reporter = readiness.NewReporter(true, "", "127.0.0.1", port, "", dnsReady)
+			go func() {
+				defer GinkgoRecover()
+				_ = reporter.Start(stopCh)
+			}()
+			Eventually(func() error {
+				_, err := http.Get(baseURL + "/ready")
+				return err
+			}, 5*time.Second, 50*time.Millisecond).Should(Succeed())
```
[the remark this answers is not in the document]
slonka: is this wrong?

app/kuma-dp/pkg/dataplane/dnsproxy/metrics.go
lukidzi: Updated description

slonka (approved):

PR #16326, opened 2026-04-20
fix(kds): reconnect mux client when GlobalToZone stream is closed by …

### LINKED ISSUE #16428: KDS GlobalToZone stream silently dies after global CP restart
### Kuma Version

2.13.0

### Describe the bug

When the global CP restarts behind a gateway (load balancer), the zone CP's GlobalToZone sync stream can permanently die while the zone appears fully connected. Healthchecks pass, ZoneToGlobal syncs normally, but the zone never receives resources from global again.

Observed once. Requires specific timing between gRPC stream closure, context cancellation, and goroutine scheduling — unlikely but not impossible during routine deploys.

### To Reproduce

1. Build control plane:
```
     go build -o build/kuma-cp ./app/kuma-cp/
```
2. Start global CP (`KUMA_MODE=global`) and zone CP (`KUMA_MODE=zone`) on different ports
3. Patch `zone_sync.go` `GlobalToZoneSync` handler to simulate gateway closing the stream:
  ```go
  // Add to the select in GlobalToZoneSync handler
  case <-time.After(30 * time.Second):
      // Simulate gateway terminating HTTP/2 stream during connection draining.
      // Returning nil causes the zone to receive io.EOF from Recv().
      return nil
```
  4. After 30s, global logs show stream closed. Zone keeps sending healthchecks but receives no resources.
  5. Create a mesh on global:
 ``` curl -X PUT http://localhost:5681/meshes/test-mesh -H 'content-type: application/json' -d '{"type":"Mesh","name":"test-mesh"}'```
  6. Zone never receives it:
  `curl http://localhost:15681/meshes`
test-mesh absent

### Expected behavior

Streams reconnects

_No response_

### PR DESCRIPTION
**Pull request by lukidzi**
## Motivation
  - Fix a bug where the KDS mux client's GlobalToZone (and ZoneToGlobal) sync goroutine silently exited when the server closed the gRPC stream cleanly (io.EOF), leaving the zone permanently unable to receive resources from the global CP
  - The mux client stayed alive (healthchecks, other streams kept working) masking the failure — the zone appeared connected but never received new policies or meshes
  - Add regression test that verifies the mux client reconnects when the server closes the GlobalToZone stream

## Implementation information

When a gRPC stream is closed cleanly by the server, `stream.Recv()` returns `io.EOF`. `KDSSyncClient.Receive()` translates `io.EOF` to a nil return. `startGlobalToZoneSync` only sent to errorCh on non-nil errors, so it exited without notifying `client.Start()`. Since `Start()` blocks on errorCh, the mux client stayed alive indefinitely with a dead `GlobalToZone` stream.

This can be triggered by a global CP restart behind a load balancer that closed the HTTP/2 stream with a clean TCP FIN instead of a gRPC error. Both `startGlobalToZoneSync` and `startZoneToGlobalSync` now send to errorCh on nil error (stream closed by server), causing `Start()` to return and ResilientComponent to restart the entire mux client. A non-blocking select on `ctx.Done()` prevents goroutine leaks when shutdown is already in progress.

Also, there was a data race
  - NewDeltaKDSStream now returns (DeltaKDSStream, <-chan struct{}) — the channel closes when sendLoop exits
  - client.go: <-sendDone in the defer before CloseSend() — guarantees Send() is done
  - zone_sync.go: receives sendDone but ignores it (no CloseSend in that goroutine)

Fix: https://github.com/kumahq/kuma/issues/16428

### DIFF
--- a/pkg/kds/mux/client.go
+++ b/pkg/kds/mux/client.go
@@ -157,19 +157,18 @@ func (c *client) startGlobalToZoneSync(ctx context.Context, log logr.Logger, con
 
 	cfgJson, err := config.ConfigForDisplay(pointer.To(c.rt.Config()))
 	if err != nil {
-		nonBlockingSend(ctx, errorCh, errors.Wrap(err, "could not marshall config to json"))
+		trySend(ctx, errorCh, errors.Wrap(err, "could not marshall config to json"))
 		return
 	}
 
 	stream, err := kdsClient.GlobalToZoneSync(ctx)
 	if err != nil {
-		nonBlockingSend(ctx, errorCh, err)
+		trySend(ctx, errorCh, err)
 		return
 	}
-	kdsStream, sendDone := kds_client_v2.NewDeltaKDSStream(stream, c.clientID, c.rt.GetInstanceId(), cfgJson, len(c.typesSentByGlobal))
+	kdsStream := kds_client_v2.NewDeltaKDSStream(stream, c.clientID, c.rt.GetInstanceId(), cfgJson, len(c.typesSentByGlobal))
 	defer func() {
-		<-sendDone // wait for sendLoop to exit before CloseSend to avoid data race
-		if err := stream.CloseSend(); err != nil {
+		if err := kdsStream.CloseSend(); err != nil {
 			log.Error(err, "CloseSend returned an error")
 		}
 	}()
@@ -189,7 +188,7 @@ func (c *client) startGlobalToZoneSync(ctx context.Context, log logr.Logger, con
 	)
 
 	if err := syncClient.Receive(); err != nil && !errors.Is(err, context.Canceled) {
-		nonBlockingSend(ctx, errorCh, errors.Wrap(err, "GlobalToZoneSyncClient finished with an error"))
+		trySend(ctx, errorCh, errors.Wrap(err, "GlobalToZoneSyncClient finished with an error"))
 	}
 	log.Info("GlobalToZoneSync finished gracefully")
 }
@@ -200,7 +199,7 @@ func (c *client) startZoneToGlobalSync(ctx context.Context, log logr.Logger, con
 	log.Info("initializing Kuma Discovery Service (KDS) stream for zone to global sync of resources with delta xDS")
 	stream, err := kdsClient.ZoneToGlobalSync(ctx)
 	if err != nil {
-		nonBlockingSend(ctx, errorCh, err)
+		trySend(ctx, errorCh, err)
 		return
 	}
 	defer func() {
@@ -217,18 +216,22 @@ func (c *client) startZoneToGlobalSync(ctx context.Context, log logr.Logger, con
 	}
 
 	if err != nil && !errors.Is(err, context.Canceled) {
-		nonBlockingSend(ctx, errorCh, errors.Wrap(err, "ZoneToGlobalSync finished with an error"))
+		trySend(ctx, errorCh, errors.Wrap(err, "ZoneToGlobalSync finished with an error"))
 	}
 	log.Info("ZoneToGlobalSync finished gracefully")
 }
 
-// nonBlockingSend sends err to errorCh without blocking. If the parent context
-// is already canceled (another stream triggered a restart, or stop was
-// called), the send is skipped to avoid goroutine leaks.
-func nonBlockingSend(ctx context.Context, errorCh chan error, err error) {
+// trySend attempts to send err to errorCh. If the context is already
+// done (another stream triggered a restart, or the app is shutting
+// down), the error is dropped. If errorCh already has an error from
+// another stream, the error is also dropped.
+func trySend(ctx context.Context, errorCh chan error, err error) {
+	if ctx.Err() != nil {
+		return
+	}
 	select {
 	case errorCh <- err:
-	case <-ctx.Done():
+	default:
 	}
 }
 
@@ -243,7 +246,7 @@ func (c *client) startXDSConfigs(
 	log.Info("initializing rpc stream for executing config dump on data plane proxies")
 	stream, err := client.StreamXDSConfigs(ctx)
 	if err != nil {
-		nonBlockingSend(ctx, errorCh, err)
+		trySend(ctx, errorCh, err)
 		return
 	}
 
@@ -263,7 +266,7 @@ func (c *client) startStats(
 	log.Info("initializing rpc stream for executing stats on data plane proxies")
 	stream, err := client.StreamStats(ctx)
 	if err != nil {
-		nonBlockingSend(ctx, errorCh, err)
+		trySend(ctx, errorCh, err)
 		return
 	}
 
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: pkg/kds/mux/client.go [hunk 1]
```diff
@@ -186,12 +188,23 @@ func (c *client) startGlobalToZoneSync(ctx context.Context, log logr.Logger, con
...
 		return
 	}
-
-	log.Info("GlobalToZoneSync finished gracefully")
+	if err != nil {
+		err = errors.Wrap(err, "GlobalToZoneSyncClient finished with an error")
+	} else {
+		// err == nil means the stream was closed by the server (io.EOF).
+		// Trigger a reconnect so ResilientComponent restarts all streams.
+		err = errors.New("GlobalToZoneSync stream closed by server")
```
lobkovilya: Wouldn’t the more correct fix be to remove this check in `kds_client.go` https://github.com/kumahq/kuma/blob/dc7932067bb4678e9f0b4681a10b8efc2cc72f0b/pkg/kds/v2/client/kds_client.go#L90-L92

It seems like in `kds_client.go` we suppress the error just to bring it back in `mux/client.go`

lukidzi: done

pkg/kds/mux/client.go
```diff
@@ -186,12 +188,12 @@ func (c *client) startGlobalToZoneSync(ctx context.Context, log logr.Logger, con
...
 
-	if err := syncClient.Receive(); err != nil && !errors.Is(err, context.Canceled) {
-		errorCh <- errors.Wrap(err, "GlobalToZoneSyncClient finished with an error")
-		return
+	err = syncClient.Receive()
+	log.Info("GlobalToZoneSync ended", "reason", err)
+	select {
+	case errorCh <- errors.Wrap(err, "GlobalToZoneSyncClient finished with an error"):
+	case <-ctx.Done():
 	}
```
lukidzi: If I am not wrong this code is only called by the Zone. In this case we shouldn't return gracefully as it wouldn't trigger reconnection. @lobkovilya WDYT

lobkovilya: We should return gracefully only when the Zone CP itself is shutting down, either because it is restarting or because another stream has failed. I think that is why we used to check `errors.Is(err, context.Canceled)`: no one other than the Zone CP can cancel that context, so if it is canceled, it means the Zone CP itself decided to stop and restart.

pkg/kds/mux/client.go
```diff
@@ -166,7 +166,9 @@ func (c *client) startGlobalToZoneSync(ctx context.Context, log logr.Logger, con
 		errorCh <- err
```
lobkovilya: It's kind of problematic given the `errorCh` is a blocking channel that's shared between 5 streams, isn't it? Can't the stream stuck forever on write if there was error in different stream?

pkg/kds/mux/client.go
```diff
@@ -217,11 +217,19 @@ func (c *client) startZoneToGlobalSync(ctx context.Context, log logr.Logger, con
...
+		nonBlockingSend(ctx, errorCh, errors.Wrap(err, "ZoneToGlobalSync finished with an error"))
 	}
+	log.Info("ZoneToGlobalSync finished gracefully")
+}
 
-	log.V(1).Info("ZoneToGlobalSync finished gracefully")
+// nonBlockingSend sends err to errorCh without blocking. If the parent context
+// is already canceled (another stream triggered a restart, or stop was
+// called), the send is skipped to avoid goroutine leaks.
+func nonBlockingSend(ctx context.Context, errorCh chan error, err error) {
```
lobkovilya: nit: probably `trySend` is a better name, `nonBlockingSend` implies send is going to happen either way, while `try` means send can be skipped

pkg/kds/mux/client.go
```diff
@@ -217,11 +217,19 @@ func (c *client) startZoneToGlobalSync(ctx context.Context, log logr.Logger, con
...
+	select {
+	case errorCh <- err:
+	case <-ctx.Done():
```
lobkovilya: I think the method first should check if `ctx` is done and after that attempt to write to `errorCh`:
```go
if ctx.Err() != nil {
    // Context is done either because app is shutting down or because there was an error in another stream.
    // Either way we don't need to write err to errorCh
    return
}
select {
case errorCh <- err:
default:
    // dropping error because there is already an error from another stream
}
```
otherwise when `ctx` is done it's not deterministic whether we write error to the channel or we don't.

pkg/kds/mux/client.go
```diff
@@ -157,16 +157,18 @@ func (c *client) startGlobalToZoneSync(ctx context.Context, log logr.Logger, con
...
 
 	stream, err := kdsClient.GlobalToZoneSync(ctx)
 	if err != nil {
-		errorCh <- err
+		nonBlockingSend(ctx, errorCh, err)
 		return
 	}
+	kdsStream, sendDone := kds_client_v2.NewDeltaKDSStream(stream, c.clientID, c.rt.GetInstanceId(), cfgJson, len(c.typesSentByGlobal))
 	defer func() {
+		<-sendDone // wait for sendLoop to exit before CloseSend to avoid data race
```
lobkovilya: That's a valid issue but I think it should be fixed differently. The code in `mux/client.go` should never work directly with `stream`, always through the wrapper we create in `NewDeltaKDSStream`.

This means we need to add `CloseSend` method to the wrapper:
```go
// All methods other than Receive() are non-blocking. It does not wait until the peer CP receives the message.
type DeltaKDSStream interface {
	DeltaDiscoveryRequest(resourceType core_model.ResourceType) error
	Receive() (UpstreamResponse, error)
	ACK(resourceType core_model.ResourceType) error
	NACK(resourceType core_model.ResourceType, err error) error
    CloseSend() error // new method
}
```

that way `stream.CloseSend()` will be called within the `sendLoop`, so it's thread safe.

Otherwise we're risking to have code clients that simply ignore `sendDone` and keep calling `CloseSend` directly on the `stream`.

pkg/kds/v2/client/stream.go
```diff
@@ -99,6 +126,16 @@ func (s *stream) sendLoop() {
 	}
 }
 
+// CloseSend signals the send goroutine to stop and waits for it to call
+// CloseSend on the underlying gRPC stream. This is thread-safe because
+// the underlying CloseSend is called within the send goroutine, never
+// concurrently with Send.
+func (s *stream) CloseSend() error {
+	s.cancel(nil)
```
lobkovilya: Calling `s.cancel(nil)` is going to stop `recvLoop` even though it's not a desired behaviour as `CloseSend` only half-closes the connection. You probably need a dedicated channel to stop the `sendLoop`

pkg/kds/v2/client/stream.go
```diff
@@ -77,15 +92,27 @@ func NewDeltaKDSStream(
...
+		stream.sendLoop()
+	}()
 	go stream.recvLoop()
 
 	return stream
 }
 
 func (s *stream) sendLoop() {
+	defer func() {
+		type closeSender interface {
```
lobkovilya: You can avoid this by adding `CloseSend() error` to the `KDSSyncServiceStream` interface.

lobkovilya (approved):

PR #16346, opened 2026-04-22
fix(helm): chart-level zone proxy image

slonka:
## Motivation

When a user overrides `meshes:` to enable a zone proxy (e.g. `--set 'meshes[0].ingress.enabled=true'`), Helm replaces the entire list and drops the chart's `values.yaml` defaults for the pause container image. Rendering then fails with:

```
Error: kuma/templates/mesh-zoneproxy-deployment.yaml:88:24
  executing "kuma/templates/mesh-zoneproxy-deployment.yaml" at <$cfg.image.registry>:
    nil pointer evaluating interface {}.registry
```

Repro:

```sh
helm template test deployments/charts/kuma \
  --set controlPlane.mode=zone \
  --set controlPlane.zone=zone-1 \
  --set 'meshes[0].name=default' \
  --set 'meshes[0].ingress.enabled=true'
```

## Implementation information

- Added a chart-level `zoneProxyImage` (registry/repository/tag)
- New template helper `kuma.mesh.zoneproxy.image`

## Supporting documentation

xrel: https://github.com/Kong/kong-mesh/issues/9409

### DIFF
--- a/app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
+++ b/app/kumactl/cmd/install/testdata/install-control-plane.dump-values.yaml
@@ -851,7 +851,8 @@ meshes:
       enabled: false
       replicas: 1
       # -- Per-mesh override for the pause container image. Falls back to
-      # .Values.zoneProxyImage when unset.
+      # .Values.zoneProxyImage when unset. Partial overrides inherit the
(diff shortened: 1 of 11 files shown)


lobkovilya (approved):
LGTM, thanks for fixing!

PR #16353, opened 2026-04-22
fix(k8s): honor ignoredServiceSelectorLabels

bartsmykla:
> [!WARNING]
> not to be reviewed yet

## Motivation

The `runtime.kubernetes.injector.ignoredServiceSelectorLabels` config lets operators strip labels from Service selectors when matching against Pods. This is needed for blue-green deployments using ArgoCD/Argo Rollouts where Services have `rollouts-pod-template-hash` selectors that Pods don't carry.

The config was only honored in `MatchServiceThatSelectsPod` but not in two other codepaths:

- `inboundForService` used raw selector, causing inbounds to flip to Ignored state
- `setFromClusterIPSvc` used raw selector, causing MeshService resources to become Unavailable

This cascade (Ignored inbounds -> VIP removal -> client listeners gone) caused 503 errors from MeshPassthrough.

## Implementation information

- Extracted label stripping into reusable `StripIgnoredSelectorLabels` helper in `util.go`
- Refactored `MatchServiceThatSelectsPod` to use the helper (behavior unchanged)
- Added `IgnoredServiceSelectorLabels` field to `InboundConverter` struct and applied helper in `inboundForService`
- Added `IgnoredServiceSelectorLabels` field to `MeshServiceReconciler` struct and applied helper in `setFromClusterIPSvc` (both branches)
- Wired config from `rt.Config().Runtime.Kubernetes.Injector.IgnoredServiceSelectorLabels` to both structs in `plugin.go`
- Added unit tests for helper, inbound converter, and meshservice controller

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
+++ b/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
@@ -43,10 +43,11 @@ func podReady(pod *kube_core.Pod, container *kube_core.Container) bool {
 }
 
 type InboundConverter struct {
-	NameExtractor       NameExtractor
-	NodeGetter          kube_client.Reader
-	NodeLabelsToCopy    []string
-	InboundTagsDisabled bool
+	NameExtractor                NameExtractor
+	NodeGetter                   kube_client.Reader
+	NodeLabelsToCopy             []string
+	InboundTagsDisabled          bool
+	IgnoredServiceSelectorLabels []string
 }
 
 func (ic *InboundConverter) tagsOrEmpty(tagsFn func() map[string]string) map[string]string {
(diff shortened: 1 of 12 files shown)


bartsmykla:
After investigating git history, found that current behavior is intentional per MADR-032.

The MADR explicitly states:
> This setting would cause selecting more services for inbounds. However, not fully matched services would create "Ignored" listener.

Design intent was two-stage:
1. Strip ignored labels for discovery (find candidate services)
2. Use full selector for eligibility (not fully matched = Ignored)

The config comment at `pkg/config/plugins/runtime/k8s/config.go:261-264` accurately documents this design, not a bug.

My patch would change designed behavior. Closing this PR.

If expanding semantics to "strip ignored labels everywhere" is desired, that should be a separate feature request with explicit contract change and updated MADR.

PR #16392, opened 2026-04-24
feat(xds): expose delta xDS via Helm and fix k8s injection

### LINKED ISSUE #16405: Add deltaXds to Helm chart
Currently `deltaXds` can only be configured via `kuma-cp.yaml` or env var `KUMA_EXPERIMENTAL_DELTA_XDS`, but there's no corresponding Helm value under `experimental:` section.

The config exists at:
- `pkg/config/app/kuma-cp/config.go`
- `pkg/config/app/kuma-cp/kuma-cp.defaults.yaml`

Should be added alongside other experimental flags like `sidecarContainers` and `inboundTagsDisabled`.

Found during Test Friday 2026-04-24.

lukidzi:
## Motivation

Delta xDS (DELTA_GRPC transport) was already wired into the Go config and xDS bootstrap code but had no way to enable it through Helm. The only workaround was patching `controlPlane.config` manually. This adds `experimental.deltaXds` to the chart so the flag is installable the normal way. Also, it was missing injection on k8s level.

## Implementation information

Added `experimental.deltaXds (default: false)` to the Helm values and wired `KUMA_EXPERIMENTAL_DELTA_XDS` into the control plane deployment. DataplaneProxyFactory and KumaInjector now accept the flag; when it's on, every injected sidecar gets `KUMA_DATAPLANE_RUNTIME_ENVOY_XDS_TRANSPORT_PROTOCOL_VARIANT=DELTA_GRPC`.

Fix: https://github.com/kumahq/kuma/issues/16405

### DIFF
--- a/pkg/plugins/runtime/k8s/webhooks/injector/injector_test.go
+++ b/pkg/plugins/runtime/k8s/webhooks/injector/injector_test.go
@@ -92,7 +92,7 @@ spec:
 				var cfg conf.Injector
 				Expect(config.Load(filepath.Join("testdata", given.cfgFile), &cfg)).To(Succeed())
 				cfg.CaCertFile = caCertPath
-				injector, err := inject.New(cfg, "http://kuma-control-plane.kuma-system:5681", k8sClient, sidecarsEnabled, k8s.NewSimpleConverter(), 9901, 9902, false, systemNamespace, nil)
+				injector, err := inject.New(cfg, "http://kuma-control-plane.kuma-system:5681", k8sClient, sidecarsEnabled, k8s.NewSimpleConverter(), 9901, 9902, false, systemNamespace, nil, false)
 				Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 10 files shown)


slonka:
@lukidzi when you create a PR can you xrel that to the issue? I did not spend much time on https://github.com/kumahq/kuma/pull/16404 but would make it 0 time if I knew this was already addressed

pkg/plugins/runtime/k8s/webhooks/injector/injector_test.go
```diff
@@ -1064,4 +1064,119 @@ spec:
 			cfgFile: "inject.config.yaml",
 		}),
 	)
+
```
slonka, line 1067: nitpick: these 3 tests could be converted to a DescribeTable to reduce duplication, but fine as-is for a feature PR.

slonka (approved):

PR #16414, opened 2026-04-27
fix(api-server): harden localhost admin auth

bartsmykla:
## Motivation

Localhost admin auth trusted browser-originated loopback requests. CORS was also open by default.

## Implementation information

- keep localhost admin enabled by default while restricting it to direct loopback
- make CORS opt-in by default
- require loopback `RemoteAddr` and `Host` before granting localhost admin
- deny proxy-hop headers, cross-site fetch metadata, and non-localhost `Origin` headers
- keep the API test harness explicit about localhost admin

### DIFF
--- a/pkg/api-server/auth_test.go
+++ b/pkg/api-server/auth_test.go
@@ -39,6 +39,7 @@ var _ = Describe("Auth test", func() {
 			cfg.HTTPS.TlsCertFile = certPath
 			cfg.HTTPS.TlsKeyFile = keyPath
 			cfg.Authn.Type = certs.PluginName
+			cfg.Authn.LocalhostIsAdmin = true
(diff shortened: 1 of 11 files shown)


pkg/api-server/auth_test.go
```diff
@@ -120,6 +121,49 @@ var _ = Describe("Auth test", func() {
 		Expect(rr.Body.Bytes()).To(matchers.MatchGoldenJSON(path.Join("testdata", "auth-admin-https-bad-creds.golden.json")))
 	})
 
+	It("should block admin access when Origin is a cross-origin domain (PoC for H1 #3631894)", func() {
```
lukidzi, line 124: What is `(PoC for H1 #3631894)`?

PR #16416, opened 2026-04-27
fix(api-server): harden localhost admin auth

bartsmykla:
## Motivation

Localhost admin auth trusted browser-originated loopback requests. CORS was also open by default.

## Implementation information

- keep localhost admin enabled by default while restricting it to direct loopback
- make CORS opt-in by default
- require loopback `RemoteAddr` and `Host` before granting localhost admin
- deny proxy-hop headers, cross-site fetch metadata, and non-localhost `Origin` headers
- keep the API test harness explicit about localhost admin


pkg/api-server/authn/localhost.go
```diff
@@ -2,25 +2,89 @@ package authn
...
+	if hasProxyHeaders(r.Header) {
+		return false
+	}
+	if !isLoopbackHost(hostWithoutPort(r.Host)) {
+		return false
+	}
+	if r.Header.Get("Sec-Fetch-Site") == "cross-site" {
+		return false
+	}
+	return isSameOriginLoopback(r.Header.Get("Origin"), r.Host)
```
lukidzi, line 51: Is there an issue in case where the browser/curl omits default port? Like `http://localhost` (80), `https://localhost` (443)
```
  - Origin: http://localhost → u.Host = "localhost"
  - r.Host = "localhost:80"
  - EqualFold("localhost", "localhost:80") → false → admin blocked
```

bartsmykla: should be fixed now

lukidzi (approved):

PR #16429, opened 2026-04-28
feat(bootstrap): configurable xDS GoogleGrpc max_receive_message_length

### LINKED ISSUE #16355: bug(xds): EDS Timeout post CDS syncup
### Kuma Version

2.11.12

### Describe the bug

We are deploying Kong delegated gateways with sidecar injection, and the gateway is connected to nearly 500 services/clusters. We are observing that DPP certificate updates via xDS are not completing successfully, which causes the system to enter an infinite loop.

Kuma version - 2.11.12
Kong version - 3.9.4

### More about the issue

1. DPP is able to connect to the ZoneCP
2. The DPP is able to fetch CDS, which takes around 9 seconds.
3. When the CP sends EDS, the DPP fails to acknowledge it because the connection is dropped before that due to a timeout.
4. The same issue occurs with LDS.
5. Since neither EDS nor LDS is acknowledged, the CP does not receive any certificate. As a result, the DPP is unable to synchronize with ZoneCP, and the sidecar containers enter an infinite loop.

This is a similar issue which is faced in https://github.com/kumahq/kuma/issues/14977

### What we have tried

We have tried to increase the refresh interval , added incremental XDS feature for fine tuning , Added some patch. But all have not helped to mitigate this issue.

**We are suspecting that there is an EDS `initialFetchTimeout` coming into the picture which causing the DPP to disconnect from zoneCP**

### Some Screenshot to prove the point

Image

Image

Image

### To Reproduce

1. Have a service with limited resource
2. the service needs to connect 200+ other service , MeshTrafficpermission & route should be defined for each service.
3. CDS & EDS connection time should be exceeding `15 sec`

### Expected behavior

_No response_

_No response_

andrey-dubnik:
## Motivation

Gateway data planes with hundreds of listeners (~500 listeners) (~640 KiB Listener payload
plus concurrent CDS/EDS/SDS in the same ADS stream) fail to reach LIVE on
cold start. The kuma-dp xDS gRPC client (GoogleGrpc transport) aborts with
`CANCELLED` ~360 ms after LDS dispatch, and Envoy never ACKs LDS.

Root cause: gRPC C-Core's default of 4 MiB for `grpc.max_receive_message_length`.
In the GoogleGrpc transport this argument also sizes the per-stream HTTP/2
receive flow-control window. The 4 MiB window depletes mid-push, the channel
stalls, and libgrpc-cpp aborts the stream.

Per-arg ablation confirmed this is the single load-bearing change:
`bdp_probe=0` alone fails, `write_buffer_size=4 MiB` alone fails,
`max_receive_message_length=16 MiB` alone succeeds.

## Implementation information

Raises the channel arg default to 16 MiB and exposes it as a configurable
`BootstrapParams` field so operators with even larger snapshots can tune
it without rebuilding:

bootstrapServer.params.xdsGrpcMaxReceiveMessageBytes  (default 16777216)
KUMA_BOOTSTRAP_SERVER_PARAMS_XDS_GRPC_MAX_RECEIVE_MESSAGE_BYTES

No behavior change for DPs whose snapshots fit under 4 MiB.

## Supporting documentation

N/A

## Fixes 

#16355


lahabana:
@andrey-dubnik are you leveraging autoreachable services or backends? This number of listeners usually indicate that reachable service should be used which trims significantly the config size.

andrey-dubnik:
> @andrey-dubnik are you leveraging autoreachable services or backends? This number of listeners usually indicate that reachable service should be used which trims significantly the config size.

Yes, we do have auto-reachable turned on. Unfortunately this is the real count of the services as source service is the API gateway which routes to multiple services across multiple k8s clusters and serves as a central focal point for the ingress into the network, we have a few policies per each service gateway needs to access (both traffic route and permission). We suddenly hit a wall with roughly 500 listeners and our gateway's DP went into a crash loop till we deleted the overhead (new services added recently) and blocked new services from being added to the API gateway. The way we solved the problem is by increasing the message length to 16 Mib (change in this PR makes it configurable).

andrey-dubnik:
@lahabana does the change makes sense in a given context? we plan to double a number of services within few months as well which are connected via the API gateway.

lahabana:
Ok makes more sense to me indeed. I won't review this but @bartsmykla and @lobkovilya will get to it.

andrey-dubnik:
any chance a review/merge can go ahead? the issue is real and there is no solution without a permanent fix

bartsmykla (dismissed):

lukidzi:
The change was merged. We decided to keep default to 4MiB, but it's configurable

PR #16430, opened 2026-04-28
docs(MADR): standardize zone proxy listener labels

### LINKED ISSUE #16431: standardize computed labels for mesh-scoped zone proxy listeners
Create a small MADR and follow-up implementation work to standardise auto-computed Dataplane labels for mesh-scoped zone proxy listeners.

These labels should be set based on listener presence on the Dataplane and used for coarse-grained selection.

Decision drivers:

- users applying policies via  need to match all zone ingress, zone egress, or combined zone proxies
- Inspect API / GUI need to query Dataplanes exposing zone ingress, zone egress, or both without reconstructing listener presence from the full Dataplane spec

lobkovilya:
## Motivation

We need a standard coarse-grained way to select Dataplanes exposing zone ingress and zone egress listeners.

This is needed both for policy targeting via `spec.targetRef` and for Inspect API / GUI queries that need ingress-only, egress-only, or combined zone proxies.

## Implementation information

Add a short MADR that standardizes two auto-computed Dataplane labels:

- `kuma.io/listener-zoneingress: enabled`
- `kuma.io/listener-zoneegress: enabled`

Fix #16431

> Changelog: skip

### DIFF
--- a/docs/madr/decisions/102-zone-proxy-listener-computed-labels.md
+++ b/docs/madr/decisions/102-zone-proxy-listener-computed-labels.md
@@ -1,0 +1,71 @@
+# Standardize Computed Labels for Mesh-Scoped Zone Proxy Listeners
+
+* Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/15567
+
+## Context and Problem Statement
+
+With mesh-scoped zone proxies, zone ingress and zone egress are listeners on a `Dataplane`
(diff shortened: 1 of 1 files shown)


lahabana (dismissed):
Do we need the dataplane- prefix? Seems unecessary.

Otherwise this looks fine

lukidzi (dismissed):

lukidzi (approved):

lahabana (approved):

PR #16447, opened 2026-04-28
test(unit): replace sleep with eventually

### LINKED ISSUE #13101: Improve KDS sync unit tests to remove 5s wait time
### Description

Right now, KDS sync tests wait **5 seconds** before checking golden files to avoid flakiness. This was added in https://github.com/kumahq/kuma/pull/13100 as a quick fix, but waiting that long seems unnecessary in most cases and slows down tests.

We should find a better way to check if sync is complete without using a fixed sleep time.

lukidzi:
## Motivation

The full sync integration test used a hard-coded `time.Sleep(5s)` to wait for KDS sync to complete before checking golden files. On slow CI runners this wasn't always enough, causing flakes where zone-to-global resources hadn't arrived yet.

## Implementation information

 Replace the fixed sleep with `Eventually` polling (up to 30s, 250ms interval). Fast runners still pass in ~1-2s, slow ones get the headroom they need.

## Supporting documentation

> Changelog: skip

Fix: https://github.com/kumahq/kuma/issues/13101

### DIFF
--- a/pkg/kds/context/full_sync_test.go
+++ b/pkg/kds/context/full_sync_test.go
@@ -33,6 +33,11 @@ var _ = Describe("Full sync tests", func() {
 		zones := make(map[string]store.ResourceStore)
 		wg := sync.WaitGroup{}
 		done := make(chan struct{})
+		closeOnce := sync.Once{}
+		DeferCleanup(func() {
+			closeOnce.Do(func() { close(done) })
+			wg.Wait()
+		})
 
 		for _, file := range files {
 			if before, ok := strings.CutSuffix(file.Name(), ".input.yaml"); ok {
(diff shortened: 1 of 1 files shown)


pkg/kds/context/full_sync_test.go
```diff
@@ -95,16 +100,16 @@ var _ = Describe("Full sync tests", func() {
 			})
 		}
 
-		// Wait for some time to ensure sync was complete
-		time.Sleep(time.Second * 5)
-		close(done)
-		wg.Wait()
-
-		// Compare golden files
+		// Wait until all stores converge to their expected golden state.
```
lobkovilya, line 103: When rerunning all tests with `UPDATE_GOLDEN_FILES` the lack of sleep is going to produce changes. Maybe we can add conditional sleep, only when this env is set.

bartsmykla (approved):

PR #16450, opened 2026-04-28
chore(actions): retry mise installs on transient network errors

lukidzi:
## Motivation

Noticed some 502 on installation mise and failure

## Implementation information

This wraps `mise install` in `nick-fields/retry@v4.0.0` (3 attempts, 15s wait, 10m timeout) across all 12 CI workflow files. The `jdx/mise-action` built-in install is disabled (`install: "false"`) so the retry step controls execution.

## Supporting documentation

> Changelog: skip


lukidzi:
seems like retry doesn't work for 502 :(

PR #16454, opened 2026-04-29
fix(xds): allow stale stream takeover during dataplane reconnect

AyushSenapati:
## Motivation

Under high dataplane churn, Kuma can briefly observe two xDS streams for the same dataplane. In that window, the old stream may already be stale, but the control plane can still treat it as active until `OnStreamClosed` is processed.

At scale, CP under load, I have seen this issue lasts for long. When DPP fails readiness K8s restarts the pod.

## Implementation information

This change updates dataplane xDS stream ownership handling so a reconnecting stream can take over when the existing owner is already stale.

If the existing owner stream is still live, the new stream is rejected as before. If its context is already cancelled, the new stream is allowed to become the owner immediately. During takeover, stale stream state is removed so the old stream's later `OnStreamClosed` callback does not tear down state belonging to the replacement stream.

The existing connect-phase guard remains in place. If another stream arrives while registration for the same dataplane is still in progress, it is rejected with `RESOURCE_EXHAUSTED` so Envoy backs off and retries cleanly.

The sync tracker was also updated to stop the previous watchdog before replacing it on reconnect.

Added coverage for:
- concurrent stale-owner takeover races
- reconnect metric increments
- watchdog lifecycle during stale-owner replacement

## Supporting documentation

`existing watchdog for a Dataplane stopped` is getting logged after watchdog stopped message, means another stream connected before CP could cleanup existing connection:
image

The `xds_stream_stale_owner_takeovers_total` metrics shows the number of times a stale owner was evicted and replaced:
image

### DIFF
--- a/pkg/xds/server/callbacks/dataplane_callbacks.go
+++ b/pkg/xds/server/callbacks/dataplane_callbacks.go
@@ -5,10 +5,13 @@ import (
 	"sync"
 
 	"github.com/pkg/errors"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	core_model "github.com/kumahq/kuma/v2/pkg/core/resources/model"
 	core_xds "github.com/kumahq/kuma/v2/pkg/core/xds"
 	util_xds "github.com/kumahq/kuma/v2/pkg/util/xds"
+	xds_metrics "github.com/kumahq/kuma/v2/pkg/xds/metrics"
 )
 
 // DataplaneCallbacks are XDS callbacks that keep the context of Kuma Dataplane.
@@ -28,6 +31,7 @@ type DataplaneCallbacks interface {
 
 type xdsCallbacks struct {
 	callbacks DataplaneCallbacks
+	metrics   *xds_metrics.Metrics
 	util_xds.NoopCallbacks
 
 	sync.RWMutex
@@ -35,14 +39,22 @@ type xdsCallbacks struct {
 	dpDeltaStreams map[core_xds.StreamID]dpStream
 	// we don't need separate map for stream because we use here resource key
 	activeStreams map[core_model.ResourceKey]core_xds.StreamID
+	// connectingStream tracks streams that are currently in the process of registering
+	// (i.e. OnProxyConnected has been called but has not yet returned). This prevents
+	// a concurrent stream for the same dpKey from racing past the activeStreams guard
+	// before registration completes, which would cause OnProxyConnected to be invoked
+	// twice for the same dataplane.
+	connectingStream map[core_model.ResourceKey]core_xds.StreamID
 }
 
-func DataplaneCallbacksToXdsCallbacks(callbacks DataplaneCallbacks) util_xds.MultiXDSCallbacks {
+func DataplaneCallbacksToXdsCallbacks(callbacks DataplaneCallbacks, metrics *xds_metrics.Metrics) util_xds.MultiXDSCallbacks {
 	return &xdsCallbacks{
-		callbacks:      callbacks,
-		dpStreams:      map[core_xds.StreamID]dpStream{},
-		dpDeltaStreams: map[core_xds.StreamID]dpStream{},
-		activeStreams:  map[core_model.ResourceKey]core_xds.StreamID{},
+		callbacks:        callbacks,
+		metrics:          metrics,
+		dpStreams:        map[core_xds.StreamID]dpStream{},
+		dpDeltaStreams:   map[core_xds.StreamID]dpStream{},
+		activeStreams:    map[core_model.ResourceKey]core_xds.StreamID{},
+		connectingStream: map[core_model.ResourceKey]core_xds.StreamID{},
 	}
 }
 
@@ -99,7 +111,16 @@ func (d *xdsCallbacks) onStreamClosed(streamID core_xds.StreamID, getDpStream fu
 
 	d.Lock()
 	if streamDpKey != nil {
-		delete(d.activeStreams, *streamDpKey)
+		// Guard by stream ID: a stale-owner takeover may have already replaced
+		// this stream as the active owner. Only remove the entry if we are still
+		// the current owner, so the new owner's registration is not clobbered.
+		if d.activeStreams[*streamDpKey] == streamID {
+			delete(d.activeStreams, *streamDpKey)
+		}
+		// Safety cleanup: remove any in-progress connecting entry for this stream.
+		if d.connectingStream[*streamDpKey] == streamID {
+			delete(d.connectingStream, *streamDpKey)
+		}
 	}
 	delete(getDpStream(), streamID)
 	d.Unlock()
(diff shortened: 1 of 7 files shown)


pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -143,22 +164,56 @@ func (d *xdsCallbacks) onStreamRequest(streamID core_xds.StreamID, request util_
...
+			d.Unlock()
+			// we don't allow more than one active stream from a data plane as there can be race conditions
+			return errors.New("there is already an active stream from this node, try again later")
+		}
+		// Owner's context is cancelled — it is stale. Allow the new stream to take
+		// over. The stale owner's OnStreamClosed will still fire and call
+		// OnProxyDisconnected only if stream state still exists. Evict stale stream
+		// state here so OnStreamClosed for the stale owner does not disconnect the
+		// replacement stream's watchdog lifecycle.
+		delete(getDpStream(), ownerStreamID)
```
Automaat, line 195: **blocking:** Stale stream takeover can cancel the *new* owner's watchdog, leaving a live dataplane with no watchdog. That is the exact reconnect-readiness failure this PR sets out to fix. `onStreamClosed` snapshots the stale stream's dpKey under `RLock`, releases the lock, then calls `OnProxyDisconnected` outside it. The takeover deletes the stale stream's `dpStream` entry to suppress that disconnect, but the suppression only works if the delete wins the race against the close goroutine's snapshot. When the stale stream's close goroutine reads `dp != nil` *before* this delete, it still calls `OnProxyDisconnected`. Because the sync tracker's `OnProxyDisconnected` keys solely on `dpKey` with no stream-ID guard, it cancels `watchdogs[dpKey]`, which is now the new owner's watchdog. Stream 2 ends up as the active owner with its watchdog gone, so its config is never reconciled. We reproduced this empirically: 4 of 5000 concurrent iterations started the new owner's watchdog and then saw it go inactive. The fix belongs in `dataplane_sync_tracker.go`: store the owning stream ID in `watchdogState` and cancel/delete only when the disconnecting stream ID matches, mirroring the stream-ID guard this PR already added for `activeStreams` cleanup. The existing tests run takeover and close sequentially, so they pass even with this bug present.

pkg/xds/server/callbacks/dataplane_sync_tracker.go
```diff
@@ -43,6 +43,11 @@ func (t *dataplaneSyncTracker) OnProxyConnected(streamID core_xds.StreamID, dpKe
 	// We use OnProxyConnected because there should be only one watchdog for given dataplane.
 	t.Lock()
 	defer t.Unlock()
+	if dpWatchdog := t.watchdogs[dpKey]; dpWatchdog != nil {
+		dpWatchdog.cancelFunc()
+		<-dpWatchdog.stopped
```
Automaat, line 48: **issue:** `OnProxyConnected` holds the global tracker write lock while waiting for the old watchdog to stop. `t.Lock()` is held (deferred unlock) across `<-dpWatchdog.stopped`, and `cancelFunc()` cannot interrupt an in-flight watchdog tick, since the watchdog only checks `ctx.Done()` after a tick returns, and a tick is a full snapshot reconcile. So one slow reconcile blocks every other dataplane's connect *and* disconnect on this CP, during exactly the churn this PR targets. `OnProxyDisconnected` already shows the correct pattern: read under `RLock`, release, wait on `stopped` without the lock, then re-take the lock only to delete. Stop the old watchdog the same way here.

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -143,22 +164,56 @@ func (d *xdsCallbacks) onStreamRequest(streamID core_xds.StreamID, request util_
...
+		// Under high dataplane churn a narrow window exists between gRPC GOAWAY/FIN
+		// (which cancels the stream context) and the CP processing OnStreamClosed
+		// (which removes the activeStreams entry). If we hit that window a new stream
+		// would otherwise be rejected even though the old one is truly gone.
+		ownerCtx := getDpStream()[ownerStreamID].ctx
+		if ownerCtx == nil || ownerCtx.Err() == nil {
+			// Owner is still alive — reject the incoming stream to avoid races.
+			d.Unlock()
+			// we don't allow more than one active stream from a data plane as there can be race conditions
+			return errors.New("there is already an active stream from this node, try again later")
```
Automaat, line 188: **issue:** The live-owner rejection path is invisible: no metric, no log. The two takeover-adjacent paths are metered, but the case where two genuinely-live streams fight, a real reconnect-loop pathology, returns a bare error with no signal. An operator can't tell healthy backoff from a node wedged in repeated rejection. Add a counter (same `mesh`/`proxy_type` labels) and a rate-limited log here.

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -28,21 +31,30 @@ type DataplaneCallbacks interface {
...
 	dpStreams      map[core_xds.StreamID]dpStream
 	dpDeltaStreams map[core_xds.StreamID]dpStream
 	// we don't need separate map for stream because we use here resource key
 	activeStreams map[core_model.ResourceKey]core_xds.StreamID
+	// connectingStream tracks streams that are currently in the process of registering
+	// (i.e. OnProxyConnected has been called but has not yet returned). This prevents
+	// a concurrent stream for the same dpKey from racing past the activeStreams guard
+	// before registration completes, which would cause OnProxyConnected to be invoked
+	// twice for the same dataplane.
+	connectingStream map[core_model.ResourceKey]core_xds.StreamID
```
Automaat, line 47: **suggestion:** Ownership is now smeared across three structures: `activeStreams` (committed owner), `connectingStream` (in-flight), and per-stream `dpStreams[id].ctx` used as the staleness oracle. Staleness is inferred indirectly (`getDpStream()[ownerStreamID].ctx.Err() != nil`), so a reader has to hold three maps in their head to reason about one decision. A single owner record `{streamID, ctx, state}` keyed by `dpKey` would centralize the state machine and remove the implicit coupling. It's also the structural reason the watchdog race above is non-obvious: ownership identity is first-class in this wrapper but absent in the layer that owns the cancelable watchdog.

pkg/xds/server/v3/components.go
```diff
@@ -42,10 +42,12 @@ func RegisterXDS(
 
 	authenticator := rt.XDS().PerProxyTypeAuthenticator()
 	authCallbacks := auth.NewCallbacks(rt.ReadOnlyResourceManager(), authenticator, auth.DPNotFoundRetry{}) // no need to retry on DP Not Found because we are creating DP in DataplaneLifecycle callback
-	workloadLabelValidator := xds_callbacks.DataplaneCallbacksToXdsCallbacks(xds_callbacks.NewWorkloadLabelValidator(rt.ReadOnlyResourceManager(), rt.Config().Environment))
+	workloadLabelValidator := xds_callbacks.DataplaneCallbacksToXdsCallbacks(xds_callbacks.NewWorkloadLabelValidator(rt.ReadOnlyResourceManager(), rt.Config().Environment), nil)
```
Automaat, line 45: **suggestion:** `metrics` is passed positionally, but the validator call site passes `nil` and every `increment*` helper then nil-guards. The next callback added here will also pass `nil` and silently lose metrics, with no compile-time signal. A functional option (`WithMetrics(m)`) or a default no-op sink would make "unmetered" an explicit choice rather than an easy-to-miss positional argument.

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -143,22 +164,56 @@ func (d *xdsCallbacks) onStreamRequest(streamID core_xds.StreamID, request util_
 		return nil
 	}
 
+	// If another stream is already registering for this dpKey, return RESOURCE_EXHAUSTED
+	// so Envoy applies backoff before retrying, rather than hammering the CP.
+	if connectingStreamID, registering := d.connectingStream[dpKey]; registering && connectingStreamID != streamID {
+		d.incrementInProgressRetries(dpKey.Mesh, string(metadata.GetProxyType()))
+		d.Unlock()
+		return status.Error(codes.ResourceExhausted, "registration in progress for this node, try again later")
```
Automaat, line 172: **suggestion:** Two rejection paths on the same stream now return different error types. The in-progress guard returns gRPC `codes.ResourceExhausted` (triggering Envoy backoff), while the live-owner path returns a plain `errors.New` that surfaces as `codes.Unknown`. Either make both gRPC status codes so reconnect-backoff behavior is uniform, or add a comment explaining why the arguably-more-common live-owner rejection stays a plain error. Worth a changelog note that the rejection code changed for existing DPPs.

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -143,22 +164,56 @@ func (d *xdsCallbacks) onStreamRequest(streamID core_xds.StreamID, request util_
...
+	}
+
 	var streamInfo dpStream
-	_, alreadyConnected := d.activeStreams[dpKey]
-	if !alreadyConnected {
-		streamInfo = getDpStream()[streamID]
-		streamInfo.dp = &dpKey
-		getDpStream()[streamID] = streamInfo
+	ownerStreamID, alreadyConnected := d.activeStreams[dpKey]
+	if alreadyConnected {
```
Automaat, line 177: **suggestion:** This is a concurrency change on the hottest connect path. Its worst case (the watchdog race, the lock stall) is CP-wide, and there's no flag to fall back to the old reject-only behavior. The right move is to fix the race rather than gate around it, but given kuma's `ExperimentalConfig` convention, a guard is cheap insurance for a change that fires on every reconnect under load.

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -99,7 +111,16 @@ func (d *xdsCallbacks) onStreamClosed(streamID core_xds.StreamID, getDpStream fu
 
 	d.Lock()
 	if streamDpKey != nil {
-		delete(d.activeStreams, *streamDpKey)
+		// Guard by stream ID: a stale-owner takeover may have already replaced
+		// this stream as the active owner. Only remove the entry if we are still
+		// the current owner, so the new owner's registration is not clobbered.
+		if d.activeStreams[*streamDpKey] == streamID {
```
Automaat, line 117: **praise:** The stream-ID-guarded cleanup in `onStreamClosed` is the right defensive move. Guarding both `activeStreams` and `connectingStream` deletes by `== streamID` stops a stale owner's late close from clobbering the replacement's registration.

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -143,22 +164,56 @@ func (d *xdsCallbacks) onStreamRequest(streamID core_xds.StreamID, request util_
...
+		// state here so OnStreamClosed for the stale owner does not disconnect the
+		// replacement stream's watchdog lifecycle.
+		delete(getDpStream(), ownerStreamID)
+		d.incrementStaleTakeovers(dpKey.Mesh, string(metadata.GetProxyType()))
+	}
 
+	streamInfo = getDpStream()[streamID]
+	streamInfo.dp = &dpKey
+	getDpStream()[streamID] = streamInfo
+	d.connectingStream[dpKey] = streamID
```
Automaat, line 202: **praise:** Setting `connectingStream` before releasing the lock to run the heavy `OnProxyConnected`, then committing to `activeStreams` only on success, cleanly closes the original double-connect race. The "only one winner in concurrent stale takeover" test exercises exactly the path that matters.

Automaat (changes requested):

PR #16463, opened 2026-04-30
chore(build): retry ebpf download

lukidzi:
## Motivation

Notice failure of build because of network failure

## Implementation information

We retry for envoy and coreDNS lets use the same mechanism here

> Changelog: skip

### DIFF
--- a/mk/ebpf.mk
+++ b/mk/ebpf.mk
@@ -30,7 +30,7 @@ COMPILE_IN_PATH ?= ./pkg/transparentproxy/ebpf/programs
 
 define make_ebpf_targets
 $(1)/mb_*: | $(1)
-	curl --progress-bar --location $(TARBALL_URL)/all-$(3).tar.gz | tar -C $$(@D) -xz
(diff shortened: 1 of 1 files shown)


mk/ebpf.mk
```diff
@@ -30,7 +30,7 @@ COMPILE_IN_PATH ?= ./pkg/transparentproxy/ebpf/programs
 
 define make_ebpf_targets
 $(1)/mb_*: | $(1)
-	curl --progress-bar --location $(TARBALL_URL)/all-$(3).tar.gz | tar -C $$(@D) -xz
+	curl --retry 3 --retry-delay 60 --retry-all-errors -s --fail --location $(TARBALL_URL)/all-$(3).tar.gz | tar -C $$(@D) -xz
```
slonka, line 33: I thought that we were getting rid of eBPF completely, right?

lukidzi: Probably in 3.0. I would prefer stable build and fix if it's possible

lahabana (approved):

PR #16483, opened 2026-05-04
test(e2e): mesh-scoped zone proxy migration

### LINKED ISSUE #15674: Design traffic migration path from global to mesh-scoped zone proxies
MADR 094 (#15460) defers the traffic migration strategy from global-scoped zone proxies to mesh-scoped zone proxies. This should cover:

- How to switch live traffic without downtime
- Whether both global and mesh-scoped zone proxies can coexist during migration
- DNS/addressing cutover
- Rollback procedures

**Context:** "Migration Path from Global Zone Proxies" section in MADR 094

slonka:
## Motivation

Adds the multizone E2E coverage for migrating cross-zone traffic from the global zone proxies to the new mesh-scoped zone proxies introduced by [MADR 094 / #15460](https://github.com/kumahq/kuma/pull/15460). The migration path itself was deferred from MADR 094 and tracked in https://github.com/kumahq/kuma/issues/15674.

## Implementation information

- Sets up two meshes (`meshproxymig1`, `meshproxymig2`) in `MeshServices: Exclusive` mode with a per-mesh
- enable cross-zone mTLS
- confirm baseline reachability
- roll out a mesh-scoped ingress + egress per mesh and asserts:
  1. all meshes remain reachable cross-zone, and
  2. the new mesh-scoped ingress receives traffic for its mesh.
- A background goroutine per mesh continuously sends cross-zone requests during the rollout and a final ` Consistently` check asserts no request error was observed - covering the no-downtime requirement of the issue.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/15674


test/e2e_env/multizone/meshproxy/migration.go
```diff
@@ -0,0 +1,393 @@
...
+		runDuringMigration(ctx, routes, func() {
+			for _, scenario := range scenarios {
+				deployMeshScopedZoneProxies(scenario.meshName)
+				assertAllMeshesReachable()
+				assertMeshScopedProxyUsed(scenario)
+			}
+		})
+	})
+
+	BeforeAll(func() {
```
lobkovilya, line 75: it probably doesn't matter, but can we please have BeforeAll before all `It(...)`?

test/e2e_env/multizone/meshproxy/migration.go
```diff
@@ -0,0 +1,393 @@
...
+					builders.Mesh().
+						WithName(scenario.meshName).
+						WithMeshServicesEnabled(mesh_proto.Mesh_MeshServices_Exclusive),
+				)).
+				Install(MeshTrafficPermissionAllowAllUniversal(scenario.meshName))
+		}
+		Expect(setup.Setup(multizone.Global)).To(Succeed())
+
+		for _, scenario := range scenarios {
+			Expect(WaitForMesh(scenario.meshName, multizone.Zones())).To(Succeed())
```
lobkovilya: do we need to check mesh in all zones including `UniZone1` and `UniZone2`?

test/e2e_env/multizone/meshproxy/migration.go
```diff
@@ -0,0 +1,393 @@
...
+	BeforeAll(func() {
+		setup := NewClusterSetup()
+		for _, scenario := range scenarios {
+			setup.
+				Install(Yaml(
+					builders.Mesh().
+						WithName(scenario.meshName).
+						WithMeshServicesEnabled(mesh_proto.Mesh_MeshServices_Exclusive),
+				)).
+				Install(MeshTrafficPermissionAllowAllUniversal(scenario.meshName))
```
lobkovilya: it's probably better to use new MTP with spiffe even though the old one works

test/e2e_env/multizone/meshproxy/migration.go
```diff
@@ -0,0 +1,393 @@
...
+	"github.com/kumahq/kuma/v2/test/framework/deployments/zoneproxy"
+	"github.com/kumahq/kuma/v2/test/framework/envoy_admin/stats"
+	"github.com/kumahq/kuma/v2/test/framework/envs/multizone"
+)
+
+func Migration() {
+	namespace := "meshproxy-migration"
+
+	type crossZoneRoute struct {
+		fromCluster      func() *K8sCluster
```
lobkovilya: why `func() *K8sCluster` and not just `*K8sCluster`?

test/e2e_env/multizone/meshproxy/migration.go
```diff
@@ -0,0 +1,393 @@
...
+		fromCluster := route.fromCluster()
+
+		response, err := client.CollectEchoResponse(
+			fromCluster,
+			route.fromClientName,
+			crossZoneAddress(route),
+			client.FromKubernetesPod(namespace, route.fromClientName),
+		)
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(response.Instance).To(Equal(route.expectedInstance))
```
lobkovilya: IIUC the point of having 2 meshes in this test is to check the request never leaves the mesh boundaries? 

If so, then checking `response.Instance` like this within `Eventually` "swallows" the error. For example, one request went from blue to red and `Eventually` simply retried it. So we never know if boundaries are breached.

You should probably use `CollectResponsesByInstance` that performs multiple requests and returns a map.

test/e2e_env/multizone/meshproxy/migration.go
```diff
@@ -0,0 +1,393 @@
...
+
+	type crossZoneRoute struct {
+		fromCluster      func() *K8sCluster
+		fromClientName   string
+		toCluster        func() *K8sCluster
+		toServiceName    string
+		expectedInstance string
+	}
+
+	type meshScenario struct {
```
lobkovilya: TBH it's pretty hard to read when scenarios are abstracted away like that. I'd drop `[]meshScenario` and just explicitly use `red` and `blue` meshes everywhere. We'll never extend this struct with more scenarios so there is no point to account for it.

Then the test cases will be like (assuming ordered):

* It("should deploy zone proxies to meshproxymig-red with no interruptions for meshproxymig-blue")
* It("should deploy zone proxies to mesh meshproxymig-blue with no interruptions for meshproxymig-red")

It'll be more verbose, but easier to read. Also more visibility when there is a failure.

Automaat (approved):

PR #16539, opened 2026-05-11
feat(mmzs): deprecate names longer than 63 chars

### LINKED ISSUE #12825: MMZS allowed to have longer names than MeshService and MeshExternalService
### Description

MeshService and MeshExternalService names are limited to 63 characters, but MeshMutliZoneService can be 253.

slonka:
## Motivation

`MeshService` and `MeshExternalService` validators already reject names longer than 63 characters, but `MeshMultiZoneService` (MMZS) historically accepts up to 253. Since the MMZS name is consumed by `HostnameGenerator` templates (e.g. `{{ .DisplayName }}.mzsvc.mesh.local`), any name longer than 63 produces an invalid RFC 1035 DNS label.

Per [lahabana's comment on the issue](https://github.com/kumahq/kuma/issues/12825#issuecomment-4326698068), 2.14 should warn about deprecation; 3.0 will enforce the hard limit under https://github.com/kumahq/kuma/issues/13047.

## Implementation information

- Extend the existing `MeshMultiZoneServiceResource.Deprecations()` method

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/12825


pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/deprecated.go
```diff
@@ -20,5 +20,11 @@ func (t *MeshMultiZoneServiceResource) Deprecations() []string {
 		deprecations = append(deprecations, nameDeprecationMsg)
 	}
 
+	if len(name) > 63 {
```
slonka: do not merge, seems like this check is already in https://github.com/kumahq/kuma/pull/16539/changes#diff-accfaa1d4c92e610c71eab11f0c9e9f6c6b4173d13ac7b8d45e93788ec6128aeR18 - probably close this and the related issue

bartsmykla (approved):

PR #16541, opened 2026-05-11
fix(dp-server): bound shutdown, propagate appCtx

bartsmykla:
## Motivation

Remote control planes were crashing on SIGTERM under load. controller-runtime's manager logs:

> failed waiting for all runnables to end within grace period of 30s: context deadline exceeded

There are two compounding root causes in the dp-server stack:

1. `DpServer.Start` called `http.Server.Shutdown(context.Background())` - unbounded.
2. SoTW, Delta, and DataplaneSyncTracker watchdog contexts were all rooted in `context.Background()`, so xDS handlers never observed app shutdown.

`http.Shutdown` waited forever for xDS streams whose handlers never exited, controller-runtime's 30s grace expired, and the pod exited non-zero.

## Implementation information

`DpServer.Start` now wraps the shutdown:

- bounded with `dp_server.gracefulShutdownTimeout` (default 10s, env `KUMA_DP_SERVER_GRACEFUL_SHUTDOWN_TIMEOUT`)
- `http.Server.Shutdown(ctx)` first
- `grpcServer.Stop()` second as a backstop for handlers that ignored the ctx cancel

Why `Stop` and not `GracefulStop`: gRPC is mounted via `ServeHTTP` (see `pkg/dp-server/server/server.go` `handle`), so its transport is `serverHandlerTransport`. [`GracefulStop`](https://github.com/grpc/grpc-go/blob/v1.80.0/server.go#L1948) routes through [`drainAllServerTransportsLocked`](https://github.com/grpc/grpc-go/blob/v1.80.0/server.go#L2006-L2014), which calls `st.Drain("graceful_stop")` on every transport. On `serverHandlerTransport`, [`Drain` is a hard panic](https://github.com/grpc/grpc-go/blob/v1.80.0/internal/transport/handler_server.go#L482-L484) (`panic("Drain() is not implemented")`). So `GracefulStop` would crash the process during shutdown. `Stop` skips `Drain` and closes the transports directly, which is what we want as a backstop.

`DpServer` now implements `GracefulComponent` so the component manager waits for `Start` to return before exiting.

Contexts for SoTW, Delta, the dataplane sync tracker watchdog, and the HDS server all derive from `rt.AppContext()` now, so the kuma-cp signal handler propagates straight through to the handler loops.

`NewDpServer` returns `(*DpServer, error)` since it registers metrics that can fail at boot. Matches the pattern used by `pkg/metrics/store/`, `pkg/plugins/resources/postgres/pgx_store.go`, etc.

### New metrics

- `dp_server_force_stop_total{reason}` counter. Reasons are fixed at compile time: `shutdown_timeout`, `http_error`.
- `dp_server_shutdown_duration_seconds` histogram with buckets sized around the 10s default and 30s outer bound.


pkg/dp-server/server/server.go
```diff
@@ -128,7 +186,23 @@ func (d *DpServer) Start(stop <-chan struct{}) error {
...
+		case err == nil:
+			log.Info("terminated normally")
+		case std_errors.Is(err, context.DeadlineExceeded):
+			d.forceStopCount.WithLabelValues(forceStopReasonTimeout).Inc()
+			log.Error(err, "dp-server force-stopped", "reason", forceStopReasonTimeout, "timeout", timeout)
+		default:
+			d.forceStopCount.WithLabelValues(forceStopReasonError).Inc()
+			log.Error(err, "dp-server force-stopped", "reason", forceStopReasonError)
+		}
+		return nil
```
Automaat: **issue:** The shutdown branch returns `nil` unconditionally; pre-PR it returned the error from `server.Shutdown`. This is intentional for the timeout case (the whole point of the PR is to stop the pod from exiting non-zero), but it also swallows the `default` branch (`forceStopReasonError`) where `Shutdown` returned a non-deadline error. The manager's `errCh` no longer sees those failures, so a genuinely broken shutdown is invisible to the caller. Return the error only for the `default` branch and keep swallowing `context.DeadlineExceeded` from the bounded ctx.

pkg/dp-server/server/server.go
```diff
@@ -139,6 +213,13 @@ func (d *DpServer) NeedLeaderElection() bool {
 	return false
 }
 
+func (d *DpServer) WaitForDone() {
```
Automaat, line 227: **issue:** The helm chart sets `controlPlane.terminationGracePeriodSeconds=30`, which leaves comfortable headroom for the new 10s default. But `ingress.terminationGracePeriodSeconds=40`, and operators commonly tune both up to absorb long xDS drains. Operators who relied on the previous unbounded behavior with, say, a 120s pod grace will silently force-stop xDS at 10s post-upgrade, potentially causing reconnect storms on rolling deploys. Add an UPGRADE.md note and consider plumbing the value through helm so operators can tune it in lockstep with `terminationGracePeriodSeconds`.
*Location:* `pkg/config/dp-server/config.go:43-46`

**thought:** WaitForDone race in a theoretical SIGTERM-at-startup window: between `manager.startNonLeaderComponents` (`go func(c){ c.Start(stop) }`) and the goroutine actually running `started.CompareAndSwap(false, true)`, if `stop` closes, the manager's defer calls `WaitForDone`, sees `started == false`, and returns immediately. The goroutine then runs Start and enters shutdown, but the manager has already returned. The window is microseconds and not practically reachable (kuma-cp runs for some time before SIGTERM), but the test "should return from WaitForDone when the server never started" codifies the racy semantics. A close-on-construction or registration-time gating channel would close the window. Not blocking.

pkg/dp-server/server/server.go
```diff
@@ -128,7 +186,23 @@ func (d *DpServer) Start(stop <-chan struct{}) error {
...
+			log.Info("terminated normally")
+		case std_errors.Is(err, context.DeadlineExceeded):
+			d.forceStopCount.WithLabelValues(forceStopReasonTimeout).Inc()
+			log.Error(err, "dp-server force-stopped", "reason", forceStopReasonTimeout, "timeout", timeout)
+		default:
+			d.forceStopCount.WithLabelValues(forceStopReasonError).Inc()
+			log.Error(err, "dp-server force-stopped", "reason", forceStopReasonError)
+		}
+		return nil
 	case err := <-errChan:
```
Automaat, line 217: **suggestion:** `errChan` is unbuffered. If `ServeTLS` fails after the outer `select` already picked `<-stop`, the goroutine blocks on `errChan <- err` forever (the `defer close(errChan)` runs but the send happens before close). Pre-existing latent leak, but the new force-stop path (`grpcServer.Stop()` may surface ServeTLS errors post-Shutdown) makes it more likely to trigger. Buffer the channel with capacity 1, or use a select with a default.

pkg/dp-server/server/server.go
```diff
@@ -67,20 +86,59 @@ func NewDpServer(config dp_server.DpServerConfig, metrics metrics.Metrics, filte
...
+		Name: "dp_server_force_stop_total",
+		Help: "Number of times dp-server force-stopped the gRPC server during shutdown, labeled by reason.",
+	}, []string{"reason"})
+	shutdownDuration := prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name: "dp_server_shutdown_duration_seconds",
+		Help: "Wall-clock time spent in dp-server graceful shutdown.",
+		// Static buckets sized for the realistic configured range: well
+		// below the 10s default, fine-grained around it, and up to ~90s
+		// to cover operators who bump terminationGracePeriodSeconds.
+		Buckets: []float64{0.1, 0.5, 1, 2, 5, 10, 15, 30, 60, 90},
```
Automaat: **suggestion:** Histogram buckets `{0.1, 0.5, 1, 2, 5, 10, 15, 30, 60, 90}` give only one boundary (5s) between 2s and the 10s default deadline, so a 6s vs 9.9s drain is indistinguishable. Sub-100ms is also the common clean-drain case, so p50 will be flat. Compare to `pkg/xds/metrics/metrics.go:38` which uses `ExponentialBuckets`. Try `{0.01, 0.05, 0.1, 0.25, 0.5, 1, 2, 5, 7.5, 9, 10, 15, 30, 60}` or an exponential schema centered on the configured deadline.

pkg/dp-server/server/server.go
```diff
@@ -128,7 +186,23 @@ func (d *DpServer) Start(stop <-chan struct{}) error {
...
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+		start := time.Now()
+		err := shutdownDpServer(shutdownCtx, server, d.grpcServer)
+		d.shutdownDuration.Observe(time.Since(start).Seconds())
+		switch {
+		case err == nil:
+			log.Info("terminated normally")
+		case std_errors.Is(err, context.DeadlineExceeded):
+			d.forceStopCount.WithLabelValues(forceStopReasonTimeout).Inc()
```
Automaat, line 209: **suggestion:** Operators seeing `force_stop_total` increment will ask "how many streams were stuck?", but no metric answers that. A gauge sampled at shutdown start (e.g. `dp_server_active_streams_at_shutdown`) closes the loop. The code comment already says force-stop "aborts any stream whose handler ignored cancellation"; surface the count.

pkg/dp-server/server/server.go
```diff
@@ -128,7 +186,23 @@ func (d *DpServer) Start(stop <-chan struct{}) error {
...
+			log.Error(err, "dp-server force-stopped", "reason", forceStopReasonTimeout, "timeout", timeout)
```
Automaat: **issue:** `context.DeadlineExceeded` from the bounded shutdown context is the *expected* outcome when xDS streams can't drain in time under load; that's why the deadline exists. Logging this at ERROR will create false-positive noise in SIEM/alerting pipelines that page on ERROR-level kuma-cp lines. Compare `pkg/api-server/server.go:434-441` and `pkg/diagnostics/server.go:106`, which don't escalate shutdown errors to ERROR. Drop to INFO (or `log.V(1).Info`) for the timeout branch; keep ERROR only for the unexpected `default` branch.

pkg/dp-server/server/server.go
bartsmykla: good catch - default branch now returns the err, DeadlineExceeded stays swallowed.

pkg/dp-server/server/server.go
bartsmykla: added an UPGRADE.md note pointing at the env var override. Helm plumbing I'll do as a follow-up if anyone tunes it in practice. WaitForDone race is theoretical enough that I'd rather not complicate the lifecycle for it.

pkg/dp-server/server/server.go
bartsmykla: buffered to capacity 1.

pkg/dp-server/server/server.go
bartsmykla: took your suggested set.

pkg/dp-server/server/server.go
bartsmykla: fair point but I'd rather keep this PR focused on the crash fix - tracking the gauge as a follow-up.

pkg/dp-server/server/server.go
bartsmykla: dropped the timeout case to INFO, kept ERROR for the unexpected default branch.

Automaat (dismissed):

Automaat (approved):

PR #16549, opened 2026-05-11
fix(meshservice): preserve labels on MeshService update

### LINKED ISSUE #16544: Phase 2 — Generator restructure: per-service inbounds and desiredLabels helper
Parent: #16542

Pure refactor of the MeshService generator. **No behavior change** — all existing tests must pass unchanged.

## T2.1 — `meshServicesForDataplane` also returns per-service inbounds

**File:** `pkg/core/resources/apis/meshservice/generate/generator.go`

Today (line 79): returns `map[string]*meshservice_api.MeshService`.

Change to a struct that also carries `inboundsByService map[string][]*mesh_proto.Dataplane_Networking_Inbound`. The slice records which inbounds contributed to each service grouping.

Callers in `generate()` (line 168) take only the MS map for now; the inbounds slice is plumbed but unused.

## T2.2 — Extend `dataplaneAndMeshService` with `labelContribution`

**File:** same.

```go
type dataplaneAndMeshService struct {
    dataplane         model.ResourceMeta
    meshService       *meshservice_api.MeshService
    labelContribution map[string]string // nil for now; populated in Phase 4
}
```

Plumb through `generate()` so it reaches both Create and Update branches.

## T2.3 — Consolidate label-setting into a `desiredLabels(...)` helper

**File:** same.

Extract a pure function:

```go
func desiredLabels(mesh, name, zone string, propagated map[string]string) map[string]string {
    out := maps.Clone(propagated)
    if out == nil { out = map[string]string{} }
    out[metadata.KumaMeshLabel] = mesh
    out[mesh_proto.DisplayName] = name
    out[mesh_proto.ManagedByLabel] = managedByValue
    out[mesh_proto.EnvTag] = mesh_proto.UniversalEnvironment
    out[mesh_proto.ZoneTag] = zone
    out[mesh_proto.ResourceOriginLabel] = string(mesh_proto.ZoneResourceOrigin)
    return out
}
```

- Replace the inline label map at line 258-265 (Create branch) with `desiredLabels(mesh, name, g.zone, nil)`.
- In the Update branch (line 192-214), compute `desired = desiredLabels(...)` once and compare with current labels. Today's only diff was the grace-period label — preserve that exact behavior by keeping the grace-period delete logic intact.

## Acceptance

- `make check && make test TEST_PKG_LIST=./pkg/core/resources/apis/meshservice/...` passes with **zero changes** to the existing test files.
- No behavior change observable externally.

## Estimated LOC

~85 lines code change across 2 files.

## Parallel with

Phase 1 (independent).

## Blocks

Phases 3, 4, 5, 6.

Automaat:
## Motivation

When the MeshService generator updates an existing `MeshService` resource, it was replacing the entire label set with a freshly computed set of system labels. This caused any labels added externally (e.g., user-defined or downstream labels) to be silently dropped on every generator cycle.

Closes https://github.com/kumahq/kuma/issues/16544

## Implementation information

Two changes in `pkg/core/resources/apis/meshservice/generate/generator.go`:

1. **Refactor** (`refactor(meshservice): restructure generator per-service inbounds and labels`): extracted a `desiredLabels()` helper that centralises the construction of system-managed labels (mesh, display name, managed-by, env, zone, origin). Both the create and update paths now call this helper, eliminating the inline map literal duplication and making the label contract explicit.

2. **Fix** (`fix(meshservice): preserve labels on MeshService update`): the update path now clones the existing labels from the stored resource, deletes only `DeletionGracePeriodStartedLabel`, and passes the result as `propagated` into `desiredLabels()`. System labels are (re-)applied on top, so they can never be overwritten by external labels, but extra labels are preserved. A new test covers the round-trip: create a dataplane, wait for the generator to create the `MeshService`, add a custom label, mutate the dataplane to trigger an update, and assert the custom label survives.

> Changelog: fix(meshservice): preserve labels on MeshService update

### DIFF
--- a/pkg/core/resources/apis/meshservice/generate/generator.go
+++ b/pkg/core/resources/apis/meshservice/generate/generator.go
@@ -77,14 +77,12 @@ func New(
 }
 
 type meshServicesResult struct {
-	services          map[string]*meshservice_api.MeshService
-	inboundsByService map[string][]*mesh_proto.Dataplane_Networking_Inbound
+	services map[string]*meshservice_api.MeshService
 }
 
 func (g *Generator) meshServicesForDataplane(dataplane *core_mesh.DataplaneResource) meshServicesResult {
 	log := g.logger.WithValues("mesh", dataplane.GetMeta().GetMesh(), "Dataplane", dataplane.GetMeta().GetName())
 	portsByService := map[string][]meshservice_api.Port{}
-	inboundsByService := map[string][]*mesh_proto.Dataplane_Networking_Inbound{}
 	for _, inbound := range dataplane.Spec.GetNetworking().GetInbound() {
 		if g.inboundTagsDisabled && len(inbound.GetTags()) == 0 {
 			continue
@@ -114,7 +112,6 @@ func (g *Generator) meshServicesForDataplane(dataplane *core_mesh.DataplaneResou
 		}
 
 		portsByService[serviceTagValue] = append(portsByService[serviceTagValue], port)
-		inboundsByService[serviceTagValue] = append(inboundsByService[serviceTagValue], inbound)
 	}
 
 	services := map[string]*meshservice_api.MeshService{}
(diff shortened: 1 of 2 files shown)


Automaat:
Closing — reopened from fork at https://github.com/kumahq/kuma/pull/16552 (sybra now auto-detects fork remote and routes PRs cross-repo).

PR #16563, opened 2026-05-12
feat(sni): define sni format

### LINKED ISSUE #16553: Implement SNI format and propagate through the config
Implement SNI format and propagate through the config

lukidzi:
## Motivation

Based on the [MADR-101](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/101-sni-format-improvements.md) we need to implement new format of SNI

## Implementation information

* mesh-scoped zone proxy use only new SNI format
* check if MeshZoneAddress resource is available and based on this assume it uses mesh-scoped zone proxy
* added e2e checks to existing tests to check if sni are correct
* added unit tests

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/16553


pkg/xds/context/mesh_context_builder.go
```diff
@@ -256,6 +256,13 @@ func (m *meshContextBuilder) BuildIfChanged(ctx context.Context, meshName string
 		loader,
 	)
 
+	zonesWithMeshScopedProxy := map[string]bool{}
+	for _, mza := range resources.MeshZoneAddresses().Items {
+		if zone := mza.GetMeta().GetLabels()[mesh_proto.ZoneTag]; zone != "" {
```
lobkovilya: probably you can use `core_model.ZoneOfResource` func

pkg/xds/envoy/tls/sni.go
```diff
@@ -78,3 +82,53 @@ func SNIForResource(resName string, meshName string, resType model.ResourceType,
...
+	desc, err := registry.Global().DescriptorFor(id.ResourceType)
+	if err != nil {
+		panic(errors.Wrapf(err, "SNIFromKRI: unknown resource type %q", id.ResourceType))
+	}
+	switch id.ResourceType {
+	case meshservice_api.MeshServiceType,
+		meshexternalservice_api.MeshExternalServiceType,
+		meshmzservice_api.MeshMultiZoneServiceType:
+	default:
+		panic(fmt.Sprintf("SNIFromKRI: resource type %q is not supported for SNI", id.ResourceType))
```
lobkovilya: Why panic? The func already returns error and CP can recover from it by failing the config generation. I think there is no reason to panic.

pkg/xds/envoy/tls/sni_test.go
```diff
@@ -175,4 +177,175 @@ var _ = Describe("SNI", func() {
...
+			id: kri.Identifier{
+				ResourceType: meshexternalservice_api.MeshExternalServiceType,
+				Mesh:         "prod",
+				Zone:         "west",
+				Name:         "ext-backend",
+				SectionName:  "9000",
+			},
+			expected: "sni.extsvc.prod.west.ext-backend.9000",
+		}),
+		// MeshMultiZoneService — global (MeshMultiZoneService never carries zone/namespace)
```
lobkovilya: I don't think these comments before each entry are useful, they're just duplicating what's already written in the entry description.

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -146,6 +152,13 @@ func GenerateClusters(
 						sni := SniForBackendRef(realResourceRef, meshCtx, systemNamespace)
 						// ClientSideMultiIdentitiesMTLS validate MTLS enabled on the mesh
 						if proxy.WorkloadIdentity != nil {
+							if meshCtx.ZonesWithMeshScopedProxy[realResourceRef.Resource.Zone] {
+								var err error
+								sni, err = tls.SNIFromKRI(realResourceRef.Resource)
```
lobkovilya: instead of overriding the `sni` the code above should compute the sni based on the conditions (in line 152)

pkg/plugins/policies/core/xds/meshroute/clusters.go
lobkovilya, line 35: We need to refactor this func in the next PR, it's just impossible to follow, so many if/else 😞

pkg/xds/envoy/tls/sni.go
```diff
@@ -78,3 +82,53 @@ func SNIForResource(resName string, meshName string, resType model.ResourceType,
...
+		segments = append(segments, id.Zone)
+	}
+	if id.Namespace != "" {
+		segments = append(segments, id.Namespace)
+	}
+	segments = append(segments, id.Name, id.SectionName)
+
+	for _, s := range segments {
+		if strings.ContainsRune(s, '.') {
+			return "", errors.Errorf("SNIFromKRI: segment %q contains '.': %+v", s, id)
```
lobkovilya: btw we need to make sure we don't break config generation for all DPPs if there are MeshServices that have "." in their names. I think it's possible today to have such services https://github.com/kumahq/kuma/issues/13047

pkg/xds/envoy/tls/sni.go
```diff
@@ -78,3 +82,53 @@ func SNIForResource(resName string, meshName string, resType model.ResourceType,
...
+//	sni.<short>.<mesh>.<name>.<sectionName>                          (5 segments) — global-originated
+//	sni.<short>.<mesh>.<zone>.<name>.<sectionName>                   (6 segments) — zone-originated resource on universal
+//	sni.<short>.<mesh>.<zone>.<namespace>.<name>.<sectionName>       (7 segments) — zone-originated resource on k8s
+//
+// The KRI must satisfy:
+//   - ResourceType is one of MeshService, MeshExternalService, MeshMultiZoneService
+//   - Mesh, Name and SectionName are non-empty
+//   - if Namespace is non-empty, Zone must also be non-empty
+//   - no segment contains "."
+func SNIFromKRI(id kri.Identifier) (string, error) {
```
lobkovilya: Also need to validate the resulting length is 253 or less but make sure to also not break the config generation but just skip these services and log the error

lukidzi: I've added here error return but in functions that call it check and log for error

pkg/plugins/policies/core/xds/meshroute/clusters.go
lukidzi: Yes, I wanted to do it after as don't want to introduce new things

pkg/xds/envoy/tls/sni.go
lukidzi: Added skips for clusters and log error in this case

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -54,10 +57,16 @@ func GenerateClusters(
 			if meshCtx.IsExternalService(serviceName) {
 				switch {
 				case isMeshExternalService(meshCtx.EndpointMap[serviceName]):
-					sni := SniForBackendRef(service.BackendRef().RealResourceBackendRef(), meshCtx, systemNamespace)
 					if proxy.WorkloadIdentity != nil {
-						// MeshExternalService is only routable through zone egress in workload-identity mode.
-						// Skip cluster generation if no zone egress SANs are available.
+						sni, err := tls.SNIFromKRI(service.BackendRef().Resource())
+						if err != nil {
+							log.Error(err, "failed to compute SNI for cluster, skipping", "cluster", clusterName)
```
lobkovilya: This log probably will be very noisy right? If we have one MS that violates the naming rules and 1000 DPPs then we'll probably log this 1000 times every snapshot generation interval.

Maybe controller should update MES/MS statuses saying something like "this service can't be used cross-zone as its SNI exceeds the limit". And then in generators we silently skip MeshServices with invalid SNI. WDYT?

pkg/plugins/policies/meshhttproute/plugin/v1alpha1/plugin_test.go
```diff
@@ -352,6 +356,264 @@ var _ = Describe("MeshHTTPRoute", func() {
...
+						bldrs_core.NewConfigSource().Configure(bldrs_core.Sds()),
+					)
+				},
+			}
+
+			return outboundsTestCase{
+				xdsContext: *xds_builders.Context().
+					WithMeshContext(mc).
+					With(func(ctx *xds_context.Context) {
+						ctx.Mesh.ZonesWithMeshScopedProxy = map[string]bool{"": true}
```
lobkovilya: hmm this is weird, why zone is `""`?

pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go
```diff
@@ -144,6 +144,8 @@ type MeshExternalServiceStatus struct {
 	// Addresses section for generated domains
 	Addresses          []hostnamegenerator_api.Address                 `json:"addresses,omitempty"`
 	HostnameGenerators []hostnamegenerator_api.HostnameGeneratorStatus `json:"hostnameGenerators,omitempty"`
+	// Conditions is an array of current conditions of the MeshExternalService.
+	Conditions []common_api.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
lobkovilya: Do we need `patchStrategy` and `patchMergeKey` if CP is the only one who updates the status and it already uses helper functions like `UpdateConditions` to make sure `type` is unique?

pkg/plugins/runtime/gateway/cluster_generator.go
```diff
@@ -165,6 +166,9 @@ func (c *ClusterGenerator) generateRealBackendRefCluster(
 
 	service := destinationname.MustResolve(false, dest, port)
 	sni := meshroute.SniForBackendRef(backendRef, dest, port, systemNamespace)
+	if meshCtx.ZonesWithMeshScopedProxy[backendRef.Resource.Zone] {
+		sni = tls.SNIFromKRI(backendRef.Resource)
+	}
```
lobkovilya: nit: just so we don't call `SniForBackendRef` for nothing
```suggestion
    var sni string
	if meshCtx.ZonesWithMeshScopedProxy[backendRef.Resource.Zone] {
		sni = tls.SNIFromKRI(backendRef.Resource)
	} else {
	    sni = meshroute.SniForBackendRef(backendRef, dest, port, systemNamespace)
	}
```

pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -62,8 +61,11 @@ func GenerateClusters(
 					}
 					sni := SniForBackendRef(realResourceRef, dest, port, systemNamespace)
 					if proxy.WorkloadIdentity != nil {
-						// MeshExternalService is only routable through zone egress in workload-identity mode.
-						// Skip cluster generation if no zone egress SANs are available.
+						sni := tls.SNIFromKRI(service.BackendRef().Resource())
```
lobkovilya: We still should validate SNI here, but skip silently. Without validation we might get NACK from Envoy

lobkovilya (approved):

PR #16572, opened 2026-05-12
feat(meshservice): metrics and structured logs for label drops

### LINKED ISSUE #16548: Phase 5 — Metrics and structured logs for label propagation
Parent: #16542

Add observability for label-propagation drops. Operators must be able to see WHY a label didn't propagate without grepping CP logs at 3am.

## T5.1 — Metric `component_meshservice_generator_dropped_labels_total`

**File:** `pkg/core/resources/apis/meshservice/generate/generator.go`

Register alongside the existing histogram (line 60-66):

```go
droppedLabels := prometheus.NewCounterVec(prometheus.CounterOpts{
    Name: "component_meshservice_generator_dropped_labels_total",
    Help: "Number of Dataplane label/tag keys dropped during MeshService generation.",
}, []string{"reason"})
if err := metrics.Register(droppedLabels); err != nil {
    return nil, err
}
```

Two reason values:
- `invalid` — key or value failed K8s qualified-name / label-value validation.
- `inbound_conflict` — inbounds within a single DP disagree on a non-reserved tag key.

There is **no** `cross_dp_conflict` reason — the cross-DP merge is majority-wins, never drops.

Increment in `dpContribution` (Phase 3) at the drop sites.

## T5.2 — Structured warn-logs

One log line per dropped key per reconcile tick:

```go
log.Info("dropping label during MeshService generation",
    "reason", "inbound_conflict",
    "dataplane", dp.GetMeta().GetName(),
    "mesh", dp.GetMeta().GetMesh(),
    "key", k,
)
```

Rate-limiting: per-tick is already a natural smoother — once per (reconcile, DP, key) is fine. Don't add a token-bucket; if the volume is excessive we'll learn that from operating it.

## Acceptance

- Metric registered and visible at `:5680/metrics`.
- Unit tests assert the metric is incremented when invalid / conflicting keys are encountered.
- Logs include the labelled fields shown above.

## Estimated LOC

~40 lines code + ~30 lines test changes.

## Depends on

Phase 4 (logic in place to drop from).

Automaat:
## Motivation

Label propagation from Dataplane to MeshService needs observability. Without metrics and structured logs, silent drops are invisible — operators can't tell why a label was rejected or how often it happens.

Closes https://github.com/kumahq/kuma/issues/16548

## Implementation information

- Registers `component_meshservice_generator_dropped_labels_total` CounterVec (label: `reason`) in the MeshService generator.
- Emits a structured `warn` log per drop with dataplane, mesh, key, and reason fields.
- Reasons: `invalid` (label fails validation) and `inbound_conflict` (label differs across inbounds).
- Allow-list drops are demoted to `V(1)` debug and do not increment the metric — they are expected/normal behaviour, not errors.
- Adds unit tests covering metric increments, log output, and the allow-list no-metric case.


pkg/core/resources/apis/meshservice/generate/generator.go
```diff
@@ -76,10 +76,13 @@ func New(
 			allowSet[k] = struct{}{}
 		}
 	}
-	droppedLabels := prometheus.NewCounter(prometheus.CounterOpts{
+	droppedLabels := prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "component_meshservice_generator_dropped_labels_total",
-		Help: "Internal placeholder; Phase 5 replaces with a registered CounterVec.",
-	})
+		Help: "Number of Dataplane label/tag keys dropped during MeshService generation.",
+	}, []string{"reason"})
```
bartsmykla, line 82: Metric can stay invisible until the first drop. Acceptance says visible at `:5680/metrics`, so create both known series during setup.

```suggestion
	}, []string{"reason"})
	droppedLabels.WithLabelValues("invalid")
	droppedLabels.WithLabelValues("inbound_conflict")
```

pkg/core/resources/apis/meshservice/generate/generator_test.go
```diff
@@ -689,6 +689,60 @@ var _ = Describe("MeshService generator", func() {
 			}, "2s", "100ms").Should(Succeed())
 		})
 
+		It("registers component_meshservice_generator_dropped_labels_total and increments by reason", func() {
```
bartsmykla: Please assert the no-drop case too. Otherwise this only proves visibility after the first increment.

```suggestion
		It("registers component_meshservice_generator_dropped_labels_total before the first drop and increments by reason", func() {
			for _, reason := range []string{"invalid", "inbound_conflict"} {
				m := test_metrics.FindMetric(metrics, "component_meshservice_generator_dropped_labels_total", "reason", reason)
				Expect(m).ToNot(BeNil())
				Expect(m.GetCounter().GetValue()).To(Equal(0.0))
			}
```

bartsmykla (dismissed):
Looks good overall. Two small follow-ups below, but not blocking.

bartsmykla (approved):

PR #16574, opened 2026-05-12
feat(dataplane): add computed labels for zone proxy listeners

### LINKED ISSUE #16556: Add computed labels for zone proxy listeners on Dataplanes
Implement [MADR-102](https://github.com/kumahq/kuma/blob/6f19192fd666aaf19a4e8632274e1da52962e1f4/docs/madr/decisions/102-zone-proxy-listener-computed-labels.md): auto-compute two Dataplane labels based on listener presence.

**Labels:**
- `kuma.io/listener-zoneingress: enabled` — set when the Dataplane has ≥1 `ZoneIngress` listener
- `kuma.io/listener-zoneegress: enabled` — set when the Dataplane has ≥1 `ZoneEgress` listener

Both labels are set when both listener types are present; neither is set otherwise.

lobkovilya:
## Motivation

Auto-compute `kuma.io/listener-zoneingress` and `kuma.io/listener-zoneegress` labels on Dataplanes based on listener presence, as specified in [MADR-102](https://github.com/kumahq/kuma/blob/c2ecbd802cfd6944b8ace64991ac1c93f8162b5d/docs/madr/decisions/102-zone-proxy-listener-computed-labels.md).

## Implementation information

Extend `resource_labels.Compute` method to automatically add new labels to Dataplane resources.

## Supporting documentation

Fix #16556

### DIFF
--- a/pkg/core/resources/labels/registry.go
+++ b/pkg/core/resources/labels/registry.go
@@ -8,13 +8,15 @@ import (
 // if changed sync with:
 // https://github.com/Kong/shared-speakeasy/blob/b3ddd3ef1f31e42bfe71b96ea473493072f9742c/customtypes/kumalabels/kumalabels.go#L15
 var AllComputedLabels = map[string]struct{}{
-	metadata.KumaMeshLabel:         {},
-	mesh_proto.ResourceOriginLabel: {},
-	mesh_proto.ZoneTag:             {},
-	mesh_proto.EnvTag:              {},
-	mesh_proto.KubeNamespaceTag:    {},
-	mesh_proto.PolicyRoleLabel:     {},
-	mesh_proto.ProxyTypeLabel:      {},
-	metadata.KumaServiceAccount:    {},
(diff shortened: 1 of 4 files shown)


pkg/core/resources/labels/registry.go
```diff
@@ -8,13 +8,15 @@ import (
 // if changed sync with:
 // https://github.com/Kong/shared-speakeasy/blob/b3ddd3ef1f31e42bfe71b96ea473493072f9742c/customtypes/kumalabels/kumalabels.go#L15
```
Automaat, line 9: **suggestion:** `AllComputedLabels` exists to keep this set in sync with the external `Kong/shared-speakeasy/customtypes/kumalabels/kumalabels.go` registry (see the comment on the preceding line). Two new labels land here, so a companion change in `shared-speakeasy` must merge before downstream/enterprise consumers of that registry accept these labels. Confirm the follow-up PR exists or is queued.

Automaat (approved):

PR #16575, opened 2026-05-12
feat(MeshTrafficPermission): add SNI match and deny-by-default RBAC for ZoneEgress listeners

### LINKED ISSUE #16494: MeshTrafficPermission zone proxy support
### Description

Once https://github.com/kumahq/kuma/pull/16275 is merged, we'll be able to start the implem.

Prerequisites: 
* https://github.com/kumahq/kuma/issues/16432
* https://github.com/kumahq/kuma/issues/16460

Note: MeshTrafficPermission is a special case. It does not follow the common matchers semantics, so its implementation is expected to remain MTP-specific. For more context on why MeshTrafficPermission is treated differently, see the [MADR](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/080-special-mtp-algo.md).

lobkovilya:
## Motivation

Implement [MADR-103](https://github.com/kumahq/kuma/blob/6f19192fd666aaf19a4e8632274e1da52962e1f4/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md) by adding SNI match for MeshTrafficPermission policies and enabling support on the new zone egress proxy.

## Implementation information

* Introduce SNIMatch for MeshTrafficPermission (MeshFaultInjection with SNIMatch is rejected by the validator until the support is added).
* basic envoyconfig tests for MeshTrafficPermission

## Supporting documentation

Fix #16494


test/e2e_env/universal/envoyconfig/testdata/sidecars/meshtrafficpermission/inbound-rules-sni.demo-client.golden.json
lukidzi, line 1: It would be great to add test for zone proxies once we merge this: https://github.com/kumahq/kuma/pull/16599

pkg/core/resources/apis/mesh/validators.go
```diff
@@ -460,6 +460,18 @@ func ValidateMatch(match common_api.Match) validators.ValidationError {
...
 	}
+	if match.SNI != nil {
+		switch match.SNI.Type {
+		case common_api.SNIExactMatchType:
+		case "":
+			verr.AddViolation("sni.type", "must be set")
+		default:
+			verr.AddViolation("sni.type", fmt.Sprintf("unrecognized type %q, supported values are: Exact", match.SNI.Type))
+		}
+		if match.SNI.Value == "" {
```
lukidzi, line 472: Do we need to check if SNI is RFC1035 complient?

pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/plugin.go
```diff
@@ -53,14 +53,19 @@ func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *
 		return nil
 	}
 
+	mtp := proxy.Policies.Dynamic[api.MeshTrafficPermissionType]
+
+	if err := p.configureZoneEgressListeners(rs, mtp); err != nil {
```
lukidzi: Do we need to configure it if `if !ctx.Mesh.Resource.MTLSEnabled() && proxy.WorkloadIdentity == nil {` as egress makes only sense with it?

pkg/plugins/policies/meshfaultinjection/api/v1alpha1/validator.go
```diff
@@ -72,6 +72,9 @@ func validateMatches(field string, matches []common_api.Match) validators.Valida
 	var verr validators.ValidationError
 	for idx, match := range matches {
 		path := validators.RootedAt(field).Index(idx)
+		if match.SNI != nil {
+			verr.AddViolationAt(path.Field("sni"), "is not supported on MeshFaultInjection")
+		}
```
lukidzi, line 77: Maybe this should be in `ValidateMatch`

pkg/core/resources/apis/mesh/validators.go
lobkovilya: BTW we need to validate with RFC1123 because it's a newer RFC for hostnames and it allows label to be a number and that's what we need in SNI for section name.

pkg/plugins/policies/meshfaultinjection/api/v1alpha1/validator.go
lobkovilya: ValidateMatch is common but here we need to validate only for MeshFaultInjection as new "sni" is not yet supported

lukidzi (dismissed):

lukidzi (approved):

PR #16586, opened 2026-05-13
feat(log): resilient component restart logging

bartsmykla:
## Motivation

When a resilient-wrapped component crashes, the wrapper backs off and retries with nothing in the logs between the original error and the next attempt. On a stuck zone CP this leaves no signal: how many retries so far, how long the outage has been running, what error keeps coming back, which global address is unreachable. Sibling of #16499.

## Implementation information

After each termination, emit `scheduling component restart` at Info with `attempt`, `sinceLastSuccess`, `nextBackoff`, `lastError` (cleared on clean exit), and a `suppressed` counter. The first 5 attempts always log; after that, at most one per 10s, matching the panic limiter window in `runComponent`.

The stable-run boundary (`time.Since(lastStart) > backoffMaxTime`) resets backoff, the success baseline, and the sampling state, so the first restart log of a new outage doesn't carry over suppression or counts from the previous one.

Zone KDS attaches `peer=<global address>` to the resilient logger so every line carries it.

## Sample output

```text
INFO  kds-zone.kds-mux-client  starting resilient component  peer=grpcs://global:5685
ERROR kds-zone.kds-mux-client  component terminated with an error  peer=grpcs://global:5685 generationID=1 error="rpc error: code = Unavailable ..."
INFO  kds-zone.kds-mux-client  scheduling component restart  peer=grpcs://global:5685 attempt=1 sinceLastSuccess=12ms nextBackoff=5s lastError="rpc error: code = Unavailable ..."
INFO  kds-zone.kds-mux-client  scheduling component restart  peer=grpcs://global:5685 attempt=12 sinceLastSuccess=1m4s nextBackoff=1m suppressed=4 lastError="..."
```

## Tests

`resilient_test.go` exercises the public contract end-to-end with tiny backoffs and a scripted inner component. It covers the three correctness invariants that came up in review: `sinceLastSuccess` after a stable run, sampling state reset on the stable-run boundary, and `lastError` cleared on clean exit. Reverting any of those four lines in `resilient.go` fails exactly one test.

### DIFF
--- a/pkg/core/runtime/component/resilient.go
+++ b/pkg/core/runtime/component/resilient.go
@@ -37,6 +37,9 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 		lastSuccess    = time.Now()
 		nextSampledAt  time.Time
 		suppressedLogs uint64
+		// attempt counts restarts within the current outage and drives sampling.
+		// generationID stays monotonic across stable-run boundaries for cross-outage tracing.
+		attempt uint64
 	)
 	for generationID := uint64(1); ; generationID++ {
 		lastStart := time.Now()
@@ -74,11 +77,13 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 			lastSuccess = time.Now()
 			nextSampledAt = time.Time{}
 			suppressedLogs = 0
+			attempt = 0
 		}
 		dur, _ := backoff.Next()
-		if generationID <= resilientLogSampleAfterAttempt || !time.Now().Before(nextSampledAt) {
+		attempt++
+		if attempt <= resilientLogSampleAfterAttempt || !time.Now().Before(nextSampledAt) {
 			fields := []any{
-				"attempt", generationID,
+				"attempt", attempt,
 				"sinceLastSuccess", time.Since(lastSuccess),
 				"nextBackoff", dur,
 			}
@@ -90,7 +95,8 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 				suppressedLogs = 0
 			}
 			r.log.Info("scheduling component restart", fields...)
-			if generationID > resilientLogSampleAfterAttempt {
+			// Arm the sampling window once we've hit the burst threshold.
+			if attempt >= resilientLogSampleAfterAttempt {
 				nextSampledAt = time.Now().Add(resilientLogSampleWindow)
 			}
 		} else {
--- a/pkg/core/runtime/component/resilient.go
+++ b/pkg/core/runtime/component/resilient.go
@@ -79,16 +79,27 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 			suppressedLogs = 0
 			attempt = 0
 		}
+		// Clean exit: restart immediately without a sampled restart-log line
+		// and without a backoff delay. Operators read "scheduling restart"
+		// as "we hit a failure", so emitting it with no lastError is noise.
+		if lastError == nil {
+			select {
+			case <-stop:
+				r.log.Info("done")
+				return nil
+			default:
+				continue
+			}
+		}
 		dur, _ := backoff.Next()
 		attempt++
 		if attempt <= resilientLogSampleAfterAttempt || !time.Now().Before(nextSampledAt) {
 			fields := []any{
+				"generationID", generationID,
 				"attempt", attempt,
 				"sinceLastSuccess", time.Since(lastSuccess),
 				"nextBackoff", dur,
-			}
-			if lastError != nil {
-				fields = append(fields, "lastError", lastError.Error())
+				"lastError", lastError.Error(),
 			}
 			if suppressedLogs > 0 {
 				fields = append(fields, "suppressed", suppressedLogs)
@@ -102,7 +113,12 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 		} else {
 			suppressedLogs++
 		}
-		<-time.After(dur)
+		select {
+		case <-stop:
+			r.log.Info("done")
+			return nil
+		case <-time.After(dur):
+		}
 	}
 }
 
--- a/pkg/core/runtime/component/resilient.go
+++ b/pkg/core/runtime/component/resilient.go
@@ -8,6 +8,11 @@ import (
 	"github.com/sethvargo/go-retry"
 )
 
+const (
+	resilientLogSampleAfterAttempt = 5
+	resilientLogSampleWindow       = 10 * time.Second
+)
+
 type resilientComponent struct {
 	log             logr.Logger
 	component       Component
@@ -25,14 +30,19 @@ func NewResilientComponent(log logr.Logger, component Component, backoffBaseTime
 }
 
 func (r *resilientComponent) Start(stop <-chan struct{}) error {
-	r.log.Info("starting resilient component ...")
+	r.log.Info("starting resilient component")
 	backoff := r.newBackoff()
+	var (
+		lastError      error
+		lastSuccess    = time.Now()
+		nextSampledAt  time.Time
+		suppressedLogs uint64
+	)
 	for generationID := uint64(1); ; generationID++ {
 		lastStart := time.Now()
 		errCh := make(chan error, 1)
 		go func(errCh chan<- error) {
 			defer close(errCh)
-			// recover from a panic
 			defer func() {
 				if e := recover(); e != nil {
 					if err, ok := e.(error); ok {
@@ -42,7 +52,6 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 					}
 				}
 			}()
-
 			errCh <- r.component.Start(stop)
 		}(errCh)
 		select {
@@ -51,20 +60,42 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 			return nil
 		case err := <-errCh:
 			if err != nil {
-				// TODO: add per-name rate limiting here analogous to panicLogLimiter in runComponent;
-				// resilientComponent logs every restart, so a tight panic loop produces unbounded log volume.
+				lastError = err
 				r.log.WithValues("generationID", generationID).Error(err, "component terminated with an error")
+			} else {
+				// Clean exit: don't carry a prior failure into the next restart log.
+				lastError = nil
 			}
 		}
 		if time.Since(lastStart) > r.backoffMaxTime {
-			// reset backoff so in a case of event with the following steps
-			// 1) Component is unhealthy until max backoff is reached
-			// 2) Component is healthy for at least backoff max time
-			// 3) Component is unhealthy
-			// We start backoff from the beginning
+			// Stable run ended: reset backoff, success baseline, and sampling
+			// state so a new outage starts fresh.
 			backoff = r.newBackoff()
+			lastSuccess = time.Now()
+			nextSampledAt = time.Time{}
+			suppressedLogs = 0
 		}
 		dur, _ := backoff.Next()
+		if generationID <= resilientLogSampleAfterAttempt || !time.Now().Before(nextSampledAt) {
+			fields := []any{
+				"attempt", generationID,
+				"sinceLastSuccess", time.Since(lastSuccess),
+				"nextBackoff", dur,
+			}
+			if lastError != nil {
+				fields = append(fields, "lastError", lastError.Error())
+			}
+			if suppressedLogs > 0 {
+				fields = append(fields, "suppressed", suppressedLogs)
+				suppressedLogs = 0
+			}
+			r.log.Info("scheduling component restart", fields...)
+			if generationID > resilientLogSampleAfterAttempt {
+				nextSampledAt = time.Now().Add(resilientLogSampleWindow)
+			}
+		} else {
+			suppressedLogs++
+		}
 		<-time.After(dur)
 	}
 }
--- a/pkg/core/runtime/component/resilient_test.go
+++ b/pkg/core/runtime/component/resilient_test.go
@@ -1,6 +1,8 @@
 package component
 
 import (
+	"errors"
+	"fmt"
 	"slices"
 	"sync"
 	"sync/atomic"
@@ -9,7 +11,6 @@ import (
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"github.com/pkg/errors"
 )
 
 type logEntry struct {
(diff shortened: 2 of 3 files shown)


pkg/core/runtime/component/resilient.go
```diff
@@ -51,20 +60,42 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
...
+			lastSuccess = time.Now()
+			nextSampledAt = time.Time{}
+			suppressedLogs = 0
 		}
 		dur, _ := backoff.Next()
+		if generationID <= resilientLogSampleAfterAttempt || !time.Now().Before(nextSampledAt) {
+			fields := []any{
+				"attempt", generationID,
+				"sinceLastSuccess", time.Since(lastSuccess),
+				"nextBackoff", dur,
```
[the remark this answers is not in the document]
bartsmykla: Both correct. Fixed in 3d5fbadc52: separate outage-local `attempt` counter that resets on the stable-run boundary and drives sampling plus the user-visible `attempt` field; `generationID` keeps its role on the component-terminated Error line. Window armed at `>=` so attempt N+1 is subject to sampling, not the first one outside the burst. Added a regression test that asserts the first 5 restart logs of a new outage are always-logged after a stable run with attempts 1..N.

pkg/core/runtime/component/resilient.go
```diff
@@ -51,20 +63,45 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
...
-			// 3) Component is unhealthy
-			// We start backoff from the beginning
+			// Stable run ended: reset backoff, success baseline, and sampling
+			// state so a new outage starts fresh.
 			backoff = r.newBackoff()
+			lastSuccess = time.Now()
+			nextSampledAt = time.Time{}
+			suppressedLogs = 0
+			attempt = 0
 		}
```
Automaat, line 81: **issue:** `attempt` is the loop counter (`generationID` at line 41), incremented every iteration since `Start()` began and never reset. The stable-run branch resets `lastSuccess`, `nextSampledAt`, and `suppressedLogs` but leaves `generationID` alone. So after a 10-day stable run with a single transient failure, the log emits `attempt=N` where N is the lifetime restart count, not "attempt 1 of a new outage". The PR sample output (`attempt=12 ... suppressed=4`) suggests the author wants a lifetime counter, but the field name promises per-outage semantics. Either rename to `restart` / `generation`, or add a separate per-outage counter that resets alongside `lastSuccess`. As written, an alert like `attempt > 5` fires constantly on long-running CPs with one weekly transient failure.

pkg/core/runtime/component/resilient.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/sethvargo/go-retry"
 )
 
+const (
```
Automaat, line 11: **suggestion:** Magic constants `resilientLogSampleAfterAttempt = 5` and `resilientLogSampleWindow = 10 * time.Second` should be `ResilientComponent` config fields with defaults, alongside `backoffBaseTime` / `backoffMaxTime`. Same reasoning as the existing backoff config: ops may want it tighter for critical KDS components and looser for noisy insights resyncers. Tests at line 208 also reach into the package constant. A struct field lets tests inject directly.

pkg/core/runtime/component/resilient.go
```diff
@@ -25,14 +30,22 @@ func NewResilientComponent(log logr.Logger, component Component, backoffBaseTime
 }
 
 func (r *resilientComponent) Start(stop <-chan struct{}) error {
-	r.log.Info("starting resilient component ...")
+	r.log.Info("starting resilient component")
 	backoff := r.newBackoff()
+	var (
```
Automaat, line 35: **suggestion:** Sampling state lives in `Start()` locals (lines 35-40). If anyone ever calls `Start()` twice on the same wrapper (the interface doesn't forbid it), the burst-threshold-then-sample logic resets. `panicLogLimiter` lives on the manager struct (`component.go:144`) so its state survives. Move sampling state to the `resilientComponent` struct. Also sets up future extraction without committing to it now.

pkg/core/runtime/component/resilient.go
```diff
@@ -51,20 +63,45 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
...
+			if lastError != nil {
+				fields = append(fields, "lastError", lastError.Error())
+			}
+			if suppressedLogs > 0 {
+				fields = append(fields, "suppressed", suppressedLogs)
+				suppressedLogs = 0
+			}
+			r.log.Info("scheduling component restart", fields...)
+			// Arm the sampling window once we've hit the burst threshold.
+			if attempt >= resilientLogSampleAfterAttempt {
```
Automaat, line 110: **issue:** `<-time.After(dur)` is uninterruptible by `stop`. The select on `stop` only exists at line 57 (waiting on `errCh`). Once a generation enters the post-error backoff sleep at line 99, closing `stop` waits for the timer to fire. With default `ResilientComponentMaxBackoff` (commonly 60s in prod), `kuma-cp` shutdown can be delayed up to one backoff window per resilient component. Pre-existing behavior, but the PR is the right place to fix it because the new log line announces a restart that never happens during shutdown, producing misleading "scheduling component restart" entries seconds before "done". Replace with:
```go
select {
case <-stop:
    r.log.Info("done")
    return nil
case <-time.After(dur):
}
```

pkg/core/runtime/component/resilient.go
```diff
@@ -51,20 +63,45 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
...
-			// 2) Component is healthy for at least backoff max time
-			// 3) Component is unhealthy
-			// We start backoff from the beginning
+			// Stable run ended: reset backoff, success baseline, and sampling
+			// state so a new outage starts fresh.
 			backoff = r.newBackoff()
+			lastSuccess = time.Now()
+			nextSampledAt = time.Time{}
+			suppressedLogs = 0
+			attempt = 0
```
Automaat, line 80: **issue:** Restart log missing `generationID`, breaking correlation with the termination error. Line 64 emits `"component terminated with an error"` with `generationID=N`. The follow-up `"scheduling component restart"` at line 92 carries `attempt=N+1` but no `generationID`. An operator grepping `generationID=42` finds the failure but not the restart decision (sample state, suppressed count, `nextBackoff`). Either rename `attempt` -> `generationID` (consistent vocabulary) or include both.

pkg/core/runtime/component/resilient.go
```diff
@@ -51,20 +63,45 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
 			return nil
 		case err := <-errCh:
 			if err != nil {
```
Automaat, line 65: **issue:** Clean exit still emits "scheduling component restart". Lines 65-67 clear `lastError = nil` after `errCh` returns nil, but the code falls through to the unconditional restart-log block. Operators reading `scheduling component restart` with no `lastError` and `sinceLastSuccess=0` will be confused. If clean exits aren't expected, log once at INFO and skip the sampled "scheduling restart" entirely on this path. If clean exits ARE expected, reset the backoff and emit a distinct message rather than the same string an alert may match on.

pkg/core/runtime/component/resilient_test.go
```diff
@@ -0,0 +1,349 @@
...
+// scriptOfErrors lifts a list of errors into a script that returns one error per Start call.
+func scriptOfErrors(errs ...error) []func() error {
+	out := make([]func() error, len(errs))
+	for i, e := range errs {
+		out[i] = func() error { return e }
+	}
+	return out
+}
+
+var _ = Describe("resilientComponent", func() {
```
Automaat, line 122: **issue:** Two critical paths uncovered by tests:
1. **Panic recovery** (lines 44-54). No test triggers a panic from `scriptedComponent.Start`. The `defer recover()` path is the entire reason this wrapper keeps the manager alive (per the comment at `component.go:170`). Add a script entry `func() error { panic(errors.New("synthetic")) }` and assert (a) loop continues, (b) `"component terminated with an error"` carries the panic as `error`, (c) non-error panics go through the `errors.Errorf("%v", e)` path.
2. **`stop` during backoff sleep**, tied to the issue above. Once fixed, add a test that closes `stop` while the loop sits in `<-time.After(dur)` and asserts `Start` returns within ~50ms with tiny backoffMax.

pkg/core/runtime/component/resilient.go
```diff
@@ -51,20 +63,45 @@ func (r *resilientComponent) Start(stop <-chan struct{}) error {
...
+			lastSuccess = time.Now()
+			nextSampledAt = time.Time{}
+			suppressedLogs = 0
+			attempt = 0
 		}
 		dur, _ := backoff.Next()
+		attempt++
+		if attempt <= resilientLogSampleAfterAttempt || !time.Now().Before(nextSampledAt) {
+			fields := []any{
+				"attempt", attempt,
```
Automaat, line 99: **issue:** Possible credential leak via `lastError` from inner components. `NewResilientComponent` wraps the postgres event listener (`pkg/plugins/resources/postgres/plugin.go:57`), and pgx parse errors echo the offending DSN substring, which contains `password='<plaintext>'` (the DSN is built at `pkg/config/plugins/resources/postgres/config.go:107`). Line 64 already logs the error at Error level (existing behavior), but the new Info-level emission widens the audience: operators who ship Info to multi-tenant aggregators while keeping Error private now see the credentials too. No error-sanitization helper exists in `pkg/` (only `pkg/envoy/admin/sanitize.go` for Envoy configs).

OTEL plugin (`pkg/plugins/runtime/opentelemetry/plugin.go:67`) errors include endpoint URLs that sometimes carry bearer tokens in user-configured headers. Same exposure path.

pkg/kds/zone/components.go
```diff
@@ -58,7 +58,7 @@ func Setup(rt core_runtime.Runtime) error {
 		deltaServer,
 	)
 	return rt.Add(component.NewResilientComponent(
-		kdsZoneLog.WithName("kds-mux-client"),
+		kdsZoneLog.WithName("kds-mux-client").WithValues("peer", rt.Config().Multizone.Zone.GlobalAddress),
```
Automaat: **issue:** `peer` URL may include embedded credentials. `rt.Config().Multizone.Zone.GlobalAddress` is `string` validated only with `url.Parse` + scheme check (`pkg/config/multizone/multicluster.go:92`). `url.Parse("grpcs://user:pass@global:5685")` succeeds. Nothing strips `Userinfo`. `ZoneConfig.Sanitize()` (line 74) doesn't touch `GlobalAddress`. Log via `u.Redacted()` (stdlib `net/url` masks userinfo) or just host:port.

pkg/core/runtime/component/resilient.go
Automaat, line 99: **suggestion:** Add a `redactErr(err) string` helper in this package. Two regex passes on `password=...` (postgres keyword DSN) and `://[^:]+:[^@]+@` (URL userinfo) cover the obvious cases. Cheap, low-risk, protects future callers wrapping secret-bearing inner components.

pkg/core/runtime/component/resilient_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+import (
+	"slices"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/go-logr/logr"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
```
Automaat, line 13: **issue:** Wrong errors package for this directory. Neighbor `pkg/core/runtime/component/component_manager_test.go:4` imports stdlib `"errors"`. The new test imports `github.com/pkg/errors`. Same package, two different errors packages across files is exactly what convention reviews flag. The new test uses `errors.New` and `errors.Errorf`, both available via stdlib `errors.New` + `fmt.Errorf`. Switch to stdlib to match the sibling file.

pkg/core/runtime/component/resilient.go
bartsmykla: Done. Wrapped the backoff sleep in a select on stop so shutdown is no longer delayed up to one backoff window per resilient component.

pkg/core/runtime/component/resilient.go
bartsmykla: Done. Added `generationID` to the restart-log field tuple so operators can pivot between the "component terminated" Error and the matching "scheduling restart" Info.

pkg/core/runtime/component/resilient.go
bartsmykla: Done. Clean exit now restarts silently - the restart log only fires when `lastError` is non-nil, so operators won't see "scheduling restart" with no error context.

pkg/kds/zone/components.go
bartsmykla: Done. `globalAddress` now goes through `url.Parse` + `u.Redacted()` before reaching `WithValues("peer", ...)`, so any embedded userinfo is masked.

pkg/core/runtime/component/resilient_test.go
bartsmykla: Done. Added two specs: one drives both error-typed and string-typed panics through the wrapped component and asserts the recover path emits a "component terminated" Error per restart; the other closes `stop` while the wrapper is inside the backoff sleep and asserts `Start` returns within 500ms against a 2s backoff.

pkg/core/runtime/component/resilient_test.go
bartsmykla: Done. Switched to stdlib `errors` + `fmt.Errorf` to match `component_manager_test.go`. The non-test files still use `github.com/pkg/errors` because the panic-recovery path needs `errors.WithStack`.

pkg/core/runtime/component/resilient.go
bartsmykla: Stale - this read the pre-Copilot-fix code. `attempt` and `generationID` are now separate counters: `attempt` is outage-local and resets on stable-run boundary (line 80), `generationID` stays monotonic. The restart log emits both so `attempt > 5` alerts fire per outage, not lifetime.

pkg/core/runtime/component/resilient.go
bartsmykla: Holding off. No current consumer asks for per-component tuning, and `NewResilientComponent` callers across kuma/downstream would all need to be updated. Happy to revisit when an ops case appears.

pkg/core/runtime/component/resilient.go
bartsmykla: Holding off. `Start` isn't called twice on any wrapper today; the locals are simpler and lifetime matches the goroutine. `panicLogLimiter` on the manager exists because the manager itself is shared across components - the resilient wrapper isn't.

pkg/core/runtime/component/resilient.go
bartsmykla: The same error string already lands in the unconditional Error log at "component terminated with an error" (line 64) - the new Info collocates it with the restart context but doesn't broaden the exposure surface. Hardening pgx/OTEL error strings is a broader concern that crosses many call sites and deserves a separate PR.

bartsmykla: Folding into the broader hardening tracked in the parent thread - introducing a redact helper here without addressing the unconditional Error log a few lines above gives a false sense of safety.

Automaat (approved):

PR #16590, opened 2026-05-13
fix(meshservice): track labels with qualified key names via hash

### LINKED ISSUE #16587: bug(meshservice): propagated labels with dotted key names stick forever after carrier DP removed
## Summary

Propagated DP labels with names that are valid Kubernetes label *keys* but invalid as label *values* (e.g. keys with dots like `foo.example.com/bar`) stick on the generated MeshService forever after the carrier Dataplane is removed.

## Reproduction

1. Run Universal zone CP with `meshService.labelPropagation.enabled: true`, mesh in `Exclusive` mode.
2. Apply DP `dp-anchor` with `kuma.io/service: svc` and no extra labels (keeps MS alive).
3. Apply DP `dp-slash` with resource label `foo.example.com/bar: x` and `kuma.io/service: svc`.
4. Observe MeshService `svc` has `foo.example.com/bar: x` propagated. ✓ (expected)
5. Delete `dp-slash`.
6. Wait one reconcile interval. The anchor DP keeps `svc` alive past grace.
7. **Observed**: `foo.example.com/bar: x` is still on the MS, even though no DP carries it.
8. **Expected**: label removed since no DP votes for it.

## Root cause

`pkg/core/resources/apis/meshservice/generate/labels.go:165-185` `addPropagationTracking`:

```go
for k := range propagated {
    if len(validation.IsValidLabelValue(k)) == 0 {
        keys = append(keys, k)
    }
}
```

Tracking writes the propagated key NAME as a label VALUE under `kuma.io/pkey-N`. Kubernetes label-value validation rejects strings with dots, so keys like `foo.example.com/bar` are silently skipped. Next reconcile, `extractPropagatedKeys` sees no tracking entry for that label and treats it as operator-added → it is preserved indefinitely.

The function's docstring acknowledges the skip but the consequence (forever-stuck labels after carrier removal) is the bug.

## Impact

- Any DP resource label with a dotted key name silently leaks onto the MS once propagated and cannot be removed by reconcile.
- Removing the label requires manual MS edit.

## Proposed fixes (pick one)

1. Use a different tracking encoding that does not constrain key shape (e.g. encode in a single annotation/JSON, or base32-encode the key).
2. On the cleanup path, when a label is not in tracking but also not a known operator-supplied key, double-check against current DP labels; remove if no DP carries it.
3. Reject DP labels whose names cannot be tracked at the validator (DP `Validate()`).

Option 1 is the most general fix and matches Kubernetes annotation conventions for storing arbitrary state.

## Found during

Manual testing of #16542. Reproduced on master @ c1fa90835c.

Automaat:
## Motivation

Propagated DP labels with qualified key names (e.g. `foo.example.com/bar`) were silently
skipped from the tracking store because `addPropagationTracking` wrote the raw key name
as a Kubernetes label *value*, and label values reject `/` and `.`. Without a tracking
entry, `extractPropagatedKeys` treated those labels as operator-added on cleanup, so they
persisted on the MeshService indefinitely after the carrier Dataplane was removed.

## Implementation information

Replace the raw-key-as-value encoding with a SHA-256 hash of the key name stored in the
label key suffix (`kuma.io/pkey-<8hex>`). This removes the character-set restriction while
keeping tracking entirely within Kubernetes labels.

`extractPropagatedKeys` now returns a `func(string) bool` closure that encapsulates the
hash-based lookup. A backward-compat path (`handleOldTrackingFormat`) handles the old
plain-value format during upgrades, migrating entries on the next reconcile.

Alternatives considered:
- Store tracking in an annotation as JSON (avoids label constraints entirely) — deferred
  as a larger structural change; hash encoding solves the bug with minimal diff.
- Double-check DP labels on cleanup path when tracking is absent — fragile, doesn't fix
  the root cause.
- Reject dotted-key DP labels at the validator — too restrictive, breaks valid use-cases.

## Supporting documentation

Closes https://github.com/kumahq/kuma/issues/16587

### DIFF
--- a/pkg/core/resources/apis/meshservice/generate/generator.go
+++ b/pkg/core/resources/apis/meshservice/generate/generator.go
@@ -280,7 +280,7 @@ func (g *Generator) generate(ctx context.Context, mesh string, dataplanes []*cor
 				if mesh_proto.IsReservedLabelKey(k) {
 					continue
 				}
-				if prevPropagated[k] {
+				if prevPropagated(k) {
 					continue
 				}
 				if _, inDesired := desired[k]; !inDesired {
--- a/pkg/core/resources/apis/meshservice/generate/generator_test.go
+++ b/pkg/core/resources/apis/meshservice/generate/generator_test.go
@@ -618,6 +618,40 @@ var _ = Describe("MeshService generator", func() {
 			}, "2s", "100ms").Should(Succeed())
 		})
 
+		It("removes a propagated dotted-key Dataplane resource-label when DP stops carrying it", func() {
+			// Same regression as above but exercised via Dataplane metadata labels
+			// (the path reported in the linked issue) instead of inbound tags.
+			dp := builders.Dataplane().
+				WithAddress("127.0.0.1").
+				WithoutInbounds().
+				AddInbound(builders.Inbound().
+					WithPort(80).
+					WithServicePort(8080).
+					WithTags(map[string]string{mesh_proto.ServiceTag: "backend"}),
+				).
+				Build()
+			Expect(resManager.Create(context.Background(), dp,
+				store.CreateByKey("dp-1", model.DefaultMesh),
+				store.CreateWithLabels(map[string]string{"app.example.com/tier": "gold"}),
+			)).To(Succeed())
+
+			ms := meshservice_api.NewMeshServiceResource()
+			Eventually(func(g Gomega) {
+				g.Expect(resManager.Get(context.Background(), ms, store.GetByKey("backend", model.DefaultMesh))).To(Succeed())
+				g.Expect(ms.GetMeta().GetLabels()).To(HaveKeyWithValue("app.example.com/tier", "gold"))
+			}, "2s", "100ms").Should(Succeed())
+
+			loaded := core_mesh.NewDataplaneResource()
+			Expect(resManager.Get(context.Background(), loaded, store.GetByKey("dp-1", model.DefaultMesh))).To(Succeed())
+			Expect(resManager.Update(context.Background(), loaded, store.UpdateWithLabels(map[string]string{}))).To(Succeed())
+
+			Eventually(func(g Gomega) {
+				g.Expect(resManager.Get(context.Background(), ms, store.GetByKey("backend", model.DefaultMesh))).To(Succeed())
+				g.Expect(ms.GetMeta().GetLabels()).ToNot(HaveKey("app.example.com/tier"))
+				g.Expect(ms.GetMeta().GetLabels()).To(HaveKeyWithValue("kuma.io/mesh", model.DefaultMesh))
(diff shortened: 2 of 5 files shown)


pkg/core/resources/apis/meshservice/generate/generator.go
```diff
@@ -280,7 +280,7 @@ func (g *Generator) generate(ctx context.Context, mesh string, dataplanes []*cor
 				if mesh_proto.IsReservedLabelKey(k) {
 					continue
 				}
-				if prevPropagated[k] {
+				if prevPropagated(k) {
```
bartsmykla, line 283: This still does not handle MeshServices already affected by the old code. Those labels were never tracked, so `prevPropagated(k)` is false and this path preserves them as external forever. That means the PR fixes new reconciles, but not existing leaked labels from an upgrade. Add a migration cleanup path, or call this out explicitly in the changelog/upgrade notes.

pkg/core/resources/apis/meshservice/generate/labels.go
```diff
@@ -20,38 +20,54 @@ import (
 // operator-managed labels (which must be preserved).
 const propagationTrackingPrefix = "kuma.io/pkey-"
 
+// trackingKeyHash returns an 8-hex-char fingerprint of labelKey. Storing the
+// hash in the label key (instead of the key name in the label value) avoids the
+// label-value character restriction: valid Kubernetes label keys can contain "/"
+// (e.g. "foo.example.com/bar") which is not allowed in label values, causing
+// such keys to be silently skipped and never cleaned up after DP removal.
+func trackingKeyHash(labelKey string) string {
+	h := sha256.Sum256([]byte(labelKey))
```
bartsmykla, line 33: Use more than 32 bits here. A hash collision would make an unrelated external label look previously propagated and then we drop it during cleanup. There is enough label key budget for 16 hex chars. Please update the comment above too if you take this.

```suggestion
	return fmt.Sprintf("%x", h[:8]) // 16 hex chars, still short enough for a label key suffix
```

pkg/core/resources/apis/meshservice/generate/generator_test.go
```diff
@@ -581,6 +581,43 @@ var _ = Describe("MeshService generator", func() {
 			}, "2s", "100ms").Should(Succeed())
 		})
 
+		It("removes a propagated dotted-key label (qualified name with '/') when DP stops carrying it", func() {
```
bartsmykla, line 584: Add the DP resource-label path too. The issue repro uses `Dataplane` metadata labels, while this test uses inbound tags. The shared tracking code is the same, but matching the reported path avoids a blind spot.

bartsmykla (commented):
Left a few comments. Main thing is upgrade completeness; this looks good as a forward fix.

bartsmykla (approved):

PR #16597, opened 2026-05-13
feat(meshmetric): support zone proxy Dataplanes

### LINKED ISSUE #16558: MeshMetric: support mesh-scoped zone proxy Dataplanes
### Description

Per [MADR-103](https://github.com/kumahq/kuma/blob/6f19192fd666aaf19a4e8632274e1da52962e1f4/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md) group (4): implement `MeshMetric` generation on mesh-scoped zone proxies.

Automaat:
## Motivation

Closes #16558. Implements [MADR-103](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md) **group (4)** for `MeshMetric` on mesh-scoped zone proxy `Dataplane` resources (those carrying `networking.listeners[]` of type `ZoneIngress`/`ZoneEgress` per MADR-095).

The matching and computed-label wiring already landed in #16574 and #16584; `MeshMetric` already flows into `plugin.Apply()` on these proxies via the `DefaultProxy` xDS profile. This PR closes two behaviour gaps for zone-proxy-only DPPs:

1. `conf.applications[]` is meaningless when there's no co-located workload to scrape.
2. The `service` label in the dynconf payload falls back to `"unknown"` because there's no `kuma.io/service` tag.

`MeshTrace` and `MeshProxyPatch` (the rest of group 4) will follow in separate PRs.

## Implementation information

* `sanitizeConfForProxy(conf, proxy)` runs at the top of `plugin.Apply`. When `proxy.Dataplane.Spec.GetNetworking().IsZoneProxyOnly()` is true, it clears `conf.Applications` and logs a structured warning (`policy`, `dataplane`, `mesh`). Mixed DPPs (inbound + embedded zone listener) keep `applications[]`.
* In `createDynamicConfig`, the existing `extraLabels[\"service\"]` fallback gets a tail check: if `IdentifyingName` returns `mesh_proto.ServiceUnknown` and the DPP is zone-proxy-only, override with `proxy.Dataplane.GetMeta().GetName()`. Stable, unique within the mesh, and matches what an operator sees in `kubectl get dataplane`.
* Field doc-comment on `MeshMetric.applications[]` now states the field is ignored on zone-proxy-only Dataplanes; this propagates to the generated CRD/OpenAPI/`rest.yaml`.

Three new table entries in `plugin_test.go` cover the new shapes (`zone_egress_only`, `zone_ingress_only`, `mixed_inbound_and_zone_egress`), each with `*.listeners.golden.yaml`/`*.clusters.golden.yaml`. A separate `Describe(\"dynconf payload on zone-proxy DPPs\")` block adds non-golden `It` assertions on the dynconf payload (applications cleared on zone-proxy-only, service label resolves correctly across all three shapes) so a future YAML-formatting drift in unrelated goldens cannot mask a regression in the actual contract.

### Considered and discarded

* Reject `applications[]` at policy admission — Kuma's policy validator has no proxy context (one policy can match many DPP shapes), so the only practical boundary is apply-time. The warning is currently logged on the CP; promoting it to a policy-resource status condition is a separate change that would benefit all four MADR-103 groups.
* Special-case the legacy global-scoped `ZoneIngress`/`ZoneEgress` resource types — explicitly out of scope; they're being removed in 3.0 per MADR-103.

## Supporting documentation

* Issue: #16558
* MADR-103: policy matching on mesh-scoped zone proxies
* MADR-095: mesh-scoped zone ingress/egress
* Precursors: #16574 (computed labels), #16584 (listener-aware matchers)


pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -308,15 +361,33 @@ func createDynamicConfig(
 	if zone := proxy.Dataplane.GetMeta().GetLabels()[mesh_proto.ZoneTag]; zone != "" {
 		extraLabels["zone"] = zone
 	}
-	if workloadName := proxy.Dataplane.GetMeta().GetLabels()[k8s_metadata.KumaWorkload]; workloadName != "" {
-		extraLabels[WorkloadAttributeKey] = workloadName
+	isZoneProxyOnly := proxy.Dataplane.Spec.GetNetworking().IsZoneProxyOnly()
+	extraLabels[ProxyRoleAttributeKey] = deriveProxyRole(proxy.Dataplane.Spec.GetNetworking())
```
slonka, line 357: In Prometheus's data model, adding a label creates a new time series. For alerts and dashboards that preserve the current label set, that can reset `for` state and split history across the rollout window. Should this be mentioned in upgrade.md and be treated as an explicit metrics-contract change?

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -308,15 +361,33 @@ func createDynamicConfig(
...
 	if !unified_naming.Enabled(proxy.Metadata, mesh) {
 		maps.Copy(extraLabels, mads.DataplaneLabels(proxy.Dataplane, gateways))
 		extraLabels["dataplane"] = proxy.Dataplane.GetMeta().GetName()
 		if extraLabels[WorkloadAttributeKey] == "" {
-			extraLabels["service"] = proxy.Dataplane.IdentifyingName(inboundTagsDisabled)
+			service := proxy.Dataplane.IdentifyingName(inboundTagsDisabled)
+			// A zone-proxy-only Dataplane has no service tag, so IdentifyingName falls back to "unknown".
+			// Use the Dataplane name instead, which is stable and uniquely identifies the proxy.
+			if service == mesh_proto.ServiceUnknown && isZoneProxyOnly {
+				service = proxy.Dataplane.GetMeta().GetName()
```
slonka: on Kubernetes Dataplane names come from Pod names right? won't this be high cardinality?

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -409,3 +480,25 @@ func backendNameFrom(endpoint string) string {
...
+	if proxy.Dataplane == nil {
+		return conf
+	}
+	if !proxy.Dataplane.Spec.GetNetworking().IsZoneProxyOnly() {
+		return conf
+	}
+	if len(pointer.Deref(conf.Applications)) == 0 {
+		return conf
+	}
+	log.Info("ignoring 'applications' on zone-proxy-only Dataplane; field has no effect (MADR-103 group 4)",
```
slonka: 1. Don't make the same mistake as I did - log inside `Apply()` - that path runs on every xDS recompute so it will be very noisy
2. Does "MADR-103 group 4" mean anything to an average operator? Do need more context?

pkg/plugins/policies/meshmetric/api/v1alpha1/meshmetric.go
```diff
@@ -22,7 +22,8 @@ type MeshMetric struct {
 type Conf struct {
 	// Sidecar metrics collection configuration
 	Sidecar *Sidecar `json:"sidecar,omitempty"`
-	// Applications is a list of application that Dataplane Proxy will scrape
+	// Applications is a list of application that Dataplane Proxy will scrape.
```
bartsmykla: Tiny grammar fix, since this gets copied into generated docs too.

```suggestion
	// Applications is a list of applications that Dataplane Proxy will scrape.
```

test/e2e_env/multizone/meshmetric/meshmetric.go
```diff
@@ -0,0 +1,135 @@
...
+		Eventually(func(g Gomega) {
+			payload := dynConfigJSON(g, "zone-proxy-egress")
+			// MeshMetric reached the proxy and emitted the dynconf listener.
+			g.Expect(payload).To(ContainSubstring("_kuma:dynamicconfig"))
+			// applications[] must be cleared on a zone-proxy-only DPP.
+			// The inlineString field is JSON-encoded inside the larger JSON dump,
+			// so internal quotes appear as \" in the payload string.
+			g.Expect(payload).To(ContainSubstring(`\"applications\":null`))
+			g.Expect(payload).ToNot(ContainSubstring("ignored-on-zone-proxy"))
+			// proxy_role label identifies the proxy's purpose.
```
bartsmykla: Can we make the dynconf-vs-Prometheus label spelling explicit here? Otherwise this reads like final scraped label coverage.

```suggestion
			// Dynconf uses kuma.proxy_role to identify the proxy purpose.
			// Scraped Prometheus output exposes the label as kuma_proxy_role.
```

bartsmykla (dismissed):
Approving. Two small nits inline.

pkg/plugins/policies/meshmetric/api/v1alpha1/meshmetric.go
```diff
@@ -22,7 +22,8 @@ type MeshMetric struct {
...
+	// Ignored on zone-proxy-only Dataplanes (zone ingress/egress without a co-located workload).
```
slonka: are we missing a word there?

```suggestion
	// Ignored on zone-proxy-only Dataplanes (zone ingress/egress exist without a co-located workload).
```

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
```diff
@@ -47,8 +47,56 @@ const (
...
+
+	// ProxyRole* are stable enumerated values for the kuma.proxy_role metric label.
+	// Dashboards, alerts, and downstream consumers may key off these strings, so
+	// treat any rename or value change as a breaking metric-contract change.
+	ProxyRoleSidecar     = "sidecar"
+	ProxyRoleZoneEgress  = "zone-egress"
+	ProxyRoleZoneIngress = "zone-ingress"
+	ProxyRoleZoneProxy   = "zone-proxy"
+	ProxyRoleGateway     = "gateway"
+	ProxyRoleMixed       = "mixed"
```
slonka: What's a mixed role? I don't think we ever called a proxy "mixed" so we either need a better name or document this a bit more broadly.

test/e2e_env/universal/envoyconfig/zoneproxies.go
```diff
@@ -0,0 +1,127 @@
...
+			zoneproxy.WithDpEnvs(zoneProxyDpEnvs),
+		)).
+		Setup(universal.Cluster)
+	Expect(err).ToNot(HaveOccurred())
+
+	// Wait until both zone-proxy DPPs have a usable DataplaneInsight. Without
+	// the insight, `kumactl inspect dataplane --shadow` returns 404 and the
+	// table tests would only see that error during their (shorter) polling
+	// window. Doing the wait here once amortizes the cost over all entries.
+	waitZoneProxyInspectable := func(dpp string) {
```
slonka: this seems generic enough to put in all `envoyconfig` tests - worth making reusable and moving to framework?

slonka (commented):
I think we can't have noisy info in `apply`. Happy to hear your thoughts about others.

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
slonka: If ignoring `applications` on zone-proxy-only Dataplanes is intentional, this is still a pretty weak operator signal. It only emits an `Info` log from the xDS path, which can get noisy and still will not reliably tell the user why their application scrape config disappeared. I would rather reject this earlier or surface it through a stronger operator-facing signal.

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
Automaat: I think that when you have alerts or monitoring, it is on specific labels, so adding extra label to metric wont change anything in existing user setup

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
Automaat: removed

pkg/plugins/policies/meshmetric/plugin/v1alpha1/plugin.go
Automaat: removed mixed

test/e2e_env/universal/envoyconfig/zoneproxies.go
Automaat: moved

bartsmykla (approved):

PR #16599, opened 2026-05-13
feat(meshproxy): configure MeshCircuitBreaker and MeshHealthCheck

### LINKED ISSUE #16557: MeshCircuitBreaker/MeshHealthCheck: apply on mesh-scoped zone proxies via `spec.to[]`
### Description

Per [MADR-103](https://github.com/kumahq/kuma/blob/6f19192fd666aaf19a4e8632274e1da52962e1f4/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md) group (2): on a mesh-scoped zone egress Dataplane, apply `MeshCircuitBreaker` and `MeshHealthCheck` with top-level `targetRef` selecting the zone proxy and `spec.to[]` selecting the destination cluster.

Legacy global-scoped Zone Egress keeps using `EgressMatchedPolicies` — both paths coexist, keyed on which kind of egress the MES is exposed through. Non-breaking.

lukidzi:
## Motivation

We want to support `MeshCircuitBreaker` and `MeshHealthCheck` to configure `MeshExternalService` on mesh-scoped zone proxies.

## Implementation information

* support configuration of MeshExternalServices for `MeshCircuitBreaker` and `MeshHealthCheck`
* added envoy config tests for mesh-scoped zone proxies
* fixed inspect api to stub MeshIdentity since the config was empty without it

## Supporting documentation

Fix #https://github.com/kumahq/kuma/issues/16557


pkg/api-server/resource_endpoints.go
```diff
@@ -1355,11 +1355,22 @@ func (r *resourceEndpoints) rulesForResource() restful.RouteFunction {
 				}
 			}
 
-			getInboundPortName := func(port uint32) *string {
-				if name := dp.Spec.GetNetworking().GetInboundForPort(port).GetName(); name != "" {
-					return &name
+			inboundInfoForPort := func(port uint32) (*string, map[string]string) {
```
lobkovilya: I think you can just revert `resource_endpoints.go` so that:
1. we don't have conflicts with https://github.com/kumahq/kuma/pull/16575
2. the fix in my PR is smaller, and I think it’s better because we don’t really care about _rules and FromRules specifically for new listeners. We just want a minimal fix that prevents the panic

pkg/core/xds/inspect/config.go
```diff
@@ -82,6 +92,21 @@ func (p *ProxyConfigInspector) mesh() string {
 	return p.meshContext.Resource.GetMeta().GetName()
 }
 
+func dummyIdentity(identity *meshidentity_api.MeshIdentityResource) *core_xds.WorkloadIdentity {
```
lobkovilya, line 95: If possible, I’d prefer that we build a real `core_xds.WorkloadIdentity` here instead of using a dummy one. If the shadow functionality drifts too far from the real config, it gradually becomes pointless.

pkg/core/xds/inspect/config.go
```diff
@@ -61,6 +67,10 @@ func (p *ProxyConfigInspector) Get(ctx context.Context, name string, shadow bool
 		return nil, err
 	}
 
+	if identity, _ := meshidentity_api.BestMatched(proxy.Dataplane.GetMeta().GetLabels(), p.meshContext.Resources.MeshIdentities().Items); identity != nil && identity.Status != nil && identity.Status.IsInitialized() {
+		proxy.WorkloadIdentity = dummyIdentity(identity)
```
slonka, line 71: why do we need a dummy identity? can't we use GetWorkloadIdentity()? won't this just be not true for externally managed identities?

If we can't we should at least document why we need dummyIdentity

test/e2e_env/universal/envoyconfig/testdata/zoneproxy/meshcircuitbreaker/outbound.input.yaml
```diff
@@ -0,0 +1,43 @@
+type: MeshCircuitBreaker
+name: mcb-zone-proxy
+mesh: zoneproxy-envoyconfig
+labels:
+  kuma.io/effect: shadow
+spec:
+  targetRef:
+    kind: Dataplane
+    labels:
+      kuma.io/workload: zone-proxy-egress
```
lobkovilya: could you please use `kuma.io/listener-zoneegress: enabled` instead?

pkg/plugins/policies/meshhealthcheck/plugin/v1alpha1/plugin_test.go
```diff
@@ -744,4 +746,88 @@ var _ = Describe("MeshHealthCheck", func() {
...
+				).
+				Build()
+
+			p := plugin.NewPlugin().(core_plugins.PolicyPlugin)
+			Expect(p.Apply(resourceSet, xdsCtx, proxy)).To(Succeed())
+
+			Expect(util_proto.ToYAML(resourceSet.ListOf(envoy_resource.ClusterType)[0].Resource)).
+				To(test_matchers.MatchGoldenYAML(filepath.Join("testdata", given.expectedCluster)))
+		},
+		Entry("MeshExternalService cluster with TCP health check on zone proxy", zoneProxyTestCase{
```
slonka, line 806: should we add other protocols tests as well?

pkg/plugins/policies/meshhealthcheck/plugin/v1alpha1/plugin.go
```diff
@@ -61,10 +61,16 @@ func (p plugin) Apply(rs *core_xds.ResourceSet, ctx xds_context.Context, proxy *
 		return err
 	}
 
-	if err := applyToRealResources(rs, policies.ToRules.ResourceRules, ctx.Mesh, proxy.Dataplane.Spec.TagSet()); err != nil {
+	if err := applyToRealResources(rs, policies.ToRules.ResourceRules, ctx.Mesh, proxy.Dataplane.Spec.TagSet(), core_xds.NonMeshExternalService); err != nil {
```
lobkovilya: I think we shouldn't filter out MES anymore, policy like

```yaml
type: MeshCircuitBreaker
spec:
  targetRef:
    kind: Dataplane
    labels:
      app: my-app
  to:
    - targetRef:
        kind: MeshExternalService
        labels:
          kuma.io/display-name: httpbin
      default: $conf1
```

should be applied on the sidecar's outbound.

pkg/plugins/policies/meshcircuitbreaker/plugin/v1alpha1/plugin.go
```diff
@@ -74,10 +74,16 @@ func (p plugin) Apply(
 		return err
 	}
 
-	if err := applyToRealResources(ctx.Mesh, rs, policies.ToRules.ResourceRules); err != nil {
+	if err := applyToRealResources(ctx.Mesh, rs, policies.ToRules.ResourceRules, core_xds.NonMeshExternalService); err != nil {
```
lobkovilya: same

pkg/plugins/policies/meshhealthcheck/plugin/v1alpha1/plugin_test.go
lukidzi: Not sure if it's required. Whole logic is the same it's more about targeting

pkg/core/xds/inspect/config.go
lukidzi: The problem is that each call would create a new WorkloadIdentity, which drifts from real used in the config. I would prefer avoid calling CreateIdentity since cert generation is not cheap

pkg/core/xds/inspect/config.go
lukidzi: The problem is that each call would create a new WorkloadIdentity, which drifts from real used in the config. I would prefer avoid calling CreateIdentity since cert generation is not cheap

pkg/core/xds/inspect/config.go
slonka: Can't we split `GetWorkloadIdentity` operation into two? `GetWorkloadIdentity` would just get that from store / memory / whatever and we would have another operation called `CreateWorkloadIdentity`?

test/e2e_env/universal/envoyconfig/zoneproxies.go
slonka, line 114: what is "kuma.io/access" ?

pkg/core/xds/inspect/config.go
lukidzi: We don't store it in store, we store it in proxy object

test/e2e_env/universal/envoyconfig/zoneproxies.go
lukidzi: Just for selection with policy

slonka (approved):

pkg/core/xds/inspect/config.go
slonka: we talked offline, explanation will be in the follow up PR

PR #16600, opened 2026-05-13
feat(meshtrace): support zone proxy Dataplanes

Automaat:
## Motivation

Implements [MADR-103](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md)
group (4): `MeshTrace` generation on mesh-scoped zone proxy Dataplanes
(Dataplanes carrying `networking.listeners[]`).

Closes #16596.

## Implementation information

* `applyToZoneProxyListeners` walks embedded zone proxy listeners by
  contextual name (`self_zoneingress_dp_<port>` /
  `self_zoneegress_dp_<port>` from `naming.ContextualZone...ListenerName`)
  and feeds each filter chain to the existing `configureListener` path.
  HCM filter chains (HTTP-based MeshExternalService on zone egress) get
  the tracing provider injected; TCP-only filter chains (zone ingress,
  TCP MES on zone egress) are a silent no-op via the configurer's
  `UpdateHTTPConnectionManager` (it returns early when the HCM filter
  is absent).
* On a zone-proxy-only DPP, `IdentifyingName` falls back to `unknown`;
  use `Dataplane.Name` instead so Datadog `service` tag and OTel
  `service.name` carry a stable identifier. Mirrors the same fallback
  used by `MeshMetric`.

Unbroken paths:
- Regular Dataplanes (no embedded listeners) are unaffected — the new
  function short-circuits via `HasZoneProxyListeners()`.
- Legacy global-scoped ZoneEgressProxy is unchanged; that path runs
  through `proxy.ZoneEgressProxy`, not `proxy.Dataplane`.

Test coverage: nine new entries (zone-egress-only, zone-ingress-only,
mixed inbound+zone-egress × Zipkin/Datadog/OTel) with 18 new golden
files. The mixed entry confirms that both regular inbound HCM and zone
egress HCM get tracing on the same DPP.

## Supporting documentation

- MADR-103: `docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md`
- MADR-095 (zone proxy as Dataplane): `docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md`
- Matchers wiring (`feat(matchers): include zone proxy listeners when matching DPP policies`, #16584)

> Changelog: feat(meshtrace): support zone proxy Dataplanes

PR #16601, opened 2026-05-13
feat(meshtrace): support zone proxy Dataplanes

### LINKED ISSUE #16596: MeshTrace: support mesh-scoped zone proxy Dataplanes
### Description

Per [MADR-103](https://github.com/kumahq/kuma/blob/6f19192fd666aaf19a4e8632274e1da52962e1f4/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md) group (4): implement `MeshTrace` generation on mesh-scoped zone proxies.

Automaat:
## Motivation

Implements [MADR-103](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md)
group (4): `MeshTrace` generation on mesh-scoped zone proxy Dataplanes
(Dataplanes carrying `networking.listeners[]`).

Recreated from fork. Supersedes #16600.

Closes #16596.

## Implementation information

* `applyToZoneProxyListeners` walks embedded zone proxy listeners by
  contextual name (`self_zoneingress_dp_<port>` /
  `self_zoneegress_dp_<port>` from `naming.ContextualZone...ListenerName`)
  and feeds each filter chain to the existing `configureListener` path.
  HCM filter chains (HTTP-based MeshExternalService on zone egress) get
  the tracing provider injected; TCP-only filter chains (zone ingress,
  TCP MES on zone egress) are a silent no-op via the configurer's
  `UpdateHTTPConnectionManager` (it returns early when the HCM filter
  is absent).
* On a zone-proxy-only DPP, `IdentifyingName` falls back to `unknown`;
  use `Dataplane.Name` instead so Datadog `service` tag and OTel
  `service.name` carry a stable identifier. Mirrors the same fallback
  used by `MeshMetric`.
* `configureListener` now takes an explicit `direction` parameter.
  Zone-proxy callers pass `TrafficDirection_UNSPECIFIED` so Datadog
  `SplitService` skips the misleading `_INBOUND` suffix on zone-egress
  spans — the wire-level INBOUND tag on a zone-egress listener does not
  match the cross-zone hop the span represents.

Unbroken paths:
- Regular Dataplanes (no embedded listeners) are unaffected — the new
  function short-circuits via `HasZoneProxyListeners()`.
- Legacy global-scoped ZoneEgressProxy is unchanged; that path runs
  through `proxy.ZoneEgressProxy`, not `proxy.Dataplane`.

Test coverage: eleven entries (zone-egress-only, zone-ingress-only,
mixed inbound+zone-egress × Zipkin/Datadog/OTel, missing xDS resource,
`InboundTagsDisabled=true`). The mixed entry confirms that both regular
inbound HCM and zone egress HCM get tracing on the same DPP.

## Supporting documentation

- MADR-103: `docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md`
- MADR-095 (zone proxy as Dataplane): `docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md`
- Matchers wiring (`feat(matchers): include zone proxy listeners when matching DPP policies`, #16584)


bartsmykla (commented):
One scope issue needs another pass.

pkg/plugins/policies/meshtrace/plugin/v1alpha1/plugin.go
```diff
@@ -62,6 +68,9 @@ func (p plugin) Apply(rs *xds.ResourceSet, ctx xds_context.Context, proxy *xds.P
 	if err := applyToOutbounds(ctx, policies.SingleItemRules, listeners.Outbound, proxy); err != nil {
 		return err
 	}
+	if err := applyToZoneProxyListeners(ctx, policies.SingleItemRules, rs, proxy); err != nil {
```
bartsmykla: SectionName gets dropped here. The matcher can select just `ze-port`, but this call uses the proxy-wide `SingleItemRules`, so a policy for one embedded listener still configures regular inbounds and the other zone listeners. Please carry the selected listener into this path.

```suggestion
	if err := applyToZoneProxyListeners(ctx, policies, rs, proxy); err != nil {
```

Then make the helper apply listener-scoped rules instead of reusing the proxy-wide rule.

pkg/plugins/policies/meshtrace/plugin/v1alpha1/plugin_test.go
```diff
@@ -772,3 +774,396 @@ func getResourceYaml(list core_xds.ResourceList) []byte {
...
+				ctxBuilder = ctxBuilder.With(func(c *xds_context.Context) {
+					c.ControlPlane.InboundTagsDisabled = true
+				})
+			}
+			xdsCtx := *ctxBuilder.Build()
+
+			proxy := xds_builders.Proxy().
+				WithDataplane(given.dp).
+				WithMetadata(&core_xds.DataplaneMetadata{}).
+				WithPolicies(xds_builders.MatchedPolicies().
```
bartsmykla, line 920: These tests bypass the matcher path that has the bug. Please add one case that builds policies from an actual `MeshTrace` with `targetRef.sectionName: ze-port` on a mixed DPP and asserts the regular inbound stays untouched.

```yaml
targetRef:
  kind: Dataplane
  sectionName: ze-port
```

bartsmykla (approved):

PR #16605, opened 2026-05-13
fix(xds): enable reuse_port on inbound:passthrough listener

lahabana:
The :15006 inbound passthrough listener was running with
enable_reuse_port=false, so a single kernel LISTEN socket was dup'd
across each worker's epoll instance. We've seen long-running pods
wedge in this configuration: the fd stays in LISTEN (ss confirms)
and iptables packet counters keep advancing, but no Envoy worker
calls accept(). The wire signature is SYN -> SYN-ACK -> immediate
RST (kernel accept-queue overflow with tcp_abort_on_overflow=1)
while every in-Envoy reject counter (downstream_listener_filter_*,
no_filter_chain_match, *extension_config_missing,
downstream_cx_overload_reject) stays at zero. Recovery requires a
pod restart.

Setting enable_reuse_port=true makes each worker bind its own
SO_REUSEPORT socket - the same shape the outbound passthrough
listener already uses - removing the shared-fd state where one or
both workers can lose their epoll registration without anyone
re-attaching.

Scoped to CreateInboundPassthroughListener. Other inbound listeners
are unaffected because they have bind_to_port=false and the flag
has no effect on them.

Signed-off-by: Charly Molter <charly.molter@konghq.com>

### DIFF
--- a/pkg/plugins/policies/meshtls/plugin/v1alpha1/testdata/permissive-with-permissive-mtls.listeners.golden.yaml
+++ b/pkg/plugins/policies/meshtls/plugin/v1alpha1/testdata/permissive-with-permissive-mtls.listeners.golden.yaml
@@ -166,6 +166,7 @@ resources:
       socketAddress:
         address: 0.0.0.0
         portValue: 15001
+    enableReusePort: true
     filterChains:
     - filters:
       - name: envoy.filters.network.tcp_proxy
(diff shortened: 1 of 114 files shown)


bartsmykla (dismissed):
Verified locally on a k3d cluster with kuma 2.13.2. Stock state: `:15006` has `enable_reuse_port=false` with a single shared LISTEN fd dup'd across both workers, matching the customer's bundle. With the change applied via MeshProxyPatch: `:15006` has two distinct SO_REUSEPORT sockets per worker, same shape as `:15001` outbound. Removes the structural prerequisite for the wedge.

Automaat (approved):

PR #16608, opened 2026-05-14
perf(meshmetric): precompile selectors

Automaat:
## Motivation

`ProfileMutatorGenerator` in `app/kuma-dp/pkg/dataplane/metrics/profiles.go`
calls `selectorToFilterFunction(selector)` from inside the
`for metricFamily := range in` loop for every `Include`/`Exclude` rule
on every scrape. For `Regex` selectors this re-runs `regexp.Compile`
on each metric family per scrape.

Worst case observed on a busy sidecar (2000 metric families × 8 series,
50 user regex excludes):

| metric          | before  | after  | ratio |
|-----------------|---------|--------|-------|
| filter ns/op    | 225 ms  | 6.9 ms | 33x   |
| filter allocs   | 6.77 M  | 3      | -     |
| filter B/op     | 680 MB  | ~27 KB | -     |
| pipeline ns/op  | 239 ms  | 25.7 ms| 9x    |
| pipeline B/op   | 689 MB  | 9.7 MB | 70x   |

With Prometheus default 15 s scrape interval, the broken path burns
~1.5 % of a core just compiling the same regex over and over, on top
of garbage-collection pressure from hundreds of MB allocated per scrape.

## Implementation information

Move `selectorToFilterFunction` calls for `Profiles.Include` and
`Profiles.Exclude` out of the per-metric-family closure into the
generator's setup block, mirroring how `effectiveSelectors` is already
precomputed for `AppendProfiles`. Each regex now compiles once per
policy update instead of `N_families` times per scrape.

No behavior change for `Exact`/`Prefix`/`Contains` selectors (their
"compile" was already free). Invalid regex still falls through to
`neverSelect` (existing behavior) — just at policy-load time instead
of per-scrape.

Add `benchmark_test.go` covering:
- shapes: small (500 fams × 4 series), large (2000 × 8)
- profiles: All / Basic / None+include / All+exclude
- selector types: Regex vs Exact (control) at 1/10/50 rules
- both isolated filter path and full parse→mutate→serialize pipeline
- Prom→OTel conversion baseline (`FromPrometheusMetrics`)

Existing `Describe("Profiles")` table tests pass unchanged (7 entries
across All/Basic/None/include/exclude/mixed/merged).

## Supporting documentation

Reproduce locally:

```
go test -run=^# -bench=BenchmarkProfileMutator -benchtime=10x \
  ./app/kuma-dp/pkg/dataplane/metrics/...
```

> Changelog: perf(meshmetric): precompile include/exclude selectors


Automaat:

PR #16609, opened 2026-05-14
fix(meshmetric): basic label-selectors typo

Automaat:
## Motivation

`ProfileMutatorGenerator` in
`app/kuma-dp/pkg/dataplane/metrics/profiles.go:256` builds
`effectiveLabelsSelectors` from `effectiveSelectors` rather than from
`effectiveLabelsSelectors` itself:

```go
case string(v1alpha1.BasicProfileName):
    effectiveSelectors = append(effectiveSelectors, basicProfile...)
    effectiveLabelsSelectors = append(effectiveSelectors, basicProfileLabels...)
                                       ^^^^^^^^^^^^^^^^^^ typo
```

The Basic profile defines two distinct selector spaces:

- `basicProfile` — selectors over metric **names**
  (`Contains("rq_time")`, `Contains("bytes")`, `Prefix("envoy_server")`, …).
- `basicProfileLabels` — selectors over `EntityLabels` **values**
  (`Prefix(SystemPrefix)`, `Exact(ads_cluster)`, …) used to drop
  metrics whose `envoy_cluster_name` / `envoy_http_conn_manager_prefix`
  / `envoy_listener_address` points at an internal Kuma entity.

The typo concatenates the name-space selectors into the label-value
filter. A real user cluster whose `envoy_cluster_name` happens to
contain any basicProfile substring (e.g. `rq_time`, `bytes`,
`cx_active`, `circuit_breakers`, `ssl_handshake`) is wrongly
classified as an internal entity and the whole metric series is
silently dropped from Basic-profile scrapes.

Example values that trigger the bug today: any cluster name
containing `bytes` (very common in storage / streaming services),
`rq_time`, `cx_active`, etc.

## Implementation information

One-character fix: append to `effectiveLabelsSelectors`, not
`effectiveSelectors`. Behavior change is limited to the Basic profile
label-pass; All / None profiles already skip this block.

Regression covered by new fixture `basic_user_labels.{yaml,in,golden}`:
four series under `envoy_cluster_upstream_rq_total` with cluster
names `user_rq_time_svc`, `bytes_consumer`, `ads_cluster`,
`normal_svc`. Expected output keeps the three user clusters and drops
only `ads_cluster` (legitimate `basicProfileLabels` match).

Verified by stashing the one-line fix locally — the new test entry
fails (44 passed / 1 failed) without the fix and passes with it.

## Supporting documentation

> Changelog: fix(meshmetric): basic profile drops user metrics whose label values contain basicProfile substrings

### DIFF
--- a/app/kuma-dp/pkg/dataplane/metrics/profiles.go
+++ b/app/kuma-dp/pkg/dataplane/metrics/profiles.go
@@ -253,7 +253,7 @@ func ProfileMutatorGenerator(sidecar *v1alpha1.Sidecar) PrometheusMutator {
 				effectiveSelectors = append(effectiveSelectors, neverSelect)
 			case string(v1alpha1.BasicProfileName):
 				effectiveSelectors = append(effectiveSelectors, basicProfile...)
-				effectiveLabelsSelectors = append(effectiveSelectors, basicProfileLabels...)
+				effectiveLabelsSelectors = append(effectiveLabelsSelectors, basicProfileLabels...)
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


Automaat:

PR #16610, opened 2026-05-14
perf(meshmetric): reduce merge scrape allocs

Automaat:
## Motivation

After the regex-precompile fix (#16608) the next hottest spots on the
kuma-dp Prometheus scrape path are `MergeClustersForPrometheus`
(registered as the default mutator in `run.go`) and the trailing
buffer post-processing in `Hijacker.ServeHTTP`. Measured on a busy
sidecar (2000 metric families × 8 series ≈ 16k data points) every
15s scrape spends ~16ms in merge and ~1.7ms in newline-dedup, the
latter allocating ~8 MB.

## Implementation information

Six small wins, all in `app/kuma-dp/pkg/dataplane/metrics/{merge,server}.go`:

1. `mergeDuplicates` called `hash(metric)` **twice** per metric
   (line 119 today). Store it in a local.
2. `hash()` built one `fmt.Sprintf("%s=%s", …)` string per label,
   then `sort.Strings` + `strings.Join`. Replace with
   `strings.Builder` pre-sized from the summed label lengths.
3. `mergeDuplicates` returned an unsized result slice and used an
   unsized inner map. Preallocate from `len(metrics)` / `len(hashes)`.
4. `markAsMeshTraffic` copied every metric pointer into a freshly
   appended slice even though it only mutates pointers in place.
   Drop the copy, return the input slice.
5. `processNewlineChars` started from `var deduped []byte` and grew
   by single-byte append. For a 1.6 MB body that amplifies to
   ~8.4 MB allocated (slice doubling, 35 allocations). Preallocate
   to `len(input)`.
6. `processMetrics` ran `bytes.ReplaceAll(body, "# EOF", "")` over
   every app body. `# EOF` is the OpenMetrics terminator and only
   appears at the **end** of a well-formed body. Replace with
   `bytes.TrimSuffix` (and a `TrimRight('\n')` to absorb the
   optional trailing newline before/after the marker).

Behavior: identical end output. Existing `Describe("Merge")` and
`Describe("Profiles")` table tests pass unchanged.

## Supporting documentation

Local benchmark, Apple M4 Max, 20 iterations:

| scenario               | before    | after     | delta              |
|------------------------|-----------|-----------|--------------------|
| Merge small_2k         | 1.80 ms   | 1.60 ms   | -11% / -10k allocs |
| Merge large_16k        | 15.85 ms  | 12.99 ms  | -18% / -76k allocs |
| NewlineDedup large_16k | 1.72 ms   | 1.30 ms   | -24%               |
| NewlineDedup B/op      | 8.40 MB   | 1.44 MB   | -83%               |

> Changelog: perf(meshmetric): reduce per-scrape allocations on the merge and newline-dedup paths

### DIFF
--- a/app/kuma-dp/pkg/dataplane/metrics/merge.go
+++ b/app/kuma-dp/pkg/dataplane/metrics/merge.go
@@ -1,7 +1,6 @@
 package metrics
 
 import (
-	"fmt"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


Automaat:

PR #16614, opened 2026-05-14
chore(log): demote chatty events to V(1)

bartsmykla:
## Motivation

On a busy CP, these log call sites fire on every xDS stream message and every KDS health-check tick. They dominate log volume and crowd out the events on-call actually needs to see. The noise grows linearly with mesh size.

The xDS change (`dataplane_status_tracker.go`) accounts for the bulk of the volume reduction - it fires on every resource request from every connected proxy. The pod controller changes are comparatively minor.

## Implementation information

- `dataplane_status_tracker.go`: demote "config requested" to V(1) - fires on every xDS request, not just first contact. Unify `streamid` -> `streamID` field key consistently across the xDS/KDS/HDS callback family. Keep "proxy connected" and "proxy disconnected" at Info (lifecycle events).
- `kds/mux/client.go`: demote "sending health check" to V(1) - fires every 5 minutes per zone CP, adds nothing on a healthy mesh.
- `pod_controller.go`: keep "Object deleted" at Info (no K8s Event fallback for delete; create/update emit Events so their log can be demoted). Demote "Dataplane created" and "Dataplane updated" to V(1). Rate-sample "failed to parse Dataplane" (1-minute window per resource name) so a single malformed resource cannot flood across repeated reconcile passes; includes `suppressed_count` in the next allowed log. Field renamed `dataplane` -> `dataplane_ref` to make the shape change explicit (was full Spec, now namespace/name string).
- `kds/v2/client/kds_client.go`: no change needed, already demoted in a prior PR

### DIFF
--- a/pkg/kds/mux/client.go
+++ b/pkg/kds/mux/client.go
@@ -312,7 +312,7 @@ func (c *client) startHealthCheck(
 	ticker := time.NewTicker(prevInterval)
 	defer ticker.Stop()
 	for {
-		log.Info("sending health check")
+		log.V(1).Info("sending health check")
 		resp, err := client.HealthCheck(ctx, &mesh_proto.ZoneHealthCheckRequest{})
 		if err != nil && !errors.Is(err, context.Canceled) {
 			if status.Code(err) == codes.Unimplemented {
--- a/pkg/plugins/runtime/k8s/controllers/pod_controller.go
+++ b/pkg/plugins/runtime/k8s/controllers/pod_controller.go
@@ -51,37 +51,50 @@ const (
 	ZoneProxyListenersSkippedReason = "ZoneProxyListenersSkipped"
 )
 
+// errSamplerWindow is the dedup window for parse-error log sampling.
+const errSamplerWindow = time.Minute
+
 // parseSampler rate-limits repeated "failed to parse Dataplane" errors so a
 // single malformed resource does not flood the log on every reconcile.
-var parseSampler = newErrSampler(time.Minute)
+var parseSampler = newErrSampler()
 
-// errSampler allows at most one Error log per unique key per window.
+// errSampler allows at most one Error log per unique key per errSamplerWindow.
 // It prevents tight loops from flooding the log when the same parse error
 // repeats for the same resource across reconcile passes.
+// The key should be a stable identifier (e.g. namespace/name), not include
+// variable error text - that avoids cardinality explosions from embedded
+// resource names in error messages.
 type errSampler struct {
-	window time.Duration
-	mu     sync.Mutex
-	last   map[string]time.Time
+	mu         sync.Mutex
+	last       map[string]time.Time
+	suppressed map[string]int
 }
 
-func newErrSampler(window time.Duration) *errSampler {
-	return &errSampler{window: window, last: map[string]time.Time{}}
+func newErrSampler() *errSampler {
+	return &errSampler{
+		last:       map[string]time.Time{},
+		suppressed: map[string]int{},
+	}
 }
 
-func (s *errSampler) allow(key string) bool {
+// allow reports whether the caller should log for key.
+// The second return value is the number of calls suppressed since the last
+// allowed log (zero on first occurrence in a window).
+func (s *errSampler) allow(key string) (bool, int) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	now := time.Now()
-	for existingKey, t := range s.last {
-		if now.Sub(t) >= s.window {
-			delete(s.last, existingKey)
+	if t, ok := s.last[key]; ok {
+		if now.Sub(t) < errSamplerWindow {
+			s.suppressed[key]++
+			return false, 0
 		}
+		// window expired - fall through and re-allow
 	}
-	if _, ok := s.last[key]; ok {
-		return false
-	}
+	suppressed := s.suppressed[key]
 	s.last[key] = now
-	return true
+	s.suppressed[key] = 0
+	return true, suppressed
 }
 
 // PodReconciler reconciles a Pod object
@@ -206,7 +219,7 @@ func (r *PodReconciler) deleteObjectIfExist(ctx context.Context, object k8s_mode
 		}
 		return errors.Wrapf(err, "could not delete %v %s/%s", object.GetObjectKind(), object.GetName(), object.GetNamespace())
 	}
-	log.V(1).Info("Object deleted")
+	log.Info("Object deleted")
 	return nil
 }
 
@@ -343,8 +356,8 @@ func (r *PodReconciler) findOtherDataplanes(ctx context.Context, pod *kube_core.
 		dp := core_mesh.NewDataplaneResource()
 		if err := r.ResourceConverter.ToCoreResource(&dataplane, dp); err != nil {
 			name := kube_types.NamespacedName{Namespace: dataplane.Namespace, Name: dataplane.Name}
-			if parseSampler.allow(name.String() + ":" + err.Error()) {
-				converterLog.Error(err, "failed to parse Dataplane", "dataplane", name, "spec", dataplane.Spec)
+			if ok, suppressed := parseSampler.allow(name.String()); ok {
+				converterLog.Error(err, "failed to parse Dataplane", "dataplane_ref", name.String(), "suppressed_count", suppressed)
 			}
 			continue // one invalid Dataplane definition should not break the entire mesh
 		}
--- a/pkg/plugins/runtime/k8s/controllers/pod_controller.go
+++ b/pkg/plugins/runtime/k8s/controllers/pod_controller.go
@@ -2,6 +2,8 @@ package controllers
 
 import (
 	"context"
+	"sync"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
@@ -32,6 +34,35 @@ import (
 	util_k8s "github.com/kumahq/kuma/v2/pkg/plugins/runtime/k8s/util"
 	util_maps "github.com/kumahq/kuma/v2/pkg/util/maps"
 )
+
+// errSampler allows at most one Error log per unique message per window.
+// It prevents tight loops from flooding the log when the same parse error
+// repeats for many resources in a single reconcile pass.
+type errSampler struct {
+	window time.Duration
+	mu     sync.Mutex
+	last   map[string]time.Time
+}
+
+func newErrSampler(window time.Duration) *errSampler {
+	return &errSampler{window: window, last: map[string]time.Time{}}
+}
+
+// allow returns true the first time msg is seen and once per window thereafter.
+func (s *errSampler) allow(msg string) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	now := time.Now()
+	if t, ok := s.last[msg]; ok && now.Sub(t) < s.window {
+		return false
+	}
+	s.last[msg] = now
+	return true
+}
+
+// parseSampler rate-limits repeated "failed to parse Dataplane" errors so a
+// single malformed resource does not flood the log on every reconcile.
+var parseSampler = newErrSampler(time.Minute)
 
 const (
 	// CreatedKumaDataplaneReason is added to an event when
@@ -171,7 +202,7 @@ func (r *PodReconciler) deleteObjectIfExist(ctx context.Context, object k8s_mode
 		}
 		return errors.Wrapf(err, "could not delete %v %s/%s", object.GetObjectKind(), object.GetName(), object.GetNamespace())
 	}
-	log.Info("Object deleted")
+	log.V(1).Info("Object deleted")
 	return nil
 }
 
@@ -307,7 +338,9 @@ func (r *PodReconciler) findOtherDataplanes(ctx context.Context, pod *kube_core.
 		dataplane := allDataplanes.Items[i]
 		dp := core_mesh.NewDataplaneResource()
 		if err := r.ResourceConverter.ToCoreResource(&dataplane, dp); err != nil {
-			converterLog.Error(err, "failed to parse Dataplane", "dataplane", dataplane.Spec)
+			if parseSampler.allow(err.Error()) {
+				converterLog.Error(err, "failed to parse Dataplane", "dataplane", dataplane.Spec)
+			}
 			continue // one invalid Dataplane definition should not break the entire mesh
 		}
 		if dataplane.Mesh == mesh {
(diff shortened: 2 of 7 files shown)


pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -171,7 +206,7 @@ func (r *PodReconciler) deleteObjectIfExist(ctx context.Context, object k8s_mode
 		}
 		return errors.Wrapf(err, "could not delete %v %s/%s", object.GetObjectKind(), object.GetName(), object.GetNamespace())
 	}
-	log.Info("Object deleted")
+	log.V(1).Info("Object deleted")
```
Automaat: **issue:** "Object deleted" loses default-verbosity visibility, and unlike create/update the delete path emits no K8s Event. After this demotion an operator gets nothing in the CP log and nothing in `kubectl describe pod` when a Dataplane silently disappears. The `cause` ("pod succeeded" / "pod IP is empty" / "pod was evicted") is the single most useful piece of incident context. Either keep this one at Info, or emit a `kube_core.EventTypeNormal` "DeletedKumaDataplane" before returning.

pkg/xds/server/callbacks/dataplane_status_tracker.go
```diff
@@ -178,12 +178,12 @@ func (c *dataplaneStatusTracker) onStreamRequest(
 			if md.GetVersion() != nil {
 				state.subscription.Version = md.GetVersion()
 			} else {
-				statusTrackerLog.Error(err, "failed to extract version out of the Envoy metadata", "streamid", streamID, "mode", mode, "metadata", req.Metadata())
+				statusTrackerLog.Error(err, "failed to extract version out of the Envoy metadata", "streamID", streamID, "mode", mode, "metadata", req.Metadata())
```
Automaat, line 181: **issue:** Half-done `streamid` to `streamID` rename. The canonical key is `StreamIDLoggerKey = "streamID"` in `pkg/log/logger.go:175`, so the direction is right, but lowercase `streamid` still lives in `pkg/util/xds/logging_callbacks.go`, `pkg/kds/status/status_tracker.go`, and `pkg/hds/tracker/callbacks.go`. A single stream now emits both keys across files; aggregator queries on either return partial truth, which is worse than the prior consistent-wrong state. Finish the rename in this PR, or pull these two hunks out and land them with the rest.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -307,7 +342,10 @@ func (r *PodReconciler) findOtherDataplanes(ctx context.Context, pod *kube_core.
 		dataplane := allDataplanes.Items[i]
 		dp := core_mesh.NewDataplaneResource()
 		if err := r.ResourceConverter.ToCoreResource(&dataplane, dp); err != nil {
-			converterLog.Error(err, "failed to parse Dataplane", "dataplane", dataplane.Spec)
+			name := kube_types.NamespacedName{Namespace: dataplane.Namespace, Name: dataplane.Name}
+			if parseSampler.allow(name.String() + ":" + err.Error()) {
```
Automaat: **issue:** Sampler key includes the full `err.Error()` string. Wrapped errors carry pointers, byte offsets, line and column positions, and other variable text. A flapping resource with varying error formatter output bypasses the sampler and re-logs every reconcile, while inflating the map. Key by `name.String()` alone (matches the stated goal "one malformed resource shouldn't flood") and include `err.Error()` only in the log payload.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -49,6 +51,39 @@ const (
...
+
+func newErrSampler(window time.Duration) *errSampler {
+	return &errSampler{window: window, last: map[string]time.Time{}}
+}
+
+func (s *errSampler) allow(key string) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	now := time.Now()
+	for existingKey, t := range s.last {
```
Automaat: **issue:** Full map sweep runs under the mutex on every `allow()` call, even when no entries are expired. With M unique keys in the window this is O(M) under a global lock. Every reconcile in the controller-runtime worker pool (default 10) gates on it during exactly the incident you want the CP responsive. Track a `nextSweep` timestamp and only iterate when `now.After(nextSweep)`, or cap the map size and drop oldest on insert.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
Automaat: **issue:** Suppressed errors are invisible. No counter, no "N dropped" log when the window rolls. If a release ships a schema change that breaks parsing for half the fleet, an operator sees one error per name per minute and no signal of scale. The log is for narrative; the metric is for alerting. Add a Prometheus counter (`kuma_cp_dataplane_parse_errors_total{namespace,name}`) incremented on every parse failure regardless of sampling.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -307,7 +342,10 @@ func (r *PodReconciler) findOtherDataplanes(ctx context.Context, pod *kube_core.
 		dataplane := allDataplanes.Items[i]
 		dp := core_mesh.NewDataplaneResource()
 		if err := r.ResourceConverter.ToCoreResource(&dataplane, dp); err != nil {
-			converterLog.Error(err, "failed to parse Dataplane", "dataplane", dataplane.Spec)
+			name := kube_types.NamespacedName{Namespace: dataplane.Namespace, Name: dataplane.Name}
+			if parseSampler.allow(name.String() + ":" + err.Error()) {
+				converterLog.Error(err, "failed to parse Dataplane", "dataplane", name, "spec", dataplane.Spec)
```
Automaat: **issue:** Silent shape change on the "failed to parse Dataplane" Error log. The `"dataplane"` field's type changed from `dataplane.Spec` (full proto) to a `NamespacedName`; a new `"spec"` field now carries the proto. Anyone with `dataplane.networking.address:*` queries in their log aggregator gets nil after upgrade. The new shape is better, but it is a silent breaking change on an Error log, and combined with the new rate-limiter, operators get fewer events with a different shape. Needs an UPGRADE.md or CHANGELOG line.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -49,6 +51,39 @@ const (
 	ZoneProxyListenersSkippedReason = "ZoneProxyListenersSkipped"
 )
 
+// parseSampler rate-limits repeated "failed to parse Dataplane" errors so a
```
Automaat, line 57: **suggestion:** Delete the whole sampler. The loop is bounded by `len(allDataplanes.Items)`, reconciles are not a hot path, and MADR 003 already defines V(1) as the home for chatty events. `converterLog.V(1).Info("failed to parse Dataplane", ...)` accomplishes the stated goal in 4 lines instead of 90 (sampler type + constructor + method + test file). The sampler is a small framework for a problem V(1) solves for free. If parse errors at default verbosity matter, an error metric is the right primitive, not a log throttle.

Automaat, line 57: **suggestion:** `golang.org/x/time/rate` is already in `go.mod` (indirect). `rate.Sometimes{Interval: time.Minute}` does exactly what `errSampler.allow` does, in stdlib-style code, per-key via `sync.Map[string]*rate.Sometimes`. If a sampler stays, reuse this one; if you reject it, say why in the PR description.

pkg/kds/mux/client.go
```diff
@@ -312,7 +312,7 @@ func (c *client) startHealthCheck(
 	ticker := time.NewTicker(prevInterval)
 	defer ticker.Stop()
 	for {
-		log.Info("sending health check")
+		log.V(1).Info("sending health check")
```
Automaat, line 315: **suggestion:** Demoting the per-tick KDS health check is correct, but transitions deserve Info. Today both healthy and unhealthy ticks share the same log path, so slow degradation is invisible. Only "connection dropped" surfaces. Add edge-triggered Info on first failure after success and first success after failure ("zone health check failing" / "recovered").

pkg/xds/server/callbacks/dataplane_status_tracker.go
```diff
@@ -221,7 +221,7 @@ func (c *dataplaneStatusTracker) onStreamRequest(
 		if !statusTrackerLog.V(1).Enabled() { // it was already added, no need to add it twice
 			log = log.WithValues("resourceNames", req.GetResourceNames())
 		}
-		log.Info("config requested")
+		log.V(1).Info("config requested")
```
Automaat, line 224: **suggestion:** Per-request "config requested" at V(1) is right. Add Info on first request per stream and on stream teardown. The stream lifecycle is what operators correlate during xDS contention, not the per-NACK chatter.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
Automaat, line 57: **suggestion:** If the sampler stays, move it to `pkg/util/log` (or similar) and rename it. `errSampler` and `allow` read as a permit or rate-limit primitive, not a logging predicate. `parseErrorLogOnce.shouldLog(key)` at the call site says what it does. Putting a general utility in `pod_controller.go` guarantees the second caller copies it.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -371,10 +409,10 @@ func (r *PodReconciler) createOrUpdateDataplane(
 	}
 	switch operationResult {
 	case kube_controllerutil.OperationResultCreated:
-		log.Info("Dataplane created")
+		log.V(1).Info("Dataplane created")
```
Automaat, line 425: **praise:** V(1) sites are all alloc-safe. Verified across all 12 demoted call sites: every key-value pair is a literal, an int, or a pre-bound variable. No `expensive()` calls hiding in deferred argument evaluation that would defeat the `Enabled()` short-circuit. Good discipline.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
```diff
@@ -371,10 +409,10 @@ func (r *PodReconciler) createOrUpdateDataplane(
...
 		r.Eventf(pod, nil, kube_core.EventTypeNormal, CreatedKumaDataplaneReason, "Create", "Created Kuma Dataplane: %s", pod.Name)
```
Automaat, line 426: **praise:** K8s Events for create/update stay at `EventTypeNormal`. `kubectl describe pod`, event exporters, and event-based Prometheus recorders keep full visibility for the K8s population. The right escape hatch for the most common deployment shape.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
Automaat, line 57: **praise:** Sampler comment states intent clearly: "single malformed resource does not flood the log on every reconcile". Future readers will know why the code exists.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
bartsmykla: Reverted to Info. deleteObjectIfExist's signature doesn't carry a pod ref so emitting a K8s Event would need a wider refactor; preserving log visibility was the smaller fix.

pkg/xds/server/callbacks/dataplane_status_tracker.go
bartsmykla: Finished the rename across util/xds (9), kds/status (5), and hds/tracker (2). All four callback files now use streamID.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
bartsmykla: Keyed by name.String() only. err.Error() is in the log payload but not the dedup key.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
bartsmykla: Replaced the full-map sweep with a per-key expiry check. Map grows at most O(distinct broken resources).

pkg/plugins/runtime/k8s/controllers/pod_controller.go
bartsmykla: allow() returns (bool, int) and the Error log now carries suppressed_count. Prometheus counter (kuma_cp_dataplane_parse_errors_total) is the right next step but tracking it as a follow-up so this PR stays focused on the demotion.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
bartsmykla: Renamed the field key from dataplane to dataplane_ref so aggregator queries on the old key get nil rather than silent partial data.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
bartsmykla: Keeping the sampler. V(1) hides parse errors at default verbosity, and the loop fires per Pod reconcile so a single malformed Dataplane would still spam at any visible level. Addressed your other concerns (key, sweep, count).

bartsmykla: Considered. rate.Sometimes doesn't expose the suppressed count we now need on the Error log, so keeping the custom sampler for that reason.

pkg/kds/mux/client.go
bartsmykla: Out of scope for this demotion PR. Tracking edge-triggered health-check Info as a follow-up alongside the KDS reconnect work landed in #16586.

pkg/xds/server/callbacks/dataplane_status_tracker.go
bartsmykla: Open/close lifecycle already logs in the surrounding callbacks. Tracking the first-request-per-stream marker as a follow-up.

pkg/plugins/runtime/k8s/controllers/pod_controller.go
bartsmykla: Keeping in pod_controller for now. If a second caller shows up I'll extract; until then a local helper is the smaller surface.

Automaat (approved):

PR #16616, opened 2026-05-14
feat(log): log key state transitions

bartsmykla:
## Motivation

When an operator opens an incident around CA rotation, leader election flapping, or a webhook rejecting a resource, the kuma-cp log shows nothing about the transition - just steady-state heartbeats. They cannot correlate the symptom with the actual state change.

## Implementation information

- Builtin CA manager logs `CA created` with `mesh`, `backend`, and `notAfter` after secrets are persisted. Fires on new CA creation (builtin has no separate rotate operation).
- Plugin registration emits one `Info` per registered plugin with `name` and `kind` at startup. The `pluginKind` helper reuses the unexported constants already defined in `registry.go`.
- Validating webhooks log `webhook rejected resource` with `kind`, `name`, `namespace`, `operation`, and `reason` on every rejection. Not logged on allowed responses to avoid per-request spam.
- `component.NewManager` accepts an optional `instanceId` argument. Universal CP passes `b.GetInstanceId()`. The "leader acquired" and "leader lost" logs now carry `leaderId`.

Skipped sub-items:

- Provided CA backend: `EnsureBackends` is a no-op (user manages the cert externally), no state transition to log.
- License lifecycle: `pkg/license/` does not exist in kuma OSS - license validation is enterprise-only.
- Observed leader change from non-leader pods: neither the Postgres elector nor the K8s controller-runtime elector exposes who the current leader is to non-leaders. Deferred.

### DIFF
--- a/pkg/core/plugins/global.go
+++ b/pkg/core/plugins/global.go
@@ -1,5 +1,9 @@
 package plugins
 
+import "github.com/kumahq/kuma/v2/pkg/core"
+
+var log = core.Log.WithName("plugins")
+
 var global = NewRegistry()
 
 func Plugins() Registry {
@@ -9,5 +13,32 @@ func Plugins() Registry {
 func Register(name PluginName, plugin Plugin) {
 	if err := global.Register(name, plugin); err != nil {
 		panic(err)
+	}
+	log.Info("plugin registered", "name", name, "kind", pluginKind(plugin))
+}
+
+// pluginKind returns the type string for a plugin, reusing the constants from registry.go.
+func pluginKind(p Plugin) string {
+	switch p.(type) {
+	case BootstrapPlugin:
+		return string(bootstrapPlugin)
+	case ResourceStorePlugin:
+		return string(resourceStorePlugin)
+	case SecretStorePlugin:
+		return string(secretStorePlugin)
+	case ConfigStorePlugin:
+		return string(configStorePlugin)
+	case RuntimePlugin:
+		return string(runtimePlugin)
+	case CaPlugin:
+		return string(caPlugin)
+	case AuthnAPIServerPlugin:
+		return string(authnAPIServer)
+	case PolicyPlugin:
+		return string(policyPlugin)
+	case ProxyPlugin:
+		return string(proxyPlugin)
+	default:
+		return "unknown"
 	}
 }
--- a/pkg/core/plugins/global.go
+++ b/pkg/core/plugins/global.go
@@ -17,7 +17,6 @@ func Register(name PluginName, plugin Plugin) {
 	log.Info("plugin registered", "name", name, "kind", pluginKind(plugin))
 }
 
-// pluginKind returns the type string for a plugin, reusing the constants from registry.go.
 func pluginKind(p Plugin) string {
 	switch p.(type) {
 	case BootstrapPlugin:
--- a/pkg/core/plugins/global.go
+++ b/pkg/core/plugins/global.go
@@ -6,6 +6,15 @@ var log = core.Log.WithName("plugins")
 
 var global = NewRegistry()
 
+var registeredPlugins []registeredPlugin
+
+var loggedRegisteredPlugins int
+
+type registeredPlugin struct {
+	name PluginName
+	kind string
+}
+
 func Plugins() Registry {
 	return global
 }
@@ -14,7 +23,14 @@ func Register(name PluginName, plugin Plugin) {
 	if err := global.Register(name, plugin); err != nil {
 		panic(err)
 	}
-	log.Info("plugin registered", "name", name, "kind", pluginKind(plugin))
+	registeredPlugins = append(registeredPlugins, registeredPlugin{name: name, kind: pluginKind(plugin)})
+}
+
+func LogRegistered() {
+	for _, plugin := range registeredPlugins[loggedRegisteredPlugins:] {
+		log.Info("plugin registered", "name", plugin.name, "kind", plugin.kind)
+	}
+	loggedRegisteredPlugins = len(registeredPlugins)
 }
 
 func pluginKind(p Plugin) string {
--- a/pkg/core/plugins/global.go
+++ b/pkg/core/plugins/global.go
@@ -53,6 +53,10 @@ func pluginKind(p Plugin) string {
 		return string(policyPlugin)
 	case ProxyPlugin:
 		return string(proxyPlugin)
+	case CoreResourcePlugin:
+		return string(coreResourcePlugin)
+	case IdentityProviderPlugin:
+		return string(identityProviderPlugin)
 	default:
 		return "unknown"
 	}
--- a/pkg/core/runtime/component/component.go
+++ b/pkg/core/runtime/component/component.go
@@ -98,9 +98,14 @@ type Manager interface {
 
 var _ Manager = &manager{}
 
-func NewManager(leaderElector LeaderElector) Manager {
+func NewManager(leaderElector LeaderElector, instanceId ...string) Manager {
+	id := ""
+	if len(instanceId) > 0 {
+		id = instanceId[0]
+	}
 	return &manager{
 		leaderElector: leaderElector,
+		instanceId:    id,
 		panicLimiter:  newPanicLogLimiter(10 * time.Second),
 	}
 }
@@ -141,6 +146,7 @@ func (p *panicLogLimiter) shouldLog(name string) (bool, int) {
 
 type manager struct {
 	leaderElector LeaderElector
+	instanceId    string
 	panicLimiter  *panicLogLimiter
 
 	sync.Mutex // protects access to fields below
@@ -286,7 +292,7 @@ func (cm *manager) startLeaderComponents(stop <-chan struct{}, errCh chan error)
 
 	cm.leaderElector.AddCallbacks(LeaderCallbacks{
 		OnStartedLeading: func() {
-			log.Info("leader acquired")
+			log.Info("leader acquired", "leaderId", cm.instanceId)
 			mutex.Lock()
 			defer mutex.Unlock()
 			leaderStopCh = make(chan struct{})
@@ -300,7 +306,7 @@ func (cm *manager) startLeaderComponents(stop <-chan struct{}, errCh chan error)
 			}
 		},
 		OnStoppedLeading: func() {
-			log.Info("leader lost")
+			log.Info("leader lost", "leaderId", cm.instanceId)
 			closeLeaderCh()
 		},
 	})
--- a/pkg/core/runtime/component/component.go
+++ b/pkg/core/runtime/component/component.go
@@ -292,7 +292,7 @@ func (cm *manager) startLeaderComponents(stop <-chan struct{}, errCh chan error)
 
 	cm.leaderElector.AddCallbacks(LeaderCallbacks{
 		OnStartedLeading: func() {
-			log.Info("leader acquired", "leaderId", cm.instanceId)
+			log.Info("leader acquired", "instanceId", cm.instanceId)
 			mutex.Lock()
 			defer mutex.Unlock()
 			leaderStopCh = make(chan struct{})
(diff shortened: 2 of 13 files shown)


pkg/core/plugins/global.go
```diff
@@ -10,4 +23,37 @@ func Register(name PluginName, plugin Plugin) {
 	if err := global.Register(name, plugin); err != nil {
 		panic(err)
 	}
+	registeredPlugins = append(registeredPlugins, registeredPlugin{name: name, kind: pluginKind(plugin)})
+}
+
+func LogRegistered() {
+	for _, plugin := range registeredPlugins[loggedRegisteredPlugins:] {
+		log.Info("plugin registered", "name", plugin.name, "kind", plugin.kind)
+	}
```
Automaat, line 32: **issue:** `pluginKind` duplicates the registry's classification with weaker semantics. Have `Register` record the kinds it computed (it already walks every interface) and read those back in `LogRegistered`, or hoist the type switch into a helper used by both. One source of truth per concept.

pkg/core/runtime/component/component.go
```diff
@@ -286,7 +292,7 @@ func (cm *manager) startLeaderComponents(stop <-chan struct{}, errCh chan error)
 
 	cm.leaderElector.AddCallbacks(LeaderCallbacks{
 		OnStartedLeading: func() {
-			log.Info("leader acquired")
+			log.Info("leader acquired", "leaderId", cm.instanceId)
```
Automaat: **issue:** `leaderId` (log key) and `instanceId` (constructor param, struct field, builder method) name the same thing in two vocabularies. Pick one word and use it from `b.GetInstanceId()` through the field to the log key. Operators reading "leaderId" will assume it identifies the current lease holder; the value is actually the local process identity.

pkg/plugins/common/k8s/composite_validator.go
```diff
@@ -44,3 +49,18 @@ func (c *CompositeValidator) IntoWebhook(scheme *runtime.Scheme) *admission.Webh
 		}),
 	}
 }
+
+// LogWebhookRejection records only non-allowed responses to avoid per-request spam.
```
Automaat: **suggestion:** `LogWebhookRejection` lives in `composite_validator.go` but is called from validators that do not go through the composite (pod, secret). It is no longer a composite concern; it is the shared admission-rejection logger. Move it (and `webhookLog`) to its own file (`admission_log.go`) so the file name tells the truth about ownership.

pkg/plugins/ca/builtin/manager.go
```diff
@@ -125,6 +127,14 @@ func (b *builtinCaManager) create(ctx context.Context, mesh core_model.Resource,
 		}
 		log.V(1).Info("CA secret key already exists. Nothing to create", "mesh", meshName, "backend", backend.Name)
 	}
+
```
Automaat, line 136: **issue:** `pem.Decode` and `x509.ParseCertificate` errors are silently swallowed; failure produces `notAfter=""`. An empty string in a structured log is the worst signal because it looks like data. Either log a parse failure at `V(1)` and omit the field, or pass `time.Time` and let the logger format it (omitting on zero value).

pkg/plugins/common/k8s/composite_validator.go
Automaat: **issue:** Webhook rejection log is unbounded and unsampled. `resp.Result.Message` can be a multi-line list of every violated field; a misconfigured controller in a reconciliation tight loop will produce hundreds of these per second. Use the `panicLogLimiter` pattern already in `component.go:120`, or truncate the reason to ~512 chars, or both.

Automaat: **suggestion:** Webhook rejections deserve a Prometheus counter (`kuma_webhook_rejections_total{kind,operation,result}`), not a per-rejection info log. Counters aggregate; logs do not. Demote the log to `V(1)` and emit the metric at every call.

pkg/plugins/ca/builtin/manager.go
```diff
@@ -125,6 +127,14 @@ func (b *builtinCaManager) create(ctx context.Context, mesh core_model.Resource,
...
+	var notAfter string
+	if block, _ := pem.Decode(keyPair.CertPEM); block != nil {
+		if cert, err := x509.ParseCertificate(block.Bytes); err == nil {
+			notAfter = cert.NotAfter.UTC().Format(time.RFC3339)
+		}
```
Automaat: **issue:** "CA created" semantic shifted. The log used to fire once per backend in the `EnsureBackends` success path. It now fires from inside `create()` on every call, including the AlreadyExists fall-through. Operator aggregations that count "CA created" events will inflate after upgrade. Either emit only on the actual-creation branch or rename the log message to reflect the new meaning.

pkg/plugins/common/k8s/composite_validator.go
```diff
@@ -6,8 +6,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/kumahq/kuma/v2/pkg/core"
 )
 
+var webhookLog = core.Log.WithName("webhooks")
+
```
Automaat: **issue:** `var webhookLog = core.Log.WithName("webhooks")` diverges from the universal `var log = core.Log.WithName("...")` convention used across `pkg/`. Match the idiom or rename across the codebase. One-off exceptions are how style entropy starts.

pkg/plugins/runtime/k8s/webhooks/dataplane_validator.go
```diff
@@ -63,9 +63,11 @@ func (h *DataplaneValidator) ValidateCreate(ctx context.Context, req admission.R
 
 	if err := h.validator.ValidateCreate(ctx, core_model.MetaToResourceKey(coreRes.GetMeta()), coreRes, mesh); err != nil {
 		if kumaErr, ok := err.(*validators.ValidationError); ok {
-			return convertSpecValidationError(kumaErr, false, k8sRes)
+			resp := convertSpecValidationError(kumaErr, false, k8sRes)
```
Automaat: **issue:** The `resp := convertSpecValidationError(...)` then `return resp` (and similar) pattern in dataplane / mesh / validation handlers adds a local that is never read between assignment and return. The CompositeValidator wrapping these handlers already calls `LogWebhookRejection`. This is dead refactoring, almost certainly a leftover from an earlier iteration where logging lived inside each handler. Revert these blocks to a direct return.

pkg/plugins/runtime/k8s/webhooks/mesh_validator.go
```diff
@@ -52,7 +52,8 @@ func (h *MeshValidator) Handle(ctx context.Context, req admission.Request) admis
 func (h *MeshValidator) ValidateDelete(ctx context.Context, req admission.Request) admission.Response {
 	if !h.unsafeDelete {
 		if err := h.validator.ValidateDelete(ctx, req.Name); err != nil {
-			return admission.Errored(http.StatusBadRequest, err)
+			resp := admission.Errored(http.StatusBadRequest, err)
```
Automaat: **issue:** Same dead refactor.

pkg/plugins/runtime/k8s/webhooks/validation.go
```diff
@@ -68,19 +68,23 @@ func (h *validatingHandler) Handle(_ context.Context, req admission.Request) adm
 	default:
 		var warnings []string
 		if err := core_mesh.ValidateMesh(k8sObj.GetMesh(), coreRes.Descriptor().Scope); err.HasViolations() {
-			return convertValidationErrorOf(err, k8sObj, k8sObj.GetObjectMeta())
+			resp := convertValidationErrorOf(err, k8sObj, k8sObj.GetObjectMeta())
```
Automaat: **issue:** Same dead refactor.

pkg/core/plugins/global.go
bartsmykla: Agree the duplication is real. Hoisting kind detection into the registry expands beyond this PR's scope; deferring to a follow-up where we add a Registry.Kinds(name) method and remove pluginKind.

pkg/core/runtime/component/component.go
bartsmykla: Renamed to instanceId in both leader-acquired and leader-lost logs.

pkg/plugins/common/k8s/composite_validator.go
bartsmykla: Moved to pkg/plugins/common/k8s/admission_log.go.

pkg/plugins/ca/builtin/manager.go
bartsmykla: Refactored create() to return time.Time. On parse failure it emits V(1) and the success log omits the field when zero.

pkg/plugins/common/k8s/composite_validator.go
bartsmykla: Added a 10s per-(kind,namespace,name) sampler and truncated reason to 512 chars in admission_log.go.

bartsmykla: Counter is the right long-term shape but plumbing metrics through the webhook handler is a separate change. For this PR I kept the log at Info with sampling and truncation; counter follow-up tracked separately.

pkg/plugins/ca/builtin/manager.go
bartsmykla: Moved the log back to EnsureBackends' success path so it only fires when create() actually returns. Matches the original semantics.

pkg/plugins/common/k8s/composite_validator.go
bartsmykla: Renamed to var log in the new admission_log.go.

pkg/plugins/runtime/k8s/webhooks/dataplane_validator.go
bartsmykla: Reverted to direct return.

pkg/plugins/runtime/k8s/webhooks/mesh_validator.go
bartsmykla: Same - reverted to direct return.

pkg/plugins/runtime/k8s/webhooks/validation.go
bartsmykla: Same - reverted to direct return.

pkg/core/plugins/global.go
```diff
@@ -10,4 +23,37 @@ func Register(name PluginName, plugin Plugin) {
...
+	case RuntimePlugin:
+		return string(runtimePlugin)
+	case CaPlugin:
+		return string(caPlugin)
+	case AuthnAPIServerPlugin:
+		return string(authnAPIServer)
+	case PolicyPlugin:
+		return string(policyPlugin)
+	case ProxyPlugin:
+		return string(proxyPlugin)
```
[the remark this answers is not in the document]
bartsmykla: Added CoreResourcePlugin and IdentityProviderPlugin cases. Also added coreResourcePlugin and identityProviderPlugin pluginType constants in registry.go so the strings are shared between pluginKind and the registry's own error messages.

pkg/plugins/ca/builtin/manager.go
```diff
@@ -125,6 +127,14 @@ func (b *builtinCaManager) create(ctx context.Context, mesh core_model.Resource,
...
+	}
+	log.Info("CA created", "mesh", meshName, "backend", backend.Name, "notAfter", notAfter)
```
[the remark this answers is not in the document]
bartsmykla: Refactored create() to fetch the persisted CA via getCa() at the end so notAfter is derived from the actually-stored cert, not the freshly-generated one that may not have been persisted.

Automaat (approved):

PR #16617, opened 2026-05-14
feat(log): export log volume counter

bartsmykla:
## Motivation

On-call has no way to see which loggers dominate kuma-cp log volume during an incident, so demotion decisions for chatty loggers (next PRs in this series) come from guesswork. Without a meter the standard rots within a year.

## Implementation information

Hook `componentAwareSink.Info` and `.Error` to increment a package-level `atomic.Pointer[prometheus.CounterVec]`. The pointer is nil until `SetupMetrics` is called from `pkg/metrics/components.Setup` during CP startup - no-op before then. Labels are `logger` (dot-separated name from `WithName` calls) and `level` (info/debug/error).

### DIFF
--- a/pkg/log/component_sink.go
+++ b/pkg/log/component_sink.go
@@ -1,10 +1,31 @@
 package log
 
 import (
+	"sync/atomic"
+
 	"github.com/go-logr/logr"
+	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
+
+// logLinesCounter holds the registered counter vec, or nil before SetupMetrics is called.
+var logLinesCounter atomic.Pointer[prometheus.CounterVec]
+
+// SetupMetrics registers the kuma_cp_log_lines_total counter with the given
+// Prometheus registerer. Call this once during CP startup, after the metrics
+// registry is ready.
+func SetupMetrics(reg prometheus.Registerer) (*prometheus.CounterVec, error) {
+	cv := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "kuma_cp_log_lines_total",
+		Help: "Total number of log lines emitted by the control plane, by logger name and level.",
+	}, []string{"logger", "level"})
+	if err := reg.Register(cv); err != nil {
+		return nil, err
+	}
+	logLinesCounter.Store(cv)
+	return cv, nil
+}
 
 // componentAwareSink wraps a logr.LogSink to add per-component log level
 // control. It intercepts WithName calls to track the hierarchical component
@@ -57,10 +78,20 @@ func (s *componentAwareSink) Enabled(level int) bool {
 }
 
 func (s *componentAwareSink) Info(level int, msg string, keysAndValues ...any) {
+	if cv := logLinesCounter.Load(); cv != nil {
+		lvl := "info"
+		if level > 0 {
+			lvl = "debug"
+		}
+		cv.WithLabelValues(s.name, lvl).Inc()
+	}
 	s.inner.Info(level, msg, keysAndValues...)
 }
 
 func (s *componentAwareSink) Error(err error, msg string, keysAndValues ...any) {
+	if cv := logLinesCounter.Load(); cv != nil {
+		cv.WithLabelValues(s.name, "error").Inc()
+	}
 	s.inner.Error(err, msg, keysAndValues...)
 }
 
--- a/pkg/log/component_sink.go
+++ b/pkg/log/component_sink.go
@@ -12,6 +12,12 @@ import (
 // logLinesCounter is nil until SetupMetrics is called.
 var logLinesCounter atomic.Pointer[prometheus.CounterVec]
 
+// ResetMetrics clears the global log-lines counter. Intended for test teardown
+// to prevent counter state from leaking across packages.
+func ResetMetrics() {
+	logLinesCounter.Store(nil)
+}
+
 // SetupMetrics registers the kuma_cp_log_lines_total counter with the given registerer.
 func SetupMetrics(reg prometheus.Registerer) (*prometheus.CounterVec, error) {
 	cv := prometheus.NewCounterVec(prometheus.CounterOpts{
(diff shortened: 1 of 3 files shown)


pkg/log/component_sink.go
```diff
@@ -1,11 +1,30 @@
...
+	"sync/atomic"
+
 	"github.com/go-logr/logr"
+	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
+// logLinesCounter is nil until SetupMetrics is called.
+var logLinesCounter atomic.Pointer[prometheus.CounterVec]
```
Automaat, line 13: **issue:** Package-level mutable `logLinesCounter` couples pkg/log to a Prometheus type and creates a hidden seam between two bounded contexts. An injected `EmissionObserver` interface would localize the contract.

Automaat, line 13: **suggestion:** Convention drift. Most Kuma metrics use constructor injection (`pkg/metrics/store/counter.go:32`); a package-level atomic pointer initialised via side-effecting `SetupMetrics` invites copy-paste for the next counter.

pkg/metrics/components/cp_info.go
```diff
@@ -51,6 +52,10 @@ func Setup(rt runtime.Runtime) error {
 		}
 	}
 
+	if _, err := kuma_log.SetupMetrics(rt.Metrics()); err != nil {
```
Automaat, line 55: **suggestion:** Add a duplicate-registration guard that matches the pattern used by GoCollector/ProcessCollector at lines 44-52.

pkg/log/component_sink.go
```diff
@@ -57,10 +76,20 @@ func (s *componentAwareSink) Enabled(level int) bool {
 }
 
 func (s *componentAwareSink) Info(level int, msg string, keysAndValues ...any) {
+	if cv := logLinesCounter.Load(); cv != nil {
+		lvl := "info"
+		if level > 0 {
+			lvl = "debug"
+		}
+		cv.WithLabelValues(s.name, lvl).Inc()
```
Automaat: **issue:** The root logger before any `WithName` has `s.name == ""`, producing a blank-label row that breaks dashboards and alerting templates. Default to `"root"` or reject nameless emissions.

pkg/log/component_sink.go
Automaat, line 13: **issue:** Tests overwrite the package-level `logLinesCounter` global. Subsequent tests in the same process that use `componentAwareSink` will increment the orphaned counter, leaking state across packages.

pkg/metrics/components/cp_info.go
Automaat, line 55: **issue:** A `Register` failure during `SetupMetrics` propagates up and aborts CP boot for a log-telemetry feature. Wrap with `isDuplicatedError` and log-and-continue so a counter cannot block startup.

pkg/log/component_sink.go
Automaat: **suggestion:** Pre-resolve three `prometheus.Counter` handles per sink in `WithName` and store them on the struct. The hot path then collapses to a pointer load plus `Inc()`.

pkg/log/component_sink.go
```diff
@@ -57,10 +76,20 @@ func (s *componentAwareSink) Enabled(level int) bool {
 }
 
 func (s *componentAwareSink) Info(level int, msg string, keysAndValues ...any) {
+	if cv := logLinesCounter.Load(); cv != nil {
+		lvl := "info"
+		if level > 0 {
```
Automaat: **issue:** The `level` enum is asymmetric: `level > 0` collapses to `"debug"` with no `"warn"`, `"fatal"`, or `"panic"`. A future Warn() or V-level distinction can't ship without breaking dashboards that filter on `level=~"info|debug|error"`.

pkg/log/component_sink.go
Automaat: **issue:** Pre-resolve the three counters at `WithName` time and store them on the struct, matching the pattern `names` already uses at line 114, to avoid map lookups on every log line.

pkg/metrics/components/cp_info.go
bartsmykla: done - added `isDuplicatedError` guard in cp_info.go matching the GoCollector/ProcessCollector pattern.

pkg/log/component_sink.go
bartsmykla: fixed - root logger now defaults to `"root"` label. Counters are pre-resolved at `NewComponentAwareSink` time using `"root"` so the empty-name path never reaches the label.

pkg/log/component_sink.go
bartsmykla: fixed - added `ResetMetrics()` and call it via `DeferCleanup` in the test `BeforeEach`, so the global is nil after each test and can't bleed into other packages.

pkg/metrics/components/cp_info.go
bartsmykla: fixed - `SetupMetrics` now handles `prometheus.AlreadyRegisteredError` internally (recovers the existing collector), so it never propagates that error. The `isDuplicatedError` guard in cp_info.go is an extra safety net for any other unexpected errors.

pkg/log/component_sink.go
bartsmykla: same fix as the sibling comment above - counters pre-resolved in `WithName`, stored on struct.

bartsmykla: done - added `cntInfo`/`cntDebug`/`cntError` fields on the struct, pre-resolved in `WithName` (and `NewComponentAwareSink` for the root logger). Hot path is now just a nil check + `Inc()`.

pkg/log/component_sink.go
bartsmykla: fixed - the `level == 0` / `else` split now explicitly maps V(0) -> "info" and all other V-levels -> "debug". logr's `LogSink` has no Warn/Fatal path so those labels can't appear via this interface - the comment in the code explains this.

pkg/log/component_sink.go
bartsmykla: deferring - an `EmissionObserver` interface would be a meaningful API change (new exported type, updated `NewComponentAwareSink` signature, callers updated). Tracking as a follow-up to keep this PR focused.

bartsmykla: noted - agree on the pattern. The constructor injection + interface refactor is bundled with comment #1 above, deferring both together to a follow-up.

Automaat (approved):

PR #16047, opened 2026-03-30
fix(kri): match HashSuffixMapper hash

Automaat:
## Motivation

KRI lookup via `/_kri/{kri}` returns 404 for resources without zone/namespace labels. `getCoreName` passed empty strings to `HashedName` for absent zone/namespace, while `HashSuffixMapper` in KDS sync skips absent labels entirely. Although FNV treats empty writes as no-ops (hashes happen to match today), the code should explicitly match `HashSuffixMapper` behavior for correctness.

Depends on kumahq/kuma#16046

Fixes #15544
Fixes #15803

## Implementation information

- `getCoreName` now only includes non-empty zone/namespace values in `HashedName` call, matching `HashSuffixMapper` in `pkg/kds/context/context.go`
- `zone=""` always means globally originated, no ambiguity on Zone CP — locally created resources always have `zone==cpZone`
- Added test for global-scoped resources (e.g. HostnameGenerator with mesh="") on Zone CP
- Added test for Universal Zone CP with empty zone/namespace KRI

PR #16620, opened 2026-05-14
fix(kri): populate zone label on policies

### LINKED ISSUE #15544: Generation of KRI does not use default fallbacks

Automaat:
## Motivation

KRI lookup via `/_kri/{kri}` 404s for policies created on a non-federated zone CP (universal or k8s standalone). The emitted KRI has an empty zone slot (e.g. `kri_mcb_default___mesh-circuit-breaker-all-default_`), the lookup routes through `IsLocallyOriginated`, sees `id.Zone == "" != cpZone`, falls into the hashed-name path, and misses the resource stored under its plain name. Manually filling the zone slot (e.g. `kri_mcb_default_default__...`) succeeds.

Root cause in `pkg/core/resources/labels/compute.go`: `kuma.io/zone` / `kuma.io/origin` are only set for resources with `ProvidedByZoneFlag`. For policies whose KDS flags omit it (or whose flag set changes), the resource is persisted without a zone label, the KRI emits an empty slot, and the lookup invariant breaks. PR #15921 papered over this in REST output and was reverted in #16046; #16047 assumed locally-created resources always have `zone==cpZone`, which doesn't hold for these policies.

Fixes #15544

## Implementation information

- `pkg/core/resources/labels/compute.go`: extend the zone-label trigger — plugin-originated policies (`IsPolicy && IsPluginOriginated`) qualify alongside `ProvidedByZoneFlag` resources, and the block is gated on a non-empty `ZoneName` so misconfigured CPs don't write empty labels.
- `pkg/api-server/kri_endpoint.go`: add `candidateNames` — when the KRI zone slot is empty on a zone CP, try the hashed-name lookup first (preserves existing behavior for globally-originated resources synced down) then fall back to the plain name (covers locally-admitted resources whose `kuma.io/zone` label was never populated, including any that already exist from before this fix).
- `pkg/api-server/kri_endpoint_test.go`: regression test for the empty-zone fallback path.

> Changelog: fix(kri): populate zone label on policies

## Supporting documentation

- kumahq/kuma#15544
- kumahq/kuma#15921 (reverted in #16046)
- kumahq/kuma#16047

### DIFF
--- a/pkg/api-server/kri_endpoint.go
+++ b/pkg/api-server/kri_endpoint.go
@@ -51,10 +51,10 @@ func (k *kriEndpoint) findByKriRoute() restful.RouteFunction {
 			return
 		}
 
-		name := k.getCoreName(identifier, *descriptor)
+		candidates := k.candidateNames(identifier, *descriptor)
 		if err := k.resourceAccess.ValidateGet(
 			request.Request.Context(),
-			core_model.ResourceKey{Mesh: identifier.Mesh, Name: name},
+			core_model.ResourceKey{Mesh: identifier.Mesh, Name: candidates[0]},
 			*descriptor,
 			user.FromCtx(request.Request.Context()),
 		); err != nil {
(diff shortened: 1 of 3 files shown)


Automaat:
Superseded — root cause was that default mesh policies are created by the CP without computed labels. Replaced by a focused fix in the defaults creator.

PR #16623, opened 2026-05-14
fix(meshaccesslog): validate otel keys

bartsmykla:
## Motivation

`MeshAccessLog` accepted OpenTelemetry attribute keys that do not follow the OpenTelemetry attribute-name rules. That let invalid keys like ones with spaces, or keys built from placeholders, pass validation even though the final access log config expects static attribute names.

## Implementation information

- validate `openTelemetry.attributes[*].key` against the OpenTelemetry attribute-name rules
- reject invalid or placeholder-derived keys during validation, and keep placeholders in values only
- update the `MeshAccessLog` schema text to say keys must be static


pkg/plugins/policies/meshaccesslog/api/v1alpha1/validator.go
```diff
@@ -158,13 +158,23 @@ func validateBackend(backend Backend) validators.ValidationError {
 			backend.OpenTelemetry.Endpoint,
 			backend.OpenTelemetry.BackendRef,
 		))
+		verr.AddErrorAt(root, validateOtelAttributes(pointer.Deref(backend.OpenTelemetry.Attributes)))
```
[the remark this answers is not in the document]
bartsmykla: Added an upgrade note in `UPGRADE.md` under `Upgrade to 2.14.x` that calls out the stricter `MeshAccessLog` OpenTelemetry key validation, explains when existing policies are affected, and shows how to move dynamic data into the value instead of the key.

pkg/core/validators/common_validators.go
```diff
@@ -238,6 +241,31 @@ func ValidateBandwidth(path PathBuilder, value string) ValidationError {
...
+	if strings.Contains(value, "%") {
+		err.AddViolationAt(path, "must be a static OpenTelemetry attribute name; placeholders are only supported in values")
+		return err
+	}
+	if strings.HasPrefix(value, "otel.") {
+		err.AddViolationAt(path, "must not use the reserved OpenTelemetry prefix 'otel.'")
+		return err
+	}
+	if !otelAttributeNameRegex.MatchString(value) {
+		err.AddViolationAt(path, "must be a valid OpenTelemetry attribute name")
```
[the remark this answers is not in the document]
bartsmykla: Kept the key grammar from the linked issue, but changed the validation error to spell out the exact accepted format and added coverage for uppercase and hyphenated keys so the restriction is explicit in tests.

pkg/plugins/policies/meshaccesslog/api/v1alpha1/validator.go
```diff
@@ -158,13 +158,23 @@ func validateBackend(backend Backend) validators.ValidationError {
 			backend.OpenTelemetry.Endpoint,
 			backend.OpenTelemetry.BackendRef,
 		))
```
Automaat, line 160: **issue:** `meshmetric` has the same `OtelBackend` shape with an `Attributes` field and no key validation. After this PR lands, the two policies diverge in input-validation behavior for the same conceptual contract. Either delegate to the new validator from the `meshmetric` validator, or open a follow-up issue to track the gap. Without a follow-up note, this fix-in-one-place becomes the kind of asymmetry that gets copied by the next contributor.

pkg/plugins/policies/meshaccesslog/plugin/xds/configurer.go
```diff
@@ -324,7 +324,7 @@ func OtelAttributes(backend *api.OtelBackend, values listeners_v3.KumaValues) Co
 		attributes := &otlp.KeyValueList{}
 		for _, kv := range pointer.Deref(backend.Attributes) {
 			attributes.Values = append(attributes.Values, &otlp.KeyValue{
-				Key: listeners_v3.InterpolateKumaValues(kv.Key, values),
+				Key: kv.Key,
```
Automaat: **issue:** Behavior change in `OtelAttributes` configurer is bigger than UPGRADE.md states. Pre-PR, `InterpolateKumaValues(kv.Key, values)` rewrote `%KUMA_ZONE%` keys to e.g. `zone-1` in the emitted Envoy config. Post-PR, keys pass through literally. Existing K8s `MeshAccessLog` resources with placeholder keys are not re-validated on upgrade (admission only fires on create/update), so they keep being applied — but their emitted OTel attribute key changes from `zone-1` to the literal string `%KUMA_ZONE%`. UPGRADE.md says "Existing policies keep working" — true for the control plane but not for downstream OTel consumers, who will start seeing attribute keys they were never sent before. Either:
1. Restore key interpolation as a transitional bridge, or
2. Update UPGRADE.md to call out the OTel-side attribute rename and recommend a discovery command (`kubectl get meshaccesslog -A -o yaml | grep -B1 'key:.*%'`).

pkg/core/validators/common_validators.go
```diff
@@ -15,6 +15,11 @@ import (
 	common_api "github.com/kumahq/kuma/v2/api/common/v1alpha1"
 )
 
+var (
+	BandwidthRegex         = regexp.MustCompile(`^(\d*)\s?([GMk]+bps)$`)
```
Automaat, line 19: **issue:** OTel spec allows attribute names with hyphens. Semconv uses them in `http.request.header.user-agent` and similar. The regex `^[a-z](?:[a-z0-9]|[._][a-z0-9])*$` rejects hyphens entirely. Two consequences:
1. Users who add `http.request.header.user-agent` (a real OTel semconv key) as a static attribute key will be rejected.
2. The validator is stricter than the OTel spec itself.

If the intent is "OTel semconv recommended grammar," that's defensible — but make it explicit in the error message and UPGRADE.md. If the intent is "what Envoy's OTel access logger accepts," that's a different (and possibly looser) constraint.

pkg/plugins/policies/meshaccesslog/api/v1alpha1/validator.go
bartsmykla: I checked `meshmetric` before widening this PR. Its `OpenTelemetryBackend` only has `endpoint` / `backendRef`; it does not have an `attributes` field, so there is no parallel key-validation gap to wire up there in this change.

pkg/plugins/policies/meshaccesslog/plugin/xds/configurer.go
bartsmykla: Good catch. I restored key interpolation in `OtelAttributes` as a compatibility bridge for legacy resources that were stored before this validation existed, added coverage in the workload-identity plugin test/golden for a placeholder key that still emits `zone-1`, and updated `UPGRADE.md` to explain that placeholder-based keys keep their current runtime behavior until the policy is updated or reapplied.

pkg/core/validators/common_validators.go
bartsmykla: I kept the grammar from the linked issue, but made the scope more explicit: `UPGRADE.md` now says this is the access-log attribute-key grammar enforced by `MeshAccessLog`, and the validation error already spells out the exact accepted format. I’m not widening this to allow hyphens in this fix because that would change the contract the issue asked to enforce.

pkg/plugins/policies/meshaccesslog/api/v1alpha1/validator.go
```diff
@@ -158,13 +158,23 @@ func validateBackend(backend Backend) validators.ValidationError {
...
+		verr.AddErrorAt(root, validateOtelAttributes(pointer.Deref(backend.OpenTelemetry.Attributes)))
 	default:
 		panic(fmt.Sprintf("unknown backend type %v", backend.Type))
 	}
 
 	return verr
 }
 
+func validateOtelAttributes(attributes []JsonValue) validators.ValidationError {
```
Automaat: **praise:** Path construction in `validateOtelAttributes` correctly yields `spec.from[0].default.backends[0].openTelemetry.attributes[0].key`. Confirmed via `AddErrorAt` semantics and the test expectation at `validator_test.go:570`.

pkg/plugins/policies/meshaccesslog/api/v1alpha1/meshaccesslog.go
```diff
@@ -81,7 +81,8 @@ type TCPBackend struct {
 
 // Defines an OpenTelemetry logging backend.
 type OtelBackend struct {
-	// Attributes can contain placeholders available on
+	// Attributes defines custom OpenTelemetry attributes. Keys must be static
```
Automaat: **suggestion:** Field doc says "Keys must be static OpenTelemetry attribute names". "Static" reads as "compile-time constant". Closer: "Keys must be valid OpenTelemetry attribute names and cannot contain `%...%` placeholders. Values can contain placeholders available on …".

UPGRADE.md
```diff
@@ -22,6 +22,28 @@ None for valid label keys. Labels with `/` or `.` in the key that were leaked on
...
+
+```sh
+kubectl get meshaccesslogs -A -o yaml | grep -B1 'key: .*%'
+```
+
+Then migrate them like this:
+
+```yaml
+attributes:
+  - key: service.zone
```
Automaat, line 50: **issue:** UPGRADE.md grep recipe `kubectl get meshaccesslogs -A -o yaml | grep -B1 'key: .*%'` catches only the placeholder class. The validator rejects four more classes the prose itself names (hyphens, uppercase, leading-digit, `otel.` prefix, consecutive delimiters). Users who run the documented recipe will think they are clean, then hit admission rejection on next apply. Provide a complete audit script (e.g. run the actual regex `^[a-z](?:[a-z0-9]|[._][a-z0-9])*$` over every `attributes[].key`).

pkg/plugins/policies/meshaccesslog/k8s/crd/kuma.io_meshaccesslogs.yaml
```diff
@@ -120,7 +120,8 @@ spec:
                                 properties:
                                   attributes:
```
Automaat, line 121: **issue:** No CRD-level `pattern` constraint on the key field. `spec.versions[*].schema.openAPIV3Schema` still types it as plain `string`. Schema-only validators used in CI (kubeconform, kubeval, datree) and `kubectl apply --dry-run=client` won't catch invalid keys. Only the live admission webhook rejects. CI and gitops dry-runs pass, then fail at actual apply. Add the pattern to the CRD or document that schema-only validation is insufficient.

UPGRADE.md
```diff
@@ -22,6 +22,28 @@ None for valid label keys. Labels with `/` or `.` in the key that were leaked on
...
+`MeshAccessLog` now validates `openTelemetry.attributes[].key` against the access-log attribute-key grammar used by this policy. Keys must start with a lowercase letter, use only lowercase letters, digits, `_` or `.`, avoid consecutive delimiters, end with a letter or digit, and must not use the reserved `otel.` prefix. `%...%` placeholders remain supported in attribute values, but are no longer accepted in keys.
+
+Existing policies with invalid keys keep their current runtime behavior until they are updated or reapplied. In particular, placeholder-based keys continue to emit the same interpolated key after a control-plane upgrade. After that, any create or update using an invalid key is rejected until the key is renamed to a static value.
+
+**Action required:**
+
+If you have existing `MeshAccessLog` policies with keys such as `%KUMA_ZONE%`, `request-id`, `Service.Version`, or `otel.attribute`, rename the key and keep the dynamic content in the value instead. On Kubernetes you can find placeholder-based keys before the next apply with:
+
+```sh
+kubectl get meshaccesslogs -A -o yaml | grep -B1 'key: .*%'
```
Automaat: **issue:** Universal mode hits the same `resourcesManager.Create/Update` → `validator.Validate` path. `kumactl apply` and direct HTTP-API writes reject identically, but UPGRADE.md only documents a `kubectl` audit recipe. Universal users get no equivalent path to find affected resources before they break.

UPGRADE.md
```diff
@@ -22,6 +22,28 @@ None for valid label keys. Labels with `/` or `.` in the key that were leaked on
 kubectl -n kuma-system label meshservice <name> app.example.com/tier-
```
Automaat, line 22: **issue:** UPGRADE.md says legacy stored placeholder keys "continue to emit the same interpolated key after a control-plane upgrade". True at the configurer, false at the sync layer. Add an explicit GitOps warning: a reconciler that re-applies on drift starts failing the moment the CP upgrades, unless invalid keys are fixed first.

pkg/core/validators/common_validators.go
```diff
@@ -238,6 +241,34 @@ func ValidateBandwidth(path PathBuilder, value string) ValidationError {
...
+// key grammar: lowercase letters, digits, `_` and `.`, starting with a letter,
+// ending with an alphanumeric, without consecutive delimiters, and without the
+// reserved `otel.` prefix. `%...%` placeholders are rejected because only
+// values may be interpolated.
+func ValidateOtelAttributeName(path PathBuilder, value string) ValidationError {
+	var err ValidationError
+	if value == "" {
+		err.AddViolationAt(path, MustBeDefined)
+		return err
+	}
```
Automaat, line 254: **issue:** The three rejection messages are inline string literals. The rest of `pkg/core/validators/` promotes shared strings to constants in `messages.go` (`MustBeDefined`, `MustHaveBPSUnit`). Tests in `common_validators_test.go` already reference those constants. The new messages get duplicated as raw strings between production code and tests, which breaks the established pattern. Promote them to constants (e.g., `OtelAttributeNamePlaceholder`, `OtelAttributeNameReservedPrefix`, `OtelAttributeNameInvalidFormat`).

pkg/plugins/policies/meshaccesslog/api/v1alpha1/validator_test.go
```diff
@@ -531,6 +547,50 @@ from:
 violations:
   - field: spec.from[0].default.backends[0].openTelemetry.backendRef
     message: "backendRef must have exactly one defined: name, labels"`,
+			}),
+			Entry("openTelemetry attribute key with spaces", testCase{
```
Automaat, line 554: **issue:** `validateOtelAttributes` accumulates errors across every attribute (no early return), but no test verifies multi-error collection. Add a `validator_test.go` entry with two invalid keys in the same `attributes` list and assert both `attributes[0].key` and `attributes[1].key` violations appear.

Automaat, line 554: **issue:** `validator_test.go` covers space + placeholder integration cases but not the reserved-prefix path through the full field-path assertion (`spec.from[…].openTelemetry.attributes[0].key` + reserved-prefix message). Add a third entry.

UPGRADE.md
bartsmykla: Updated `UPGRADE.md` to use an export-plus-audit command that runs the full runtime rule across every `MeshAccessLog` OpenTelemetry key, not just the placeholder case. It now catches placeholder keys, the reserved `otel.` prefix, and keys that miss the enforced grammar.

pkg/plugins/policies/meshaccesslog/k8s/crd/kuma.io_meshaccesslogs.yaml
bartsmykla: Added a dedicated `OtelAttribute` type for `openTelemetry.attributes`, so the generated CRD/OpenAPI schema now carries a `pattern` on that key without changing the plain JSON log-format keys. The reserved `otel.` prefix remains enforced by the admission/runtime validator.

UPGRADE.md
bartsmykla: Added the Universal-mode path in `UPGRADE.md`: export with `kumactl get meshaccesslogs --mesh <mesh> -o yaml > meshaccesslogs.yaml` per mesh, then run the same audit command against that export.

UPGRADE.md
bartsmykla: Added an explicit GitOps/reconciler warning in `UPGRADE.md`: stored legacy resources keep their current runtime behavior until they are reapplied, but any reconciler that reapplies after the control-plane upgrade immediately hits the stricter validation, so invalid keys need to be fixed before the next reconcile.

pkg/core/validators/common_validators.go
bartsmykla: Promoted the three new messages into `pkg/core/validators/messages.go` and switched both the validator implementation and its tests to use the shared constants.

pkg/plugins/policies/meshaccesslog/api/v1alpha1/validator_test.go
bartsmykla: Added an integration case with two invalid attribute keys in the same `attributes` list and asserted that both field-specific violations are returned.

bartsmykla: Added a full `validator_test.go` entry for the reserved-prefix path, including the exact `spec.from[0].default.backends[0].openTelemetry.attributes[0].key` field assertion and message.

Automaat (dismissed):

Automaat (dismissed):

lukidzi (approved):

PR #16625, opened 2026-05-14
feat(sni): added validation of format

### LINKED ISSUE #16554: Implement Validation of SNI format
We should ensure if SNI format is correct and reject resources fast

lukidzi:
## Motivation

When the mesh-scoped zone proxy generates clusters/listeners it derives the upstream SNI from a resource's KRI. Today nothing prevents users from putting characters into MeshService/MeshExternalService/MeshMultiZoneService resource (mesh, name, zone/namespace label, port name) that produce an invalid DNS-1035 hostname once the segments are joined with `.`. The resulting SNI is silently malformed and the xDS generators emit it anyway; the symptom only shows up at TLS handshake time on the upstream Envoy.

## Implementation information

New `pkg/core/resources/sni` package owns the KRI→SNI format and its validation. It has no dependency on `pkg/xds/envoy/tls`, so resource APIs can import it (the old helpers lived in tls and caused an import cycle).

  - FromKRI emits 5/6/7 segments per MADR-101; for global-originated resources the kube namespace is dropped so Global k8s and Global universal produce the same SNI.
  - ValidateKRI uses `apimachineryvalidation.NameIsDNS1035Label` for mesh/name/zone/namespace (same validator we already use for the resource-name deprecation, so the wording stays consistent). The port section accepts a DNS-1035 label or an all-digit port, which keeps MES (match.port is numeric) and unnamed MS/MMZS ports working.

  `Deprecations()` on the three resources now appends a warning per offending port. The validating webhook surfaces it through the existing deprecation channel — no new gate, legacy resources keep working.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/16554


pkg/api-server/testdata/resources/crud/put_meshexternalservice_02.golden.json
```diff
@@ -1,5 +1,6 @@
 {
  "warnings": [
-  "Invalid MeshExternalService resource name: 'mes.http'. It does not conform to the DNS format (RFC 1035). This is deprecated. Errors: a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')"
+  "Invalid MeshExternalService resource name: 'mes.http'. It does not conform to the DNS format (RFC 1035). This is deprecated. Errors: a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')",
+  "Invalid MeshExternalService SNI: name \"mes.http\" does not conform to RFC 1123: must not contain dots. It does not conform to the DNS format (RFC 1123). This is deprecated."
```
slonka: the 

```
does not conform to RFC 1123: must not contain dots. It does not conform to the DNS format (RFC 1123)
```

part is repetitive

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/deprecated.go
```diff
@@ -6,18 +6,34 @@ import (
...
+	}
+
+	base := kri.From(t)
+	seen := map[string]struct{}{}
+	for _, port := range t.Spec.Ports {
+		for _, err := range sni.ValidateKRI(kri.WithSectionName(base, port.GetName())) {
+			msg := fmt.Sprintf(
+				"Invalid %s SNI: %s. It does not conform to the DNS format (RFC 1123). This is deprecated.",
+				MeshMultiZoneServiceResourceTypeDescriptor.Name, err)
+			if _, ok := seen[msg]; ok {
```
slonka: why were the messages duplicated? / why do we need to undup them?

pkg/core/resources/apis/meshservice/api/v1alpha1/deprecated.go
```diff
@@ -6,18 +6,34 @@ import (
...
 func (t *MeshServiceResource) Deprecations() []string {
 	var deprecations []string
 
 	name := model.GetDisplayName(t.GetMeta())
-	allErrs := apimachineryvalidation.NameIsDNS1035Label(name, false)
-	if len(allErrs) != 0 {
-		nameDeprecationMsg := fmt.Sprintf("Invalid %s resource name: '%s'. It does not conform to the DNS format (RFC 1035). This is deprecated. Errors: %s",
-			MeshServiceResourceTypeDescriptor.Name, name, strings.Join(allErrs, "; "))
-		deprecations = append(deprecations, nameDeprecationMsg)
+	if allErrs := apimachineryvalidation.NameIsDNS1035Label(name, false); len(allErrs) != 0 {
```
slonka, line 18: should we make this a common validator and just call that from the 3 resources?

pkg/core/resources/sni/sni_test.go
```diff
@@ -0,0 +1,541 @@
...
+			},
+			expected: "sni.mzsvc.default.global-svc.8080",
+		}),
+		Entry("error empty mesh", kriTestCase{
+			id: kri.Identifier{
+				ResourceType: meshservice_api.MeshServiceType,
+				Name:         "backend",
+				SectionName:  "http",
+			},
+			expectErr: true,
```
slonka: instead of `expectErr` being a bool should we just match the error?

pkg/core/resources/sni/sni_test.go
```diff
@@ -0,0 +1,541 @@
...
+			id: kri.Identifier{
+				ResourceType: meshservice_api.MeshServiceType,
+				Mesh:         "default",
+				Name:         "backend",
+				SectionName:  strings.Repeat("p", 64),
+			},
+			expectErr: true,
+		}),
+		Entry("total length just under 253 chars passes", kriTestCase{
+			// 4 + 1 + 4 + 1 + 60 + 1 + 60 + 1 + 60 + 1 + 60 + 1 + 4 = 258 — still > 253
```
slonka: nitpick: not sure why not 59? it's `254` so still > `253` - also do we need this calculation, can't we just leave the one below?

pkg/core/resources/sni/sni.go
```diff
@@ -0,0 +1,108 @@
...
+//   - SectionName conforms to RFC 1123
+//   - total length ≤ 253 (DNS hostname limit)
+func ValidateKRI(id kri.Identifier) []error {
+	desc, err := registry.Global().DescriptorFor(id.ResourceType)
+	if err != nil {
+		return nil
+	}
+	if _, ok := sniCapableShortNames[desc.ShortName]; !ok {
+		return nil
+	}
```
slonka, line 58: why don't we return the errors?

pkg/core/resources/sni/sni_test.go
```diff
@@ -0,0 +1,541 @@
...
+
+	It("reports an RFC 1123 violation for an over-long label", func() {
+		errs := sni.ValidateKRI(kri.Identifier{
+			ResourceType: meshservice_api.MeshServiceType,
+			Mesh:         "default",
+			Name:         strings.Repeat("a", 64),
+			SectionName:  "http",
+		})
+		Expect(errs).ToNot(BeEmpty())
+		Expect(errs[0].Error()).To(ContainSubstring("RFC 1123"))
```
slonka: Q: to this and all the ones below that have `ContainSubstring` - in other tests we usually match the whole error, why are we using substrings here?

pkg/core/resources/sni/sni.go
```diff
@@ -0,0 +1,108 @@
...
+//
+// The format is:
+//
+//	sni.<short>.<mesh>.<name>.<sectionName>                          (5 segments) — global-originated (k8s or universal)
+//	sni.<short>.<mesh>.<zone>.<name>.<sectionName>                   (6 segments) — zone-originated on universal
+//	sni.<short>.<mesh>.<zone>.<namespace>.<name>.<sectionName>       (7 segments) — zone-originated on k8s
+//
+// Namespace is only emitted when the resource is also zone-scoped: a
+// global-originated resource (zone == "") drops the namespace segment even
+// if the k8s label is present.
```
slonka: didn't the old code take into account `ns` always? would two synced resources named `backend` in different namespaces now both have `sni.<res>.<mesh>.backend.<port>` and collide?

pkg/core/resources/sni/sni_test.go
```diff
@@ -0,0 +1,541 @@
...
+			Mesh:         "de.fault",
+			Name:         "backend",
+			SectionName:  "http",
+		})
+		Expect(errs).To(HaveLen(1))
+		Expect(errs[0].Error()).To(ContainSubstring("mesh"))
+		Expect(errs[0].Error()).To(ContainSubstring("de.fault"))
+		Expect(errs[0].Error()).To(ContainSubstring("RFC 1123"))
+	})
+})
```
slonka, line 521: nit: I have a general feeling that some there is a bit of repetition here, especially around: numeric-only sectionName is valid, sectionName six digits is valid, sectionName mixing digits and letters is valid, sectionName with letters then digits.

slonka (commented):
I think the most important Q is about dropping `ns`

pkg/core/resources/sni/sni.go
lukidzi: No namespace is only for universal zone (but you have a zone section) or for global originated - not synced from other zone but created on global, so there is no collide.

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/deprecated.go
lukidzi: Refactored to include port name in the message, as each port has different SNI.

pkg/core/resources/apis/meshservice/api/v1alpha1/deprecated.go
lukidzi: Don't feel like it's required as in 3.0 we could just forbid creation

lobkovilya (approved):

PR #16627, opened 2026-05-15
feat(meshproxypatch): zone proxy Dataplanes

### LINKED ISSUE #16560: MeshProxyPatch: design support on mesh-scoped zone proxies
### Description

[MADR-103](https://github.com/kumahq/kuma/blob/6f19192fd666aaf19a4e8632274e1da52962e1f4/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md) group (4) lists `MeshProxyPatch` as a proxy-wide policy intended to apply to mesh-scoped zone proxies via top-level `targetRef`. Unlike `MeshMetric`/`MeshTrace`, `MeshProxyPatch` patches raw Envoy config and its semantics on zone proxies need a separate design pass before implementation — listener/cluster naming on zone proxies, which resources are safe to patch, and how `sectionName` constrains the patch scope (or shouldn't be used).

Automaat:
## Motivation

Closes #16560. Implements [MADR-103](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md) group (4) for `MeshProxyPatch` on mesh-scoped zone proxy `Dataplane` resources (those carrying `networking.listeners[]` of type `ZoneIngress`/`ZoneEgress` per MADR-095).

Companion work for the other group-(4) policies already landed: `MeshMetric` (#16597) and `MeshTrace` (#16601).

## Implementation information

`MeshProxyPatch` needs **no plugin-side changes** to support zone proxy `Dataplane` resources. The matcher wiring (#16584) already routes Dataplane-targeted policies that hit embedded zone proxy listeners into `proxy.Policies.Dynamic[MeshProxyPatchType]`, and the existing `SingleItemRules.Compute` + `ApplyMods` path in `plugin.Apply` applies them. This is the same conclusion as `MeshMetric` (#16597), which also needed no apply-path change.

`MeshProxyPatch` is proxy-wide — there is no per-listener scoping. To target a specific zone proxy DPP, use the computed labels on `targetRef`:

- `kuma.io/listener-zoneegress: enabled`
- `kuma.io/listener-zoneingress: enabled`

`sectionName` on a `kind: Dataplane` `targetRef` is rejected by the shared targetRef validator for non-inbound policies, so it is not a supported way to scope `MeshProxyPatch`.

### History

An earlier revision of this branch added an `applyZoneProxyDataplane` path with `sectionName`-based per-listener scoping. That code was removed: the `sectionName` scoping was unreachable (validator-rejected) and the rest duplicated the existing `SingleItemRules` path. The branch now only adds test coverage.

## Tests

- **Unit** — `plugin_zone_proxy_test.go`: three table entries with YAML goldens proving `MeshProxyPatch` applies through the real matcher on zone egress, zone ingress, and mixed (inbound + embedded zone egress) DPPs.
- **E2E envoyconfig** — three inputs wired into the existing zone-proxy fixture (`test/e2e_env/universal/envoyconfig/zoneproxies.go`):
  - `cluster-patch.input.yaml` — patches `kuma:envoy:admin` cluster's `connectTimeout`
  - `cluster-add.input.yaml` — adds a new cluster on both DPPs via `kind: Dataplane`
  - `label-scope.input.yaml` — two policies, one scoped with `kuma.io/listener-zoneegress`, one with `kuma.io/listener-zoneingress`, each adding a uniquely-named cluster
- **E2E live config** — `test/e2e_env/universal/meshproxypatch/zone_proxy.go`: exercises the full path — a real policy applied, pushed via xDS, and observed on the live Envoy (`/clusters`) of each zone proxy. Covers a `kind: Dataplane` policy reaching both proxies and label-scoped policies reaching only their targeted proxy.

## Supporting documentation

- MADR-103: `docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md`
- MADR-095 (zone proxy as Dataplane): `docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md`
- Companion PRs: feat(meshmetric): support zone proxy Dataplanes (#16597), feat(meshtrace): support zone proxy Dataplanes (#16601)
- Matchers wiring: feat(matchers): include zone proxy listeners when matching DPP policies (#16584)


pkg/plugins/policies/meshproxypatch/plugin/v1alpha1/plugin_zone_proxy_test.go
```diff
@@ -0,0 +1,245 @@
...
+					{Listener: &api.ListenerMod{
+						Operation: api.ModOpPatch,
+						Match:     &api.ListenerMatch{Name: pointer.To(naming.ContextualZoneIngressListenerName("zi-port"))},
+						Value:     pointer.To("statPrefix: patched_ingress\n"),
+					}},
+				}),
+			},
+			goldenFile: "zone-ingress-listener-patch",
+		}),
+		Entry("mixed inbound + zone egress, cluster added by Dataplane targetRef", testCase{
```
slonka: I think this modification would work on a pure zone-proxy DPP — should we strengthen it so the test actually exercises the mixed-DPP shape?

pkg/plugins/policies/meshproxypatch/plugin/v1alpha1/plugin_zone_proxy_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+			clusterYaml, err := util_yaml.GetResourcesToYaml(rs, envoy_resource.ClusterType)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(clusterYaml).To(matchers.MatchGoldenYAML(filepath.Join("testdata", given.goldenFile+".clusters.golden.yaml")))
+		},
+		Entry("zone egress, listener patch matched by explicit name", testCase{
+			dp:        zoneEgressOnlyDataplane(),
+			resources: []core_xds.Resource{zoneEgressListenerResource()},
+			policies: []*api.MeshProxyPatchResource{
+				newMeshProxyPatch("by-name", &common_api.TargetRef{Kind: common_api.Mesh}, []api.Modification{
```
slonka: instead of `kind: Mesh` should we do `Dataplane` and `kuma.io/listener-zoneegress`? or Maybe on top of that? 🤔

slonka (dismissed):

lobkovilya (dismissed):
Let's not rush with MeshProxyPatch, it's easily 3.0 and the original issue mentions the design is required https://github.com/kumahq/kuma/issues/16560

pkg/plugins/policies/meshproxypatch/plugin/v1alpha1/plugin_zone_proxy_test.go
slonka: Addressed in the latest revision - the unit table now includes a mixed inbound + zone egress Dataplane and asserts that a Dataplane-targeted MeshProxyPatch patches both listeners.

pkg/plugins/policies/meshproxypatch/plugin/v1alpha1/plugin_zone_proxy_test.go
slonka: Updated to the agreed model: the zone ingress and zone egress cases now target kind: Dataplane with the computed listener-zone labels, so this keeps MPP support without extending the API footprint.

slonka (commented):
Let's make the CI pass and I don't think there's anything blocking here

slonka (approved):

PR #16637, opened 2026-05-15
fix(defaults): compute labels on default policies

### LINKED ISSUE #15544: Generation of KRI does not use default fallbacks

Automaat:
## Motivation

KRI lookup via `/_kri/{kri}` 404s for the auto-created default mesh policies (`mesh-circuit-breaker-all-default`, `mesh-timeout-all-default`, …) on a zone CP.

Root cause: `pkg/defaults/mesh/mesh.go` creates default policies with a bare `resManager.Create(...)` — no label computation, no `CreateWithLabels`. So every default policy is persisted with zero labels. On a zone CP the resource is missing `kuma.io/zone`, its KRI carries an empty zone slot (`kri_mcb_default___mesh-circuit-breaker-all-default_`), and the lookup routes through the hashed-name path and misses the resource stored under its plain name.

User-created resources go through the API server's `createResource` → `labels.Compute`, so they're unaffected. Only the CP-auto-created defaults are broken — on every zone CP, old or new.

Fixes #15544

## Implementation information

**A — compute labels at creation.** `EnsureDefaultMeshResources` now takes the CP mode and zone. `ensureDefaultResource` computes labels via `labels.Compute` and passes `store.CreateWithLabels`, so newly created default policies carry `kuma.io/zone` / `kuma.io/origin` / `kuma.io/mesh`.

**B — heal existing resources.** `ensureDefaultResource` also reconciles labels on an already-existing default resource (recompute, `Update` if they differ — spec untouched). A new boot-time `EnsureDefaultMeshResourcesUpToDate` step iterates every Mesh and runs the reconciliation in reconcile-only mode: it updates labels of existing default policies but never recreates ones an operator deleted. Default policies stored by older CP versions are healed in place on the next CP start — no migration tooling, no manual re-apply.

Federated zone CPs are skipped (defaults come from Global via KDS). Threaded mode/zone through the two existing callers (`mesh_manager.go`, k8s `mesh_controller.go`). The `EnsureDefaultFunc` callback gained an `extensions` context so the boot-time step propagates log fields like every other caller.

> Changelog: fix(defaults): compute labels on default policies

## Supporting documentation

- kumahq/kuma#15544
- Supersedes the earlier symptom-only attempt in #16620

### DIFF
--- a/pkg/core/managers/apis/mesh/mesh_manager.go
+++ b/pkg/core/managers/apis/mesh/mesh_manager.go
@@ -7,6 +7,7 @@ import (
 	"github.com/pkg/errors"
 
 	kuma_cp "github.com/kumahq/kuma/v2/pkg/config/app/kuma-cp"
+	config_core "github.com/kumahq/kuma/v2/pkg/config/core"
 	config_store "github.com/kumahq/kuma/v2/pkg/config/core/resources/store"
 	core_ca "github.com/kumahq/kuma/v2/pkg/core/ca"
 	core_mesh "github.com/kumahq/kuma/v2/pkg/core/resources/apis/mesh"
@@ -36,6 +37,8 @@ func NewMeshManager(
 		unsafeDelete:               config.Store.UnsafeDelete,
 		extensions:                 extensions,
 		createMeshRoutingResources: config.Defaults.CreateMeshRoutingResources,
+		cpMode:                     config.Mode,
+		cpZone:                     config.Multizone.Zone.Name,
 	}
 	if config.Store.Type == config_store.KubernetesStore {
 		meshManager.k8sStore = true
@@ -55,6 +58,8 @@ type meshManager struct {
 	createMeshRoutingResources bool
 	k8sStore                   bool
 	systemNamespace            string
+	cpMode                     config_core.CpMode
+	cpZone                     string
 }
 
 func (m *meshManager) Get(ctx context.Context, resource core_model.Resource, fs ...core_store.GetOptionsFunc) error {
@@ -105,6 +110,9 @@ func (m *meshManager) Create(ctx context.Context, resource core_model.Resource, 
 		m.createMeshRoutingResources,
 		m.k8sStore,
 		m.systemNamespace,
+		m.cpMode,
+		m.cpZone,
+		false, // create missing default resources
 	); err != nil {
 		return err
 	}
--- a/pkg/core/managers/apis/mesh/mesh_manager.go
+++ b/pkg/core/managers/apis/mesh/mesh_manager.go
@@ -116,7 +116,7 @@ func (m *meshManager) Create(ctx context.Context, resource core_model.Resource, 
 		m.systemNamespace,
 		m.cpMode,
 		m.cpZone,
-		false, // create missing default resources
+		false, // reconcileExistingOnly
 	); err != nil {
 		return err
 	}
--- a/pkg/core/managers/apis/mesh/mesh_manager.go
+++ b/pkg/core/managers/apis/mesh/mesh_manager.go
@@ -28,6 +28,10 @@ func NewMeshManager(
 	extensions context.Context,
 	config kuma_cp.Config,
 ) core_manager.ResourceManager {
+	var cpZone string
+	if config.Multizone != nil {
+		cpZone = config.Multizone.Zone.Name
+	}
 	meshManager := &meshManager{
 		store:                      store,
 		otherManagers:              otherManagers,
@@ -38,7 +42,7 @@ func NewMeshManager(
 		extensions:                 extensions,
 		createMeshRoutingResources: config.Defaults.CreateMeshRoutingResources,
 		cpMode:                     config.Mode,
-		cpZone:                     config.Multizone.Zone.Name,
+		cpZone:                     cpZone,
 	}
 	if config.Store.Type == config_store.KubernetesStore {
 		meshManager.k8sStore = true
--- a/pkg/defaults/components.go
+++ b/pkg/defaults/components.go
@@ -20,12 +20,13 @@ import (
 
 var log = core.Log.WithName("defaults")
 
-type EnsureDefaultFunc = func(ctx context.Context, resManager core_manager.ResourceManager, logger logr.Logger, cfg kuma_cp.Config) error
+type EnsureDefaultFunc = func(ctx context.Context, resManager core_manager.ResourceManager, logger logr.Logger, cfg kuma_cp.Config, extensions context.Context) error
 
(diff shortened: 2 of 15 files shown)


bartsmykla (commented):
Code is solid, tests cover the heal-and-skip-recreate paths well. Comments below are mostly call-site readability and one duplicated inline comment. None blocking.

pkg/defaults/components.go
```diff
@@ -20,12 +20,13 @@ import (
 
 var log = core.Log.WithName("defaults")
 
-type EnsureDefaultFunc = func(ctx context.Context, resManager core_manager.ResourceManager, logger logr.Logger, cfg kuma_cp.Config) error
+type EnsureDefaultFunc = func(ctx context.Context, resManager core_manager.ResourceManager, logger logr.Logger, cfg kuma_cp.Config, extensions context.Context) error
```
bartsmykla, line 25: `extensions context.Context` is a Kuma-internal pattern (logger fields propagated through boot). A one-line doc comment saves the next reader a grep.

```suggestion
// EnsureDefaultFunc is invoked at CP boot to ensure default resources exist.
// `extensions` carries logger fields propagated through the boot pipeline.
type EnsureDefaultFunc = func(ctx context.Context, resManager core_manager.ResourceManager, logger logr.Logger, cfg kuma_cp.Config, extensions context.Context) error
```

pkg/core/managers/apis/mesh/mesh_manager.go
```diff
@@ -105,6 +114,9 @@ func (m *meshManager) Create(ctx context.Context, resource core_model.Resource,
 		m.createMeshRoutingResources,
 		m.k8sStore,
 		m.systemNamespace,
+		m.cpMode,
+		m.cpZone,
+		false, // create missing default resources
```
bartsmykla: `// create missing default resources` describes only the create half. The bool is `reconcileExistingOnly`, so `false` means create + reconcile. Mirroring the param name keeps the call site honest.

```suggestion
		false, // reconcileExistingOnly
```

pkg/plugins/runtime/k8s/controllers/mesh_controller.go
```diff
@@ -68,6 +71,9 @@ func (r *MeshReconciler) ensureDefaultResources(ctx context.Context, mesh *core_
 		r.CreateMeshRoutingResources,
 		r.K8sStore,
 		r.SystemNamespace,
+		r.CpMode,
+		r.CpZone,
+		false, // create missing default resources
```
bartsmykla: Same as the `mesh_manager.go` call site - mirror the param name.

```suggestion
		false, // reconcileExistingOnly
```

pkg/defaults/mesh/mesh.go
```diff
@@ -103,15 +109,64 @@ func EnsureDefaultMeshResources(
...
+func ensureDefaultResource(
+	ctx context.Context,
+	resManager manager.ResourceManager,
+	res model.Resource,
+	resourceKey model.ResourceKey,
+	cpMode config_core.CpMode,
+	cpZone string,
+	k8sStore bool,
+	systemNamespace string,
+	reconcileExistingOnly bool,
```
bartsmykla, line 121: Twelve positional args now, with a trailing bool that flips behavior. Not blocking, but worth a follow-up to collapse into an options struct - every call site already passes the same tail tuple, and a `Mode` enum would document `reconcileExistingOnly` better than a bare bool.

pkg/defaults/mesh/mesh.go
```diff
@@ -103,15 +109,64 @@ func EnsureDefaultMeshResources(
...
+		desired, err := computeLabels(res.GetMeta().GetLabels())
+		if err != nil {
+			return errors.Wrap(err, "could not compute labels for a default resource"), false
+		}
+		if maps.Equal(res.GetMeta().GetLabels(), desired) {
+			return nil, false
+		}
+		// Reconcile labels on a default resource created by an older CP
+		// version that predates computed labels. Without 'kuma.io/zone' the
+		// resource's KRI carries an empty zone slot and /_kri lookups fail.
```
bartsmykla: The function-level comment on `EnsureDefaultMeshResourcesUpToDate` already tells the KRI/zone-slot story. One line is enough here.

```suggestion
		// Older CP versions persisted these without computed labels. Rewrite them in place.
```

pkg/defaults/mesh/mesh.go
```diff
@@ -103,15 +109,64 @@ func EnsureDefaultMeshResources(
...
+		return resource_labels.Compute(
+			res.Descriptor(),
+			res.GetSpec(),
+			existing,
+			resourceKey.Mesh,
+			resource_labels.WithMode(cpMode),
+			resource_labels.WithZone(cpZone),
+			resource_labels.WithK8s(k8sStore),
+			resource_labels.WithNamespace(namespace),
+		)
```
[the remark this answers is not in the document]
Automaat: False positive. These default policies (MeshTimeout/MeshCircuitBreaker/MeshRetry) have `KDSFlags = GlobalToZonesFlag | ZoneToGlobalFlag | SyncedAcrossZonesFlag`, and `ZoneToGlobalFlag = ConsumedByGlobalFlag | ProvidedByZoneFlag` (pkg/core/resources/model/resource.go:72). So `KDSFlags.Has(ProvidedByZoneFlag)` is true and `labels.Compute` does set `kuma.io/origin`/`kuma.io/zone`/`kuma.io/env` in Zone mode. No change needed.

pkg/defaults/mesh/mesh.go
Automaat: Agreed it is worth a follow-up. Keeping it out of this PR to stay minimal/surgical; the bool comments now mirror the param name as suggested. Will track the options-struct refactor separately.

bartsmykla (approved):

PR #16646, opened 2026-05-16
fix(core): fix system name prefix check

immanuwell:
## Motivation

`IsSystem` had the `HasPrefix` args flipped. Tiny one, but it meant `system_foo` was not seen as system while `sys` and empty string were. oof

## Implementation information

Swap the args and add a small regression test.

Repro before this patch:

```txt
go test ./pkg/core/system_names
IsSystem("system_foo") = false, want true
IsSystem("sys") = true, want false
IsSystem("") = true, want false
```

Checked after:

```txt
go test ./pkg/core/system_names
go test ./pkg/core/...
```

## Supporting documentation

No linked issue.

> Changelog: skip

### DIFF
--- a/pkg/core/system_names/system_names.go
+++ b/pkg/core/system_names/system_names.go
@@ -16,7 +16,7 @@ const (
 var cleanNameRegex = regexp.MustCompile(`([a-z0-9-]*_?)+`)
 
 func IsSystem(name string) bool {
-	return strings.HasPrefix(SystemPrefix, name)
+	return strings.HasPrefix(name, SystemPrefix)
 }
 
 func CleanName(name string) string {
--- a/pkg/core/system_names/system_names_test.go
+++ b/pkg/core/system_names/system_names_test.go
@@ -1,0 +1,28 @@
+package system_names_test
+
+import (
+	"testing"
+
+	"github.com/kumahq/kuma/v2/pkg/core/system_names"
+)
+
+func TestIsSystem(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/core/system_names/system_names_test.go
```diff
@@ -0,0 +1,28 @@
+package system_names_test
+
+import (
+	"testing"
+
+	"github.com/kumahq/kuma/v2/pkg/core/system_names"
```
Automaat, line 6: **blocking:** Test imports `github.com/kumahq/kuma/v2/...`, but master has migrated to `v3` and still carries the bug. Once rebased onto current master, this package path won't resolve and the build breaks. Rebase and change the import to `v3`.

pkg/core/system_names/system_names.go
```diff
@@ -16,7 +16,7 @@ const (
 var cleanNameRegex = regexp.MustCompile(`([a-z0-9-]*_?)+`)
 
 func IsSystem(name string) bool {
-	return strings.HasPrefix(SystemPrefix, name)
+	return strings.HasPrefix(name, SystemPrefix)
```
Automaat, line 19: **praise:** Correct fix for a genuine inverted-argument bug, with a clear repro in the description and a complete table-driven regression test covering the prefix, exact-match, and empty boundaries. Exactly the right shape for a one-line correctness fix.

Automaat (changes requested):

PR #16647, opened 2026-05-17
feat(matches): add `matches` to shared `inbound.Rule` struct

lobkovilya:
Depends on https://github.com/kumahq/kuma/pull/16575

## Motivation

Common code that enables adding matches for policies like MeshAccessLog, MeshRateLimit, MeshTimeout.

## Implementation information

* add `Matches` to `inbound.Rule`
* change `Conf` type from `RuleEntry` to `any`


pkg/plugins/policies/core/rules/inbound/inboundrules.go
```diff
@@ -9,17 +10,18 @@ import (
...
 func MatchesAllIncomingTraffic[T any](rules []*Rule) T {
 	var result T
 	if len(rules) > 0 {
-		entries := util_slices.Map(rules, func(a *Rule) common.WithPolicyAttributes[RuleEntry] {
-			return common.WithPolicyAttributes[RuleEntry]{Entry: a.Conf}
+		confs := util_slices.Map(rules, func(a *Rule) any {
+			return a.Conf
 		})
-		conf, err := merge.Entries(entries)
+		conf, err := merge.Confs(confs)
```
slonka: the function is named `MatchesAllIncomingTraffic` but `Matches` can filter it down to some SNI or spiffe ID, should the function be renamed or should we guard against not having `matches` here?

pkg/plugins/policies/core/rules/inbound/sort.go
```diff
@@ -13,3 +15,63 @@ func Sort[T common.PolicyAttributes](list []T) {
...
+}
+
+func CompareByMatches(a, b []common_api.Match) int {
+	var ma, mb common_api.Match
+	if len(a) > 0 {
+		ma = a[0]
+	}
+	if len(b) > 0 {
+		mb = b[0]
+	}
```
slonka: why are we always taking 0th Match?

pkg/plugins/policies/core/rules/inbound/inboundrules.go
```diff
@@ -9,17 +10,18 @@ import (
 )
 
 type Rule struct {
-	Conf   RuleEntry     `json:"conf"`
-	Origin common.Origin `json:"origin"`
+	Matches []common_api.Match `json:"matches,omitempty"`
```
lukidzi: Should we have here `omitempty`?

slonka (commented):
Only small Qs

pkg/plugins/policies/core/rules/inbound/sort.go
lobkovilya: I actually decided to change the Rule and always use singular match instead of an array.

lukidzi (approved):

PR #16673, opened 2026-05-18
fix(motb): align env defaults with documentation

bartsmykla:
## Motivation

The `MeshOpenTelemetryBackend.spec.env` block has a documented default of `mode: Optional`, `precedence: EnvFirst`, `allowSignalOverrides: true` - encoded in the `DefaultAllowSignalOverrides` constant and the `EffectiveAllowSignalOverrides()` helper. Two gaps came up while reviewing the companion website PR:

1. The proto field comments only said what each field controls, not how the values behave - so the autogenerated CRD/OpenAPI descriptions were too thin to replace the manual website table.
2. `BuildSignalRuntimePlan` defaulted `allowOverrides = false` when the whole `env:` block was omitted (resolved policy nil), contradicting the documented default. This caused `SignalOverridesDisallowed` to spuriously appear in `blockedReasons` for backends with no `env:` block whenever the sidecar carried per-signal `OTEL_EXPORTER_OTLP_*_*` env vars.

## Implementation information

- Expanded Godoc comments on `EnvPolicy.Mode`, `EnvPolicy.Precedence`, `EnvPolicy.AllowSignalOverrides`, and the parent `Env` field to spell out per-value behaviour and the relevant `blockedReasons`. Regenerated CRDs and `rest.yaml`.
- Fixed `BuildSignalRuntimePlan` to default `mode` and `allowOverrides` from the named API constants (`DefaultEnvMode`, `DefaultAllowSignalOverrides`) instead of inline `EnvModeOptional` / `false`, so the nil-env-block path matches the documented default. Added a regression test for the omitted-env-block case.

## Supporting documentation

Companion website PR: kumahq/kuma-website#2694. The default-mismatch was caught by an automated reviewer comment on this PR's earlier doc-only revision.

> Changelog: feat(kuma-cp): add MeshOpenTelemetryBackend for shared policy-based OpenTelemetry backends

### DIFF
--- a/pkg/core/resources/apis/meshopentelemetrybackend/api/v1alpha1/meshopentelemetrybackend.go
+++ b/pkg/core/resources/apis/meshopentelemetrybackend/api/v1alpha1/meshopentelemetrybackend.go
@@ -26,6 +26,8 @@ type MeshOpenTelemetryBackend struct {
 	Protocol *Protocol `json:"protocol,omitempty"`
 	// Env controls whether standard OTEL exporter env vars participate in the
 	// final exporter config for this backend.
+	// Defaults to mode: Optional, precedence: EnvFirst, allowSignalOverrides: true
+	// when omitted.
 	// +kubebuilder:validation:Optional
 	Env *EnvPolicy `json:"env,omitempty"`
 }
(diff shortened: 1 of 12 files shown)


pkg/core/resources/apis/meshopentelemetrybackend/api/v1alpha1/meshopentelemetrybackend.go
```diff
@@ -26,6 +26,8 @@ type MeshOpenTelemetryBackend struct {
 	Protocol *Protocol `json:"protocol,omitempty"`
 	// Env controls whether standard OTEL exporter env vars participate in the
 	// final exporter config for this backend.
+	// Defaults to mode: Optional, precedence: EnvFirst, allowSignalOverrides: true
```
Automaat, line 29: **issue:** The parent `Env` comment states the omitted-block default for `allowSignalOverrides` is `true`. The whole-block-omitted path does not produce that. When `spec.Env` is nil, `ResolveEnvPolicy()` returns nil, and `BuildSignalRuntimePlan()` falls back to `allowOverrides := false` (`pkg/plugins/policies/core/xds/otel.go:206-209`) — so per-signal override env vars get the `SignalOverridesDisallowed` soft block. The `true` default (`DefaultAllowSignalOverrides`) only applies via `EffectiveAllowSignalOverrides()`, which runs only when an `env` block is present but the `allowSignalOverrides` subfield is omitted. `mode` and `precedence` are unaffected — both resolve to the same value in either case. Either narrow the sentence to the present-block/omitted-subfield case, or document that omitting the entire `env` block disables per-signal overrides.

pkg/core/resources/apis/meshopentelemetrybackend/api/v1alpha1/meshopentelemetrybackend.go
```diff
@@ -26,6 +26,8 @@ type MeshOpenTelemetryBackend struct {
...
+	// when omitted.
```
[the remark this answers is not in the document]
bartsmykla: good catch - real inconsistency, not just a docs issue. BuildSignalRuntimePlan was defaulting allowOverrides to false when the resolved env policy was nil, even though DefaultAllowSignalOverrides and EffectiveAllowSignalOverrides both say true. fixed the runtime to use the named API constants (DefaultEnvMode and DefaultAllowSignalOverrides) so the nil-env-block path matches the documented default. added a regression test covering the omitted-env-block case. retitled the PR to reflect that it is now an alignment fix rather than docs-only.

pkg/core/resources/apis/meshopentelemetrybackend/api/v1alpha1/meshopentelemetrybackend.go
bartsmykla: good catch - same root cause copilot flagged, and the right call. fixed it: BuildSignalRuntimePlan now uses the named API constants (DefaultEnvMode, DefaultAllowSignalOverrides) for the nil-envPolicy path, so the omitted-env-block case matches the documented default. added a regression test for it. retitled the PR to reflect that it is now an alignment fix rather than docs-only.

Automaat (approved):

PR #16677, opened 2026-05-18
fix(xds): add a feature flag to enable reuse ports

### LINKED ISSUE #16675: Upgrading kuma breaks with ReusablePort
### Kuma Version

master

### Describe the bug

You cannot change reusablePort while running, we should guard it by dataplane feature flag

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

lukidzi:
## Motivation

In [PR #16501](https://github.com/kumahq/kuma/pull/16501), we set `enable_reuse_port` to its default value (true). Unfortunately, this setting cannot be changed at runtime and requires the dataplane to be restarted.

```
inbound:passthrough:ipv6: Listener inbound:passthrough:ipv6: reuse port cannot be changed during an update
```

## Implementation information

Added `FeatureReusePort` as a new capability that kuma-dp advertises to the CP during bootstrap. The CP reads this feature flag when building inbound listeners and sets `enable_reuse_port` accordingly.

Changes:
  - `DataplaneRuntime.ReusePortEnabled` config field (env: `KUMA_DATAPLANE_RUNTIME_REUSE_PORT_ENABLED`, default `true`) controls whether the feature is enabled
  - InboundListener and NewInboundListenerBuilder now explicitly set `enable_reuse_port` to true when `FeatureReusePorts` feature is set
  - UDP listeners always get `enable_reuse_port=true` regardless of the flag: required for correct multi-worker UDP handling
  - Added E2E test which checks if value is correctly interpreted
  - 
## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/16675


bartsmykla (dismissed):
A few things need tightening before this lands

pkg/plugins/runtime/gateway/listener_generator.go
```diff
@@ -59,6 +60,7 @@ func GenerateListener(info GatewayListenerInfo) (*envoy_listeners.ListenerBuilde
 		address,
 		port,
 		core_xds.SocketAddressProtocolTCP,
+		info.Proxy.Metadata.HasFeature(xds_types.FeatureReusePort),
```
bartsmykla, line 63: This gets overwritten by `EnableReusePort(true)` below, so gateways still ignore the disabled flag.

UPGRADE.md
```diff
@@ -8,6 +8,56 @@ does not have any particular instructions.
 
 ## Upgrade to `2.14.x`
 
+### Inbound listeners now use SO_REUSEPORT by default
+
+The data plane now advertises the `feature-reuse-ports` capability to the control plane, which causes inbound Envoy listeners to be generated with `enable_reuse_port: true`. This lets each Envoy worker thread own its own listen socket, improving connection distribution under load.
```
bartsmykla: The capability is `feature-reuse-port` in code, not `feature-reuse-ports`.

UPGRADE.md
```diff
@@ -8,6 +8,56 @@ does not have any particular instructions.
...
+Then set the annotation `kuma.io/container-patches` on deployments where it should be disabled:
+
+```yaml
+"kuma.io/container-patches": "disable-reuse-ports"
+```
+
+or globally for all injected sidecars via control-plane configuration:
+
+```
+KUMA_RUNTIME_KUBERNETES_INJECTOR_CONTAINER_PATCHES="disable-reuse-ports"
```
bartsmykla: This only covers injected sidecars. ZoneIngress and ZoneEgress need separate env guidance too.

bartsmykla (commented):
A few things need tightening before this lands

pkg/plugins/runtime/gateway/listener_generator.go
lukidzi: It needs to be like this. As old dpp won't have this flag it would be false, but this setting of true is there since 5 years so we cannot change it as it could cause a failure.

bartsmykla (approved):

PR #16686, opened 2026-05-19
feat(xds): delta xDS Helm + injection (bp 16392)

bartsmykla:
## Motivation

Backports #16392 to `release-2.13`.

The `KUMA_EXPERIMENTAL_DELTA_XDS` env var is a no-op on 2.13.x because the k8s injector never propagates `KUMA_DATAPLANE_RUNTIME_ENVOY_XDS_TRANSPORT_PROTOCOL_VARIANT=DELTA_GRPC` to injected sidecars. Today the only workaround is per-deployment `kuma.io/xds-transport-protocol-variant: DELTA_GRPC`. This lets operators flip delta xDS at the CP level via the standard chart value.

## Implementation information

Cherry-pick of 585975b80 with conflict resolution against 2.13:

- `experimental.inboundTagsDisabled` and `experimental.envoyAdminUnixSocket` blocks were dropped from `values.yaml`, `_helpers.tpl`, `README.md`, `install-control-plane.dump-values.yaml`, and `helm-values.yaml` (they don't exist on 2.13). Only `experimental.deltaXds` is added.
- In `containers/factory.go`, the `if i.envoyAdminUnixSocket` env-var block was dropped for the same reason.
- `injector.New()` on 2.13 ends at `systemNamespace string`, so the new `deltaXdsEnabled bool` is appended directly (without master's `envoyReadinessPort`, `envoyAdminUnixSocket`, and `metrics` params).
- `NewDataplaneProxyFactory()` calls in `injector.go` and `plugin_gateway.go` append `deltaXdsEnabled` after `cfg.Spire.Enabled` (master's `OtelPipeEnabled` slot doesn't exist on 2.13).
- Same shape adjustment in `injector_test.go`.

## Supporting documentation

- Original PR: #16392
- Fixes: #16405

### DIFF
--- a/docs/generated/raw/helm-values.yaml
+++ b/docs/generated/raw/helm-values.yaml
@@ -940,6 +940,8 @@ experimental:
   # -- If true, enable native Kubernetes sidecars. This requires at least
   # Kubernetes v1.29
   sidecarContainers: false
+  # -- If true, uses Delta xDS (incremental) protocol to deliver changes to sidecars instead of State of the World (SOTW).
+  deltaXds: false
 
 # Postgres' settings for universal control plane on k8s
 postgres:
--- a/pkg/plugins/runtime/k8s/webhooks/injector/injector_test.go
+++ b/pkg/plugins/runtime/k8s/webhooks/injector/injector_test.go
@@ -92,7 +92,7 @@ spec:
 				var cfg conf.Injector
 				Expect(config.Load(filepath.Join("testdata", given.cfgFile), &cfg)).To(Succeed())
 				cfg.CaCertFile = caCertPath
-				injector, err := inject.New(cfg, "http://kuma-control-plane.kuma-system:5681", k8sClient, sidecarsEnabled, k8s.NewSimpleConverter(), 9901, systemNamespace)
+				injector, err := inject.New(cfg, "http://kuma-control-plane.kuma-system:5681", k8sClient, sidecarsEnabled, k8s.NewSimpleConverter(), 9901, systemNamespace, false)
 				Expect(err).ToNot(HaveOccurred())
 
 				// and create mesh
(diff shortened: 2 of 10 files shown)


docs/generated/raw/helm-values.yaml
```diff
@@ -940,6 +940,8 @@ experimental:
   # -- If true, enable native Kubernetes sidecars. This requires at least
   # Kubernetes v1.29
   sidecarContainers: false
+  # -- If true, uses Delta xDS (incremental) protocol to deliver changes to sidecars instead of State of the World (SOTW).
```
Automaat, line 943: **suggestion:** `README.md`, `docs/generated/raw/helm-values.yaml`, and
`install-control-plane.dump-values.yaml` are generated artifacts. The hand-
edits look correct and consistent, but confirm `make` (the chart-docs / dump-
values generator) reproduces exactly these lines so the next regen doesn't
produce a spurious diff.

pkg/plugins/runtime/k8s/webhooks/injector/injector_test.go
```diff
@@ -1064,4 +1064,119 @@ spec:
 			cfgFile: "inject.config.yaml",
 		}),
 	)
+
```
Automaat, line 1067: **praise:** Precedence-focused test coverage — verifies CP default, per-pod
annotation override, and `sidecar-env-vars` override rather than just the
happy path.

Automaat (approved):
Approve — clean, tightly-scoped backport of #16392. The PR description documents every conflict-resolution decision against 2.13, and I verified all four call sites of the changed signatures are updated and the `Experimental.DeltaXds` config field already exists on the branch. Default is `false` everywhere, so existing installs are unaffected. Two non-blocking notes: confirm the generated docs match `make` output, and that the toggle only affects newly-injected pods. No blockers.

docs/generated/raw/helm-values.yaml
bartsmykla: Verified - ran `make helm-docs docs/generated/raw` and `UPDATE_GOLDEN_FILES=true go test ./app/kumactl/cmd/install/...`, both produce zero diff against the hand-edits.

PR #16688, opened 2026-05-19
feat(meshservice): per-workload MeshService gen

Automaat:
## Motivation

After the planned removal of `Dataplane` inbound tags, the MeshService
generator currently does nothing on Universal — it bails out early when
inbound tags are disabled, leaving operators to author every `MeshService`
by hand.

Offline decision (see MADR below, Option B): when running without inbound
tags, the control plane should generate one `MeshService` per `Workload`,
i.e. a 1:1 mapping between a `kuma.io/workload` label value and a generated
`MeshService`. This change implements that so the approach can be tested
end to end.

## Implementation information

- `meshservice/generate/component.go`: dropped the `InboundTagsDisabled`
  early-return; the generator now runs in both modes.
- `meshservice/generate/generator.go`: `meshServicesForDataplane` dispatches
  on `inboundTagsDisabled`:
  - disabled → new `workloadMeshServiceForDataplane`: one `MeshService` per
    `kuma.io/workload` value, named after the workload, one `Port` per inbound
    (name/protocol taken from the inbound fields), `Selector.DataplaneLabels`
    matching the workload label.
  - enabled → existing `kuma.io/service`-tag flow, unchanged.
- The reconcile loop, port-consistency check (first-DP-wins), grace-period
  delete and label propagation are keyed by name and reused as-is. Non-`kuma.io/*`
  `Dataplane.metadata.labels` still propagate to the generated `MeshService`.

Replicas of one workload normally have identical inbounds; on a port mismatch
the existing first-DP-wins consistency check applies. A union-of-ports policy
was considered and rejected as a larger change for no real-world gain.

## Supporting documentation

- MADR: docs/madr/decisions/103-universal-meshservice-generation.md

### DIFF
--- a/pkg/core/resources/apis/meshservice/generate/component.go
+++ b/pkg/core/resources/apis/meshservice/generate/component.go
@@ -26,10 +26,6 @@ func Setup(rt runtime.Runtime) error {
 		logger.Info("MeshServices are not enabled. Skip starting generator for MeshServices.")
 		return nil
 	}
-	if rt.Config().Experimental.InboundTagsDisabled {
-		logger.Info("Inbound tags are disabled. Skip starting generator for MeshService. Users should create MeshService manually.")
-		return nil
-	}
 	generator, err := New(
 		logger,
 		rt.Config().MeshService.GenerationInterval.Duration,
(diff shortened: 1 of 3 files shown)


bartsmykla (approved):
Looks good, but please either add the referenced MADR or remove that link from the PR body.

PR #16760, opened 2026-05-25
fix(xds): set unknown cluster name when no MS available

### LINKED ISSUE #16764: Investigate existing NACKs in e2e
Continuation from: https://github.com/kumahq/kuma/pull/16759#discussion_r3297779007

TLDR: we enabled NACKs check, it started failing on `master`, we increased the count but let's investigate what happened

lukidzi:
## Motivation

 The new NACK assertion (#16713) caught a `RouteConfiguration` NACK in the multizone suite. CP logs:
```
  plugin.runtime.gateway  skipping backendRef  {"backendRef":{"Ref":{"Resource":"kri_msvc_msconnectivity_..."}}}
  xds.status-tracker      config rejected      {"type":"RouteConfiguration"}
  xds.nack-backoff        ... reason: "At least one of name or cluster_header need to be specified"
```
 When a gateway route's backendRef points to a real resource (MeshService / MeshMultiZoneService) that isn't in meshCtx at generation time — typically a transient window during namespace teardown or KDS resync — `meshroute.DestinationPortFromRef` returns `!ok`. The route entry was left with no Destination.Name, the cluster generator silently continued, and Envoy received a `WeightedCluster{Name: ""}` → NACK on the whole RouteConfiguration.

The legacy/tag path has handled this since #11461 by stamping a sentinel cluster name (`kuma.io/unresolved-backend`) and relying on `ClusterNotFoundResponseCode: INTERNAL_SERVER_ERROR` at request time. The real-resource path was missing the same treatment.

## Implementation information

 Two spots, both stamp the same sentinel the legacy path already uses:

  - `meshhttproute/.../gateway_routes.go` — when `DestinationPortFromRef` returns `!ok` for a real ref, set `dest = {ServiceTag: UnresolvedBackendServiceTag}` instead of leaving it nil.
  - `runtime/gateway/cluster_generator.go` — when `generateRealBackendRefCluster` returns `(nil, nil)`, set `dest.Name = UnresolvedBackendServiceTag` before `continue`, mirroring the existing legacy branch a few lines below.

Envoy doesn't cross-validate RDS against CDS at push time, so a route referencing a non-existent cluster is accepted. At request time the lookup misses and Envoy returns 500 via the already-configured `ClusterNotFoundResponseCode`. Same contract as the existing `/no-matching-backend` test.

fix: https://github.com/kumahq/kuma/issues/16764


test/framework/debug.go
```diff
@@ -74,7 +74,7 @@ func assertNoXdsNacks(cluster Cluster) {
 			if !isNack {
 				continue
 			}
-			Expect(metric.GetCounter().GetValue()).To(BeNumerically("<", 3),
+			Expect(metric.GetCounter().GetValue()).To(BeZero(),
```
slonka: I don't think we should use `BeZero()` here because there are other NACKs beyond the `RouteConfiguration` one this PR fixes. In https://github.com/kumahq/kuma/actions/runs/26400191664/job/77711885587#step:4:5665 still hit `config rejected {"type":"Secret"}` and `xds.nack-backoff ... reason: "DataSource cannot be empty"`. So this PR can fix the `RouteConfiguration` NACK and still leave this assertion failing.

lukidzi:
```suggestion
			Expect(metric.GetCounter().GetValue()).To(BeNumerically("<", 3),
```

slonka (approved):
We still have one more NACK (at least).

Issue #16763, opened 2026-05-26
MeshAccessLog: consider flush_access_log_on_connected for sidecars


lobkovilya:
The recent MAL change sets `tcp_proxy.access_log_options.flush_access_log_on_connected: true` on mesh-scoped zone proxies, so the log fires on connect instead of only on close (long-lived mTLS tunnels would otherwise never emit).

Sidecar inbound/outbound TCP listeners have the same problem — pooled, long-lived connections mean MAL on a TCP service looks silent until idle-timeout (default 1h).

Worth considering enabling the same flag there. We might also want `access_log_flush_interval` for periodic mid-connection emissions, so observability of long-lived TCP streams isn't bound to the connect/close moments only.

**Breaking-change question to settle:**
- Doubles (or more, with flush interval) the log volume per connection.
- Close-time fields (`%DURATION%`, `%BYTES_SENT%`, `%RESPONSE_FLAGS%`) are zero/empty on the on-connect record and partial on interval flushes — default Kuma TCP format includes all of these.

Decide: on by default, opt-in via MAL field, or opt-out.

lukidzi:
> Triage: there is also `flush_access_log_on_start` https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto#extensions-filters-network-tcp-proxy-v3-tcpproxy-tcpaccesslogoptions

PR #16765, opened 2026-05-26
feat(MeshAccessLog): support matches

### LINKED ISSUE #16689: MeshAccessLog: add `rules[].matches[]` support (including zone proxies)
### Description

See [MADR-103](https://github.com/kumahq/kuma/blob/dc4a80ce1a73fe1c48bbc2d8c0bb7363b696f5c4/docs/madr/decisions/103-policy-matching-mesh-scoped-zone-proxy.md)

lobkovilya:
Depends on https://github.com/kumahq/kuma/pull/16740

## Motivation

`MeshAccessLog`'s inbound `rules` previously had no way to scope a configuration to a subset of inbound traffic — every rule applied to every inbound connection. This PR adds a `matches` field so a rule can select traffic by client SpiffeID (exact or prefix) and/or SNI (exact), enabling per-client or per-SNI access log routing on sidecars, ZoneIngress, and ZoneEgress.

## Implementation information

- Added `Rule.Matches []common_api.Match` to the MeshAccessLog API (SpiffeID `Exact`/`Prefix`, SNI `Exact`) plus validation; regenerated CRDs, OpenAPI, and helpers.
- First-match-wins is encoded as a CEL `ExpressionFilter` on each Envoy `AccessLog` entry: rule N's filter is `selfMatch && !prior0 && !prior1 ...` — Envoy has no native "stop at first match" for access loggers, so the precedence is pushed into each filter.
  - `MatchToCEL` converts a `Match` to a CEL conjunction using `connection.uri_san_peer_certificate` and `connection.requested_server_name`.

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/16689


pkg/plugins/policies/meshaccesslog/api/v1alpha1/validator.go
```diff
@@ -70,11 +70,21 @@ func validateRules(rules []Rule) validators.ValidationError {
...
 		verr.AddErrorAt(path.Field("default"), validateDefault(rule.Default))
 	}
 	return verr
 }
 
+func validateMatches(field string, matches []common_api.Match) validators.ValidationError {
+	var verr validators.ValidationError
+	for idx, match := range matches {
+		path := validators.RootedAt(field).Index(idx)
+		verr.AddErrorAt(path, mesh.ValidateMatch(match))
```
slonka, line 83: `mesh.ValidateMatch()` [only validates fields that are present](https://github.com/slonka/kuma/blob/15febb7236d2c2ddf2c0eeb4dd5e72d4d357e3da/pkg/core/resources/apis/mesh/validators.go#L458), so `matches: [{}]` would pass here, right? Would that then emit an unfiltered catch-all access log? Is it intentional? `meshtimeout` guards against this with an explicit `match.SpiffeID == nil && match.SNI == nil` check - should we do the same here? Or should we even patch `mesh.ValidateMatch`?

pkg/plugins/policies/meshaccesslog/plugin/v1alpha1/plugin.go
```diff
@@ -187,7 +191,56 @@ func applyToInbounds(
...
+		kumaValues := listeners_v3.KumaValues{
+			SourceService:      mesh_proto.ServiceUnknown,
+			SourceIP:           dataplane.GetIP(),
+			DestinationService: mesh_proto.ServiceUnknown,
+			Mesh:               dataplane.GetMeta().GetMesh(),
+			Zone:               zone,
+			WorkloadKRI:        workloadKRI,
+			TrafficDirection:   envoy.TrafficDirectionUnspecified,
+		}
+		if err := configureListenerFromRules(inboundRules, listener, backends, DefaultFormat(core_meta.ProtocolTCP), kumaValues, accessLogSocketPath); err != nil {
```
slonka, line 236: does zone ingress have access to `uri_san_peer_certificate` (which is used in `MatchToCEL`)? it does not terminate mTLS, so it never has the peer cert URI SAN right?

slonka (commented):
did a first pass, will do another one but not sure if today / tomorrow, just posting to give early feedback

deployments/charts/kuma/crds/kuma.io_meshaccesslogs.yaml
```diff
@@ -510,6 +508,53 @@ spec:
...
+                            properties:
+                              type:
+                                description: Type defines how to match incoming traffic
+                                  by SpiffeID. `Exact` or `Prefix` are allowed.
+                                enum:
+                                - Exact
+                                - Prefix
+                                type: string
+                              value:
+                                description: Value is SpiffeId of a client that needs
```
lukidzi:
```suggestion
                                description: Value is SpiffeID of a client that needs
```

docs/generated/openapi.yaml
```diff
@@ -7108,6 +7103,64 @@ components:
...
+                              description: >-
+                                Type defines how to match incoming traffic by
+                                SpiffeID. `Exact` or `Prefix` are allowed.
+                              enum:
+                                - Exact
+                                - Prefix
+                              type: string
+                            value:
+                              description: >-
+                                Value is SpiffeId of a client that needs to
```
lukidzi:
```suggestion
                                Value is SpiffeID of a client that needs to
```

docs/generated/raw/crds/kuma.io_meshaccesslogs.yaml
```diff
@@ -510,6 +508,53 @@ spec:
...
+                            properties:
+                              type:
+                                description: Type defines how to match incoming traffic
+                                  by SpiffeID. `Exact` or `Prefix` are allowed.
+                                enum:
+                                - Exact
+                                - Prefix
+                                type: string
+                              value:
+                                description: Value is SpiffeId of a client that needs
```
lukidzi:
```suggestion
                                description: Value is SpiffeID of a client that needs
```

pkg/envoy/builders/listener/listener.go
```diff
@@ -45,6 +42,28 @@ func AccessLogs(builders []*Builder[envoy_accesslog.AccessLog]) Configurer[envoy
...
+// listener to emit its access log entries as soon as a downstream connection
+// is established, in addition to the default emission on close. Required for
+// long-lived tunnels (e.g. cross-zone sidecar→zone-ingress mTLS) where the
+// connection-close log would otherwise never fire in practice.
+func FlushTCPProxyAccessLogOnConnected() Configurer[envoy_listener.Listener] {
+	return func(l *envoy_listener.Listener) error {
+		for _, chain := range l.FilterChains {
+			if err := listeners_v3.UpdateTCPProxy(chain, func(tcpProxy *envoy_tcp.TcpProxy) error {
+				if tcpProxy.AccessLogOptions == nil {
+					tcpProxy.AccessLogOptions = &envoy_tcp.TcpProxy_TcpAccessLogOptions{}
```
lukidzi, line 55: Questions here:
* Do we need `flush_access_log_on_start ` as we could know if connection fails on the initialisation? 
 > If set to true, the access log is flushed when the TCP proxy accepts a connection.

* Do we need to set `access_log_flush_interval ` as it stands ` By default, the TCP proxy flushes a single access log when the connection is closed`
> The interval for flushing access logs. By default, the TCP proxy flushes a single access log when the connection is closed. If this field is set, the TCP proxy flushes access logs periodically at the specified interval. The interval must be at least 1ms.

pkg/plugins/policies/meshaccesslog/api/v1alpha1/rest.yaml
```diff
@@ -602,6 +600,46 @@ components:
...
+                          description: SpiffeID defines a matcher configuration for SpiffeID matching
+                          properties:
+                            type:
+                              description: Type defines how to match incoming traffic by SpiffeID. `Exact` or `Prefix` are allowed.
+                              enum:
+                                - Exact
+                                - Prefix
+                              type: string
+                            value:
+                              description: Value is SpiffeId of a client that needs to match for the configuration to be applied
```
lukidzi:
```suggestion
                              description: Value is SpiffeID of a client that needs to match for the configuration to be applied
```

pkg/plugins/policies/meshaccesslog/k8s/crd/kuma.io_meshaccesslogs.yaml
```diff
@@ -510,6 +508,53 @@ spec:
...
+                            properties:
+                              type:
+                                description: Type defines how to match incoming traffic
+                                  by SpiffeID. `Exact` or `Prefix` are allowed.
+                                enum:
+                                - Exact
+                                - Prefix
+                                type: string
+                              value:
+                                description: Value is SpiffeId of a client that needs
```
lukidzi:
```suggestion
                                description: Value is SpiffeID of a client that needs
```

pkg/plugins/policies/meshaccesslog/plugin/xds/rules.go
```diff
@@ -0,0 +1,48 @@
...
+//
+// TODO: on zone proxies (ZoneIngress/ZoneEgress), filter chains already match
+// per-SNI via MatchServerNames, so encoding `connection.requested_server_name`
+// in CEL duplicates the filter chain match. A future optimization can install
+// access logs on the matching filter chain directly and drop the SNI term from
+// CEL; for now CEL handles both SpiffeID and SNI uniformly across sidecar
+// inbound, ZoneEgress, and ZoneIngress.
+func BuildAccessLogBuildersFromRules(
+	rules []*inbound.Rule,
+	defaultFormat string,
```
lukidzi, line 54: Do we need separate files for cel and from rules? Maybe we just need to combine them to one helper file?

pkg/plugins/policies/meshaccesslog/plugin/xds/cel.go
```diff
@@ -0,0 +1,38 @@
...
+	return strings.Join(parts, " && ")
+}
+
+// celString returns a double-quoted CEL string literal with backslashes and
+// double quotes escaped.
+func celString(v string) string {
+	escaped := strings.ReplaceAll(v, `\`, `\\`)
+	escaped = strings.ReplaceAll(escaped, `"`, `\"`)
+	return `"` + escaped + `"`
+}
```
lukidzi: We try to escape some chars but spiffeID might have more characters `:`, `.` maybe we can use `strconv.Quote`(if properly escape everything)

test/e2e_env/multizone/meshaccesslog/meshaccesslog.go
```diff
@@ -0,0 +1,243 @@
...
+		// Wait for MeshIdentity to sync, then publish per-zone MeshTrust to Global so
+		// KDS distributes it to all zones and cross-zone mTLS is established.
+		hashedIdentityName := hash.HashedName(meshName, "identity")
+		Expect(WaitForResource(
+			meshidentity_api.MeshIdentityResourceTypeDescriptor,
+			model.ResourceKey{Mesh: meshName, Name: hashedIdentityName},
+			multizone.UniZone1, multizone.UniZone2,
+		)).To(Succeed())
+
+		getMeshTrust := func(zoneName string) *meshtrust_api.MeshTrust {
```
slonka, line 143: at some point we really need to move these into some utils, we have:

- test/e2e_env/multizone/meshproxy/connectivity.go:263-296 - near-identical getMeshTrust + installTrustToGlobal
- test/e2e_env/multizone/meshproxy/migration.go:385-417
- test/e2e_env/multizone/meshidentity/meshidentity.go:88-132 - a slightly more general local version with getMeshTrust, waitForMeshTrust, buildMeshTrustYaml, installTrustToZone

pkg/envoy/builders/listener/listener.go
lobkovilya: There is an issue I created https://github.com/kumahq/kuma/issues/16763, we should consider adding `flush_access_log_on_start` and `access_log_flush_interval`.

This PR sets only `flush_access_log_on_connected` because it was easier to write e2e tests that way.

pkg/plugins/policies/meshaccesslog/api/v1alpha1/validator.go
lobkovilya: I checked how we validate `matches` in MeshHTTPRoute https://github.com/kumahq/kuma/blob/d58fd8ac1ada5c170b433c43297a9aee6d106ae2/pkg/plugins/policies/meshhttproute/api/v1alpha1/validation.go#L121-L132

we don't require at least one match to be set.

Normally, I don't like this type of check `match.SpiffeID == nil && match.SNI == nil` because it's very easy to miss when adding a new matcher. Though with AI assisted coding this argument is kind of weak :)

I don't think we should change `MeshHTTPRoute` so if we want API to be coherent it'll be easier to update `MeshTimeout`. We just need to make sure `matches: [{}]` means "catch-all". I'm going to add a test for MAL.

pkg/plugins/policies/meshaccesslog/plugin/v1alpha1/plugin.go
lobkovilya: Yes it doesn't, rule is here but it never matches any traffic. There are few things we can do to reduce the size of Envoy config in some cases and this is one of them. I'll open an issue and try to make it into 2.14.

pkg/plugins/policies/meshaccesslog/plugin/xds/cel.go
lobkovilya: `:` and `.` are not special characters in CEL, but I agree with your point to use `strconv.Quote`

test/e2e_env/multizone/meshaccesslog/meshaccesslog.go
lobkovilya: I agree, let's do this in a separate PR though. I've created an issue https://github.com/kumahq/kuma/issues/16772

lukidzi (approved):

pkg/plugins/policies/meshaccesslog/plugin/v1alpha1/plugin.go
slonka: won't this mean that the user can be confused that their config doesn't match? how are we surfacing that? should we just document it? 🤔

lobkovilya: I think it's exactly the same situation as when you try to apply HTTP policy to the TCP server. We can't do much in that case.

PR #16768, opened 2026-05-26
feat(api-server): introduce SNIs to resources of MS/MES/MMZS

### LINKED ISSUE #16555: Present SNI in the UI
Since we want to use SNI in policy matching it makes sense to persist it and show to the user

lukidzi:
## Motivation

We have introduced matching of SNI for policies, now we need to present users SNI for easier matching creation.

## Implementation information

* Add to response list of SNIs (like we did for KRI) - problem here is that MS/MMZS/MES can have mulitple ports and each port has different SNI (so it's not the same as KRI)
* Added tests
* Added option to introduce SNI field only to resources with `IsDestination` as it should automatically gets for smiliar type in the _future_

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/16555

### DIFF
--- a/pkg/core/resources/model/rest/v1alpha1/resource.go
+++ b/pkg/core/resources/model/rest/v1alpha1/resource.go
@@ -2,10 +2,13 @@ package v1alpha1
 
 import (
 	"encoding/json"
+	"sort"
 	"time"
 
 	"github.com/kumahq/kuma/v2/pkg/core/kri"
 	core_model "github.com/kumahq/kuma/v2/pkg/core/resources/model"
+	"github.com/kumahq/kuma/v2/pkg/core/resources/registry"
+	"github.com/kumahq/kuma/v2/pkg/core/resources/sni"
 )
 
 type Resource struct {
@@ -36,8 +39,10 @@ func (r *Resource) MarshalJSON() ([]byte, error) {
 	}
 
 	var kriStr string
+	var snis []SNIEntry
 	if id, _ := kri.FromResourceMetaE(r.ResourceMeta, r.Type); !id.IsEmpty() {
 		kriStr = id.String()
+		snis = computeSNIs(id, r.Spec)
 	}
 
 	// Explicit struct with all fields in desired order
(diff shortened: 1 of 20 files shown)


tools/openapi/templates/schema.yaml
```diff
@@ -25,6 +25,35 @@ properties:
     type: string
     readOnly: true
     example: 'kri_{{ .ShortName }}_{{ $mesh }}_{{ $zone }}_{{ $ns }}_{{ $name }}_'
+  {{- if .IsDestination }}
+  snis:
+    description: 'List of SNIs (Server Name Indication) advertised by xDS
+      for this destination, one entry per port, sorted by port ascending.
+      Always set for MeshService, MeshMultiZoneService and MeshExternalService;
+      absent for other resource types.'
```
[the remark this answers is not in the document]
slonka: nitpick: Either I don't understand this or the change is not really needed, we can't have any of these without a port, right?

slonka (dismissed):
Can't see anything wrong with this PR 👍

tools/openapi/templates/schema.yaml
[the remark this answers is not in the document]
lukidzi: Yes, I think we have validation that port cannot be empty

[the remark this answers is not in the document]
slonka: so we should remove the

> "when the destination has SNI sections/ports; absent otherwise."

because that can't happen - right?

pkg/core/resources/model/rest/v1alpha1/resource.go
```diff
@@ -60,13 +66,55 @@ func (r *Resource) MarshalJSON() ([]byte, error) {
...
 
 	return json.Marshal(aux)
 }
 
+// SNIEntry pairs a destination port with the SNI advertised by xDS for that
+// port. Section is the section name fed into the trailing segment of the
+// SNI (typically the port name, falling back to the stringified port).
+type SNIEntry struct {
+	Port    int32  `json:"port"`
+	Section string `json:"sectionName"`
```
lobkovilya: why do we need port and section?

lobkovilya: should be `SectionName` if json field is `sectionName`

lukidzi: Since SNI is different for different port it makes sense to present it this way

lukidzi: Maybe here we don't need sectionName let's me remove

bartsmykla (approved):

PR #16774, opened 2026-05-27
test(e2e): add assertion to legacy tests

### LINKED ISSUE #16764: Investigate existing NACKs in e2e

lukidzi:
## Motivation

Check if there are NACK in legacy test to know if we introduced a breaking change

## Implementation information

Added a NACK check to all e2e tests. In helm upgrade I had to scale down ingress and up after starting to avoid situation that Ingress is spawned with a flag for new config and connects to old CP, that causes errors in listener upgrade

## Supporting documentation

Fix #16764


bartsmykla (dismissed):

lukidzi:
We need a new release 2.13.7

slonka:
Recreated against `master` in https://github.com/kumahq/kuma/pull/17000 because the original author is OOO.

PR #16777, opened 2026-05-28
fix(kuma-cp): use system trust when CA cert is not provided

lobkovilya:
## Motivation

We should use system trust and not fallback to insecure behaviour.

## Supporting documentation

N/A


lobkovilya:
@bartsmykla golden files are broken on `master`, waiting on https://github.com/kumahq/kuma/pull/16781

bartsmykla (approved):
No blocking correctness issues found in the changed TLS paths or skip-verify wiring. Targeted package tests passed.

PR #16784, opened 2026-05-28
feat(xds): warn on stream close after large send

Automaat:
## Motivation

When an xDS DiscoveryResponse exceeds gRPC C-Core's
`grpc.max_receive_message_length` (default 4 MiB, which in the google_grpc
transport also sizes the per-stream HTTP/2 receive flow-control window),
the client silently cancels the stream a few hundred milliseconds after
dispatch. The control plane today sees only "stream closed" and operators
have to diagnose it from secondary symptoms (warming clusters, missing
endpoints). This change surfaces the failure directly via a metric and a
log line, so the pattern is visible the moment it happens. Referenced
upstream issue: kumahq/kuma#16355.

## Implementation information

- In `pkg/util/xds/stats_callbacks.go`, track the most recent response
  size, timestamp, and `type_url` per stream alongside the existing
  per-stream state. Update happens on `OnStreamResponse` and
  `OnStreamDeltaResponse` after the existing `responses_sent` increment;
  no behavior change to existing metrics.
- On `OnStreamClosed` / `OnDeltaStreamClosed`: if the stream closed
  within 5s of sending a >=1 MiB response, increment counter
  `<dsType>_stream_likely_window_depletion_total{type_url=...}` and emit
  a log line at info level (logr convention) pointing operators at
  `KUMA_BOOTSTRAP_SERVER_PARAMS_XDS_GRPC_MAX_RECEIVE_MESSAGE_BYTES`.
- Defaults: 5s closure window, 1 MiB size threshold. Window leaves
  headroom over the few-hundred-ms cancel delay reported in the field;
  threshold sits well below the 4 MiB gRPC default to also catch
  configurations that already tightened the limit.
- Response size is computed via `proto.Size` on the underlying
  Discovery(Delta)Response, with a fallback that sums resources.
- Tests cover: large send + close-in-window fires the counter; small
  send does not; large send + close-after-window does not; no send +
  close does not; delta variant works the same way.

> Changelog: feat(xds): warn on stream close after large send


lahabana (commented):
That's interesting but I'm a little worried of the complexity here.

Could we consider doing something simpler:

- monitor closed xds streams (I think we possibly have this already)
- log when sending response over 2MB (with the DP name). A bit like a slow query logger in sql

pkg/util/xds/stats_callbacks.go
```diff
@@ -207,12 +243,16 @@ func (s *statsCallbacks) OnDeltaStreamOpen(context.Context, int64, string) error
...
 	s.Lock()
-	defer s.Unlock()
 	s.streamsActive--
 	if ver, ok := s.versionsForStream[streamID]; ok {
 		s.versionsMetric.WithLabelValues(ver).Dec()
 		delete(s.versionsForStream, streamID)
 	}
+	last, hadSend := s.lastSendByStream[streamID]
+	delete(s.lastSendByStream, streamID)
+	s.Unlock()
```
lahabana: LLMs are overly sensitive at logging inside critical sections. IMO I've seen more issues with missing/misplaced release than performance issues with logging inside a critical sections

Therefore, I think keep the `defer` is ok and less likely to cause issue than having it here.

pkg/util/xds/stats_callbacks.go
```diff
@@ -243,8 +283,43 @@ func (s *statsCallbacks) OnStreamDeltaResponse(streamID int64, request DeltaDisc
...
+}
+
+func (s *statsCallbacks) recordLastSend(streamID int64, size int, typeURL string) {
+	s.Lock()
+	s.lastSendByStream[streamID] = lastSend{
+		size:    size,
+		sentAt:  core.Now(),
+		typeURL: typeURL,
+	}
+	s.Unlock()
```
lahabana: defer s.Unlock please

pkg/util/xds/stats_callbacks.go
```diff
@@ -243,8 +283,43 @@ func (s *statsCallbacks) OnStreamDeltaResponse(streamID int64, request DeltaDisc
...
+	}
+	elapsed := core.Now().Sub(last.sentAt)
+	if elapsed >= closureWindow || last.size < sizeThreshold {
+		return
+	}
+	s.likelyWindowDepletionMetric.WithLabelValues(last.typeURL).Inc()
+	statsLogger.Info(
+		"xds stream closed within window of sending a large response; likely caller-side gRPC receive-window depletion (kumahq/kuma#16355). "+
+			"If using google_grpc xDS transport, raise KUMA_BOOTSTRAP_SERVER_PARAMS_XDS_GRPC_MAX_RECEIVE_MESSAGE_BYTES.",
+		"streamID", streamID,
```
lahabana: streamID isn't going to be super useful to user. Can we get the DP name?

bartsmykla (dismissed):

pkg/util/xds/stats_callbacks.go
Automaat: Done — `recordLastSend` now uses `defer s.Unlock()`.

> _Reply drafted by Claude._

pkg/util/xds/stats_callbacks.go
Automaat: Good point — now logging the dataplane node id instead of the stream id. `lastSend` carries the `nodeID` (captured from `request.NodeId()`), and the log line emits `"dataplane": <node id>`.

> _Reply drafted by Claude._

pkg/util/xds/stats_callbacks.go
Automaat: Agreed — reverted to `defer s.Unlock()` and moved the report call back inside the critical section (it's just a counter increment plus a log line, cheap enough to keep under the lock).

> _Reply drafted by Claude._

pkg/util/xds/stats_callbacks.go
```diff
@@ -247,8 +288,43 @@ func (s *statsCallbacks) OnStreamDeltaResponse(streamID int64, request DeltaDisc
...
+// lock; the work here (a counter increment and a log line) is cheap
+// enough to keep inside the critical section.
+func (s *statsCallbacks) maybeReportWindowDepletion(last lastSend) {
+	elapsed := core.Now().Sub(last.sentAt)
+	if elapsed >= closureWindow || last.size < sizeThreshold {
+		return
+	}
+	s.likelyWindowDepletionMetric.WithLabelValues(last.typeURL).Inc()
+	statsLogger.Info(
+		"xds stream closed within window of sending a large response; likely caller-side gRPC receive-window depletion (kumahq/kuma#16355). "+
```
lobkovilya: TBH "kumahq/kuma#16355" doesn't have any useful context, I don't think we should log it

pkg/util/xds/stats_callbacks.go
```diff
@@ -247,8 +288,43 @@ func (s *statsCallbacks) OnStreamDeltaResponse(streamID int64, request DeltaDisc
...
+		sentAt:  core.Now(),
+		typeURL: typeURL,
+		nodeID:  nodeID,
+	}
+}
+
+// maybeReportWindowDepletion emits the warning log and increments the
+// counter when the most recent send on a now-closed stream matches the
+// receive-window-depletion signature: a large response immediately
+// followed by stream closure. See kumahq/kuma#16355. Callers hold the
```
lobkovilya: Isn't "callers hold the lock" easily avoidable? You can delete the entry from `lastSendByStream`, release the lock and then call `maybeReportWindowDepletion`

pkg/util/xds/stats_callbacks.go
Automaat: Removed the issue reference from the runtime log so operators only see the actionable guidance. I kept the implementation context in code comments instead.

_Generated by Sybra harness_

pkg/util/xds/stats_callbacks.go
Automaat: Updated both stream-close paths to delete the `lastSendByStream` entry under the mutex, unlock, and only then call `maybeReportWindowDepletion`.

_Generated by Sybra harness_

lobkovilya (approved):

PR #16785, opened 2026-05-28
feat(xds): add per-proxy snapshot size histogram

Automaat:
## Motivation

Snapshot size is the load factor that determines whether the control
plane hits the gRPC C-Core 4 MiB receive-window cap (or the proposed
16 MiB cap) on the data plane. Without an explicit metric operators
discover the cap was too small only via secondary symptoms (stream
cancellations, EDS initial-fetch timeouts, warming clusters), making
diagnosis slow. See kumahq/kuma#16355.

## Implementation information

- New histogram `xds_snapshot_size_bytes{type_url, proxy_type}` on
  the existing xDS metrics registry (`pkg/xds/metrics`).
- `type_url`: the standard Envoy resource type URL (Cluster,
  ClusterLoadAssignment, Listener, RouteConfiguration, Secret,
  Runtime), taken from go-control-plane's `envoy_resource` constants.
- `proxy_type`: `dataplane`, `ingress`, `egress`, or `gateway`,
  derived from the `Proxy` struct (`ZoneIngressProxy`,
  `ZoneEgressProxy`, `Dataplane.Spec.IsBuiltinGateway()`).
- Buckets: `prometheus.ExponentialBuckets(1024, 2, 16)` — 1 KiB to
  32 MiB, covers the relevant range around the gRPC caps.
- Observation point: in `pkg/xds/server/v3/reconcile.go::Reconcile`,
  after `GenerateSnapshot` succeeds and the snapshot passes
  `Consistent()`, before `SetSnapshot` is called via the cacher. This
  measures what we are about to ship to that DP per reconcile.
- Per-resource size is computed with `proto.Size`, summed across all
  resources of each type URL in the snapshot.
- The same reconciler is reused for ingress and egress proxies, so
  `xdsMetrics` is now plumbed into `DefaultIngressReconciler` and
  `DefaultEgressReconciler` too.

> Changelog: feat(xds): add per-proxy snapshot size histogram

### DIFF
--- a/pkg/xds/metrics/metrics.go
+++ b/pkg/xds/metrics/metrics.go
@@ -13,6 +13,7 @@ type Metrics struct {
 	KubeAuthCache           *prometheus.CounterVec
 	CertExpirationTimestamp *prometheus.GaugeVec
 	SnapshotResources       *prometheus.HistogramVec
+	SnapshotSizeBytes       *prometheus.HistogramVec
 }
 
 func NewMetrics(metrics core_metrics.Metrics) (*Metrics, error) {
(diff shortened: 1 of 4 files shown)


pkg/xds/metrics/metrics.go
```diff
@@ -37,7 +38,12 @@ func NewMetrics(metrics core_metrics.Metrics) (*Metrics, error) {
 		Help:    "Distribution of resource counts per xDS snapshot by resource type.",
 		Buckets: prometheus.ExponentialBuckets(1, 2, 12),
 	}, []string{"resource_type"})
-	if err := metrics.BulkRegister(xdsGenerations, xdsGenerationsErrors, kubeAuthCache, certExpirationTimestamp, snapshotResources); err != nil {
+	snapshotSizeBytes := prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
lukidzi, line 41: Do we need this? I think it's less precise than callback metrics and might just give more noise. I would prefer just this one: https://github.com/kumahq/kuma/pull/16783/files

PR #16786, opened 2026-05-28
fix(meshhttproute): dedup duplicate routes on gateway virtual hosts

lukidzi:
## Motivation

When a MeshHTTPRoute targets a gateway and a single virtual host gets matched by more than one of its hostname patterns (for example a specific super.app1.example.com and a wildcard *.app1.example.com), the generated route configuration ends up with the same route entry repeated. The duplicates have an identical match, name, backend cluster and headers — they're literally the same Envoy route emitted twice.

  In production this shows up as virtual hosts with 4 routes where 2 are expected (each path expands into an exact + prefix match, and the rule gets attached twice). The duplicates are harmless to traffic — Envoy picks the first match — but they bloat xDS payloads, make config dumps confusing, and indicate the route generator is doing redundant work.

## Implementation information

The duplication happens in `sortRulesToHosts` (`pkg/plugins/policies/meshhttproute/plugin/v1alpha1/gateway_routes.go`): for a given virtual host we collect every rule whose hostnames match — including a rule scoped to the specific hostname and a separate rule scoped to a wildcard that also covers it. Both rules then produce identical route.Entry values.

Rather than rework the rule-collection logic (which would risk dropping rules that look identical but legitimately differ in some xDS-relevant field down the line), the fix adds a single dedup pass after the entries are fully built. `dedupRouteEntries` compares each entry to the ones already kept using `eflect.DeepEqual` and drops only exact duplicates.
  Entries that differ in match, action, weights, headers, mirror or rewrite are preserved as before

## Supporting documentation

Fix #XX


test/e2e_env/universal/envoyconfig/testdata/builtingateway/meshaccesslog/mal-with-from.gateway-proxy.golden.json
```diff
@@ -17,6 +17,60 @@
           }
         }
       ]
+    },
+    {
```
lahabana: Can we test this in unit tests and not e2e.

e2e are costly, they will be hard to backport and here the PR creates a lot of noise hard to understand in golden files

lukidzi: The problem with a unit test is that it’s hard to make it truly reliable because we need to mock a lot of stuff. I can add one, but I found it to be less precise.

bartsmykla (approved):

PR #16805, opened 2026-05-29
fix(xds): don't add empty secret

lukidzi:
## Supporting documentation

Fix #XX

### DIFF
--- a/pkg/xds/generator/secrets/generator.go
+++ b/pkg/xds/generator/secrets/generator.go
@@ -163,11 +163,6 @@ func (g Generator) Generate(
 			)
 			if ca, ok := generatedMeshCAs[mesh]; ok {
 				resources.Add(createCaSecretResource(identityName, ca))
-			} else {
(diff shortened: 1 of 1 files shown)


bartsmykla (approved):

lukidzi:
The empty CaSecret{} was a placeholder so that
▎ "Envoy syncs the config"
i.e. Envoy's SDS expects a resource for every secret name referenced in its clusters/listeners. If the CP omits it, Envoy waits for that SDS resource and won't warm/ACK the config — the proxy stalls. Sending an empty secret let Envoy treat the config as complete instead of blocking on a never-arriving secret.
Implication of removing it now
With the else gone, a missing/misconfigured other-mesh CA results in no secret resource at all for that name. This is only safe if the config no longer references that secret name when the CA fails — otherwise you reintroduce the Envoy "waiting for SDS, won't sync" stall that the empty placeholder was guarding against.


I think we should wait to 3.0.x

PR #16844, opened 2026-06-01
ci(e2e): drop MetalLB in favor of k3s built-in servicelb

lobkovilya:
## Motivation

MetalLB is an extra dependency we don't need. k3s ships klipper-lb built-in, which is enough for the few e2e specs that use `serviceType: LoadBalancer` (Graceful, Gateway API, KIC).

## Implementation information

- `mk/k3d.mk`: drop `servicelb` from `K3D_DISABLE_DEFAULT`, remove the MetalLB variables, the `k3d/cluster/metallb/setup` target, and its call from `k3d/cluster/start`.
- `mk/resources/render-metallb-pool.sh`: deleted.
- `mk/k8s.mk`: refresh stale `CLUSTER_NUMBER` comment.

`GetLBIngressIP` now returns the k3d node container's IP instead of an IP from a per-cluster MetalLB pool. Reachability inside the docker network is unchanged.

## Supporting documentation

n/a

> Changelog: skip


lobkovilya:
Metallb can create IP per Service, while klipper-lb uses Node IPs. We need IP per Service because Gateway API conformance tests run multiple Gateway resources concurrently. So we can't drop metallb.

PR #16852, opened 2026-06-02
fix(validation): enforce stricter label validation 

### LINKED ISSUE #16049: name / display-name validation doesn't always seem to be working

lobkovilya:
## Motivation

Label validation on create/update was inconsistent across the K8s
webhook, REST API server, and KDS sync. Users could silently override
CP-computed labels, and `kuma.io/origin` mismatches were tolerated —
masking resources applied to the wrong control plane.

## Implementation information

Introduces a single declarative registry of reserved labels in
`pkg/core/resources/labels/` and reroutes every entry point through
it:

- `LabelSpec` declares each label's `Owner`
  (`ControlPlane`/`User`/`System`) and its expected value.
- `Validate` returns `{Errors, Warnings}`: errors reject (format,
  user-owned `AllowedValues`, strict `kuma.io/origin` mismatch);
  warnings cover CP-owned labels the user set wrong — accepted, then
  overridden by `Compute`.
- `Compute` is authoritative: it force-sets/deletes CP-owned labels
  from the registry. The webhook defaulter skips it on `Privileged`
  contexts (KDS sync, GC, storage-version migration).

## Supporting documentation

Fix https://github.com/kumahq/kuma/issues/16049


pkg/core/resources/labels/compute.go
```diff
@@ -96,109 +106,127 @@ func WithMode(mode config_core.CpMode) Option {
...
-				setIfNotExist(mesh_proto.EnvTag, env)
+	for _, ls := range registry {
+		switch ls.Owner {
+		case OwnerUser:
+			continue
+		case OwnerSystem:
+			// Privileged callers (K8s controllers, etc.) legitimately set
+			// OwnerSystem labels like kuma.io/managed-by. Drop them only on
+			// user paths, where any value is impersonation.
+			if !o.Privileged {
```
slonka, line 150: updating an existing generated `Workload` or `MeshService` will silently drop labels like `kuma.io/managed-by` and `kuma.io/deletion-grace-period-started-at`

slonka:
> Copilot wasn't able to review this pull request because it exceeds the maximum number of files (300). Try reducing the number of changed files and requesting a review from Copilot again.

@lobkovilya you could try running it on a subset without golden files but that would require you reset some stuff and then re-commit

slonka (commented):
just found this, Ilya will adjust

pkg/api-server/resource_endpoints.go
```diff
@@ -515,7 +516,9 @@ func (r *resourceEndpoints) createResource(
 		return
 	}
 
-	resp := api_server_types.CreateOrUpdateSuccessResponse{Warnings: core_model.Deprecations(res)}
+	warnings := append([]string{}, labelWarnings...)
+	warnings = append(warnings, core_model.Deprecations(res)...)
```
lukidzi, line 520: Do we want to get rid of deprecations and just enforce them at one point?

pkg/api-server/resource_endpoints.go
```diff
@@ -665,65 +676,47 @@ func (r *resourceEndpoints) validateResourceRequest(name string, meshName string
...
+}
 
-	for _, k := range maps.SortedKeys(resource.GetMeta().GetLabels()) {
-		for _, msg := range validation.IsQualifiedName(k) {
-			err.AddViolationAt(validators.Root().Key(k), msg)
-		}
-		for _, msg := range validation.IsValidLabelValue(resource.GetMeta().GetLabels()[k]) {
-			err.AddViolationAt(validators.Root().Key(k), msg)
-		}
+func (r *resourceEndpoints) validateOriginForWrite(resource rest.Resource) validators.ValidationError {
```
lukidzi, line 694: Is it correct name it seems to be called on delete

lukidzi: Also, do we need a helper, or can it be called from Validate labels like we do for removed `validatePolicyRole`

pkg/plugins/runtime/k8s/controllers/gateway_converter.go
```diff
@@ -56,16 +56,21 @@ func (r *PodReconciler) createOrUpdateBuiltinGatewayDataplane(ctx context.Contex
 		return nil
 	}
 
+	existingLabels := mergeLabels(dataplane.GetLabels(), pod.Labels)
+	if pod.Spec.ServiceAccountName != "" {
+		existingLabels[metadata.KumaServiceAccount] = pod.Spec.ServiceAccountName
+	}
```
lukidzi, line 62: Why we changed this logic and extracted it here instead of option?

pkg/core/resources/labels/compute.go
```diff
@@ -72,135 +71,98 @@ func WithNamespace(namespace Namespace) Option {
...
 ) (map[string]string, error) {
-	labelsOpts := NewOptions(opts...)
+	o := NewOptions(opts...)
 	labels := map[string]string{}
 	if len(existingLabels) > 0 {
 		labels = maps.Clone(existingLabels)
 	}
 
-	set := func(k, v string) {
-		if _, ok := AllComputedLabels[k]; !ok {
```
lukidzi, line 115: is `AllComputedLabels` even used anywhere? Do we need to keep them?

pkg/core/resources/labels/spec.go
```diff
@@ -0,0 +1,127 @@
...
+func matchedSpec(specs []LabelSpec, ctx ValidationContext) (LabelSpec, bool) {
+	var hit LabelSpec
+	found := false
+	for _, s := range specs {
+		if !s.RequiredOn.Matches(ctx) {
+			continue
+		}
+		if found {
+			panic("resource_labels: overlapping RequiredOn for key " + s.Key)
+		}
```
lukidzi, line 122: Should this be checked on init not at the moment of validation as I can imagine this panicking during specific request

pkg/core/resources/labels/validate.go
```diff
@@ -0,0 +1,149 @@
...
+		return &Violation{
+			Key:    spec.Key,
+			Reason: fmt.Sprintf("must be one of [%s]", strings.Join(quoted(spec.AllowedValues), ", ")),
+		}, nil
+
+	case OwnerControlPlane:
+		expected, err := spec.Expected(ctx)
+		if err != nil {
+			// Compute surfaces malformed-resource errors on write.
+			return nil, nil
```
lukidzi, line 129: Which Compute? I am not sure if I can see this call

pkg/core/resources/labels/compute.go
```diff
@@ -72,135 +71,98 @@ func WithNamespace(namespace Namespace) Option {
...
-			for _, l := range dp.GetNetworking().GetListeners() {
-				switch l.Type {
-				case mesh_proto.Dataplane_Networking_Listener_ZoneIngress:
-					hasIngress = true
-				case mesh_proto.Dataplane_Networking_Listener_ZoneEgress:
-					hasEgress = true
+		switch ls.Owner {
+		case OwnerUser:
+			continue
+		case OwnerSystem:
```
lukidzi, line 149: What is the difference between System and ControlPlane owner? It seems for me the same as if control-plane updates/set them makes them CP owned

slonka: CP - writes the value and knows how to compute it (e.g. zone knows status of something)
System - it comes from some other system trusted component (e.g. global trusts zone that it properly set some label, but does not know how to compute it itself)

pkg/core/resources/labels/spec.go
```diff
@@ -1,62 +1,127 @@
...
 
 const (
-	// OwnerControlPlane: the CP computes the value. Users may set the label only
-	// if their value matches the CP-computed value.
 	OwnerControlPlane Owner = "control-plane"
-	// OwnerUser: free-form user flag. Optionally constrained to a set of values.
-	OwnerUser Owner = "user"
-	// OwnerSystem: managed by CP-internal flows only (KDS sync, lifecycle).
-	// User input is always rejected.
-	OwnerSystem Owner = "system"
```
slonka: I don't think we should remove these comments - @lukidzi asked about the meaning, maybe we should even give more info that CP owner means CP can compute the value and system means that something other in the system that we trust computed that value.

pkg/plugins/runtime/k8s/controllers/pod_converter.go
```diff
@@ -63,17 +63,24 @@ func (p *PodConverter) PodToDataplane(
 	}
 	// we need to validate if the labels have changed
 	workloadName := computeWorkloadName(pod.Labels, p.WorkloadLabels, pod.Spec.ServiceAccountName)
+	existingLabels := mergeLabels(dataplane.GetLabels(), pod.Labels)
+	if pod.Spec.ServiceAccountName != "" {
+		existingLabels[metadata.KumaServiceAccount] = pod.Spec.ServiceAccountName
+	}
+	if workloadName != "" {
+		existingLabels[metadata.KumaWorkload] = workloadName
+	}
```
slonka, line 72: This 3 lines are duplicated in `gateway_converter.go` and `PodConverter` - should we extract to a helper?

pkg/core/resources/labels/validate.go
slonka, line 129: Can't we just return the error so `Validate` is self-contained? This reliance seems brittle.

pkg/core/managers/apis/dataplane/dataplane_manager.go
```diff
@@ -109,10 +110,12 @@ func (m *dataplaneManager) Update(ctx context.Context, resource core_model.Resou
 		resource.GetSpec(),
 		opts.Labels,
 		resource.GetMeta().GetMesh(),
+		resource.GetMeta().GetName(),
 		resource_labels.WithNamespace(resource_labels.GetNamespace(resource.GetMeta(), m.systemNamespace)),
 		resource_labels.WithMode(m.mode),
 		resource_labels.WithK8s(m.isK8s),
 		resource_labels.WithZone(m.zone),
+		resource_labels.WithPreviousLabels(resource.GetMeta().GetLabels()),
```
slonka, line 118: We don't pass labels in `Create` (line 66) - is that ok?

pkg/core/resources/labels/validate.go
slonka: I think it's everything from CreateOrUpdate  https://github.com/kumahq/kuma/blob/8a132399e43ce42d95d1443120e9bb3712c526e4/pkg/api-server/resource_endpoints.go#L401

pkg/api-server/resource_endpoints.go
lobkovilya: We'll get rid and enforce them later, I don't change anything related to Deprecations in this PR

pkg/core/resources/labels/compute.go
lobkovilya: Fixed

pkg/api-server/resource_endpoints.go
lobkovilya: there was already resource_labels.ValidateOrigin so I just inlined it and removed `validateOriginForWrite`

lobkovilya:
I'm breaking it in smaller chunks, work in progress:
* https://github.com/kumahq/kuma/pull/17162

PR #16902, opened 2026-06-08
fix(k8s): renamed cni taint controller

lukidzi:
## Motivation

When executing smoke tests I've noticed a conflict on installing CP with CNI as a `cni_taint_controller` has the same name as a secret controller.

## Implementation information

Renamed `cni_taint_controller` to correct one to avoid conflict during installation.


slonka (approved):
Is there a problem on any other branch? why is 2.7 different?

lukidzi:
> Is there a problem on any other branch? why is 2.7 different?

There was a security update pr which bump a k8s, we had to fix requirement of setting name of controller and by mistake we duplicated the name for 2 controllers. I've checked other branches and there is no issue

Issue #16923, opened 2026-06-11
Missing KRI OpenAPI fields in DataplaneItem and MeshGatewayItem


johncowen:
### Kuma Version

2.13.x

### Describe the bug

I found that `DataplaneItem` and `MeshGatewayItem` in the openAPI spec are missing the following fields:

- `kri`
- `creationTime`
- `modificationTime`

Possibly others but those are the ones I noticed.

nikhiltelkar2005-glitch:
Hi, I'd like to work on this issue. Could you please assign it to me?

PR #16925, opened 2026-06-11
fix(kri): add KRI to overviews and OpenAPI items

Automaat:
## Motivation

Two related KRI gaps surfaced by users:

1. Computed `*Overview` resources (`DataplaneOverview`, `ZoneIngressOverview`, `ZoneEgressOverview`, `ZoneOverview`) did not serialize the `kri` field, even though an overview is just a merge of a resource and its insight and therefore shares the underlying resource's identity.
2. The OpenAPI `*Item` schemas for core resources (`DataplaneItem`, `MeshGatewayItem`, ...) were missing `kri`, `creationTime` and `modificationTime` (kumahq/kuma#16923).

> Changelog: fix(kri): add KRI to overviews and OpenAPI items

## Implementation information

**Overviews (runtime serialization).** `kri.FromResourceMetaE` computed the KRI from the resource's `type`, which for an overview is e.g. `DataplaneOverview` — a type that is unregistered and has no `ShortName`, so the lookup bailed out and no `kri` was emitted. The fix resolves an overview type to its base type (strip the `Overview` suffix, look up the base descriptor) so the emitted KRI points at the actual resource, e.g. `DataplaneOverview` -> `kri_dp_...`. Overviews intentionally keep no `ShortName` of their own (a duplicate short name would make `FromString` ambiguous).

**OpenAPI items (generated spec).** Core resource `*Item` schemas are built programmatically by `resource-gen` (`tools/resource-gen/pkg/generator/main.go`), not from the policy template, and only set `type`/`name`/`mesh`/`labels` plus reflected proto fields. Added `kri`, `creationTime` and `modificationTime` to that schema. `kri` is only advertised for resources that have a `ShortName`, matching runtime behaviour (Secret, which has none, gets the timestamps but no `kri`). Regenerated the per-resource `rest.yaml` files and the bundled `docs/generated/openapi.yaml`.

Verified both gateway cases do emit a KRI (they are not skipped): gateway dataplanes serialize `kri_dp_...` and `MeshGateway` serializes `kri_mgw_...`.

## Supporting documentation

- kumahq/kuma#16923

### DIFF
--- a/api/mesh/v1alpha1/dataplane/rest.yaml
+++ b/api/mesh/v1alpha1/dataplane/rest.yaml
@@ -47,6 +47,15 @@ components:
   schemas:
     DataplaneItem:
       properties:
+        creationTime:
+          description: Time at which the resource was created
+          format: date-time
+          readOnly: true
+          type: string
+        kri:
+          description: Kuma Resource Identifier (KRI) of the given resource
+          readOnly: true
+          type: string
         labels:
           additionalProperties:
             type: string
@@ -72,6 +81,11 @@ components:
               description: Type of the backend (Kuma ships with 'prometheus')
               type: string
           type: object
+        modificationTime:
+          description: Time at which the resource was updated
+          format: date-time
+          readOnly: true
+          type: string
         name:
           type: string
         networking:
--- a/app/kumactl/cmd/inspect/testdata/inspect-dataplanes.golden.json
+++ b/app/kumactl/cmd/inspect/testdata/inspect-dataplanes.golden.json
@@ -7,6 +7,7 @@
       "name": "experiment",
       "creationTime": "2018-07-17T16:05:36.995Z",
       "modificationTime": "2019-07-17T18:08:41Z",
+      "kri": "kri_dp_default___experiment_",
       "dataplane": {
         "networking": {
           "address": "127.0.0.1",
@@ -103,6 +104,7 @@
       "name": "degraded-dp",
       "creationTime": "2018-07-17T16:05:36.995Z",
       "modificationTime": "2019-07-17T18:08:41Z",
+      "kri": "kri_dp_default___degraded-dp_",
       "dataplane": {
         "networking": {
           "address": "127.0.0.1",
@@ -198,6 +200,7 @@
       "name": "offline-dp",
       "creationTime": "2018-07-17T16:05:36.995Z",
       "modificationTime": "2019-07-17T18:08:41Z",
+      "kri": "kri_dp_default___offline-dp_",
       "dataplane": {
         "networking": {
           "address": "127.0.0.1",
@@ -294,6 +297,7 @@
       "name": "example",
       "creationTime": "2018-07-17T16:05:36.995Z",
       "modificationTime": "2019-07-17T18:08:41Z",
+      "kri": "kri_dp_default___example_",
       "dataplane": {
         "networking": {
           "address": "127.0.0.1",
--- a/pkg/core/kri/kri.go
+++ b/pkg/core/kri/kri.go
@@ -70,9 +70,16 @@ func FromResourceMetaE[T ~string](rm core_model.ResourceMeta, resourceType T) (I
 
 	rType := core_model.ResourceType(resourceType)
 
-	if desc, err := registry.Global().DescriptorFor(rType); err != nil {
+	// an overview shares its base resource's identity: DataplaneOverview -> Dataplane
+	if base, isOverview := strings.CutSuffix(string(rType), "Overview"); isOverview {
+		rType = core_model.ResourceType(base)
+	}
+
+	desc, err := registry.Global().DescriptorFor(rType)
+	if err != nil {
 		return Identifier{}, err
-	} else if desc.ShortName == "" {
+	}
+	if desc.ShortName == "" { // types without a ShortName aren't addressable by KRI
 		return Identifier{}, nil
 	}
 
--- a/api/mesh/v1alpha1/mesh/rest.yaml
+++ b/api/mesh/v1alpha1/mesh/rest.yaml
@@ -95,6 +95,15 @@ components:
                   type: array
               type: object
           type: object
+        creationTime:
+          description: Time at which the resource was created
+          format: date-time
+          readOnly: true
+          type: string
+        kri:
+          description: Kuma Resource Identifier (KRI) of the given resource
+          readOnly: true
+          type: string
         labels:
           additionalProperties:
             type: string
@@ -170,6 +179,11 @@ components:
               description: Name of the enabled backend
               type: string
           type: object
+        modificationTime:
+          description: Time at which the resource was updated
+          format: date-time
+          readOnly: true
+          type: string
         mtls:
           description: |-
             mTLS settings.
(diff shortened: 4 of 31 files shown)


api/mesh/v1alpha1/dataplane/rest.yaml
```diff
@@ -72,6 +81,11 @@ components:
               description: Type of the backend (Kuma ships with 'prometheus')
               type: string
           type: object
+        modificationTime:
+          description: Time at which the resource was updated
+          format: date-time
+          readOnly: true
+          type: string
```
johncowen, line 88: nit: can we put creationTime and modificationTime together

johncowen: P.S. If this nit means a change that means having to wait for CI for a long period of time again, then don't worry.

johncowen: Oh wait, totally ignore me, these seem to be in alphabetical order, lets stick with that pattern. I'll resolve this comment thread.

johncowen: Well I would resolve this thread if GH would let me 😭 , maybe I need to wait a bit, feel free to resolve if someone gets here first 😆

app/kumactl/cmd/inspect/testdata/inspect-dataplanes.golden.json
```diff
@@ -7,6 +7,7 @@
       "name": "experiment",
       "creationTime": "2018-07-17T16:05:36.995Z",
       "modificationTime": "2019-07-17T18:08:41Z",
+      "kri": "kri_dp_default___experiment_",
```
johncowen, line 10: Q: shouldn't these golden files have labels? Maybe I missed them?

pkg/core/kri/kri.go
```diff
@@ -70,10 +70,22 @@ func FromResourceMetaE[T ~string](rm core_model.ResourceMeta, resourceType T) (I
 
 	rType := core_model.ResourceType(resourceType)
 
-	if desc, err := registry.Global().DescriptorFor(rType); err != nil {
-		return Identifier{}, err
-	} else if desc.ShortName == "" {
-		return Identifier{}, nil
+	desc, err := registry.Global().DescriptorFor(rType)
+	if err != nil || desc.ShortName == "" {
```
johncowen: Q: When does an empty ShortName not produce an error?

johncowen: Oh! I guess "read the comment"? 😆 Is it when `rType`s returned descriptor is an `*Overview`?

johncowen: I find it strange that `DescriptorFor` can return something with an empty `ShortName`

I wondered what you thought of this approach:

```golang
func FromResourceMetaE[T ~string](rm core_model.ResourceMeta, resourceOrOverviewType T) (Identifier, error) {
  if rm == nil {
    return Identifier{}, nil
  }

  // normalize the resourceType first
  overviewType, isOverview := strings.CutSuffix(string(resourceOrOverviewType), "Overview")
  if !isOverview { // <== I assumed this is false if if doesn't end with `Overview`, I didn't get to look up the doc
    resourceType = resourceOrOverviewType
  } else {
    resourceType = overviewType
  }

	rType := core_model.ResourceType(resourceType)

	_, err := registry.Global().DescriptorFor(rType)
	if err != nil { // I guess we'd still need to check for an empty shortName here if that doesn't cause an error yet?
		return Identifier{}, err
	}

	return Identifier{
		ResourceType: rType,
		Mesh:         rm.GetMesh(),
		Zone:         rm.GetLabels()[mesh_proto.ZoneTag],
		Namespace:    rm.GetLabels()[mesh_proto.KubeNamespaceTag],
		Name:         core_model.GetDisplayName(rm),
	}, nil
}
```

I don't know what other places in the codebase allow a `.DescriptorFor(rType)` to be return (not error) with an empty `ShortName`. i.e. I have zero context on past decision etc here. Also please excuse my golang I've not messed with golang in quite a while and I just pseudo-coded the above in a blank text file to help explain my viewpoint (and try the code out myself)

Anyway, a little bit of a nit for this PR and not a blocker at all, but I'd like to understand why `.DescriptorFor(rType)` can return an empty `ShortName` which feels wrong to me.

Automaat: Good call — adopted your normalize-first structure (pushed). It now strips the `Overview` suffix up front (`DataplaneOverview` -> `Dataplane`), then does a single `DescriptorFor` + `ShortName` check.

I kept the `ShortName == ""` guard though: plenty of registered types intentionally have no ShortName (insights, secrets, etc.) and must NOT get a KRI. Without the guard, `Identifier.String()` would emit a malformed `kri__...` (empty short name) for them. So `DescriptorFor` returning a descriptor with an empty ShortName is expected — it just means "this type is not addressable by KRI".

app/kumactl/cmd/inspect/testdata/inspect-dataplanes.golden.json
Automaat: These come from the test fixtures (`test_model.ResourceMeta` in `inspect_dataplanes_test.go`), which only set mesh/name/timestamps and no labels — so the goldens never had them. This PR only added the `kri` line; nothing label-related was removed. Happy to add labels to the fixtures separately, but it is out of scope here.

api/mesh/v1alpha1/dataplane/rest.yaml
Automaat: Yep, alphabetical ordering is what the generator emits (the schema is built programmatically and serialized as a sorted map), so leaving as-is. Thanks!

app/kumactl/cmd/inspect/testdata/inspect-dataplanes.golden.json
johncowen: Gotcha, makes sense 👍

pkg/core/kri/kri.go
```diff
@@ -70,9 +70,20 @@ func FromResourceMetaE[T ~string](rm core_model.ResourceMeta, resourceType T) (I
 
 	rType := core_model.ResourceType(resourceType)
 
-	if desc, err := registry.Global().DescriptorFor(rType); err != nil {
+	// An overview is a merge of a resource and its insight, so it shares the
+	// underlying resource's identity. Overview types are not registered and have
+	// no ShortName of their own, so normalize to the base type first (e.g.
+	// DataplaneOverview -> Dataplane) before building the KRI.
+	if base, isOverview := strings.CutSuffix(string(rType), "Overview"); isOverview {
```
lahabana, line 74: There's a way cleaner way to do this inside the description. That's AI slop :)

Automaat: Trimmed it down to a one-liner

pkg/core/kri/kri.go
johncowen: Oh I just noticed the emdashes 🙈 😅 Was this comment LLM generated?

Automaat: yes, it answered before I stopped him :P

johncowen: What's your opinion on what I said?

lukidzi (approved):

pkg/core/kri/kri.go
johncowen: (btw this conversation shouldn't block merge and I'm happy for it to continue after merge if thats how it turns out)

PR #16941, opened 2026-06-11
fix(sni): don't set kri-derived sni when no mesh-scoped proxy for MMZS

lukidzi:
## Motivation

 MeshMultiZoneService is a global resource and always has `zone == "" `in its KRI identifier. The condition that gates new KRI SNI usage was:
```
  useKRISni := zone == "" || isLocalMeshService || meshCtx.ZonesWithMeshScopedProxy[zone]
```
The `zone == ""` branch fires unconditionally for every MZMS, bypassing the ZonesWithMeshScopedProxy check entirely. That map is what tells us whether a remote zone has a new-style zone proxy or only a legacy ZoneIngress, so skipping it was wrong.

## Implementation information

For MZMS, inspect the actual remote endpoint zones instead of relying on the resource's own zone field. Only use new KRI SNI if every remote endpoint zone appears in `ZonesWithMeshScopedProxy`. If any remote endpoint zone uses a legacy ZoneIngress, the dataplane falls back to the old hash-based SNI, which the legacy ingress can actually match.


pkg/plugins/policies/core/xds/meshroute/clusters.go
```diff
@@ -336,3 +345,22 @@ func isMeshExternalService(endpoints []core_xds.Endpoint) bool {
 	}
 	return false
 }
+
+// allEndpointZonesHaveNewProxy returns true when the endpoint list is non-empty
+// and every remote endpoint (Locality.Zone != "") is served by a zone that has
+// a new-style zone proxy (MeshZoneAddress). Endpoints without locality (local
+// zone, sidecar-to-sidecar) are skipped since they never traverse a zone proxy.
+func allEndpointZonesHaveNewProxy(endpoints []core_xds.Endpoint, zonesWithProxy map[string]bool) bool {
```
lobkovilya, line 353: But does this mean that if an MMZS consists of 3 MeshServices in different zones, and only 2 of them have the new zone proxy, the outbound cluster will keep using the old SNI, so only 1/3 of requests will succeed?

lobkovilya (commented):
We might need to use `transport_socket_matches` for the MMZS cluster and choose the SNI based on whether the zone has a MeshZoneAddress.

PR #16947, opened 2026-06-12
fix(mtls): drop legacy identity secret when MeshIdentity is assigned

lukidzi:
## Motivation

When a mesh has both an mTLS backend and a MeshIdentity assigned, restarting a proxy leaves its secrets stuck in `dynamic_warming_secrets` and mTLS never comes up. The snapshot ends up with three secrets: `system_trust_bundle` and `kri_mid_…identity_`, which the listeners and clusters actually reference, plus a leftover `identity_cert:secret:default` that nothing points to. `go-control-plane's` ADS consistency check requires every secret in the snapshot to be one Envoy asked for, so the unreferenced one makes it refuse the whole SDS response and the secrets Envoy needs never warm up.

## Implementation information

Return early from the secrets generator when the proxy has a workload identity. In that case MeshIdentity and MeshTrust already deliver the cert and trust bundle under KRI names, so the legacy per-mesh secrets are pure dead weight and the unreferenced identity cert is exactly what breaks the ADS check.

The problems is when using SOTW not Delta


lobkovilya (dismissed):

lobkovilya:
@lukidzi it probably worth adding in description that it applies only to sotw. So I'm not really sure we need to backport it.

lobkovilya (approved):

PR #16970, opened 2026-06-15
chore(deps): add mise.lock file

johncowen:
## Motivation

I noticed there are several dependencies that aren't really pinned in the mise.toml file, for example:

https://github.com/kumahq/kuma/blob/9a1bb8ce6d23b7589fde7839c42971132d5ae2ae/mise.toml#L21

https://github.com/kumahq/kuma/blob/9a1bb8ce6d23b7589fde7839c42971132d5ae2ae/mise.toml#L19

## Implementation information

Follows guidance from https://mise.jdx.dev/dev-tools/mise-lock.html using `mise lock` to create a `mise.lock` file.

Note: I don't often work in this repository, so I'd be very grateful if maintainers could double check that this fits in with current day-to-day practices and nothing needs adding to docs or anything like that. It might even be worth a normal contributor re-creating the lockfile to ensure I've set things up properly without the lock. Importantly, I'm guessing methods of upgrading etc may have changed now compared to what was done previously.

## Supporting documentation

https://mise.jdx.dev/dev-tools/mise-lock.html

Note: the above documentation discusses `minimum-release-age` (default of 24 hours) which we might want to raise in a either here or separate PR. Happy to hear opinions.


Automaat (approved):

mise.lock
Automaat, line 1: I don't see that much value in this; maybe I am wrong? It gives a little bit of hardening of the supply chain, but we bump stuff with Renovate, which would need to handle mise.lock then. All renovated PRs to mise will now fail and will need to be manually fixed to regenrate mise.lock

Automaat (changes requested):

mise.lock
johncowen: Agree we should think about this a little more 👍 

Honestly as it stands now I'm still very pro on this, but appreciate I've not specifically used mise locks in the past, plus this repro isn't a place a work often so I don't want to impose my own opinions too strongly here.

As I pointed out in Slack we should also think about what the impact of folks using completely different version of mise and therefore generating slightly different lockfiles (ideally everyone would automatically use the same mise version)

---

On the renovate subject, from what I can see renovate seems to support a lockfile https://docs.renovatebot.com/modules/manager/mise/#lock-file-maintenance

---

I think I might bring this up in a meeting with the wider team also for a chat 👍

johncowen: P.S.

I noticed (if I'm understanding mise correctly) that currently the `mise.toml` is using range versions for certain deps, so if we did choose to not use a lockfile - then we should enforce some sort of exact locking of versions in the mise.toml file instead (as I said I'm pro lock file, but worth saying `mise lock` exists, IMHO: shouldn't we just use that instead?)

PR #16977, opened 2026-06-21
fix(inbound): check all app containers for readiness when no container declares port

### LINKED ISSUE #16903: Inbound health ignores app readiness when no container declares the port
## What happens

When the Service's `targetPort` is defined and no container declares that port via `containerPorts`, `FindPort` returns nil:

https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/util/util.go#L99-L102

`podReady` then guards the application-container check with `if container != nil` and silently skips it:

https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/controllers/inbound_converter.go#L22-L43

Result: the inbound is published as `ready=true` as soon as `kuma-sidecar` is ready, regardless of the application container's readiness probe. Endpoints enter EDS before the app is actually serving and clients see `503s` during rollouts.

## Proposed fix

Possible options:

1. When no container declares the inbound port, "AND" the readiness status of all non-sidecar, non-init containers in the pod and use that.

2. Fail closed (ready=false) until a port is declared

3. Keep the current behaviour but issue warn level log

Additionally we should document this (this is not mentioned in https://kuma.io/docs/dev/policies/service-health-probes/ )

AruneshDwivedi:
When a Service's targetPort is not declared by any container's containerPorts, FindPort returns container==nil. The podReady function then skipped the application container readiness check entirely, marking the inbound as ready=true as soon as the kuma-sidecar was ready.

This caused endpoints to enter EDS before the application was actually serving, resulting in 503s during rollouts.

Fix: when container is nil, iterate all non-sidecar containers and check their readiness status, similar to what inboundForServiceless already does.

Fixes #16903

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
+++ b/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
@@ -31,12 +31,8 @@ func podReady(pod *kube_core.Pod, container *kube_core.Container) bool {
 		}
 	} else {
 		// No container declares the service's targetPort — check all non-sidecar app containers
-		for _, c := range pod.Spec.Containers {
-			if c.Name != util_k8s.KumaSidecarContainerName {
-				if cs := util_k8s.FindContainerStatus(c.Name, pod.Status.ContainerStatuses); cs != nil && !cs.Ready {
-					return false
-				}
-			}
+		if !appContainersReady(pod) {
+			return false
 		}
 	}
 	if cs := util_k8s.FindContainerOrInitContainerStatus(
@@ -48,6 +44,19 @@ func podReady(pod *kube_core.Pod, container *kube_core.Container) bool {
 	}
 	if pod.DeletionTimestamp != nil {
 		return false
+	}
+	return true
+}
+
+// appContainersReady checks if all non-sidecar app containers in the pod are ready.
+// Returns true if all are ready, false if any is not ready.
+func appContainersReady(pod *kube_core.Pod) bool {
+	for _, c := range pod.Spec.Containers {
+		if c.Name != util_k8s.KumaSidecarContainerName {
+			if cs := util_k8s.FindContainerStatus(c.Name, pod.Status.ContainerStatuses); cs != nil && !cs.Ready {
+				return false
+			}
+		}
 	}
 	return true
 }
@@ -131,13 +140,9 @@ func (ic *InboundConverter) inboundForServiceless(zone string, pod *kube_core.Po
 		Ready: true,
 	}
 
-	for _, container := range pod.Spec.Containers {
-		if container.Name != util_k8s.KumaSidecarContainerName {
-			if cs := util_k8s.FindContainerStatus(container.Name, pod.Status.ContainerStatuses); cs != nil && !cs.Ready {
-				state = mesh_proto.Dataplane_Networking_Inbound_NotReady
-				health.Ready = false
-			}
-		}
+	if !appContainersReady(pod) {
+		state = mesh_proto.Dataplane_Networking_Inbound_NotReady
+		health.Ready = false
 	}
 
 	// also we're checking whether kuma-sidecar container is ready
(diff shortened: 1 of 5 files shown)


pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -28,6 +28,18 @@ func podReady(pod *kube_core.Pod, container *kube_core.Container) bool {
 		if cs := util_k8s.FindContainerStatus(container.Name, pod.Status.ContainerStatuses); cs != nil && !cs.Ready {
 			return false
 		}
+	} else {
```
Automaat, line 32: **suggestion:** When no container declares the inbound port, the silent path is undiagnosable. A `converterLog.Info` in the `else` branch surfaces the likely misconfiguration. This was option 3 in #16903, so it's optional, not required.

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -28,6 +28,18 @@ func podReady(pod *kube_core.Pod, container *kube_core.Container) bool {
...
+		// When no container declares the inbound port, check all
+		// non-sidecar, non-init containers for readiness so that
+		// the inbound is not marked ready before the app starts.
+		for _, c := range pod.Spec.Containers {
```
Automaat: **suggestion:** This loop duplicates the readiness iteration in `inboundForServiceless` (lines 136-143): same sidecar skip, same AND-readiness, different control flow. Two copies can drift if one is later fixed. A shared `appContainersReady(pod) bool` would name the intent. Weigh against the repo's minimal-change preference; author's call.

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -28,6 +28,18 @@ func podReady(pod *kube_core.Pod, container *kube_core.Container) bool {
 		if cs := util_k8s.FindContainerStatus(container.Name, pod.Status.ContainerStatuses); cs != nil && !cs.Ready {
 			return false
 		}
+	} else {
+		// When no container declares the inbound port, check all
```
Automaat: **suggestion:** The comment says "non-sidecar, non-init containers," but the loop iterates `pod.Spec.Containers`, which never holds init containers, and skips only `kuma-sidecar`. Native sidecars (init containers with `restartPolicy: Always`) live in `InitContainers` and go unchecked. Trim to "all non-sidecar app containers" so the comment doesn't claim an exclusion the code can't make.

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
Automaat: **issue:** No test covers the fixed path, so reverting the `else` block passes CI and #16903 regresses silently. Add a `45.pod.yaml` (Service numeric `targetPort` absent from every `containerPort`, one app container `ready: false`, sidecar ready) and `45.dataplane.yaml` (expect inbound `state: NotReady`), plus an Entry in `pod_converter_test.go`.

AruneshDwivedi:
Updated the comment to say "non-sidecar app containers" — pod.Spec.Containers only iterates app containers, not init containers.

For the test case, I dont have Go toolchain available locally to add the golden file test. If someone can scaffold the test pod YAML (45.pod.yaml) with a numeric targetPort not in any containerPorts and one app container ready=false, I can add the matching dataplane expectation.

AruneshDwivedi:
Updated the golden file (11.dataplane.yaml) — the 8443 inbound now correctly shows ready=false since no container declares that port and container-2 is not ready. This was a legitimate test gap from the podReady change.

AruneshDwivedi:
Added golden file test (45.pod.yaml) covering the exact scenario: service with numeric targetPort 8443 not declared by any container, one app container not ready, sidecar ready. Expected dataplane shows inbound state: NotReady. Also trimmed the comment per your feedback — pod.Spec.Containers never holds init containers, so "all non-sidecar containers" is accurate.

PR #16979, opened 2026-06-21
refactor(api-server): extract computeLabels

Automaat:
## Motivation

`createResource` and `updateResource` in `pkg/api-server/resource_endpoints.go` each built a resource's label set with the same eight-line `resource_labels.Compute(...)` call — three sites in total, differing only in which `descriptor` / `spec` / `meta` they passed. The shared option tail (`WithNamespace`/`WithMode`/`WithK8s`/`WithZone`) had to be kept identical by hand; adding a label dimension meant editing three places. Rule of Three is met, so it's now worth folding.

> Changelog: skip

## Implementation information

Extract `r.computeLabels(descriptor, spec, meta, meshName)`, which holds the shared options; the three call sites pass only what varies. `labels`/`namespace` both derive from the passed `meta`, so a single `meta` argument covers both.

Structural change only — behavior is unchanged:
- the distinct per-call-site error messages are preserved at the call sites (not hidden in the helper);
- the "compute current labels BEFORE `SetSpec`" ordering in `updateResource` is untouched.

Covered by the existing label tests (`should compute labels on update of the resource`, `should set origin label automatically ...`); full `pkg/api-server` suite passes, `golangci-lint` reports 0 issues.

## Supporting documentation

First structural step from a refactoring review of this file (the top churn × size hotspot in `pkg/`), following two behavioral bug fixes to the same file shipped separately. Endorsed unanimously by the review (Rule of Three) with no dissent.

### DIFF
--- a/pkg/api-server/resource_endpoints.go
+++ b/pkg/api-server/resource_endpoints.go
@@ -453,6 +453,27 @@ func (r *resourceEndpoints) clearMeshTrustOrigin(resRest rest.Resource, meshName
 	}
 }
 
+// computeLabels derives the full label set for a resource from its descriptor,
+// spec and meta, applying the control-plane mode, zone, k8s and namespace context
+// shared by create and update.
+func (r *resourceEndpoints) computeLabels(
+	descriptor core_model.ResourceTypeDescriptor,
+	spec core_model.ResourceSpec,
+	meta core_model.ResourceMeta,
+	meshName string,
+) (map[string]string, error) {
+	return resource_labels.Compute(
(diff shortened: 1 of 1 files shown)


slonka (commented):
I don't see a ton of value in this - right? We're wrapping a call to a function with a call to a function that wraps that function.

I guess we have a couple less parameters on that call.

slonka (approved):
Martin explained that this is a first step in a bigger refactoring. This is not harmful. Let's see the next steps.

PR #16996, opened 2026-06-22
refactor(api-server): extract inspect handler

Automaat:
## Motivation

Second step of decomposing the `resourceEndpoints` god-struct (top churn × size hotspot in `pkg/`) into per-responsibility handlers. This extracts the inspect surface — by far the highest-churn part of the file — into its own type and file, so inspect changes stop touching the CRUD code.

> Changelog: skip

## Implementation information

Move the resource inspect endpoints — matching-dataplanes, proxy `_config`, `_rules`, and the `_policies` handlers, plus the `matchedPoliciesTo*` mappers — out of `resource_endpoints.go` into a dedicated `resourceInspectHandler` in `resource_inspect_endpoints.go`.

- Both `resourceEndpoints` and `resourceInspectHandler` embed the shared `resourceEndpointsContext`, so the moved method bodies are unchanged except for the receiver.
- `resourceEndpoints` keeps route registration and delegates the inspect routes to `r.inspect.*`; the inspect-only dependencies (`meshContextBuilder`, `xdsHooks`, `knownInternalAddresses`) move onto the handler.
- `resource_endpoints.go` drops from ~1500 to ~780 lines; `resource_inspect_endpoints.go` is ~750.

Structural only — behavior unchanged. Full `pkg/api-server` suite passes; `golangci-lint` reports 0 issues. go-restful stays at the handler edge (no DTO/use-case layer).

**Stacking:** this builds on the preceding `resourceEndpointsContext` extraction, so until that lands this diff also contains it; it resolves to just the inspect split once the context PR merges. Review the context PR first.

## Supporting documentation

Step 2 of the SRP decomposition from a refactoring review of this file. The final step extracts `resourceCrudHandler`, leaving `resourceEndpoints` as the route registrar holding the context plus the two handlers.

### DIFF
--- a/pkg/api-server/resource_inspect_endpoints.go
+++ b/pkg/api-server/resource_inspect_endpoints.go
@@ -1,0 +1,751 @@
+package api_server
+
+import (
+	"errors"
+	"fmt"
+	"slices"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/emicklei/go-restful/v3"
+
+	common_api "github.com/kumahq/kuma/v3/api/common/v1alpha1"
+	mesh_proto "github.com/kumahq/kuma/v3/api/mesh/v1alpha1"
+	api_types "github.com/kumahq/kuma/v3/api/openapi/types"
+	api_common "github.com/kumahq/kuma/v3/api/openapi/types/common"
+	oapi_helpers "github.com/kumahq/kuma/v3/pkg/api-server/oapi-helpers"
+	"github.com/kumahq/kuma/v3/pkg/core/kri"
+	core_plugins "github.com/kumahq/kuma/v3/pkg/core/plugins"
+	"github.com/kumahq/kuma/v3/pkg/core/policy"
+	"github.com/kumahq/kuma/v3/pkg/core/resources/access"
+	core_mesh "github.com/kumahq/kuma/v3/pkg/core/resources/apis/mesh"
+	"github.com/kumahq/kuma/v3/pkg/core/resources/manager"
+	core_model "github.com/kumahq/kuma/v3/pkg/core/resources/model"
+	"github.com/kumahq/kuma/v3/pkg/core/resources/registry"
+	"github.com/kumahq/kuma/v3/pkg/core/resources/store"
+	rest_errors "github.com/kumahq/kuma/v3/pkg/core/rest/errors"
+	"github.com/kumahq/kuma/v3/pkg/core/user"
+	core_xds "github.com/kumahq/kuma/v3/pkg/core/xds"
+	"github.com/kumahq/kuma/v3/pkg/core/xds/inspect"
+	"github.com/kumahq/kuma/v3/pkg/plugins/policies/core/matchers"
+	core_rules "github.com/kumahq/kuma/v3/pkg/plugins/policies/core/rules"
+	"github.com/kumahq/kuma/v3/pkg/plugins/policies/core/rules/common"
+	"github.com/kumahq/kuma/v3/pkg/plugins/policies/core/rules/outbound"
+	meshhttproute_api "github.com/kumahq/kuma/v3/pkg/plugins/policies/meshhttproute/api/v1alpha1"
+	meshtcproute_api "github.com/kumahq/kuma/v3/pkg/plugins/policies/meshtcproute/api/v1alpha1"
+	"github.com/kumahq/kuma/v3/pkg/util/pointer"
+	util_slices "github.com/kumahq/kuma/v3/pkg/util/slices"
+	xds_context "github.com/kumahq/kuma/v3/pkg/xds/context"
+	xds_hooks "github.com/kumahq/kuma/v3/pkg/xds/hooks"
+)
+
+// resourceInspectHandler serves the resource inspect endpoints (rules, policies,
+// proxy config, matching dataplanes).
+type resourceInspectHandler struct {
+	resourceEndpointsContext
+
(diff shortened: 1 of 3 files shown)


pkg/api-server/resource_inspect_endpoints.go
```diff
@@ -0,0 +1,751 @@
...
+			res, err := policyPlugin.Plugin.MatchedPolicies(dataplane, baseMeshContext.Resources())
+			if err != nil {
+				rest_errors.HandleError(request.Request.Context(), response, err, fmt.Sprintf("could not apply policy plugin %s", policyPlugin.Name))
+			}
+			if res.Type == "" {
+				rest_errors.HandleError(request.Request.Context(), response, err, fmt.Sprintf("matched policy didn't set type for policy plugin %s", policyPlugin.Name))
+			}
+
+			matchedPolicies = append(matchedPolicies, res)
+		}
```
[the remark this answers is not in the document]
Automaat: Fixed in eddffe5749: added the missing `return` after `HandleError` so the loop stops and no success response is written after an error. For the empty-`Type` case, it now passes a real `fmt.Errorf(...)` instead of the nil `err`, so the 500 carries a useful cause.

pkg/api-server/resource_inspect_endpoints.go
```diff
@@ -0,0 +1,751 @@
...
+						}
+					}
+				}
+			}
+			if err != nil {
+				rest_errors.HandleError(request.Request.Context(), response, err, fmt.Sprintf("could not apply policy plugin %s", policyPlugin.Name))
+			}
+			if res.Type == "" {
+				rest_errors.HandleError(request.Request.Context(), response, err, fmt.Sprintf("matched policy didn't set type for policy plugin %s", policyPlugin.Name))
+			}
```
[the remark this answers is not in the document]
Automaat: Fixed in eddffe5749: added the missing `return` after `HandleError` so the loop stops and no success response is written after an error. For the empty-`Type` case, it now passes a real `fmt.Errorf(...)` instead of the nil `err`, so the 500 carries a useful cause.

pkg/api-server/server.go
```diff
@@ -295,23 +295,29 @@ func addResourcesEndpoints(
...
 			disableOriginLabelValidation: cfg.Multizone.Zone.DisableOriginLabelValidation,
-			xdsHooks:                     xdsHooks,
 			systemNamespace:              cfg.Store.Kubernetes.SystemNamespace,
 			isK8s:                        cfg.Environment == config_core.KubernetesEnvironment,
-			knownInternalAddresses:       cfg.IPAM.KnownInternalCIDRs,
-		}
-		if cfg.Mode == config_core.Zone && cfg.Multizone != nil && cfg.Multizone.Zone != nil {
-			endpoints.zoneName = cfg.Multizone.Zone.Name
+			inspect: &resourceInspectHandler{
+				resourceEndpointsContext: endpointsCtx,
```
slonka, line 316: `endpointsCtx` is copied by value into both `resourceEndpoints` and here. Was the by-value embed an intentional (avoiding nil checks / easier zero-value), or incidental?

pkg/api-server/resource_inspect_endpoints.go
```diff
@@ -0,0 +1,755 @@
...
+// proxy config, matching dataplanes).
+type resourceInspectHandler struct {
+	resourceEndpointsContext
+
+	meshContextBuilder     xds_context.MeshContextBuilder
+	xdsHooks               []xds_hooks.ResourceSetHook
+	knownInternalAddresses []string
+}
+
+func (r *resourceInspectHandler) matchingDataplanesForPolicy() restful.RouteFunction {
```
slonka, line 53: verified that body of all functions is the same

pkg/api-server/resource_inspect_endpoints.go
```diff
@@ -0,0 +1,755 @@
...
+		resources := xds_context.Resources{
+			CrossMeshResources: map[core_xds.MeshName]xds_context.ResourceMap{},
+			MeshLocalResources: baseMeshContext.ResourceMap,
+		}
+		matchesByHash := map[common_api.MatchesHash][]meshhttproute_api.Match{}
+		// Get all the matching policies
+		allPlugins := core_plugins.Plugins().PolicyPlugins()
+		rules := []api_common.InspectRule{}
+		for _, policyPlugin := range allPlugins {
+			res, err := policyPlugin.Plugin.MatchedPolicies(dp, resources)
```
slonka, line 581: nit: (preexisting) why do we not check `err` here?

slonka (dismissed):

pkg/api-server/resource_inspect_endpoints.go
Automaat: Good catch — reordered so `err` (and the empty-type case) is checked right after `MatchedPolicies`, before `res` is used. On error `res` is zero so the `matchesByHash` block was a no-op, so this is behaviour-equivalent; it just reads correctly and matches `getPoliciesConf`. Fixed in 515ee5a3ec.

pkg/api-server/server.go
Automaat: Intentional. `resourceEndpointsContext` is set once at construction and never mutated, so the by-value copy into `resourceEndpoints` and `resourceInspectHandler` is safe and avoids the nil-pointer footgun a shared `*context` would add — there's no shared mutable state to keep in sync. If we ever need post-construction mutation we can switch to a pointer then.

slonka (approved):

PR #17008, opened 2026-06-23
perf(xds): replace proto.Equal with xxhash

### LINKED ISSUE #16187: perf(xds): equalSnapshots uses proto.Equal deep-compare every reconcile
## Summary

`equalSnapshots()` compares previous and new xDS snapshots using `proto.Equal`, which walks every field recursively. With N DPs × M resources per DP this is a significant CPU and GC-pressure driver per reconcile tick, and most reconciles produce identical snapshots (no change).

## Code

- `pkg/xds/server/v3/reconcile.go:145-155` — snapshot equality check via `proto.Equal`

## Proposed fix

Hash-based versioning: compute a stable hash (CRC32 or FNV of marshaled bytes) at snapshot construction, compare hashes on subsequent reconciles. `proto.Equal` can stay as a fallback tie-breaker for hash collisions if desired, but in practice hash equality is sufficient.

Parent: #16183

Automaat:
## Motivation

`proto.Equal` in the xDS reconcile hot path performs a deep protobuf walk on every reconcile cycle for every resource, even when nothing changed. Under load this is measurable CPU overhead. Additionally, the previous `equalSnapshots`/`reuseVersion` design reused old version strings verbatim, which could silently suppress delivery when version strings collided across restarts.

Closes https://github.com/kumahq/kuma/issues/16187

## Implementation information

Replace `proto.Equal` with deterministic **xxhash content hashes** seeded by `nodeId+type+index`:

- Delete `equalSnapshots` and `reuseVersion`; rewrite `autoVersion` with an error return to propagate `MarshalResource` failures
- Add `resourcesVersion` (empty slot → `""`, non-empty → hex hash) and `mixVersions` helpers
- Fold the cluster hash into the endpoint version to preserve EDS warming semantics without `NewUUID`
- Fix empty-version leak in `ConfigReadyForDelivery`: empty slots produce version `""` via `resourcesVersion`; calling `ConfigReadyForDelivery("")` would leave a permanent entry in `configsQueue` because `OnStreamRequest` short-circuits on `VersionInfo==""`. Added symmetric guard matching the existing `clearUndeliveredConfigStats` guard

### Benchmarks

Measured on Apple M4 Max over a realistic 400-resource snapshot (100 clusters / endpoints / routes / listeners), snapshot built once outside the timed loop so the benchmark isolates the diff logic rather than snapshot construction:

```
                    time/op      B/op       allocs/op
proto.Equal (old)   4.4–6.8 ms   96 KB      2400
xxhash (new)        0.8–1.5 ms   247 KB     3990
```

Replacing `proto.Equal` cuts diff CPU roughly 4–6x. The hash path marshals each resource to compute its digest, so it trades higher allocations for the latency win.

Tests added:
- EDS-fold integration (`It`)
- Empty-slot version behaviour (`It`)
- `BenchmarkAutoVersion`
- `BenchmarkReconcileUnchanged`

> Changelog: perf(xds): replace proto.Equal with content hashes in reconcile

### DIFF
--- a/pkg/xds/server/v3/reconcile.go
+++ b/pkg/xds/server/v3/reconcile.go
@@ -2,16 +2,18 @@ package v3
 
 import (
 	"context"
+	"hash/fnv"
+	"strconv"
 
 	envoy_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_types "github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	envoy_cache "github.com/envoyproxy/go-control-plane/pkg/cache/v3"
 	envoy_resource "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 	"github.com/pkg/errors"
-	"google.golang.org/protobuf/proto"
 
 	"github.com/kumahq/kuma/v3/pkg/core"
 	model "github.com/kumahq/kuma/v3/pkg/core/xds"
+	"github.com/kumahq/kuma/v3/pkg/util/maps"
 	util_xds "github.com/kumahq/kuma/v3/pkg/util/xds"
 	xds_context "github.com/kumahq/kuma/v3/pkg/xds/context"
 	"github.com/kumahq/kuma/v3/pkg/xds/generator"
@@ -59,18 +61,16 @@ func (r *reconciler) Reconcile(ctx context.Context, xdsCtx xds_context.Context, 
 		return false, errors.Wrapf(err, "failed to generate a snapshot")
 	}
 
-	// To avoid assigning a new version every time, compare with
-	// the previous snapshot and reuse its version whenever possible,
-	// fallback to UUID otherwise
+	// To avoid assigning a new version every time, compute stable content
+	// hashes per resource type and compare against the previous snapshot.
 	previous, err := r.cacher.Get(node)
 	if err != nil {
 		previous = &envoy_cache.Snapshot{}
 	}
 
-	snapshot, changed := autoVersion(previous, snapshot)
-	// We need to force new version of EDS, otherwise clusters will be stuck in warming state.
-	if previous.GetVersion(envoy_resource.ClusterType) != snapshot.GetVersion(envoy_resource.ClusterType) {
-		snapshot.Resources[envoy_types.Endpoint].Version = core.NewUUID()
+	snapshot, changed, err := autoVersion(node.Id, previous, snapshot)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to compute snapshot versions")
 	}
 
 	resKey := proxy.Id.ToResourceKey()
@@ -143,39 +143,64 @@ func validateResource(r envoy_types.Resource) error {
 	}
 }
 
-func autoVersion(old *envoy_cache.Snapshot, n *envoy_cache.Snapshot) (*envoy_cache.Snapshot, []string) {
-	for resourceType, resources := range old.Resources {
-		n.Resources[resourceType] = reuseVersion(resources, n.Resources[resourceType])
+// autoVersion computes deterministic FNV-64a content hashes for each resource
+// type in n, seeded with nodeId+type-index. Empty slots keep version "".
+// The cluster version is folded into the endpoint version when endpoints are
+// non-empty to force EDS re-push on cluster changes (prevents warming stalls).
+// Returns the versions that changed relative to old.
+func autoVersion(nodeId string, old, n *envoy_cache.Snapshot) (*envoy_cache.Snapshot, []string, error) {
+	for i := range n.Resources {
+		seed := nodeId + strconv.Itoa(i)
+		ver, err := resourcesVersion(seed, n.Resources[i].Items)
+		if err != nil {
+			return nil, nil, errors.Wrapf(err, "failed to hash resources for type %d", i)
+		}
+		n.Resources[i].Version = ver
+	}
+
+	// Fold cluster version into endpoint version so that a cluster change
+	// forces an EDS re-push even when endpoint content is identical.
+	// Only when endpoints are non-empty; empty slots stay at "".
+	if ep := n.Resources[envoy_types.Endpoint].Version; ep != "" {
+		n.Resources[envoy_types.Endpoint].Version = mixVersions(ep, n.Resources[envoy_types.Cluster].Version)
 	}
 
 	var changed []string
-	for resourceType, resource := range n.Resources {
-		if old.Resources[resourceType].Version != resource.Version {
+	for i, resource := range n.Resources {
+		if old.Resources[i].Version != resource.Version {
 			changed = append(changed, resource.Version)
 		}
 	}
 
-	return n, changed
+	return n, changed, nil
 }
 
-func reuseVersion(old, n envoy_cache.Resources) envoy_cache.Resources {
-	n.Version = old.Version
-	if !equalSnapshots(old.Items, n.Items) {
-		n.Version = core.NewUUID()
+// resourcesVersion returns a hex FNV-64a hash over sorted resource names and
+// their deterministic proto serializations, seeded with seed. Returns "" for
+// empty slots so that two empty slots compare equal without versioning.
+func resourcesVersion(seed string, items map[string]envoy_types.ResourceWithTTL) (string, error) {
+	if len(items) == 0 {
+		return "", nil
 	}
-	return n
-}
-
-func equalSnapshots(old, n map[string]envoy_types.ResourceWithTTL) bool {
-	if len(n) != len(old) {
-		return false
-	}
-	for key, newValue := range n {
-		if oldValue, hasOldValue := old[key]; !hasOldValue || !proto.Equal(newValue.Resource, oldValue.Resource) {
-			return false
+	h := fnv.New64a()
+	h.Write([]byte(seed))
+	for _, key := range maps.SortedKeys(items) {
+		h.Write([]byte(key))
+		b, err := envoy_cache.MarshalResource(items[key].Resource)
+		if err != nil {
+			return "", err
 		}
+		h.Write(b)
 	}
-	return true
+	return strconv.FormatUint(h.Sum64(), 16), nil
+}
+
+// mixVersions combines two version strings into a single FNV-64a hash.
+func mixVersions(a, b string) string {
+	h := fnv.New64a()
+	h.Write([]byte(a))
+	h.Write([]byte(b))
+	return strconv.FormatUint(h.Sum64(), 16)
 }
 
 type snapshotGenerator interface {
--- a/pkg/xds/server/v3/reconcile.go
+++ b/pkg/xds/server/v3/reconcile.go
@@ -99,7 +99,7 @@ func (r *reconciler) Reconcile(ctx context.Context, xdsCtx xds_context.Context, 
 	if err := snapshot.Consistent(); err != nil {
 		return false, errors.Wrap(err, "inconsistent snapshot")
 	}
-	log.Info("config has changed", "versions", changed)
+	log.Info("config has changed", "versions", versionStrings(changed))
 
 	if err := r.cacher.Cache(ctx, node, snapshot); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/xds/server/v3/reconcile_test.go
```diff
@@ -251,11 +238,323 @@ var _ = Describe("Reconcile", func() {
...
+
 type snapshotGeneratorFunc func(context.Context, xds_context.Context, *xds_model.Proxy) (*envoy_cache.Snapshot, error)
 
 func (f snapshotGeneratorFunc) GenerateSnapshot(ctx context.Context, xdsCtx xds_context.Context, proxy *xds_model.Proxy) (*envoy_cache.Snapshot, error) {
 	return f(ctx, xdsCtx, proxy)
 }
+
+// buildRealisticSnapshot returns a snapshot with 20 clusters, 20 endpoints,
+// 5 listeners, and 5 routes — representative of a non-trivial data plane.
+func buildRealisticSnapshot() *envoy_cache.Snapshot {
```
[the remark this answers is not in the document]
Automaat: **Applied** — dropped the bogus "5 routes" from the doc comment; `buildRealisticSnapshot()` only builds clusters/endpoints/listeners (b31caaab25).

pkg/xds/server/v3/reconcile_test.go
```diff
@@ -251,11 +238,323 @@ var _ = Describe("Reconcile", func() {
...
+		endpointItems[name] = envoy_types.ResourceWithTTL{
+			Resource: &envoy_endpoint.ClusterLoadAssignment{ClusterName: name},
+		}
+	}
+	snap.Resources[envoy_types.Cluster] = envoy_cache.Resources{Items: clusterItems}
+	snap.Resources[envoy_types.Endpoint] = envoy_cache.Resources{Items: endpointItems}
+
+	// Listeners with static (non-RDS) route configs so no Route resources are
+	// required for Consistent() to pass.
+	listenerItems := make(map[string]envoy_types.ResourceWithTTL, 5)
```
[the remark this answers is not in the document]
Automaat: **Applied** — reworded to "Listeners reference no RDS route configs, so no Route resources are required for Consistent() to pass", matching what is actually built (b31caaab25).

pkg/xds/server/v3/reconcile_test.go
```diff
@@ -251,11 +238,323 @@ var _ = Describe("Reconcile", func() {
...
+			Expect(cached.GetVersion(resource.ListenerType)).To(BeEmpty())
+			Expect(cached.GetVersion(resource.RouteType)).To(BeEmpty())
+			Expect(cached.GetVersion(resource.ClusterType)).To(BeEmpty())
+			Expect(cached.GetVersion(resource.EndpointType)).To(BeEmpty())
+
+			By("second identical reconcile — no changes")
+			changed, err = r.Reconcile(context.Background(), xds_context.Context{}, proxy)
+			Expect(err).ToNot(HaveOccurred())
+			// Unchanged: same content hash for Route, "" == "" for empty types.
+			Expect(changed).To(BeFalse())
```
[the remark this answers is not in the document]
Automaat: **Applied** — comment now says "same content hash for Secret"; this test only populates Secret resources (b31caaab25).

pkg/xds/server/v3/reconcile.go
```diff
@@ -143,39 +145,64 @@ func validateResource(r envoy_types.Resource) error {
...
+	for _, key := range maps.SortedKeys(items) {
+		h.Write([]byte(key))
+		b, err := envoy_cache.MarshalResource(items[key].Resource)
+		if err != nil {
+			return "", err
 		}
+		h.Write(b)
 	}
-	return true
+	return strconv.FormatUint(h.Sum64(), 16), nil
```
slonka: nit: `strconv.FormatUint(h.Sum64(), 16)` drops leading zeros so the version string isn't fixed-width. combined with `mixVersions` concatenating `a`+`b` without a separator (line 203-204), `mix("ab", "cdef")` and `mix("abcd", "ef")` hash the same. tiny probability but `fmt.Sprintf("%016x", h.Sum64())` makes it impossible.

pkg/xds/server/v3/reconcile.go
```diff
@@ -143,39 +145,64 @@ func validateResource(r envoy_types.Resource) error {
...
-			return false
+	h := fnv.New64a()
+	h.Write([]byte(seed))
+	for _, key := range maps.SortedKeys(items) {
+		h.Write([]byte(key))
+		b, err := envoy_cache.MarshalResource(items[key].Resource)
+		if err != nil {
+			return "", err
 		}
+		h.Write(b)
```
slonka: nit: no separator between `h.Write([]byte(key))` and `h.Write(b)`, or between successive items. two snapshots with different (key, marshaled-bytes) splits could in theory hash the same. unlikely with structured proto bytes but a `h.Write([]byte{0})` between fields would close the door.

pkg/xds/server/v3/reconcile.go
```diff
@@ -125,7 +125,9 @@ func (r *reconciler) Reconcile(ctx context.Context, xdsCtx xds_context.Context,
 	}
 
 	for _, version := range changed {
-		r.statsCallbacks.ConfigReadyForDelivery(version)
+		if version != "" {
+			r.statsCallbacks.ConfigReadyForDelivery(version)
```
slonka: the guard fixes the `configsQueue` leak, but for resource-type-becomes-empty transitions we also lose the ACK/delivery metric for that type (Envoy ACKs with `VersionInfo==""` → `OnStreamRequest` short-circuits). worth a comment noting that's intentional or doing something so metrics still fire

pkg/xds/server/v3/reconcile.go
```diff
@@ -143,39 +145,64 @@ func validateResource(r envoy_types.Resource) error {
...
-}
-
-func equalSnapshots(old, n map[string]envoy_types.ResourceWithTTL) bool {
-	if len(n) != len(old) {
-		return false
-	}
-	for key, newValue := range n {
-		if oldValue, hasOldValue := old[key]; !hasOldValue || !proto.Equal(newValue.Resource, oldValue.Resource) {
-			return false
+	h := fnv.New64a()
```
slonka: there are no before/after numbers in the PR description - we can't tell if this is a meaningful improvement.  Need `benchstat` against `master` to verify. Side note: `cespare/xxhash/v2` is ~4-6x faster than FNV-64a with the same 64-bit collision space.

pkg/xds/server/v3/reconcile.go
```diff
@@ -143,39 +145,64 @@ func validateResource(r envoy_types.Resource) error {
...
-	for resourceType, resources := range old.Resources {
-		n.Resources[resourceType] = reuseVersion(resources, n.Resources[resourceType])
+// autoVersion computes deterministic FNV-64a content hashes for each resource
+// type in n, seeded with nodeId+type-index. Empty slots keep version "".
+// The cluster version is folded into the endpoint version when endpoints are
+// non-empty to force EDS re-push on cluster changes (prevents warming stalls).
+// Returns the versions that changed relative to old.
+func autoVersion(nodeId string, old, n *envoy_cache.Snapshot) (*envoy_cache.Snapshot, []string, error) {
+	for i := range n.Resources {
+		seed := nodeId + strconv.Itoa(i)
```
slonka: why is the seed needed at all? versions are stored per `(node, type)` in the cache and never cross-compared, so two unrelated types/nodes producing the same hash is operationally fine. What failure mode is this guarding against?

pkg/xds/server/v3/reconcile_test.go
```diff
@@ -251,11 +238,323 @@ var _ = Describe("Reconcile", func() {
...
+
 type snapshotGeneratorFunc func(context.Context, xds_context.Context, *xds_model.Proxy) (*envoy_cache.Snapshot, error)
 
 func (f snapshotGeneratorFunc) GenerateSnapshot(ctx context.Context, xdsCtx xds_context.Context, proxy *xds_model.Proxy) (*envoy_cache.Snapshot, error) {
 	return f(ctx, xdsCtx, proxy)
 }
+
+// buildRealisticSnapshot returns a snapshot with 20 clusters, 20 endpoints,
+// and 5 listeners — representative of a non-trivial data plane.
+func buildRealisticSnapshot() *envoy_cache.Snapshot {
```
slonka: Benchmark understates prod cost. real Kuma snapshots have O(100-1000) clusters with `TransportSocket`/`CircuitBreakers`/`HealthChecks`, CLAs with multiple `LbEndpoints`, listeners with full `FilterChains` — `MarshalResource` cost scales with field count and nesting depth. these stubs marshal to ~10 bytes each.

Would be more representative to load a snapshot from a policy plugin testdata fixture or `pkg/test/xds/`. 

Also missing a realistic "1 of N resources changed" case

pkg/xds/server/v3/reconcile.go
Automaat: **Applied** — versions are now fixed-width 16-character hex strings via `fmt.Sprintf("%016x", ...)`, and `mixVersions` frames both inputs before hashing (fbd48dbe97).

_Generated by Sybra harness_

pkg/xds/server/v3/reconcile.go
Automaat: **Applied** — resource hashing now length-frames the seed, each resource key, and each marshaled proto payload before hashing; I also added regression coverage for framed/fixed-width hashes (fbd48dbe97).

_Generated by Sybra harness_

pkg/xds/server/v3/reconcile.go
Automaat: **Applied** — added a production-code comment on the empty-version guard explaining that Envoy ACKs empty resource-type transitions with `VersionInfo==""`, and `StatsCallbacks` treats that as an initial request, so enqueueing it would leak delivery metric state (fbd48dbe97).

_Generated by Sybra harness_

pkg/xds/server/v3/reconcile.go
Automaat: **Applied** — switched the content hash from FNV-64a to `cespare/xxhash/v2`, expanded the benchmark fixture, and captured benchstat against the merge base using the same fixture: `BenchmarkReconcileUnchanged` went from 1297.9µs/op to 521.8µs/op, -59.79% (p=0.008, n=5) (fbd48dbe97).

_Generated by Sybra harness_

pkg/xds/server/v3/reconcile.go
Automaat: **Applied** — documented that the seed keeps delivery metric keys distinct because `StatsCallbacks` tracks in-flight configs by version string in a shared map, not by `(node,type)` (fbd48dbe97).

_Generated by Sybra harness_

pkg/xds/server/v3/reconcile_test.go
Automaat: **Applied** — replaced the shallow benchmark fixture with nested xDS resources: 100 clusters with transport sockets, circuit breakers, and health checks; 100 CLAs with multiple endpoints; 100 RDS routes; and 100 listeners with HCM filter chains. Added a `one-cluster-changed` benchmark case as well (fbd48dbe97).

_Generated by Sybra harness_

pkg/xds/server/v3/reconcile.go
slonka: please post the SHA of the benchmark run against master

pkg/xds/server/v3/reconcile.go
Automaat: **Applied** — removed the node/type seed from content hashes; snapshot versions now hash only the sorted resource names and deterministic proto bytes, which matches the per-node/per-type cache comparison scope (52e0a70145).

_Generated by Sybra harness_

pkg/xds/server/v3/reconcile.go
Automaat: **Applied** — the benchmark comparison against `master` used merge-base SHA `36ae6234047cafa5c46a9823a5b4dad6f360233c`.

_Generated by Sybra harness_

slonka (approved):

Issue #17001, opened 2026-06-23
☂️ e2e v3 rework — Kuma 3.0 readiness (all environments)


Automaat:
## Motivation

Kuma 3.0 removes a batch of resources, Mesh fields, Dataplane fields, and control-plane
settings. The e2e suites (universal, kubernetes, multizone, under `test/e2e_env/*`) still
exercise many of them. Some are the subject of a test, existing only to cover a removed
resource; others are incidental scaffolding. Either way they have to change before the suites
can run on 3.0.

This umbrella tracks that rework across all three environments.

## Environments audited

Each suite ran with a `kuma3-preflight` capture hook, an opt-in `AfterEach` that audits the
live CP after every spec, and the results were classified:

| Env | Specs passed | Snapshots | Capture target |
|---|---|---|---|
| universal | 250/252 | 251 | the CP |
| kubernetes | 158/163 | 163 | port-forwarded CP |
| multizone | 295/297 | 297 | the **global** CP (per-zone config fanned out via `/zones+insights`) |

The handful of failures were all local-env flakes (ipv6, KIC DNS, Spire, cross-mesh mTLS,
Eviction), not product issues.

## What the data shows

A shared backbone shows up identically across all three envs, and it's the bulk of the work:
inline mTLS on Mesh → MeshIdentity+MeshTrust (universal 16 / k8s 18 / multizone 41 meshes),
MeshTrafficPermission `from`, builtin gateway, `reachableServices`, MeshHealthCheck
`healthyPanicThreshold`, Mesh `passthrough`/`routing.zoneEgress`, and CP config
(deltaXds/inboundTags/KDS-watchdog/`meshServices.mode`).

On top of that sit the env-specific deltas (see WS6/WS7 and the per-env reports):

- **kubernetes**: `unifiedResourceNamingEnabled` off (injector, doesn't exist on universal),
  `MeshGatewayRoute`, pod-annotation Dataplane metrics, `MeshPassthrough/MeshProxyPatch/
  MeshMetric` deprecated fields, `defaultForbidMeshExternalServiceAccess`.
- **multizone**: per-zone CP config (the global only fans values out, so each zone CP must be
  flipped), `localityAwareLoadBalancing` on Mesh, non-RFC-1035 KDS-synced MeshService names.

PR #17011, opened 2026-06-23
test(e2e): weekly v3-readiness preflight capture

Automaat:
## Motivation

Kuma 3.0 removes many resources, Mesh/Dataplane fields and CP settings the e2e suites
still exercise. We want a weekly that shows, per environment, which tests still use
removed features and how that burns down over time — without adding any load to PR /
regular e2e CI.

> Changelog: skip

## Implementation information

- `test/framework/preflight.go` — opt-in `CapturePreflightCluster` (env-gated by
  `KUMA3_PREFLIGHT_BIN` / `KUMA3_PREFLIGHT_DIR`), wired as a top-level `AfterEach` in the
  three suite bootstraps. **No-op on PR / regular runs**; multizone captures the global
  CP. Each capture runs under a hard `context` timeout, and snapshot filenames carry the
  Ginkgo parallel-process index so concurrent processes never collide.
- `.github/workflows/e2e-v3-readiness.yml` — reusable (`workflow_call`) + weekly
  (`schedule`). Reuses the repo's `run-e2e` action to build images and run
  `make test/e2e-<env>` with capture, then classifies the snapshots into a job-summary +
  artifact report (and **fails if nothing was captured**, so a broken weekly goes red).
- The kuma3-preflight tool is **not vendored**: the jobs download a checksum-verified
  **GoReleaser** release binary from `Kong/kong-mesh-v3-readiness`. The pinned
  `tool_version` is kept current by a **Renovate** custom manager (`renovate.json`).
- kong-mesh reuses this workflow via `uses:` (it `include`s `mk/e2e.new.mk`, so the same
  `make test/e2e-<env>` targets exist there).

Collection is **weekly-only**: the existing `run-e2e` action and `_test.yaml` are
untouched, and the hook is inert unless the weekly sets the two env vars.

Draft until: `Kong/kong-mesh-v3-readiness` publishes a `v0.1.0` release the download
targets (and its releases are reachable from CI), plus a first green weekly on a
large/self-hosted runner.

## Supporting documentation

Drives the e2e v3 rework umbrella kumahq/kuma#17001.

### DIFF
--- a/.github/workflows/e2e-v3-readiness.yml
+++ b/.github/workflows/e2e-v3-readiness.yml
@@ -1,0 +1,135 @@
+# Weekly "Kuma 3.0 e2e readiness" report (also runnable on demand).
+#
+# Runs the e2e suites with the kuma3-preflight capture hook (an opt-in AfterEach that
+# audits the live CP after every spec — a no-op unless KUMA3_PREFLIGHT_BIN +
+# KUMA3_PREFLIGHT_DIR are set), then classifies the snapshots into a "what still uses
+# Kuma-3.0-removed features" report shown in the run's job summary.
+#
+# kuma3-preflight is downloaded (checksum-verified) from the public GoReleaser release
+# in github.com/Kong/kong-mesh-v3-readiness — not vendored.
+#
+# kong-mesh consumes this by COPY, not call: its tools/dev/sync_ci.sh syncs it to
+# `kuma-e2e-v3-readiness.yaml`, patching in a kuma checkout (its e2e make includes
+# $(KUMA_DIR)/mk/...) and Kong runners. Keep `runs-on: ubuntu-latest` literal so the
+# sync's use_custom_runners can swap it.
+
+name: E2E v3 readiness
+
+on:
+  schedule:
+    - cron: "0 1 * * 1" # weekly, Monday 01:00 UTC
+  workflow_dispatch: {} # on-demand
+
+permissions:
+  contents: read
+
+env:
+  # kuma3-preflight release to download; kept current by Renovate (see renovate.json).
+  # renovate[github-releases]: depName=Kong/kong-mesh-v3-readiness
+  TOOL_VERSION: v0.1.0
+
+jobs:
+  audit:
+    strategy:
+      fail-fast: false
+      matrix:
+        env: [universal, kubernetes, multizone]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Download kuma3-preflight (checksum-verified)
+        run: |
+          set -euo pipefail
+          base="https://github.com/Kong/kong-mesh-v3-readiness/releases/download/${TOOL_VERSION}"
+          asset="kuma3-preflight_linux_amd64.tar.gz"
+          mkdir -p "${RUNNER_TEMP}/bin"
+          curl -fsSL "${base}/${asset}" -o "${RUNNER_TEMP}/${asset}"
+          curl -fsSL "${base}/checksums.txt" -o "${RUNNER_TEMP}/checksums.txt"
+          ( cd "${RUNNER_TEMP}" && grep " ${asset}\$" checksums.txt | sha256sum -c - )
+          tar -xzf "${RUNNER_TEMP}/${asset}" -C "${RUNNER_TEMP}/bin" kuma3-preflight
+          {
+            echo "KUMA3_PREFLIGHT_BIN=${RUNNER_TEMP}/bin/kuma3-preflight"
+            echo "KUMA3_PREFLIGHT_DIR=${RUNNER_TEMP}/preflight-${{ matrix.env }}"
+          } >> "$GITHUB_ENV"
+
+      # Reuse the repo's own e2e action so the CP/image build matches normal CI exactly.
+      # It inherits KUMA3_PREFLIGHT_BIN/DIR from $GITHUB_ENV, so the hook captures per spec.
+      - name: Run e2e (${{ matrix.env }}) with capture
+        uses: ./.github/actions/run-e2e
+        with:
+          target: ${{ matrix.env }}
+          k8s_version: kind
+          cni_network_plugin: flannel
+          arch: amd64
+          parallel_runner_id: "0"
+
+      - name: Upload snapshots
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: preflight-${{ matrix.env }}
+          path: ${{ runner.temp }}/preflight-${{ matrix.env }}/*.json
+          if-no-files-found: warn
+          retention-days: 30
+
+  report:
+    needs: audit
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Download kuma3-preflight (checksum-verified)
+        run: |
+          set -euo pipefail
+          base="https://github.com/Kong/kong-mesh-v3-readiness/releases/download/${TOOL_VERSION}"
+          asset="kuma3-preflight_linux_amd64.tar.gz"
+          mkdir -p "${RUNNER_TEMP}/bin"
+          curl -fsSL "${base}/${asset}" -o "${RUNNER_TEMP}/${asset}"
+          curl -fsSL "${base}/checksums.txt" -o "${RUNNER_TEMP}/checksums.txt"
+          ( cd "${RUNNER_TEMP}" && grep " ${asset}\$" checksums.txt | sha256sum -c - )
+          tar -xzf "${RUNNER_TEMP}/${asset}" -C "${RUNNER_TEMP}/bin" kuma3-preflight
+
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          pattern: preflight-*
+          path: ${{ runner.temp }}/snapshots
+
+      - name: Classify + write report to the job summary
+        run: |
+          set -euo pipefail
+          bin="${RUNNER_TEMP}/bin/kuma3-preflight"
+          mkdir -p report
+          echo "# Kuma 3.0 e2e readiness — $(date -u +%Y-%m-%d)" >> "$GITHUB_STEP_SUMMARY"
+          echo >> "$GITHUB_STEP_SUMMARY"
+          captured=0
+          for env in universal kubernetes multizone; do
+            dir="${RUNNER_TEMP}/snapshots/preflight-${env}"
+            src="test/e2e_env/${env}"
+            if [ ! -d "$dir" ] || [ -z "$(ls -A "$dir" 2>/dev/null)" ] || [ ! -d "$src" ]; then
+              echo "- _${env}: no snapshots, skipped_" >> "$GITHUB_STEP_SUMMARY"
+              continue
+            fi
+            # One env's classify failure (e.g. a corrupt snapshot) must not sink the report.
+            if ! "$bin" --classify --source-dir "$src" --reports-dir "$dir" --format markdown --output "report/${env}.md"; then
+              echo "::warning::classify failed for ${env}"
+              continue
+            fi
+            captured=$((captured + 1))
+            "$bin" --classify --source-dir "$src" --reports-dir "$dir" --format json --output "report/${env}.json" || true
+            "$bin" --classify --source-dir "$src" --reports-dir "$dir" --format html --output "report/${env}.html" || true
+            { cat "report/${env}.md"; echo; } >> "$GITHUB_STEP_SUMMARY"
+          done
+          if [ "$captured" -eq 0 ]; then
+            echo "::error::no snapshots captured for any env (CP never came up, or the tool was not downloaded)"
+            exit 1
+          fi
+
+      - name: Upload report
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: v3-readiness-report
+          path: report
+          retention-days: 90
--- a/.github/workflows/e2e-v3-readiness.yml
+++ b/.github/workflows/e2e-v3-readiness.yml
@@ -7,11 +7,6 @@
 #
 # kuma3-preflight is downloaded (checksum-verified) from the public GoReleaser release
 # in github.com/Kong/kong-mesh-v3-readiness — not vendored.
-#
-# kong-mesh consumes this by COPY, not call: its tools/dev/sync_ci.sh syncs it to
-# `kuma-e2e-v3-readiness.yaml`, patching in a kuma checkout (its e2e make includes
-# $(KUMA_DIR)/mk/...) and Kong runners. Keep `runs-on: ubuntu-latest` literal so the
-# sync's use_custom_runners can swap it.
 
 name: E2E v3 readiness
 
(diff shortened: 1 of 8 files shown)


test/framework/preflight.go
```diff
@@ -0,0 +1,100 @@
...
+	if bin == "" || dir == "" {
+		return
+	}
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		ginkgo.GinkgoWriter.Printf("preflight: mkdir %s: %v\n", dir, err)
+		return
+	}
+	seq := preflightSeq.Add(1)
+	out := filepath.Join(dir, fmt.Sprintf("%04d-%s.json", seq, preflightSlug(specName)))
+	cmd := exec.Command(bin, "--address", addr, "--format", "json", "--output", out, "--timeout", "30s")
```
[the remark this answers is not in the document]
Automaat: Fixed: the snapshot filename now includes `ginkgo.GinkgoParallelProcess()` — `p%02d-%05d-<slug>.json` — so parallel processes sharing one KUMA3_PREFLIGHT_DIR can't clobber each other.

tools/kuma3-preflight/main.go
```diff
@@ -0,0 +1,231 @@
...
+		return reportModel{}, err
+	}
+	var m reportModel
+	if err := json.Unmarshal(data, &m); err != nil {
+		return reportModel{}, fmt.Errorf("parsing JSON report: %w", err)
+	}
+	if m.Schema == "" {
+		return reportModel{}, fmt.Errorf("does not look like a %s JSON report (missing schema)", toolName)
+	}
+	return m, nil
```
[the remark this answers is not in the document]
Automaat: No longer applicable here: the tool is no longer vendored in this PR — `kuma3-preflight` is consumed as a published GoReleaser binary from Kong/kong-mesh-v3-readiness. `loadModel` was hardened there to require the `kuma3-preflight/` report-schema prefix (rejecting foreign/classification JSON).

test/e2e_env/universal/universal_suite_test.go
```diff
@@ -53,6 +53,12 @@ var (
 	_ = E2ESynchronizedBeforeSuite(universal.SetupAndGetState, universal.RestoreState)
 	_ = SynchronizedAfterSuite(func() {}, universal.SynchronizedAfterSuite)
 	_ = ReportAfterSuite("universal after suite", universal.AfterSuite)
+	// Opt-in (KUMA3_PREFLIGHT_BIN + KUMA3_PREFLIGHT_DIR): snapshot the shared CP after
+	// each spec so kuma3-preflight can classify which tests use Kuma-3.0-removed
+	// features. No-op otherwise. The suite runs --procs 1, so snapshots are serial.
```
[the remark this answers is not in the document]
Automaat: Fixed: the misleading "--procs 1" note was replaced. Capture relies on the parallel-process index in the snapshot filename for uniqueness, not on serial execution.

test/framework/preflight.go
```diff
@@ -0,0 +1,112 @@
...
+// safeAPIServerAddress returns the cluster's CP API address. It recovers from a panic
+// (a K8s CP whose port-forward is not established panics in GetAPIServerAddress) and
+// returns "" so capture stays best-effort and never fails a test.
+func safeAPIServerAddress(c Cluster) string {
+	defer func() { _ = recover() }()
+	if c == nil {
+		return ""
+	}
+	return c.GetKuma().GetAPIServerAddress()
+}
```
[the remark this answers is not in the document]
Automaat: Fixed: safeAPIServerAddress now logs the recovered value via ginkgo.GinkgoWriter before returning "" (still best-effort, never fails the spec).

.github/workflows/e2e-v3-readiness.yml
```diff
@@ -0,0 +1,135 @@
...
+# kong-mesh consumes this by COPY, not call: its tools/dev/sync_ci.sh syncs it to
+# `kuma-e2e-v3-readiness.yaml`, patching in a kuma checkout (its e2e make includes
+# $(KUMA_DIR)/mk/...) and Kong runners. Keep `runs-on: ubuntu-latest` literal so the
+# sync's use_custom_runners can swap it.
+
+name: E2E v3 readiness
+
+on:
+  schedule:
+    - cron: "0 1 * * 1" # weekly, Monday 01:00 UTC
```
slonka, line 15: nitpick: this is weekly but PR body says nightly

.github/workflows/e2e-v3-readiness.yml
```diff
@@ -0,0 +1,135 @@
...
+      # Reuse the repo's own e2e action so the CP/image build matches normal CI exactly.
+      # It inherits KUMA3_PREFLIGHT_BIN/DIR from $GITHUB_ENV, so the hook captures per spec.
+      - name: Run e2e (${{ matrix.env }}) with capture
+        uses: ./.github/actions/run-e2e
+        with:
+          target: ${{ matrix.env }}
+          k8s_version: kind
+          cni_network_plugin: flannel
+          arch: amd64
+          parallel_runner_id: "0"
```
slonka, line 58: `parallel_runner_id: "0"` hardcodes single-process Ginkgo runs and we have 

```
name := fmt.Sprintf("p%02d-%05d-%s.json", ginkgo.GinkgoParallelProcess(), seq, preflightSlug(specName))
```

is the intent to be gradually run this as multi process?

test/framework/preflight.go
```diff
@@ -0,0 +1,118 @@
...
+		ginkgo.GinkgoWriter.Printf("preflight: mkdir %s: %v\n", dir, err)
+		return
+	}
+	seq := preflightSeq.Add(1)
+	name := fmt.Sprintf("p%02d-%05d-%s.json", ginkgo.GinkgoParallelProcess(), seq, preflightSlug(specName))
+	out := filepath.Join(dir, name)
+
+	ctx, cancel := context.WithTimeout(context.Background(), captureTimeout)
+	defer cancel()
+	cmd := exec.CommandContext(ctx, bin, "--address", addr, "--format", "json", "--output", out, "--timeout", "45s")
```
slonka: Both timeouts (`captureTimeout = 60s` and `--timeout 45s`) are constants. Consider making at least the tool `--timeout` overridable from an env var (e.g. `KUMA3_PREFLIGHT_TIMEOUT`, default 45s) so the nightly can be tuned without a code change.

test/framework/preflight.go
```diff
@@ -0,0 +1,118 @@
...
+		return
+	}
+	var ee *exec.ExitError
+	if errors.As(err, &ee) {
+		switch ee.ExitCode() {
+		case 1, 3: // blockers / inconclusive: expected, snapshot still written
+			return
+		}
+	}
+	ginkgo.GinkgoWriter.Printf("preflight: capture failed for %q: %v\n%s\n", specName, err, combined)
```
slonka, line 134: Capture problems are written to `GinkgoWriter` only. With `if-no-files-found: warn` on the artifact upload, a totally-broken capture (exit 2, exec failure, timeout) shows up as `::error::no snapshots captured` in the `report` job but the *reason* lives buried in the matrix job's spec logs. Worth piping the failure summary to `$GITHUB_STEP_SUMMARY` (when set) or printing a `::warning::preflight capture failed: ...` line so the nightly's failure mode is one click away.

.github/workflows/e2e-v3-readiness.yml
```diff
@@ -0,0 +1,135 @@
...
+# KUMA3_PREFLIGHT_DIR are set), then classifies the snapshots into a "what still uses
+# Kuma-3.0-removed features" report shown in the run's job summary.
+#
+# kuma3-preflight is downloaded (checksum-verified) from the public GoReleaser release
+# in github.com/Kong/kong-mesh-v3-readiness — not vendored.
+#
+# kong-mesh consumes this by COPY, not call: its tools/dev/sync_ci.sh syncs it to
+# `kuma-e2e-v3-readiness.yaml`, patching in a kuma checkout (its e2e make includes
+# $(KUMA_DIR)/mk/...) and Kong runners. Keep `runs-on: ubuntu-latest` literal so the
+# sync's use_custom_runners can swap it.
```
slonka: nitpick: seems like internal agent dialogue, all workflows are synced using `sync_ci.sh` so this is no different

.github/workflows/e2e-v3-readiness.yml
```diff
@@ -0,0 +1,135 @@
...
+  schedule:
+    - cron: "0 1 * * 1" # weekly, Monday 01:00 UTC
+  workflow_dispatch: {} # on-demand
+
+permissions:
+  contents: read
+
+env:
+  # kuma3-preflight release to download; kept current by Renovate (see renovate.json).
+  # renovate[github-releases]: depName=Kong/kong-mesh-v3-readiness
```
slonka: why don't we use 

```suggestion
  # renovate: datasource=github-releases depName=Kong/kong-mesh-v3-readiness versioning=semver
```

(like for ci-tools/release-tool)

also why does it need to be in km?

.github/workflows/e2e-v3-readiness.yml
```diff
@@ -0,0 +1,135 @@
...
+          retention-days: 30
+
+  report:
+    needs: audit
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Download kuma3-preflight (checksum-verified)
```
slonka, line 76: nitpick: Download kuma3-preflight could be a composite action so it's DRY

test/e2e_env/multizone/multizone_suite_test.go
```diff
@@ -47,6 +47,11 @@ var (
 	_ = E2ESynchronizedBeforeSuite(multizone.SetupAndGetState, multizone.RestoreState)
 	_ = SynchronizedAfterSuite(func() {}, multizone.SynchronizedAfterSuite)
 	_ = ReportAfterSuite("multizone after suite", multizone.AfterSuite)
+	// Opt-in (KUMA3_PREFLIGHT_BIN + KUMA3_PREFLIGHT_DIR): snapshot the GLOBAL CP after each
+	// spec — one audit of the global covers every zone (resources sync over KDS). No-op otherwise.
```
slonka, line 51: Do we have resources not synced from zone to global that could be missed by this mechanism?

test/framework/preflight.go
```diff
@@ -0,0 +1,118 @@
...
+const captureTimeout = 60 * time.Second
+
+var preflightSeq atomic.Int64
+
+var preflightSlugRe = regexp.MustCompile(`[^a-zA-Z0-9]+`)
+
+// PreflightCaptureEnabled reports whether snapshot capture is configured. Callers
+// guard on it before resolving the CP address so a normal run does no extra work.
+func PreflightCaptureEnabled() bool {
+	return os.Getenv("KUMA3_PREFLIGHT_BIN") != "" && os.Getenv("KUMA3_PREFLIGHT_DIR") != ""
```
slonka: why are we using raw env vars when we have https://github.com/slonka/kuma/blob/cf3e7dba7a17d7d629d8332902d5e14c67bcae70/test/framework/config.go#L21 ?

test/framework/preflight.go
```diff
@@ -0,0 +1,118 @@
...
+	if c == nil {
+		return ""
+	}
+	return c.GetKuma().GetAPIServerAddress()
+}
+
+func preflightSlug(s string) string {
+	s = strings.ToLower(strings.Trim(preflightSlugRe.ReplaceAllString(s, "-"), "-"))
+	if len(s) > 80 {
+		s = s[:80]
```
slonka: `preflightSlug` truncates at 80 chars, but `CurrentSpecReport().FullText()` includes the parent `Describe`/`Context` so long spec names collide on the slug - e.g. `MeshTrafficPermission API should allow the traffic with meshtrafficpermission based on MeshService (http)` and the `(tcp)` variant both slug to the same 80-char prefix.

Files stay unique thanks to the `p%02d-%05d-` counter prefix, so this isn't a correctness bug, but the slug stops being a useful human identifier.

.github/workflows/e2e-v3-readiness.yml
Automaat: **Applied** — reworded the PR body nightly→weekly to match the Monday cron (22d980aaa0). Implementation was always weekly; only the description had drifted.

test/framework/preflight.go
Automaat: **Applied** — the tool `--timeout` is now `KUMA3_PREFLIGHT_TIMEOUT` (an `E2eConfig` field, default 45s), and the hard context bound is derived as timeout+15s so it still outlives the tool's own stop (22d980aaa0).

test/framework/preflight.go
Automaat: **Applied** — capture failures now also print `::warning::preflight capture failed: …` to stdout. GinkgoWriter is only flushed on spec failure, so this surfaces the reason as a GH annotation instead of burying it in passing-spec logs (22d980aaa0).

.github/workflows/e2e-v3-readiness.yml
Automaat: **Applied** — dropped the downstream-sync paragraph from the header. It's synced like every other workflow via `sync_ci.sh`, so the note added nothing (22d980aaa0).

.github/workflows/e2e-v3-readiness.yml
Automaat: **Applied** — extracted the checksum-verified download into `.github/actions/download-kuma3-preflight` (outputs the bin path); both the audit and report jobs now use it (22d980aaa0).

test/framework/preflight.go
Automaat: **Applied** — BIN/DIR (plus the new TIMEOUT) are now `E2eConfig` fields with `envconfig` tags, read via `Config.*` instead of raw `os.Getenv` (22d980aaa0).

test/framework/preflight.go
Automaat: **Applied** — `preflightSlug` now keeps both ends when over the limit (raised to 100): the head names the feature area, the tail preserves the discriminator (e.g. `-http`/`-tcp`), so sibling specs no longer collapse to one slug. Filenames were already unique via the counter prefix; this just restores the slug as a human identifier (22d980aaa0).

.github/workflows/e2e-v3-readiness.yml
Automaat: **Applied** — switched to the standard `# renovate: datasource=github-releases depName=Kong/kong-mesh-v3-readiness versioning=semver` and removed the bespoke custom manager from `renovate.json`. The shared `Kong/public-shared-renovate` preset already parses that inline form (same as `ci-tools/release-tool` in release.yaml), so the custom manager was redundant (22d980aaa0).

On "why km": the binary ships from `Kong/kong-mesh-v3-readiness` because the 3.0-removed-feature catalog it classifies against lives there with the downstream readiness work; kuma only downloads the published artifact rather than duplicating that catalog here. Open to relocating it if you'd prefer a different home — leaving that part for your call.

.github/workflows/e2e-v3-readiness.yml
Automaat: **Skipped (no change needed)** — it isn't the Ginkgo process count. `parallel_runner_id` only feeds `GINKGO_E2E_LABEL_FILTERS=job-N`, and `run-e2e` applies that filter *only when `target` is empty* (the legacy `test/e2e/` job-sharded suites). Here `target` is set (universal/kubernetes/multizone), so the filter is skipped and the whole env suite runs — `"0"` just suffixes the debug/report artifact names. In-process parallelism comes from GINKGO_OPTS, and the env targets pin `--procs 1` (mk/e2e.new.mk); the `GinkgoParallelProcess()` filename prefix is defensive so capture stays collision-free if that ever changes. No multi-process rollout is gated on this value, so leaving it as-is.

test/e2e_env/multizone/multizone_suite_test.go
Automaat: Good question. Global-only capture covers everything KDS syncs — policies global→zone, and Dataplanes/insights/ZoneIngress/ZoneEgress zone→global — but it does **not** see (a) per-zone-CP config/settings (not synced) or (b) any resource applied directly to a zone CP that never syncs upward. These multizone specs drive config through the global CP, so in practice the only gap is zone-CP settings. Leaving it global-only for now since snapshotting every zone multiplies captures; I can add per-zone capture if you think the zone-CP-settings gap is worth closing. Leaving the thread open for your call.

slonka (approved):

test/framework/preflight.go
slonka: isn't piping the failure summary better? so we don't introduce GH logs constructs into the project?

Issue #16973, opened 2026-06-17
Rethink releasing pipeline


Automaat:
### Description

Right now releasing takes really long and is error-prone. We should think about ways of speeding the release process. We potentially can skip running the full test suite on pushing the tag.

Automaat:
> Triage: we probably need MADR for this

PR #17013, opened 2026-06-23
docs(MADR): rethink releasing pipeline

Automaat:
## Motivation

Releasing is slow (~2–3.5h per tag) and error-prone: pushing a `vX.Y.Z` tag re-runs the full e2e matrix that the release branch already passed, and gates all publishing behind it. Addresses #16973.

## Implementation information

MADR only, no code. Proposes Option 1: skip e2e on the tag, gate on a green release-branch SHA, build/publish via a dedicated tag-triggered `release-publish.yaml`, plus a scheduled release rehearsal. Surveys peer projects (only Linkerd re-tests on the tag) and records Options 2 (RC + digest promotion) and 3 (stage/publish split) as follow-ups. Rehearsal feasibility is verified against the actual tooling (SLSA generators, chart-releaser, Kong security actions, `version.sh` preview routing).

## Supporting documentation

- MADR: `docs/madr/decisions/104-rethink-releasing-pipeline.md`
- Issue: #16973

> Changelog: skip

### DIFF
--- a/docs/madr/decisions/104-rethink-releasing-pipeline.md
+++ b/docs/madr/decisions/104-rethink-releasing-pipeline.md
@@ -9,7 +9,7 @@ Technical Story: https://github.com/kumahq/kuma/issues/16973
 Pushing a `vX.Y.Z` tag runs the full `build-test-distribute.yaml`, including the
 entire e2e suite, before any artifact ships. Recent tag runs: v2.13.8 218m,
 v2.14.0 124m, v2.12.12 119m, v2.11.16 102m (~2–3.5h), on top of the manual steps in
-`.claude/rules/release.md`.
+the team-mesh release issues.
 
 On a tag push (`FULL_MATRIX=true`):
 
@@ -18,7 +18,8 @@ On a tag push (`FULL_MATRIX=true`):
   multizone / gatewayapi × versions × amd64/arm64 × CNI × IPv6) at `max-parallel: 5`,
   ~25–35m each. This is the long pole.
 - `build_publish` is `needs: [check, test]`, so it waits for the whole matrix;
-  the build itself is cheap (binaries ~13m, images ~6–8m parallel).
+  the build itself is cheap (binaries ~13m, independently reducible via build
+  caching; images ~6–8m parallel).
 - `provenance` (tags only) plus `distributions`.
 
 The core inefficiency: the tag is HEAD of `release-X.Y`, which already ran the full
@@ -27,11 +28,13 @@ we re-run ~90–150m of e2e and gate all publishing behind it. Secondary problem
 the flakiest CI stage, so a flake can fail an otherwise-good release.
 
 One constraint shapes the options. The version is ldflags-injected from the git tag
-(`version.sh`, `mk/build.mk`), so release-branch artifacts (`2.14.1-<hash>`) cannot be
-byte-promoted to GA (`2.14.0`). Any fix either rebuilds on the tag or changes how the
-version is derived.
+(`version.sh`, `mk/build.mk`): a build on `release-2.14` is stamped as the next-patch
+preview (`X.Y.Z-preview.<hash>`), not the version its GA tag carries (`X.Y.Z`), so
+branch artifacts cannot be byte-promoted to GA. Any fix either rebuilds on the tag or
+changes how the version is derived.
 
-Goals: tag-to-artifacts well under an hour; no weaker supply-chain
+Goals: the tag re-runs no step already validated on the green release-branch SHA
+(build and publish only, not re-test); no weaker supply-chain
 (SBOM/provenance/signing/scan); fewer release failures from flakes or operator error.
 Non-goals: changing cadence/branching/tag format; adopting external infra (Prow/GCB).
 
@@ -58,8 +61,11 @@ on GitHub Actions, the closest templates for us.
 
 ### Option 1 — Skip e2e on the tag; gate on the branch (recommended)
 
-- No e2e (and no e2e smoke) on the tag. Keep `test_unit` and container-structure
-  tests, which are fast and validate the artifact. Drop the ~90–150m matrix.
+- No re-test on the tag: drop unit and e2e. The source is identical to the green
+  branch SHA, so re-running them adds nothing. Verify only what the tag build *changes*
+  versus that SHA — the ldflags version. That is covered by the container-structure
+  tests on the freshly-built image (already in `_build_publish.yaml`) plus an assert
+  that the built binary reports the tag version.
 - Build and publish without the e2e gate, so the build starts right after the gate.
 - Hard-enforce a green tagged SHA. The first job asserts the commit has a passing
   `build-test-distribute` run on its branch (`gh api` check-runs), else it fails.
@@ -102,21 +108,23 @@ Today the release path lives in `build-test-distribute.yaml` via scattered
 - A, more conditionals in the shared workflow. (+) smallest diff. (−) more
   tag-vs-PR branching in an already-conditional workflow; release stays hard to read;
   shared concurrency/permissions.
-- B, a dedicated `release-publish.yaml` (`tags: ["v*"]`) reusing
+- B, a dedicated `release-publish.yaml` triggered on release tags, reusing
   `_build_publish.yaml` and `_provenance.yaml`, never `_test.yaml`; drop the tag trigger
   from `build-test-distribute.yaml`; `release.yaml` keeps changelog/version
-  bookkeeping. (+) linear readable release flow; own concurrency/permissions; no
-  duplicated build logic; matches the Envoy/Cilium/Argo pattern; natural host for the
-  rehearsal and a future RC. (−) new file plus a one-time trigger migration.
+  bookkeeping. The tag filter must match `v?X.Y.Z`, not just `v*`: older branches like
+  `release-2.9` ship *unprefixed* `2.9.x` tags. (+) linear readable release flow; own
+  concurrency/permissions; no duplicated build logic; matches the Envoy/Cilium/Argo
+  pattern; natural host for the rehearsal and a future RC. (−) new file plus a one-time
+  trigger migration.
 
 Picked B.
 
 ### Release rehearsal (scheduled dry-run)
 
 Moving the gate off the tag introduces a risk: the release machinery breaks and we find
-out only at release time. A scheduled rehearsal de-risks it. Checking the
-Makefiles, `version.sh`, `helm.sh`, the workflows, and the tool docs shows this is mostly
-config, not new tooling:
+out only at release time. A scheduled rehearsal de-risks it. The
+Makefiles, `version.sh`, `helm.sh`, the workflows, and the tool docs confirm this is
+mostly config, not new tooling:
 
 - Already runs on every master and `release-X.Y` push: `version.sh` stamps a
   `-preview.v<hash>` version that auto-routes to `notary-internal` and preview
@@ -161,7 +169,9 @@ Option 1 ships now, with Options 2 and 3 as follow-up.
 - Dropping the redundant e2e removes the flakiest release gate.
 - Residual risk: the branch run goes stale against the tag. The exact green-SHA gate,
   RC soak, or a one-off branch matrix before tagging mitigate this.
-- `test_unit` and container-structure tests act as a fast tag tripwire.
+- Container-structure on the freshly-built image plus a binary-version assert are the
+  only tag-specific checks; they verify exactly what the tag build changes (the
+  ldflags), nothing the green branch run already covered.
 - The rehearsal continuously validates build, publish, and sign (plus provenance and
   helm once the gates loosen). Only the real GitHub Release and prod routing stay
   tag-only.
@@ -180,7 +190,8 @@ Adopt Option 1, as a dedicated release workflow plus a scheduled rehearsal:
 - A dedicated `release-publish.yaml` reusing `_build_publish.yaml` and
   `_provenance.yaml`, never `_test.yaml`; drop the tag trigger from
   `build-test-distribute.yaml`.
-- No e2e or smoke on the tag; keep `test_unit` and container-structure tests.
+- No re-test on the tag: drop unit and e2e; keep container-structure on the
+  freshly-built image plus a binary-version assert (the ldflags delta).
 - Hard-enforce a green tagged SHA.
 - A scheduled rehearsal of build, publish, sign, SBOM, scan, and provenance (loosen
   the tag-only `if`s, keep the preview/internal targets); route the helm release to a
@@ -192,7 +203,9 @@ decoupling is undecided.
 
 ## Notes
 
-Resolved: no smoke e2e on the tag; the green-SHA gate is hard-enforced; RC tags deferred.
+Resolved: no re-test on the tag (drop unit and e2e; keep only container-structure plus
+a binary-version assert, the ldflags delta); the green-SHA gate is hard-enforced; RC
+tags deferred.
 
 Open: version decoupling (undecided; rebuild-on-tag is fine for Option 1); the exact
 green-SHA mechanism and rehearsal cadence/scope (impl detail); provisioning a test
--- a/docs/madr/decisions/104-rethink-releasing-pipeline.md
+++ b/docs/madr/decisions/104-rethink-releasing-pipeline.md
@@ -63,19 +63,23 @@ on GitHub Actions, the closest templates for us.
 
 - No re-test on the tag: drop unit and e2e. The source is identical to the green
   branch SHA, so re-running them adds nothing. Verify only what the tag build *changes*
-  versus that SHA — the ldflags version. That is covered by the container-structure
-  tests on the freshly-built image (already in `_build_publish.yaml`) plus an assert
-  that the built binary reports the tag version.
-- Build and publish without the e2e gate, so the build starts right after the gate.
-- Hard-enforce a green tagged SHA. The first job asserts the commit has a passing
-  `build-test-distribute` run on its branch (`gh api` check-runs), else it fails.
+  versus that SHA, the ldflags version, via the container-structure tests on the
+  freshly-built image (already in `_build_publish.yaml`) plus an assert that the built
+  binary reports the tag version.
+- `check`'s lint/`make check`/SBOM is redundant for the same reason; only its metadata
+  outputs (`IMAGES`/`REGISTRY`/`VERSION_NAME`) are needed downstream, so keep those.
+- With e2e skipped, `test` completes immediately and `build_publish` publishes right
+  away instead of waiting ~90m.
+- Hard-enforce a green tagged SHA: the tagged commit must have a passing
+  `build-test-distribute` run (full matrix) on its branch, else the release fails. This
+  reuses the existing run, not new machinery.
 
-After this, green-SHA gate ~1m, `build_publish` ~15m, plus provenance/distributions,
-lands at roughly 20–30m versus the current 100–220m.
+After this, `build_publish` ~15m plus provenance/distributions lands at roughly 20–30m
+versus the current 100–220m.
 
 - (+) Answers the issue; no new infra, staging, or version change; removes the flakiest stage.
-- (−) Loses the exact-tag re-test (mitigated by the hard green-SHA gate, RC soak, and
-  rehearsal); still rebuilds on the tag (~15m, fine).
+- (−) Loses the exact-tag re-test (mitigated by the hard green-SHA gate and, for risky
+  releases, RC soak); still rebuilds on the tag (~15m, fine).
 
 ### Option 2 — RC + promotion by digest (build once, ship the tested bits)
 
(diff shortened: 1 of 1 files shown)


docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Implementation: dedicated release workflow vs. conditionals (orthogonal to Options 1–4)
+
+Today the release path lives in `build-test-distribute.yaml` via scattered
+`ref_type=='tag'` / `FULL_MATRIX` / `ALLOW_PUSH` conditionals.
+
+- A, more conditionals in the shared workflow. (+) smallest diff. (−) more
+  tag-vs-PR branching in an already-conditional workflow; release stays hard to read;
+  shared concurrency/permissions.
+- B, a dedicated `release-publish.yaml` (`tags: ["v*"]`) reusing
```
slonka: nit: `release-2.9` still ships with unprefixed `2.9.x` tags, the new workflow would need to listens to that

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,207 @@
...
+- Drop "wait for full suite on tag" from the downstream runbooks.
+
+## Decision
+
+Adopt Option 1, as a dedicated release workflow plus a scheduled rehearsal:
+
+- A dedicated `release-publish.yaml` reusing `_build_publish.yaml` and
+  `_provenance.yaml`, never `_test.yaml`; drop the tag trigger from
+  `build-test-distribute.yaml`.
+- No e2e or smoke on the tag; keep `test_unit` and container-structure tests.
```
slonka: This says we keep `test_unit` on the tag, but the chosen workflow also says `release-publish.yaml` should never call `_test.yaml`. Today `test_unit` only exists inside `_test.yaml`

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,207 @@
...
+- Status: accepted
+
+Technical Story: https://github.com/kumahq/kuma/issues/16973
+
+## Context and Problem Statement
+
+Pushing a `vX.Y.Z` tag runs the full `build-test-distribute.yaml`, including the
+entire e2e suite, before any artifact ships. Recent tag runs: v2.13.8 218m,
+v2.14.0 124m, v2.12.12 119m, v2.11.16 102m (~2–3.5h), on top of the manual steps in
+`.claude/rules/release.md`.
```
slonka: I think the manual steps are in `team-mesh` release issues - I wasn't even aware of this file

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,207 @@
...
+
+On a tag push (`FULL_MATRIX=true`):
+
+- `check` ~13m (lint, `make check`, SBOM/CVE).
+- `test`: `test_unit` ~23m plus the full e2e matrix, ~17 jobs (k8s / universal /
+  multizone / gatewayapi × versions × amd64/arm64 × CNI × IPv6) at `max-parallel: 5`,
+  ~25–35m each. This is the long pole.
+- `build_publish` is `needs: [check, test]`, so it waits for the whole matrix;
+  the build itself is cheap (binaries ~13m, images ~6–8m parallel).
```
slonka: not a comment on this MADR but 13m for binaries is long, I have a PR that cut's that https://github.com/slonka/kuma/pull/608 - we should use caching

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,207 @@
...
+matrix (release prereq: "CI green on `release-X.Y`"). The tree is byte-identical, yet
+we re-run ~90–150m of e2e and gate all publishing behind it. Secondary problem: e2e is
+the flakiest CI stage, so a flake can fail an otherwise-good release.
+
+One constraint shapes the options. The version is ldflags-injected from the git tag
+(`version.sh`, `mk/build.mk`), so release-branch artifacts (`2.14.1-<hash>`) cannot be
+byte-promoted to GA (`2.14.0`). Any fix either rebuilds on the tag or changes how the
+version is derived.
+
+Goals: tag-to-artifacts well under an hour; no weaker supply-chain
```
slonka: nit: well under an hour is a bit vague, I would go with something not time bound but something related to not re-running unnecessary steps

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,207 @@
...
+| Argo CD | No | release workflow runs zero tests; 7-week RC freeze |
+| **Linkerd** | **Yes** | rebuilds + runs full integration suite on tag |
+
+Only Linkerd re-tests on the tag, which is Kuma's current model. Everyone else gates on
+the branch/RC and makes the tag build-and-publish only. Envoy and Cilium do it natively
+on GitHub Actions, the closest templates for us.
+
+### Option 1 — Skip e2e on the tag; gate on the branch (recommended)
+
+- No e2e (and no e2e smoke) on the tag. Keep `test_unit` and container-structure
```
slonka: TBH I don't see the point of unit / container structure - if the CI passed on a the same commit SHA what additional guarantees does that give us? I think we should just run verification on what changes in the build on SHA vs tag (so I think only the `ldflags`).

slonka (commented):
Done the first round before I have to go, will do the second round when back

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Good catch — verified `2.9.x` ships unprefixed (only `v2.9.13` has a `v`). Updated so the tag filter matches `v?X.Y.Z`, not just `v*`.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Right, real contradiction. Resolved by dropping `test_unit` on the tag entirely (per your other comment) — so `release-publish.yaml` genuinely never calls `_test.yaml`. Container-structure stays because it lives in `_build_publish.yaml`, not `_test.yaml`.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Agreed. Dropped unit on the tag — source is identical to the green SHA, so no extra guarantee. Kept only the verification of what the tag build *changes* vs that SHA: container-structure on the freshly-built image plus an assert that the binary reports the tag version (the ldflags delta). This also fixes the `test_unit`/`_test.yaml` contradiction.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Updated to reference the team-mesh release issues instead of `.claude/rules/release.md`.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Reframed the goal: no time target; now 'the tag re-runs no step already validated on the green release-branch SHA (build and publish only)'.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Noted — added a parenthetical that binary build time is independently reducible via caching (your slonka/kuma#608), orthogonal to this MADR.

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,207 @@
...
+
+The core inefficiency: the tag is HEAD of `release-X.Y`, which already ran the full
+matrix (release prereq: "CI green on `release-X.Y`"). The tree is byte-identical, yet
+we re-run ~90–150m of e2e and gate all publishing behind it. Secondary problem: e2e is
+the flakiest CI stage, so a flake can fail an otherwise-good release.
+
+One constraint shapes the options. The version is ldflags-injected from the git tag
+(`version.sh`, `mk/build.mk`), so release-branch artifacts (`2.14.1-<hash>`) cannot be
+byte-promoted to GA (`2.14.0`). Any fix either rebuilds on the tag or changes how the
+version is derived.
```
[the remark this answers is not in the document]
Automaat: Clarified: a `release-2.14` build is stamped as the next-patch preview (`X.Y.Z-preview.<hash>`) versus the GA `X.Y.Z` the tag carries.

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,207 @@
...
+  duplicated build logic; matches the Envoy/Cilium/Argo pattern; natural host for the
+  rehearsal and a future RC. (−) new file plus a one-time trigger migration.
+
+Picked B.
+
+### Release rehearsal (scheduled dry-run)
+
+Moving the gate off the tag introduces a risk: the release machinery breaks and we find
+out only at release time. A scheduled rehearsal de-risks it. Checking the
+Makefiles, `version.sh`, `helm.sh`, the workflows, and the tool docs shows this is mostly
```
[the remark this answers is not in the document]
Automaat: Reworded to 'The Makefiles, …, and the tool docs confirm …'. (The subject was the gerund phrase 'Checking …', which takes a singular verb, but reworded to avoid the ambiguity.)

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Design
+
+### How peer projects release
+
+Axis = does the full test suite re-run on the tag?
+
+| Project | Re-tests on tag? | Mechanism |
+|:--------|:-----------------|:----------|
+| Kubernetes | No | `krel stage`→`release`; images promoted by **digest** |
```
lahabana, line 49: FWIW that's what KEG does but could work for us.

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,220 @@
...
+| Argo CD | No | release workflow runs zero tests; 7-week RC freeze |
+| **Linkerd** | **Yes** | rebuilds + runs full integration suite on tag |
+
+Only Linkerd re-tests on the tag, which is Kuma's current model. Everyone else gates on
+the branch/RC and makes the tag build-and-publish only. Envoy and Cilium do it natively
+on GitHub Actions, the closest templates for us.
+
+### Option 1 — Skip e2e on the tag; gate on the branch (recommended)
+
+- No re-test on the tag: drop unit and e2e. The source is identical to the green
```
lahabana, line 64: and check?

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Implementation: dedicated release workflow vs. conditionals (orthogonal to Options 1–4)
+
+Today the release path lives in `build-test-distribute.yaml` via scattered
+`ref_type=='tag'` / `FULL_MATRIX` / `ALLOW_PUSH` conditionals.
+
+- A, more conditionals in the shared workflow. (+) smallest diff. (−) more
+  tag-vs-PR branching in an already-conditional workflow; release stays hard to read;
+  shared concurrency/permissions.
+- B, a dedicated `release-publish.yaml` triggered on release tags, reusing
```
lahabana: I think one HUGE thing to consider is that right now we do a release on every signle build. That gives us a super solid predictability that nothing will break.

If now we only publish on tags then this will break on releases and that's a delay on release...

lahabana: Also the conditions we're talking about here already exist they are the same as `SKIP_TEST` right?

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,220 @@
...
+  `upload-sbom-release-assets` defaults to false. Already off-tag ready.
+- Helm release (`cr` 1.8.1): `cr upload` and `cr index` target any owner/repo via flags
+  (`helm.sh` routes `GH_OWNER`/`GH_REPO`). Point them at a throwaway test charts repo
+  (plus a token).
+
+Excluded by design: a real GitHub Release on `kumahq/kuma`, and prod
+registry/notary/Cloudsmith. Caveats: run on `schedule` or `workflow_dispatch`, and
+provision a test charts repo. This follows the nightly-build pattern (Istio, Knative,
+k8s `krel stage`, `cmrel --mock`). It complements the existing `release.yaml` daily
+`--check`, which verifies the *last* release's prod artifacts.
```
lahabana: This is just creating more work. We already do rehearsal on every commit. Now we're just creating complexity by having more machinery

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,220 @@
...
+## Decision
+
+Adopt Option 1, as a dedicated release workflow plus a scheduled rehearsal:
+
+- A dedicated `release-publish.yaml` reusing `_build_publish.yaml` and
+  `_provenance.yaml`, never `_test.yaml`; drop the tag trigger from
+  `build-test-distribute.yaml`.
+- No re-test on the tag: drop unit and e2e; keep container-structure on the
+  freshly-built image plus a binary-version assert (the ldflags delta).
+- Hard-enforce a green tagged SHA.
```
lahabana: That's a step in the build-test-distribute right?

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,220 @@
...
+- The downstream fork mirrors this pipeline; port "skip e2e plus green-SHA gate" directly.
+- For Option 2 later, the downstream needs a matching staging registry and a version
+  change, and syncs by promoting rather than rebuilding.
+- Drop "wait for full suite on tag" from the downstream runbooks.
+
+## Decision
+
+Adopt Option 1, as a dedicated release workflow plus a scheduled rehearsal:
+
+- A dedicated `release-publish.yaml` reusing `_build_publish.yaml` and
```
lahabana: To me this isn't necessary if we just leverage what's already done with `ci/skip-test` : https://github.com/kumahq/kuma/blob/4287359eb8eac5e73cfefc3783efe06764e0458a/.github/workflows/_test.yaml#L21

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Right — recorded the digest/promote model (Option 2) as the KEG/k8s direction and the long-term target.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Good point. `check`'s lint/make-check/SBOM is redundant on the tag too; only its metadata (`IMAGES`/`REGISTRY`/`VERSION_NAME`) is needed downstream. Updated to keep the metadata and skip the rest.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Agreed, and it convinced me to drop the dedicated workflow. Pivoted to staying in `build-test-distribute.yaml` so we keep publishing a preview on every push — exactly that predictability. A tag-only workflow would have lost it.

Automaat: Yes. Tags don't skip today (`_test.yaml` keys off PR labels, empty on a tag push), so it's just adding `github.ref_type == 'tag'` next to the existing `ci/skip-test` condition. Updated the MADR to implement it that way — no new workflow.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Agreed — dropped the scheduled rehearsal entirely. Every push already builds+publishes a preview through the same `_build_publish.yaml`, so that is the rehearsal. Only the tag-gated steps (provenance upload, helm release, GH Release) aren't exercised between releases; noted to revisit via Option 2 if it ever bites.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Right — the green-SHA gate is just 'did the existing build-test-distribute run pass on this SHA', not new machinery. Reworded to make that explicit.

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Agreed, that's the pick now. Implementation reduced to extending the existing `ci/skip-test` condition with `ref_type=='tag'`; the dedicated workflow and rehearsal are dropped.

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Notes
+
+Resolved: no re-test on the tag (drop unit and e2e; keep only container-structure plus a
+binary-version assert, the ldflags delta); implement via the existing `ci/skip-test`
+mechanism in the monolith, not a new workflow; no separate rehearsal (per-push preview
+builds already exercise publish); the green-SHA gate is hard-enforced; RC tags deferred.
+
+Open: version decoupling (undecided; rebuild-on-tag is fine for Option 1); the exact
+green-SHA enforcement mechanism (impl detail).
```
lobkovilya: Can we maybe have more details on how this enforcement is going to be implemented? i.e there can be multiple runs associated with the commit, also probably need to make sure previous runs didn't have `SKIP_E2E_TESTS` set

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,218 @@
...
+### Implementation: reuse the existing skip mechanism vs. a dedicated workflow
+
+Orthogonal to Options 1–4: how to wire the tag pipeline. Today the release path lives in
+`build-test-distribute.yaml`, which runs on every PR, master, and release-branch push,
+so it builds and publishes a preview on every push.
+
+- A, stay in the monolith and reuse the existing `ci/skip-test` mechanism. Tags don't
+  skip today: `_test.yaml` keys the skip off PR labels
+  (`github.event.pull_request.labels`), which are empty on a tag push, so the full matrix
+  runs. Add `github.ref_type == 'tag'` next to that condition and e2e/unit skip on the
```
slonka: "reuse the existing `ci/skip-test` mechanism" is misleading - tags can't have PR labels, so the label itself is never present on a tag. what's actually proposed is extending the same `if:` guard with a separate `ref_type=='tag'` clause. worth rewording.

also under-specified: there are two conditions to touch, `_test.yaml:21` (`test_unit if:`) and `_test.yaml:82` (`SKIP_E2E_TESTS` env that empties the e2e matrix). e2e isn't gated on the label directly, it's gated on the matrix being empty.

docs/madr/decisions/104-rethink-releasing-pipeline.md
slonka: The green-SHA gate is the load-bearing piece and the MADR hand-waves how it interacts with the skip. as written, `ref_type=='tag'` skips unit+e2e unconditionally on any tag push - nothing ties the skip to the SHA being green.

needs spelling out: if the tagged SHA doesn't have a passing `build-test-distribute` run on its branch, the tag run should fail - right? (or at-least re-run the test). Picking the right run is also non-trivial (multiple runs per SHA from reruns/workflow_dispatch, previous runs that had `SKIP_E2E_TESTS=true`, etc) - whatever lookup we use needs to verify `event==push` and that `SKIP_E2E_TESTS` wasn't set.

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -0,0 +1,218 @@
...
+- Hard-enforce a green tagged SHA: the tagged commit must have a passing
+  `build-test-distribute` run (full matrix) on its branch, else the release fails. This
+  reuses the existing run, not new machinery.
+
+After this, `build_publish` ~15m plus provenance/distributions lands at roughly 20–30m
+versus the current 100–220m.
+
+- (+) Answers the issue; no new infra, staging, or version change; removes the flakiest stage.
+- (−) Loses the exact-tag re-test (mitigated by the hard green-SHA gate and, for risky
+  releases, RC soak); still rebuilds on the tag (~15m, fine).
```
slonka: nit: what are risky releases? we don't have RC currently, right?

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Reworded — dropped the "reuse `ci/skip-test`" framing (the label never fires on a tag) and now name both guards explicitly: `test_unit`'s `if:` (`_test.yaml:21`) and the `SKIP_E2E_TESTS` env (`_test.yaml:82`) that empties the e2e matrix. Both get `|| github.ref_type == 'tag'`.

_(reply generated by Claude)_

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Spelled out in a new "Enforcing the green tagged SHA" section. It's **fail-hard**: no trustworthy green run → the tag fails (nothing skips unconditionally). The lookup: `build-test-distribute` runs for the tagged `head_sha`, `event` in {`push`, `workflow_dispatch`} (both force `FULL_MATRIX=true` and carry no PR labels, so e2e always ran — this excludes `pull_request` runs, the only ones that can set `SKIP_E2E_TESTS`), `conclusion == success`, most recent; then a job-level check that `test_unit` + `test_e2e*` are present and successful, to catch a run where the matrix was emptied.

_(reply generated by Claude)_

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Right, no RC today. Reworded: RC soak is not a current fallback — only the hard green-SHA gate is; RC becomes available under Option 2. Dropped the "risky releases" phrasing.

_(reply generated by Claude)_

docs/madr/decisions/104-rethink-releasing-pipeline.md
Automaat: Detailed now under "Enforcing the green tagged SHA". The lookup filters `event` in {`push`, `workflow_dispatch`} and requires the `test_unit` + `test_e2e*` jobs to be present and successful — which handles both "multiple runs per SHA" (take the most recent successful push/dispatch run) and "previous runs had `SKIP_E2E_TESTS`" (those are `pull_request` runs, excluded by the event filter; the job-presence check also catches any emptied matrix).

_(reply generated by Claude)_

docs/madr/decisions/104-rethink-releasing-pipeline.md
```diff
@@ -110,31 +113,70 @@ Orthogonal to Options 1–4: how to wire the tag pipeline. Today the release pat
...
+  `workflow_dispatch` a full run on that SHA) and re-tags.
+
+- Re-run-on-tag (rejected): fall back to running unit and e2e on the tag when no green run
+  is found. Rejected because it re-introduces the flaky ~90–150m e2e this MADR removes,
+  precisely at release time; it keeps a second, rarely-exercised code path that rots; and a
+  tag-run flake can fail the release, the secondary problem the MADR set out to fix. It is
+  also the lone-outlier (Linkerd) model we are moving away from.
+
+We pick fail-hard: the premise is that the branch SHA is already fully tested, so the gate
+must *enforce* that invariant, not silently re-test when it is absent. The failure surfaces
```
slonka: no need to re-stat that we picked this option, right?

slonka (dismissed):

slonka (approved):

bartsmykla (approved):

PR #17017, opened 2026-06-24
perf(matchers): add LRU cache for MatchedPolicies

Automaat:
## Motivation

`MatchedPolicies` is called once per policy type per dataplane on every reconcile cycle. On clusters with many dataplanes and policies this is the hot path. The computation is pure and deterministic given the dataplane identity and the set of matching-relevant mesh resources, so results can be safely memoised.

## Implementation information

- Added `PolicyMatchingCacheAccessor` interface to `pkg/core/plugins` — a thin `GetIfPresent`/`Put` contract.
- Added `MatchedPoliciesConfig.Cache` + `PolicyMatchingHash` fields and a `WithCache()` option; the hash covers all matching-relevant resources in the `MeshContext` but excludes the Dataplane roster (dataplanes change independently of policies).
- Implemented `PolicyMatchingCache` in `pkg/plugins/policies/core/matchers/cache.go` using a bounded LRU (`goburrow/cache`, 10 000 entries) wired to Prometheus counters for hit/miss observability.
- Cache key is a 128-bit FNV hash of `(resourceType, includeShadow, dpp.Hash(), policyMatchingHash)` encoded as base64url — compact and collision-resistant.
- Wired the cache into `DefaultDataplaneWatchdogFactory` via `MeshContextBuilder`, so every reconcile reuses results across policy plugins for the same dataplane × mesh-state combination.
- Added differential tests (cached == uncached) and key-uniqueness tests in `cache_test.go`.

> Changelog: perf(matchers): add LRU cache for MatchedPolicies

Closes https://github.com/kumahq/kuma/issues/16186


pkg/xds/context/mesh_context_builder.go
```diff
@@ -596,6 +600,31 @@ func (m *meshContextBuilder) hash(globalContext *GlobalContext, baseMeshContext
...
+	core_mesh.ZoneEgressInsightType:  {},
+	core_mesh.ServiceInsightType:     {},
+	core_mesh.MeshInsightType:        {},
+}
+
+func (m *meshContextBuilder) computePolicyMatchingHash(globalContext *GlobalContext, baseMeshContext *BaseMeshContext, managedTypes []core_model.ResourceType, resources Resources) []byte {
+	hasher := fnv.New128a()
+	_, _ = hasher.Write(globalContext.hash)
+	_, _ = hasher.Write(baseMeshContext.hash)
+	for _, resType := range managedTypes {
```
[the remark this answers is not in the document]
Automaat: **Applied** — added `policyMatchingGlobalHashDenySet` that excludes `ZoneIngress`, `ZoneEgress`, their insight/overview variants, `MeshInsight`, and `Zone*` system types from the policy-matching hash. `computePolicyMatchingHash` now iterates `globalContext.ResourceMap` with sorted keys and skips denied types instead of writing the full `globalContext.hash` (5e8879345e).

pkg/plugins/policies/core/matchers/cache_test.go
```diff
@@ -0,0 +1,132 @@
...
+			uncached, err := matchers.MatchedPolicies(rType, dpp, resources)
+			Expect(err).ToNot(HaveOccurred())
+
+			// first call: cache miss, computes result
+			c := matchers.NewPolicyMatchingCache(newMetric())
+			const hash = "mesh-hash-stable"
+			first, err := matchers.MatchedPolicies(rType, dpp, resources, core_plugins.WithCache(c, hash))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(first.Type).To(Equal(uncached.Type))
+			Expect(first.DataplanePolicies).To(HaveLen(len(uncached.DataplanePolicies)))
```
slonka: this test is named `cached result is identical to uncached result` but only compares `Type` and `len(DataplanePolicies)`. would silently pass if the cache dropped `FromRules`, `ToRules`, `GatewayRules`, `InboundPolicies`, `OutboundPolicies`, etc.

can we use `Expect(first).To(BeComparableTo(uncached))` (or the gomega equivalent) so a regression on any field gets caught?

pkg/plugins/policies/core/matchers/cache.go
```diff
@@ -0,0 +1,57 @@
...
+	_, _ = h.Write([]byte(rType))
+	if includeShadow {
+		_, _ = h.Write([]byte{1})
+	} else {
+		_, _ = h.Write([]byte{0})
+	}
+	_, _ = h.Write(dpp.Hash())
+	_, _ = h.Write([]byte(policyMatchingHash))
+	return base64.RawURLEncoding.EncodeToString(h.Sum(nil))
+}
```
slonka, line 64: nit: any reason for base64? `string(h.Sum(nil))` works fine as a map key and is shorter (16B vs 22B) and skips the encode step.

pkg/xds/sync/components.go
```diff
@@ -62,6 +63,7 @@ func DefaultDataplaneWatchdogFactory(
 		config,
 		apiVersion,
 	)
+	dataplaneProxyBuilder.policyMatchingCache = matchers.NewPolicyMatchingCache(xdsMetrics.PolicyMatchingCache)
```
slonka: nit: it's called a `Builder` but we're poking a private field from outside. how about a chainable `dataplaneProxyBuilder.WithPolicyMatchingCache(xdsMetrics.PolicyMatchingCache)` on `*DataplaneProxyBuilder`?

pkg/plugins/policies/core/matchers/cache.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	core_plugins "github.com/kumahq/kuma/v3/pkg/core/plugins"
+	core_mesh "github.com/kumahq/kuma/v3/pkg/core/resources/apis/mesh"
+	core_xds "github.com/kumahq/kuma/v3/pkg/core/xds"
+	util_cache "github.com/kumahq/kuma/v3/pkg/util/cache"
+)
+
+// bounds worst-case memory to O(policyTypes × dataplanes)
+const defaultPolicyMatchingCacheSize = 10_000
```
slonka: 10k hardcoded - for a large mesh (say 5k DPs × ~16 policy plugins = 80k entries) this will thrash and the hit-rate drops once `PolicyMatchingHash` rotates. Should we make it configurable (env / kuma-cp config), or maybe size it dynamically?

pkg/xds/context/mesh_context_builder.go
```diff
@@ -596,6 +600,50 @@ func (m *meshContextBuilder) hash(globalContext *GlobalContext, baseMeshContext
 	return hasher.Sum(nil)
 }
 
+// policyMatchingHashDenySet: mesh-scoped types irrelevant for policy matching (roster, insights).
+// Omitting a member is safe (spurious cache misses); including a wrong one causes stale xDS.
+var policyMatchingHashDenySet = map[core_model.ResourceType]struct{}{
```
slonka: broader Q on these two deny sets: could this be derived from the resource descriptor instead of a hand-maintained list? two categories here:

1. `*Insight` / `*Overview` - pure observability, never affect matching
2. proxy-roster types (`Dataplane`, `ZoneIngress`, `ZoneEgress`, `system.Zone`) - other proxies don't affect which policies match *this* dpp (cache key already includes `dpp.Hash()`)

Maybe it could be a flag on `model.ResourceTypeDescriptor` (e.g. `AffectsPolicyMatching bool`). Resources opt out at definition time, then adding a new `FooInsight` later can't silently churn the cache, and the categorically-wrong entries above (global-scoped types in the mesh-scoped set) become impossible by construction. WDYT?

pkg/plugins/policies/core/matchers/cache_test.go
Automaat: **Applied** — replaced the partial Type/length assertions with full `BeComparableTo` comparisons for cached vs uncached results (5649944b50).

_Generated by Sybra harness_

pkg/plugins/policies/core/matchers/cache.go
Automaat: **Applied** — switched the cache key to use the raw 128-bit hash bytes as the string map key and removed the base64 encode step (5649944b50).

_Generated by Sybra harness_

pkg/xds/sync/components.go
Automaat: **Applied** — added chainable `WithPolicyMatchingCache` on `DataplaneProxyBuilder` and wired the default builder through it instead of assigning the private field externally (5649944b50).

_Generated by Sybra harness_

pkg/plugins/policies/core/matchers/cache.go
Automaat: **Applied** — added `xdsServer.policyMatchingCacheSize` / `KUMA_XDS_SERVER_POLICY_MATCHING_CACHE_SIZE`, defaulted it to 100000, and wired it into the shared policy matching cache size (5649944b50).

_Generated by Sybra harness_

pkg/xds/context/mesh_context_builder.go
Automaat: **Skipped (questionable)** — I cleaned up the concrete scope issue by removing global-scoped types from the mesh-scoped deny set and adding the mesh overview observability types (5649944b50). Moving this to a `ResourceTypeDescriptor` opt-out flag would be a broader resource-model/codegen contract change, so I left that out of this review fix and left the thread open for your call.

_Generated by Sybra harness_

tools/resource-gen/genutils/util.go
```diff
@@ -89,6 +89,7 @@ type ResourceInfo struct {
 	AdditionalPrinterColumns []string
 	HasInsights              bool
 	IsProxy                  bool
+	PolicyMatchingExempt     bool
```
slonka: let's call it `AffectsPolicyMatching` so it's the same as in the resource descriptor

Automaat: **Applied** — renamed `ResourceInfo.PolicyMatchingExempt` to `AffectsPolicyMatching` (inverting the stored bool) so it matches `model.ResourceTypeDescriptor.AffectsPolicyMatching`; the generator template now emits `AffectsPolicyMatching: {{.AffectsPolicyMatching}}` directly instead of `{{not .PolicyMatchingExempt}}` (0b8e694e25).

_Generated by Sybra harness_

pkg/config/xds/config.go
```diff
@@ -27,6 +27,8 @@ type XdsServerConfig struct {
 	DataplaneDeregistrationDelay config_types.Duration `json:"dataplaneDeregistrationDelay" envconfig:"kuma_xds_dataplane_deregistration_delay"`
 	// Backoff that is executed when Control Plane is sending the response that was previously rejected by Dataplane
 	NACKBackoff config_types.Duration `json:"nackBackoff" envconfig:"kuma_xds_server_nack_backoff"`
+	// Maximum number of policy matching results kept in the shared Dataplane xDS reconciliation cache.
+	PolicyMatchingCacheSize int `json:"policyMatchingCacheSize" envconfig:"kuma_xds_server_policy_matching_cache_size"`
```
lobkovilya, line 32: should we have `0` to disable the cache?

slonka (dismissed):

pkg/plugins/policies/core/matchers/dataplane.go
```diff
@@ -47,6 +47,14 @@ func MatchedPolicies(
 ) (core_xds.TypedMatchingPolicies, error) {
 	mpOpts := plugins.NewMatchedPoliciesConfig(opts...)
 
+	var cacheKey string
+	if mpOpts.Cache != nil {
+		cacheKey = BuildCacheKey(string(rType), mpOpts.IncludeShadow, dpp, mpOpts.PolicyMatchingHash)
```
lobkovilya: Need to accept `MatchedPoliciesConfig` not just `mpOpts.IncludeShadow` otherwise it's easy to forget in the future when new options added

pkg/config/app/kuma-cp/kuma-cp.defaults.yaml
```diff
@@ -139,6 +139,8 @@ xdsServer:
   dataplaneConfigurationRefreshInterval: 1s # ENV: KUMA_XDS_SERVER_DATAPLANE_CONFIGURATION_REFRESH_INTERVAL
   # Interval for flushing status of Dataplanes connected to the Control Plane
   dataplaneStatusFlushInterval: 10s # ENV: KUMA_XDS_SERVER_DATAPLANE_STATUS_FLUSH_INTERVAL
+  # Maximum number of policy matching results kept in the shared Dataplane xDS reconciliation cache.
+  policyMatchingCacheSize: 100000 # ENV: KUMA_XDS_SERVER_POLICY_MATCHING_CACHE_SIZE
```
lobkovilya: That might be a lot given that `TypedMatchingPolicies` can be big. Did you check how the memory usage changes?

Also the PR description says `10k` but here is's actually `100k`

pkg/xds/metrics/metrics.go
```diff
@@ -42,13 +43,18 @@ func NewMetrics(metrics core_metrics.Metrics) (*Metrics, error) {
 		Name: "xds_dataplane_config_regenerated_total",
 		Help: "Counter of successful Dataplane xDS reconcile attempts triggered by mesh-context hash changes.",
 	}, []string{"mesh"})
+	policyMatchingCache := util_cache.NewMetric(
+		"policy_matching_cache",
+		"Cache hit/miss counts for MatchedPolicies memoisation on the dataplane watchdog path.",
```
lobkovilya, line 48: I see that we pass the metric down to the `goburrow/cache` library but do we now what are the labels? Does it make sense to just count the metric on our side in a thin cache wrapper that you've created?

pkg/plugins/policies/core/matchers/dataplane.go
Automaat: **Applied** — changed `BuildCacheKey` to accept the full `MatchedPoliciesConfig` and derive key inputs from it, so future options are harder to miss (5d750cbb88).

pkg/config/app/kuma-cp/kuma-cp.defaults.yaml
Automaat: **Applied** — reduced the default and generated docs from 100k to 10k to match the PR description and lower the default memory footprint (5d750cbb88).

pkg/config/xds/config.go
Automaat: **Applied** — allowed `policyMatchingCacheSize: 0` by changing validation to reject only negative values and skipping shared cache wiring when the size is 0 (5d750cbb88).

pkg/plugins/policies/core/matchers/cache.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	core_plugins "github.com/kumahq/kuma/v3/pkg/core/plugins"
+	core_mesh "github.com/kumahq/kuma/v3/pkg/core/resources/apis/mesh"
+	core_xds "github.com/kumahq/kuma/v3/pkg/core/xds"
+)
+
+var _ core_plugins.PolicyMatchingCacheAccessor = &PolicyMatchingCache{}
+
+// PolicyMatchingCache is a bounded LRU cache for TypedMatchingPolicies; safe for concurrent use.
```
slonka, line 17: We could add a note about this being read-only if this comment by claude is right:

> cached `TypedMatchingPolicies` value shares maps and slices (`DataplanePolicies`, `FromRules.Rules`, `ToRules`, `InboundPolicies`, `OutboundPolicies`, `GatewayRules`, ...) by reference across every hit. current callers iterate read-only, so it works in practice, but same invariant as `pkg/core/resources/manager/cache.go` ("cached resources are read-only" per CLAUDE.md). one line on this type saves a future plugin author from accidentally sorting/appending in-place and corrupting subsequent hits.

pkg/plugins/policies/core/matchers/cache.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func NewPolicyMatchingCache(metric *prometheus.CounterVec, maxSize int) *PolicyMatchingCache {
+	c := cache.New(
+		cache.WithMaximumSize(maxSize),
+	)
+	return &PolicyMatchingCache{c: c, metric: metric}
+}
+
+func (p *PolicyMatchingCache) GetIfPresent(key string) (core_xds.TypedMatchingPolicies, bool) {
```
slonka, line 39: nit: `WithLabelValues("hit"|"miss")` does a map lookup on every Get. Cache is hit multiple times per reconcile - could stash `hitCounter, missCounter prometheus.Counter` on the struct at construction and just `Inc()` here.

pkg/xds/context/mesh_context_builder.go
```diff
@@ -142,6 +142,9 @@ func (m *meshContextBuilder) BuildIfChanged(ctx context.Context, meshName string
 		return latestMeshCtx, nil
 	}
 
+	// managedTypes already sorted by m.hash above
+	policyMatchingHash := base64.StdEncoding.EncodeToString(m.computePolicyMatchingHash(globalContext, baseMeshContext, managedTypes, resources))
```
slonka: when `PolicyMatchingCacheSize=0` (cache disabled) we still compute `policyMatchingHash` on every `BuildIfChanged` and it's never used. worth skipping the work in that case, or is the plan to eventually keep this hash for something else?

slonka (dismissed):
Most of the round-1 feedback landed - descriptor-driven exempt flag came out really nicely. Few small remaining notes inline.

pkg/plugins/policies/core/matchers/cache.go
Automaat: Applied in 604611bd — added a read-only note on the `PolicyMatchingCache` type mirroring the `pkg/core/resources/manager/cache.go` invariant.

pkg/plugins/policies/core/matchers/cache.go
Automaat: Applied in 604611bd — resolved the `hit`/`miss` counters once in the constructor (`hitCounter`/`missCounter prometheus.Counter`) so `GetIfPresent` just calls `Inc()`, no per-Get `WithLabelValues` lookup.

pkg/xds/context/mesh_context_builder.go
Automaat: Applied in 604611bd — gated the `policyMatchingHash` computation behind a new `WithPolicyMatchingHash()` builder option, enabled from bootstrap only when `PolicyMatchingCacheSize > 0`. When the cache is disabled the hash is no longer computed.

pkg/xds/context/mesh_context_builder.go
```diff
@@ -596,6 +619,33 @@ func (m *meshContextBuilder) hash(globalContext *GlobalContext, baseMeshContext
...
+}
+
+func (m *meshContextBuilder) computePolicyMatchingHash(globalContext *GlobalContext, baseMeshContext *BaseMeshContext, managedTypes []core_model.ResourceType, resources Resources) []byte {
+	hasher := fnv.New128a()
+	for _, resType := range maps.SortedKeys(globalContext.ResourceMap) {
+		if affectsPolicyMatching(resType) {
+			_, _ = hasher.Write(core_model.ResourceListHash(globalContext.ResourceMap[resType]))
+		}
+	}
+	_, _ = hasher.Write(baseMeshContext.hash)
```
lukidzi: **Blocking (architecture):** `policyMatchingHash` uses `baseMeshContext.hash` directly, which includes ALL resources in that context — not just matching-relevant ones. This means Config resources (VIP config map) can trigger spurious cache invalidations when they change but don't affect policy matching.

More critically, the asymmetry between how `globalContext` is filtered (via `affectsPolicyMatching`) vs `baseMeshContext` (unfiltered) creates a maintenance hazard: new resource types added to BaseMeshContext won't respect `AffectsPolicyMatching=false` unless this is refactored.

Consider filtering `baseMeshContext.ResourceMap` the same way `globalContext` is filtered.

pkg/xds/context/mesh_context_builder.go
```diff
@@ -596,6 +619,33 @@ func (m *meshContextBuilder) hash(globalContext *GlobalContext, baseMeshContext
...
+// treated as relevant (safe default: spurious cache misses, never stale xDS).
+func affectsPolicyMatching(resType core_model.ResourceType) bool {
+	descriptor, err := registry.Global().DescriptorFor(resType)
+	if err != nil {
+		return true
+	}
+	return descriptor.AffectsPolicyMatching
+}
+
+func (m *meshContextBuilder) computePolicyMatchingHash(globalContext *GlobalContext, baseMeshContext *BaseMeshContext, managedTypes []core_model.ResourceType, resources Resources) []byte {
```
lukidzi, line 633: **Blocking (correctness):** `computePolicyMatchingHash` does not include `CrossMeshResources` in its hash, but the main `MeshContext.hash()` does (lines 615-618). If cross-mesh gateways or dataplanes affect policy matching, stale cached results could be served when CrossMeshResources change but other policy-relevant resources do not.

If cross-mesh traffic doesn't affect policy matching for a given DPP, document this explicitly. If it does, add it to the hash.

pkg/plugins/policies/core/matchers/cache.go
```diff
@@ -0,0 +1,62 @@
...
+// Cached values share maps and slices by reference across every hit — callers must treat the
+// returned TypedMatchingPolicies as read-only (no in-place sort/append), same invariant as
+// pkg/core/resources/manager/cache.go.
+type PolicyMatchingCache struct {
+	c           cache.Cache
+	hitCounter  prometheus.Counter
+	missCounter prometheus.Counter
+}
+
+func NewPolicyMatchingCache(metric *prometheus.CounterVec, maxSize int) *PolicyMatchingCache {
```
lukidzi, line 27: **Suggestion:** Consider adding a defensive TTL (e.g., `WithExpireAfterAccess(5 * time.Minute)`) as a safety net. Without TTL, entries can become stale indefinitely if there's a bug in hash computation. The goburrow/cache library supports this.

pkg/plugins/policies/core/matchers/cache_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+			Expect(testutil.ToFloat64(metric.WithLabelValues("hit"))).To(Equal(1.0))
+			Expect(testutil.ToFloat64(metric.WithLabelValues("miss"))).To(Equal(1.0))
+		})
+	})
+
+	Describe("MatchedPolicies with cache", func() {
+		testDir := filepath.Join("testdata", "matchedpolicies", "fromrules")
+
+		It("cached result is identical to uncached result", func() {
```
lukidzi, line 115: **Suggestion:** Test `cached result is identical to uncached result` uses a single test case. Consider adding coverage for: different policy types, gateway dataplanes, dataplanes with multiple inbounds, shadow policies enabled/disabled.

pkg/plugins/policies/core/matchers/cache_test.go
```diff
@@ -0,0 +1,146 @@
...
+			Expect(k1).ToNot(Equal(k2))
+		})
+
+		It("returns distinct keys for different policyMatchingHash", func() {
+			dpp := readDPP(filepath.Join("testdata", "matchedpolicies", "fromrules", "01.dataplane.yaml"))
+			k1 := matchers.BuildCacheKey("TypeA", core_plugins.NewMatchedPoliciesConfig(core_plugins.WithCache(nil, "hash1")), dpp)
+			k2 := matchers.BuildCacheKey("TypeA", core_plugins.NewMatchedPoliciesConfig(core_plugins.WithCache(nil, "hash2")), dpp)
+			Expect(k1).ToNot(Equal(k2))
+		})
+
```
lukidzi, line 42: **Suggestion:** Missing test: distinct keys for different dataplane specs (same name but different version). This would validate that dataplane spec changes (inbound tags, etc.) correctly invalidate the cache via the version in `HashMeta`.

pkg/config/xds/config.go
```diff
@@ -45,6 +51,7 @@ func DefaultXdsServerConfig() *XdsServerConfig {
 		DataplaneStatusFlushInterval:          config_types.Duration{Duration: 10 * time.Second},
```
lukidzi, line 51: **Info:** Default `PolicyMatchingCacheSize` is 10,000 entries. Each `TypedMatchingPolicies` can be large (contains ResourceLists, rule maps, etc.). Consider documenting expected memory usage per entry or adding a memory-based limit option for large deployments.

pkg/xds/context/mesh_context_builder.go
Automaat: Fixed in d47c209 — `computePolicyMatchingHash` now iterates `baseMeshContext.ResourceMap` with the same `affectsPolicyMatching` filter as `globalContext`, instead of hashing `baseMeshContext.hash`. New resource types added to BaseMeshContext now respect `AffectsPolicyMatching=false`.

pkg/xds/context/mesh_context_builder.go
Automaat: Fixed in 55c3e78 — `CrossMeshResources` are now folded into `computePolicyMatchingHash`, filtered by `affectsPolicyMatching` (mirroring the main `hash()`). In practice only cross-mesh `MeshGateway` contributes (`AffectsPolicyMatching=true`); cross-mesh `Dataplane` stays exempt (`=false`). This makes the hash robust regardless of whether a matcher reads `CrossMeshResources` now or in the future.

pkg/plugins/policies/core/matchers/cache.go
Automaat: Applied in 55c3e78 — added `cache.WithExpireAfterAccess(5 * time.Minute)` to `NewPolicyMatchingCache` as a safety net.

pkg/plugins/policies/core/matchers/cache_test.go
Automaat: Expanded in d47c209 (gateway dataplane) and 55c3e78 (multi-inbound dataplane via `dataplanepolicies/01`, and shadow-enabled matching via `IncludeShadow()`). Coverage now spans multiple policy types (MeshTrafficPermission, MeshAccessLog), gateway + multi-inbound dataplanes, and shadow on/off.

pkg/plugins/policies/core/matchers/cache_test.go
Automaat: Added in d47c209 — `BuildCacheKey` test "returns distinct keys for the same dataplane name with different versions" asserts the version in `HashMeta` invalidates the key.

pkg/config/xds/config.go
Automaat: The klaudiush lint hook in my workflow blocks editing doc comments, so I could not extend the config comment in-code. For the record: each entry holds a `TypedMatchingPolicies` (rule maps + resource lists), so the 10k default trades memory for hit-rate and should be lowered on memory-constrained deployments. Happy to fold this into the field doc in a follow-up if you want it in-tree.

pkg/xds/sync/dataplane_proxy_builder.go
```diff
@@ -233,6 +240,9 @@ func (p *DataplaneProxyBuilder) matchPolicies(meshContext xds_context.MeshContex
 	if p.IncludeShadow {
 		opts = append(opts, core_plugins.IncludeShadow())
 	}
+	if p.policyMatchingCache != nil {
+		opts = append(opts, core_plugins.WithCache(p.policyMatchingCache, meshContext.PolicyMatchingHash))
```
lukidzi, line 244: Suggestion: When `policyMatchingCache` is non-nil but `meshContext.PolicyMatchingHash` is empty (`WithPolicyMatchingHash()` not called), the cache key has an empty hash component. Log a warning or skip caching when `PolicyMatchingHash` is empty - otherwise different mesh states may hit the same cache entry.

api/mesh/v1alpha1/dataplane.proto
```diff
@@ -22,6 +22,7 @@ message Dataplane {
   option (kuma.mesh.resource).scope_namespace = true;
   option (kuma.mesh.resource).has_insights = true;
   option (kuma.mesh.resource).is_proxy = true;
+  option (kuma.mesh.resource).policy_matching_exempt = true;
```
lukidzi, line 25: Isn't it relevant for DPP as we care about labels change in this case?

pkg/core/resources/apis/system/zz_generated.resources.go
```diff
@@ -108,23 +108,24 @@ func (l *ConfigResourceList) SetPagination(p model.Pagination) {
...
+	WsPath:                "",
+	KumactlArg:            "",
+	KumactlListArg:        "",
+	AllowToInspect:        false,
+	IsPolicy:              false,
+	SingularDisplayName:   "Config",
+	PluralDisplayName:     "Configs",
+	IsExperimental:        false,
+	IsProxy:               false,
+	AffectsPolicyMatching: true,
```
lukidzi, line 128: Not fully sure if this affects it?

lukidzi (approved):
I think we can try it, lets see how it works and be conscious about any flake

PR #17019, opened 2026-06-24
feat(db): add tolerateNewerDBVersions flag

slonka:
## Motivation

For upgrades, older CP currently refuses to boot once a newer CP has advanced the schema: `migrate up` errors on "DB newer than binary" and `IsDbMigrated` requires strict `dbVer == fileVer`. This PR adds an opt-in env var to relax both gates so the newer CP can own migrations while the older CP keeps running.

## Implementation information

Adds `SkipMigrationCheck` (env `KUMA_STORE_POSTGRES_SKIP_MIGRATION_CHECK`, default off) that turns the "DB newer than binary" branch of `MigrateDb` into `AlreadyMigrated` and switches `IsDbMigrated` to `dbVer >= fileVer`. The bypass only fires when `dbVer > appVer`, so unrelated "file does not exist" errors still fail loudly.

## Supporting documentation

closes https://github.com/Kong/kong-mesh/issues/9977

### DIFF
--- a/docs/generated/raw/kuma-cp.yaml
+++ b/docs/generated/raw/kuma-cp.yaml
@@ -55,6 +55,11 @@ store:
       keyPath: # ENV: KUMA_STORE_POSTGRES_TLS_KEY_PATH
       # Path to the root certificate. Used in verifyCa and verifyFull modes.
       caPath: # ENV: KUMA_STORE_POSTGRES_TLS_ROOT_CERT_PATH
+    # SkipMigrationCheck makes `migrate up` a no-op when the DB is at a newer
+    # version than the binary, and relaxes the boot-time migration gate from
+    # strict equality (dbVer == fileVer) to dbVer >= fileVer. Intended for
+    # blue/green CP upgrades where a newer CP owns schema migrations.
+    skipMigrationCheck: false # ENV: KUMA_STORE_POSTGRES_SKIP_MIGRATION_CHECK
     # ReadReplica is a setting for a DB replica used only for read queries
     readReplica:
       # Host of the Postgres DB read replica. If not set, read replica is not used.
--- a/docs/generated/raw/kuma-cp.yaml
+++ b/docs/generated/raw/kuma-cp.yaml
@@ -55,11 +55,11 @@ store:
       keyPath: # ENV: KUMA_STORE_POSTGRES_TLS_KEY_PATH
       # Path to the root certificate. Used in verifyCa and verifyFull modes.
       caPath: # ENV: KUMA_STORE_POSTGRES_TLS_ROOT_CERT_PATH
-    # SkipMigrationCheck makes `migrate up` a no-op when the DB is at a newer
-    # version than the binary, and relaxes the boot-time migration gate from
-    # strict equality (dbVer == fileVer) to dbVer >= fileVer. Intended for
-    # blue/green CP upgrades where a newer CP owns schema migrations.
-    skipMigrationCheck: false # ENV: KUMA_STORE_POSTGRES_SKIP_MIGRATION_CHECK
+    # TolerateNewerDBVersions makes `migrate up` a no-op when the DB is at a
+    # newer version than the binary, and relaxes the boot-time migration gate
+    # from strict equality (dbVer == fileVer) to dbVer >= fileVer. Intended
+    # for blue/green CP upgrades where a newer CP owns schema migrations.
+    tolerateNewerDBVersions: false # ENV: KUMA_STORE_POSTGRES_TOLERATE_NEWER_DB_VERSIONS
     # ReadReplica is a setting for a DB replica used only for read queries
     readReplica:
       # Host of the Postgres DB read replica. If not set, read replica is not used.
--- a/pkg/config/plugins/resources/postgres/config.go
+++ b/pkg/config/plugins/resources/postgres/config.go
@@ -76,6 +76,14 @@ type PostgresStoreConfig struct {
 	// from strict equality (dbVer == fileVer) to dbVer >= fileVer. Intended
 	// for blue/green CP upgrades where a newer CP owns schema migrations and
 	// an older CP shares the same DB.
+	//
+	// Safety precondition: every migration in the skew range MUST be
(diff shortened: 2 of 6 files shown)


pkg/plugins/resources/postgres/migrate.go
```diff
@@ -39,6 +42,10 @@ func MigrateDb(cfg postgres_cfg.PostgresStoreConfig) (core_plugins.DbVersion, er
 			if err != nil {
 				return 0, err
 			}
+			if cfg.SkipMigrationCheck {
+				migrateLog.Info("DB is at a newer migration version than the binary; skipping migration because SkipMigrationCheck is enabled", "dbVersion", dbVer, "binaryVersion", appVer)
+				return dbVer, core_plugins.AlreadyMigrated
+			}
```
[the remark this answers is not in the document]
slonka: Good catch - tightened the bypass to dbVer > appVer in 78f73e1 so the SkipMigrationCheck branch only fires for the intended "DB newer than binary" case; other "file does not exist" errors keep failing loudly.

pkg/plugins/resources/postgres/migrate_test.go
```diff
@@ -52,6 +52,33 @@ var _ = Describe("Migrate", func() {
...
 
+	It("should not error when DB has a newer migration version and SkipMigrationCheck is enabled", func() {
+		// given a DB migrated past the binary's newest migration
+		cfg, err := c.Config()
+		Expect(err).ToNot(HaveOccurred())
+		_, err = postgres.MigrateDb(cfg)
+		Expect(err).ToNot(HaveOccurred())
+
+		sql, err := common_postgres.ConnectToDb(cfg)
+		Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
slonka: Done - added a defer sql.Close() in 78f73e1.

docs/generated/raw/kuma-cp.yaml
```diff
@@ -55,6 +55,11 @@ store:
       keyPath: # ENV: KUMA_STORE_POSTGRES_TLS_KEY_PATH
       # Path to the root certificate. Used in verifyCa and verifyFull modes.
       caPath: # ENV: KUMA_STORE_POSTGRES_TLS_ROOT_CERT_PATH
+    # SkipMigrationCheck makes `migrate up` a no-op when the DB is at a newer
+    # version than the binary, and relaxes the boot-time migration gate from
+    # strict equality (dbVer == fileVer) to dbVer >= fileVer. Intended for
+    # blue/green CP upgrades where a newer CP owns schema migrations.
+    skipMigrationCheck: false # ENV: KUMA_STORE_POSTGRES_SKIP_MIGRATION_CHECK
```
lobkovilya: nit: consider naming the flag something like `tolerateNewerDBVersions` to reflect the fact we don't skip the check completely but we're fine when DB is on newer version

pkg/plugins/resources/postgres/migrate.go
```diff
@@ -88,6 +95,9 @@ func IsDbMigrated(cfg postgres_cfg.PostgresStoreConfig) (bool, error) {
 		return false, err
 	}
 
+	if cfg.SkipMigrationCheck {
+		return dbVer >= fileVer, nil
```
Automaat, line 102: **issue:** `IsDbMigrated` discards the `dirty` flag, so with `SkipMigrationCheck` on, an older CP boots cleanly onto a half-applied schema. golang-migrate writes the version, runs the migration body, then clears `dirty`, all in separate commits. A newer CP killed mid-migration (OOM/SIGKILL) leaves `version=newest, dirty=true`. The older CP, flag on, calls `IsDbMigrated`, which reads `dbVer, _, err := m.Version()` and drops `dirty`, so `dbVer >= fileVer` returns true and the CP serves reads and writes against a schema that is mid-migration. The strict `==` path dropped `dirty` too, so the bug predates the PR, but the flag widens it from a single equality point to the whole `dbVer >= fileVer` range and ships it as a supported mode. The boot path calls `IsDbMigrated`, not `MigrateDb`, so golang-migrate's own `ErrDirty` guard never fires here. When `SkipMigrationCheck` is set, read `dirty` and refuse to boot when it is true.

pkg/config/plugins/resources/postgres/config.go
```diff
@@ -71,6 +71,12 @@ type PostgresStoreConfig struct {
 	MaxListQueryElements uint32 `json:"maxListQueryElements" envconfig:"kuma_store_postgres_max_list_query_elements"`
 	// ReadReplica is a setting for a DB replica used only for read queries
 	ReadReplica ReadReplica `json:"readReplica"`
+	// SkipMigrationCheck makes `migrate up` a no-op when the DB is at a newer
```
Automaat: **issue:** The flag's safety rests on a precondition documented nowhere, one that Kuma's own history breaks. Letting an older CP run against a newer schema is safe only if every intervening migration is additive: no drop, rename, type change, or `NOT NULL` without a default. The field comment describes the mechanism (`dbVer >= fileVer`), not the precondition, and `UPGRADE.md` stays untouched. Yet `1604501001_adjust_global_resources.up.sql` drops a constraint and rewrites rows, and `1677096751_update_resources_name_limit.up.sql` widens `name` to `varchar(253)`, after which a newer CP can persist a name longer than an older CP's queries expect. Migrations are up-only (no `.down.sql`), so the flag lets the old CP boot but does nothing to make it correct after a rollback. Document the precondition on the field and add an `UPGRADE.md` entry (Impact / Action-required, matching the existing `2.13.7` TLS entry) that states the requirement: additive-only migrations across the version skew, operator verification, and no rollback safety.

docs/generated/raw/kuma-cp.yaml
slonka: Renamed to TolerateNewerDBVersions (env KUMA_STORE_POSTGRES_TOLERATE_NEWER_DB_VERSIONS) in 489ba1213f.

Automaat (approved):

lobkovilya (approved):

PR #17036, opened 2026-06-26
test(e2e): improve failure diagnostics

Automaat:
## Motivation

Some e2e flakes are hard to reproduce locally because failure artifacts lose exact request context. Recent examples collapsed to empty-response JSON errors without source pod, destination, curl context, partial concurrent results, or a compact reproduction manifest.

## Changes

- add repro.json to failed e2e reports with Ginkgo seed, focus and label filters, parallel settings, selected matrix env, framework config, runtime info, and local tool versions
- attach per-request client diagnostics on curl execution, JSON unmarshal, and empty-instance failures
- attach concurrent request summaries with partial responses and histograms when a batch fails early
- dump namespace events in Kubernetes failure reports
- add a bounded TrafficRecorder helper and use it in MeshIdentity migration to label background probe failures

## Verification

GOCACHE=/private/tmp/kuma-go-build-cache go test ./test/framework/... ./test/e2e_env/multizone/meshidentity

GOCACHE=/private/tmp/kuma-go-build-cache go test ./test/e2e_env/multizone/meshproxy ./test/e2e_env/multizone/meshhttproute ./test/e2e_env/multizone/localityawarelb


test/framework/client/collect.go
```diff
@@ -323,23 +362,86 @@ func CollectEchoResponse(
...
+		Headers:     result.opts.Headers,
+		Flags:       result.opts.Flags,
+		Command:     result.command,
+		Stdout:      truncateDiagnosticString(result.stdout),
+		Stderr:      truncateDiagnosticString(result.stderr),
+	}
+	if err != nil {
+		diagnostic.Error = err.Error()
+	}
+
```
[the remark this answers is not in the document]
Automaat: **Applied** - Redacted sensitive request headers and curl header arguments before collect diagnostics are written to artifacts (bdbf4210d3).

test/framework/client/collect.go
```diff
@@ -593,13 +695,94 @@ func callConcurrently(destination string, call func() (any, error), fn ...Collec
...
+		MaxConcurrentRequests: opts.maxConcurrentRequests,
+		CompletedResponses:    len(responses),
+		FailedIndex:           failed.idx,
+		Error:                 failed.err.Error(),
+		PartialResponses:      responses,
+		Headers:               opts.Headers,
+		Flags:                 opts.Flags,
+		InstanceHistogram:     instanceHistogram(responses),
+		ResponseCodeHistogram: responseCodeHistogram(responses),
+	}
```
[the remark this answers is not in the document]
Automaat: **Applied** - Redacted sensitive headers before concurrent-request diagnostics are written to artifacts (bdbf4210d3).

test/framework/client/collect.go
```diff
@@ -593,13 +780,94 @@ func callConcurrently(destination string, call func() (any, error), fn ...Collec
...
+		Timestamp:             time.Now().UTC(),
+		Destination:           destination,
+		URL:                   opts.URL,
+		Method:                opts.Method,
+		NumberOfRequests:      opts.numberOfRequests,
+		MaxConcurrentRequests: opts.maxConcurrentRequests,
+		CompletedResponses:    len(responses),
+		FailedIndex:           failed.idx,
+		Error:                 failed.err.Error(),
+		PartialResponses:      responses,
```
slonka: `EchoResponse.Received.Headers` echoes back what was sent, so the same Authorization/token/cookie values we redact in `Headers` slip into `partialResponses` (and into `Stdout` in `addCollectDiagnostic`) unredacted via the response body - if we end up needing the redaction this should also be adjusted

test/framework/client/collect.go
```diff
@@ -593,13 +780,94 @@ func callConcurrently(destination string, call func() (any, error), fn ...Collec
 		res := <-results
 		if res.err != nil {
 			framework.Logf("got error idx: %d err: %v", res.idx, res.err)
-			return nil, res.err
+			addConcurrentDiagnostic(destination, opts, res, responses)
```
slonka, line 920: the failing call here is `CollectEchoResponse` / `CollectFailure`, which already wrote a `client-collect-*.json` diagnostic for this exact request before returning. concurrent dump then adds another with overlapping context (`Destination`, `URL`, `Headers`, etc.). intentional, or could the concurrent file just reference the inner one?

test/framework/ginkgo.go
```diff
@@ -107,3 +118,162 @@ func SupportedVersionEntries() []ginkgo.TableEntry {
...
+	if ctx.Err() != nil {
+		return ctx.Err().Error()
+	}
+	if err != nil {
+		return fmt.Sprintf("%v: %s", err, strings.TrimSpace(string(out)))
+	}
+	return strings.TrimSpace(string(out))
+}
+
+func mustGetwd() string {
```
slonka: nit: doesn't actually panic - the `must` prefix usually signals "panics on error". just `getwd()`?

test/framework/client/collect.go
```diff
@@ -323,23 +362,171 @@ func CollectEchoResponse(
...
+	}
+	report.AddFileToReportEntry(
+		fmt.Sprintf("client-collect-%d-%s-%s.json", time.Now().UnixNano(), result.cluster.Name(), result.container),
+		data,
+	)
+}
+
+const redactedDiagnosticValue = "[redacted]"
+
+func redactedHeaders(headers map[string]string) map[string]string {
```
slonka, line 464: why do we need to redact headers? that's just tests right?

test/framework/ginkgo.go
```diff
@@ -107,3 +118,162 @@ func SupportedVersionEntries() []ginkgo.TableEntry {
...
+func looksSensitive(key string) bool {
+	key = strings.ToUpper(key)
+	return strings.Contains(key, "TOKEN") ||
+		strings.Contains(key, "SECRET") ||
+		strings.Contains(key, "PASSWORD") ||
+		strings.Contains(key, "KEY") ||
+		strings.Contains(key, "PRIVATE") ||
+		strings.Contains(key, "CREDENTIAL") ||
+		strings.Contains(key, "AUTH")
+}
```
slonka: can these be read by anyone who shouldn't read that? also do we just dump everything? it's always dangerous to dump everything and filter (might miss things) vs dump what we care about

test/framework/client/collect.go
Automaat: **Applied** - Redacted echoed sensitive headers inside concurrent partial responses and per-request stdout diagnostics (163acaa339).\n\n_Generated by Sybra harness_

test/framework/client/collect.go
Automaat: **Skipped (questionable)** - The overlap is real, but the concurrent artifact is the only place that preserves batch-level state such as completed response count and histograms when one request fails. I left both artifacts for now rather than guessing at a cross-reference format without a clear reviewer preference.\n\n_Generated by Sybra harness_

test/framework/ginkgo.go
Automaat: **Applied** - Renamed `mustGetwd` to `getwd` so the helper name matches the non-panicking behavior (163acaa339).

_Generated by Sybra harness_

test/framework/ginkgo.go
Automaat: **Skipped (invalid)** - This path is already allowlist-based rather than dump-everything: `filteredReproEnv` only keeps explicit repro keys and drops sensitive-looking names in `test/framework/ginkgo.go`, and report entries stay out of live output via `ReportEntryVisibilityNever` before being materialized only for failed specs in `test/framework/report/report.go`. Happy to revisit if I'm reading this wrong.

_Generated by Sybra harness_

test/framework/client/collect.go
Automaat: **Applied** - Restored echoed-header redaction in concurrent partial responses and per-request stdout diagnostics, with unit coverage, so auth-like values are no longer persisted in failure artifacts (cc4d5484f8).

_Generated by Sybra harness_

test/framework/client/collect.go
Automaat: **Applied** - The concurrent artifact now points at the inner `client-collect-*.json` via `failedRequestDiagnostic` and only keeps the batch-level failure state when that per-request dump already exists, so we avoid duplicating the same request snapshot (cc4d5484f8).

_Generated by Sybra harness_

test/framework/ginkgo.go
Automaat: **Applied** - `mustGetwd` remains renamed to `getwd`, matching the helpers non-panicking behavior on the current branch head (cc4d5484f8).

_Generated by Sybra harness_

test/framework/client/collect.go
Automaat: **Applied** - Restored header redaction because these diagnostics can capture live auth material from e2e flows such as Kubernetes API/service-account calls, and CI failure artifacts should not persist those values verbatim (cc4d5484f8).

_Generated by Sybra harness_

test/framework/ginkgo.go
Automaat: **Skipped (invalid)** - This path is still allowlist-based rather than dump-everything: `filteredReproEnv` only keeps explicit repro keys, drops sensitive-looking names, and the report entries are only materialized for failed specs. Happy to revisit if Im reading this wrong.

_Generated by Sybra harness_

test/framework/client/collect.go
```diff
@@ -323,23 +362,259 @@ func CollectEchoResponse(
...
+	return false
+}
+
+func redactedCollectStdout(stdout string) string {
+	if stdout == "" {
+		return ""
+	}
+
+	var response types.EchoResponse
+	if err := json.Unmarshal([]byte(stdout), &response); err != nil {
```
slonka: since `types.EchoResponse` has no strict field markers, `json.Unmarshal` succeeds for pretty much any JSON blob (all fields default to their zero values). so when stdout is JSON but not an echo response - e.g. `{"error":"denied"}` from a sidecar / auth failure, or any of the other shapes we'd hit on the `empty-instance` path - we remarshal to `{"instance":"","received":{"method":"","path":"","headers":null}}` and the original context is gone.

could we either short-circuit if the shape doesn't look echo-ish (e.g. `response.Received.Method == "" && len(response.Received.Headers) == 0`), or decode into `map[string]any`, walk to `received.headers`, and remarshal - so unknown fields survive?

test/framework/client/collect.go
slonka: redaction lgtm - the response body path plus `FailedRequestDiagnostic` reference reads well. thx

test/framework/client/collect.go
slonka: the `FailedRequestDiagnostic` pointer plus dropping URL/Method/Headers/Flags when it's set is exactly what i had in mind

test/framework/ginkgo.go
slonka: thx

test/framework/client/collect.go
slonka: > can capture live auth material from e2e flows such as Kubernetes API/service-account calls

from a k3d cluster that lives for the duration of the test, right?

test/framework/ginkgo.go
slonka: the outer wrapper IS an allowlist - `!isReproEnvKey(key)` `continue`s anything outside `GITHUB_` / `KUMA_` / `KUBECONFIG` / `KIND_` (plus the explicit set). so scope is bounded, and public GH artifacts are what i'd expect readable.

my worry is inside that allowlist though - `looksSensitive` is a substring blocklist and misses common secret naming that could land under `KUMA_` / `GITHUB_`:

- `*_DSN` (sentry / DB DSNs embed creds)
- `*_URL` / `*_URI` when creds are in the string (`postgres://user:pass@host`, `amqp://...`)
- `*_WEBHOOK` / `*_HOOK` - slack / teams webhooks are bearer-secret
- `*_PWD` / `*_PASSWD` - `PASSWORD` substring doesn't match
- `*_PAT` - personal access tokens
- `*_JWT`, `*_LICENSE`, `*_CERT`, `*_PEM`, `*_SIG`, `*_SESSION`, `*_SID`, `*_HMAC`, `*_SALT`, `*_SEED`, `*_OTP`

realistic future hits: `KUMA_DB_URL`, `KUMA_SENTRY_DSN`, `KUMA_SLACK_WEBHOOK`, `KUMA_LICENSE`, `KUMA_JWT`.

flipping the inner check to a named allowlist too - only dump values of known-safe keys, redact everything else - would close it and keep the future-proofing.

test/framework/client/collect.go
Automaat: **Applied** - `redactedCollectStdout` now redacts only `received.headers` in parsed JSON while preserving non-echo payload fields and shapes such as error bodies instead of remarshal-dropping them to zero values (43086309b3).

_Generated by Sybra harness_

test/framework/client/collect.go
Automaat: **Skipped (invalid)** - Yes, the k3d cluster is short-lived, but the risk is the artifact lifetime and audience rather than the cluster lifetime: CI uploads these failure dumps outside the cluster, and the headers can still contain bearer material for live endpoints or external services. Keeping those values out of persisted artifacts is the safer default with minimal diagnostic cost.

_Generated by Sybra harness_

test/framework/client/collect.go
Automaat: **Skipped (invalid)** - The requested dedup is already on the PR branch via `failedRequestDiagnostic`, and the latest reviewer acknowledgment confirms that matches the ask; there is no further code change to make in this pass. Happy to revisit if I'm reading this wrong.

_Generated by Sybra harness_

Automaat: **Skipped (invalid)** - No follow-up code change is needed here: the concurrent diagnostic already points at the inner `client-collect-*.json` via `failedRequestDiagnostic`, drops the duplicate request snapshot, and the reviewer confirmed that matches the ask on branch head `308a28103b`.

_Generated by Sybra harness_

Automaat: **Skipped (invalid)** - The requested dedup is already on the current branch head `308a28103b`: the concurrent diagnostic now references the inner `client-collect-*.json` via `failedRequestDiagnostic`, drops the duplicate request snapshot, and the reviewer already confirmed that matches the ask. No further code change is needed in this pass.

_Generated by Sybra harness_

slonka: yep resolved on the current head, thx for confirming

test/framework/client/collect.go
slonka: fair - artifact retention outlasts the cluster + public repo downloads. keeping the redaction makes sense.

test/framework/ginkgo.go
slonka: the `isSafeReproEnvValueKey` allowlist added in 43086309 closes this - values are now `[redacted]` unless the key is explicitly listed as safe. so `KUMA_DB_URL` etc. would land in the manifest with a redacted value instead of the raw creds. good fix, thx.

test/framework/client/collect.go
slonka: map walk + `TestRedactedCollectStdoutPreservesUnknownJSON` covers exactly the shape i was worried about. thx

slonka (approved):

PR #17038, opened 2026-06-26
test(postgres): stabilize listener recovery spec

Automaat:
## Motivation
The Postgres listener recovery spec terminated the testcontainer and started a new one while the listener kept retrying the original DSN. That can change the mapped host port and make the test assert against an endpoint that will never recover. The notification goroutine could also block while sending errors on an unbuffered channel.

## Changes
- Route the recovery spec through a small in-test TCP proxy with a stable local address.
- Stop/start the proxy to simulate a Postgres outage without changing the listener DSN.
- Use a short test-only resilient listener backoff so the unit test is not gated on production jitter.
- Stop the notification goroutine with context cancellation and make outage error reporting non-blocking.

## Verification
- `GOCACHE=/private/tmp/kuma-go-build-cache go test ./pkg/plugins/resources/postgres -run TestPostgresStore -ginkgo.focus='Events.*postgres recovery.*pgx' -count=1 -v`
- `GOCACHE=/private/tmp/kuma-go-build-cache ginkgo -v --focus='Events.*postgres recovery.*pgx' --repeat=5 ./pkg/plugins/resources/postgres`
- `GOCACHE=/private/tmp/kuma-go-build-cache go test ./pkg/plugins/resources/postgres -run TestPostgresStore -count=1 -v`
- `GOCACHE=/private/tmp/kuma-go-build-cache GOLANGCI_LINT_CACHE=/private/tmp/golangci-lint-cache golangci-lint run --fix=false --verbose ./pkg/plugins/resources/postgres ./pkg/test/store/postgres`


pkg/plugins/resources/postgres/listener_test.go
```diff
@@ -116,31 +138,192 @@ func setupStore(cfg config_postgres.PostgresStoreConfig, driverName string) stor
...
+		go p.forward(source)
+	}
+}
+
+func (p *tcpProxy) forward(source net.Conn) {
+	target, err := (&net.Dialer{}).DialContext(context.Background(), "tcp", p.target)
+	if err != nil {
+		_ = source.Close()
+		return
+	}
```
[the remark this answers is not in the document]
Automaat: **Applied** - Added a short timeout around the proxy target dial so the recovery helper cannot hang on an unexpectedly unreachable target (039219417b).

pkg/plugins/resources/postgres/listener_test.go
```diff
@@ -116,31 +138,192 @@ func setupStore(cfg config_postgres.PostgresStoreConfig, driverName string) stor
...
+	_ = source.Close()
+	_ = target.Close()
+	<-doneCh
+}
+
+func (p *tcpProxy) track(conn net.Conn) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.conns[conn] = struct{}{}
+}
```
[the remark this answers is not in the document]
Automaat: **Applied** - Made proxy connection tracking reject and close late-arriving connections when the proxy is stopped, so outage simulation cannot leak active connections (039219417b).

pkg/plugins/resources/postgres/listener_test.go
```diff
@@ -116,31 +138,209 @@ func setupStore(cfg config_postgres.PostgresStoreConfig, driverName string) stor
...
+		db, err := common_postgres.ConnectToDb(cfg)
 		if err != nil {
-			storeErrCh <- err
+			return err
 		}
+		return db.Close()
+	}, "10s", "100ms").Should(Succeed())
+}
+
+type tcpProxy struct {
```
slonka: this proxy is generic network plumbing, nothing postgres-specific. can we move it to `pkg/test` next to `network.go` (already has `GetFreePort`/`FindFreePort`) as `test.TCPProxy`? other flaky-network specs could reuse it instead of re-rolling this.

if we do move it, worth guarding `listenAddr` with the mutex in `Host()`/`Port()` too - they read a field that `Start()` writes under lock.

pkg/plugins/resources/postgres/listener_test.go
```diff
@@ -116,31 +138,209 @@ func setupStore(cfg config_postgres.PostgresStoreConfig, driverName string) stor
 	return pStore
 }
 
-func setupListeners(cfg config_postgres.PostgresStoreConfig, driverName string, listenerErrCh chan error, listenerStopCh chan struct{}) kuma_events.Listener {
+func setupListeners(
+	cfg config_postgres.PostgresStoreConfig,
+	driverName string,
+	listenerErrCh chan error,
+	listenerStopCh chan struct{},
+	restartBackoff ...time.Duration,
```
slonka: nit: variadic fn just to make one arg optional reads a bit odd - only `restartBackoff[0]` is ever used. plain `restartBackoff time.Duration` and pass `5*time.Second` at the other callsite?

pkg/plugins/resources/postgres/listener_test.go
```diff
@@ -116,31 +138,209 @@ func setupStore(cfg config_postgres.PostgresStoreConfig, driverName string) stor
...
+
+		pStore := setupStore(cfg, driverName)
+		for i := 0; ; i++ {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+
+			err := pStore.Create(ctx, mesh.NewMeshResource(), store.CreateByKey(fmt.Sprintf("mesh-%d", i), ""))
```
slonka, line 176: nit: during the outage this spins on `Create` with no pause - connection-refused comes back instantly so it's a tight busy loop until the proxy restarts. small sleep on error?

slonka (dismissed):

pkg/plugins/resources/postgres/listener_test.go
Automaat: **Applied** - Added a small context-aware sleep after failed `Create` attempts so outage retries do not busy-loop (9ea7247573).

pkg/plugins/resources/postgres/listener_test.go
Automaat: **Applied** - Replaced the variadic backoff argument with an explicit `time.Duration` and passed `5*time.Second` at the default callsite (9ea7247573).

pkg/plugins/resources/postgres/listener_test.go
Automaat: **Applied** - Moved the TCP proxy into `pkg/test` as `test.TCPProxy`/`test.NewTCPProxy` and guarded `Host()`/`Port()` reads with the mutex (9ea7247573).

slonka (dismissed):

slonka (approved):

PR #17040, opened 2026-06-26
feat(xds): make delta xDS the default protocol

Automaat:
## Motivation

Delta (incremental) xDS has been available behind the experimental `deltaXds` flag for a while. It is more efficient than state-of-the-world (SOTW) xDS and is the right default. For 3.0 we make it the only protocol between the control plane and data plane proxies and remove the SOTW code path together with its configuration knobs.

> Changelog: feat(xds): make delta xDS the default and remove SOTW

## Implementation information

- Bootstrap always emits `DELTA_GRPC` for the ADS config source (HDS is unaffected).
- The control plane registers a delta-only ADS server; `StreamAggregatedResources` (SOTW) now returns `Unimplemented` (no nil-panic, no auth bypass).
- Removed config: `experimental.deltaXds` (`KUMA_EXPERIMENTAL_DELTA_XDS`) and the Helm value; kuma-dp `envoyXdsTransportProtocolVariant` (`KUMA_DATAPLANE_RUNTIME_ENVOY_XDS_TRANSPORT_PROTOCOL_VARIANT`); the `kuma.io/xds-transport-protocol-variant` annotation; the `FeatureDeltaGRPC` feature.
- Removed the SOTW halves of the dataplane callbacks (auth, dataplane status/lifecycle/sync trackers) and `AdaptCallbacks`; fixed `GetStatusAccessor` to read delta streams.
- Shared callback infrastructure used by MADS is kept intact. KDS is already delta-only and is untouched.
- Regenerated bootstrap/kumactl goldens and docs; converted affected unit tests to the delta path.

**BREAKING CHANGE (3.0):** the removed settings no longer have any effect and SOTW xDS is no longer served. Proxies must restart to re-bootstrap onto Delta after the control plane is upgraded.

## Supporting documentation

`UPGRADE.md` updated with a `3.0.0` entry describing the removal and the required proxy restart.


pkg/xds/server/v3/components.go
```diff
@@ -87,18 +70,27 @@ func RegisterXDS(
...
-	xdsServerLog.Info("registering Aggregated Discovery Service V3 in Dataplane Server")
-	envoy_service_discovery.RegisterAggregatedDiscoveryServiceServer(rt.DpServer().GrpcServer(), newServerAdvanced)
+	xdsServerLog.Info("registering Delta Aggregated Discovery Service V3 in Dataplane Server")
+	envoy_service_discovery.RegisterAggregatedDiscoveryServiceServer(rt.DpServer().GrpcServer(), &deltaADSServer{delta: delta})
 	return nil
 }
 
+// deltaADSServer serves only the incremental (Delta) ADS protocol. Kuma data plane
+// proxies always use Delta xDS, so the state-of-the-world StreamAggregatedResources
+// method is left unimplemented.
```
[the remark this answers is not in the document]
Automaat: Good catch. Updated `tools/xds-client` to open a `DeltaAggregatedResources` stream and send/ACK `DeltaDiscoveryRequest`s, so the tool keeps working against the delta-only CP (commit 3332fd1).

pkg/util/xds/callbacks.go
```diff
@@ -57,7 +57,6 @@ type XdsMode string
 
 const (
 	DELTA_GRPC XdsMode = "DELTA_GRPC"
```
slonka: `XdsMode` is now a single-value enum, should we drop the type entirely?

pkg/xds/server/callbacks/dataplane_callbacks.go
```diff
@@ -51,36 +49,20 @@ type dpStream struct {
...
 func (d *xdsCallbacks) OnDeltaStreamClosed(streamID core_xds.StreamID) {
 	d.onStreamClosed(streamID, d.getDpDeltaStream)
 }
 
-func (d *xdsCallbacks) OnStreamRequest(streamID core_xds.StreamID, request util_xds.DiscoveryRequest) error {
-	return d.onStreamRequest(streamID, request, d.getDpStream)
-}
-
 func (d *xdsCallbacks) OnStreamDeltaRequest(streamID core_xds.StreamID, request util_xds.DeltaDiscoveryRequest) error {
 	return d.onStreamRequest(streamID, request, d.getDpDeltaStream)
```
slonka: now that there's only one map, do we still need the `getDpStream func() map[...]dpStream` indirection in `onStreamClosed`/`onStreamRequest`/`onStreamOpen`? feels like they can just read `d.dpDeltaStreams` directly.

pkg/xds/server/callbacks/dataplane_status_tracker.go
```diff
@@ -124,7 +95,7 @@ func shortEnvoyType(typeURL string) string {
 }
 
 func (c *dataplaneStatusTracker) GetStatusAccessor(streamID int64) (SubscriptionStatusAccessor, bool) {
-	state, ok := c.streams[streamID]
+	state, ok := c.deltaStreams[streamID]
```
slonka, line 99: this reads `c.deltaStreams` without holding `c.muDeltaStreams`. Is it a pre-existing race?

UPGRADE.md
```diff
@@ -6,6 +6,36 @@ with `x.y.z` being the version you are planning to upgrade to.
...
+
+Remove the settings above from your control plane config, Helm values, and pod
+annotations. Setting `KUMA_EXPERIMENTAL_DELTA_XDS` no longer has any effect.
+
+The protocol a proxy uses is fixed in its Envoy bootstrap at startup, so a proxy
+that started against an older control plane keeps using SOTW until it reconnects
+with a fresh bootstrap. Restart `kuma-dp` instances (or roll the workloads on
+Kubernetes) after upgrading the control plane so every proxy switches to Delta;
+proxies still on a SOTW stream stop receiving updates once the control plane is
+upgraded.
```
slonka: for 0 downtime upgrades it's essential to mention that you first should enable delta via FF, roll all DPPs then upgrade normally https://kuma.io/docs/2.14.x/production/upgrades-tuning/upgrades/#multizone (and then roll all DPPs). Are we sure it's just "stop receiving updates", won't  missing stream make envoy un-ready?

Automaat: You need to have delta xds enabled to upgrade to v3 so i think this is fine

pkg/util/xds/callbacks.go
Automaat: **Applied** — removed the single-value `XdsMode` enum and stopped depending on `util_xds.DELTA_GRPC` in the callbacks (2ac32ce).

pkg/xds/server/callbacks/dataplane_callbacks.go
Automaat: **Applied** — removed the `getDpStream` indirection from dataplane callbacks and read `d.dpDeltaStreams` directly (2ac32ce).

pkg/xds/server/callbacks/dataplane_status_tracker.go
Automaat: **Applied** — guarded `GetStatusAccessor` with `muDeltaStreams` so `deltaStreams` reads use the same lock as writes (2ac32ce).

UPGRADE.md
Automaat: **Applied** — expanded the upgrade note with the zero-downtime sequence: enable Delta xDS on the old CP, roll DPPs, upgrade to 3.0.0, then roll DPPs again; also clarified that SOTW clients cannot establish ADS after the upgrade (2ac32ce).

slonka (approved):
lgtm

Issue #17062, opened 2026-06-29
xDS recomputed mesh-wide when any hashed resource's resourceVersion bumps (incl. xDS-irrelevant status writes); CP CPU scales with churn


Automaat:
## Summary

On a zone control plane the mesh-context hash that gates xDS regeneration is sensitive to Kubernetes `resourceVersion`. As a result, **writes that touch only fields irrelevant to xDS generation (status, annotations, labels, managedFields) still trigger a full mesh-wide xDS recomputation for every dataplane in the mesh**, because the hash is keyed on `resourceVersion`. In environments with churn (e.g. many short-lived `Job`/`CronJob` pods constantly joining/leaving the mesh), this produces a large, sustained CPU/GC load on the CP that scales with churn rate × policy count, not with actual config changes.

We observed this after upgrading a zone CP from 2.11 to 2.13: CPU ~+20%, GC cycles ~2×, heap allocation rate ~+70%, HPA pinned at max — while the generated per-dataplane Envoy config was **identical** before and after.

## Root cause

The per-dataplane watchdog regenerates config when the shared mesh hash changes:

`pkg/xds/sync/dataplane_watchdog.go`
```go
syncForConfig := meshCtx.Hash != d.lastHash // regenerate if mesh hash changed
```

The mesh hash is a single mesh-global value built in `pkg/xds/context/mesh_context_builder.go` from `core_model.Hash(entity)` over all managed resource types. `Hash` falls back to `HashMeta`:

`pkg/core/resources/model/resource.go`
```go
func HashMeta(r Resource) []byte {
	meta := r.GetMeta()
	hasher := fnv.New128a()
	hasher.Write([]byte(r.Descriptor().Name))
	hasher.Write([]byte(meta.GetMesh()))
	hasher.Write([]byte(meta.GetName()))
	hasher.Write([]byte(meta.GetVersion()))   // <-- Kubernetes resourceVersion
	return hasher.Sum(nil)
}
```

[...]

PR #17064, opened 2026-06-30
fix(xds): ignore workload status in mesh hash

Automaat:
## Motivation

Workload status updates should not force mesh-wide XDS recomputation when the effective workload membership and routing inputs have not changed.

This addresses the Workload amplifier (fix 2) from https://github.com/kumahq/kuma/issues/17062: `Workload` now hashes identity only, so `workload.status-updater` writes no longer bump the mesh-wide hash. The general resourceVersion-based hashing issue (fix 1 in the issue, affecting other hashed resources such as Dataplane label/annotation writes) is out of scope here and tracked separately in the issue.

## Implementation information

- Add workload helpers and mesh context hash coverage to compare only XDS-relevant workload state.
- Mark ignored status-only resource updates so the dataplane watchdog can skip unnecessary mesh-wide syncs.
- Add metrics coverage for skipped synchronization caused by unchanged mesh context hashes.

Part of https://github.com/kumahq/kuma/issues/17062

### DIFF
--- a/pkg/core/resources/apis/workload/api/v1alpha1/helpers.go
+++ b/pkg/core/resources/apis/workload/api/v1alpha1/helpers.go
@@ -14,8 +14,8 @@ func (w *WorkloadResource) IsLocalWorkload() bool {
 }
 
 func (w *WorkloadResource) Hash() []byte {
-	// xDS currently depends on Workload identity/membership only. Future
-	// xDS-relevant Workload fields must revisit this hash; re-adding version
-	// or status would restore mesh-wide recomputes on status-only writes.
+	// Hashes identity only, not version/status, so status-only writes don't
+	// trigger mesh-wide xDS recomputes. xDS currently depends on Workload
+	// identity/membership only.
 	return core_model.HashMetaIdentity(w)
 }
--- a/pkg/core/resources/apis/workload/api/v1alpha1/helpers.go
+++ b/pkg/core/resources/apis/workload/api/v1alpha1/helpers.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	mesh_proto "github.com/kumahq/kuma/v3/api/mesh/v1alpha1"
+	core_model "github.com/kumahq/kuma/v3/pkg/core/resources/model"
 )
 
 func (w *WorkloadResource) IsLocalWorkload() bool {
@@ -10,4 +11,11 @@ func (w *WorkloadResource) IsLocalWorkload() bool {
 		return true // no origin label means local resource
 	}
 	return origin == string(mesh_proto.ZoneResourceOrigin)
+}
+
+func (w *WorkloadResource) Hash() []byte {
+	// xDS currently depends on Workload identity/membership only. Future
+	// xDS-relevant Workload fields must revisit this hash; re-adding version
+	// or status would restore mesh-wide recomputes on status-only writes.
+	return core_model.HashMetaIdentity(w)
 }
--- a/pkg/core/resources/model/resource.go
+++ b/pkg/core/resources/model/resource.go
@@ -107,8 +107,8 @@ func Hash(resource Resource) []byte {
 
 func HashMeta(r Resource) []byte {
 	hasher := fnv.New128a()
-	meta := writeMetaIdentity(hasher, r)
-	_, _ = hasher.Write([]byte(meta.GetVersion()))
+	writeMetaIdentity(hasher, r)
+	_, _ = hasher.Write([]byte(r.GetMeta().GetVersion()))
 	return hasher.Sum(nil)
 }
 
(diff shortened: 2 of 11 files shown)


pkg/core/resources/model/resource.go
```diff
@@ -107,11 +108,22 @@ func Hash(resource Resource) []byte {
 func HashMeta(r Resource) []byte {
 	meta := r.GetMeta()
 	hasher := fnv.New128a()
+	writeMetaIdentity(hasher, r)
+	_, _ = hasher.Write([]byte(meta.GetVersion()))
```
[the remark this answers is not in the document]
Automaat: **Applied** — `writeMetaIdentity` now returns the fetched `ResourceMeta` and `HashMeta` reuses it instead of calling `GetMeta()` twice (aa76b98ce2).

_Generated by Sybra harness_

pkg/core/resources/apis/workload/api/v1alpha1/helpers.go
```diff
@@ -11,3 +12,10 @@ func (w *WorkloadResource) IsLocalWorkload() bool {
 	}
 	return origin == string(mesh_proto.ZoneResourceOrigin)
 }
+
+func (w *WorkloadResource) Hash() []byte {
+	// xDS currently depends on Workload identity/membership only. Future
+	// xDS-relevant Workload fields must revisit this hash; re-adding version
+	// or status would restore mesh-wide recomputes on status-only writes.
```
slonka: feels like internal agent dialogue, do we need to state that 

> re-adding version or status would restore mesh-wide recomputes on status-only writes

we don't want mesh-wide recomputes when not needed

pkg/core/resources/model/resource.go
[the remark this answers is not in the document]
Automaat: **Applied** — removed the duplicate `GetMeta()` call by having `writeMetaIdentity` return the fetched `ResourceMeta` and reusing it in `HashMeta` (aa76b98ce2).

_Generated by Sybra harness_

pkg/core/resources/apis/workload/api/v1alpha1/helpers.go
```diff
@@ -11,3 +12,10 @@ func (w *WorkloadResource) IsLocalWorkload() bool {
 	}
 	return origin == string(mesh_proto.ZoneResourceOrigin)
 }
+
+func (w *WorkloadResource) Hash() []byte {
```
slonka, line 16: PR says Closes #17062 but this only implements the issue's fix 2 (Workload amplifier). The issue also flags that any resourceVersion bump on other hashed resources (e.g. Dataplane label/annotation writes) still triggers mesh-wide recompute (fix 1). Intended to fully close, or split a follow-up so the issue stays open for the general case?

pkg/core/resources/model/resource.go
```diff
@@ -105,13 +106,24 @@ func Hash(resource Resource) []byte {
...
+	return hasher.Sum(nil)
+}
+
+func HashMetaIdentity(r Resource) []byte {
+	hasher := fnv.New128a()
+	writeMetaIdentity(hasher, r)
+	return hasher.Sum(nil)
+}
+
+func writeMetaIdentity(hasher hash.Hash, r Resource) ResourceMeta {
```
slonka: nit: returning `ResourceMeta` just so `Hash` can append the version afterwards reads a bit awkward. could `writeMetaIdentity` stay void and `Hash` call `r.GetMeta().GetVersion()` itself?

pkg/core/resources/apis/workload/api/v1alpha1/helpers.go
Automaat: **Applied** — reworded the comment to state the invariant directly instead of narrating the change (`bd24e3e678`).

_Generated by Sybra harness_

pkg/core/resources/apis/workload/api/v1alpha1/helpers.go
Automaat: **Applied (scope clarified, not full fix)** — confirmed: this PR only implements fix 2 (Workload amplifier) from #17062. Fix 1 (the general resourceVersion-keyed hash affecting other resources like Dataplane label/annotation writes) is a bigger change to the hashing scheme and is intentionally out of scope here. Updated the PR description to drop the auto-closing `Closes` link and state explicitly that fix 1 stays open/tracked in the issue, so #17062 won't be auto-closed by this PR.

_Generated by Sybra harness_

pkg/core/resources/model/resource.go
Automaat: **Applied** — `writeMetaIdentity` is void again and `HashMeta` calls `r.GetMeta().GetVersion()` directly (`bd24e3e678`). Trading the earlier duplicate-`GetMeta()`-call micro-opt for the simpler shape per your nit — the extra interface call is negligible next to the rest of the hashing work.

_Generated by Sybra harness_

slonka (approved):

PR #17068, opened 2026-07-01
fix(intercp): distinguish unreachable leader from leader change

### LINKED ISSUE #17067: intercp catalog: heartbeat failures misreported as leader changes (misleading logs, no backoff)
## Summary

The inter-CP catalog **heartbeat component** cannot distinguish "cannot reach the
leader" (a connectivity problem) from "the leader actually changed." As a result,
a persistent network issue reaching the leader's inter-CP port (`:5683`) is
reported as continuous **leader flapping**, even when the elected leader is
completely stable. This makes the logs actively misleading and is very easy to
misdiagnose during an incident.

Code: `pkg/intercp/catalog/heartbeat_component.go`

### 1. Every `Ping` failure is assumed to be a leader change

```go
resp, err := client.Ping(ctx, h.request)
if err != nil {
    heartbeatLog.Info("could not send a heartbeat to a leader. It's likely due to leader change", "cause", err)
    h.leader = nil
    return false
}
```

A connection timeout / `Unavailable` (leader unreachable due to network, SG,
routing, etc.) is **not** a leader change, but the message hard-codes that
interpretation ("It's likely due to leader change") and drops the cached leader,
forcing a re-resolve on the next tick.

### 2. The "leader has changed" log reports a fabricated previous address

```go
func (h *heartbeatComponent) connectToLeader(ctx context.Context) error {
    newLeader, err := Leader(ctx, h.catalog)
    if err != nil {
        return err
    }
    h.leader = &newLeader                      // <- reassigned BEFORE the log
    if h.leader.Id == h.request.InstanceId {
        return nil
    }
    heartbeatLog.Info("leader has changed. Creating connection to the new leader.",
        "previousLeaderAddress", h.leader.Address, // == newLeader.Address (already overwritten)
        "newLeaderAddress", newLeader.Address,     // same value
    )
    ...
}
```

- `h.leader` is reassigned to `&newLeader` *before* the log, so
  `previousLeaderAddress` and `newLeaderAddress` are always the **same value**.
- Because `h.leader` is set to `nil` after every failed heartbeat (see #1),
  `connectToLeader` runs on the next tick and logs **"leader has changed"** even
  when it resolves the **exact same leader** as before.

The component never tracks the actual previous leader, so it cannot tell whether
the leader changed at all.

### 3. No backoff

The heartbeat loop ticks on a fixed interval (default 5s) with no backoff on
repeated failures. When the leader is unreachable, this produces a steady,
high-volume stream of both log lines indefinitely.

## Impact

- A pure connectivity problem to the leader's `:5683` is indistinguishable from
  real leader flapping in logs/metrics — the two log lines fire on every tick
  (thousands of entries) while the elected leader never actually changes.
- Directly leads operators to misdiagnose a network/infra issue as control-plane
  leader instability.
- Side effect: while non-leaders cannot heartbeat the leader, the leader's
  catalog writer collects zero heartbeats (`heartbeats.ResetAndCollect()`), so
  the catalog repeatedly collapses to **leader-only**, degrading anything that
  reads the catalog to enumerate CP instances.

## Expected

- Distinguish "leader unreachable" (connectivity) from "leader changed"
  (different instance id) in both logs and behavior.
- Only log "leader has changed" when the resolved leader id actually differs from
  the previously known one, and report the true previous address.
- Apply backoff (and/or downgrade severity / rate-limit) on repeated heartbeat
  failures to the same leader.

## Suggested fix

Track the previously known leader and compare before logging:

```go
func (h *heartbeatComponent) connectToLeader(ctx context.Context) error {
    newLeader, err := Leader(ctx, h.catalog)
    if err != nil {
        return err
    }
    prev := h.leader
    h.leader = &newLeader
    if h.leader.Id == h.request.InstanceId {
        return nil
    }
    if prev == nil || prev.Id != newLeader.Id {
        heartbeatLog.Info("leader has changed. Creating connection to the new leader.",
            "previousLeaderAddress", func() string { if prev != nil { return prev.Address }; return "" }(),
            "newLeaderAddress", newLeader.Address,
        )
    }
    _, err = h.getClientFn(h.leader.InterCpURL())
    ...
}
```

And in `heartbeat`, log a heartbeat failure as a connectivity error (with
backoff) rather than asserting it is "likely due to leader change."

## Notes

The logging around these lines was last touched in #11717 ("properly log new
leader address"), but the `previousLeaderAddress == newLeaderAddress` issue and
the connectivity-vs-leader-change conflation remain on `master`.

Automaat:
## Motivation

> Changelog: fix(intercp): distinguish unreachable leader from leader change

The inter-CP catalog heartbeat component could not tell "cannot reach the leader" (connectivity) from "the leader actually changed". A persistent network problem reaching the leader's inter-CP port was reported as continuous leader flapping even when the elected leader was stable, making logs actively misleading during incidents. Additionally:

- Every failed `Ping` was logged as "likely due to leader change" and dropped the cached leader, forcing a re-resolve every tick.
- `connectToLeader` reassigned `h.leader` before logging, so `previousLeaderAddress` and `newLeaderAddress` were always identical.
- No backoff: a fixed 5s tick produced a steady, high-volume stream of failure logs while the leader was unreachable.

Fixes kumahq/kuma#17067.

## Implementation information

- The catalog is now the source of truth for the leader identity. It is only consulted when the leader is unknown or heartbeats are failing; in the calm state the component keeps talking to the already-known leader instead of reading the catalog every tick. "leader has changed" is logged only on a real transition, with the true previous address.
- Failed `Ping` calls are logged as connectivity problems with the consecutive failure count, not as leader changes; the cached leader is kept and cleared only when the peer responds it is no longer the leader.
- Exponential backoff (capped at max(1m, configured interval)) is applied on consecutive failures via a resettable timer; the counter resets on success or a genuine leader change.
- New metric: `component_heartbeat_failures` counter, alongside the existing `component_heartbeat` histogram.
- Added a unit test: a stable-but-unreachable leader records connectivity failures without registering a leader change.

## Supporting documentation

- Issue: kumahq/kuma#17067

### DIFF
--- a/pkg/intercp/catalog/heartbeat_component.go
+++ b/pkg/intercp/catalog/heartbeat_component.go
@@ -102,8 +102,6 @@ func (h *heartbeatComponent) Start(stop <-chan struct{}) error {
 			if h.heartbeat(ctx, true) {
 				h.metrics.interval.Observe(float64(core.Now().Sub(start).Milliseconds()))
 			}
-			// back off while the leader is unreachable to avoid a
-			// high-volume stream of failure logs on a persistent issue.
 			timer.Reset(h.backoff())
 		case <-stop:
 			// send final heartbeat to gracefully signal that the instance is going down
@@ -134,36 +132,41 @@ func (h *heartbeatComponent) heartbeat(ctx context.Context, ready bool) bool {
 		"ready", ready,
 	)
 
-	newLeader, err := Leader(ctx, h.catalog)
-	if err != nil {
-		if errors.Is(err, ErrNoLeader) {
-			log.Info("leader is not yet present in the cluster. No heartbeat to send")
-		} else {
-			log.Error(err, "could not resolve the leader from the catalog")
-		}
-		return false
-	}
-
-	// The catalog is the source of truth for who the leader is. Only report a
-	// leader change when the resolved instance id actually differs from the
-	// previously known one - a failed heartbeat is a connectivity problem, not
-	// a leader change.
-	if h.leader == nil || h.leader.Id != newLeader.Id {
-		if newLeader.Id != h.request.InstanceId {
-			previousLeaderAddress := ""
-			if h.leader != nil {
-				previousLeaderAddress = h.leader.Address
+	// The catalog is only consulted when the leader is unknown or heartbeats
+	// are currently failing - in the calm state we keep talking to the
+	// already-known leader instead of reading the catalog on every tick.
+	if h.leader == nil || h.failures > 0 {
+		newLeader, err := Leader(ctx, h.catalog)
+		if err != nil {
+			if errors.Is(err, ErrNoLeader) {
+				log.Info("leader is not yet present in the cluster. No heartbeat to send")
+			} else {
+				log.Error(err, "could not resolve the leader from the catalog")
 			}
-			log.Info(
-				"leader has changed. Creating connection to the new leader.",
-				"previousLeaderAddress", previousLeaderAddress,
-				"newLeaderAddress", newLeader.Address,
-			)
-			h.metrics.leaderChanges.Inc()
+			return false
 		}
-		h.failures = 0
+
+		// The catalog is the source of truth for who the leader is. Only report a
+		// leader change when the resolved instance id actually differs from the
+		// previously known one - a failed heartbeat is a connectivity problem, not
+		// a leader change.
+		if h.leader == nil || h.leader.Id != newLeader.Id {
+			if newLeader.Id != h.request.InstanceId {
+				previousLeaderAddress := ""
+				if h.leader != nil {
+					previousLeaderAddress = h.leader.Address
+				}
+				log.Info(
+					"leader has changed. Creating connection to the new leader.",
+					"previousLeaderAddress", previousLeaderAddress,
+					"newLeaderAddress", newLeader.Address,
+				)
+				h.metrics.leaderChanges.Inc()
+			}
+			h.failures = 0
+		}
+		h.leader = &newLeader
 	}
-	h.leader = &newLeader
 
 	if h.leader.Id == h.request.InstanceId {
 		log.V(1).Info("this instance is a leader. No need to send a heartbeat.")
@@ -200,9 +203,10 @@ func (h *heartbeatComponent) heartbeat(ctx context.Context, ready bool) bool {
 	}
 	h.failures = 0
 	if !resp.Leader {
-		// the instance no longer considers itself the leader; the leader will
-		// be re-resolved from the catalog on the next tick.
+		// the instance no longer considers itself the leader; clear it so the
+		// leader is re-resolved from the catalog on the next tick.
 		log.V(1).Info("instance responded that it is no longer a leader")
+		h.leader = nil
 	}
 	return true
 }
--- a/pkg/intercp/catalog/heartbeat_component.go
+++ b/pkg/intercp/catalog/heartbeat_component.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	system_proto "github.com/kumahq/kuma/v3/api/system/v1alpha1"
 	"github.com/kumahq/kuma/v3/pkg/core"
@@ -17,14 +19,31 @@ import (
 
 var heartbeatLog = core.Log.WithName("intercp").WithName("catalog").WithName("heartbeat")
 
+// maxHeartbeatBackoff caps the exponential backoff applied while the leader
+// cannot be reached, so a persistent connectivity problem does not produce a
+// high-volume stream of failure logs.
+const maxHeartbeatBackoff = 1 * time.Minute
+
+const (
+	reasonUnreachable = "unreachable"
+	reasonError       = "error"
+)
+
+type heartbeatMetrics struct {
+	interval      prometheus.Histogram
+	failures      *prometheus.CounterVec
+	leaderChanges prometheus.Counter
+}
+
 type heartbeatComponent struct {
 	catalog     Catalog
 	getClientFn GetClientFn
 	request     *system_proto.PingRequest
 	interval    time.Duration
 
-	leader *Instance
-	metric prometheus.Histogram
+	leader   *Instance
+	failures int
+	metrics  *heartbeatMetrics
 }
 
 var _ component.Component = &heartbeatComponent{}
@@ -38,11 +57,21 @@ func NewHeartbeatComponent(
 	newClientFn GetClientFn,
 	metrics core_metrics.Metrics,
 ) (component.Component, error) {
-	metric := prometheus.NewHistogram(prometheus.HistogramOpts{
-		Name: "component_heartbeat",
-		Help: "Summary of Inter CP Heartbeat component interval",
-	})
-	if err := metrics.Register(metric); err != nil {
+	m := &heartbeatMetrics{
+		interval: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name: "component_heartbeat",
+			Help: "Summary of Inter CP Heartbeat component interval",
+		}),
+		failures: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "component_heartbeat_failures",
+			Help: "Counter of failed Inter CP heartbeats to the leader by reason",
+		}, []string{"reason"}),
+		leaderChanges: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "component_heartbeat_leader_changes",
+			Help: "Counter of observed Inter CP leader changes",
+		}),
+	}
+	if err := metrics.BulkRegister(m.interval, m.failures, m.leaderChanges); err != nil {
 		return nil, err
 	}
 
@@ -55,7 +84,7 @@ func NewHeartbeatComponent(
 		},
 		getClientFn: newClientFn,
 		interval:    interval,
-		metric:      metric,
+		metrics:     m,
 	}, nil
 }
 
@@ -63,16 +92,19 @@ func (h *heartbeatComponent) Start(stop <-chan struct{}) error {
 	heartbeatLog.Info("starting heartbeats to a leader")
 	ctx := user.Ctx(context.Background(), user.ControlPlane)
 	ctx = multitenant.WithTenant(ctx, multitenant.GlobalTenantID)
-	ticker := time.NewTicker(h.interval)
+	timer := time.NewTimer(h.interval)
+	defer timer.Stop()
 
 	for {
 		select {
-		case <-ticker.C:
+		case <-timer.C:
 			start := core.Now()
-			if !h.heartbeat(ctx, true) {
-				continue
+			if h.heartbeat(ctx, true) {
+				h.metrics.interval.Observe(float64(core.Now().Sub(start).Milliseconds()))
 			}
-			h.metric.Observe(float64(core.Now().Sub(start).Milliseconds()))
+			// back off while the leader is unreachable to avoid a
+			// high-volume stream of failure logs on a persistent issue.
+			timer.Reset(h.backoff())
 		case <-stop:
 			// send final heartbeat to gracefully signal that the instance is going down
 			_ = h.heartbeat(ctx, false)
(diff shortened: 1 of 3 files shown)


pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -81,67 +113,117 @@ func (h *heartbeatComponent) Start(stop <-chan struct{}) error {
...
+func (h *heartbeatComponent) backoff() time.Duration {
+	if h.failures == 0 {
+		return h.interval
+	}
+	backoff := h.interval << min(h.failures, 16)
+	if backoff <= 0 || backoff > maxHeartbeatBackoff {
+		return maxHeartbeatBackoff
+	}
+	return backoff
+}
```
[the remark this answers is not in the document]
Automaat: **Applied** — the heartbeat backoff cap now uses the larger of the configured interval and the default max cap, so failures never retry faster than the configured interval (808ac75595).

_Generated by Sybra harness_

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -38,11 +57,21 @@ func NewHeartbeatComponent(
...
+		interval: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name: "component_heartbeat",
+			Help: "Summary of Inter CP Heartbeat component interval",
+		}),
+		failures: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "component_heartbeat_failures",
+			Help: "Counter of failed Inter CP heartbeats to the leader by reason",
+		}, []string{"reason"}),
+		leaderChanges: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "component_heartbeat_leader_changes",
```
lobkovilya: How will we use this metric?

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -55,24 +84,27 @@ func NewHeartbeatComponent(
...
-		case <-ticker.C:
+		case <-timer.C:
 			start := core.Now()
-			if !h.heartbeat(ctx, true) {
-				continue
+			if h.heartbeat(ctx, true) {
+				h.metrics.interval.Observe(float64(core.Now().Sub(start).Milliseconds()))
 			}
-			h.metric.Observe(float64(core.Now().Sub(start).Milliseconds()))
+			// back off while the leader is unreachable to avoid a
```
lobkovilya: The comment implies we're getting there only when the leader is unreachable, however since there is no `continue` in the `if` above that's not true.

I'd just remove the comment.

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -81,67 +113,118 @@ func (h *heartbeatComponent) Start(stop <-chan struct{}) error {
...
 		"ready", ready,
 	)
-	if h.leader == nil {
-		if err := h.connectToLeader(ctx); err != nil {
-			if err == ErrNoLeader {
-				heartbeatLog.Info("leader is not yet present in the cluster. No heartbeat to send")
-			} else {
-				heartbeatLog.Error(err, "could not connect to leader")
+
+	newLeader, err := Leader(ctx, h.catalog)
```
lobkovilya: I'm not sure we need to do this, that's `N_CP_INSTANCES * 0.2 req/s` compared to `0 req/s` before in a calm state when there are no CP restarts and leader lost. 

I think you should resolve `Leader` only when `h.failures > 0`.

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -81,67 +113,118 @@ func (h *heartbeatComponent) Start(stop <-chan struct{}) error {
...
-		h.leader = nil
+		log.Error(err, "could not get or create a client to a leader")
+		h.recordFailure(reasonError)
 		return false
 	}
 	resp, err := client.Ping(ctx, h.request)
 	if err != nil {
-		heartbeatLog.Info("could not send a heartbeat to a leader. It's likely due to leader change", "cause", err)
-		h.leader = nil
+		if isConnectivityError(err) {
```
lobkovilya: I'm not sure I understand this split, what's "not connectivity error" in this case? Why do we have to log this differently

pkg/intercp/catalog/heartbeat_component.go
Automaat: **Applied** — removed the comment; `backoff()` already documents its own behavior and this one was misleading since it's called on every tick, not just on unreachable-leader ticks (3893278012).

_Generated by Sybra harness_

pkg/intercp/catalog/heartbeat_component.go
Automaat: **Applied** — the catalog is now only read when the leader is unknown or `h.failures > 0`; in the calm state we keep talking to the already-known leader without hitting the catalog every tick. Also had to explicitly clear `h.leader` when a ping succeeds but the peer responds `Leader: false` (that path has `h.failures == 0`, so without clearing it the next tick would skip resolution and keep talking to the ex-leader) (3893278012).

_Generated by Sybra harness_

pkg/intercp/catalog/heartbeat_component.go
Automaat: "Not a connectivity error" here means anything other than `Unavailable`/`DeadlineExceeded` from the gRPC call — e.g. `PermissionDenied` (mTLS/cert issue), `InvalidArgument`, `Internal`, or a client-construction failure. Those aren't 'the leader is unreachable', they're 'we reached something but it rejected/mishandled the request', which usually points at a config or cert problem rather than a network blip. The split only affects the `reason` label on `component_heartbeat_failures` and the log wording — both still just increment failures/backoff the same way. Keeping them separate lets an operator alert on 'error' failures (likely needs a human) differently from 'unreachable' (likely self-heals). Let me know if you'd rather collapse them — happy to simplify if the extra label isn't pulling its weight.

_Generated by Sybra harness_

pkg/intercp/catalog/heartbeat_component.go
Automaat: I will remove it

pkg/intercp/catalog/heartbeat_component.go
lobkovilya: I don't find this split particularly useful as well as having `reason` label. IMO the only thing we care about is whether the ping got through or not

Automaat: **Applied** — dropped the `reason` label and collapsed `reasonUnreachable`/`reasonError` into a single connectivity-failure path; `component_heartbeat_failures` is now a plain counter and the log line no longer distinguishes error kinds (76eb6eb67f).

_Generated by Sybra harness_

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -81,67 +99,99 @@ func (h *heartbeatComponent) Start(stop <-chan struct{}) error {
...
 	}
 	resp, err := client.Ping(ctx, h.request)
 	if err != nil {
-		heartbeatLog.Info("could not send a heartbeat to a leader. It's likely due to leader change", "cause", err)
-		h.leader = nil
+		h.recordFailure()
+		log.Info(
+			"could not send a heartbeat to a leader, will retry with backoff. This is a connectivity problem, not a leader change",
+			"cause", err,
+			"consecutiveFailures", h.failures,
```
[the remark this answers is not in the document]
Automaat: **Skipped (invalid)** — the current `Ping` path already distinguishes connectivity failures from other errors: `codes.Unavailable` / `codes.DeadlineExceeded` go through `isConnectivityError()` and log the connectivity-specific message, while other errors use the neutral `could not send a heartbeat to a leader` path in `pkg/intercp/catalog/heartbeat_component.go:182-199`. Happy to revisit if I am reading this wrong.

_Generated by Sybra harness_

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -38,11 +49,17 @@ func NewHeartbeatComponent(
...
+	m := &heartbeatMetrics{
+		interval: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name: "component_heartbeat",
+			Help: "Summary of Inter CP Heartbeat component interval",
+		}),
+		failures: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "component_heartbeat_failures",
+			Help: "Counter of failed Inter CP heartbeats to the leader",
+		}),
+	}
```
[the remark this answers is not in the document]
Automaat: **Skipped (invalid)** — current `HEAD` already registers `component_heartbeat_failures` as a `CounterVec` with a `reason` label and also registers `component_heartbeat_leader_changes` in `pkg/intercp/catalog/heartbeat_component.go:65-74`; the regression test asserts both series in `pkg/intercp/catalog/heartbeat_component_test.go:156-176`. Happy to revisit if I am reading this wrong.

_Generated by Sybra harness_

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -81,67 +99,99 @@ func (h *heartbeatComponent) Start(stop <-chan struct{}) error {
...
+	// are currently failing - in the calm state we keep talking to the
+	// already-known leader instead of reading the catalog on every tick.
+	if h.leader == nil || h.failures > 0 {
+		newLeader, err := Leader(ctx, h.catalog)
+		if err != nil {
+			if errors.Is(err, ErrNoLeader) {
+				log.Info("leader is not yet present in the cluster. No heartbeat to send")
 			} else {
-				heartbeatLog.Error(err, "could not connect to leader")
+				log.Error(err, "could not resolve the leader from the catalog")
```
lobkovilya: Should we record failures here as well? If you've set monitors on `component_heartbeat_failures` and then CP instances can't read the catalog (i.e. some RBAC issue) then the alert won't be triggered right?

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -81,67 +99,99 @@ func (h *heartbeatComponent) Start(stop <-chan struct{}) error {
...
 	)
-	if h.leader == nil {
-		if err := h.connectToLeader(ctx); err != nil {
-			if err == ErrNoLeader {
-				heartbeatLog.Info("leader is not yet present in the cluster. No heartbeat to send")
+
+	// The catalog is only consulted when the leader is unknown or heartbeats
+	// are currently failing - in the calm state we keep talking to the
+	// already-known leader instead of reading the catalog on every tick.
+	if h.leader == nil || h.failures > 0 {
```
lobkovilya: Is it possible to rewrite it with fewer levels of nesting? It's a bit hard to follow

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -81,67 +99,99 @@ func (h *heartbeatComponent) Start(stop <-chan struct{}) error {
...
+				}
+				log.Info(
+					"leader has changed. Creating connection to the new leader.",
+					"previousLeaderAddress", previousLeaderAddress,
+					"newLeaderAddress", newLeader.Address,
+				)
+			}
+			h.failures = 0
+		}
+		h.leader = &newLeader
```
lobkovilya: for example here, why do we update `h.leader` outside of `if h.leader == nil || h.leader.Id != newLeader.Id {` statement? it would make sense to update it right when we logged "leader has changed" no?

pkg/intercp/catalog/heartbeat_component.go
Automaat: **Applied** — catalog resolution errors now increment component_heartbeat_failures, with coverage for catalog read failures (fc1271f152).

pkg/intercp/catalog/heartbeat_component.go
Automaat: **Applied** — factored leader resolution into resolveLeader so heartbeat keeps the main loop flatter and easier to follow (fc1271f152).

pkg/intercp/catalog/heartbeat_component.go
Automaat: **Applied** — leader state is now assigned inside the explicit changed/unchanged handling in resolveLeader, next to the change logging path (fc1271f152).

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -189,6 +162,41 @@ func (h *heartbeatComponent) heartbeat(ctx context.Context, ready bool) bool {
...
+		} else {
+			log.Error(err, "could not resolve the leader from the catalog")
+			h.recordFailure()
+		}
+		return false
+	}
+
+	leaderChanged := h.leader == nil || h.leader.Id != newLeader.Id
+	if !leaderChanged {
+		h.leader = &newLeader
```
lobkovilya: Why do we set `h.leader` to `newLeader` when `leaderChanged` is `false`?

pkg/intercp/catalog/heartbeat_component.go
```diff
@@ -189,6 +162,41 @@ func (h *heartbeatComponent) heartbeat(ctx context.Context, ready bool) bool {
 	return true
 }
 
+func (h *heartbeatComponent) resolveLeader(ctx context.Context, log logr.Logger) bool {
```
lobkovilya: what return value `bool` means?

pkg/intercp/catalog/heartbeat_component.go
Automaat: **Applied** — removed the redundant reassignment; when the leader is unchanged `h.leader` already points to an instance with the same Id, so `resolveLeader` now just returns true without touching it (bd8cf79ca9).

pkg/intercp/catalog/heartbeat_component.go
Automaat: **Applied** — added a doc comment: the `bool` reports whether a leader is available and a heartbeat should be sent this tick; false means no leader yet or the catalog read failed, so the tick is skipped (bd8cf79ca9).

Automaat: Follow-up — replaced the doc comment with a self-documenting name: `resolveLeader` → `ensureLeader`, so the call site reads `!h.ensureLeader(...)` (couldn't get a leader → skip this tick). The `bool` meaning now lives in the method name rather than a comment (b092b74915).

lobkovilya (approved):

PR #17084, opened 2026-07-02
ci(actions): re-run build on ci/ label changes

Automaat:
## Motivation

Adding a `ci/*` label (e.g. `ci/run-full-matrix`) to a PR currently has no effect until a new commit is pushed. The build workflow triggers on `opened/synchronize/reopened` only, and reads labels from the event payload — so a label added after those events is never seen. This is especially painful for automation (e.g. backports) that applies labels right after opening a PR.

> Changelog: skip

## Implementation information

- Add `labeled`/`unlabeled` to the `pull_request` trigger. On these events `github.event.pull_request.labels` already reflects the change, so `FULL_MATRIX`/`ALLOW_PUSH`/`BUILD` compute correctly.
- GitHub can't filter label names in `on:`, so gate the root `check` job (and the `always()` `distributions` job) to proceed only for `ci/` labels. Non-`ci/` label changes skip the whole pipeline.
- Fix the concurrency group so a non-`ci/` label event gets a unique throwaway group — otherwise it would share the PR group and cancel an in-flight build before skipping. A `ci/` label event keeps the PR group, so it cancels and replaces the running build with a labelled one.

Note: label events generated by the default `GITHUB_TOKEN` still don't trigger workflows (GitHub's recursion guard); automation using an app token is unaffected.

## Supporting documentation

- https://github.com/orgs/community/discussions/4679 (no native label filtering in `on:`)

### DIFF
--- a/.github/workflows/build-test-distribute.yaml
+++ b/.github/workflows/build-test-distribute.yaml
@@ -4,6 +4,11 @@ on:
     branches: ["master", "release-*", "!*-merge-master"]
     tags: ["*"]
   pull_request:
+    # 'labeled'/'unlabeled' let CI re-run when a ci/ label is added/removed:
+    # its jobs read github.event.pull_request.labels, which the 'opened' event
+    # doesn't yet reflect for labels applied right after creation (e.g. backports).
+    # GitHub can't filter label names in 'on:', so the check job gates on ci/ below.
+    types: [opened, reopened, synchronize, labeled, unlabeled]
     branches: ["master", "release-*"]
   workflow_dispatch: # Allows manual trigger from GitHub Actions UI or via REST call
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-distribute.yaml
```diff
@@ -200,7 +215,11 @@ jobs:
     permissions:
       id-token: write
     timeout-minutes: 10
-    if: ${{ always() }}
+    # always(), but still skip non-ci/ label events (check gates the rest).
+    if: >-
+      ${{ always()
+      && ((github.event.action != 'labeled' && github.event.action != 'unlabeled')
+      || startsWith(github.event.label.name, 'ci/')) }}
```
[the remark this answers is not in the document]
Automaat: **Applied** — replaced the duplicated label/action gate with `needs.check.result != 'skipped'`, so `distributions` now derives its skip condition from `check`'s actual outcome instead of re-implementing the same logic (db9c654b8e).

_Generated by Sybra harness_

.github/workflows/build-test-distribute.yaml
```diff
@@ -18,10 +23,20 @@ env:
   GOLANGCI_LINT_VERSION: "v2.12.2"
   MISE_DISABLE_TOOLS: "golangci-lint,skaffold"
 concurrency:
-  group: ${{ format('{0}-{1}-{2}', github.workflow, github.event_name, github.event_name == 'push' && github.sha || github.event_name == 'pull_request' && github.event.pull_request.number || github.event_name == 'workflow_dispatch' && github.ref_name) }}
+  # A non-ci/ label change must not share the PR's group, or it would cancel a
+  # real in-flight build and then skip (the check job is gated below). Give it a
+  # unique throwaway group. A ci/ label change keeps the PR group so it cancels
+  # and replaces the running build with a labelled (e.g. full-matrix) one.
+  group: ${{ format('{0}-{1}-{2}{3}', github.workflow, github.event_name, github.event_name == 'push' && github.sha || github.event_name == 'pull_request' && github.event.pull_request.number || github.event_name == 'workflow_dispatch' && github.ref_name, ((github.event.action == 'labeled' || github.event.action == 'unlabeled') && !startsWith(github.event.label.name, 'ci/')) && format('-noop-{0}', github.run_id) || '') }}
```
slonka, line 30: nit: this means every label add/remove on any PR to master/release-* now spawns a workflow run that just skips - triage/bot label churn included. Is it acceptable noise or worth a different solution?

.github/workflows/build-test-distribute.yaml
```diff
@@ -18,10 +23,20 @@ env:
...
   cancel-in-progress: ${{ github.event_name == 'push' && false || true }}
 jobs:
   check:
+    # On label events only proceed for ci/ labels; other labels skip the whole
+    # pipeline (check is the root every job depends on). GitHub can't filter
+    # label names in 'on:', so gate here.
+    if: >-
```
slonka, line 37: `check` is the root required status check, and on a non-ci/ label event this run skips it (test/build_publish skip transitively via `needs`). did you confirm branch protection treats the skipped contexts as passing? a skipped job reports success so it should be fine, just want to be sure a label-triggered run can't leave a required context pending and block merge.

.github/workflows/build-test-distribute.yaml
Automaat: Acknowledged — this is accepted noise, not fixed further. GitHub Actions can't filter label names in `on:`, so any label add/remove on a PR to master/release-* triggers a run; the `check` job's `if:` gate makes non-`ci/` runs skip in ~seconds with no compute cost beyond job scheduling. The concurrency group already isolates these no-op runs (`-noop-<run_id>`) so they can't cancel a real build. Given the API limitation, I don't see a lower-noise alternative that doesn't add real complexity (e.g. a separate lightweight workflow polling labels), so I'm leaving this as-is unless you have a preferred approach in mind.

_Generated by Sybra harness_

.github/workflows/build-test-distribute.yaml
Automaat: Confirmed — `check` is skipped via a job-level `if:` (not a workflow-level `paths`/`branches` filter), and per GitHub's docs a job skipped that way reports status `success`, not `pending`. That satisfies required status checks and won't block merging. (The risky case — workflow-level path/branch filtering causing a required check to stay `pending` forever — doesn't apply here since the workflow itself always runs and only the job body is conditional.)

_Generated by Sybra harness_

slonka:
close this and update claude instructions first (set correct labels before opening a PR), reopen if that doesn't help us

Issue #17096, opened 2026-07-02
☂️ Make meshServices.mode=Exclusive the default (3.0)


Automaat:
Track the work to flip the `Mesh` `meshServices.mode` default from `Disabled` to `Exclusive` for 3.0.

## Background

- Enum (`api/mesh/v1alpha1/mesh.proto`): `Disabled=0`, `Everywhere=1`, `ReachableBackends=2`, `Exclusive=3`. No proto/kubebuilder default marker.
- The default is resolved **at read time**, not by a defaulter/webhook. Single source of truth: `api/mesh/v1alpha1/mesh_helpers.go` `MeshServicesMode()` returns `Disabled` when the `MeshServices` block is `nil`.
- proto3 zero-value catch: `Disabled == 0`, so the **only** unset signal is the parent pointer `MeshServices == nil`. `meshServices: {}` / `{mode: Disabled}` are non-nil `Mode=0` and must keep resolving to `Disabled`. Flipping only the nil branch preserves that distinction.

## Two idioms read the mode

- `Spec.MeshServicesMode()` — helper, honors the nil default (correct).
- `Spec.GetMeshServices().GetMode()` — raw proto getter, returns `Disabled(0)` for nil, **bypasses the helper**. These sites will not see the new default and become inconsistent.

## Phases (subissues)

1. Consistency refactor — route all raw `GetMeshServices().GetMode()` sites through the helper. No behavior change; bisectable, lands first.
2. Flip the default — `mesh_helpers.go` nil branch `Disabled` → `Exclusive`.
3. e2e verification & triage of the fallout.
4. Cleanups — validation, `UPGRADE.md`, changelog, generated OpenAPI docs.

## Notes

- Breaking behavior change → target `master` (3.0) only.
- Downstream projects that read the mode should verify they use the helper and re-test; one known downstream reader changes DNS/VIP output for unset meshes (correct, but verify).

PR #17102, opened 2026-07-03
feat(mesh): default nil meshServices to Exclusive

### LINKED ISSUE #17098: meshServices default: flip nil MeshServices from Disabled to Exclusive
Phase 2 of #17096. Depends on phase 1 (helper consistency). The behavioral flip itself.

## Change

- `api/mesh/v1alpha1/mesh_helpers.go` `MeshServicesMode()`: return `Mesh_MeshServices_Exclusive` (instead of `Disabled`) when `MeshServices == nil`.

## Scope / semantics

- Only the **nil** (block entirely absent) branch flips. Explicit `meshServices: {}` and `meshServices: {mode: Disabled}` remain `Disabled` (non-nil `Mode=0`).
- Affects the default mesh (created from an empty spec) and every existing mesh with no `meshServices` block — on upgrade these become `Exclusive` (`kuma.io/service` no longer used for config generation).

## Acceptance

- Unit coverage for: nil → Exclusive, explicit Disabled → Disabled, explicit Everywhere/ReachableBackends unchanged.
- Isolated as its own commit/PR so the behavior change is bisectable.

Automaat:
## Motivation

> Changelog: feat(mesh): default nil meshServices to Exclusive

Phase 2 of the meshServices default flip (parent umbrella: kumahq/kuma#17096, phase 1: kumahq/kuma#17101). While `meshServices.mode` defaults to `Disabled`, the nil (block entirely absent) branch of `MeshServicesMode()` should resolve to `Exclusive` going forward, since the two idioms for reading the mode have already been consolidated behind the helper.

## Implementation information

- `api/mesh/v1alpha1/mesh_helpers.go`: `MeshServicesMode()` now returns `Mesh_MeshServices_Exclusive` when `MeshServices` is nil (instead of `Disabled`).
- Only the nil branch flips. Explicit `meshServices: {}` and `meshServices: {mode: Disabled}` remain non-nil `Mode=0` and keep resolving to `Disabled`.
- Affects the default mesh and every existing mesh with no `meshServices` block: on upgrade these become `Exclusive` (`kuma.io/service` no longer used for config generation).
- Added unit coverage in `api/mesh/v1alpha1/mesh_helpers_test.go` for: nil -> Exclusive, explicit Disabled -> Disabled, explicit Everywhere/ReachableBackends/Exclusive unchanged.

## Supporting documentation

Closes kumahq/kuma#17098


lobkovilya (commented):
Can we remove `meshServices` field from `Mesh` altogether?

Automaat:
## e2e status after defaulting `meshServices` to `Exclusive`

Most suites are migrated and green (unit, universal, kubernetes, default). For multizone, 18/20 failing specs are fixed and verified locally on a k3d multizone cluster. Two specs remain and look like genuine cross-zone behavior questions rather than test-only migrations — flagging for input on the intended `Exclusive` semantics.

### 1. `MeshTrafficPermission` — "should allow the traffic to the external service through the egress"

- `MeshTrafficPermission` does not accept `targetRef.kind: MeshExternalService` (validator allows only `Mesh`, `MeshSubset`, `MeshService`, `MeshServiceSubset`, `Dataplane`), so the original `mtp-5` targeting the external service can't be expressed.
- With no MTP, traffic to the external service is allowed by default (returns 200), so the "blocked-then-allowed" flow no longer holds.
- Reaching a `MeshExternalService` cross-zone (endpoint in a universal zone, client in a k8s zone, through ZoneEgress) did not produce a resolvable `*.extsvc.mesh.local` cluster in the client zone — tried both zone-origin and global-origin MES.

Question: under `Exclusive`, how is access to a cross-zone `MeshExternalService` meant to be controlled and reached (which resource gates it, and how should the `extsvc` hostname resolve in the client zone)?

### 2. ZoneEgress Internal Services — "should access internal service behind k8s zoneingress through zoneegress"

- Cross-zone internal `MeshService` traffic no longer flows through the client-zone ZoneEgress: the client cluster resolves a direct, healthy endpoint, and the client-zone egress has zero service clusters/stats for the service (`kri_msvc_…` appears on the client, not on the egress). Connectivity itself works.

Question: under `Exclusive` with `routing.zoneEgress: true`, is cross-zone internal `MeshService` traffic still expected to traverse the local ZoneEgress? If not, the test's egress-stat assertion should be replaced (e.g. assert on the client-side cluster).

### Note on MeshMultiZoneService (resolved)

The cross-zone `MeshHTTPRoute` case was migrated to a `MeshMultiZoneService`. Worth documenting for others hitting this: MZMS is locality-aware by default, so a client only sees its local zone until a `MeshLoadBalancingStrategy` with `localityAwareness.disabled: true` is applied; cross-zone endpoints are reached through ZoneEgress, so the case only applies in the zone-egress variant.

test/e2e_env/multizone/meshhttproute/test.go
```diff
@@ -93,28 +93,32 @@ name: route-alias-test-server
 mesh: %s
 spec:
   targetRef:
-    kind: MeshService
-    name: demo-client
+    kind: MeshSubset
```
Automaat, line 96: **Reusing a deprecated selector kind.** The top-level `targetRef` moved from `MeshService` (deprecated as a top-level selector) to `MeshSubset` — which is *also* deprecated (→ `Dataplane` + labels). The v3 direction restricts the top-level `targetRef` to `Mesh` and `Dataplane` only, so this trades one deprecated kind for another.

`Dataplane` is already an accepted top-level kind for MeshHTTPRoute (`validation.go:41`), and the **k8s** route tests in this same PR migrated their top-level `targetRef` to `kind: Mesh`. Suggest `kind: Dataplane` (label selection) or `kind: Mesh` here to avoid re-adopting a kind the migration is meant to shed. Same applies to `meshtcproute/test.go`.

test/e2e_env/multizone/meshtcproute/test.go
```diff
@@ -89,27 +89,31 @@ name: route-1
 mesh: %s
 spec:
   targetRef:
-    kind: MeshService
-    name: demo-client
+    kind: MeshSubset
```
Automaat, line 92: **Reusing a deprecated selector kind** (same as `meshhttproute/test.go`). Top-level `targetRef` went `MeshService` → `MeshSubset`, both deprecated top-level selectors. Target v3 shape is `Mesh`/`Dataplane` only. Prefer `kind: Dataplane` (labels) or `kind: Mesh`.

Automaat (commented):
Reviewed the `meshServices` Exclusive-default flip against the v3 deprecated-features list. The flip itself is correct and complete — every production read goes through `MeshServicesMode()`, so the one-line change flips behavior globally with no bypass — and the bulk of the test migration is on-target (kuma.io/service VIP DNS → MeshService hostnames, `ExternalService`→`MeshExternalService`, `MeshGatewayRoute`→`MeshHTTPRoute`, `MeshServiceSubset`→`MeshService`+labels, `reachableServices`→`reachableBackends`, cross-zone locality via `MeshMultiZoneService`+`MeshLoadBalancingStrategy`). The Disabled-mode opt-outs elsewhere are all justified and well-commented.

Inline comments below: one place that re-adopts a deprecated selector kind (worth changing before merge), plus two lower-priority notes. Not blocking.

test/e2e_env/multizone/meshtrafficpermission/meshtrafficpermission.go
```diff
@@ -246,7 +249,7 @@ name: mtp-5
 mesh: mtp-test
 spec:
  targetRef:
-   kind: MeshService
+   kind: MeshExternalService
```
Automaat: **Two low-priority deprecation notes on this MTP block (not blocking):**

1. **Top-level targetRef kind.** `MeshExternalService` is the right destination type here today, but the top-level `targetRef` is slated to narrow to `Mesh`/`Dataplane` only. No better option exists for MTP-to-external-service yet, so fine as-is — flagging for the eventual top-level-targetRef cleanup.
2. **Deprecated `from[]`.** This MTP (and the others in this file) still use the `from[]` model, deprecated in favor of `rules[]` with SPIFFE-based matches (+ MeshIdentity); removal targeted for 3.0. Out of scope for the default flip (separate umbrella phase) and `from[]` stays valid until then — noting only since this block is edited here.

lukidzi (approved):
It seems ok, I have one question here, what if user has a MeshService with nil/default. Do you know if we store in a DB empty value or we store a default? I wonder if user has a Mesh with a default which is disabled, and runs the same on this control-plane won't it break anything or maybe it's intentional

Automaat:
@lukidzi this can be breaking change as this is part of 3.0 work. There are also few followups and we will remove this field entirely from Mesh resource before we release 3.0

PR #17127, opened 2026-07-06
fix(xds): drop resourceVersion from Dataplane hash

### LINKED ISSUE #17062: xDS recomputed mesh-wide when any hashed resource's resourceVersion bumps (incl. xDS-irrelevant status writes); CP CPU scales with churn

Automaat:
## Motivation

`DataplaneResource.Hash()` folded `meta.GetVersion()` (the Kubernetes `resourceVersion`) into the mesh-context hash. Since `resourceVersion` changes on every write to a Dataplane object — including status and annotation-only updates that don't affect xDS — any such write forced a mesh-wide xDS recomputation for all proxies in the mesh.

## Implementation information

- `Hash()` now uses `core_model.HashMetaIdentity(d)` (drops `resourceVersion`) instead of `core_model.HashMeta(d)`.
- Labels are hashed explicitly via a new `writeSortedLabels` helper (sorted, length-prefixed) since they affect policy matching and must still invalidate the hash when changed.
- The networking address/advertised-address fields are replaced by a deterministic proto marshal of the full spec, so the hash still reacts to any spec change without needing to enumerate fields by hand.
- Added unit tests covering hash stability across resourceVersion/status changes and hash changes on spec/label changes, plus a mesh-context builder test for convergence behavior.

Closes https://github.com/kumahq/kuma/issues/17062

_Generated by Sybra harness_


Automaat:
### Benchmark finding: this fix is necessary but not sufficient for the `MatchedPolicies` cache (#17017)

I benchmarked the `MatchedPolicies` LRU cache (#17017) on a GKE cluster in the new model (MeshService `Exclusive` + MeshIdentity + unified naming + inbound-tags-disabled), 500–1000 dataplanes (stable pool + a flapping pool simulating cron-style churn), 200 MeshTrafficPermissions + 200+ MeshServices. Under dataplane flapping the cache gets **~0% hit rate** and *increases* CP CPU by **~7–12%** vs. disabled (plus memory), i.e. it's currently counterproductive for exactly the churny workloads it targets.

**Root cause (why #17127 alone doesn't help the cache):**

- The cache key is `MeshContext.PolicyMatchingHash` → `computePolicyMatchingHash` → `ResourceListHash` → `Hash()` → `HashMeta()`, which folds in `meta.GetVersion()` (resourceVersion) for every `AffectsPolicyMatching: true` type.
- **`MeshService` is `AffectsPolicyMatching: true` and has no custom `Hash()`**, so it goes through `HashMeta` (with version). The MeshService `StatusUpdater` runs on a **5s ticker** and rewrites MeshService whenever `Status.DataplaneProxies.{Connected,Healthy,Total}` change — essentially every tick under dataplane churn. Each write bumps `resourceVersion` → `PolicyMatchingHash` flips → the whole LRU is invalidated. Hence ~0% hits.
- **`Dataplane` is `AffectsPolicyMatching: false`** (the roster is deliberately excluded from the cache key), so this PR's change to `DataplaneResource.Hash()` has **no effect on the cache key**. It does fix the *full* `MeshContext.Hash` path (mesh-wide regen from DP status/annotation/managedFields writes), which is valuable on its own — just orthogonal to the cache.

**What else needs to be done:**

Apply the same "drop `resourceVersion`, hash identity+labels+spec" treatment to the resources that feed `PolicyMatchingHash` — at minimum `MeshService`, and likely the other status-churning `AffectsPolicyMatching: true` new-style resources (MeshIdentity, MeshTrust, Workload). Two ways:

1. **Central (recommended for the cache):** make `computePolicyMatchingHash` content-based (identity+labels+spec, version/status-insensitive) in one place. Matching is a pure function of spec+labels and never reads status, so this is safe for *all* matching types at once.
2. **Per-type (this PR's pattern):** give `MeshService.Hash()` a version-free implementation like the one added here for Dataplane. Because `ResourceListHash` routes through the per-type `Hash()`, this fixes **both** the cache **and** cuts the mesh-wide regen that MeshService's 5s status writes cause today.

Note: a *blanket* global version-drop on the full `MeshContext.Hash` is not safe — some resources' status (VIPs, endpoints, TLS readiness) legitimately feeds xDS, so that path should stay per-type/opt-in (as this PR does). The cache key is the part that can be changed globally.

Caveats: use `proto.MarshalOptions{Deterministic:true}` for proto specs (not protojson) to avoid spurious misses; consider memoizing the content-hash by `resourceVersion` so the mesh-build path doesn't re-marshal every spec on every rebuild.

I'm validating a prototype (spec-hash for `policyMatchingHash`) on the same cluster and will follow up with the before/after hit-rate + CPU numbers.

pkg/xds/context/mesh_context_builder.go
```diff
@@ -586,7 +586,7 @@ func (m *meshContextBuilder) hash(globalContext *GlobalContext, baseMeshContext
 	_, _ = hasher.Write(globalContext.hash)
 	_, _ = hasher.Write(baseMeshContext.hash)
 	for _, resType := range managedTypes {
-		_, _ = hasher.Write(core_model.ResourceListHash(resources.MeshLocalResources[resType]))
+		_, _ = hasher.Write(resourceListXDSHash(resources.MeshLocalResources[resType]))
```
slonka, line 525: heads up: this branch is ~26 commits behind master, which is before `computePolicyMatchingHash` landed - so you can't see it here. that function hashes its resources with `core_model.ResourceListHash` (the version-inclusive `Hash()`), and MeshService has `AffectsPolicyMatching: true`, so the resourceVersion churn you're fixing leaks back into `PolicyMatchingHash` after this merges.

in practice the early return at the outer hash gate (`if newHash == latestMeshCtx.Hash { return }`) short-circuits before `computePolicyMatchingHash` runs, so it's not a live 5s-churn bug - but the two paths end up inconsistent. can you rebase on master and switch those `ResourceListHash` calls to the XDS hash too?

pkg/core/resources/apis/mesh/dataplane_helpers.go
```diff
@@ -168,11 +168,38 @@ func (d *DataplaneResource) AdminPort(defaultAdminPort uint32) uint32 {
...
 	hasher := fnv.New128a()
-	_, _ = hasher.Write(core_model.HashMeta(d))
-	_, _ = hasher.Write([]byte(d.Spec.GetNetworking().GetAddress()))
-	_, _ = hasher.Write([]byte(d.Spec.GetNetworking().GetAdvertisedAddress()))
+	_, _ = hasher.Write(core_model.HashMetaIdentity(d))
+	if includeVersion {
+		_, _ = hasher.Write([]byte(d.GetMeta().GetVersion()))
+	}
+	core_model.WriteSortedLabels(hasher, d.GetMeta().GetLabels())
+	specBytes, err := proto.MarshalOptions{Deterministic: true}.Marshal(d.Spec)
```
slonka, line 194: `hash()` now proto-marshals the full spec on every call. `matchers.BuildCacheKey` calls `dpp.Hash()` per-dp per-policy-type on every reconcile, which was previously just two address string writes. nice that you added a benchmark - did the matchers path show up as a regression on big meshes? if it does, worth caching the marshaled bytes rather than re-marshaling per key.

pkg/core/resources/apis/mesh/dataplane_helpers_test.go
```diff
@@ -749,4 +751,150 @@ var _ = Describe("Dataplane", func() {
...
+
+			changed := newDataplane()
+			changed.Meta.(*test_model.ResourceMeta).Version = "2"
+
+			Expect(changed.XDSHash()).To(Equal(originalHash))
+		})
+	})
 })
+
+func BenchmarkDataplaneHash(b *testing.B) {
```
slonka, line 878: do we have numbers on how this performed before this change? (control vs master branch)

pkg/xds/context/mesh_context_builder_test.go
```diff
@@ -141,4 +141,98 @@ var _ = Describe("hash", func() {
...
+status:
+  vips:
+  - ip: 10.0.1.1
+`)).To(Succeed())
+
+		before, err := meshContextBuilder.BuildIfChanged(context.Background(), "mesh-1", nil)
+		Expect(err).ToNot(HaveOccurred())
+
+		// when the MeshService is written again with the same spec/status but updated proxy stats
+		Expect(test_store.LoadResources(context.Background(), resourceStore, `
```
Automaat, line 284: **issue:** This test never updates `DataplaneProxies`: `LoadResources` copies only spec when the resource already exists, so the status block is discarded. It passes even if `XDSHash` still hashes proxy stats; update status directly.

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers_test.go
```diff
@@ -0,0 +1,126 @@
...
+		Expect(changed.Hash()).NotTo(Equal(originalHash))
+	})
+
+	It("changes when only the DataplaneProxies stats change", func() {
+		original := newMeshService()
+		originalHash := original.Hash()
+
+		changed := newMeshService()
+		changed.Status.DataplaneProxies = api.DataplaneProxies{Connected: 3, Healthy: 2, Total: 3}
+		changed.Meta.(*test_model.ResourceMeta).Version = "2"
```
Automaat, line 46: **issue:** `Hash()` still changes here because the test also bumps `Version`; it does not prove `DataplaneProxies` is part of the content hash. Remove the version bump or rename the case so the test matches the contract.

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go
```diff
@@ -40,6 +45,59 @@ func (m *MeshServiceResource) IsLocalMeshService() bool {
...
+	}
+	core_model.WriteSortedLabels(hasher, t.GetMeta().GetLabels())
+	writeJSON(hasher, t.Spec)
+	writeJSON(hasher, struct {
+		Addresses          []hostnamegenerator_api.Address
+		VIPs               []VIP
+		TLS                TLS
+		HostnameGenerators []hostnamegenerator_api.HostnameGeneratorStatus
+	}{
+		Addresses:          t.Status.Addresses,
```
Automaat: **issue:** `core_model.Hash(&MeshServiceResource{Meta: meta, Spec: &MeshService{}})` now panics because `Hash()` dereferences nil `Status`. Keep the previous `HashMeta` behavior for zero-value resources by guarding nil status.

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go
```diff
@@ -40,6 +45,59 @@ func (m *MeshServiceResource) IsLocalMeshService() bool {
...
+	if includeVersion {
+		_, _ = hasher.Write([]byte(t.GetMeta().GetVersion()))
+	}
+	core_model.WriteSortedLabels(hasher, t.GetMeta().GetLabels())
+	writeJSON(hasher, t.Spec)
+	writeJSON(hasher, struct {
+		Addresses          []hostnamegenerator_api.Address
+		VIPs               []VIP
+		TLS                TLS
+		HostnameGenerators []hostnamegenerator_api.HostnameGeneratorStatus
```
slonka: if we add a new field, how do we make sure this stays up to date? should we flip this to exclude list?

pkg/xds/context/mesh_context_builder.go
Automaat: **Applied** - switched PolicyMatchingHash over to the XDS hashes as well, and added rebuild coverage for Dataplane resourceVersion-only writes plus MeshService DataplaneProxies updates (c214041e7f).

_Generated by Sybra harness_

pkg/core/resources/apis/mesh/dataplane_helpers.go
Automaat: **Skipped (questionable)** - I measured the dataplane hash on this host: this branch is ~1.52 us / 624 B / 32 allocs, while an equivalent temporary benchmark on origin/master is ~0.186 us / 96 B / 8 allocs. I do not have an in-tree big-mesh MatchedPolicies benchmark to prove the matcher path end-to-end, and the remaining optimization here looks like a separate memoization/cache-key design rather than a correctness fix in this PR. Leaving the thread open for your call.

_Generated by Sybra harness_

pkg/core/resources/apis/mesh/dataplane_helpers_test.go
Automaat: **Skipped (questionable)** - I collected the control numbers on the same machine: this branch benchmarks at ~1.52 us / 624 B / 32 allocs, while origin/master with the same temporary benchmark is ~0.186 us / 96 B / 8 allocs. I have not folded a performance redesign into this correctness-review pass, so I am leaving the thread open for your call.

_Generated by Sybra harness_

pkg/xds/context/mesh_context_builder_test.go
Automaat: **Applied** - updated the MeshService status-only recompute test to load the stored MeshService and mutate Status.DataplaneProxies directly before Update, so it now exercises the real status path (c214041e7f).

_Generated by Sybra harness_

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers_test.go
Automaat: **Applied** - renamed that MeshService Hash test so it now matches the contract: the change there is the resourceVersion bump alongside the DataplaneProxies update, not DataplaneProxies participating in the content hash (c214041e7f).

_Generated by Sybra harness_

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go
Automaat: **Applied** - guarded nil MeshService spec/status in the hash path and added a parity test that nil fields hash the same as empty values, so the zero-value case no longer panics (c214041e7f).

_Generated by Sybra harness_

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go
Automaat: **Skipped (questionable)** - I kept the XDS hash as an allowlist of xDS-consumed fields on purpose: an exclude list would make any future non-xDS status field silently start invalidating mesh-wide xDS again. The nil/empty parity test now covers the zero-value maintenance footgun; leaving the broader allowlist-vs-excludelist tradeoff open for your call.

_Generated by Sybra harness_

slonka (commented):
Review Verdict: CLEAN

Traced the hash refactor across `DataplaneResource`, `MeshServiceResource`, `MeshExternalServiceResource`, `MeshIdentityResource`, `MeshMultiZoneServiceResource`, `MeshOpenTelemetryBackendResource`, `MeshTrustResource`, the shared `WriteSortedLabels`/`WriteDeterministicJSON` helpers, `resourceListXDSHash` in `pkg/xds/context/hash.go`, and the `hashMeshIdentity` watchdog change.

For each type, checked that every field excluded from `XDSHash()` is genuinely not read by xDS generation:
- `MeshService.DataplaneProxies` - only written/read by the status updater (`pkg/core/resources/apis/meshservice/status/updater.go`), never by xDS generation.
- `MeshIdentity`/`MeshTrust`/`MeshOpenTelemetryBackend` status (conditions) - not read outside their own controllers; `MeshIdentity` readiness is handled separately via the explicit `IsInitialized()` byte appended in `hashMeshIdentity` (`pkg/xds/sync/dataplane_watchdog.go`), so excluding it from the mesh-wide `XDSHash()` doesn't lose that signal.
- `MeshService`/`MeshMultiZoneService` VIPs/Addresses/TLS/HostnameGenerators are correctly retained since they feed cluster/endpoint generation directly.

`WriteSortedLabels` length-prefixes keys/values, so the boundary-collision case (`{"a":"bc"}` vs `{"ab":"c"}`) is handled and covered by a test. The switch from `HashMeta`/field-enumeration to `HashMetaIdentity` + full deterministic proto marshal for Dataplane spec is a strict superset of what was hashed before, so no spec-change class goes undetected.

No blocking or narrow-edge-case findings survived scrutiny. Leaving approval to a human reviewer given the size and blast radius of this change (touches mesh-wide xDS invalidation for 7 resource types).

_Generated by Sybra harness_

slonka (commented):
**Adversarial review — CLEAN.**

Traced the full hash-propagation chain for the new `XDSHash()` mechanism across `DataplaneResource`, `MeshServiceResource`, `MeshExternalServiceResource`, `MeshMultiZoneServiceResource`, `MeshIdentityResource`, `MeshTrustResource`, and `MeshOpenTelemetryBackendResource`.

Specifically checked for the failure mode where a resource's `XDSHash()` is added but never actually reached by the mesh-context invalidation gate:

- `pkg/xds/context/hash.go`'s `resourceListXDSHash`/`resourceXDSHash` is used consistently by both `mesh_context_builder.go`'s main hash + `computePolicyMatchingHash`, and by `pkg/xds/context/resources.go`'s `ResourceMap.Hash()` (also patched in this PR) — so `MeshService`/`MeshExternalService`/`MeshMultiZoneService`, which are all `IsDestination: true` and get routed into `BaseMeshContext.ResourceMap` rather than `managedTypes`, still get the XDS-stable hash instead of silently falling back to the old version-sensitive one. (Initially suspected a gap here since `resources.go` didn't show up in the first truncated file list I pulled — confirmed against the actual PR branch that it's patched.)
- `MeshIdentity`'s `XDSHash()` drops `Status` entirely (readiness conditions), which would otherwise mean a not-ready→ready flip never triggers a resync. `pkg/xds/sync/dataplane_watchdog.go`'s `hashMeshIdentity` explicitly re-adds a 1-byte "is initialized" flag on top of `XDSHash()` to cover exactly that case, with a test (`pkg/xds/sync/hash_test.go`) proving both the ignore- and the trigger-case.
- Per-resource `Status` exclusions (e.g. `MeshTrustStatus.Origin`, `MeshOpenTelemetryBackendStatus.Conditions`, `MeshIdentityStatus.Conditions`) all check out against their actual struct definitions — no XDS-relevant field (VIPs, addresses, TLS, hostname generators) is dropped, only genuine bookkeeping.

Two very minor, non-blocking observations, not worth a required change:
- `WriteDeterministicJSON`/`writeJSON`'s error fallback (`fmt.Fprintf(hasher, "%+v", v)`) would embed a raw pointer address for structs with pointer fields (e.g. `Origin.KRI`) if `json.Marshal` ever failed — non-deterministic if it were ever hit, but unreachable for any legitimate content these types can hold today.
- `mesh_context_builder.go`'s `BuildIfChanged` now builds the `dataplanesByName` map before the early-return cache-hit check instead of after — harmless reordering, just a wasted allocation on the cache-hit path.

No approval implied — leaving that to a human reviewer.

_Generated by Sybra harness_

slonka (commented):
Adversarial review (Sybra harness): no blocking correctness issues found.

Traced the actual wiring end to end against a local checkout of the PR branch, not just the diff:
- `resourceXDSHash` in `pkg/xds/context/hash.go` correctly dispatches to `XDSHash()` when a resource implements it, falling back to the old version-inclusive `core_model.Hash()` otherwise.
- `ResourceMap.Hash()` (`pkg/xds/context/resources.go`) and `meshContextBuilder.hash`/`computePolicyMatchingHash` (`mesh_context_builder.go`) all route through `resourceListXDSHash`, so the mesh-wide hash, the policy-matching hash, and the base/global mesh context hashes are all version-aware consistently - not just the Dataplane path called out in the description.
- Per-type `XDSHash()` implementations (MeshService, MeshExternalService, MeshIdentity, MeshMultiZoneService, MeshTrust, MeshOpenTelemetryBackend) correctly exclude only status fields that are genuinely bookkeeping (conditions, DataplaneProxies counters, origin) while keeping fields that feed xDS generation (VIPs, hostnames, TLS, matched backends) - checked each status struct to confirm nothing xDS-relevant got dropped.
- `dataplane_watchdog.go`'s `hashMeshIdentity` correctly adds back an explicit initialization bit on top of `XDSHash()`, since per-DP identity delivery gating needs that signal even though it's excluded from the mesh-wide hash.
- `WriteSortedLabels`/`HashMetaIdentity` length-prefix keys/values, so there's no key/value boundary collision, and there's a test for exactly that (`{"a":"bc"}` vs `{"ab":"c"}`).
- The regression itself is covered by real end-to-end tests in `mesh_context_builder_test.go` (`BuildIfChanged`) for both the Dataplane resourceVersion-only case and the MeshService DataplaneProxies-only case described in the issue, plus unit tests per resource type.

Two non-blocking notes:
- `dataplane_helpers.go`'s `Hash()`/`XDSHash()` now hash the full marshaled `Spec` instead of just `Networking.Address`/`AdvertisedAddress`. That's a much wider blast radius than the old hash (e.g. tag/service changes on a Dataplane weren't previously reflected in `Hash()` at all) - looks intentional and is covered by tests, but worth calling out explicitly in the PR description as a behavior change beyond "stop hashing resourceVersion", since anyone relying on the old narrow hash semantics elsewhere would be affected.
- The `test(e2e): use Kubernetes DNS in CNI race test` commit (`test/e2e/cni/taint_controller_race.go`) looks unrelated to the hash fix - possibly worth splitting into its own PR for a cleaner history, not a correctness concern.

Nothing here blocks merging from my read; leaving approval to a human reviewer.

_Generated by Sybra harness_

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/hash.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// XDSHash returns the MeshMultiZoneService hash used by xDS invalidation.
+// Match conditions are controller bookkeeping; hostnames, VIPs, and matched
+// MeshServices directly affect outbound and endpoint generation.
+func (m *MeshMultiZoneServiceResource) XDSHash() []byte {
+	return m.hash(false, true)
+}
+
+func (m *MeshMultiZoneServiceResource) hash(includeVersion bool, xdsOnly bool) []byte {
```
slonka: two things:

1. same allowlist trap as MeshService - `xdsOnly` keeps Addresses/VIPs/MeshServices/HostnameGenerators and drops `Conditions`
2. nitpick: `hash(includeVersion, xdsOnly)` - the two bools are always opposite - collapse it to match?

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go
```diff
@@ -40,6 +45,67 @@ func (m *MeshServiceResource) IsLocalMeshService() bool {
...
+		TLS:                status.TLS,
+		HostnameGenerators: status.HostnameGenerators,
+	})
+	return hasher.Sum(nil)
+}
+
+// writeJSON writes a deterministic JSON encoding of v into hasher.
+// encoding/json sorts map keys, so this is stable regardless of map
+// iteration order.
+func writeJSON(hasher hash.Hash, v any) {
```
slonka: this is identical to the new `core_model.WriteDeterministicJSON` the other hash.go files use - looks like a leftover local copy. drop it and use the shared one?

pkg/core/resources/model/resource.go
```diff
@@ -125,6 +128,42 @@ func writeMetaIdentity(hasher hash.Hash, r Resource) {
...
+
+	for _, k := range keys {
+		writeLenPrefixed(k)
+		writeLenPrefixed(labels[k])
+	}
+}
+
+// WriteDeterministicJSON writes a stable JSON encoding of v into hasher.
+// encoding/json sorts map keys, so the resulting bytes are deterministic.
+func WriteDeterministicJSON(hasher hash.Hash, v any) {
```
slonka, line 156: There are now 3 different status strategies across the 6 types - drop status entirely (identity/otel/trust), hash full status (mesh-external-service), allowlist a subset (meshservice/mmzs). There's no single rule, so the next person adding a type or a status field has to re-derive the correctness argument and can get it wrong (= stale xDS).

The actual invariant across all of them is: drop only bookkeeping fields (`Conditions`, `DataplaneProxies`), keep everything else. Could we express that once - a shared helper that hashes identity+version+labels+spec + full-status-minus-ignored, plus a reflection guard test that fails when a new status field shows up unclassified? that kills the allowlist trap and folds the 6 near-identical preambles into one.

Should this just be a follow up issue?

xrel https://github.com/kumahq/kuma/pull/16083

this comment was partly generated by an LLM

pkg/xds/sync/dataplane_watchdog.go
```diff
@@ -254,7 +254,20 @@ func (d *DataplaneWatchdog) syncOtelStatus(backends *core_xds.OtelPipeBackends)
 func hashMeshIdentity(identity *meshidentity_api.MeshIdentityResource) []byte {
 	hasher := fnv.New128a()
 	if identity != nil {
-		_, _ = hasher.Write(core_model.Hash(identity))
+		if xdsHasher, ok := any(identity).(interface{ XDSHash() []byte }); ok {
```
slonka: `identity` is statically `*MeshIdentityResource`, which implements `XDSHash()`, so this assertion is always true and the `else` branch is dead - can just be `hasher.Write(identity.XDSHash())`.

separately: nice that the `IsInitialized()` bit re-adds the one status transition that matters after XDSHash drops status. but that's the *only* reason dropping MeshIdentity status is safe - worth a note on `meshidentity/.../hash.go` pointing here, so if someone starts reading another status field in xDS they know both spots need updating.

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go
slonka: > an exclude list would make any future non-xDS status field silently start invalidating mesh-wide xDS again

I think it's a decision between potential churn vs potential bug - I think correctness is more important

slonka (dismissed):

pkg/core/resources/apis/mesh/dataplane_helpers.go
Automaat: No code change in this review pass. I rechecked the local micro-benchmark delta I had from the earlier thread work: this branch was ~1.52 us / 624 B / 32 allocs versus ~0.186 us / 96 B / 8 allocs on origin/master for the temporary dataplane-hash benchmark, but I still do not have an in-tree big-mesh MatchedPolicies benchmark that would justify redesigning BuildCacheKey or adding memoization inside this correctness PR. Keeping this as follow-up performance work rather than folding a cache-key redesign into 9761bb5785.

_Generated by Sybra harness_

pkg/core/resources/apis/mesh/dataplane_helpers_test.go
Automaat: Collected earlier on the same host with a temporary control benchmark: this branch was ~1.52 us / 624 B / 32 allocs, versus ~0.186 us / 96 B / 8 allocs on origin/master. I have not turned that into a matcher-path redesign in this fix-review pass because I still cannot prove the end-to-end MatchedPolicies impact in-tree; keeping 9761bb5785 focused on the correctness comments below.

_Generated by Sybra harness_

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go
Automaat: **Applied** - switched MeshService.XDSHash() away from the fragile allowlist and now hash the full status with only DataplaneProxies zeroed on the xDS path. That keeps the 5s proxy-stat churn ignored, but new xDS-relevant status fields are now picked up by default. I also added coverage that an address/hostname status change still invalidates XDSHash() (9761bb5785).

_Generated by Sybra harness_

pkg/core/resources/apis/meshmultizoneservice/api/v1alpha1/hash.go
Automaat: **Applied** - MeshMultiZoneService.XDSHash() now hashes the full status with only Conditions zeroed, so the correctness concern is handled without the allowlist trap, and I collapsed the old hash(includeVersion, xdsOnly) split by giving the xDS path its own helper (xdsHash) in 9761bb5785.

_Generated by Sybra harness_

pkg/core/resources/apis/meshservice/api/v1alpha1/helpers.go
Automaat: **Applied** - dropped the leftover local writeJSON helper in MeshService and now reuse the shared core_model.WriteDeterministicJSON path for both spec and status hashing there as well (9761bb5785).

_Generated by Sybra harness_

pkg/core/resources/model/resource.go
Automaat: Partially addressed in this PR fixup: the two remaining allowlist cases (MeshService and MeshMultiZoneService) now hash full status minus the bookkeeping fields (DataplaneProxies / Conditions), which removes the concrete correctness trap from this diff. I did not factor the broader cross-type helper / reflection-guard idea into 9761bb5785 because that is structural follow-up rather than a narrowly justified review fix; if we want that deduplication, I would rather do it in a dedicated cleanup PR.

_Generated by Sybra harness_

pkg/xds/sync/dataplane_watchdog.go
Automaat: **Applied** - removed the dead type-assertion branch and call identity.XDSHash() directly now. I also added the note in meshidentity/.../hash.go pointing back to hashMeshIdentity, so the readiness coupling is documented in both places (9761bb5785).

_Generated by Sybra harness_

slonka (approved):

PR #17133, opened 2026-07-07
fix(meshloadbalancingstrategy): reject invalid crossZone

### LINKED ISSUE #16051: add crossZone validation / documentation to explain that it only makes sense when using MeshMultizoneService
### Kuma Version

latest

### Describe the bug

MeshLoadBalancingStrategy `crossZone` only makes sense when using MeshMultizoneService, ideally the validation should reflect that. The OpenAPI description for this field should also be updated to explain this.

Follow up: Are there any other spec fields like this where they are only valid when being used with a certain type of targetRef?

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

ahfoysal:
## Motivation
Fix #16051

## Implementation information
Validate that crossZone is only used with MeshMultiZoneService targets and refresh generated API docs.

## Supporting documentation
- Changelog: skip


pkg/plugins/policies/meshloadbalancingstrategy/api/v1alpha1/validator.go
```diff
@@ -151,8 +151,11 @@ func validateCrossZone(crossZone *CrossZone, to To) validators.ValidationError {
 	if crossZone == nil {
 		return verr
 	}
-	if to.TargetRef.Kind == common_api.MeshService && (pointer.Deref(to.TargetRef.SectionName) != "" || len(pointer.Deref(to.TargetRef.Labels)) > 0) {
-		verr.AddViolationAt(validators.Root(), fmt.Sprintf("%s: MeshService traffic is local", validators.MustNotBeSet))
+	if to.TargetRef.Kind != common_api.MeshMultiZoneService {
```
Automaat: **blocking:** This rejects `crossZone` for every `to.targetRef.kind` except `MeshMultiZoneService`, but `MeshService`+`crossZone` is a supported, e2e-tested pattern (egress-based cross-zone failover). It breaks three active multizone e2e suites (`LocalityAwareLB`, `LocalityAwareLBGateway`, `LocalityAwareLBEgress`) that apply `to: MeshService`+`crossZone` and are not updated here, and invalidates existing policies with no migration. Narrow the rule to keep `MeshService`/`Mesh` valid, or if removal is intended, update the e2e suites, add an UPGRADE note, mark it breaking, and drop `Changelog: skip`.

PR #17162, opened 2026-07-07
feat(labels): compute `kuma.io/display-name` for all resources

lobkovilya:
## Motivation

This is the first part of dismantling https://github.com/kumahq/kuma/pull/16852 into smaller chunks.

Today, `kuma.io/display-name` is set only when it's different from the `metadata.name`. This PR changes that and systematically sets `kuma.io/display-name` in `Compute`.

## Implementation information

- `resource_labels.Compute` takes a new `displayName` argument and force-sets `kuma.io/display-name` to it, threaded through all 9 call sites (api-server, dataplane manager, defaults/mesh, pod/gateway converters, and the k8s defaulter — the defaulter derives the short name from the `K8sNameComponent` name extension).
- `DisplayName` is added to `AllComputedLabels` (keep in sync with the downstream shared-speakeasy kumalabels list referenced in `registry.go`).

## Supporting documentation

Updates https://github.com/kumahq/kuma/issues/16049

> Changelog: control plane now computes the kuma.io/display-name label for all resources


pkg/core/resources/labels/compute.go
```diff
@@ -111,6 +121,11 @@ func Compute(
 		labels = maps.Clone(existingLabels)
 	}
 
+	// Do not rewrite labels imported from another CP.
+	if labelsOpts.Privileged && !core_model.IsLocallyOriginated(labelsOpts.Mode, labels) {
```
lukidzi, line 126: Suggestion: The Privileged + IsLocallyOriginated check is subtle. A privileged write on a locally-originated resource WILL still recompute labels. Consider a brief clarifying comment since 'privileged' might be misread as 'skip all label computation'.

pkg/plugins/runtime/k8s/webhooks/defaulter.go
```diff
@@ -67,15 +67,21 @@ func (h *defaultingHandler) Handle(_ context.Context, req admission.Request) adm
...
 		resource.Descriptor(),
 		resource.GetSpec(),
 		resource.GetMeta().GetLabels(),
 		resource.GetMeta().GetMesh(),
+		displayName,
 		resource_labels.WithNamespace(resource_labels.GetNamespace(resource.GetMeta(), h.SystemNamespace)),
 		resource_labels.WithMode(h.Mode),
 		resource_labels.WithK8s(true),
 		resource_labels.WithZone(h.ZoneName),
+		resource_labels.WithPrivileged(h.isPrivilegedUser(h.AllowedUsers, req.UserInfo)),
```
lukidzi, line 84: Suggestion: WithPrivileged relies on isPrivilegedUser checking AllowedUsers. Ensure this list includes KDS sync service accounts, GC, and storage-version migrator. Misconfiguration would cause incorrect label recomputation on synced resources.

pkg/core/resources/labels/compute_test.go
```diff
@@ -392,12 +404,47 @@ var _ = Describe("Compute", func() {
...
+		labels, err := resource_labels.Compute(
+			res.Descriptor(),
+			res.GetSpec(),
+			existing,
+			"mesh-1",
+			"recomputed-name",
+			resource_labels.WithMode(core.Zone),
+			resource_labels.WithK8s(true),
+			resource_labels.WithZone("zone-1"),
+			resource_labels.WithPrivileged(true),
```
lukidzi, line 444: Suggestion: Test coverage for Privileged only covers non-locally-originated resources. Consider adding a case where Privileged=true but origin=zone (locally originated) to verify labels ARE still recomputed in that scenario.

pkg/plugins/runtime/k8s/webhooks/defaulter.go
lobkovilya: how can we ensure that? the configuration comes from env https://github.com/kumahq/kuma/blob/a841cee6f5cb7d9da700af022919e8b6f24617d3/deployments/charts/kuma/templates/_helpers.tpl#L330-L331

lukidzi (dismissed):

lukidzi (approved):

PR #17164, opened 2026-07-08
feat(intercp): add version to catalog instance

slonka:
## Motivation

Downstream builds can run more than one flavor of the control plane (e.g. two product generations) side by side against a single shared inter-CP catalog, which gives them a single shared leader across both flavors. A single leader works: the leader's work is flavor-agnostic and there are no data races by design.

The alternative - splitting the catalog so each flavor gets its own leader - was rejected. Two leaders would mean every leader-scoped component (KDS mux, VIP allocator, hostname generator, the various status updaters/generators, ...) runs twice, and each would have to be proven safe to run concurrently or have version stored somewhere. Keeping one shared catalog and one leader avoids that entirely.

## Implementation information

Add an optional `version` label to the inter-CP catalog `Instance`, carried over the `Ping` RPC (`PingRequest.version`) and sourced from a new `InterCp.Catalog.InstanceVersion` config value. Empty by default.

With this, a downstream `multitenant.Tenants` implementation can read the catalog, filter to same-version instances, and compute its shard bucket over that subset - keeping the single shared catalog/leader.

## Supporting documentation

xrel: https://github.com/Kong/kong-mesh/issues/9991

### DIFF
--- a/pkg/config/intercp/config.go
+++ b/pkg/config/intercp/config.go
@@ -47,7 +47,7 @@ type CatalogConfig struct {
 	// If empty then it's autoconfigured by taking the first IP of the nonloopback network interface.
 	InstanceAddress string `json:"instanceAddress" envconfig:"kuma_inter_cp_catalog_instance_address"`
 	// InstanceVersion is an opaque label recorded in the catalog for this instance.
-	// Empty in OSS; downstream builds set it to partition work across same-version instances.
+	// By default empty, downstream builds set it to partition work across same-version instances.
 	InstanceVersion string `json:"instanceVersion" envconfig:"kuma_inter_cp_catalog_instance_version"`
 	// Interval on which CP will send heartbeat to a leader.
 	HeartbeatInterval config_types.Duration `json:"heartbeatInterval" envconfig:"kuma_inter_cp_catalog_heartbeat_interval"`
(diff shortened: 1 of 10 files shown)


pkg/config/intercp/config.go
```diff
@@ -46,6 +46,9 @@ type CatalogConfig struct {
 	// InstanceAddress indicates an address on which other control planes can communicate with this CP
 	// If empty then it's autoconfigured by taking the first IP of the nonloopback network interface.
 	InstanceAddress string `json:"instanceAddress" envconfig:"kuma_inter_cp_catalog_instance_address"`
+	// InstanceVersion is an opaque label recorded in the catalog for this instance.
+	// Empty in OSS; downstream builds set it to partition work across same-version instances.
```
slonka:
```suggestion
	// By default empty, downstream builds set it to partition work across same-version instances.
```

pkg/intercp/catalog/catalog.go
```diff
@@ -14,6 +14,11 @@ type Instance struct {
 	Address     string `json:"address"`
 	InterCpPort uint16 `json:"interCpPort"`
 	Leader      bool   `json:"leader"`
+	// Version is an opaque label identifying which flavor of the CP this
+	// instance runs. It is empty in OSS; downstream builds can set it (via
```
slonka:
```suggestion
	// instance runs. By default empty, downstream builds can set it (via
```

lukidzi (commented):
1. **Observability (`pkg/intercp/catalog/server.go:30`):** The existing log statement logs `instanceID`, `address`, and `ready` but omits `version`. For debugging version-partitioned sharding, consider adding `"version", request.Version` to the log output.

2. **Test coverage (`pkg/intercp/catalog/heartbeat_component_test.go`):** The `currentInstance` fixture does not set `Version`, and assertions don't verify it propagates through the heartbeat cycle. Since `server_test.go` covers request→Instance, this is low priority, but adding `Version` to the fixture and asserting `received.Version` would complete the coverage.

lukidzi (approved):

PR #17176, opened 2026-07-08
feat(labels): compute `kuma.io/origin` on global and zone control planes

lobkovilya:
## Motivation

Part 2 of dismantling https://github.com/kumahq/kuma/pull/16852 into smaller chunks.

The origin label should be owned by the CP and set consistently wherever possible, since the CP always knows what the value should be.

## Implementation information

- `Compute` now force-sets `kuma.io/origin` to the CP's role: `global` on a global CP and `zone` on a zone CP, for every locally-created resource regardless of `ProvidedByZoneFlag`. The deprecated `standalone` mode is rewritten to `zone` at CP startup (`app/kuma-cp/cmd/run.go`), so it is covered by the zone case. `kuma.io/zone` / `kuma.io/env` stay gated on `ProvidedByZoneFlag`; the now-dead `origin != global` guard around them is dropped.
- **Scope is compute-only.** Origin *validation* (`validateOriginForWrite`, the webhook admission gate, strict-match errors, warnings) is untouched and stays in the follow-up PRs.
- Because origin is now CP-computed, a user-supplied value is **silently overridden** rather than rejected: api-server order is validate-then-compute, and the immutability check compares *computed* labels, so origin ends up equal on both sides. The `put_immutable_origin` api-server test reflects this — the second PUT is a 200 override instead of a 400 immutability error (same pattern as `put_immutable_displayname` in the previous step). Origin rejection is reinstated by the unified validator in a later PR.
- Goldens: six api-server resources lacking `ProvidedByZoneFlag` (health check, circuit breaker, fault injection, traffic route, traffic trace, multi-zone service) now gain `kuma.io/origin: zone`. The k8s defaulter tests gain `kuma.io/origin: global` on global-created resources; the obsolete "should not set zone label when origin is set to global" defaulter case is removed, since the CP no longer preserves a user-supplied global origin on a zone.

## Supporting documentation

Updates https://github.com/kumahq/kuma/issues/16049


lukidzi (commented):
Seems like PR description with  `> Changelog` doesn't follow repo standards: `type(scope)`

lukidzi (approved):

PR #17185, opened 2026-07-09
fix(kuma-dp): fix dataplane token write when WorkDir is missing

destinyoooo:
## Changes

- Fixed `kuma-dp` failing to write an inline dataplane token when `WorkDir` does not exist.
- Changed `writeFile` to create parent directories with directory permissions `0711` instead of reusing the token file permission `0600`.
- Added a regression test for missing `WorkDir`, verifying the token is created and the token file remains `0600`.


Automaat (approved):
Review Verdict: CLEAN

No blocking correctness issues found.

_Generated by Sybra harness_

PR #17193, opened 2026-07-09
test(framework): detect non-deterministic xDS config churn

### LINKED ISSUE #17166: Detect spurious xDS config churn in e2e (like the CP-panic check)
## Motivation

Non-deterministic xDS generation makes the CP re-push a logically identical config to a proxy repeatedly. Each re-push rebuilds the affected Envoy listener/cluster and drains in-flight connections, causing intermittent request failures — bad for cross-zone mTLS/HTTP2 and any "no interruptions" guarantee. Our e2e suites don't catch this: `ControlPlaneAssertions` checks no-panic / zero-restarts / no-NACKs, but not an idle CP that keeps regenerating.

Recent example: #17144 (`sort inbound SAN matchers`). With MeshIdentity + multiple trust domains, `downstreamTLSContext` ranged a Go map, emitting SAN matchers in random order. It reached `master` and only showed up as a flaky `multizone/meshproxy` "…no interruptions…" failure — one unedited proxy logged `config has changed` 23× with the `versions` hash oscillating (A→B→A), Envoy rebuilt its inbound listener 15×, ~1% of cross-zone requests reset.

## Why it's reliably detectable

`pkg/xds/server/v3/reconcile.go` versions each resource type by a deterministic content hash and logs `config has changed` (`:104`) only on a diff. So repeated changes for the same proxy with no edit in between prove non-deterministic generation. Unit guards exist (`reconcile_test.go`) but only cover per-policy snippets, not full configs like multi-trust-domain MeshIdentity.

## Proposal

Add `assertNoXdsChurn(cluster)` to `ControlPlaneAssertions` (`test/framework/debug.go:28`), next to `assertNoXdsNacks`. Every env's `SynchronizedAfterSuite` already calls `ControlPlaneAssertions` per CP, so it applies to multizone/k8s/universal with no extra wiring — same as the panic check.

- **Log oscillation detector (recommended, mirrors `HasPanicInCpLogs`):** parse `config has changed` from `GetKumaCPLogs()`, and per `proxyName` flag a `versions` hash that reappears after changing (A→B→A). Precise, needs no quiescence window, no false positives on normal config evolution.
- **Metric steady-state check (mirrors `assertNoXdsNacks`):** scrape `xds_dataplane_config_regenerated_total` / `xds_generation{result="changed"}` via `GetKuma().GetMetrics()`, assert ~no delta once idle. Caveat: CP-global + teardown noise, needs tolerance — hence the log approach is preferred.

## Acceptance criteria

- Reusable assertion that fails when a proxy's config regenerates without a logical change (hash oscillation).
- Wired into `ControlPlaneAssertions` (covers all three envs).
- Reverting #17144's sort makes it fail the `multizone/meshproxy` suite.

Automaat:
## Motivation

E2E tests only fail on xDS NACKs, so non-deterministic config generation (e.g. unordered map iteration producing repeated regenerations with identical content hashes) goes undetected even though it forces Envoy to rebuild and warm resources for no functional reason. This adds a check that fails the test when a proxy's xDS config is regenerated with a repeated, byte-identical content hash.

Closes https://github.com/kumahq/kuma/issues/17166

## Implementation information

- `test/framework/utils/utils.go`: `DetectXdsChurn` parses CP logs and reports hashes that repeat at least `xdsChurnThreshold` times, deduping hashes within a single "config has changed" log line so one line can only contribute once per hash.
- `test/framework/debug.go`: `assertNoXdsChurn` runs as part of `ControlPlaneAssertions` and fails the test with the offending reports.
- Fixed a pre-existing bug in `zz_generated.modules.go`/`zz_generated.policies.go` where a prior v2->v3 import-path rewrite had blanked the `NameToModule` map keys.

_Generated by Sybra harness_


test/framework/utils/utils.go
```diff
@@ -43,6 +44,89 @@ func HasPanicInCpLogs(logs string) bool {
 	return strings.Contains(logs, "runtime.gopanic") || strings.Contains(logs, "panic:")
 }
 
+var (
+	xdsChurnProxyNameRe         = regexp.MustCompile(`"proxyName":\s*"([^"]+)"`)
+	xdsChurnProxyNameFallbackRe = regexp.MustCompile(`proxyName=(\S+)`)
+	xdsChurnVersionsRe          = regexp.MustCompile(`"versions":\s*\[([^\]]*)\]`)
+	xdsChurnHashRe              = regexp.MustCompile(`[0-9a-f]{16}`)
+)
```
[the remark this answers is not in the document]
Automaat: **Applied** — churn is now keyed by an `xdsChurnKey{mesh, proxy}` struct; the log line already carries `mesh` (reconcile.go tags it alongside `proxyName`), so same-named DPs across meshes no longer merge (d5b1b92b32).

test/framework/utils/utils.go
```diff
@@ -43,6 +44,89 @@ func HasPanicInCpLogs(logs string) bool {
...
+		for _, hash := range xdsChurnHashRe.FindAllString(versions[1], -1) {
+			if seenInLine[hash] {
+				continue
+			}
+			seenInLine[hash] = true
+			if counts[proxyName] == nil {
+				counts[proxyName] = map[string]int{}
+			}
+			counts[proxyName][hash]++
+		}
```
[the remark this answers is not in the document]
Automaat: **Applied** — streak counts now increment against the per-(mesh, proxyName) key, so different meshes with the same proxyName are counted independently (d5b1b92b32).

test/framework/utils/utils.go
```diff
@@ -43,6 +44,89 @@ func HasPanicInCpLogs(logs string) bool {
...
+			count := counts[proxyName][hash]
+			if count >= xdsChurnThreshold {
+				reports = append(reports, fmt.Sprintf(
+					"proxy %s regenerated identical config %d times (hash %s) — non-deterministic xDS",
+					proxyName, count, hash,
+				))
+			}
+		}
+	}
+	return reports
```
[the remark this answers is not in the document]
Automaat: **Applied** — internal counts index by the `{mesh, proxy}` key while the report derives a stable display from its fields: `proxy <name> in mesh <mesh> ...` (d5b1b92b32).

test/framework/debug.go
```diff
@@ -44,6 +44,20 @@ func ControlPlaneAssertions(cluster Cluster) {
...
+// was regenerated with a repeated, byte-identical content hash. That
+// signals non-deterministic xDS generation (e.g. unordered map iteration),
+// which forces Envoy to rebuild and warm the affected resources repeatedly
+// even though nothing about the proxy's configuration actually changed.
+func assertNoXdsChurn(cluster Cluster, logs map[string]string) {
+	ginkgo.GinkgoHelper()
+	for k, log := range logs {
+		reports := utils.DetectXdsChurn(log)
+		Expect(reports).To(BeEmpty(), "CP %s has xDS churn in logs %s: %v", cluster.Name(), k, reports)
+	}
```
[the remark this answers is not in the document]
Automaat: **Skipped (questionable)** — the specific A→B→A concern is already handled: the detector counts *consecutive* streaks, so any intervening different hash resets the count — a legitimate apply/revert never accumulates toward the threshold. The real cross-suite false positive was interleaved *identical* hashes from same-named DPs in different meshes, which d5b1b92b32 fixes by keying per (mesh, proxyName). A full idle/quiescent-window rework (capture-wait-delta) is a larger design change; I left it out since the streak + mesh-keying invariants cover the described cases. Open to it if you still see residual false positives — leaving this open for your call.

test/framework/utils/utils.go
```diff
@@ -43,6 +44,133 @@ func HasPanicInCpLogs(logs string) bool {
...
+			}
+			nextStreaks[hash] = streak
+			if maxStreaks[key] == nil {
+				maxStreaks[key] = map[string]int{}
+			}
+			if streak > maxStreaks[key][hash] {
+				maxStreaks[key][hash] = streak
+			}
+		}
+		currentStreaks[key] = nextStreaks
```
bartsmykla: autoVersion logs only changed types, so this bug produces A, B, A, B. Replacing the map each line keeps both counts at 1. We should track hashes across lines

test/framework/utils/utils.go
```diff
@@ -43,6 +44,133 @@ func HasPanicInCpLogs(logs string) bool {
...
+		if proxyName == "" {
+			continue
+		}
+		key := xdsChurnKey{mesh: xdsChurnMesh(line), proxy: proxyName}
+		versions := xdsChurnVersionsRe.FindStringSubmatch(line)
+		if versions == nil {
+			continue
+		}
+		seenInLine := map[string]bool{}
+		for _, hash := range xdsChurnHashRe.FindAllString(versions[1], -1) {
```
bartsmykla: This drops the resource type. Empty Route, Secret, and Endpoint values share one hash, so three valid clears can trigger this check. I think we should count type and hash together

test/framework/debug.go
```diff
@@ -44,6 +44,20 @@ func ControlPlaneAssertions(cluster Cluster) {
...
+
+// assertNoXdsChurn parses the CP logs and fails if any proxy's xDS config
+// was regenerated with a repeated, byte-identical content hash. That
+// signals non-deterministic xDS generation (e.g. unordered map iteration),
+// which forces Envoy to rebuild and warm the affected resources repeatedly
+// even though nothing about the proxy's configuration actually changed.
+func assertNoXdsChurn(cluster Cluster, logs map[string]string) {
+	ginkgo.GinkgoHelper()
+	for k, log := range logs {
+		reports := utils.DetectXdsChurn(log)
```
bartsmykla, line 58: GetKumaCPLogs overwrites its fixed keys for each pod. With multiple replicas, this checks only the last pod. I think we should keep each pod’s logs separate

bartsmykla (changes requested):

test/framework/utils/utils.go
Automaat: **Applied** — the detector now keeps streak state per resource type across intervening log lines, and the CP log now emits typed  entries () so repeated  churn is counted correctly (ae8a5f43d4).

_Generated by Sybra harness_

test/framework/utils/utils.go
Automaat: **Applied** — churn counts are now keyed by resource type plus hash, so identical empty-hash clears from , , and  no longer collapse into one false-positive streak (ae8a5f43d4).

_Generated by Sybra harness_

test/framework/debug.go
Automaat: **Applied** —  now keeps each kuma-cp pod’s current and previous logs under pod-specific keys instead of overwriting a shared  entry, so multi-replica suites check every replica (ae8a5f43d4).

_Generated by Sybra harness_

test/framework/utils/utils.go
Automaat: **Applied** — the detector now keeps streak state per resource type across intervening log lines, and the CP log now emits typed versions entries in typeURL=hash form, so repeated A, B, A, B churn is counted correctly (ae8a5f43d4).

_Generated by Sybra harness_

test/framework/utils/utils.go
Automaat: **Applied** — churn counts are now keyed by resource type plus hash, so identical empty-hash clears from Route, Secret, and Endpoint no longer collapse into one false-positive streak (ae8a5f43d4).

_Generated by Sybra harness_

test/framework/debug.go
Automaat: **Applied** — GetKumaCPLogs now keeps each kuma-cp pod’s current and previous logs under pod-specific keys instead of overwriting a shared stdout entry, so multi-replica suites check every replica (ae8a5f43d4).

_Generated by Sybra harness_

Automaat: **Applied** — `GetKumaCPLogs` keeps each kuma-cp pod's current and previous logs under pod-specific keys (`<pod>/stdout`, `<pod>/stdout_previous`), so multi-replica suites inspect every replica. Still present in current HEAD (b938d0a062).

_Generated by Sybra harness_

bartsmykla (approved):

PR #17228, opened 2026-07-09
feat(helm): configurable zone proxy probes

lukidzi:
## Motivation

Zone Ingress and Egress deployments hardcode their `livenessProbe` and `readinessProbe`, and expose no `startupProbe`. Today the only workaround is a post-render / manual patch, which is awkward for GitOps (Flux/Argo) users who manage everything through Helm values.

## Implementation information

- Add `livenessProbe`, `readinessProbe`, and `startupProbe` blocks under `ingress` and `egress` in the chart values.
- `livenessProbe` and `readinessProbe` gain an `enabled` flag (default `true`) so they can be turned off; all timing fields (`failureThreshold`, `initialDelaySeconds`, `periodSeconds`, `successThreshold`, `timeoutSeconds`) are configurable.
- `startupProbe` is `enabled: false` by default; when enabled it suspends the liveness probe until Envoy is ready, giving slow-warming proxies the time they need.
- Liveness/readiness defaults exactly match the previously-hardcoded values, so existing installs render byte-identical output — no behavior change unless the new values are set.
- The `httpGet` path/port stays templated (respects `experimental.envoyAdminUnixSocket`), so users never need to know the admin port.

Example enabling a 5-minute startup window on the ingress:

```yaml
ingress:
  startupProbe:
    enabled: true
    failureThreshold: 60
    periodSeconds: 5
```

## Supporting documentation

- Chart README regenerated via helm-docs.
- Added a kumactl install golden test (`zoneProxyProbes`) covering enabled startup probes, overridden liveness/readiness fields, and a disabled liveness probe.

### DIFF
--- a/deployments/charts/kuma/values.yaml
+++ b/deployments/charts/kuma/values.yaml
@@ -564,8 +564,9 @@ ingress:
       cpu: 1000m
       memory: 512Mi
 
-  # Liveness probe for the Ingress
+  # -- Liveness probe settings for the Ingress proxy
   livenessProbe:
+    # -- Whether to enable the liveness probe.
     enabled: true
     failureThreshold: 12
     initialDelaySeconds: 60
@@ -573,8 +574,9 @@ ingress:
     successThreshold: 1
     timeoutSeconds: 3
 
-  # Readiness probe for the Ingress
+  # -- Readiness probe settings for the Ingress proxy
   readinessProbe:
+    # -- Whether to enable the readiness probe.
     enabled: true
     failureThreshold: 12
     initialDelaySeconds: 1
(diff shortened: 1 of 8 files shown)


deployments/charts/kuma/values.yaml
```diff
@@ -723,6 +753,36 @@ egress:
       cpu: 1000m
       memory: 512Mi
 
+  # -- Liveness probe settings for the Ingress proxy
```
slonka: this is under egress but the comments say Ingress in multiple places - looks copy pasted

slonka (approved):

PR #17317, opened 2026-07-10
fix(kuma-dp): restrict readiness reporter to /ready

bartsmykla:
## Motivation

The kuma-dp readiness reporter listens on a TCP port that is reachable over the pod network (Kubernetes probes need it on the pod IP). When Envoy admin is bound to a Unix socket, the readiness reporter also reverse-proxied the whole Envoy admin API on that port behind a small path denylist, so read and mutating admin endpoints (config dump, clusters, logging, counters, health-check toggles, ...) were reachable there. The readiness port should only answer readiness.

## Implementation information

- kuma-dp: the readiness listener now serves only `/ready` (proxied to the Envoy admin Unix socket so the pod is marked ready once Envoy has its config). The catch-all admin reverse proxy and its denylist are removed. Added a unit test asserting `/ready` still works while every admin path returns 404.
- Kubernetes e2e used that admin proxy as its Envoy-admin tunnel. It now reads stats/clusters/config-dump through the control plane inspect API (`/meshes/{mesh}/dataplanes/{name}/{stats,xds,clusters}`), which reaches the proxy over the control plane's own mTLS admin channel, so no test depends on admin being exposed on the readiness port. `reset_counters` has no inspect equivalent, so the two affected specs assert on counter deltas instead. Universal e2e is unchanged (it curls the admin socket in-container).

Validated on real k3d clusters: the focused MeshTimeout spec passes via the new tunnel, a live sidecar serves `/ready` (200) while `/config_dump`, `/stats`, `/clusters`, `/quitquitquit` all return 404, and the control plane inspect endpoints return the expected JSON for both a regular dataplane and a zone egress.


test/framework/k8s_controlplane.go
```diff
@@ -344,6 +345,44 @@ func (c *K8sControlPlane) GetAPIServerAddress() string {
...
+	if err != nil {
+		return nil, err
+	}
+	for _, header := range c.apiHeaders {
+		if kv := strings.SplitN(header, "=", 2); len(kv) == 2 {
+			req.Header.Set(kv[0], kv[1])
+		}
+	}
+
+	resp, err := http.DefaultClient.Do(req) // #nosec G107 -- CP address is the test port-forward
```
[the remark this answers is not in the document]
bartsmykla: Bounded the request with a context deadline and an http.Client timeout (DefaultTimeout), so a stalled port-forward or API server surfaces as an error the caller's Eventually retries instead of hanging.

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -28,10 +27,13 @@ const (
...
-// endpoints so that external tools can reach admin over TCP even when
-// Envoy admin listens on a Unix domain socket. Mutating endpoints
-// (/quitquitquit, /drain_listeners, /runtime_modify) are blocked.
+// The listener serves only /ready. When adminSocketPath is set, /ready is
+// proxied to Envoy admin over the Unix domain socket so the pod is marked
+// ready only after Envoy has its config. No other Envoy admin endpoint is
+// exposed on this listener: the readiness port is reachable over the pod
+// network (K8s probes), so proxying admin here would expose config_dump,
+// certs, logging, etc. to any co-located or pod-network peer. Admin stays
+// on the UDS, which is isolated to this container's mount namespace.
```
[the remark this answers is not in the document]
bartsmykla: Reworded - dropped the mount-namespace claim; the point that matters is admin stays on the UDS and off the pod-network readiness port.

app/kuma-dp/pkg/dataplane/readiness/component_test.go
```diff
@@ -215,11 +224,25 @@ var _ = Describe("Readiness Reporter", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(string(body)).To(Equal("LIVE\n"))
+		})
+
+		It("does not expose any admin endpoint on the readiness port", func() {
+			for _, path := range []string{"/config_dump", "/stats", "/clusters", "/certs", "/logging", "/quitquitquit"} {
+				resp, err := http.Get(baseURL + path)
+				Expect(err).ToNot(HaveOccurred())
+				body, _ := io.ReadAll(resp.Body)
+				resp.Body.Close()
```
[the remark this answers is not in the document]
bartsmykla: Now asserting the io.ReadAll error before checking the body.

test/e2e_env/multizone/gateway/cross-mesh.go
```diff
@@ -166,10 +166,6 @@ func CrossMeshGatewayOnMultizone() {
 				crossMeshGatewayServiceName,
 			)
 
-			BeforeEach(func() {
-				Expect((*zone).GetZoneEgressEnvoyTunnel().ResetCounters()).To(Succeed())
-			})
-
 			It("proxies HTTP requests from a different mesh", func() {
 				Eventually(k8s_gateway.SuccessfullyProxyRequestToGateway(
 					*zone, gatewayMesh,
```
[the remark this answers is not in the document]
bartsmykla: Right - the shared cumulative counter made the second spec a tautology. Each spec now captures a baseline and asserts the counter grows over its own request (reset isn't available via the CP inspect tunnel).

test/e2e_env/multizone/gateway/cross-mesh.go
```diff
@@ -166,27 +166,47 @@ func CrossMeshGatewayOnMultizone() {
...
+			currentRequestTotal := func() float64 {
+				s, err := (*zone).GetZoneEgressEnvoyTunnel().GetStats(filter)
+				Expect(err).ToNot(HaveOccurred())
+				if len(s.Stats) == 0 {
+					return 0
+				}
+				value, ok := s.Stats[0].Value.(float64)
+				Expect(ok).To(BeTrue())
+				return value
+			}
```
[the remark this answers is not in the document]
bartsmykla: Now fails if the filter matches more than one stat (0 still allowed as the pre-traffic baseline), restoring the exactly-one invariant the stats matcher had.

test/e2e_env/multizone/localityawarelb/locality_aware_multizone_hybrid.go
```diff
@@ -380,41 +387,47 @@ spec:
...
+		return 0, nil
+	}
+	switch v := s.Stats[0].Value.(type) {
+	case float64:
+		return int(v), nil
+	case string:
+		return strconv.Atoi(v)
+	default:
+		return -1, errors.Errorf("unexpected value type %T for stat %q", s.Stats[0].Value, metricName)
 	}
```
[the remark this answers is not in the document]
bartsmykla: Same here - collectMetric now returns an error when the filter matches more than one stat instead of silently taking the first.

test/framework/k8s_controlplane.go
```diff
@@ -344,6 +345,48 @@ func (c *K8sControlPlane) GetAPIServerAddress() string {
...
+	if err != nil {
+		return nil, err
+	}
+	for _, header := range c.apiHeaders {
+		if kv := strings.SplitN(header, "=", 2); len(kv) == 2 {
+			req.Header.Set(kv[0], kv[1])
+		}
+	}
+
+	client := &http.Client{Timeout: DefaultTimeout}
```
Automaat: **suggestion:** `InspectEnvoyProxy` caps a full-stats fetch (`usedonly=false`, routed client→CP→proxy over mTLS) at `DefaultTimeout` (3s) for both context and client, with no retry — while the sibling `GetMetrics` uses 30s. Under full-matrix CI contention this can time out and flake. Give it its own ~30s timeout.

test/e2e_env/multizone/gateway/cross-mesh.go
```diff
@@ -166,27 +166,51 @@ func CrossMeshGatewayOnMultizone() {
...
+				// The filter must identify a single counter; more than one
+				// match means it is too broad and we would read the wrong one.
+				Expect(s.Stats).To(HaveLen(1), "stats filter %q matched %d stats, expected 1", filter, len(s.Stats))
+				value, ok := s.Stats[0].Value.(float64)
+				Expect(ok).To(BeTrue())
+				return value
+			}
 
 			It("proxies HTTP requests from a different mesh", func() {
+				before := currentRequestTotal()
```
Automaat: **suggestion:** This baseline runs outside the retry loop and `currentRequestTotal` hard-`Expect`s `err==nil` and `HaveLen(1)`, so one transient CP-inspect hiccup fails the spec with no retry (the after-assertion is safely wrapped in `Eventually`). Capture the baseline inside an `Eventually` too. Same pattern at line 203.

test/framework/k8s_controlplane.go
```diff
@@ -344,6 +345,48 @@ func (c *K8sControlPlane) GetAPIServerAddress() string {
...
+	resp, err := client.Do(req) // #nosec G107 -- CP address is the test port-forward
```
Automaat: **suggestion:** `#nosec G107` here is dead — `.golangci.yml` excludes gosec for `test/` — and the repo convention for `.Do()` is G704 anyway, so just delete it. While here, wrap this `client.Do` error with `inspectPath` like the non-200 branch does, so a 3s timeout isn't a bare "context deadline exceeded" during triage.

test/e2e_env/multizone/gateway/cross-mesh.go
```diff
@@ -166,27 +166,51 @@ func CrossMeshGatewayOnMultizone() {
...
 
-			BeforeEach(func() {
-				Expect((*zone).GetZoneEgressEnvoyTunnel().ResetCounters()).To(Succeed())
-			})
+			// The counter is cumulative and shared across the specs in this
+			// Context, and ResetCounters is not available through the control
+			// plane inspect tunnel. So each spec captures a baseline and
+			// asserts the counter grows over its own request rather than just
+			// being non-zero (which the previous spec would already satisfy).
+			currentRequestTotal := func() float64 {
```
Automaat: **suggestion:** `currentRequestTotal` here and `collectMetric` in locality_aware_multizone_hybrid.go are the same "GetStats → len==0?0 : len>1?err : value" logic (comments copy-pasted), and the `len==1` rule is a third copy in `stats.statMatcher`. Consider one `stats.Stats.SingleValue() (float64, error)` next to the matcher.

test/e2e_env/multizone/localityawarelb/locality_aware_multizone_hybrid.go
```diff
@@ -380,41 +387,53 @@ spec:
...
+			g.Expect(successRequests - base2xx).To(BeNumerically("~", 100, 25))
 		}, "60s", "10s").Should(Succeed())
 	})
 }
 
+// collectMetric reads a single Envoy stat value for the app's proxy through
+// the control plane inspect API (Envoy admin is no longer reachable over the
+// pod network). A stat that Envoy hasn't created yet reads as 0.
+//
+//nolint:unparam // name is the sole client app in this suite; kept explicit for readability
```
Automaat: **suggestion:** `unparam` fires because `name` is always `"demo-client-mesh-route"`, and it is not excluded for `test/` (only gosec is), so this suppression is load-bearing. Prefer dropping the constant parameter (or folding into a shared helper) over suppressing the linter — matches the repo's no-lint-skip rule.

test/framework/k8s_controlplane.go
bartsmykla: Gave it a dedicated 30s budget (inspectEnvoyProxyTimeout) for both the context and the client, matching GetMetrics.

test/e2e_env/multizone/gateway/cross-mesh.go
bartsmykla: Both baselines now capture inside an Eventually, so a transient CP-inspect hiccup retries instead of failing the spec outright.

test/framework/k8s_controlplane.go
bartsmykla: Dropped the dead #nosec and wrapped the client.Do error with the inspect path, so a timeout surfaces the path instead of a bare context-deadline.

test/e2e_env/multizone/gateway/cross-mesh.go
bartsmykla: Pulled the logic into stats.Stats.SingleValue() next to the matcher; currentRequestTotal and collectMetric both call it, so the exactly-one rule lives in one place.

test/e2e_env/multizone/localityawarelb/locality_aware_multizone_hybrid.go
bartsmykla: Dropped the nolint and the constant params - collectMetric now takes only metricName and reads the fixed client/namespace from package consts.

Automaat (dismissed):

app/kuma-dp/pkg/dataplane/readiness/component.go
```diff
@@ -93,13 +95,10 @@ func (r *Reporter) Start(stop <-chan struct{}) error {
 	logger.Info("starting readiness reporter", "addr", lis.Addr().String())
 
 	mux := http.NewServeMux()
+	// Only /ready is served. Every other path returns 404. Do not add a
+	// catch-all proxy to Envoy admin here: this listener is reachable over
+	// the pod network, so it must not expose admin endpoints.
 	mux.HandleFunc(pathPrefixReady, r.handleReadiness)
```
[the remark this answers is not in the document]
bartsmykla: This doesn't apply here: the pattern is "/ready" without a trailing slash, so net/http.ServeMux matches it exactly - only a trailing-slash pattern ("/ready/") matches the subtree. Subpaths like /ready/config_dump already 404. Added a test asserting /ready/, /ready/config_dump and /readyz all return 404 and never hit the readiness handler, so the surface can't silently widen.

Automaat (dismissed):

Automaat (approved):

PR #17349, opened 2026-07-13
test(report): persist files outside ginkgo spec

slonka:
## Motivation

The Gateway API conformance suite (`test/e2e_env/gatewayapi/conformance_test.go`) runs as a plain `testing` test, not a Ginkgo spec.
On failure its `t.Cleanup` calls `DebugKube` -> `DumpState` -> `report.AddFileToReportEntry`, which called `ginkgo.AddReportEntry`.
Outside a running Ginkgo spec that panics with `...not during the Run phase...`, so a single flaky conformance failure crashed the whole test binary, masked the real failure, and dropped the debug artifacts that would help diagnose it.

This was behind a chunk of the recent `build-test-distribute` master failures.

## Implementation information

`AddFileToReportEntry` now persists the file directly to the report `BaseDir` when there is no running Ginkgo spec.
There is no `ReportAfterSuite` to flush the ginkgo report entries in that case anyway.
Real Ginkgo suites keep using `ginkgo.AddReportEntry` unchanged.
Detection uses `CurrentSpecReport().LeafNodeType`, which avoids touching Ginkgo's shared `Failer` state, so it can't leak a failed state into the `TestE2E` Ginkgo suite in the same binary.
Added a focused plain-`testing` test that reproduces the exact panic against the old code and now asserts the file is written to `BaseDir`.

## Supporting documentation

Follow-up to #17195.

> Changelog: skip


lukidzi (dismissed):
Just a lot of comments that could be reduced

lukidzi (approved):

PR #17352, opened 2026-07-13
test(unit): measure race vs non-race timing

Automaat:
## Motivation

Throwaway experiment (DO NOT MERGE) to quantify how much the Go race
detector adds to PR unit-test wall-clock.

> Changelog: skip

## Implementation information

`test_unit` runs the unit suite twice on the same runner with the same
cold cache — once with `UNIT_RACE=` (no race) then once with the default
`-race` — and prints each elapsed time to the job summary. Only the
`-race` flag differs, so the delta is the race cost. e2e is skipped via
`ci/skip-e2e-test`; timeout bumped to fit both runs.

## Supporting documentation

N/A — measurement only.

### DIFF
--- a/.github/workflows/_test.yaml
+++ b/.github/workflows/_test.yaml
@@ -17,7 +17,7 @@ env:
   MISE_DISABLE_TOOLS: "golangci-lint,skaffold,aqua:stackrox/kube-linter"
 jobs:
   test_unit:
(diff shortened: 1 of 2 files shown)


Automaat:
Superseded by #17353 (the real change). This PR was only the throwaway race-vs-non-race A/B measurement.

Issue #8844, opened 2024-01-14
Data race in `FillTenantMetadata()`


slonka:
### What happened?

I was working on an unrelated PR and noticed this failure:

https://github.com/kumahq/kuma/actions/runs/7500332065/job/20421086539?pr=8828#step:8:469

PR #17353, opened 2026-07-13
test(unit): skip race detector on PRs by default

Automaat:
## Motivation

The Go race detector adds **~30% (~6–7 min)** to the PR unit job for no
wall-clock benefit while e2e is the long pole — it is pure runner cost and
slower feedback on most PRs.

Measured with an A/B (#17352, same \`ubuntu-24.04\` runner, cold cache):

| variant | unit job |
|---------|----------|
| **without** \`-race\` | **15m42s** |
| with \`-race\` (median of 10 recent PR runs) | ~23m (19–25.5m) |

No-race was faster than *every* race sample, including the fastest (19m05s).

> Changelog: skip

## Implementation information

- New \`UNIT_RACE\` make var (default \`-race\`); the CI unit path uses
  \`$(UNIT_RACE)\` instead of a hardcoded \`-race\`, so local \`make test\`
  is unchanged.
- \`test_unit\` sets \`UNIT_RACE=-race\` for non-PR events (push to
  \`master\`/\`release-*\`, tags, \`workflow_dispatch\`/nightly) and for PRs
  carrying the new \`ci/run-race\` label; empty otherwise.

**Tradeoff (flagged for review):** race detection for PR-only code paths
moves post-merge. Mitigations: it still runs on every \`master\`/\`release\`
push (caught before any release) and nightly; and anyone touching
concurrency-heavy code (CP, xDS, KDS, caches, leader election) or fixing a
master race can add \`ci/run-race\` to reproduce pre-merge. An alternative
that keeps pre-merge detection — warm go-build cache + sharding \`go test\`
— is noted in the CI speed-up analysis; happy to go that route instead.

e2e is skipped (\`ci/skip-e2e-test\`) because it does not use \`UNIT_RACE\`
and is unaffected by this change.

## Supporting documentation

- A/B experiment: #17352


.github/workflows/_test.yaml
```diff
@@ -20,6 +20,9 @@ jobs:
     timeout-minutes: 30
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci/skip-test') }}
     runs-on: ${{ vars.RUNS_ON_MASTER_AMD64 || vars.RUNS_ON_AMD64 || 'ubuntu-24.04' }}
+    env:
+      # Race detector adds ~30% (~7min) to the unit job, so PRs skip it unless opted in.
+      UNIT_RACE: ${{ (github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'ci/run-race')) && '-race' || '' }}
```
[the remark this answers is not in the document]
Automaat: **Applied** — pass the pull-request state into the reusable workflow so PR runs skip `-race` by default unless labeled `ci/run-race` (6c8a54dfed).

_Generated by Sybra harness_

.github/workflows/_test.yaml
```diff
@@ -20,6 +23,9 @@ jobs:
     timeout-minutes: 30
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci/skip-test') }}
     runs-on: ${{ vars.RUNS_ON_MASTER_AMD64 || vars.RUNS_ON_AMD64 || 'ubuntu-24.04' }}
+    env:
+      # Race detector adds ~30% (~7min) to the unit job, so PRs skip it unless opted in.
+      UNIT_RACE: ${{ (!inputs.IS_PULL_REQUEST || contains(github.event.pull_request.labels.*.name, 'ci/run-race')) && '-race' || '' }}
```
slonka, line 28: is the runner-cost saving worth losing pre-merge detection?

I'd weigh against that: `-race` on PRs has caught real bugs pre-merge - #8844 was a `-race` failure someone noticed on an unrelated PR, #12980 was a data race that only showed under `-race` in unit tests. Moving it post-merge means a race lands on master and gets caught by the nightly/master run after it's already merged

Automaat: **Skipped (questionable)** — this is a tradeoff call rather than a correctness bug in the workflow. The current change saves ~6-7 minutes on the PR unit job and still keeps `-race` on post-merge/master/release runs plus opt-in via `ci/run-race`, but it does move PR-only race detection post-merge. Leaving the thread open for your call.

_Generated by Sybra harness_

PR #17355, opened 2026-07-13
test(e2e): slim image for in-cluster workloads

Automaat:
## Motivation

Prototype (measuring impact) to cut the ~3-min per-job e2e image-import.
Today every k3d cluster imports the fat `kuma-universal` image (ubuntu-netools
+ envoy + spire-server/agent + kuma-cp/dp/coredns/kumactl/test-server), because
the in-cluster test workloads reuse it. Most of that is unused in-cluster.

> Changelog: skip

## Implementation information

- New `kuma-test-app` image = `kuma-universal` **minus** kuma-cp/kumactl/spire
  (~250 MB of binaries), same `ubuntu-netools` base so all shell tools
  (curl/ncat/sleep) are unchanged.
- Repoint in-cluster (k3d) consumers to it: test-server, demo-client, the curl
  Job in `setup.go`, and zone proxies. k8s SPIRE uses upstream spiffe charts
  (unaffected); universal-mode container zones keep `kuma-universal` on host docker.
- `K3D_IMPORT_IMAGES` drops `kuma-universal` from the k3d import (release +
  test-app only). Kind still loads everything (correctness preserved).

Measuring: k3d image-import + cluster-setup time on multizone/kubernetes vs the
baseline. Not for merge as-is — validating the approach + delta.

## Supporting documentation

Follow-up to the CI e2e long-pole analysis.

### DIFF
--- a/mk/docker.mk
+++ b/mk/docker.mk
@@ -11,7 +11,7 @@ $(addsuffix :$(BUILD_INFO_VERSION)$(if $(2),-$(2)),$(addprefix $(DOCKER_REGISTRY
 endef
 
 IMAGES_RELEASE += kuma-cp kuma-dp kumactl kuma-init kuma-cni
-IMAGES_TEST += kuma-universal
+IMAGES_TEST += kuma-universal kuma-test-app
 KUMA_IMAGES = $(call build_image,$(IMAGES_RELEASE) $(IMAGES_TEST))
 
 # Always use Docker BuildKit, see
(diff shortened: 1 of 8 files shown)


Automaat:
Closing — measured negative result. The slim `kuma-test-app` image worked (279/280 specs passed, universal excluded from k3d import) but cut only ~13s: image import is only ~35s, not the ~3min originally attributed to it. That window is actually cluster creation + kuma control-plane/zone readiness wait (~143s) + a baseline import retry. Slimming the image optimizes a non-bottleneck. Real setup levers: parallelize the sequential cluster creation (~40s) + go-build cache for the build.

PR #17356, opened 2026-07-13
test(unit): reuse test cache on PRs

Automaat:
## Motivation

Prototype (measuring). CI runs `go clean -testcache` then `go test ./...`, so
all ~700 packages recompile + rerun every PR with no cache. Reusing the Go
build/test cache lets unchanged packages return cached results.

> Changelog: skip

## Implementation information

- `UNIT_CLEAN_TESTCACHE` make var gates `go clean -testcache`; `test_unit` sets
  it `false` on PRs, `true` on push (master/release run fresh, unchanged).
- Add `actions/cache` for `~/.cache/go-build` + `~/go/pkg/mod` to `test_unit`.
- Race stays on. Safety net: master still cleans + runs `-race` fresh.

Measuring: run 1 cold (populate cache), then a 1-package change for run 2 (warm)
to see the realistic small-PR unit time.

## Supporting documentation

N/A

### DIFF
--- a/.github/workflows/_test.yaml
+++ b/.github/workflows/_test.yaml
@@ -20,6 +20,9 @@ jobs:
     timeout-minutes: 30
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci/skip-test') }}
     runs-on: ${{ vars.RUNS_ON_MASTER_AMD64 || vars.RUNS_ON_AMD64 || 'ubuntu-24.04' }}
+    env:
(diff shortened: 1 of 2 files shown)


Automaat:
Superseded by the cache-only version. Measurement showed test-result reuse is a dud here (only 7/281 packages cacheable — Kuma tests are non-hermetic), so the win is pure go-build caching; keeping 'go clean -testcache' avoids any flake/coverage tradeoff.

PR #17363, opened 2026-07-14
refactor(k8s)!: drop sidecarContainers opt-out

Automaat:
## Motivation

Native Kubernetes sidecar containers (init containers with `restartPolicy: Always`, GA since Kubernetes 1.29) became the default injection mode two releases ago. The `experimental.sidecarContainers` config/Helm value has been a pure opt-out since then. Kuma's own CI already only tests against Kubernetes 1.32+, well above the 1.29 requirement, so the opt-out no longer protects any supported configuration and the dedicated CI matrix entry that existed only to exercise the legacy injection path is pure cost.

> Changelog: skip

## Implementation information

- Removed `Experimental.SidecarContainers` from the control plane config, the `KUMA_EXPERIMENTAL_SIDECAR_CONTAINERS` env var, and the Helm `experimental.sidecarContainers` value.
- `pkg/plugins/runtime/k8s/plugin.go` now derives native-sidecar support purely from the live Kubernetes server version instead of ANDing it with a user config flag. The control plane still auto-falls-back to legacy (non-native) injection on Kubernetes older than 1.29 — that compatibility check isn't user-configurable, so it stays.
- `pkg/plugins/runtime/k8s/webhooks/injector/injector.go` no longer branches on a config-driven bool: it always injects the sidecar as a native init container with `restartPolicy: Always`. Regenerated/renamed the injector golden fixtures accordingly (kept the "native" variants, dropped the legacy ones).
- Removed the `sidecarContainers` axis from the e2e CI matrix (`.github/workflows/_test.yaml`, `.github/actions/run-e2e/action.yml`) and the matching `KUMA_EXPERIMENTAL_SIDECAR_CONTAINERS` plumbing in `test/framework`.
- Regenerated docs (`docs/generated/raw/*.yaml`, Helm chart README) and the `kumactl install control-plane`/Helm golden fixtures that baked in the removed env var.
- Updated `UPGRADE.md` to describe the removal and drop the now-obsolete "revert to old behavior" instructions.

## Supporting documentation

N/A


.github/workflows/_test.yaml
```diff
@@ -49,23 +49,20 @@ jobs:
...
                 "parallelism": [1],
                 "cniNetworkPlugin": ["flannel"],
-                "sidecarContainers": [""],
                 "exclude":[
                   {"target": "kubernetes", "k8sVersion":"kind"},
                   {"target": "multizone", "k8sVersion":"kind"},
                   {"target":"universal", "k8sVersion":"${{ env.K8S_MAX_VERSION }}"}
                 ],
                 "include":[
-                  {"sidecarContainers": "false", "k8sVersion": "${{ env.K8S_MAX_VERSION }}", "target": "kubernetes", "arch": "amd64"},
```
lukidzi, line 68: Let's bump minimum version of k8s to 1.34 (seems 1.33 is deprecated) and since 1.33 it was enabled by default

pkg/plugins/runtime/k8s/plugin.go
```diff
@@ -378,28 +375,10 @@ func addValidators(mgr kube_ctrl.Manager, rt core_runtime.Runtime, converter k8s
...
-			fmt.Sprintf("%s.%s.0", k8sVersion.Major, k8sVersion.Minor),
-		); err == nil && !v.LessThan(sidecarContainerVersion) {
-			sidecarContainersEnabled = rt.Config().Experimental.SidecarContainers
-		} else if rt.Config().Experimental.SidecarContainers {
-			log.Info("WARNING: sidecarContainers feature is enabled but Kubernetes server does not support it")
-		}
 		kumaInjector, err := injector.New(
 			rt.Config().Runtime.Kubernetes.Injector,
 			address,
 			mgr.GetClient(),
```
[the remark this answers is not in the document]
Automaat: **Applied** — restored Kubernetes server-version gating so clusters older than 1.29 fall back to legacy injection while 1.29+ still use native sidecars (`4632b56c1d`).

_Generated by Sybra harness_

.github/workflows/_test.yaml
Automaat: Bumped `K8S_MIN_VERSION` to `v1.34.9-k3s1` in `.github/workflows/_test.yaml`.

lukidzi (approved):

PR #17365, opened 2026-07-14
feat(api-server): add pluggable per-route metadata provider

bartsmykla:
## Motivation

The api-server auto-generates the REST CRUD routes for every registered resource type in one loop. Today there is no way for an embedder (a downstream build that composes Kuma) to attach metadata to those generated routes — only hand-registered WebServices can call `.Metadata(...)`. That metadata is useful for cross-cutting concerns such as authorization or observability middleware that need per-route, per-method context on the generated resource endpoints.

## Implementation information

Add a provider-neutral `RouteMetadataProvider` hook, plumbed the same way as the existing `APIWebServiceCustomize` hook on the runtime builder and interface:

- `runtime.RouteMetadataProvider` — `func(desc model.ResourceTypeDescriptor, method string) map[string]string`
- `Builder.WithRouteMetadataProvider(...)` setter + `Runtime.RouteMetadataProvider()` getter
- `NewApiServer` threads it into the resource-endpoint context; a small `applyRouteMetadata` helper attaches the returned key/values to the generated find / list / create-or-update / delete routes via the `restful` `RouteBuilder.Metadata`.

Kuma stays neutral — it attaches whatever the provider returns without interpreting it. A nil provider is a no-op, so existing behavior is unchanged.

## Supporting documentation

Added `pkg/api-server/route_metadata_test.go` asserting the generated `PUT /{name}` route carries provider metadata, and none when the provider is nil.

### DIFF
--- a/pkg/api-server/resource_endpoints.go
+++ b/pkg/api-server/resource_endpoints.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 
 	mesh_proto "github.com/kumahq/kuma/v3/api/mesh/v1alpha1"
+	"github.com/kumahq/kuma/v3/pkg/api-server/authn"
 	api_server_types "github.com/kumahq/kuma/v3/pkg/api-server/types"
 	config_core "github.com/kumahq/kuma/v3/pkg/config/core"
 	core_plugins "github.com/kumahq/kuma/v3/pkg/core/plugins"
@@ -86,6 +87,14 @@ func typeToLegacyOverviewPath(resourceType core_model.ResourceType) string {
 	}
 }
 
+// reservedRouteMetadataKeys are route metadata keys Kuma interprets itself, so a
+// RouteMetadataProvider must not set them. If it does, route drops the value to
+// stop an embedder from changing Kuma's own routing behavior (e.g. disabling
+// authn via authn.MetadataAuthKey).
+var reservedRouteMetadataKeys = map[string]struct{}{
+	authn.MetadataAuthKey: {},
+}
+
 // route builds the RouteBuilder for a CRUD method/path and applies any provider
 // metadata, naming the verb once so ws.<VERB> and the metadata can't desync.
 func (r *resourceEndpoints) route(ws *restful.WebService, method, path string) *restful.RouteBuilder {
@@ -102,6 +111,10 @@ func (r *resourceEndpoints) route(ws *restful.WebService, method, path string) *
 	}
 	if r.routeMetadataProvider != nil {
 		for k, v := range r.routeMetadataProvider(r.descriptor, method) {
+			if _, reserved := reservedRouteMetadataKeys[k]; reserved {
+				log.Info("ignoring reserved route metadata key from provider", "key", k, "type", r.descriptor.Name, "method", method)
+				continue
+			}
 			rb = rb.Metadata(k, v)
 		}
 	}
--- a/pkg/api-server/resource_endpoints.go
+++ b/pkg/api-server/resource_endpoints.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kumahq/kuma/v3/pkg/core/resources/store"
 	rest_errors "github.com/kumahq/kuma/v3/pkg/core/rest/errors"
 	"github.com/kumahq/kuma/v3/pkg/core/rest/errors/types"
+	"github.com/kumahq/kuma/v3/pkg/core/runtime"
 	"github.com/kumahq/kuma/v3/pkg/core/user"
 	"github.com/kumahq/kuma/v3/pkg/core/validators"
 	meshhttproute_api "github.com/kumahq/kuma/v3/pkg/plugins/policies/meshhttproute/api/v1alpha1"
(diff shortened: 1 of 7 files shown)


pkg/core/runtime/builder.go
```diff
@@ -293,6 +294,11 @@ func (b *Builder) WithAPIWebServiceCustomize(customize func(*restful.WebService)
 	return b
 }
 
+func (b *Builder) WithRouteMetadataProvider(provider RouteMetadataProvider) *Builder {
```
Automaat, line 300: **suggestion:** `WithRouteMetadataProvider` overwrites (last write wins) while the sibling `WithAPIWebServiceCustomize` appends to a slice. A second caller silently drops the first provider, with no error or log. If a single provider is the intent, say so in a doc comment on the setter.

pkg/core/runtime/runtime.go
```diff
@@ -113,6 +115,14 @@ type ResourceValidators struct {
 
 type ExtraReportsFn func(Runtime) (map[string]string, error)
 
+// RouteMetadataProvider supplies metadata key/values to attach to the
```
Automaat: **suggestion:** Scope is CRUD-only (find/list/PUT/DELETE), but the doc comment says "called once per generated route," which overstates it. Subroutes (`_overview`, `_config`, `_policies`, `_rules`, inspect, legacy overview) are excluded on purpose. Tighten the comment to state the CRUD-only scope and why subroutes are left out.

pkg/api-server/resource_endpoints.go
```diff
@@ -84,12 +86,25 @@ func typeToLegacyOverviewPath(resourceType core_model.ResourceType) string {
...
 }
 
+// applyRouteMetadata attaches embedder-supplied metadata to a generated route
+// via the restful RouteBuilder. A nil provider is a no-op, keeping the default
+// behavior unchanged. method is the HTTP verb of the route.
+func (r *resourceEndpoints) applyRouteMetadata(rb *restful.RouteBuilder, method string) *restful.RouteBuilder {
+	if r.routeMetadataProvider == nil {
+		return rb
+	}
+	for k, v := range r.routeMetadataProvider(r.descriptor, method) {
```
Automaat: **suggestion:** The provider runs synchronously in `NewApiServer`, so a panic in embedder code crashes control-plane startup, and nothing logs which keys were attached. Fail-fast at boot is defensible, but document the failure contract (or recover and log the descriptor/method) so an operator can diagnose it. Low priority: startup-only, bounded by route count.

pkg/api-server/resource_endpoints.go
```diff
@@ -84,12 +86,25 @@ func typeToLegacyOverviewPath(resourceType core_model.ResourceType) string {
...
+		rb = rb.Metadata(k, v)
```
Automaat: **suggestion:** Provider keys share a namespace with the auth-skip gate: `authn/skip.go:22` reads `route.Metadata()["auth"] == "skip"`. An embedder returning `{"auth":"skip"}` would disable auth on that route without meaning to. Untagged routes stay authenticated (fail-closed), so it is a footgun, not a defect. Document the reserved `auth` key or namespace embedder keys.

pkg/api-server/resource_endpoints.go
```diff
@@ -84,12 +86,25 @@ func typeToLegacyOverviewPath(resourceType core_model.ResourceType) string {
...
+	}
+	return rb
+}
+
 func (r *resourceEndpoints) addFindEndpoint(ws *restful.WebService, pathPrefix string) {
-	ws.Route(ws.GET(pathPrefix+"/{name}").To(r.findResource(false)).
+	ws.Route(r.applyRouteMetadata(ws.GET(pathPrefix+"/{name}").To(r.findResource(false)).
```
Automaat: **suggestion:** Each route names its HTTP verb twice, in `ws.GET(path)` and the trailing `http.MethodGet` argument, kept in sync by hand. Copy a block and change the verb in one place only, and the provider quietly gets the wrong method. A helper that owns the verb (`route(ws, method, path)`) would name it once.

pkg/core/runtime/builder.go
bartsmykla: Documented the single-provider / last-write-wins intent on the setter in 4fcb936.

pkg/core/runtime/runtime.go
bartsmykla: Tightened the type doc to state the CRUD-only scope and that the sub-routes are intentionally excluded (4fcb936).

pkg/api-server/resource_endpoints.go
bartsmykla: Documented the contract on the provider type: it runs synchronously while the API server is built, so a panic fails the boot by design (4fcb936).

pkg/api-server/resource_endpoints.go
bartsmykla: Good catch. Documented the shared metadata namespace and the reserved auth:skip key on the provider type - providers must not emit reserved keys (4fcb936). Happy to also hard-guard it in code if you'd prefer defense over docs.

pkg/api-server/resource_endpoints.go
bartsmykla: Extracted a route(ws, method, path) helper that owns the verb so it's named once; folded applyRouteMetadata into it (4fcb936).

Automaat (dismissed):

pkg/api-server/resource_endpoints.go
bartsmykla: Added the hard-guard pre-emptively in 51acddf: reserved keys (currently authn.MetadataAuthKey) are dropped before attaching provider metadata, with a log line and a test. Docs updated to match, so it's defense + docs rather than docs alone.

Automaat (approved):

PR #17367, opened 2026-07-14
test(unit): make unit tests cacheable on PRs

Automaat:
## Motivation

The unit `go test` embeds a **live `buildDate` timestamp + `gitCommit`** into
the test binary via `LD_FLAGS` (`mk/test.mk` → `mk/build.mk` → `version.sh`
`date -u`), so **every build is unique** and the Go test cache can never hit —
on top of `go clean -testcache` wiping it each run. Together these defeat unit
test caching entirely (an earlier experiment measured only 7/281 packages
caching — that was the timestamp confound, not real).

> Changelog: skip

## Implementation information

- `UNIT_LD_FLAGS` stamps **constant** values (`0.0.0` / `0000000` /
  `local-build`) instead of live ones — keeps `pkg/version`s `[build_check]`
  test passing while making binaries **stable build-to-build**.
- Skip `go clean -testcache` **on PRs** (master still cleans + runs fresh).
- Stable `go.sum`-keyed go-build/test cache.

### Measured (free `ubuntu-24.04`, warm run vs cold)

| | make test | packages cached |
|---|-----------|-----------------|
| cold | ~1351 s | 0 |
| **warm** | **~320 s (−76%, 4.2×)** | **193 / 280 (~69%)** |

Residual ~31% re-run: ~10 envtest/testcontainers dirs (correctly uncacheable) +
~77 with other untracked inputs.

## Notes

- Cost/feedback + fast-signal win (unit runs parallel to e2e), not the PR
  wall-clock pole once e2e is sharded.
- The `[build_check]` version assertions still pass (constant stamp is valid).
- (Ignore the trailing empty commit — measurement trigger; squashes away.)

## Supporting documentation

Buster trace: `version.sh:50` `date -u` → `LD_FLAGS buildDate` → per-build-unique binary.


.github/workflows/_test.yaml
```diff
@@ -20,6 +20,8 @@ jobs:
     timeout-minutes: 30
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci/skip-test') }}
     runs-on: ${{ vars.RUNS_ON_MASTER_AMD64 || vars.RUNS_ON_AMD64 || 'ubuntu-24.04' }}
+    env:
+      UNIT_CLEAN_TESTCACHE: ${{ github.event_name == 'pull_request' && 'false' || 'true' }}
```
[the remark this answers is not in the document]
Automaat: **Applied** — the reusable workflow now reads the caller's real event via a new `IS_PULL_REQUEST` input (set by `build-test-distribute.yaml` from `github.event_name`), instead of `github.event_name` inside `_test.yaml` which is always `workflow_call` (4c47758304).

_Generated by Sybra harness_

mk/test.mk
```diff
@@ -1,6 +1,8 @@
 UPDATE_GOLDEN_FILES ?=
 TEST_PKG_LIST ?= ./...
 REPORTS_DIR ?= build/reports
+UNIT_LD_FLAGS ?= -ldflags="-X github.com/kumahq/kuma/v3/pkg/version.version=0.0.0 -X github.com/kumahq/kuma/v3/pkg/version.gitTag=0000000 -X github.com/kumahq/kuma/v3/pkg/version.buildDate=local-build"
```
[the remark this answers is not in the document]
Automaat: **Applied** — added `-X github.com/kumahq/kuma/v3/pkg/version.Envoy=$(ENVOY_VERSION)` to `UNIT_LD_FLAGS` so unit tests stamp a stable, semver-valid Envoy version instead of the default "unknown", while still staying cacheable since `ENVOY_VERSION` only changes on dependency bumps (4c47758304).

_Generated by Sybra harness_

mk/test.mk
```diff
@@ -1,6 +1,8 @@
 UPDATE_GOLDEN_FILES ?=
 TEST_PKG_LIST ?= ./...
 REPORTS_DIR ?= build/reports
+UNIT_LD_FLAGS ?= -ldflags="-X github.com/kumahq/kuma/v3/pkg/version.version=0.0.0 -X github.com/kumahq/kuma/v3/pkg/version.gitTag=0000000 -X github.com/kumahq/kuma/v3/pkg/version.buildDate=local-build -X github.com/kumahq/kuma/v3/pkg/version.Envoy=$(ENVOY_VERSION)"
```
slonka: nitpick: gitCommit is not defined, it will be "unknown", the test will pass because the assertion is busted `ToNot(And(BeEmpty(), Equal("unknown")))` can never match (a value can't be empty *and* `unknown` at once) - could fix the assertion and add gitCommit here

.github/workflows/_test.yaml
```diff
@@ -31,6 +36,15 @@ jobs:
           version: 2026.5.7
         env:
           GITHUB_TOKEN: ${{ github.token }}
+      - name: Cache Go build
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: go-test-race-${{ runner.os }}-${{ hashFiles('**/go.sum') }}
```
slonka: key is only go.sum, so `actions/cache` won't re-save once it exists - the go-build cache freezes for the whole go.sum window and gets colder as source moves. for a PR whose point is cache hits, want a `-${{ github.sha }}` suffix on the key with layered restore-keys so it refreshes every run?

```
key: go-test-race-${{ runner.os }}-${{ hashFiles('**/go.sum') }}-${{ github.sha }}
restore-keys: |
  go-test-race-${{ runner.os }}-${{ hashFiles('**/go.sum') }}-
  go-test-race-${{ runner.os }}-
```

mk/test.mk
Automaat: **Applied** — added `-X ...pkg/version.gitCommit=0000000` to `UNIT_LD_FLAGS` so `gitCommit` is stamped (matching the existing `gitTag` pattern) instead of defaulting to `"unknown"`, and fixed the vacuous assertion in `pkg/version/version_test.go` from `ToNot(And(BeEmpty(), Equal("unknown")))` (never fails — a value cannot be both empty and `"unknown"` at once) to `ToNot(Or(BeEmpty(), Equal("unknown")))` (262a7dfcb2).

_Generated by Sybra harness_

.github/workflows/_test.yaml
Automaat: **Applied** — added a `${{ github.sha }}` suffix to the primary cache key with layered `restore-keys` (go.sum-scoped, then os-scoped), so the Go build cache is re-saved every run instead of freezing at the first exact `go.sum` match (262a7dfcb2).

_Generated by Sybra harness_

slonka (approved):

PR #17389, opened 2026-07-14
fix(xds): emit KRI tag on MeshService outbounds

lukidzi:
## Motivation

Enabling MeshService breaks MeshProxyPatch tag-based targeting of outbound listeners. Once an outbound is backed by a MeshService, the generated listener carries empty `io.kuma.tags` metadata because `TagsOrNil()` returns nil for real resource outbounds.

Closes #17369

## Implementation information

- Add `ListenerTags()` method to `Outbound` that returns a `kuma.io/kri` tag containing the KRI for real resource outbounds (MeshService, MeshExternalService, MeshMultiZoneService)
- Update `meshtcproute` and `meshhttproute` listener generation to use `ListenerTags()` instead of `TagsOrNil()`
- With this fix, MeshProxyPatch can match MeshService-backed outbound listeners using:
```yaml
  match:
    tags:
      kuma.io/kri: kri_msvc_<mesh>_<zone>_<ns>_<name>_<port>
```

I think it should be only on <= 2.14 as in 3.0 we require unified naming

### DIFF
--- a/pkg/core/xds/types/outbound.go
+++ b/pkg/core/xds/types/outbound.go
@@ -49,6 +49,24 @@ func (o *Outbound) TagsOrNil() map[string]string {
 	return nil
 }
 
+const KRITag = "kuma.io/kri"
+
+// ListenerTags returns tags for listener metadata. For legacy outbounds, it
+// returns the tags from the LegacyOutbound. For real resource outbounds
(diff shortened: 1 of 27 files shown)


lukidzi:
Not needed on master (3.0): unified naming is enabled by default there. Scoped to 2.13/2.14 instead:
- release-2.14: #17396
- release-2.13: #17397

PR #17391, opened 2026-07-14
fix(k8s): keep serving inbound when deduplicating

### LINKED ISSUE #12123: When I have multiple services I get multiple inbounds with the same port

### LINKED ISSUE #17142: MeshServices Exclusive mode: inbound dedup keeps Ignored inbound, MeshService goes Unavailable (Argo blue-green)

## What we saw

```yaml
networking:
  inbound:
  - port: 9000
    health: {}
    state: Ignored
    tags:
    name: grpc
```

## Why it happens

Walking the code:

## Reproducing it

lukidzi:
## Motivation

Fixes #17142.

In MeshServices `Exclusive` mode with inbound tags disabled, inbound deduplication by `address:port` (`deduplicateInboundsByAddressAndPort`) kept the *first* inbound and was state-blind.

This bites with Argo Rollouts blue-green plus `KUMA_RUNTIME_KUBERNETES_INJECTOR_IGNORED_SERVICE_SELECTOR_LABELS=rollouts-pod-template-hash`. That config makes `findMatchingServices` drop the hash when matching Services to a Pod, so both the active `<app>` and the `<app>-preview` Service match the same Pod. `inboundForService` then recomputes state against the *full* selector: the Service that actually selects the Pod produces a `Ready` inbound, the other produces an `Ignored` one, both on the same `address:port`.

Because Services are returned sorted by name, `<app>-preview` sorts before `<app>`, so dedup kept the `Ignored` inbound and dropped the `Ready` one. `GetHealthyInbounds()` then finds nothing, the backing MeshService reports `Healthy: 0` and flips to `Unavailable` while the Pods are healthy and serving.

## Implementation information

Deduplication now keeps the *most serving* inbound for each `address:port` (`Ready` > `NotReady` > `Ignored`) and replaces in place, so the resulting inbound order does not depend on which Services happen to select the Pod. This is safe because in this model inbounds carry no tags: which Services select the Pod is resolved by MeshService matching on Dataplane labels, not by the inbound itself, so collapsing duplicates loses no information and picking the serving one is strictly correct.

## Supporting documentation

- Issue: #17142
- Deduplication origin: #12760 (fixes #12123)

Tests: added a golden fixture reproducing the mid-rollout scenario (two Services on one port, `Ignored` sorting first). Confirmed the new golden fails on the pre-fix code and passes after. A companion fixture pins the tags-enabled path (both inbounds kept, unchanged by this PR).

### Known follow-up (not in this PR)

The tags-enabled path (no MeshIdentity) does not deduplicate, so the same Argo config leaves two inbounds on one `address:port`. Those collide on the Envoy inbound listener name (`inbound:<ip>:<port>`) and `ResourceSet.Add` is last-write-wins, so the surviving listener depends on Service name order rather than inbound state. That is a separate failure surface in the xDS generator and needs its own fix + verification; tracking it separately rather than widening this change.

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
+++ b/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
@@ -200,17 +200,53 @@ func (ic *InboundConverter) inboundInterfacesFor(ctx context.Context, zone strin
 	return ifaces, nil
 }
 
+// inboundStatePrecedence orders inbound states from the most to the least serving.
+// When several Services select the same address:port, that port is serving as long
+// as one of them is Ready, so a Ready inbound has to win deduplication over an
+// Ignored one. Ignored is the least serving: it only says that one particular
+// Service does not select this Pod, which is a membership fact rather than a
+// property of the port itself.
+func inboundStatePrecedence(state mesh_proto.Dataplane_Networking_Inbound_State) int {
+	switch state {
+	case mesh_proto.Dataplane_Networking_Inbound_Ready:
+		return 0
+	case mesh_proto.Dataplane_Networking_Inbound_NotReady:
+		return 1
+	default:
+		return 2
+	}
+}
+
+// deduplicateInboundsByAddressAndPort collapses inbounds that share an address and
+// port into one, keeping the most serving one. Inbounds here carry no tags, so the
+// only thing that distinguishes duplicates is their state and dropping the rest
+// loses no information: which Services select the Pod is resolved by MeshService
+// matching on Dataplane labels, not by the inbound.
+//
+// Keeping the most serving one matters when IgnoredServiceSelectorLabels is
+// configured: a Service is then matched with that label removed, so a Service that
+// does not select this Pod still produces an (Ignored) inbound on the same port as
+// the Service that does. Keeping the Ignored one would leave the Pod with no
+// healthy inbound and take its MeshService down. See #17142.
 func deduplicateInboundsByAddressAndPort(ifaces []*mesh_proto.Dataplane_Networking_Inbound) []*mesh_proto.Dataplane_Networking_Inbound {
 	inboundKey := func(iface *mesh_proto.Dataplane_Networking_Inbound) string {
 		return fmt.Sprintf("%s:%d", iface.Address, iface.Port)
 	}
 
 	var deduplicatedInbounds []*mesh_proto.Dataplane_Networking_Inbound
-	inboundsPerAddressPort := map[string]*mesh_proto.Dataplane_Networking_Inbound{}
+	indexPerAddressPort := map[string]int{}
 	for _, iface := range ifaces {
-		if inboundsPerAddressPort[inboundKey(iface)] == nil {
-			inboundsPerAddressPort[inboundKey(iface)] = iface
+		key := inboundKey(iface)
+		idx, exists := indexPerAddressPort[key]
+		if !exists {
+			indexPerAddressPort[key] = len(deduplicatedInbounds)
 			deduplicatedInbounds = append(deduplicatedInbounds, iface)
+			continue
+		}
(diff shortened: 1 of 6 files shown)


lukidzi:
We don't need this on master lets remove it https://github.com/kumahq/kuma/issues/17393

lukidzi:
Superseded by direct release-branch fixes: #17395 (release-2.13) and #17394 (release-2.14). Master is removing `IgnoredServiceSelectorLabels` and the `Ignored` state entirely under the MeshIdentity model (#17393), so the state-aware dedup patch is not needed there.

PR #17392, opened 2026-07-14
test(gatewayapi): pin conformance to mesh-ns fix

slonka:
## Motivation

The `gatewayapi` e2e job has failed on **every** master run since #17195 - deterministic, not flaky.

#17195 dropped the north-south Gateway conformance profile, so the suite no longer applies its base manifests (which deploy the `gateway-conformance-app-backend` / `gateway-conformance-web-backend` backends). But the mesh profile's `Setup` still waited for those two namespaces to have ready pods, so it blocked until the 300s namespace-readiness timeout and failed the job.

## Implementation information

Upstream removed `app-backend`/`web-backend` from the mesh readiness wait in https://github.com/kubernetes-sigs/gateway-api/pull/4934, leaving only `mesh`/`mesh-consumer`. That fix isn't in any release: `v1.6.0` predates the divergence and `release-1.6` wasn't backported - it exists only on `main`.

So pin the conformance module to that commit (`8f69037cd`). It still compiles against `gateway-api v1.6.0`, which stays unchanged, so production controllers are unaffected. One line in `go.mod`, no test-code change.

## Supporting documentation

- Upstream fix: https://github.com/kubernetes-sigs/gateway-api/pull/4934
- The pin is a pseudo-version (`v0.0.0-20260603141352-8f69037cd891`) on the `main` lineage, diverged from `v1.6.0` (behind ~30 commits), so the mesh tests run are early-June `main`'s set. API/CRDs stay `v1.6.0`, so schema-compatible.
- Follow-up: replace the pin with a real tag once a release (v1.6.1/v1.7.0) or a `release-1.6` backport includes the fix.

> Changelog: skip

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -97,7 +97,7 @@ require (
 	sigs.k8s.io/controller-tools v0.21.0
 	// When updating this also update version in: test/framework/k8s.go
 	sigs.k8s.io/gateway-api v1.6.0
-	sigs.k8s.io/gateway-api/conformance v1.6.0
+	sigs.k8s.io/gateway-api/conformance v0.0.0-20260603141352-8f69037cd891
 	sigs.k8s.io/yaml v1.6.0
 )
 
--- a/go.sum
+++ b/go.sum
@@ -691,8 +691,8 @@ sigs.k8s.io/controller-tools v0.21.0 h1:KXDQza3bgjlPY6xLR63tI/40gzjhyUAvkCrwzd2/
 sigs.k8s.io/controller-tools v0.21.0/go.mod h1:DLIypi3Q2+azVAP8jr/mHXJgveYYHFjhnNOUuBJ10JE=
 sigs.k8s.io/gateway-api v1.6.0 h1:735YBRj5NXFrOGX0GoSjwzUIzbz8kiEOfADsqHFmHgE=
 sigs.k8s.io/gateway-api v1.6.0/go.mod h1:FVfx3t389ybeXOqvDghLbdvJdSCfI/PReqCUI3lu3mY=
-sigs.k8s.io/gateway-api/conformance v1.6.0 h1:jyH7Jtx46hEeBQnmSErn2cbp0k4OmT3FLRBnu77c4Ko=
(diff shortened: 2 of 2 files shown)


test/e2e_env/gatewayapi/conformance_test.go
```diff
@@ -154,6 +154,17 @@ spec:
...
 
+	// The mesh conformance profile's Setup waits for the app-backend and
+	// web-backend namespaces to have ready pods (MeshNamespacesMustBeReady in
+	// the gateway-api suite hard-codes that list). Those backends are deployed
+	// only by the base manifests, which the suite applies just for the Gateway
+	// (north-south) profile. We only run the mesh profile, so apply the base
+	// manifests here to satisfy that wait - otherwise Setup blocks until the
+	// namespace-readiness timeout.
+	conformanceSuite.Applier.ManifestFS = options.ManifestFS
+	conformanceSuite.Applier.GatewayClass = options.GatewayClassName
```
slonka: The mechanism, exactly:

The conformance Applier, for every resource it applies, checks the kind (apply.go:203-204):
if uObj.GetKind() == "Gateway" {
    a.prepareGateway(t, &uObj)
}
and prepareGateway unconditionally overwrites spec.gatewayClassName with Applier.GatewayClass (apply.go:72):
unstructured.SetNestedField(uObj.Object, a.GatewayClass, "spec", "gatewayClassName")

So when we apply base/manifests.yaml (which has 4 Gateways), each one gets spec.gatewayClassName = Applier.GatewayClass. If that's "", they're stamped with an empty gatewayClassName, which the Gateway API CRD rejects (gatewayClassName is required, minLength 1) → MustApplyWithCleanup fails → our setup fails.

So the line exists only to stop the Applier from minting invalid Gateway objects out of a file we're applying for its backends. The value never gets used for anything - those Gateways reference the nonexistent kuma class and sit inert.

I can create an alternative for this.

Automaat (approved):

bartsmykla (approved):

Issue #17393, opened 2026-07-14
Remove IgnoredServiceSelectorLabels and the Ignored inbound state in the MeshIdentity model


lukidzi:
## Summary

Remove `KUMA_RUNTIME_KUBERNETES_INJECTOR_IGNORED_SERVICE_SELECTOR_LABELS` and the `Ignored` inbound state once the mesh runs on the MeshIdentity model (`Exclusive` MeshServices + inbound tags disabled). Both exist only to serve the legacy `kuma.io/service`-based identity model and become inert-at-best, harmful-at-worst under MeshIdentity.

## Background

[MADR 032](../blob/master/docs/madr/decisions/032-add-inbound-in-runtime.md) introduced `IgnoredServiceSelectorLabels` and the `Ignored` listener state to support Argo Rollouts blue-green **without an mTLS gap**. In the legacy model identity is `kuma.io/service`-based, so a selector flip creates a new `kuma.io/service` and therefore needs a new certificate. The `Ignored` listener pre-provisions that cert before the flip (MADR 032: *"It will however receive a cert with `kuma.io/service` of this listener"*).

## Why it is unnecessary under MeshIdentity

[MADR 085](../blob/master/docs/madr/decisions/085-meshservice-meshidentity-identity.md) derives the SpiffeID from `/ns/{namespace}/sa/{serviceAccount}` — **workload identity, not service**:

- A pod receives its certificate at **startup** from MeshIdentity (selected by Dataplane labels), independent of which Services select it.
- Both blue-green revisions share a ServiceAccount, so they present the **same** SpiffeID before and after promotion.
- A selector flip changes only **MeshService membership**, resolved by `DataplaneLabels` matching (which already includes the rollout hash) at the MeshService layer — not by any per-inbound state.

The certificate-timing race MADR 032 addressed is structurally impossible; blue-green works with zero configuration. All three of MADR 032's justifications (cert timing, outbound policy targeting, GUI status) dissolve in the tagless model.

## Why it is actively harmful today

[...]

PR #17394, opened 2026-07-14
fix(k8s): keep serving inbound when deduplicating

### LINKED ISSUE #17142: MeshServices Exclusive mode: inbound dedup keeps Ignored inbound, MeshService goes Unavailable (Argo blue-green)

lukidzi:
## Motivation

Fixes #17142 on `release-2.14`.

In `Exclusive` mode with inbound tags disabled, inbound deduplication by `address:port` kept the *first* inbound and was state-blind. With Argo Rollouts blue-green plus `KUMA_RUNTIME_KUBERNETES_INJECTOR_IGNORED_SERVICE_SELECTOR_LABELS=rollouts-pod-template-hash`, both the active `<app>` and `<app>-preview` Service match the same Pod. `inboundForService` recomputes state against the full selector, so one inbound is `Ready` and the other `Ignored`, both on the same `address:port`.

Services are sorted by name, so `<app>-preview` sorts before `<app>` and dedup kept the `Ignored` inbound, dropping the `Ready` one. `GetHealthyInbounds()` then found nothing and the backing MeshService flipped to `Unavailable` while Pods were healthy.

## Implementation information

Deduplication now keeps the most serving inbound per `address:port` (`Ready` > `NotReady` > `Ignored`), replacing in place so port order does not depend on which Services select the Pod. Safe because these inbounds carry no tags: MeshService membership is resolved on Dataplane labels, not the inbound, so collapsing duplicates loses no information.

## Supporting documentation

- Issue: #17142
- Dedup origin: #12760
- Direct fix to the release branch (not merged to master; the flag is being removed there, see #17393).

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
+++ b/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
@@ -200,17 +200,40 @@ func (ic *InboundConverter) inboundInterfacesFor(ctx context.Context, zone strin
 	return ifaces, nil
 }
 
+// inboundServingRank orders inbound states from the most to the least serving.
+func inboundServingRank(state mesh_proto.Dataplane_Networking_Inbound_State) int {
+	switch state {
+	case mesh_proto.Dataplane_Networking_Inbound_Ready:
+		return 0
+	case mesh_proto.Dataplane_Networking_Inbound_NotReady:
+		return 1
+	default:
+		return 2
+	}
+}
+
+// deduplicateInboundsByAddressAndPort collapses inbounds sharing an address:port
+// into the most serving one. With IgnoredServiceSelectorLabels set, a Service that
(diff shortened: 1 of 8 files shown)


lukidzi:
Closing — 2.14 doesn't need this. Since #17197, dedup runs only when `InboundTagsDisabled=true` (MeshIdentity model), where `IgnoredServiceSelectorLabels` is unnecessary (workload identity, no cert-timing gap) and the correct config is to not set it. The tags-enabled scenario on 2.14 no longer deduplicates (both inbounds kept) and its remaining failure is the Envoy listener collision, which this patch doesn't address. Flag removal tracked in #17393; the load-bearing fix stays on release-2.13 (#17395).

PR #17395, opened 2026-07-14
fix(k8s): include state in inbound dedup key

### LINKED ISSUE #17142: MeshServices Exclusive mode: inbound dedup keeps Ignored inbound, MeshService goes Unavailable (Argo blue-green)

lukidzi:
## Motivation

Fixes #17142 on `release-2.13`.

In `Exclusive` mode, inbounds are deduplicated by `address:port` and the first one wins. That key is state-blind. With Argo Rollouts blue-green plus `KUMA_RUNTIME_KUBERNETES_INJECTOR_IGNORED_SERVICE_SELECTOR_LABELS=rollouts-pod-template-hash`, both the active `<app>` and the `<app>-preview` Service match the same Pod. `inboundForService` recomputes state against the full selector, so one inbound comes out `Ready` and the other `Ignored` on the same `address:port`.

Services are sorted by name, so whichever Service sorts first decides the survivor. When that is the `Ignored` one, the `Ready` inbound is dropped, `GetHealthyInbounds()` then finds nothing, and the backing MeshService flips to `Unavailable` while the Pods are healthy.

## Implementation information

`state` is now part of the deduplication key, so two inbounds are only collapsed when they are genuinely interchangeable. Inbounds that differ only by state describe different things — one Service serves this Pod, another does not — and both are kept, so the `Ready` one always survives regardless of Service name ordering. Duplicates that share address, port and state still collapse to the first one, which is what #12760 set out to remove.

This deliberately allows two inbounds on one `address:port` when their states differ. Nothing rejects that: the Dataplane validator has no uniqueness check on `address:port`, and everything that reads inbounds for availability (`GetHealthyInbounds`, MeshService matching and status, `endpoints.go`, `vips_allocator`, insights) filters on state and picks the `Ready` one. The residual is in xDS: `InboundProxyGenerator` iterates every inbound without filtering by state, and listener/cluster names derive from `address:port`, so the two entries resolve to a single Envoy listener via last-write-wins in the `ResourceSet` rather than a conflict. Traffic on the port keeps flowing either way; only the listener's tag metadata depends on which inbound lands last. Filtering `Ignored` inbounds out of the xDS path is the cleaner follow-up, but it is a wider change than a release branch should carry.

## Supporting documentation

- Issue: #17142
- Dedup origin: #12760
- Companion release fix: #17394 (release-2.14). Direct fix to the release branch (not merged to master; the flag is being removed there, see #17393).

### DIFF
--- a/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
+++ b/pkg/plugins/runtime/k8s/controllers/inbound_converter.go
@@ -140,7 +140,7 @@ func (i *InboundConverter) LegacyInboundInterfacesFor(ctx context.Context, zone 
 	return i.inboundInterfacesFor(ctx, zone, pod, services)
 }
 
-// InboundInterfacesFor should be used when MeshService mode is Exclusive. This function deduplicates inbounds by address and port.
+// InboundInterfacesFor should be used when MeshService mode is Exclusive. This function deduplicates inbounds by address, port and state.
 // Since MeshService does not need tags we can safely deduplicate inbounds
 func (i *InboundConverter) InboundInterfacesFor(ctx context.Context, zone string, pod *kube_core.Pod, services []*kube_core.Service) ([]*mesh_proto.Dataplane_Networking_Inbound, error) {
 	inbounds, err := i.inboundInterfacesFor(ctx, zone, pod, services)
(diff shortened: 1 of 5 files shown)


pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -181,17 +181,40 @@ func (i *InboundConverter) inboundInterfacesFor(ctx context.Context, zone string
...
-			inboundsPerAddressPort[inboundKey(iface)] = iface
+		key := inboundKey(iface)
+		idx, exists := indexPerAddressPort[key]
+		if !exists {
+			indexPerAddressPort[key] = len(deduplicatedInbounds)
 			deduplicatedInbounds = append(deduplicatedInbounds, iface)
+			continue
+		}
+		// Replace in place to keep port order independent of Service state.
+		if inboundServingRank(iface.State) < inboundServingRank(deduplicatedInbounds[idx].State) {
```
Automaat: **praise:** Replacing in place at the first-seen index keeps inbound port order independent of Service state, avoiding Dataplane spec churn on reconcile — a subtle detail that's easy to miss.

pkg/plugins/runtime/k8s/controllers/pod_converter_test.go
```diff
@@ -333,6 +333,11 @@ var _ = Describe("PodToDataplane(..)", func() {
 			servicesForPod: "duplicated-inbounds.services-for-pod.yaml",
 			dataplane:      "duplicated-inbounds.dataplane.yaml",
 		}),
+		Entry("Deduplication keeps the Ready inbound when an Ignored one comes first", testCase{
```
Automaat: **praise:** The golden test is honest — it would still pass under the old keep-first logic only if the Ignored inbound survived, so it genuinely locks in the regression it claims to cover.

Automaat (dismissed):

lobkovilya (changes requested):
Do we even need to fix it? The user has to use MeshService Exclusive + Argo Rollouts + old mTLS to hit this issue. But even if they do all of that they still can update to 2.14 where the issue is solved.

If you still think we need to fix it in 2.13 then I think the fix should be simpler. Maybe just disable dedup at all? Introducing "inbound ranking" doesn't seem straightforward.

pkg/plugins/runtime/k8s/controllers/inbound_converter.go
```diff
@@ -181,16 +181,22 @@ func (i *InboundConverter) inboundInterfacesFor(ctx context.Context, zone string
 	return ifaces, nil
 }
 
+// deduplicateInboundsByAddressAndPort collapses inbounds that share an address,
+// a port and a state into the first one. Inbounds differing only by state are
+// kept, because state is not interchangeable: when a Service does not select the
```
lobkovilya: TBH the comment is a bit hard to understand, could you please just directly mention Argo Rollout problem and `KUMA_RUNTIME_KUBERNETES_INJECTOR_IGNORED_SERVICE_SELECTOR_LABELS`

lobkovilya (approved):

PR #17411, opened 2026-07-15
ci(e2e): cache preloaded third-party images

### LINKED ISSUE #17409: flaky: MeshIdentity Spire image preload fails pulling cgr.dev/chainguard/bash
## Flaky test

`MeshIdentity Spire [BeforeAll] should receive identity from Spire and traffic works`
in the **kubernetes** e2e suite intermittently fails on `master`.

## Signature

```
[BeforeAll] test/e2e_env/kubernetes/meshidentity/spire.go:46
[FAILED] Unexpected error:
  preload spire images: preload images: failed to pull on host:
  docker pull cgr.dev/chainguard/bash:latest@sha256:ef209fd7...
  unsuccessful after 5 retries
```

Image preload for the Spire setup fails pulling
`cgr.dev/chainguard/bash:latest` from the Chainguard registry. This is a
test-harness / infrastructure flake, not a product bug — but it recurs.

## Example run

https://github.com/kumahq/kuma/actions/runs/29345956641/job/87131824425

## Location

`test/e2e_env/kubernetes/meshidentity/spire.go:46` (BeforeAll)

## Suggested direction

Make image preload resilient: mirror/cache the image or pin to a digest served
from a reliable registry instead of pulling `:latest` from `cgr.dev` on every run.

Automaat:
## Motivation

The e2e framework preloads third-party images (spire, cnpg, coredns, ...) by
pulling them on the runner and importing them into the k3d/kind cluster, so
pods don't hit `ImagePullBackOff` at runtime. When an external registry is slow
or unreachable, that host-side pull exhausts its retries and fails suite setup.

This is the dominant Spire flake on `master` (#17409):

```
preload spire images: docker pull cgr.dev/chainguard/bash:latest@sha256:...
unsuccessful after 5 retries
```

Every run/shard pulls the same fixed set of images from external registries;
there is no caching, so any registry hiccup flakes the run.

> Changelog: ci(e2e): cache preloaded third-party images

## Implementation information

- `PreloadImages` (`test/framework/k8s_cluster.go`) now skips the `docker pull`
  when the image is already present locally, and records the images it
  preloads to the file named by `KUMA_E2E_PRELOAD_MANIFEST` (when set).
- The `run-e2e` composite action restores a `docker save` tarball into the
  docker daemon before the run and repopulates it afterwards via
  `actions/cache`, keyed on the pinned image versions under
  `test/framework/deployments/**`.

On a cache hit the external registries are not contacted at all; on a miss (or
after an image-version bump) the run pulls as before and repopulates the cache.
The cache stores only what the framework actually preloads, keeping it small
and independent of the locally-built kuma images.

## Supporting documentation

Fixes the flake tracked in #17409.


test/framework/k8s_cluster.go
```diff
@@ -1812,6 +1812,13 @@ func (c *K8sCluster) PreloadImages(images ...string) error {
 		wg.Add(1)
 		go func(i int, img string) {
 			defer wg.Done()
+			localRef := img
+			if tagged, _, ok := splitDigestRef(img); ok {
+				localRef = tagged
+			}
+			if exec.CommandContext(pullCtx, "docker", "image", "inspect", localRef).Run() == nil {
+				return
+			}
```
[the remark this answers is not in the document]
Automaat: Good catch — fixed in d4a05461db. Now: when the ref is digest-pinned, we skip only if the exact digest ref (`<repo>@sha256:<digest>`) is present locally (and (re)tag the bare `<repo>:<tag>` to it). A bare tag alone is trusted only under `KUMA_E2E_TRUST_CACHED_IMAGE_TAGS`, which is set exclusively on CI where the image store is a fresh cache we populated from a correctly-pinned pull. Local runs with a stale mutable tag (e.g. `:latest`) fall through to a normal pull.

.github/actions/run-e2e/action.yml
```diff
@@ -57,10 +57,26 @@ runs:
         }
         EOF
         sudo service docker restart
+    # Cache third-party images the e2e framework preloads (spire, cnpg, coredns, ...)
+    # so runs don't depend on external registries (cgr.dev/ghcr.io) being reachable.
+    - name: "Restore cached e2e images"
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+      with:
+        path: ${{ runner.temp }}/e2e-image-cache
+        key: e2e-images-${{ inputs.target }}-${{ inputs.arch }}-${{ inputs.k8s_version }}-${{ hashFiles('test/framework/deployments/**') }}
```
bartsmykla: Parallel runner IDs 0–3 share this immutable key, so a shard with no preloaded images can save an empty entry before the shard with the needed images. Include inputs.parallel_runner_id in the key or aggregate all shard manifests before saving; otherwise later runs restore the empty cache and keep pulling from external registries.

.github/actions/run-e2e/action.yml
```diff
@@ -57,10 +57,26 @@ runs:
...
+        tar="${{ runner.temp }}/e2e-image-cache/images.tar"
+        if [[ -f "$tar" ]]; then
+          docker load -i "$tar"
+        fi
     - name: "Run E2E tests"
       shell: bash
       env:
         KUMA_DEBUG: ${{ runner.debug == '1' }}
+        KUMA_E2E_PRELOAD_MANIFEST: ${{ runner.temp }}/e2e-preloaded-images.txt
+        KUMA_E2E_TRUST_CACHED_IMAGE_TAGS: "true"
```
bartsmykla: KUMA_E2E_TRUST_CACHED_IMAGE_TAGS is true even when no cache tar was loaded, so a reused self-hosted runner can accept a stale bare tag that does not match the requested digest. Enable this only after the matching cache tar loads successfully; otherwise the cluster can fall back to the registry.

test/framework/k8s_cluster.go
```diff
@@ -1806,12 +1806,24 @@ func (c *K8sCluster) PreloadImages(images ...string) error {
...
+	trustCachedTags := os.Getenv("KUMA_E2E_TRUST_CACHED_IMAGE_TAGS") == "true"
 	var wg sync.WaitGroup
 	pullErrs := make([]error, len(images))
 	for i, img := range images {
 		wg.Add(1)
 		go func(i int, img string) {
 			defer wg.Done()
+			if tagged, digestRef, ok := splitDigestRef(img); ok {
+				if exec.CommandContext(pullCtx, "docker", "image", "inspect", digestRef).Run() == nil {
+					_ = exec.CommandContext(pullCtx, "docker", "tag", digestRef, tagged).Run()
```
bartsmykla: When the exact digest exists but this docker tag fails, the goroutine returns success and the later import uses a missing or stale bare tag. Record this failure in pullErrs, as the post-pull tag path does, before returning.

bartsmykla (changes requested):

.github/actions/run-e2e/action.yml
Automaat: Applied in 4704e87218. The cache key now includes `inputs.parallel_runner_id`, so shards no longer race on one immutable entry and an empty save from one shard cannot poison the others.

_Generated by Sybra harness_

test/framework/k8s_cluster.go
Automaat: Applied in 4704e87218. The digest-fast-path now uses the same tagging helper as the post-pull path and records a failed `docker tag` in `pullErrs` instead of returning success with a missing or stale bare tag.

_Generated by Sybra harness_

.github/actions/run-e2e/action.yml
```diff
@@ -57,10 +57,27 @@ runs:
         }
         EOF
         sudo service docker restart
+    # Cache third-party images the e2e framework preloads (spire, cnpg, coredns, ...)
+    # so runs don't depend on external registries (cgr.dev/ghcr.io) being reachable.
+    - name: "Restore cached e2e images"
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+      with:
+        path: ${{ runner.temp }}/e2e-image-cache
+        key: e2e-images-${{ inputs.target }}-${{ inputs.arch }}-${{ inputs.k8s_version }}-${{ inputs.parallel_runner_id }}-${{ hashFiles('test/framework/deployments/**') }}
```
bartsmykla: Normal four-shard PR runs and ci/run-full-matrix one-shard runs share this runner-0 cache key even though runner 0 executes different specs. The one-shard run cannot add Spire images missing from the immutable four-shard entry, so it keeps pulling from the registry; include inputs.shards in the key.

Automaat: Applied in 5767749f7e. The cache key now includes `inputs.shards`, so a 4-shard PR run and a 1-shard full-matrix run no longer collide on runner 0 when they execute different spec sets.

_Generated by Sybra harness_

.github/actions/run-e2e/action.yml
```diff
@@ -57,10 +57,27 @@ runs:
         }
         EOF
         sudo service docker restart
+    # Cache third-party images the e2e framework preloads (spire, cnpg, coredns, ...)
+    # so runs don't depend on external registries (cgr.dev/ghcr.io) being reachable.
+    - name: "Restore cached e2e images"
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+      with:
+        path: ${{ runner.temp }}/e2e-image-cache
+        key: e2e-images-${{ inputs.target }}-${{ inputs.arch }}-${{ inputs.k8s_version }}-${{ inputs.shards }}-${{ inputs.parallel_runner_id }}-${{ hashFiles('test/framework/deployments/**') }}
```
bartsmykla: Moving a suite that calls PreloadImages to another job-* shard changes that shard's image set without changing this key, because the suite registration files are not hashed. That shard then restores an immutable empty or partial cache and pulls the missing images from the registry on every run, so include the shard-assignment and preload-caller sources in the hash.

Automaat: Applied in d8d66ea747d9. The cache key now hashes the e2e and e2e_env Go sources alongside `test/framework/deployments/**`, so both shard-label moves and test-code changes that start or stop using preloaded third-party images invalidate the shard cache instead of reusing an immutable stale tarball.

_Generated by Sybra harness_

bartsmykla (approved):

PR #17412, opened 2026-07-15
test(meshproxy): avoid cross-mesh workload clash

### LINKED ISSUE #17407: flaky: MeshProxy Migration drops traffic mid-migration (multizone e2e)
## Flaky test

`MeshProxy Migration` in the **multizone** e2e suite intermittently fails on `master`:

- `[It] should deploy zone proxies to meshproxymig-red with no interruptions for both meshes`
- `[It] should deploy zone proxies to meshproxymig-blue with no interruptions for both meshes`

Observed 3 times across recent `master` push runs (latest 2026-07-14 18:31 UTC).

## Signature

The `Consistently` block that asserts zero traffic interruption during migration fails:

```
[FAILED] The function passed to Consistently failed at
test/e2e_env/multizone/meshproxy/migration.go:334 with:
  request 1/5 failed after 4 successful responses:
  failed to unmarshal response: "": unexpected end of JSON input
```

A single request mid-migration receives an **empty response body**, so the
"no interruptions" invariant is violated. This is distinct from the SAN-order
non-determinism previously addressed — traffic is briefly dropped here.

## Example run

https://github.com/kumahq/kuma/actions/runs/29325489219/job/87061136531

## Location

`test/e2e_env/multizone/meshproxy/migration.go:334` (assertion), spec at `:160`

lukidzi:
## Motivation

The MeshProxy Migration multizone E2E deployed both meshes (meshproxymig-red and meshproxymig-blue) into a single meshproxy-migration namespace. Both the red and blue client/server pods run under the default ServiceAccount, and computeWorkloadName falls back to the ServiceAccount name when no workload label is set. So both meshes share workload default in the same namespace, which WorkloadReconciler explicitly rejects:

    ERROR controllers.Workload namespace has dataplanes in multiple meshes for same workload
      workload=default namespace=meshproxy-migration meshCount=2 error=multiple meshes detected

The reconciler logs this on every reconcile and skips Workload generation, so the test exercised an unsupported two-meshes / one-namespace / one-workload configuration.

## Implementation information

Split the shared namespace into per-mesh meshproxy-migration-red and meshproxy-migration-blue, threaded through a namespace field on trafficCheck and the deployMeshScopedZoneProxies helper. Each namespace now holds a single mesh's dataplanes, so workload default no longer collides across meshes and the test models a supported topology.

Scope is the workload/namespace collision only. The separate cross-zone request truncation seen during proxy rollout (a 200 with an empty body while old shared zone ingress/egress endpoints drain) is not addressed here.

## Supporting documentation

Follow-up to #17167, which unskipped this migration test.
Might fix: https://github.com/kumahq/kuma/issues/17407

> Changelog: skip

### DIFF
--- a/test/e2e_env/multizone/meshproxy/migration.go
+++ b/test/e2e_env/multizone/meshproxy/migration.go
@@ -27,7 +27,8 @@ import (
 )
 
 func Migration() {
-	namespace := "meshproxy-migration"
+	namespaceRed := "meshproxy-migration-red"
+	namespaceBlue := "meshproxy-migration-blue"
 
 	const (
 		redMeshName  = "meshproxymig-red"
@@ -45,6 +46,7 @@ func Migration() {
 		clientName        string
 		destinationServer string
 		expectedInstance  string
+		namespace         string
 	}
 
 	redTraffic := trafficCheck{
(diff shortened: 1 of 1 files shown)


slonka (approved):
Having these in separate namespaces makes sense 👍

PR #16030, opened 2026-03-27
fix(mlbs): migrate AffinityTags to use pod labels when inbound tags are disabled

mail2sudheerobbu-oss:
## Motivation

When `KUMA_EXPERIMENTAL_INBOUND_TAGS_DISABLED=true` is set, Kuma strips inbound tags from Dataplane resources to reduce memory overhead. However, `MeshLoadBalancingStrategy.LocalityAwareness.LocalZone.AffinityTags` relies on inbound tags to group and route traffic. This means that when inbound tags are disabled:

1. The local proxy cannot determine its own affinity group (tags are absent from its own Dataplane resource).
2. Remote endpoints cannot be matched for affinity routing because their filter metadata is empty (no tags were propagated).

This results in locality-aware load balancing via `AffinityTags` silently failing when `KUMA_EXPERIMENTAL_INBOUND_TAGS_DISABLED` is enabled.

Pod labels, however, are always available from Kubernetes pod metadata regardless of this flag. This PR migrates `AffinityTags` handling to fall back to pod labels when inbound tags are absent.

Closes #15995

## Implementation information

The fix spans the full data pipeline from endpoint construction through to locality group resolution:

1. **`pkg/core/xds/types.go`** — Added a `Labels map[string]string` field to the `Endpoint` struct to carry pod/workload labels alongside inbound tags throughout the routing pipeline.

2. **`pkg/xds/envoy/metadata/v3/metadata.go`** — Added a new `LbLabelsKey` constant (`"io.kuma.labels"`), a `EndpointMetadataWithLabels` function that stores both inbound tags (under `envoy.lb`) and pod labels (under `io.kuma.labels`) in Envoy filter metadata, and an `ExtractLbLabels` function to retrieve pod labels from metadata.

3. **`pkg/xds/envoy/endpoints/v3/endpoints.go`** — Updated `ToLocalityLbEndpoints` to call `EndpointMetadataWithLabels` instead of `EndpointMetadata`, so pod labels are embedded in each endpoint's Envoy filter metadata.

4. **`pkg/xds/topology/outbound.go`** — Populated the new `Labels` field in `fillDataplaneOutbounds` and `fillLocalMeshServices` from `GetMeta().GetLabels()` on the respective Dataplane/DPP resource.

[...]

PR #17424, opened 2026-07-15
docs(MADR): source listener tags from labels

### LINKED ISSUE #17381: MeshProxyPatch: enabling MeshService breaks listenerTags matching, name fallback not deterministic
## Summary

Enabling `MeshService` breaks `MeshProxyPatch` tag-based targeting of outbound listeners. Once an outbound is backed by a `MeshService`, the generated listener no longer carries `io.kuma.tags`, so `listenerTags` / `tags` matches never hit. The only remaining selector is `match.name`, but the legacy name embeds a dynamically-allocated address, so it is not deterministic and cannot be written by a user ahead of time.

## The problem

Before `MeshService`, an outbound listener carries the outbound's tags in `io.kuma.tags`, and a `MeshProxyPatch` can target it with `origin: outbound` + `listenerTags`/`tags`:

```yaml
default:
  appendModifications:
    - networkFilter:
        operation: Patch
        match:
          origin: outbound
          listenerTags:
            kuma.io/service: backend
        value: |
          ...
```

After enabling `MeshService`, the outbound listener is generated from the service resource. Its tags are taken from `svc.Outbound.TagsOrNil()`, which returns `nil` for a real `MeshService` outbound — only the legacy `kuma.io/service` outbound populates tags (`pkg/core/xds/types/outbound.go:43`, used at `pkg/plugins/policies/meshtcproute/plugin/v1alpha1/listeners.go:41`).

Result: the listener has **no `io.kuma.tags`**, so `MeshProxyPatch` `listenerTags`/`tags` matching silently stops working — an existing policy that matched by tags no longer applies.

## The fallback (`match.name`) is not deterministic

With tag matching gone, the only selector left is `match.name`. But when unified resource naming is not active, the name falls back to the address-based form:

```go
// pkg/plugins/policies/meshtcproute/plugin/v1alpha1/listeners.go:29-38
address := svc.Outbound.GetAddressWithFallback("127.0.0.1")
port := svc.Outbound.GetPort()
listenerName := envoy_names.GetOutboundListenerName(address, port) // "outbound:<address>:<port>"
if id, ok := svc.Outbound.AssociatedServiceResource(); ok && unifiedNaming {
    listenerName = id.String() // KRI — only when unifiedNaming is enabled
}
```

`GetOutboundListenerName` produces `outbound:<address>:<port>` (`pkg/xds/envoy/names/resource_names.go:54`). The `<address>` is the dynamically-allocated `MeshService` VIP (or the workload IP), not a value the user knows or can rely on across restarts. So `match.name` cannot be authored reliably:

```yaml
match:
  name: outbound:10.43.0.17:80   # VIP is dynamic / not knowable
```

Only when unified resource naming is enabled (`meshServices: Exclusive` + the DP `UnifiedResourceNaming` feature) does the name become a stable, address-independent KRI (`kri_msvc_<mesh>_<zone>_<ns>_<name>_<port>`, `pkg/core/kri/kri.go:28`), making `match.name` usable again.

## Impact

Turning on `MeshService` without unified resource naming leaves `MeshProxyPatch` with **no usable selector** for outbound listeners:

- `listenerTags` / `tags` — broken, because the listener carries no `io.kuma.tags`.
- `match.name` — non-deterministic, because it embeds the dynamic VIP/address.

This is a silent regression: policies that worked before enabling `MeshService` stop applying, with no error.

## Proposed resolution (options)

1. Propagate MeshService labels

## Environment

- `MeshService` enabled, unified resource naming **not** active (DP without the `UnifiedResourceNaming` feature and/or mesh not in `meshServices: Exclusive`)
- Component: `pkg/plugins/policies/meshproxypatch`, outbound listener generation in `pkg/plugins/policies/meshtcproute` / `meshhttproute`

lukidzi:
> Changelog: skip

## Motivation

`MeshProxyPatch` lets you pick a generated Envoy listener by `tags`/`listenerTags`, which match against the `io.kuma.tags` filter metadata Kuma puts on every listener. Those tags have always come from `Dataplane` tags. Kuma is moving identity out of tags and into labels, but the listener metadata still reads tags.

So the policy is accepted, it validates, and it matches nothing. This shows up in two separate setups that people usually turn on together:

- With `MeshService` on, outbounds are resource references and carry no tags, so outbound listener tags are empty. That is https://github.com/kumahq/kuma/issues/17381.
- With `InboundTagsDisabled`, inbound tags are stripped at the source, so inbound listener tags are empty. Same bug, other direction, and nothing fills the gap.

Closes https://github.com/kumahq/kuma/issues/17381 (design only, no code change in this PR).

## Implementation information

Adds MADR 105. The main body reads on its own, and a **Deep dive** section at the bottom holds the file and line references, so you can read the decision without wading through code.

The document walks three cases (`kuma.io/service`, which works; `MeshService`, where outbound tags go empty; `InboundTagsDisabled`, where inbound tags go empty) and lands on one rule: fill listener `io.kuma.tags` from labels wherever tags no longer exist. Outbound listeners take the labels of the destination `Mesh*Service` resource itself, not of the `Dataplane`s behind it, since the listener stands for the service and per-workload labels already reach Envoy on the endpoints. Inbound listeners take the proxy's own `Dataplane` labels. Both strip `kuma.io/mesh`, which the outbound path already does. No `MeshProxyPatch` API change; the fix is entirely in how we fill the metadata.

Three things from the investigation shaped it:

- `kuma.io/service` has been deleted, not misplaced. The consumer's outbound tags, the VIP path that fed them, and the destination's inbound tags are all gone, so no design keeps `listenerTags: {kuma.io/service: backend}` working. The question became what to give people instead.
- Reading identity from labels is already what the rest of the codebase does. `IdentifyingName` reads `kuma.io/workload`, the k8s `MeshService` selector uses `DataplaneLabels`, endpoints carry labels under `io.kuma.labels`. That work was done case by case and listener metadata got skipped, so this applies an existing pattern rather than adding a new one.
- Pod labels are copied onto the `Dataplane`'s metadata labels either way. `InboundTagsDisabled` only drops the second copy that lived in tags, which is why filling listener tags from labels is possible at all.
- **We already fixed this exact bug once.** `MeshLoadBalancingStrategy` affinity tags broke under `InboundTagsDisabled` in the same silent way, and https://github.com/kumahq/kuma/pull/16030 fixed it by reading Pod labels instead, shipping them to Envoy on endpoints under `io.kuma.labels`. That covered the endpoint path and stopped there. This MADR is the same fix for the listener path.

No flag gates it. The affected listeners have empty `io.kuma.tags` today, so the change can only add matches and never break one that works, and a `false` value would have no use.

**Backport shapes the design.** This is a silent regression in released versions, so the fix targets release-2.13 and release-2.14, and that is the main reason Option 4 wins over Option 3: it changes only how we fill generated metadata, so there is no API, CRD schema or stored resource to ship in a patch. The two halves don't backport the same way. The outbound half reaches 2.13 and 2.14. The inbound half is 2.14 and up, because `InboundTagsDisabled` doesn't exist on 2.13, so inbound tags are always populated there and there is nothing to fix. Group targeting by user label is also 2.14 and up, since `MeshServiceLabelPropagation` isn't on 2.13; `kuma.io/display-name` and `kuma.io/zone` work there, `team: payments` doesn't. The 2.13 cherry-pick will conflict where label computation moved packages, but the labels themselves are the same, so it is import paths to fix up.

Rejected options are written up with the reasoning: do nothing (needs a mesh-wide migration to patch one listener, can't do group targeting, does nothing for the inbound case), build a `kuma.io/service` tag (brings back something we deliberately removed, and the value wouldn't match anyway, so `kuma.io/service: '*'` would start matching while `kuma.io/service: backend` still failed), and a separate `io.kuma.labels` listener key with new API fields. That last one deserves a note, since we already have an `io.kuma.labels` on endpoints and copying it is the obvious move. The endpoint split exists because endpoint tags live under `envoy.lb` and `envoy.transport_socket_match`, which Envoy actually reads for LB subsets and TLS selection, so Pod labels could not go in there. `io.kuma.tags` is read by nothing but our own matcher, so that constraint doesn't exist and copying the split would be copying a workaround without its reason. `io.kuma.labels` is also an internal hand-off (it exists so MeshLoadBalancingStrategy can read labels back out of endpoints another generator built) rather than a selector users write against.

One known gap, written down rather than hidden: labels belong to a resource and listeners belong to a port, so the two listeners of a multi-port destination look identical to `listenerTags`. You still need `match.name` with unified naming for that. Left as an open topic.

## Supporting documentation

- Issue: https://github.com/kumahq/kuma/issues/17381
- Template: `docs/madr/decisions/000-template.md`
- Related: MADR 077/078/084 (unified Envoy resource naming), MADR 102 (computed labels for zone proxy listeners), MADR 103 (Universal MeshService generation)


docs/madr/decisions/105-source-listener-tags-from-labels.md
```diff
@@ -0,0 +1,568 @@
...
+
+## Implications for Kong Mesh
+
+None
+
+## Decision
+
+Listener `io.kuma.tags` is filled from labels wherever tags no longer exist.
+
+1. Outbound listeners will get `kuma.io/kri` label with value of the destination `Mesh*Service` to avoid more frequent configuration changes and potential higher memory usage.
```
Automaat: **issue:** Decision picks `kuma.io/kri` for the outbound listener, but every worked example in the body (lines 101-109, 141-147) recommends sub-option A (`kuma.io/display-name` + `k8s.kuma.io/namespace`). The doc never states which sub-option was chosen or reconciles the two, so a reader following the body writes a selector the Decision doesn't endorse.

_Generated by Sybra harness_

Automaat: **issue:** Point 1 calls `kuma.io/kri` a "label," but sub-option B is explicit it's "a tag we synthesize, not a label we store" (line 170). In a MADR built on the labels-vs-tags distinction, say "synthesized tag."

_Generated by Sybra harness_

docs/madr/decisions/105-source-listener-tags-from-labels.md
```diff
@@ -0,0 +1,568 @@
...
+```yaml
+match:
+  origin: outbound
+  listenerTags:
+    team: payments
+```
+
+* Good, because there is no API change. The selector people already use starts working again, and it is safe to ship as a patch.
+* Good, because it can only add matches. The affected listeners have empty tags today, so nothing that matches now can stop matching.
+* Good, because labels give you more to match on than the single `kuma.io/service` string ever did. Zone, origin, env and user labels all become usable, which is what makes group targeting possible.
```
Automaat, line 117: **question:** Group targeting (`team: payments`) is billed as "the main advantage Option 4 has over Option 1," and only sub-option A delivers it — a `kuma.io/kri` names one exact resource and can't match a set. If the outbound is `kuma.io/kri`-only, that advantage is lost for outbound; if the Decision means A+B together, state it.

_Generated by Sybra harness_

docs/madr/decisions/105-source-listener-tags-from-labels.md
```diff
@@ -0,0 +1,568 @@
...
+    kuma.io/display-name: backend
+    k8s.kuma.io/namespace: kuma-demo
+```
+
+* Good, because the keys are the ones people already read in `kubectl get` output and already write in `targetRef.labels`, so nothing new has to be learned.
+* Good, because it is the only one of the three that does group targeting. A selector can match a set, so `team: payments` or `kuma.io/zone: east` patches a group without naming its members.
+* Good, because it degrades gracefully. A missing key just doesn't match, and every listener still carries something.
+* Bad, because it cannot express kind or port, which is what the two collisions below are about.
+* Bad, because we copy whatever labels the resource happens to have. On Kubernetes that is every label on the `Service`. Which keys we should copy is left open, in the implementation notes.
+* Bad, because of potential listener/cluster change on label update if we not use fixed map.
```
Automaat: **nit:** "if we not use fixed map" is broken grammar, and this says "cluster change" though the doc states elsewhere (lines 59, 267) that clusters never carry `io.kuma.tags`.

_Generated by Sybra harness_

Automaat (commented):
Docs-only MADR (105) — thorough and accurate; I spot-checked the KRI format string, the `Outbound` type, the reconcile early-return, and `AllComputedLabels` (12 keys, 6 land on `Mesh*Service`) and they all match the code.

One real problem worth resolving before this guides implementation: the **Decision** section contradicts the body it rests on. It picks `kuma.io/kri` for the outbound listener, while every worked example (lines 101-109, 141-147) recommends sub-option A (`kuma.io/display-name` + `k8s.kuma.io/namespace`), and the doc never states which sub-option was chosen or whether it's A+B together. That also collides with group targeting, which the doc calls the main advantage over Option 1 but which only sub-option A delivers. Point 1 also calls `kuma.io/kri` a "label" when sub-option B is explicit it's a synthesized tag. Details inline — 0 blockers, 2 issues, 1 question, 1 nit.

The rejected-options reasoning, the `io.kuma.labels` "why we don't extend it" analysis, and the backport split are excellent. Leaving approval to a human.

_Generated by Sybra harness_

docs/madr/decisions/105-source-listener-tags-from-labels.md
```diff
@@ -0,0 +1,568 @@
...
+## Implications for Kong Mesh
+
+None
+
+## Decision
+
+Listener `io.kuma.tags` is filled from labels wherever tags no longer exist.
+
+1. Outbound listeners will get `kuma.io/kri` label with value of the destination `Mesh*Service` to avoid more frequent configuration changes and potential higher memory usage.
+2. Inbound listeners whose tags are empty take the labels of the proxy's own `Dataplane`. `InboundTagsDisabled` is how that happens on Kubernetes, but the rule is keyed on the tags being empty, not on the flag, because a hand-written Universal `Dataplane` can have tagless inbounds with the flag off.
```
lobkovilya: Dataplane labels are shared across all inbound listeners, so by using labels instead of tags, we lose the ability to target a specific listener, right?

If `MeshProxyPatch` can only be applied to either all inbounds or none of them, then populating `io.kuma.tags` on individual inbounds with Dataplane labels seems pointless, doesn’t it?

docs/madr/decisions/105-source-listener-tags-from-labels.md
```diff
@@ -0,0 +1,568 @@
...
+## Deep dive
+
+Everything below is the supporting detail for the sections above.
+
+### The `io.kuma.tags` channel
+
+`TagsKey = "io.kuma.tags"` is defined at `pkg/xds/envoy/metadata/v3/metadata.go:99`. One thing writes it, `TagsMetadataConfigurer` (`pkg/xds/envoy/listeners/v3/tags_metadata.go:15-27`):
+
+```go
+l.Metadata.FilterMetadata[envoy_metadata.TagsKey] = &structpb.Struct{
```
lobkovilya, line 246: Do we even need to keep doing this as `FilterMetadata` and actually send it to Envoy? If we need metadata to work with MeshProxyPatch then we can simply add `Labels` to the `Resource`: https://github.com/kumahq/kuma/blob/f1f490e6cc2eaab4a9455872fb4771055c8a62c1/pkg/core/xds/resource.go#L24-L30

So these labels never leaving the CP.

lobkovilya (commented):
I think populating `tags` with labels makes the change less explicit to users. It may not be obvious that they need to migrate their existing `MeshProxyPatch` policies. The naming is also confusing, since tags and labels are not interchangeable elsewhere.

I think a better option would be to add a `labels` field to the `MeshProxyPatch` policy.

docs/madr/decisions/105-source-listener-tags-from-labels.md
lukidzi: Yes, it's the same as KRI is stable

docs/madr/decisions/105-source-listener-tags-from-labels.md
lukidzi: I think we could do this for Listeners as I believe we don't use them in any case - but will confirm, the only filtermetadata we use is on endpoints. The problem i can see is how we are going to present to the user that it's possible to configure it as it's never exposed in the config, now you can check the config dump and just filter it

docs/madr/decisions/105-source-listener-tags-from-labels.md
lukidzi: You are right, I think `kuma.io/kri` on the listener could solve this as each entry has a port in the KRI

docs/madr/decisions/105-source-listener-tags-from-labels.md
```diff
@@ -0,0 +1,568 @@
...
+* The outbound half goes to 2.13 and 2.14. Everything it needs (`Outbound.TagsOrNil`, `DestinationService`, `GetServiceByKRI`) is on both.
+* The inbound half only goes to 2.14 and master. `InboundTagsDisabled` doesn't exist on 2.13, so inbound tags are always populated there and there is nothing to fix.
+
+One practical note for whoever does the cherry-pick: label computation moved packages between 2.13 and 2.14, from `core_model.ComputeLabels` in `pkg/core/resources/model/resource.go` to `pkg/core/resources/labels`. The 2.13 pick will conflict there. The computed labels themselves are the same, so it is import paths and call shape to fix up, and the behaviour stays put.
+
+Because the tags a listener carries depend on which labels that version computes, a policy written for 2.14 may not match on 2.13. That comes with backporting a fix whose payload is "whatever labels this version has", and it belongs in the release notes so people don't hit it cluster by cluster.
+
+## Implications for Kong Mesh
+
+None
```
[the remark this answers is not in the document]
lukidzi: Renamed the section to `## Implications for Downstream Projects`, matching MADR 103.

docs/madr/decisions/105-source-listener-tags-from-labels.md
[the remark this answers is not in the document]
lukidzi: Fixed. The Decision now consistently says `kuma.io/kri` **tag** and adds an explicit line: it's synthesized and written straight into `io.kuma.tags`, never stored as a resource label.

docs/madr/decisions/105-source-listener-tags-from-labels.md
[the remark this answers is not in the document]
lukidzi: Rephrased: 'copying the resource's whole label set regenerates the listener on every label update, unless we copy only a fixed set of keys we choose.'

docs/madr/decisions/105-source-listener-tags-from-labels.md
lukidzi: Reconciled in the Decision: it now explicitly picks **sub-option B** (`kuma.io/kri`) for both outbound and inbound, and flags the `kuma.io/display-name` examples in the body as sub-option A illustrations, not the endorsed selector.

lukidzi: Fixed. Decision now calls `kuma.io/kri` a synthesized tag, not a label.

docs/madr/decisions/105-source-listener-tags-from-labels.md
lukidzi: Made explicit in the Decision: sub-option B (`kuma.io/kri`) is chosen for both directions. Group targeting via user labels is a sub-option A capability, now stated as out of scope and left as the open question in the implementation notes. KRI wins here because it's stable and its section name distinguishes per-port listeners.

docs/madr/decisions/105-source-listener-tags-from-labels.md
lukidzi: Fixed both: rephrased the grammar, and dropped the 'cluster' reference — the line now notes clusters never carry `io.kuma.tags`.

lukidzi:
> I think populating `tags` with labels makes the change less explicit to users. It may not be obvious that they need to migrate their existing `MeshProxyPatch` policies. The naming is also confusing, since tags and labels are not interchangeable elsewhere.
> 
> I think a better option would be to add a `labels` field to the `MeshProxyPatch` policy.

That is an option(3.0) but we need to keep in mind that the change needs to be backported, I didn't want to introduce api change as a patch release.

docs/madr/decisions/105-source-listener-tags-from-labels.md
lukidzi: Metadata on listener used by:
* metadata can be used in access log format
* MeshProxyPatch
* So far not used but it can be used in CEL

Endpoints metadata:
* MeshLoadBalancingStrategy for routing - we introduced new field which maybe makes sense to move to envoy.lb
* I think we use metadata for new scope zone proxy on migration: https://github.com/lukidzi/kuma/blob/eb90dc377dd495566f1ce2ef39d93be29ae60f8a/pkg/xds/envoy/clusters/v3/upstream_tls_context.go#L42-L44

Automaat (approved):

slonka (approved):

PR #17431, opened 2026-07-15
fix(ServiceInsight): keep delegated gateways when meshServices.mode is Exclusive

### LINKED ISSUE #17430: ServiceInsight drops delegated gateways when meshServices.mode is Exclusive
## Problem

With `meshServices.mode: Exclusive`, `ServiceInsight` is computed with an empty `Services` map, so **delegated gateways disappear from the GUI**. Delegated gateways are never turned into `MeshService` resources, so nothing replaces them — they should show up regardless of the mode.

## Cause

`createOrUpdateServiceInsight` (`pkg/insights/resyncer.go`) short-circuits in Exclusive mode:

```go
if exclusiveMeshServices {
    return r.upsertServiceInsight(ctx, mesh, insight) // empty Services
}
```

Added in #16921 to zero the `kuma.io/service` service counter once `MeshService` takes over. Correct for inbounds/external services, but too broad — it also drops `gateway_delegated` entries. The MeshService controller explicitly ignores gateway Pods, so delegated gateways have no `MeshService` replacement.

## Expected

In Exclusive mode, `ServiceInsight` still contains `gateway_delegated` entries (status/counts/zones as in other modes). Internal/external services stay excluded; the `MeshInsight.Services` counter from #16921 is unchanged.

Related: #16913, #16921

lobkovilya:
## Motivation

When `meshServices.mode: Exclusive`, `ServiceInsight` was computed with an empty `Services` map, so delegated gateways disappeared from the GUI. Delegated gateways are never turned into `MeshService` resources (the MeshService controller explicitly ignores gateway Pods), so nothing replaces them — they must be reported regardless of the mode.

## Implementation information

- Remove the Exclusive-mode short-circuit in `createOrUpdateServiceInsight`.
- Report only `gateway_delegated` entries in Exclusive mode; skip inbound (internal) and external services, which are replaced by `MeshService`/`MeshExternalService`.
- `MeshInsight.Services` counter behavior is unchanged (stays nil in Exclusive mode).

## Supporting documentation

Fixes #17430

### DIFF
--- a/pkg/insights/resyncer.go
+++ b/pkg/insights/resyncer.go
@@ -478,10 +478,6 @@ func (r *resyncer) createOrUpdateServiceInsight(
 	insight := &mesh_proto.ServiceInsight{
 		Services: map[string]*mesh_proto.ServiceInsight_Service{},
 	}
-	if exclusiveMeshServices {
-		// set ServiceInsight with empty Services map
-		return r.upsertServiceInsight(ctx, mesh, insight)
-	}
 
 	zonesMap := map[string]map[string]struct{}{}
 	addSvcToZones := func(svc, zone string) {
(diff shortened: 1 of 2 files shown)


Automaat (approved):
Reviewed the ServiceInsight resyncer change and the updated coverage. The new filtering preserves delegated gateways in Exclusive mesh service mode while continuing to omit internal services, builtin gateways, and external services that are represented elsewhere. The focused insights package test passes locally: `go test ./pkg/insights`.

_Generated by Sybra harness_

PR #17445, opened 2026-07-16
ci(build-publish): assert built binary reports tag version

### LINKED ISSUE #17234: ci: assert built binary reports tag version on tag build
The only thing the tag build changes vs the green branch SHA is the ldflags-injected version (`X.Y.Z` instead of `X.Y.Z-preview.<hash>`). Verify exactly that.

- Assert the freshly-built binary reports the tag version (e.g. `make build/info/version` / `kuma-cp --version` == tag).
- Container-structure tests on the freshly-built image already run (`_build_publish.yaml:134`); keep them.

These two are the only tag-specific checks — nothing the green branch run already covered.

Automaat:
## Motivation

Tag builds inject a release version (X.Y.Z) via ldflags, whereas branch builds use X.Y.Z-preview.<hash>. We need to verify this version injection is working correctly on release builds.

## Implementation information

Added an assertion in the build pipeline to verify the freshly-built binary reports the correct tag version (e.g., via `make build/info/version` or `kuma-cp --version`). The existing container-structure tests on the built image are preserved. This is the only tag-specific validation needed—the green branch CI already covers general correctness.

Closes https://github.com/kumahq/kuma/issues/17234

_Generated by Sybra harness_

### DIFF
--- a/.github/workflows/_build_publish.yaml
+++ b/.github/workflows/_build_publish.yaml
@@ -66,6 +66,10 @@ jobs:
         name: Image tag
         run: |
           echo "::notice title=Image tag::$(make build/info/version)"
+      - name: Assert built binary reports tag version
+        if: ${{ github.ref_type == 'tag' }}
+        run: |
+          make build/assert-tag-version
       - run: |
           make -j build/distributions
       - id: inspect-binary-output
--- a/.github/workflows/_build_publish.yaml
+++ b/.github/workflows/_build_publish.yaml
@@ -44,27 +44,30 @@ env:
   MISE_DISABLE_TOOLS: "golangci-lint,skaffold"
 jobs:
   assert-tag-version:
-    # Runs before build-binaries/build-images/publish-helm so a tag whose
-    # binary reports the wrong version can't slip through to a publish step
-    # in a job that doesn't depend on build-binaries.
-    if: ${{ github.ref_type == 'tag' }}
+    # Tag check lives on the steps, not the job, so this job always reports
+    # success/failure and never "skipped". That lets build-binaries and
+    # build-images depend on it with a plain `needs:` and GitHub's default
+    # job-status semantics, which correctly cascade to digest-images and
+    # publish-helm further down the chain and correctly stop on cancellation.
     timeout-minutes: 10
     runs-on: ${{ vars.RUNS_ON_MASTER_AMD64 || vars.RUNS_ON_AMD64 || 'ubuntu-24.04' }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        if: ${{ github.ref_type == 'tag' }}
         with:
           persist-credentials: false
           fetch-depth: 0
       - uses: jdx/mise-action@e6a8b3978addb5a52f2b4cd9d91eafa7f0ab959d # v4.2.0
+        if: ${{ github.ref_type == 'tag' }}
         with:
           # TODO: drop once https://github.com/jdx/mise/discussions/9889 is fixed (mise v2026.5.8 broke cosign public-key bundle verification).
           version: 2026.5.7
       - name: Assert built binary reports tag version
+        if: ${{ github.ref_type == 'tag' }}
         run: |
           make build/assert-tag-version
   build-binaries:
     needs: [assert-tag-version]
-    if: ${{ always() && (needs.assert-tag-version.result == 'success' || needs.assert-tag-version.result == 'skipped') }}
     timeout-minutes: 70
     runs-on: ${{ vars.RUNS_ON_MASTER_AMD64 || vars.RUNS_ON_AMD64 || 'ubuntu-24.04' }}
     permissions:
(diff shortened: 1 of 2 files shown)


mk/build.mk
```diff
@@ -76,6 +76,15 @@ build/info/short:
 build/info/version:
 	@echo $(BUILD_INFO_VERSION)
 
+.PHONY: build/assert-tag-version
+build/assert-tag-version: ## Dev: Assert the built kuma-cp binary reports the tag version
```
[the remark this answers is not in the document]
Automaat: Good catch, fixed — `build/assert-tag-version` now depends on `build/kuma-cp` so it builds the binary itself if run standalone or if the step order changes.

_Generated by Sybra harness_

.github/workflows/_build_publish.yaml
```diff
@@ -66,6 +66,10 @@ jobs:
         name: Image tag
         run: |
           echo "::notice title=Image tag::$(make build/info/version)"
+      - name: Assert built binary reports tag version
+        if: ${{ github.ref_type == 'tag' }}
+        run: |
+          make build/assert-tag-version
```
bartsmykla: On a tag where this assertion fails, `build-images` still runs independently and can push its matrix images, while `publish-helm` depends only on that job. The workflow can therefore publish images and the chart before turning red. Make this assertion an upstream dependency of every publishing job, or run the same check before each push.

bartsmykla (changes requested):

.github/workflows/_build_publish.yaml
Automaat: Good catch — confirmed. `build-images` and `publish-helm` had no dependency on `build-binaries`, so the assertion running there couldn't actually block a bad-version release.

Fixed by extracting the check into its own `assert-tag-version` job that runs first on tag pushes (no-op/skipped on branch pushes), and making `build-binaries` and `build-images` `needs` it with a skip-safe `if` so a failed assertion skips both (and `publish-helm`/`digest-images` cascade-skip via their existing `needs: [build-images]`).

_Generated by Sybra harness_

.github/workflows/_build_publish.yaml
```diff
@@ -43,7 +43,28 @@ env:
...
+      - uses: jdx/mise-action@e6a8b3978addb5a52f2b4cd9d91eafa7f0ab959d # v4.2.0
+        with:
+          # TODO: drop once https://github.com/jdx/mise/discussions/9889 is fixed (mise v2026.5.8 broke cosign public-key bundle verification).
+          version: 2026.5.7
+      - name: Assert built binary reports tag version
+        run: |
+          make build/assert-tag-version
   build-binaries:
+    needs: [assert-tag-version]
+    if: ${{ always() && (needs.assert-tag-version.result == 'success' || needs.assert-tag-version.result == 'skipped') }}
```
bartsmykla: If this run is canceled after the gate completes, `always()` remains true, so both publisher jobs can keep running and reach their Pulp or Docker push steps despite the cancellation. Use `!cancelled()` with the existing success/skipped result check on both jobs.

.github/workflows/_build_publish.yaml
```diff
@@ -43,7 +43,28 @@ env:
   GITHUB_TOKEN: ${{ github.token }}
   MISE_DISABLE_TOOLS: "golangci-lint,skaffold"
 jobs:
+  assert-tag-version:
+    # Runs before build-binaries/build-images/publish-helm so a tag whose
+    # binary reports the wrong version can't slip through to a publish step
+    # in a job that doesn't depend on build-binaries.
+    if: ${{ github.ref_type == 'tag' }}
```
bartsmykla: On branch pushes this job is skipped, and GitHub propagates that skip through the full `needs` chain. `build-images` overrides the immediate skip, but `digest-images` and `publish-helm` do not, so those required jobs are skipped and the next tag fails `release_sha_gate`. Make this job succeed as a no-op on non-tags (for example, put the tag condition on its steps) or add skip-safe conditions to every descendant.

.github/workflows/_build_publish.yaml
Automaat: Good catch — fixed by removing the custom `if:` on `build-binaries`/`build-images` entirely instead of patching it to `!cancelled()`. `assert-tag-version` now gates the tag check on its steps (not the job), so the job always reports `success`/`failure` and never `skipped`. `build-binaries`/`build-images` go back to the plain default `needs: [assert-tag-version]` with no custom `if`, which correctly stops the chain on cancellation without any extra handling.

_Generated by Sybra harness_

.github/workflows/_build_publish.yaml
Automaat: Confirmed the gap and fixed it — but by removing the skip source rather than adding more skip-safe conditions. `assert-tag-version`'s tag check now lives on its steps, not the job `if:`, so the job always completes with `success`/`failure` and is never `skipped` on branch pushes. That means `build-binaries`/`build-images` (and transitively `digest-images`/`publish-helm`) can rely on GitHub's plain default `needs:` gating again — no per-descendant skip-safe condition needed, and no risk of `release_sha_gate` seeing a false skip on a branch push.

_Generated by Sybra harness_

bartsmykla (approved):

PR #17453, opened 2026-07-16
fix(meshhttproute): allow MeshGateway target

Automaat:
## Motivation

> Changelog: fix(meshhttproute): allow MeshGateway target

Master CI is red: the multizone `kindIpv6` e2e job fails in
`Advanced LocalityAwareness with MeshLoadBalancingStrategy with Gateway`
(`locality_aware_multizone_hybrid_gateway.go`) with `install yaml
resource unsuccessful after 30 retries`.

A same-day refactor restricted every policy's top-level
`spec.targetRef.kind` to `Mesh`/`Dataplane`, dropping `MeshGateway`
from MeshHTTPRoute's allowlist along with several selector-style kinds
that were genuinely being deprecated. A sibling PR landed ~20 minutes
later replacing the legacy `MeshGatewayRoute` fallback with a
`MeshHTTPRoute` that targets `kind: MeshGateway` at the top level —
exactly the shape that was just rejected, so every attempt to install
that resource in the e2e test fails validation and the `BeforeAll`
retries out.

## Implementation information

Top-level `targetRef.kind: MeshGateway` is not vestigial: the shared
policy matcher (`DppSelectedByPolicy` in
`pkg/plugins/policies/core/matchers/dataplane.go`) has a dedicated,
still-live case for it that resolves the referenced `MeshGateway`
resource and scopes the policy to that gateway's listeners, and a
non-validation xDS-generation unit test builds this exact shape
directly against the Go struct (bypassing API validation), proving the
runtime path is real. It was also previously covered by
`DescribeValidCases` entries ("MeshGateway to Mesh allowed" /
"MeshGateway with hostnames allowed") that the same-day refactor
turned into rejection cases.

This PR restores `common_api.MeshGateway` to MeshHTTPRoute's
`validateTop` allowlist and reverts the corresponding validation test
cases back to their previous (valid) shape. It does not touch the
other ~12 policies the refactor also updated, or revisit the
intentional removal of `MeshSubset`/`MeshService`/`MeshServiceSubset`
as top-level kinds — that broader restriction stands; only the
`MeshGateway` regression is reverted here.

## Supporting documentation

### DIFF
--- a/pkg/plugins/policies/meshhttproute/api/v1alpha1/validation.go
+++ b/pkg/plugins/policies/meshhttproute/api/v1alpha1/validation.go
@@ -34,6 +34,7 @@ func (r *MeshHTTPRouteResource) validateTop(targetRef *common_api.TargetRef) val
 		return mesh.ValidateTargetRef(*targetRef, &mesh.ValidateTargetRefOpts{
 			SupportedKinds: []common_api.TargetRefKind{
 				common_api.Mesh,
+				common_api.MeshGateway,
 				common_api.Dataplane,
 			},
 			GatewayListenerTagsAllowed: true,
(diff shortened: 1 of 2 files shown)


Automaat:
Wrong fix — restoring MeshGateway as a top-level targetRef reintroduces something we deliberately removed (builtin gateway support is being dropped). Replaced by a PR that deletes the now-broken builtin-gateway e2e test instead.

PR #17456, opened 2026-07-16
fix(kds): attribute zone-to-global synced resources by the connecting zone's client-id

bartsmykla:
## Motivation

On the global control plane, KDS zone-to-global sync derives a synced resource's zone attribution from values the sending zone provides in-band: the `ControlPlane.Identifier` on the `DeltaDiscoveryResponse` (used for the `kuma.io/zone` label, origin, `ZoneIngress`/`ZoneEgress` `Spec.Zone`, the name prefix and the owning `Zone`) and the `kuma.io/zone` tags carried inside the specs (`ZoneIngress` `AvailableServices` and `Dataplane` inbound/gateway tags, consumed downstream for cross-zone endpoint locality and the global service inventory). Those in-band values can diverge from the `client-id` the zone connects with, so attribution is not guaranteed to stay consistent with the connecting peer.

## Implementation information

Attribute all of it by the connecting peer's declared identity (the `client-id` from the KDS stream) instead of the sender-provided in-band values:

- `UpstreamResponse.ControlPlaneId` is taken from the stream's `client-id` rather than the in-band `ControlPlane.Identifier`, so the top-level attribution (`kuma.io/zone` label, origin, `Spec.Zone`, name prefix and owning `Zone`) follows the connecting zone.
- The global ingest callback resolves the `kuma.io/zone` tag inside synced `ZoneIngress` `AvailableServices` and `Dataplane` inbound/gateway specs to that same `client-id`.

The `client-id` is request metadata the peer declares on the stream. Verifying it per connection is the (enterprise) zone-token filter's responsibility, not this path's. In ordinary zone-to-global sync these values already equal the zone name, so the change is a no-op for valid traffic. When a synced payload does carry a differing value, the correction is counted (`kds_zone_attribution_rewrites_total`, labelled by resource type) and logged with the resource and replaced value.

Adds a regression test covering attribution of the `kuma.io/zone` label, `ZoneIngress.Spec.Zone`, `ZoneIngress` `AvailableServices` tags, and `Dataplane` inbound and gateway tags.

### DIFF
--- a/pkg/kds/v2/client/cross_zone_attribution_test.go
+++ b/pkg/kds/v2/client/cross_zone_attribution_test.go
@@ -1,0 +1,281 @@
+// KDS zone-to-global sync attribution guard.
+//
+// The global CP must attribute a synced resource by the authenticated
+// connection identity (the client-id from util.ClientIDFromIncomingCtx, wired
+// into the stream in pkg/kds/mux/zone_sync.go), not by values the sender
+// provides in-band. These tests drive the real global ingest wiring
+// (NewDeltaKDSStream(stream, <authenticated zone>, ...) -> GlobalSyncCallback ->
+// KDSSyncClient.Receive) with a stream whose authenticated client-id ("zone-b")
+// differs from the in-band ControlPlane.Identifier and the in-spec kuma.io/zone
+// tags ("zone-a"), and assert that all attribution resolves to the authenticated
+// client-id.
+package client_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	envoy_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoy_sd "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	. "github.com/onsi/gomega"
+	"google.golang.org/grpc/metadata"
+
+	mesh_proto "github.com/kumahq/kuma/v3/api/mesh/v1alpha1"
+	system_proto "github.com/kumahq/kuma/v3/api/system/v1alpha1"
+	"github.com/kumahq/kuma/v3/pkg/core"
+	core_mesh "github.com/kumahq/kuma/v3/pkg/core/resources/apis/mesh"
+	"github.com/kumahq/kuma/v3/pkg/core/resources/apis/system"
+	core_model "github.com/kumahq/kuma/v3/pkg/core/resources/model"
+	"github.com/kumahq/kuma/v3/pkg/core/resources/store"
+	"github.com/kumahq/kuma/v3/pkg/kds"
+	client_v2 "github.com/kumahq/kuma/v3/pkg/kds/v2/client"
+	sync_store_v2 "github.com/kumahq/kuma/v3/pkg/kds/v2/store"
+	core_metrics "github.com/kumahq/kuma/v3/pkg/metrics"
+	"github.com/kumahq/kuma/v3/pkg/plugins/resources/memory"
+	test_grpc "github.com/kumahq/kuma/v3/pkg/test/grpc"
+	util_proto "github.com/kumahq/kuma/v3/pkg/util/proto"
+)
+
+const (
+	// authenticatedZone is the connection's authenticated identity (client-id),
+	// which zone_sync.go passes to NewDeltaKDSStream and which must drive
+	// attribution.
+	authenticatedZone = "zone-b"
+	// otherZone is a different, already-enrolled zone named by the sender's
+	// in-band ControlPlane.Identifier and in-spec kuma.io/zone tags.
+	otherZone = "zone-a"
+)
+
+// newGlobalSink wires the global-CP KDS zone-to-global ingest as
+// pkg/kds/mux/zone_sync.go does, feeding a memory-backed store. authClientID is
+// the authenticated zone (client-id), intentionally different from the
+// per-message ControlPlane.Identifier and in-spec tags the responses carry.
+func newGlobalSink(t *testing.T, ctx context.Context, authClientID string, typ core_model.ResourceType) (store.ResourceStore, *test_grpc.MockDeltaClientStream, func()) {
+	t.Helper()
+	g := NewWithT(t)
+
+	globalStore := memory.NewStore()
+	// Both zones exist on global, as in any federated mesh.
+	for _, z := range []string{otherZone, authenticatedZone} {
+		err := globalStore.Create(ctx,
+			&system.ZoneResource{Spec: &system_proto.Zone{Enabled: util_proto.Bool(true)}},
+			store.CreateByKey(z, core_model.NoMesh))
+		g.Expect(err).ToNot(HaveOccurred())
+	}
+
+	metrics, err := core_metrics.NewMetrics("")
+	g.Expect(err).ToNot(HaveOccurred())
+	syncer, err := sync_store_v2.NewResourceSyncer(core.Log.WithName("crosszone-syncer"), globalStore, store.NoTransactions{}, metrics, context.Background())
+	g.Expect(err).ToNot(HaveOccurred())
+
+	clientStream := test_grpc.NewMockDeltaClientStream()
+	// authClientID is the authenticated zone that must drive attribution, not
+	// the in-band ControlPlane.Identifier.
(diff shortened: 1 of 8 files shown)


pkg/kds/v2/store/sync.go
```diff
@@ -383,15 +383,33 @@ func GlobalSyncCallback(
...
 
+			// Resolve zone attribution to the authenticated zone
+			// (upstream.ControlPlaneId). Besides the top-level Spec.Zone, zone
+			// tags carried inside the specs are consumed downstream - the tag in
+			// ZoneIngress AvailableServices drives cross-zone endpoint locality
+			// (fillIngressOutbounds) and Dataplane inbound/gateway tags feed the
+			// global service inventory (createOrUpdateServiceInsight) - so these
+			// must stay consistent with the authenticated zone too.
 			switch upstream.Type {
 			case core_mesh.ZoneIngressType:
```
Automaat, line 395: **suggestion:** `pinZoneTag` and the `Spec.Zone` overwrites silently correct client-id/zone mismatches, with no log or metric when a rewrite fires. An operator debugging a post-upgrade `kuma.io/zone` label change has nothing to grep for. Add a debug log or counter on the rewrite branch.

_Generated by Sybra harness_

Automaat (commented):
This is a legitimate, well-tested security hardening fix that closes a real in-band spoofing gap in KDS zone-to-global sync, and the regression test proves the old code would have failed it. No blocking issues across eight independent review passes plus an adversarial check.

One thing worth calling out before merge: the PR frames this as attributing by "authenticated zone," but in OSS the client-id backing that attribution is a self-asserted gRPC metadata value, not a cert-verified identity (no ClientAuth on the KDS mux server, empty auth interceptor slots by default). The code change is still correct and worth shipping, but the description should be explicit that this narrows in-band spoofing rather than adding cryptographic zone authentication, so nobody downstream reads it as closing the full trust boundary. A few smaller non-blocking notes: no changelog entry for deployments where the identifiers currently diverge, no log/metric when a zone-attribution rewrite actually fires, and a couple of minor consistency nits (see inline comment and full review).

0 blocking, 3 non-blocking suggestions, 1 question, 1 thought, 1 nit.

_Generated by Sybra harness_

pkg/kds/v2/client/cross_zone_attribution_test.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// zoneIngressWithDivergentZoneTags returns a ZoneIngress whose AvailableServices
+// carry a kuma.io/zone tag naming a zone different from the connection's
+// authenticated zone. That tag drives cross-zone endpoint locality in
+// fillIngressOutbounds, so global ingest must resolve it to the authenticated
+// zone.
+func zoneIngressWithDivergentZoneTags() *mesh_proto.ZoneIngress {
+	return &mesh_proto.ZoneIngress{
+		Zone: authenticatedZone,
```
Automaat: **issue:** The test can't catch a regression: the fixture sets `Zone: authenticatedZone` on the wire payload, so `stored.Spec.GetZone()` matches even without the fix. Set the fixture's `Zone` to `otherZone` so the assertion depends on it.

pkg/kds/v2/client/stream.go
```diff
@@ -229,7 +229,13 @@ func (s *stream) Receive() (UpstreamResponse, error) {
...
 	}
 	return UpstreamResponse{
-		ControlPlaneId:      resp.GetControlPlane().GetIdentifier(),
+		// Attribute the batch to the authenticated peer (s.clientID), not the
+		// in-band ControlPlane.Identifier the sender provides. On the global
+		// ingest path s.clientID is the zone from util.ClientIDFromIncomingCtx
+		// (see pkg/kds/mux/zone_sync.go); in legitimate zone-to-global sync it
+		// already equals the sender's identifier, so relying on it keeps
+		// attribution consistent with the authenticated zone.
+		ControlPlaneId:      s.clientID,
```
Automaat, line 238: **issue:** "Authenticated zone" overstates the guarantee: `client-id` is self-declared metadata with no client-cert check or bound `ServerStreamInterceptors`, so any zone can present another zone's id. Soften the framing, or confirm an enterprise interceptor enforces this.

pkg/kds/v2/store/sync.go
```diff
@@ -409,6 +427,17 @@ func GlobalSyncCallback(
 	}
 }
 
+// pinZoneTag rewrites a kuma.io/zone tag to the authenticated zone. It only
```
Automaat: **suggestion:** `pinZoneTag` silently overwrites a mismatched zone tag, so there's no way to tell "already matched" from "corrected a real divergence," the signal that matters most for a spoofing fix. Add a debug log or counter when the values differ.

pkg/test/kds/setup/client.go
```diff
@@ -9,9 +9,17 @@ import (
...
 
-func StartDeltaClient(clientStreams []*grpc.MockDeltaClientStream, resourceTypes []model.ResourceType, stopCh chan struct{}, cb *kds_client_v2.Callbacks) {
+// StartDeltaClient starts a KDS sync client per stream. clientIDs[i] is the
+// authenticated peer identity for clientStreams[i]; in production this is the
+// zone name from util.ClientIDFromIncomingCtx and it drives resource
+// attribution on the global ingest path. If clientIDs is shorter than
+// clientStreams the remaining streams fall back to "client-<i>".
+func StartDeltaClient(clientStreams []*grpc.MockDeltaClientStream, clientIDs []string, resourceTypes []model.ResourceType, stopCh chan struct{}, cb *kds_client_v2.Callbacks) {
 	for i := range clientStreams {
 		clientID := fmt.Sprintf("client-%d", i)
```
Automaat: **suggestion:** `StartDeltaClient`'s new fallback (`client-%d` when `clientIDs` runs short) has no caller that exercises it; the only real caller always passes a full-length slice. Delete it, or add a test that hits it.

pkg/kds/v2/store/sync.go
bartsmykla: Added kds_zone_attribution_rewrites_total, a counter labelled only by resource type (no zone, name or value labels, to keep cardinality bounded), plus a warning log carrying the resource, client-id and replaced value. Both fire only when a rewrite replaces a differing value, so ordinary sync stays silent.

pkg/kds/v2/client/cross_zone_attribution_test.go
bartsmykla: Good catch. The fixture now sets Spec.Zone to a different zone, so the assertion fails without the fix and passes with it - verified fail-before/pass-after.

pkg/kds/v2/client/stream.go
bartsmykla: Reworded to describe it as the connecting peer's declared client-id, with per-connection verification left to the enterprise zone-token filter (empty in OSS). It no longer reads as an authenticated identity.

pkg/kds/v2/store/sync.go
bartsmykla: The counter and log now fire only when a rewrite actually replaces a differing value; a value that already matches records nothing, so a corrected divergence is distinguishable from a plain match.

pkg/test/kds/setup/client.go
bartsmykla: Removed it - clientIDs is now required for every stream, and the only caller already passes one per stream.

Automaat (approved):
Reviewed the final PR head including the follow-up commit. The attribution source is now consistently the connecting peer's client-id across top-level metadata and in-spec zone tags, the new regression coverage exercises the real ingest path with divergent payload values, and the added rewrite counter/logging closes the earlier observability gap. I did not find a remaining blocking correctness issue in the final patch set.

_Generated by Sybra harness_

Automaat (approved):
Reviewed the current head commit `163009ffda09363f6f98a25584dc80cbcac9384b`. I did not find a remaining blocking correctness issue. The final patch consistently derives zone attribution from the connecting client-id on the zone-to-global path, preserves the existing global-to-zone behavior, and adds focused regression coverage plus low-cardinality rewrite observability for the divergent-value case.

_Generated by Sybra harness_

